PR #10, opened 2016-06-24
Hack for handling CONOPT warning message

qtothec:
This is related to an old [forum post](https://groups.google.com/d/msg/pyomo-forum/rsf4Iqy-fRE/W7adnQWNBQAJ) regarding CONOPT returning a status code 100 for nonlinear models for which it finds a local optimum.
It's a fragile solution that would be temporary until there's a better way to resolve the miscommunication. I could make it slightly less fragile by using a regex for the CONOPT version number, but this pull request is more just to open a discussion on how to accommodate the CONOPT behavior.


ghackebeil:
I think the the pull request looks fine, as in it wouldn't break the SOL parser for any other solvers. However, I don't think I'm for adding this hack to the SOL parser. I think this is technically a bug in CONOPT, and the SOL parser should return the information that is found in the SOL file. I'm not sure why CONOPT is associating finding a locally optimal solution with a status number that is supposed to represent "optimal solution indicated, **but error likely**" ([source](http://www.ampl.com/NEW/statuses.html)).

It sounds like the CONOPT folks were responsive about the empty NL file rows issue. Did they have any comment on this status behavior? Any chance of a fix in the future? Also, was the workaround I posted on the forum effective?

Perhaps a better place for this kind of solver specific workaround would be inside a CONOPT solver plugin. It could be a trivial plugin (e.g., a simple subclass of the ASL plugin) that checks for this case and makes a simple modification to the results object status. Then again, that's almost too trivial of a reason to add a new file to Pyomo.

I'm on the fence. The results object is supposed to reflect information that is in the solution file returned by the solver. Any modification to that information before the user has a chance to inspect it makes it difficult to debug certain case, and is just adding more implicit behavior to Pyomo. I'm probably overthinking this, and should just let this hack go in (but it would be nice if CONOPT fixed things).

Thoughts anyone?

qtothec:
The response that I had received from Arne was:

> The solver status 100 is an AMPL code and the AMPL people have decided that locally optimal solutions should have code 100. If you send a convex model to CONOPT, CONOPT does not know that the model is convex (there is nothing in the AMPL interface that can be used to define convexity) so it cannot return ‘Globally Optimal’. If Pyomo knows that the model is convex it could add a translation layer on the way back from AMPL/CONOPT and convert 100 to 0. But does Pyomo really know or is it just something you know?? If the Pyomo interface thinks that 100 is really bad and produces unwanted messages then this is not really something that is related to CONOPT.

Arne also implied in a later conversation that stewardship of the AMPL interface with CONOPT resided with the AMPL developers.

ghackebeil:
Maybe we can bring up the issue on the AMPL help forum and they would fix it there. Or is there a website or GitHub page specifically for ASL where we can report the issue?

I know of at least one other nonlinear solver (Ipopt) that does not know about global optimality (I don't have a license to test others, like knitro), and when it converges during normal execution (locally optimal) the status code is less than 100 because the solver executed normally and an error was not likely.

Maybe our interpretation of those status numbers is wrong and we need to just not make that 100-199 status range produce a warning.

ghackebeil:
I came across some documentation for the AMPL CONOPT executable ([here](http://ampl.com/SOLVERS/README.conopt.txt)). Here is an excerpt:

> Here is a table of solve_result_num values that "conopt" can return
> to an AMPL session, along with the text that appears in the associated
> solve_message.  Values >= 510 indicate a bug of some kind.
> 
>         Value   Message
>         0       Optimal.
>         100     Locally optimal.
>         101     Bogus "Termination by solver" (CONOPT bug).
>         ...

In what cases would the solver result num ever be set to 0? Perhaps it is when AMPL can prove convexity (such as when it is an LP)? I started a thread on the AMPL help forum, but I don't think it has been approved yet. Maybe we can get them to change the locally optimal case to 1 (or even 99).

qtothec:
In my limited testing, giving CONOPT an LP will yield a result num of 0.

ghackebeil:
That makes sense.

Given that we have a very specific table describing the status numbers for CONOPT (that I linked to in the previous comment), I think that is further justification of moving this hack into a CONOPT specific plugin (that maybe can make further use of the rest of those status descriptions), which modifies the results object the way you're currently doing in the SOL reader patch. It would be a very short plugin that probably inherits from the ASL plugin. The hard part of figuring out which method to overload (but I think one exists that gets passed the results object at some point).

qtothec:
That sounds good. I'm not familiar enough with the solver plugin system architecture to do that, so it may be prudent to close this pull request and move it into issues?

whart222:
FWIW, I agree that we should not accept this PR, but instead have a CONOPT-specific solver interface.

ghackebeil:
@qtothec: Let us know if the commit referenced above fixes the issue.

PR #33, opened 2016-09-04
Fix undefined mutable params

jsiirola:
This change started with @ghackebeil observing that the following seemingly innocuous example generates an exception:

```
from pyomo.environ import *

model = ConcreteModel()

model.p_index = Set(initialize=list(range(100)))
model.p = Param(model.p_index, mutable=True)
model.e = Expression(
    expr= sum(1 - model.p[i] for i in model.p_index))
```

His original argument was that as a _mutable_ `Param`, it should be OK to implicitly declare the Param without providing any initial value (leaving it to the user to set the value some time later before sending the model to the writer/solver).  He also pointed out that this kind of behavior _is_ supported for Simple Params, so it should work for Indexed Params.

Originally, I agreed that this was a "no brainer" change, but then fell back to how we explain "Abstract vs Concrete Models": with abstract models, we parse the model first and then load the data into it; and with concrete models, all the data is present when the model is constructed.  This change request is effectively asking for a third mode where not all of the the necessary data is present when the model is constructed.

Is this a change we want to make?

### DIFF
--- a/pyomo/core/base/param.py
+++ b/pyomo/core/base/param.py
@@ -397,11 +397,11 @@ class Param(IndexedComponent):
         # Set the parameter
         #
         if self._mutable:
-            if idx is None:
-                self._raw_setitem(idx, val)
-            else:
+            if self.is_indexed():
                 self._data[idx] = _ParamData(self, val)
                 #self._raw_setitem(idx, _ParamData(self, val), True)
+            else:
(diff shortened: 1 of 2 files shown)


ghackebeil:
Are we waiting on anything in particular to merge this? I think someone just needs to click a button. Should I?

PR #37, opened 2016-09-08
Make .name a property method

ghackebeil:
This PR includes three major changes:

- Renames the current `name` method to `getname`.
- Creates a `name` property method that returns the fully qualified name.
- Creates a `local_name` property method that returns the shortened name (only the immediate parent is taken into account).

Additionally, a setter has been defined for the `name` property method that raises an exception telling users that `.name` is no longer a public writable attribute. When I did the original `cname` -> `name` method rename, I encountered many examples where model names were being set by setting this public attribute. I changed those examples to use the `name` constructor keyword instead.

### DIFF
--- a/doc/GettingStarted/current/examples/block_iter_example.py
+++ b/doc/GettingStarted/current/examples/block_iter_example.py
@@ -26,9 +26,9 @@ model.pprint()
 
 # @compprintloop:
 for v in model.component_objects(Var):
-    print("FOUND VAR:" + v.name(True))
+    print("FOUND VAR:" + v.name)
     v.pprint()
 
 for v_data in model.component_data_objects(Var):
-    print("Found: "+v_data.name(True)+", value = "+str(value(v_data)))
+    print("Found: "+v_data.name+", value = "+str(value(v_data)))
 # @:compprintloop
--- a/examples/dae/run_Path_Constraint.py
+++ b/examples/dae/run_Path_Constraint.py
@@ -31,8 +31,8 @@ def plotter(subplot, x, *y, **kwds):
         if kwds.get('points', False):
             plt.plot(list(x), [value(_y[t]) for t in x], 'o')
     plt.title(kwds.get('title',''))
-    plt.legend(tuple(_y.name() for _y in y))
-    plt.xlabel(x.name())
+    plt.legend(tuple(_y.name for _y in y))
+    plt.xlabel(x.name)
 
 import matplotlib.pyplot as plt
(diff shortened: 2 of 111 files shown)


ghackebeil:
@jsiirola, @whart222, @blnicho:

I've compiled a list of files where I'm seeing `.local_name` (i.e., `.name()`, before the changes) being used in a way that might lead to problems with block-based models. To be more clear, I'm simply saying my understanding of the code (or the time bandwidth) was not sufficient to say for sure that it was _not_ a problem.

I just wanted to give each of you a heads up and suggest that you take look at these files in your spare time to see if the use of `.local_name` is correct. For instance, there were cases in pyomo.bilevel (I think) where the use of a list of variable `id()` values might be more efficient (and correct) for block-based models than using `.local_name`. 

Here is the list of files:
- pyomo/bilevel/plugins/lcp.py
- pyomo/bilevel/plugins/transform.py
- pyomo/core/base/connector.py
- pyomo/core/base/DataPortal.py
- pyomo/core/base/PyomoModel.py
- pyomo/core/data/TableData.py
- pyomo/core/plugins/data/csv_table.py
- pyomo/core/plugins/data/db_table.py
- pyomo/core/plugins/data/text.py
- pyomo/core/plugins/data/xml_table.py
- pyomo/core/plugins/transform/dual_transformation.py
- pyomo/core/plugins/transform/nonnegative_transform.py
- pyomo/dae/diffvar.py
- pyomo/dae/integral.py
- pyomo/dae/misc.py
- pyomo/dae/plugins/colloc.py
- pyomo/gdp/plugins/bigm.py
- pyomo/gdp/plugins/chull.py
- pyomo/repn/collect.py

jsiirola:
I do not like the blanket refusal to have a publicly settable name attribute.  For example, I think it should be OK to set the name of a model after it is constructed.  The intent in disallowing the name to be set is that the name is dictated by the Block it is attached to -- so I think I would rather see the `Component.name` setter be the following:

```python
@name.setter
def name(self, val):
    if self.parent_block() is None:
        self._name = val
    else
        raise ValueError(
            "The .name attribute (for component %s) is not settable when "
            "the component is assigned to a Block." % (self.name,))
```

ghackebeil:
@jsiirola: That seems reasonable. Feel free to make that change.

ghackebeil:
@jsiirola: I still see some use of .name as a method in connector.py and param.py. Can you take care of these?

jsiirola:
Yes.  I will get this cleaned up in a couple hours.

Issue #30, opened 2016-08-24
is_fixed checks inside constraint.set_expr need to be changed


ghackebeil:
The checks for is_fixed when setting a constraint expression need to be changed to something more restrictive. However, is_constant is too restrictive. We need another expression classifier (is_numeric?) that checks for the existence of optimization variables, or even the possibility of them appearing in the future (e.g., because a mutable expression appears somewhere). Here is some pseudo code of a problematic case:

``` !python
y = variable()
x = variable()
expr = x == y
x.fix(1.0)
c = constraint(expr)
x.free()
```

Now c.lower and c.upper are equal to x (which are always supposed to be "numeric"), but there is nothing invalid about how this constraint was defined. It is simply a side-effect of our invalid code breaking apart the constraint expression in a way that doesn't make sense.

jsiirola:
Isn't the correct solution to finally fix `Constraint.add()` so that it doesn't manipulate the expression that the user hands it?

That is, we defer all decisions of what the lower/upper bounds of a constraint are until the writer.  This becomes doubly nice after we rework the solver interface so that the solver owns the writer instance: since the writer will be the one manipulating the expressions, it will be able to inform the solver how to interpret (and map) things like duals back into the original model context.

ghackebeil:
I agree that your solution would be better in the longer term. However, that also seems like a MUCH more extensive change (it would dramatically affect constraint.py, canonical_repn.py, and all of the solver plugins), leading me to doubt it will get done anytime soon.

I have use for the kind of expression classification that I mentioned outside of constraint.add, and this seems like a more realistic fix to a very long standing problem. I think it could be done in a day if we can decide on a name for this class of expressions.

ghackebeil:
Fixed in #42

PR #42, opened 2016-09-10
is_data expression classification

ghackebeil:
This PR adds a new method called `_is_data` to the NumericValue base class and defines a module level function in numvalue.py called `is_data`. The `is_data` check returns `True` when an expression does not reference any decision variables or any mutable subexpressions that can be updated to reference decision variables (e.g., the Expression object). This PR also replaces the use of `is_fixed` with `is_data` when setting a constraint expression, as will be explained below.

Example usage of `is_data`:

``` python
import pyomo.environ as aml
from pyomo.core.base.numvalue import is_data

model = aml.ConcreteModel()
model.q = aml.Param(initialize=1.0)
model.p = aml.Param(mutable=True)
model.v = aml.Var()
model.e = aml.Expression()

print(is_data(1.0))      # True
print(is_data(model.p))  # True
print(is_data(model.q))  # True
print(is_data(model.v))  # False
print(is_data(model.e))  # False

# This check also returns True for None and string types
print(is_data(None)) # True
print(is_data('a'))  # True
```
- Where is this used?
  - Setting constraint bounds (changed from `is_fixed`)
  - I have use for this outside of pyomo.core, so I am fine modifying this PR to exclude any changes to how constraint expressions are set.
- How is this different from `is_fixed` and `is_constant`?
  - Unlike `is_fixed` and `is_constant`, this check does not require objects to be set to numeric values (e.g., mutable parameters). For instance, in the expression `x**p` both `is_fixed` and `is_constant` check if `p == 0`, which results in an exception if `p` references mutable parameters or variables that do not have a value. Note that these checks are perfectly valid at the time of writing the model, so I don’t think we should change them.
  - One can identify a _data_ expression simply by looking for variables or mutable expressions. The `is_data` check for these two objects is always `False`, and unlike `is_fixed`, it is a static property.
  - In the context of constraint bounds, the `is_fixed` check, although valid, can result in bounds being set to an expression that can become unfixed at a later time, resulting in disastrous behavior for our solver interfaces that generate a canonical representation based on the constraint body.
- What does this fix?
  - Fixes ticket #30, which highlights how a fixed variable can get moved to a constraint bound when setting the constraint expression, and result in an invalid model when the variable is unfixed (**even though the user did nothing invalid**). @jsiirola  has pointed out other ways to fix this problem, but they require large changes to constraint expression storage, canonical representation generation, and all of the solver plugins. His fixes are definitely something to consider in the future, but this relatively small change fixes the problem now. 
- What does this break?
  - Can not use Expression objects in the lower, body, and upper slots of a double sided inequality expression being stored in a constraint (same as the behavior for variables). E.g.

``` python
model = ConcreteModel()
model.e1 = Expression()
model.e2 = Expression()
model.e3 = Expression()
model.c = Constraint(expr= model.e1 <= model.e2 <= model.e3) # -> ValueError
```
- Why not use this check for variable bounds?
  - All of our solver interfaces currently validate that the lower and upper bounds pass the `is_fixed` check, and there are no intermediate steps, such as generation of a canonical representation, that would require the non-fixed elements from these bounds.
  - Note that the `is_fixed` checks currently in use by the variable bound setters can probably be relaxed (left out). They can also result in failures to set certain valid expressions just because referenced data elements have not been set to a value. I am not entirely opposed to using the `is_data` checks here in place of `is_fixed`, but we could just as easily delay any checks until the model is passed to a solver.


ghackebeil:
We might also consider an alternate Expression type, perhaps called DataExpression, that returns true for `_is_data` and always verifies that the sub-expression it stores passes the `is_data` check. Then again, this might be good functionality to place on a mutable parameter.

ghackebeil:
Notes from discussion about this PR:
  - Suggested rename to `_is_structurally_constant`
  - A reason to distinguish a `DataExpresssion` object from a mutable Param is that failing to do so would change the semantics of assignment to mutable Params, for instance, when the expression being assigned involves other mutable Params. Currently, expressions like this would be evaluated at the time they are assigned. However, this change would have the Params reference an expression whose value could be updated indirectly by changing values of mutable Params (or Expressions) that it references. Example:
```python
model = ConcreteModel()
model.p = Param(mutable=True, initialize=1.0)
model.q = Param(mutable=True)
model.q = 2.0*model.p
print(value(model.q))    # -> 2.0
model.p = 0.0
# current behavior
print(value(model.q))    # -> 2.0
```
However, if mutable Params stored expressions, then `model.q` would evaluate to 0.0 in the final print statement.

jsiirola:
Working on pyomo4 expression generation, I realized another use case for `_is_structurally_constant`: when we are generating `_LinearExpression`s, we check that the two arguments of a multiplication are a `_VarData` and something that is *not* a `_VarData` or `is_expression()`.  We would like to remove the `is_expression()` restriction so that `(m.p + 1)*m.x` would generate a`_LinearExpression`.  

The correct test here should be `isinstance(a, _VarData) and b._is_structurally_constant()`.  Given that assertion, then it is critical that `Expression`s return `_is_structurally_constant() == False`

As a side note: I think that the current Param behavior is correct and that it is a bad idea for Params to store expressions as their value...  If you want a named object that can store an expression, you should be using an `Expression`.

ghackebeil:
Here is why I don't like `_is_structurally_constant()`: We could implement a `DataExpression` (name is not important as long as you understand what I am trying to capture) that is restricted to storing "var-less" expressions, which does not preserve structural const-ness but should still fall into the class of expressions that you are checking for in the above case. That is, your check should be changed to:
```python
isinstance(a, _VarData) and b._never_allowed_to_be_a_variable_and_never_allowed_be_an_expression_that_reference_variables
```

If we can come up with a more concise name for `_never_allowed_to_be_a_variable_and_never_allowed_be_an_expression_that_reference_variables`, then we are done. `_is_data` was my first attempt at that. Maybe `_restricted_to_no_variables` is better.

To rephrase my above point, I don't see this as being related to the structural const-ness of an expression (non-public mutability, as I'm calling it). I see this as identifying a class of expressions that are not allowed to reference variables. It just so happens that we don't have any publicly mutable expression type (e.g., `Expression`) that is also restricted to "var-less" expressions, but I can easily envision adding one (e.g., `DataExpression`).

Why does this even matter. Well the objects in a `DataExpression` might communicate other information that don't have to do with variables or the polynomial degree but could be useful for other pyomo "solver" interfaces (something I am working on).

DLWoodruff:
To the extent that you are able, please forgive me for kibitzing. Two
comments:

_never_allowed_be_an_expression_that_references_Vars

seems to me to subsume the first part (I assume you mean Vars and not
variables; I go on to assume that at least when they are first written
down, expression that pass this test do, in fact, often reference variables)


Next point: I don't see what is so bad about a long name for fairly arcane
things like this. In the few code fragments that need to reference this
more than once, it can be assigned to a shorter name.  The trouble with
elegant names like _is_data is that programmers who are three levels deep
in their bug-fix tree stack and cutting and pasting lines of code late at
night will think they know that _is_data means (and they don't have time to
check the docstrings for everything in the lines they are pasting
so they will be relieved that at least something is easy to understand) but
they won't. They will understand

_never_allowed_be_an_expression_that_references_Vars

or at least they will know they better read the docstring.

  Dave


On Fri, Oct 14, 2016 at 8:50 AM, Gabriel Hackebeil <notifications@github.com

jsiirola:
@jwatsonnm and I talked through names and like `potentially_variable()` (the logical negation of `is_data`).  This is also a nice contrast to `is_fixed` and `is_constant` where the methods tell you something about the *current* status of the expression; in this case, the method is telling you about what could *potentially* be a status in the future.

DLWoodruff:
I have to admit that is pretty good. Using "variable" as an adjective
really does the trick. That is so good that maybe it doesn't even need a
leading underscore

ghackebeil:
I think it's a little too clever, but I'll take anything at this point. I would also prefer the underscore because it's quite a low level property and there is a public function in numvalue.py to accompany it.

Moving past the name momentarily:
Should we include the fix to Constraint.add in the PR? It is a very simple change and it fixes some terrible edge cases that would be around until someone re-architects Constraint.add, canonical_repn , and all of the solver plugins. I don't predict this will happen in less than a year, but feel free to prove me wrong. If _anyone_ is opposed, say so now and I'll revert that part of the PR.

jwatsonnm:
I am offended, as I was pleased with my cleverness that early in the AM ☺

ghackebeil:
Well then that was a compliment! 😉

PR #44, opened 2016-09-13
Support generating "template expressions" from a rule

jsiirola:
This PR adds the ability to generate "template expressions" (previously, we called this idea "abstract expression trees").  The idea is that we call the rule with a special class (an `IndexTemplate`) instead of the usual index value.  This changes the way that expressions are generated (in particular, through the `IndexedComponent.__getitem__()` method) so that the `__getitem__` call gets mapped to a new `_GetItemExpression` node in the expression tree.

This can only generate template expressions for simple rules (no conditionals that depend on the `IndexTemplate` object).  Rules that use the `IndexTemplate` in a Boolean context (or otherwise evaluate the numeric value of the `IndexTemplate` outside the context of `__getitem__` will generate a `TemplateExpressionError` exception.

This feature does not change the current functionality and should be 100% backwards compatible.


jsiirola:
We discussed this PR today, and identified the following problems
  - the `substitute_template_with_param` method will fail for cases where the same component is used in an expression with different indices, e.g. `m.x[t,1] + m.x[t,2]`.
  - the `substitute_template_with_index` method is poorly named

PR #56, opened 2016-10-19
Remove floating return statement in XPRESS.py solver plugin

### LINKED ISSUE #55: XPRESS Solver Error : AttributeError: 'NoneType' object has no attribute 'problem'
I've pasted the traceback below :

```
  File "model.py", line 15, in <module>
    solver.solve(model)
  File "/Users/$USER/anaconda/lib/python2.7/site-packages/pyomo/opt/base/solvers.py", line 587, in solve
    result = self._postsolve()
  File "/Users/$USER/anaconda/lib/python2.7/site-packages/pyomo/opt/solver/shellcmd.py", line 267, in _postsolve
    results = self.process_output(self._rc)
  File "/Users/$USER/anaconda/lib/python2.7/site-packages/pyomo/opt/solver/shellcmd.py", line 329, in process_output
    self.process_soln_file(results)
  File "/Users/$USER/anaconda/lib/python2.7/site-packages/pyomo/solvers/plugins/solvers/XPRESS.py", line 336, in process_soln_file
    results.problem.number_of_objectives=1
AttributeError: 'NoneType' object has no attribute 'problem'
```

kdheepak:
This problem is because of a floating `return` statement in the `process_logfile` function in the [xpress solver plugin](https://github.com/Pyomo/pyomo/blob/master/pyomo/solvers/plugins/solvers/XPRESS.py#L201). This returns a `None` object instead of the `results` which causes the function `process_soln_file` to error. This patch removes the floating `return` and fixes #55.

### DIFF
--- a/pyomo/solvers/plugins/solvers/XPRESS.py
+++ b/pyomo/solvers/plugins/solvers/XPRESS.py
@@ -198,8 +198,6 @@ class XPRESS_shell(ILMLicensedSystemCallSolver):
         log_file_contents = "".join(log_file.readlines())
         log_file.close()
 
-        return
-
         for line in log_file_contents.split("\n"):
             tokens = re.split('[ \t]+',line.strip())
 


ghackebeil:
@kdheepak: We've had trouble obtaining a test license for the XPRESS solver, so it is very likely this solver plugin is out of date. However, I believe it is possible to use the ASL interface to XPRESS with Pyomo, which might use a different executable (named something like xpressampl), but I'm not sure.

Can you clarify what interface to XPRESS you are using to test this? Are you using the solver_io keyword? If so, is it set to 'lp' or 'nl'?

kdheepak:
Hi @ghackebeil, using `xpressampl` (or its equivalent) might be the best long term solution! I've tested it in two ways,  

1) using a SolverFactory instance as shown below. 

```
solver = SolverFactory('xpress')
solver.solve(instance)
```

2) using the command line 

```pyomo solve model.py model.dat  --solver=xpress```

Both of these seem to work from my initial testing, after adding the above patch. I'm not using the `solver_io` keyword, so PYOMO must be defaulting to something. 

Let me know if you want me to run some additional tests.

**Comment by jsiirola** [comment]
@ghackebeil: The spurious `return` that this PR removes was added in 2012 by @jwatsonnm.  From the looks of things, it was a debugging statement that was put in during initial development to fence off the results processing.  JP will have to comment, but I don't see a problem with accepting the PR.

**Comment by ghackebeil** [comment]
I'm fine accepting the PR as well. It sounds like the LP interface is being tested here. My only concern is whether or not these changes break the plugin when using the ASL interface.

@kdheepak: Do you have a way to test this? Can you try with solver_io='nl' passed into SolverFactory (--solver-io=nl on the command line)? It may or may not require you to obtain another executable (e.g. xpressampl).

**Comment by kdheepak** [comment]
Yes, I am able to test this.

```pyomo solve model.py model.dat  --solver=xpress --solver-io=nl```

I ran the above and got the same results as before.

**Comment by ghackebeil** [comment]
As in it works? There is a serious issue right now where sometime if no solution is returned during this solve (e.g., because the solver execution failed) then it looks like things worked, but it really silently fails. Can you confirm that there is some kind of meaningful status info in the results object when you solve with the different interfaces (solver_io='lp' and solver_io='nl')?

**Comment by kdheepak** [comment]
This is what I ran. It does provide the same objective function value in the two cases. However, with `--solver-io=lp` the `status` is returned as `unknown`

```
$ pyomo solve model.py model.dat  --solver=xpress --solver-io=lp --json

[    0.00] Setting up Pyomo environment
[    0.00] Applying Pyomo preprocessing actions
[    0.02] Creating model
[    0.45] Applying solver
[    0.89] Processing results
    Number of solutions: 1
    Solution Information
      Gap: None
      Status: unknown
      Function Value: 90127.0682088
    Solver results file: results.json
[    1.01] Applying Pyomo postprocessing actions
[    1.01] Pyomo Finished
```

```
pyomo solve model.py model.dat  --solver=xpress --solver-io=nl --json
[    0.00] Setting up Pyomo environment
[    0.00] Applying Pyomo preprocessing actions
[    0.03] Creating model
[    0.49] Applying solver
[    1.24] Processing results
    Number of solutions: 1
    Solution Information
      Gap: None
      Status: optimal
      Function Value: 90127.0682088
    Solver results file: results.json
[    1.31] Applying Pyomo postprocessing actions
[    1.31] Pyomo Finished
```

I'm assuming that since the function value is returned, it solves in both cases.

ghackebeil:
Okay. It looks like it is working to a certain extent.

kdheepak:
Great! Let me know if there are additional quick tests that you'd like me to run with XPRESS.

ghackebeil:
I'm fine accepting the PR as well. It sounds like the LP interface is being tested here. My only concern is whether or not these changes break the plugin when using the ASL interface.

@kdheepak: Do you have a way to test this? Can you try with solver_io='nl' passed into SolverFactory (--solver-io=nl on the command line)? It may or may not require you to obtain another executable (e.g. xpressampl).

Gabe

Issue #58, opened 2016-10-21
XPRESS.py always uses lpsolve irrespective of problem type


kdheepak:
The XPRESS.py plugin using [lpsolve](https://github.com/Pyomo/pyomo/blob/master/pyomo/solvers/plugins/solvers/XPRESS.py#L159) irrespective of the problem type. This results in an incorrect solution when solving MILP problems. The correct solver is `mipoptimize` for MILP problems. XPRESS.py should either

1) Ask the user for the problem type
2) Have `pyomo` figure out the problem type

ghackebeil:
Pyomo would definitely be able to figure out the problem type (it checks everything about the model inside of the LP writer). The only challenge here is figuring out how to get that information out of the LP writer and into this method.

The easiest thing to try would be to just change that line to `mipoptimize` and see if XPRESS is smart enough to fall back to `lpoptimize` when the problem is an LP. I'm mean, surely it can solve a MIP that just _happens_ to have no discrete variables. Can you test if this works?

It's also important that this change would not cause XPRESS to no longer return dual solution information if the problem is an LP. Can you test this as well?

kdheepak:
😄 What a great idea! I'll check this out and let you know. Is there a model.py + model.dat that you can share that I can use for testing?

ghackebeil:
No need for a `.dat` file. Just paste this code into a file and run it as a python script (not using the pyomo command):
```python
from pyomo.environ import *

m = ConcreteModel()
m.x = Var()
m.y = Var(bounds=(0,1))
m.o = Objective(expr= m.x + m.y)
m.c = Constraint(expr= m.x >= 1)
m.dual = Suffix(direction=Suffix.IMPORT)
m.rc = Suffix(direction=Suffix.IMPORT)
m.slack = Suffix(direction=Suffix.IMPORT)

solver = SolverFactory("xpress",
                       solver_io="lp")
results = solver.solve(m)

print(m.o())
print(m.x())
print(m.y())
print(m.dual[m.c])
print(m.rc[m.y])
print(m.slack[m.c])

# should print                                                                                 
# 1                                                                                            
# 1                                                                                            
# 0                                                                                            
# 1                                                                                            
# 1                                                                                            
# 0  
```

kdheepak:
I changed the example to the following to run the above test.

```
from pyomo.environ import *

m = ConcreteModel()
m.x = Var()
m.y = Var(bounds=(0,1))
m.o = Objective(expr= m.x + m.y)
[...]

PR #70, opened 2016-11-14
Conditionally choose lp or mip solver based on problem for XPRESS

kdheepak:
This addresses #58

### DIFF
--- a/pyomo/solvers/plugins/solvers/XPRESS.py
+++ b/pyomo/solvers/plugins/solvers/XPRESS.py
@@ -69,6 +69,8 @@ class XPRESS_shell(ILMLicensedSystemCallSolver):
         kwds['type'] = 'xpress'
         ILMLicensedSystemCallSolver.__init__(self, **kwds)
 
+        self.is_mip = kwds.pop('is_mip', False)
+
         #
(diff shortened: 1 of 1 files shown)


ghackebeil:
I don't think this will fix this problem. The main problem file always ends with .lp (unless you are using the xpress_ampl interface). Have you tested this?

kdheepak:
You are right. I was using `solver_io='mps'` in the example that I tested
with, and it worked then. What is the way to find if the problem is a mip
in the solver instance?

kdheepak:
I've updated the pull request. This will allow the user to pass `is_mip=True` as a keyword argument when creating a solver instance to call `mipoptimize` during a solve. There may be better ways to achieve the same effect. This PR can be considered a temporary fix, while a better implementation can be discussed. This offers the choice to the user while also being backward compatible.

ghackebeil:
I think this is a good short term fix as well. I won't have time to look at the longer term fix for a least a week or two, but it will involve the problem writers (LP, MPS, NL) returning a classification of the model in some form (LP, MIP, QP, QCP, MIQCP, etc.), which I think could be useful in other contexts.

PR #72, opened 2016-11-20
Adding IPOPT Nightly Tests

whart222:
This PR adds IPOPT to the Travis Tests.  The IPOPT binary is loaded form the cachemeorg channel.  Unfortunately, this solver returns values that do not match Pyomo baseline files.


whart222:
@carldlaird Can you diagnose these test failures?  Is there a problem with the ipopt executable?

ghackebeil:
When I originally added these tests, I came across this bug in Ipopt. Bug is probably too strong of a word. The sign of the dual solution wasn't very consistent with the objective sense, and went against the convention used by most solvers. It was fixed in the Ipopt 3.10.4 release.

I will add an expected failure to the solver tests when the Ipopt version is older than 3.10.4.

PR #67, opened 2016-11-10
Tracking changes for #65

ghackebeil:
Merging this PR will resolve #65 automatically, so make any fixes for this issue within this PR branch. This initial commit updates the pickle tests in pyomo/solvers/tests/pickling to expose the problem discussed in #65.

Issue #65, opened 2016-11-09
Issue cloning an unpickled model that was solved before it was pickled


ghackebeil:
The following simple example results in a very strange error inside of __setstate__ on the ModelSolutions object:

```python
import pickle
from pyomo.environ import *

m = ConcreteModel()
m.x = Var()
m.o = Objective(expr=m.x)
m.c = Constraint(expr=m.x >= 1)

#                    
# SOLVE A MODEL
#
glpk = SolverFactory("glpk")
glpk.solve(m)

#
# PICKLE THE SOLVED MODEL
#
m_unpickled = pickle.loads(pickle.dumps(m))

#
# CLONE THE UNPICKLED MODEL
#
m_cloned = m_unpickled.clone() # ERROR
```

The exact order here is important: The unpickled version of a model THAT WAS SOLVED BEFORE IT WAS PICKLED can not be cloned. Here is a partial stack trace:
```python
  File "~/Frameworks/Python.framework/Versions/2.7/lib/python2.7/copy.py", line 182, in deepcopy
    rv = reductor(2)
  File "~/Pyomo/pyomo/pyomo/core/base/PyomoModel.py", line 159, in __getstate__
    tmp[name].update( (ComponentUID(obj()), entry) for (obj, entry) in itervalues(data) )
  File "~/Pyomo/pyomo/pyomo/core/base/PyomoModel.py", line 159, in <genexpr>
    tmp[name].update( (ComponentUID(obj()), entry) for (obj, entry) in itervalues(data) )
TypeError: 'NoneType' object is not callable
```
I will expand the pickle tests to expose this.

@whart222, @jsiirola: Mind taking a look?

ghackebeil:
I just came across a performance issue related to this that I though was worth documenting. The existence of a "solution" in the model.solutions object is slowing down the time of model.clone() on a model I am using **by a factor of 34** (~3 seconds to ~100 seconds). Calling model.solutions.clear() before model.clone() fixes the performance issue, but this begs the question of whether or not we should really be caching a solution by default (it is just duplicating what is stored in the model).

I'm guessing the timing has to do with generation of ComponentUIDs when cloning the solution object. I recently added a helper function for fast bulk generation of ComponentUID strings to block.py called generate_cuid_names that might be of use in this situation. Whoever looks into the pickle issue might want to look at this performance issue as well.

(references #67)

PR #78, opened 2016-11-27
Categorized solver tests

whart222:
This PR reworks the solver testing infrastructure to enable the enumeration of all test scenarios.  Test models are now managed in the pyomo.solver.tests.models package, and the writer and pickling tests are in the tests.checks package.

This logic is also used to support the testing logic in 
>   pyomo test-solvers
This subcommand generates a table summarizing the tests and test failures.


whart222:
@ghackebeil You should check out the PySP testers.  I had to adjust these to use the reworked test suite logic.

whart222:
Note that there are test_*.py scripts in the pyomo.solvers.tests module that need to be reviewed.  I think that further tests could be consolidated.

ghackebeil:
I think this is good to go. The only issue I came across was that the solver options were not being set prior to the solve. This is fixed now, and I've made a few other bug fixes to the solver interfaces in the process.

I still have many local failures for CBC and CPLEX about bad result statuses, but I won't have time to fix these for at least a few weeks, so I think we should go ahead and merge this into master before then.

This is also probably going to create a conflict for PR #67 (changes to pickle tests to expose bug from #65 ), but I'll manually resolve that once this is merged into master.

PR #79, opened 2016-11-27
Fix xpress solver options writer

kdheepak:
This pull request makes two changes.

1. Adds a newline character when printing options for every `option_name`
1. Pops `mipgap` from options if it exists so that it isn't rewritten in the following `for` loop for all options.


jsiirola:
This looks good to me. @jwatsonnm: I think you originally drafted the xpress interface - if you agree, I think we can merge.

whart222:
Well, I accidentally merged this.  I'm going to leave it merged since John thought it was OK.

PR #81, opened 2016-12-09
Eliminate extraneous 's' characters to fix error display

r-barnes:
Several error messages attempted to use `%` substitution of the form `"A thing '%s' had a problem '%s'" %s(a,b)`. Note that the `%s(a,b)` is invalid. I have converted these instances to `%(a,b)`.


ghackebeil:
Nothing to debate about here. Thanks for the fix.

PR #99, opened 2017-01-14
Adding support for decorating rules to simplify syntax.

jsiirola:
This adds support for declaring components on a model by decorating the
rule instead of explicitly invoking the setattr method on the block.
This "fixes" the normal pattern of the component name appearing in three
places.

This is an implementation of a feature requested by Ned Dimitrov at the PyomoFEST/Austin.

### DIFF
--- a/examples/pyomo/p-median/decorated_pmedian.py
+++ b/examples/pyomo/p-median/decorated_pmedian.py
@@ -1,0 +1,39 @@
+from pyomo.environ import *
+
+import random
+random.seed(1000)
+
+model = AbstractModel()
+
+model.N = Param(within=PositiveIntegers)
+model.P = Param(within=RangeSet(1,model.N))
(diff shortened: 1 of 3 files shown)


jsiirola:
One note: this solution is hard-coded to the `rule=` keyword argument. I am wondering if we should allow an additional keyword argument to the decorator so this could be used for, e.g. the `initialize=` keyword for Var. That still won't help when the user wants to specify two rules (e.g. initialize and bounds), but I think I am OK with that.

jsiirola:
The general consensus (of @ghackebeil, @DLWoodruff, and @jwatsonnm) is that we should keep this as-is and only support the decorator for `rule=`.  This solves 95% of the problem and does not introduce extra complexity.

whart222:
I think we should augment the Var/Param components to support rule= as a
default for one of their other options (e.g. initialize=).

--Bill

PR #109, opened 2017-01-23
Disable solver tests while importing

whart222:
This is a stab at reworking the solver logic to delay checks for solvers until they are constructed.

I haven't been able to debug this for CPLEX solvers.  Some other developer will need to help with that.

I don't have an absolute check that NO solvers are being checked.  I'm not sure we can do that.

This addresses a concern raised by @jsiirola about the time required to import pyomo.environ.  After these edits, I'm only seeing import times of ~2 seconds.


ghackebeil:
Can I propose delaying checks for availability until either the user explicitly asks (by calling solver.available()) or until after a failure occurs when we attempt to launch the solver? Otherwise, we are still going to be launching extra shell commands that we do not need to launch, and now you are moving this logic into a user’s solve loop, which might slow down their code.

whart222:
Gabe:

It'll just slow down their code the first time, right?  The checks are
short-circuited with a test to see if the solver configurations have been
loaded.

We could try moving into the available() function, but I think that one of
the solver constructors used information about the solver configuration...

--Bill

whart222:
@ghackebeil @jsiirola @jwatsonnm :  Can one of you review?

whart222:
FWIW, I'm not sure what the failure for codecov/patch means.  I don't think this should be a blocker for merging this PR.

jsiirola:
The codecov patch failure is because the patch (the cumulative edits in the PR) have a cumulative coverage less than the code base coverage (so will reduce overall test coverage). Normally, I wold say this is a blocker, but as we don't have a way to test solvers on TravisCI, we can probably see finite it in this case.

pyomo/solvers/plugins/solvers/CPLEXDirect.py
```diff
@@ -122,10 +129,10 @@ class CPLEXDirect(OptSolver):
                             doc='Direct Python interface to the CPLEX LP/MIP solver')
 
     def __init__(self, **kwds):
+        configure_cple_()
```
jsiirola: This is a typo that breaks the cplex direct plugin.

jsiirola:
@whart222: I think @ghackebeil is making the argument that the correct place for these checks is actually in the `available()` method instead of in `__init__()` where you have them now.  Conceptually, I agree with him - although this is a significant enough change that I don't want "perfect" to be the enemy of "good enough".  I think that when we get around to redesigning the solver plugin API will be the "right" time to move things from `__init__()` to `available()` (or what ever the new API equivalent method is).

My only remaining "concerns" with this PR is if you really want the extra `.tex` files added to master and if you can (easily) do something similar with the NEOS interface (see my review).  If the answer is "not easily" to the latter, then I think we should accept this PR and open an issue to track the NEOS problem.  Thoughts?

jsiirola:
We will go ahead and merge this PR and open a new issue to track the NEOS interface issue.

Issue #94, opened 2016-12-29
Solveable Blocks


eslickj:
This is a request for a new feature.  It would be nice if Blocks could be solved the by passing them to a solve function the same way models can.  The reason I would like to do this is to solve sub-problems as part of a model initialization.  Currently my two solutions are:

1) Deactivate all other blocks and solve model.  This can get a bit messy.

2) Use ConcreteModel instead of Block.  This works fine, but I think there is a problem when I want to use an IndexedBlock as I don't think there is an equivalent IndexedConcreteModel.

The main problem I see with solvable blocks is that some constraints may use variables that are actually contained in other blocks.  Unless connectors or similar are used for all variables shared between blocks, things could get ugly, and that may be a good reason not to implement this feature. I think (when a block is being solved independently) the block would need to pull in variables from other blocks not being solved, and it could be a lot more complex than it seems at first.

For very complex models this may provide some path for problem decomposition and parallelization.  Although I haven't thought much about the details.

carldlaird:
I create ConcreteModel objects in place of blocks (as described in # 2 above) frequently. This can serve as a work-around until Blocks support solving on their own.

In the construction rule for the indexed block, you can return a new object instead of "filling-in" the block that is passed to you. For example,

    def block_rule(b,i):
        block = ConcreteModel()
        ...
        return block
    
    model.blocks = Block(model.T, rule=block_rule)

While this is a reasonable work-around, I think the issue should be left open until we have decided if the Block component can (or should) be solvable on its own.

jsiirola:
Fixed by #110

whart222:
I want to re-open this issue.  This is a big enough change that I would expect some documentation of it, other than the changes in the test directory.

We haven't historically done this documentation, and I'm open to suggestions.  But what about having a blog-like documentation of the CHANGELOG?  This would contain significant changes to master.  If the online documentation has been updated, then the blog could simply reference those updates.  But in a case like this, where the design might change, the blog would contain the description of this change.

PR #110, opened 2017-01-31
Support solving Blocks as if they were Models

jsiirola:
This allows arbitrary blocks to be solved as if they were models.

This addresses an IDAES need where we would like to be able to send arbitrary sub-blocks of a model to a solver (e.g., for sequential initialization of complex nonlinear models).  The fix is conceptually simple: just promote two methods from `Model` up to `_BlockData` (`valid_problem_types` and `write`).

This isn't the cleanest solution (in particular, it has a rather ugly circular import dependency).  However, it is simple and should work, at least until we have a chance / opportunity to re-architect the solver plugins and problem writers (ideally, this would eliminate the need for attaching `SymbolMap`s to the model (the cause of the circular imports).


jsiirola:
This PR would fix #94.

Issue #52, opened 2016-10-17
pyomo.dae transformations ignore Piecewise blocks


ghackebeil:
The issue should be explained near the end of the following thread:
https://groups.google.com/forum/#!topic/pyomo-forum/Q_60FKCXqkw

It looks like Piecewise does not get re-constructed over the expanded `ContinuousSet` .

ghackebeil:
@jsiirola, @blnicho: I dug into the dae code a little and noticed that only components of type Var, Constraint, and Expression get expanded with the new index elements. In order for this to work with Piecewise (or other Blocks) it would have to include the Block type (see `update_contset_indexed_component` in pyomo/dae/misc.py)

In the short term, that function should probably be updated to raise an exception when it encounters types it does not handle. It could probably also alert users that a workaround is to add certain model components after applying the dae transformation.

jsiirola:
Fixed by #111

PR #111, opened 2017-01-31
Fix to expand Blocks indexed by a ContinuousSet

### LINKED ISSUE #89: DAE: expand blocks indexed by a ContinusSet
We need to update `pyomo.dae.misc.update_contset_indexed_component()` to also expand `Block`s that are indexed by `ContinuousSet`s.  This is an issue with models like a packed column whose length is specified as a continuous set.  Things like the physical properties for each "stage" (finite element) will need to be expanded.

Reported by @eslickj and @andrewlee94.

blnicho:
This pull request fixes #89 and modifies the "update ContinuousSet indexed component" logic to throw an error when an unrecognized ContinuousSet indexed component is encountered as suggested in #52.

### DIFF
--- a/pyomo/dae/misc.py
+++ b/pyomo/dae/misc.py
@@ -12,7 +12,7 @@ import logging
 from pyomo.core import *
 from pyomo.core.base.indexed_component import IndexedComponent
 from pyomo.core.base.misc import apply_indexed_rule
-from pyomo.core.base.block import _BlockData
+from pyomo.core.base.block import _BlockData, IndexedBlock
 from pyomo.dae import *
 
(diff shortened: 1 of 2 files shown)


blnicho:
@jsiirola thanks for the comments, I addressed everything you mentioned. 

I also added support for expanding Piecewise components indexed by a ContinuousSet fixing #52. @ghackebeil could you confirm that Piecewise components are being expanded correctly? I'm pretty sure my fix  (simply reconstructing the component) works but I've never used the Piecewise component in practice so I might have missed some subtlety.

ghackebeil:
Fix looks good for Piecewise.

PR #123, opened 2017-03-08
Pyomo4 expressions

ghackebeil:
Summary of changes on this branch:
  - Includes an additional core expression system ("pyomo4") that does not require `sys.getrefcount`. For now, the original expression system ("coopr3") is used by default, and we only fall back to this new system when `sys.getrefcount` is not available. This allows us to support the PyPy interpreter.
  - Fixes to the "coopr3" expression system to support Python3.6. Some conditional adjustments were required for the reference count constants used to control expression cloning. Due to significant reorganization of the core expression code, it was easier to make this change on this branch than master.

Ideally, we would not have two separate expression systems as this requires a large amount of duplicated code in places like canonical_repn.py, ampl_repn.py, and the NL writer. However, merging this branch in its current state will allow developers and users to start doing more thorough profiling on PyPy. After this merge takes place, a new branch will be created with the goal of moving to a single expression system that supports interpreters with and without `sys.getrefcount`.


ghackebeil:
@whart222: I wasn't able to get a PyPy build going on Travis due to issue with the conda commands you are running. Would you mind taking a stab at this? Same goes for the other Python3.6 builds that are commented out in the Travis script, but getting a single PyPy build running is the priority.

On a related note, I came across an interesting [blog post](http://steven.casagrande.io/articles/travis-ci-and-if-statements/) about issues using if statements in a Travis script. The build will not stop if a command fails inside of an if statement. You may want to deal with that while you are at it.

ghackebeil:
We are going to merge this branch in its current state as it contains a working "pyomo4" expression system. There as still some problems with PyPy, but we will revisit those on a new branch in the near future. Python 3.6 support needs to be in a release.

PR #126, opened 2017-03-19
Restrict component names

jsiirola:
We recently had a [question on StackOverflow](http://stackoverflow.com/questions/42834501/typeerror-cannot-compute-the-value-of-an-array-of-parameters) that highlighted a source of confusion with component names.  In this case, the user creates a Param names `type`.  This hid the `Block.type()` method which resulted in a very uninformative error message when the user attempted to solve the model.

This PR prevents users from declaring *components* that would override standard `Block` methods.  I originally thought about preventing ANY user-defined attributes from overriding built-in methods, but that proved to be difficult to implement (it causes a bit of a chicken & egg issue when `Block` methods update their own attributes).  So, a user could still shoot themselves in the foot by declaring non-component attributes that override the built-in methods.  I figure that that (a) is a pretty uncommon situation and (b) would be un-Pythonic if we prevented it.

### DIFF
--- a/pyomo/core/base/PyomoModel.py
+++ b/pyomo/core/base/PyomoModel.py
@@ -546,6 +546,8 @@ class Model(SimpleBlock):
 
     preprocessor_ep = ExtensionPoint(IPyomoPresolver)
 
+    _Block_reserved_words = set()
+
     def __new__(cls, *args, **kwds):
         if cls != Model:
             return super(Model, cls).__new__(cls)
@@ -980,6 +982,12 @@ class AbstractModel(Model):
     def __init__(self, *args, **kwds):
         Model.__init__(self, *args, **kwds)
 
+
+#
+# Create a Model and record all the default attributes, methods, etc.
+# These will be assumes to be the set of illegal component names.
+#
+Model._Block_reserved_words = set(dir(Model()))
 
 
 register_component(Model, 'Model objects can be used as a component of other models.')
--- a/pyomo/core/base/PyomoModel.py
+++ b/pyomo/core/base/PyomoModel.py
@@ -987,7 +987,11 @@ class AbstractModel(Model):
 # Create a Model and record all the default attributes, methods, etc.
 # These will be assumes to be the set of illegal component names.
 #
-Model._Block_reserved_words = set(dir(Model()))
+# Note that creating a Model will result in a warning, so we will
+# (arbitrarily) choose a ConcreteModel as the definitive list of
+# reserved names.
+#
+Model._Block_reserved_words = set(dir(ConcreteModel()))
 
 
 register_component(Model, 'Model objects can be used as a component of other models.')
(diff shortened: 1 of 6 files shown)


ghackebeil:
From the changes I've seen in these commits, it looks like ".index" is restricted name. Is this correct? If so, I'm surprised. I've used that as a name in many many examples.

jsiirola:
Index is the name of a method that returns the underlying indexing set. Almost nothing uses it in core, so overriding it usually doesn't cause a fatal exception. But you are overriding a part of the Block API. 

That is the point of the PR: folks could accidentally override the core and then get Very unintuitive errors. 

The simple solution was to disallow ANY components that override a declared method.

ghackebeil:
I want to say that the better solution to this would be good API documentation. It feels un-pythonic to do this, but I think you are right in that just disallowing it for components is a reasonable trade off.

jsiirola:
I would be ok with this tossing a (strong) warning. That should accomplish basically the same thing but would allow folks to override the methods with a component if they really, really wanted to...

ghackebeil:
I'm okay with either.

I think this entire issue is a motivating argument for making more of our low-level methods start with an `_` (e.g., `is_fixed`). Whatever the case, we should keep this issue in mind when we do a code review of the kernel block API (e.g., Would longer public method names reduce the chances of clashing with a user preferred name?).

blnicho:
We (@jsiirola, @carldlaird, and I) discussed this PR yesterday, specifically whether we should throw a warning or an exception when a user tries to override a declared method on Blocks. The consensus was that we should throw an exception. If the user overrides a critical method and ignores a warning they will most likely get a cryptic error message or, in the worst case, successful execution but an incorrect model. Despite the un-pythonic-ness of this fix we think it is a necessary restriction.

whart222:
FWIW, I agree that we should throw an exception, and I don't think this
behavior is un-Pythonic.  Here's a simple example:

>>> a = {}
>
> >>> dir(a)
>
> >>> ['__class__', '__cmp__', '__contains__', '__delattr__', '__delitem__',
> '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__',
> '__getitem__', '__gt__', '__hash__', '__init__', '__iter__', '__le__',
> '__len__', '__lt__', '__ne__', '__new__', '__reduce__', '__reduce_ex__',
> '__repr__', '__setattr__', '__setitem__', '__sizeof__', '__str__',
> '__subclasshook__', 'clear', 'copy', 'fromkeys', 'get', 'has_key', 'items',
> 'iteritems', 'iterkeys', 'itervalues', 'keys', 'pop', 'popitem',
> 'setdefault', 'update', 'values', 'viewitems', 'viewkeys', 'viewvalues']
>
>
> >>> a.pop = 2
>
> >>> Traceback (most recent call last):
>
>   File "<stdin>", line 1, in <module>
>
> AttributeError: 'dict' object attribute 'pop' is read-only
>

Clearly, Python also enforces read-only attributes.

--Bill

ghackebeil:
Good point.

PR #130, opened 2017-03-29
Pyomo Kernel

ghackebeil:
Summary
----------
This PR summarizes the changes on the component_storage_interface branch. The motivation for this work was to address a number with the original design in `pyomo/core/base`. These include:
  - a class hierarchy that made it difficult to subclass components
  - excessive use of memory for singleton objects (which bloated models that made significant use of indexed blocks)
  - too strong of a connection between the Pyomo `Set` object and indexed components (and the issues that go along with `Set`, such as flattening of tuples)
  - too strong of a connection between the "delayed construction" mechanism used by `AbstractModel` and the "concrete" interface for modeling components (which made the code base overly complex)

This branch includes an experimental rewrite of the core component storage hierarchy in Pyomo to address these issues. This new storage hierarchy resides in the `pyomo/core/kernel` directory.

The recommended way of using this library is to import symbols from the dummy namespace `pyomo.kernel`. This takes care of some initial setup of object categories that must be done outside of `pyomo.core.kernel` to avoid a circular dependency with `pyomo.core.base` (hopefully this is temporary). This also takes care of the `pyomo.environ` import. Example usage:
```python
import pyomo.kernel as pmo

m = pmo.block()
m.x = pmo.variable()
m.o = pmo.objective(m.x**2, sense=pmo.minimize)
m.c = pmo.constraint(m.x >= 1)

opt = pmo.SolverFactory("cplex")
result = opt.solve(m)
assert str(result.solver.termination_condition) == "optimal"
```

Core Functionality
------------------

Most (if not all) of the functionality in `pyomo.core.base` related to concrete modeling has been re-implemented in `pyomo.core.kernel`. The following standard modeling components are available:
  - `block` (a concrete model)
  - `variable`
  - `constraint`
  - `objective`
  - `expression`
  - `parameter` (always mutable, no functionality related to sparsity though)
  - `sos` (special ordered set)
  - `suffix`
  - `piecewise`

In addition, generic `dict`, `list`, and `tuple` containers exist for storing each of these component types. Examples can be found in the [pyomo/examples/kernel](https://github.com/Pyomo/pyomo/tree/component_storage_interface/examples/kernel) directory.

A significant effort has been placed on documentation. Google style doc-strings have been added to all new classes, methods, and functions in `pyomo.core.kernel`. I've also started to put together API documentation using the sphinx build system in the `pyomo/doc/OnlineDocs` directory.

Syntax Comparison with Existing AML Interface
-----------------------------------------------

The main differences between the kernel API and that of the existing AML is a more explicit syntax when it comes to building indexed objects, a lack of support for abstract models, and no concept of "delayed construction". With respect to the last point, this means that objects are fully initialized when you create them, not just after assigning them to a model.

In the AML, the standard procedure for defining an indexed collection of constraints is to define a `Set` object to act as an index, and pair that with a "constraint rule" that returns a constraint expression for each item in the index. This looks like
```python
import pyomo.environ as aml

m = aml.ConcreteModel()
m.s = aml.Set(initialize=range(10))
m.x = aml.Var(m.s)
def c_rule(m, i):
    if i ==5:
        return aml.Constraint.Skip
    else:
        return m.x[i] >= i
m.c = aml.Constraint(m.s, rule=c_rule)
```
In the kernel API, there is no concept of an index `Set` or "rules". One simply chooses a style of container (e.g., `dict`, `list`, `tuple`), and populates it by explicitly adding modeling objects of the same category to it. The previous example could be written as:
```python
import pyomo.kernel as pmo

m = pmo.block()
m.x = pmo.variable_list()
m.c = pmo.constraint_list()
for i in range(10):
    m.x.append(pmo.variable())
    if i != 5:
        m.c.append(pmo.constraint(m.x[i] >= 1))
``` 
Note that modeling objects placed in the list-style containers above are the same objects that one could assign to a block as _scalar_ (or non-indexed) components. E.g.,
```python
import pyomo.kernel as pmo

m = pmo.block()
m.x = pmo.variable()
m.x_indexed = pmo.variable_dict()
for key in ['a','b',None]:
    m.x_indexed[key] = pmo.variable()
```
Additionally, the kernel `dict`, `list`, and `tuple` containers support nested layers of containers, giving more control over the model storage hierarchy. E.g.,
```python
import pyomo.kernel as pmo

x = pmo.variable_dict()
x[None] = pmo.variable_list()
x[None].append(pmo.variable())
x[None].append(pmo.variable_list())
x[None][1].append(pmo.variable())
```

Memory / Speed Comparison with AML Components
----------------------------------------------------

The following table shows memory / speed differences for different situations between some AML components (upper case) and kernel components (lower case). These were generated on a late 2013 model Macbook Pro with Python 3.6.1 ([attached](https://github.com/Pyomo/pyomo/files/1030073/compare_components.py.txt) file has a txt extension, but it is the Python code that generated this table).

                                             Label        Bytes              Seconds        
                                               Var       2.6 KB             0.000046        
                                   _GeneralVarData      288.0 B (0.11x)     0.000008 (0.18x)
                                          variable      288.0 B (0.11x)     0.000007 (0.16x)
                                    staticvariable      144.0 B (0.05x)     0.000006 (0.12x)

                                             Label        Bytes              Seconds        
                                Indexed Var (2500)     501.3 KB             0.011827        
                              variable_dict (2500)     608.6 KB (1.21x)     0.008019 (0.68x)
                              variable_list (2500)     281.4 KB (0.56x)     0.007888 (0.67x)
                        staticvariable_list (2500)     161.4 KB (0.32x)     0.006424 (0.54x)

                                             Label        Bytes              Seconds        
                                             Block       2.8 KB             0.000089        
                                        _BlockData       1.0 KB (0.36x)     0.000023 (0.25x)
                                             block       1.1 KB (0.40x)     0.000021 (0.24x)
                                        tiny_block      504.0 B (0.18x)     0.000015 (0.17x)

                                             Label        Bytes              Seconds        
                             Block w/ 3 components       7.2 KB             0.000163        
                        _BlockData w/ 3 components       6.0 KB (0.83x)     0.000119 (0.73x)
                             block w/ 3 components       3.3 KB (0.46x)     0.000052 (0.32x)
                        tiny_block w/ 3 components       1.5 KB (0.20x)     0.000038 (0.23x)

                                             Label        Bytes              Seconds        
                  Indexed Block (2500) w/ Vars (8)      18.0 MB             0.523436        
                block_list (2500) w/ variables (8)       7.7 MB (0.43x)     0.092805 (0.18x)
           tiny_block_list (2500) w/ variables (8)       3.4 MB (0.19x)     0.080594 (0.15x)
     tiny_block_list (2500) w/ staticvariables (8)       2.4 MB (0.13x)     0.069665 (0.13x)

Of note in the table above is the case of using scalar components, which requires roughly 9x less memory and initialization time with kernel code, and the case of using an indexed set of blocks with a small number of components on each, which requires roughly 6x less memory and initialization time with kernel code (e.g., indexed `piecewise`).

Note that the above table shows an _increase_ in memory usage by a `variable_dict` with 2500 entries compared to a standard indexed `Var` object of the same size from the AML interface. This is due to the use of `collections.OrderedDict` to manage the underlying storage in the `variable_dict` implementation. This can be switched to use a plain `dict` by passing `ordered=False` when initializing the `variable_dict`. There is likely room to improve the implementation to reduce memory and remain ordered, but I don't see much worth in doing this as the memory used by containers like this is likely not a bottleneck for any real models. On a related note, I read that the built-in `dict` starting in Python 3.6.1 has ordered behavior, so the use of `collections.OrderedDict` might not be necessary in the long term.

New Functionality
------------------

The more explicit syntax used by the kernel containers allows for seamless support of user-defined classes that are derived from existing components or containers. Example:
```python
import pyomo.kernel as pmo

# a user-defined extension of the kernel 'block' type
class Widget(pmo.block):
    def __init__(self):
        super(Widget, self).__init__()
        ...
    def foo(self):
        ...

# Widgets can be placed on a block and stored in block containers
m = pmo.block()
m.w = Widget()
blist = pmo.block_list()
for i in range(10):
    blist.append(Widget())
    blist.append(pmo.block())
    blist.append(pmo.tiny_block())
```

The following is a list of new modeling tools in `pyomo.kernel` that do not yet exist in the AML interface:

  - `tiny_block`: A block implementation that uses more memory-efficient storage for the case when there are only a few child objects (used by `piecewise`).
  - `linear_constraint`: A faster and more memory efficient representation of a linear constraint that directly takes a list variables and coefficients as input (also used by `piecewise`). Example:
```python
import pyomo.kernel as pmo

x = pmo.variable()
# x == y + 1
c = pmo.linear_constraint(terms=[(x, 1), (y,-1)], rhs=1)
```
  - `piecewise_nd`: An extension of the piecewise library that can be used to model multi-variate piecewise linear functions. Requires NumPy and SciPy. Example can be found [here.](https://github.com/Pyomo/pyomo/blob/component_storage_interface/examples/kernel/piecewise_nd_functions.py)
  - `matrix_constraint`: A convenient container for a set of linear constraints defined by a matrix-vector product (e.g., `L <= Ax <= U`). Requires NumPy and SciPy. Example:
```python
import pyomo.kernel as pmo
import numpy
import scipy
import scipy.sparse

# Ax <= b
# x >= 0
m = 100, n = 1000
x = pmo.variable_list(pmo.variable(lb=0) for i in range(n))
A = scipy.sparse.random(m, n, density=0.01,format='csr')
b = numpy.random.uniform(size=m)
cons = pmo.matrix_constraint(A, ub=b, variable_order=x)
```
  - `noclone`: A lightweight object that can be used when building multi-line expressions to disable cloning. This should only be used after profiling and, obviously, verifying that it is safe to avoid cloning. Unlike the `expression` component, the `noclone` object is not mutable and there is no need to store it on a block. Example:
```python
x = pmo.variable()
e = pmo.noclone(x-1)
c = pmo.constraint(body=e + e**2 + e**3, rhs=1)
```
  - `data_expression`: A mutable expression that is only allowed to store non-variable expressions. This makes it safe to use as variable or constraint bounds.

Effect on Code Outside of `pyomo.core.kernel`
-----------------------------------------------

Some changes to the solver plugins and problem writers were required to make them compatible with kernel models. These changes are minimal, and should have no effect on models built with the existing AML code in `pyomo.core.base`.

The core expression system and virtual domain types have been moved over to the `pyomo/core/kernel` directory. In order to minimize the impact this would have outside of `pyomo/core/base`, any files associated with this in `pyomo/core/base` have been changed to simply redirect the import to the associated file in `pyomo/core/kernel`. Files in `pyomo/core/base` that reference the kernel directory include:

 - expr.py
 - expr_common.py
 - expr_coopr3.py
 - expr_pyomo4.py
 - numvalue.py
 - set_types.py
 - symbol_map.py
 - suffix.py (imports `ComponentMap` base class from kernel)
 - objective.py (imports `minimize` and `maximize` from kernel)

The one change on this branch that has the potential to affect existing AML-based models is a change to the class hierarchy of the `_VirtualSet` objects (e.g., `Reals`, `NonNegativeReals`). These originally inherited from `SimpleSetBase`, but the unit tests that were implemented basically tested that 95% of the inherited `SimpleSetBase` methods raised an exception. For this reason, I don't foresee any actual negative impact on existing user code that could come from breaking this inheritance (but merging this branch to master will help us determine that). That inheritance no longer exists on this branch, which allows the `_VirtualSet` objects (set_types.py) to live in `pyomo/core/kernel`, while the `Set` objects (sets.py) remains in `pyomo/core/base`. I propose renaming set_types.py to domain_types.py in the future.

Testing and Solver Compatibility
--------------------------------

All new code in `pyomo/core/kernel` is 100% covered via unit tests.

Additionally, I have translated a significant number of "solver test" models into kernel form and expanded that test suite.  Kernel-based models seem to be fully compatible with the .lp writer, .mps writer, .nl writer, .bar writer, and the direct solver interfaces.

Future Work
------------

For now, this code lives alongside the existing AML interface in `pyomo/core/base`. The future plan is to re-implement the AML modeling objects on top of the code in `pyomo/core/kernel`. Some time before that happens, we need to do a code review of the component interfaces in `pyomo/core/kernel`.


jsiirola:
Some (initial) thoughts:

1. I am a bit bothered by the amount of "base class implementation" that is replicated in all the derived classes in order to support abstract bases and mixins.  I see this as being a long-term maintenance problem, as it not only violates DRY, but if we want to change the storage implementation in a base class we will have to update *ALL* of the derived classes.  I am wondering if we can eliminate it through a couple "tricks":
    - can we have base classes define "`__metaslots__`"?  That is, a class-level list of "forward slot declarations"? This way slot definitions slot that conceptually belong in the base class would actually be declared there.
    - to initialize the slot list in the final derived class, we can provide a `compile_metaslots` method that walks the MRO in much the same way as the new `__getstate__` implementation and puts together the list of slots.
    - the slots would be initialized by the base classes in their normal `__init__` method.
    - Here is my draft implementation:

```python
def compile_metaslots(*classes):
    # Note: track both metaslots and slots so that if a derived class
    # inherits from a base class we do not accidentally re-declare
    # slots.
    metaslots = set()
    slots = set()
    for cls in classes:
        for base_cls in cls.__mro__:
            metaslots.update(getattr(base_cls,'__metaslots__',()))
            slots.update(getattr(base_cls,'__slots__',()))
    return tuple(metaslots - slots)

class A(object):
    __slots__ = ()

class B(object):
    __slots__ = ()
    __metaslots__ = ('b',)

    def __init__(self):
        super(B, self).__init__()
        self.b = 5

class C(B, A):
    __slots__ = ()
    __metaslots__ = ('c',)

    def __init__(self):
        super(C, self).__init__()
        self.c = 6

class D(C):
    # Note that compile_metaslots is passed all the base classes of this class
    __slots__ = compile_metaslots(C) + ('d',)

    def __init__(self):
        super(D, self).__init__()
        self.d = 7

print compile_metaslots(C) # (b, c)
print compile_metaslots(D) # () [since D declares all the metaslots as slots!

x = D()
print( dir( x ) )
print( x.b )
print( x.c )
print( x.d )
```

2. In the long-term vision, what are your thoughts about renaming the Ixxx interfaces to replace the existing AML classes?  (e.g., IVariable -> Var)?  As the AML classes are really just re-directors to the Simple*/Indexed* implementations, we could add that to the Ixxx interfaces by just declaring `__new__`?  This would fix the circular imports and keep the conceptual meaning of ctype clear.

3. Do we really need the create_XXX_dict and create_XXX_list helper functions?  These seem like alternate constructors - which would make more sense implemented as class methods.

```python
class generic_dict(dict):
    @classmethod
    def from_keys(cls, keys, *args, **kwds):
        scalar = cls._default_scalar_type
        ans = cls()
        for key in keys:
            ans[key] = scalar(*args, **kwds)
        return ans

class int_dict(generic_dict):
    _default_scalar_type = int

print( int_dict.from_keys([1,2,3], 5) )
```

4.  I'd like to drag the rest of the team into a discussion around naming.  I had advocated "child" and "parent" back when I was designing Blocks and others had objected it as being "to CS-y".  Having child, parent, container, and component is precise, but it is much more "CS-y".  In addition, your use of "component" is different from the current Pyomo AML: in the current AML, a "Component" is a container, and an individual thing (e.g. a variable) is a Component Data.  As I read the Kernel, a component is a single thing (e.g., a single variable).  I think I like this change, but it is a change and I wonder about compatibility with documentation in the Book.

ghackebeil:
Thanks for taking a look. Don't take the length of my responses as anything other than my desire to document why I landed on some of these decisions. I think your points are valid and worth discussing further.

1) **metaslots, etc.**

I like these ideas for the most part. As for whether we should place an `__init__` method on the lower-level base classes, I think this could use some more discussion. Recall that _all_ of the core components in the AML interface basically copy-pasted the base class constructors into their own for performance reasons. Now this alone might not be reason enough to avoid an `__init__` method on the base class (for all I know, this could have been someone's premature optimization), but I'd like to confront this historical fact head-on in the base class design rather than incrementally converging to what is currently in `core/base`. In that scenario, the naive developer might _think_ they only need to update some base class's `__init__` method, but they would be wrong, so so wrong. In this scenario, there is a strong deterrent to tinkering with the base class, and there is a clear understanding of what it will require doing.

Besides, I don't see much difference between a virtual base class requiring a derived class define a method vs. requiring a derived class define an attribute/property/slot in a way that is documented somewhere on the base class. I originally enforced this using the ABCMeta magic with properties (e.g., `_parent` was an abstract property for `ICategorizedObject`), but that ended up making it difficult to _not_ use `__slots__` when defining a complete implementation. In the current design, the first complete implementations need to handle declaration of all abstract attributes/slots/properties/methods documented on the base classes. Sub-classing a complete implementation then looks like what one would see in more standard Python class hierarchies, you call the base class constructor. Thus, things are easy for folks who sub-class things like `variable`, `variable_dict`, etc., and a little more knowledge of the base class organization is required to implement completely new component categories. At its worst, one needs the lines `_parent = None` and `_active = True`, and failing to declare those would make the object unusable in any real context.

FYI, the naming rules I've been using so far are that classes whose name starts with an "I" (e.g., `ICategorizedObject`) or whose name contains "Mixin" (e.g., `_ActiveComponentMixin`) do _not_ have an `__init__` method, so sub-classes do not need to call them in their constructor.

2) **IVariable -> Var, using `__new__`**

I'm trying to avoid using `__new__` this low in the interface. It introduces serious complications down the line for sub-classing, and that is one thing that we need to make less complicated.


3) **replace `create_variable_dict`, etc. with a class method, and assign a default type to the containers**

First, a little background for someone reading this. I very recently added a few utility functions to "shorten" creation of some variable containers (just a small experiment). E.g.,

```python
# long form
x = variable_list()
for i in range(10):
    x.append(variable())
y = variable_dict()
for i in range(10):
    y[i] = variable(lb=1, value=0)

# short-ish form
x = variable_list(variable() for i in range(10))
y = variable_dict((i,variable(lb=1, value=0)) for i in range(10))

# short form
x = create_variable_list(10) # first arg is size
y = create_variable_dict(range(10), lb=1, value=0) # first arg is set of keys
```

These utility functions also accept a keyword named `type_` that one can assign some other class with the same `ctype` (e.g.,  `create_variable_list(10, type_=another_variable_type)`).

First, I'm not entirely against what John is proposing. I almost went with that before pushing these functions, but let me explain why I decided not to. I'm trying to keep a strict separation between the implementation of a container and the implementation of what it can contain. What convinced me to go with a separate function for now was (a) that I really just wanted to see how much I would actually use these alternate constructor forms (rather than the expanded initialization syntax using a generator) and (b) when trying to implement what was proposed for the dict container, I ran into the issue of how to deal with the keyword clashes between the container and what it contains. For instance, `ComponentDict` accepts an 'ordered' keyword. What if the type it's asked to contain uses the same keyword? There is no completely correct way to handle this inside of an alternate constructor like this, and I don't want to start introducing edge cases this early.

Second, IMO, what you are proposing starts to put us back on the slippery slope of having to define a new container each time you sub-class a modeling component. I would almost rather delete these functions and leave it up to the AML layer to add functionality like this. These alternate constructors  really don't introduce any new functionality, so I don't think we should sacrifice any simplicity of the class design just to support them (at least without some more prototyping and testing).

PR #133, opened 2017-04-11
Constraint default fix

### LINKED ISSUE #9: Derived component IndexConstraint failed to define _default()
Got the following error trying to refer to a Constraint that I skipped definition for.

```
File "/home/qtothec/pyomo/src/pyomo/pyomo/core/base/indexed_component.py", line 516, in _default
  % (self.__class__.__name__,))
NotImplementedError: Derived component IndexedConstraint failed to define _default().
    Please report this to the Pyomo developers
```

It's quite easy to produce this error:

```
from pyomo.environ import *
m = ConcreteModel()
m.x = Var()
def _c(b, i):
    if i == 1:
        return m.x == 1
    else:
        return Constraint.NoConstraint
m.c = Constraint([1, 2, 3], rule=_c)
print(m.c[2])
# or really anything else trying to refer to that _ConstraintData
# is going to give the above error
```

I'm not opposed to getting an error message in this case, since it was an error on my part, but the message should probably say something different.

Related to this, if I'm iterating over m.c._index, would be best way to check for existence be to use `indx in m.c`?

jsiirola:
This fixes #9 (Derived IndexedConstraint failed to implement _default).  More generally, this also cleans up several components' APIs to make them more standardized.  This includes moving the `Constraint.__delitem__()` method to the `IndexedComponent` base class, making `__setitem__()` generally leverage the `__getitem__()` logic (for index flattening, etc.), and updating the writers to handle Constraints with `None` body expressions.


jsiirola:
Notes from the Developers' Meeting
  - Accessing a member that is not in the `_data` dictionary should raise an exception (KeyError).  This should also apply to mutable Params without a default.  This includes `m.c[i].set_value(...)` if `i` is not in `m.c._data`.
  - Setting an undefined member should work (in the same spirit as `dict`).

pyomo/core/base/constraint.py
```diff
@@ -706,13 +706,14 @@ def __init__(self, *args, **kwargs):
         kwargs.setdefault('ctype', Constraint)
         ActiveIndexedComponent.__init__(self, *args, **kwargs)
 
-    def _setitem(self, idx, val, new=False):
+    def _setitem_if_not_present(self, idx, val, new=False):
         if self._check_skip_add(idx, val) is None:
```
qtothec: I can see future me asking you: what's going on here?

doc/OnlineDocs/tutorial/index.rst
```diff
@@ -29,6 +29,72 @@ Concrete vs Abstract Models
 
 Modeling Components
 -------------------
+The primary modeling components defined by Pyomo are
+:class:`Set<pyomo.core.base.sets.Set>`, :class:`Var<pyomo.core.base.var.Var>`,
```
whart222: Did you intentionally omit Param?  If so, then I think it makes sense to also omit Set.  Both Set and Param provide data abstractions that are not strictly necessary to build Pyomo models.

doc/OnlineDocs/tutorial/index.rst
```diff
@@ -29,6 +29,72 @@ Concrete vs Abstract Models
...
+   ...     return m.x[i] >= m.y[i]
+   >>> m.c = pe.Constraint(m.a, rule=c_rule)
+   >>> m.c.pprint()
+   c : Size=3, Index=a, Active=True
+       Key : Lower : Body        : Upper : Active
+         1 :  -Inf : y[1] - x[1] :   0.0 :   True
+	 2 :  -Inf : y[2] - x[2] :   0.0 :   True
+	 3 :  -Inf : y[3] - x[3] :   0.0 :   True
+
+By default, the index set is fixed. New indices may not be added:
```
whart222: the index set -> an index set

doc/OnlineDocs/tutorial/index.rst
```diff
@@ -29,6 +29,72 @@ Concrete vs Abstract Models
...
+.. doctest::
+   
+   >>> m.z = pe.Var()
+   >>> m.c[4] = m.x[1] == 5 * m.z
+   Traceback (most recent call last):
+     ...
+   KeyError: "Index '4' is not valid for indexed component 'c'"
+
+This helps prevent many common mistakes. However, IndexedComponent can behave
+more like a dictionary if Any is used as the index.
```
whart222: I think we should document IndexComponent and Any as classes:   :class:`IndexComponent`  and :class:`Any`.

doc/OnlineDocs/tutorial/index.rst
```diff
@@ -29,6 +29,72 @@ Concrete vs Abstract Models
...
+   >>> m.c2[8] = m.x[2] == m.z * m.y[2]
+   >>> m.c2.pprint()
+   c2 : Size=2, Index=Any, Active=True
+       Key : Lower : Body            : Upper : Active
+         1 :   0.0 :      x[1] - 5*z :   0.0 :   True
+	 8 :   0.0 : x[2] - z * y[2] :   0.0 :   True
+   
+.. note::
+   
+   In order to use Any with an IndexedVar, set dense=False.
```
whart222: What is IndexVar?  A Pyomo class?  It's first used here without definition.

pyomo/core/base/constraint.py
```diff
@@ -705,6 +707,29 @@ def __init__(self, *args, **kwargs):
         kwargs.setdefault('ctype', Constraint)
         ActiveIndexedComponent.__init__(self, *args, **kwargs)
 
+    #
+    # TODO: Ideally we would not override these methods and instead add
+    # the contents of _check_skip_add to the set_value() metthod.
```
whart222: metthod

pyomo/core/base/expression.py
```diff
@@ -361,29 +361,12 @@ def store_values(self, new_values):
...
-                "Cannot set the value of Expression '%s' with "
-                "invalid index '%s'"
-                % (self.name, str(ndx)))
-        #
-        # Set the value
-        #
-        exprdata.set_value(val)
+    def _getitem_when_not_present(self, index):
+        # TBD: Is this desired behavior?  I can see implicitly setting
+        # an Expression if it was not ooriginally defined, but I am less
```
whart222: ooriginally

pyomo/core/base/expression.py
```diff
@@ -361,29 +361,12 @@ def store_values(self, new_values):
...
-        if exprdata is None:
-            raise KeyError(
-                "Cannot set the value of Expression '%s' with "
-                "invalid index '%s'"
-                % (self.name, str(ndx)))
-        #
-        # Set the value
-        #
-        exprdata.set_value(val)
+    def _getitem_when_not_present(self, index):
```
whart222: FWIW, I don't think it makes sense to implicitly create a named expression.  Vars are a special case in many ways.  If we think we need this, we should have motivating examples.

pyomo/core/base/indexed_component.py
```diff
@@ -25,22 +25,56 @@
...
 normalize_index.flatten = True
 
+#
+# Get the fully-qualified name for this index.  If there isn't anything
+# in the _data dict (and there shouldn't be), then add something, get
+# the name, and remove it.  This allows us to get the name of something
+# that we haven't added yet without changing the state of the constraint
+# object.
+#
+def _get_indexed_component_data_name(component, index):
```
whart222: Documentation within this method would be useful.  It's not immediately obvious why you're constructing and then deleting a constraint.

pyomo/core/base/indexed_component.py
```diff
@@ -447,89 +490,190 @@ def itervalues(self):
...
+            index = self._processUnhashableIndex(index)
+            if index.__class__ is _IndexedComponent_slicer:
+                return index
+
+        if obj is None:
+            # Not good: we have to defer this import to now
+            # due to circular imports (expr imports _VarData
+            # imports indexed_component, but we need expr
+            # here
+            from pyomo.core.base import expr as EXPR
```
whart222: With pyomo5, you won't have to defer this import.   You'll be able to add:

  from pyomo.core.expr import current as EXPR

pyomo/core/base/objective.py
```diff
@@ -267,6 +278,30 @@ def __init__(self, *args, **kwargs):
         kwargs.setdefault('ctype', Objective)
         ActiveIndexedComponent.__init__(self, *args, **kwargs)
 
+    #
+    # TODO: Ideally we would not override these methods and instead add
+    # the contents of _check_skip_add to the set_value() metthod.
```
whart222: metthod

pyomo/core/base/param.py
```diff
@@ -710,8 +707,7 @@ def _initialize_from(self, _init):
             # will take the "less surprising" route of letting the
             # source become dense, so that we get the expected copy.
             #
-            # TODO: Establish use-cases where we can use default values
-            # for sparsity.
+            # TBD: Are there use-cases where we want to maintain sparsity?
```
whart222: I think you should create a ticket to follow-up on this action.  The need for sparsity seems more obvious to me than it does to you.

pyomo/core/base/param.py
```diff
@@ -829,7 +830,9 @@ def construct(self, data=None):
         if data is not None:
             try:
                 for key, val in iteritems(data):
-                    self[key] = val
+                    #self[key] = val
```
whart222: Delete?

pyomo/core/base/param.py
```diff
@@ -908,7 +913,16 @@ def __call__(self, exception=True):
         Return the value of this parameter.
         """
         if self._constructed:
-            return _ParamData.__call__(self, exception=exception)
+            if not self._data:
```
whart222: So an indexed component can be constructed but self._data could be none?  That's surprising.  I think the indexed component documentation should include a discussion of how different construction "modes" initialize the core data structures.

pyomo/core/base/var.py
```diff
@@ -735,7 +738,7 @@ class SimpleVar(_GeneralVarData, Var):
 
     def __init__(self, *args, **kwd):
         _GeneralVarData.__init__(self,
-                                 domain=Reals,
+                                 domain=None,
```
whart222: What is the intent of this change?  If no domain is specified, then ...?

pyomo/core/base/var.py
```diff
@@ -916,26 +919,38 @@ class VarList(IndexedVar):
...
-        self._nvars = 0
         IndexedVar.__init__(self, *args, **kwds)
 
     def construct(self, data=None):
         """Construct this component."""
         if __debug__ and logger.isEnabledFor(logging.DEBUG):
             logger.debug("Constructing variable list %s", self.name)
-        IndexedVar.construct(self, data)
+
+        # We need to ensure that the indices needed for initialization are
```
whart222: I'm not quite sure what is going on in the logic below, but I'm worried that we're looping through all indices twice and constructing data up-front.  This sounds expensive.  I think there's a potential runtime and memory hit here.   Have you confirmed that that's not the case?

Can you comment on the difference between the previous design and the current one for Var?

pyomo/core/base/expression.py
jsiirola: I agree.  I put it in because that is how things currently work.  However, I thing currently work that way because in the old system we did not distinguish between being able to implicitly _create_ a `_ComponentData` when it was not already present and being able to implicitly _get_ a `_ComponentData` when it was not already present.  For Expression components, I believe that the former is fine and the latter makes no sense.  I am hoping that @ghackebeil will comment.  If he is OK with it, I would like to remove `Expression._getitem_when_not_present()`.

pyomo/core/base/var.py
jsiirola: Exactly.  It turns out that that initialization is a no-op.  I was banging my head as to why we were setting all scalar vars to be Reals.  The default value comes from the `Var.__init__` below.  I removed the `Reals` here to make it more obvious that *this* method was not responsible for setting the domain.

pyomo/core/base/var.py
jsiirola: Yes.  The change is because we are now actually validating indices for VarList.  In the past, you could provide anything as keys when initializing a VarList using a dict and we would happily chug along.  Now that we are validating that VarList keys against the underlying index set, we need to make sure that the underlying index set is correct.  There is no impact on memory: you end up with the same number of variables as before.

whart222: I see.  This is just for VarList, which makes sense.

pyomo/core/base/expression.py
whart222: I think you'll need to bug him for input if you want it before the holidays.

whart222:
@jsiirola Addressed my comments that represented real concerns.  I think the remaining action is to document the code design in the Sphinx documentation.

whart222:
I've approved the changes.  Carl added a ticket for the outstanding issue that I had concerns about:  documentation (see #287).  I think the right thing to do is to merge this PR, but @carldlaird  recommended that we let @jsiirola do that.

PR #139, opened 2017-04-26
Dapsdev

DLWoodruff:
Minor change to the csvwriter plugin in PySP to move a function out of the class so it can be called by daps.
Changed daps so the ef solver can output.

### DIFF
--- a/pyomo/pysp/daps/demodaps.py
+++ b/pyomo/pysp/daps/demodaps.py
@@ -16,7 +16,7 @@ stsolver = st.StochSolver('cref.py', tree_model)
 stsolver.solve_ef('cplex')
 # the stsolver.scenario_tree has the solution
 csvw.write_csv_soln(stsolver.scenario_tree, "testcref")
-kadsjhf
+
 ### simple tests ####
 print("\n*** 2Stage_AMPL")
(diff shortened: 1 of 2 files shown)


whart222:
I'm surprised that this reduces code coverage so much.  Is there an explanation?

DLWoodruff:
I don't have an explanation, but I do have more questions: why did the
number of files go down so much and the lines as well? The daps code itself
has tests with better than 50% coverage, btw.

ghackebeil:
I believe we merged the dapsdev branch into master last week or so, and maybe forgot to tell @DLWoodruff.

By pushing to this branch (rather than master), Github has decided to reopen a new pull request with the same name (but a different number), so that's probably the source of any weirdness.

As for the coverage, I think it drops because a bunch of new .py files were added in the original pull request, and maybe they are not being covered by tests.

DLWoodruff:
Sorry to cause this trouble, but I have to admit that I don't understand
what is going on. I did know that dapsdev was merged last week, so when I
wanted to do some minor changes I
0) merged master to dapsdev (or at least thought I did).
1) made my changes on my machine
2) pushed to dapsdev
3) made a pull require to merge dapsdev with master with the idea that my
minor changes would be eventually merged into master.


On Wed, Apr 26, 2017 at 1:12 PM, Gabriel Hackebeil <notifications@github.com

Issue #122, opened 2017-03-08
Error getting NEOS results


bpierpont:
New to Pyomo, but in 5.1.1 on both Mac and Windows, I am unable to get a NEOS job to return results. I've tested with several examples and with SolverStudio models (which solve successfully with local CBC solver).

Using CBC on NEOS, I get the following error:

> ---------- Begin Solver Output -----------
> Condor submit: 'neos.submit'
> Condor submit: 'watchdog.submit'
> Job submitted to NEOS HTCondor pool.
> [    8.99] Pyomo Finished
> ERROR: Unexpected exception while running model:
> 	local variable 'opt' referenced before assignment

Using CPLEX on NEOS, I get the following error:

> ---------- Begin Solver Output -----------
> Condor submit: 'neos.submit'
> Condor submit: 'watchdog.submit'
> Job submitted to NEOS HTCondor pool.
> [   17.02] Pyomo Finished
> ERROR: Unexpected exception while running model:
> 	Cannot load a SolverResults object with bad status: error

I've seen others reference this bug as well (https://groups.google.com/forum/#!topic/pyomo-forum/CRzG4l-3E2A).

bpierpont:
Confirmed that uninstalling Pyomo 5.1.1 and installing 4.4.1 fixes this issue.

jsiirola:
The problematic commit was reverted in c667585 (on 22 Jan 17).  It will be available in the 5.2 release.

PR #147, opened 2017-05-13
Neos fixes: timeouts, error messages, CPLEX results

### LINKED ISSUE #125: NEOS Gateway Timeout
I have been getting the following exception running jobs on NEOS in pyomo 5.1.1 in both Ubuntu and Windows:
```<ProtocolError for www.neos-server.org:3333/RPC2: 504 Gateway Time-out>```

I believe this is due to a time-out in the call to ```self.kestrel.neos.getIntermediateResults()``` here: https://github.com/Pyomo/pyomo/blob/79404732c451ec9f143dbdd47d2e372ac1695d8d/pyomo/neos/plugins/kestrel_plugin.py#L205-L215

My fix to this problem has been to check ```self.kestrel.neos.getJobStatus()``` until the job is no longer in 'Waiting' status and then proceed with  the call to ```self.kestrel.neos.getIntermediateResults()```. This seems to fix the problem for me, but it relies on checking at a regular time interval and repeated queries to NEOS. I don't believe this is the best solution. Any suggestions?

jsiirola:
This has become a bit of an omnibus PR to fix problems with the NEOS interface:

* Define a log handler so we see logging information from namespaces other than `pyomo.core` and `pyomo.opt`
* Provide extra debugging information when things go bad
* remove the `timelimit` option from the solver options when its value is `None` (the default).  This fixes problems with running CPLEX without an explicit time limit (fixes the last of #122)
* improved proxy settings detection
* timeouts getting intermediate solutions for long-running jobs should not be fatal (fixes #125)

### DIFF
--- a/pyomo/neos/__init__.py
+++ b/pyomo/neos/__init__.py
@@ -33,3 +33,5 @@ doc = {
 'path':     'Nonlinear MCP solver',
 'snopt':    'SQP NLP solver'
 }
+
+import pyomo.neos.log_config
--- a/pyomo/neos/kestrel.py
+++ b/pyomo/neos/kestrel.py
@@ -66,9 +66,19 @@ class kestrelAMPL:
         self.setup_connection()
 
     def setup_connection(self):
-        if 'HTTP_PROXY' in os.environ:
+        # on *NIX, the proxy can show up either upper or lowercase.
+        # Prefer lower case, and prefer HTTPS over HTTP if the urlscheme
+        # is https.
+        proxy = os.environ.get(
+            'http_proxy', os.environ.get(
+                'HTTP_PROXY', ''))
+        if urlscheme == 'https':
+            proxy = os.environ.get(
+                'https_proxy', os.environ.get(
+                    'HTTPS_PROXY', proxy))
+        if proxy:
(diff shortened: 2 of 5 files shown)


whart222:
John:  Can you comment on the tests that you did to verify this patch.  We don't have automated tests, so I think it's appropriate to document the testing informally.

whart222:
I think we can accept this patch even with the code coverage decline.  Testing the NEOS solver interface cannot be easily automated.

jsiirola:
@whart222: I did the following tests:
* `pmedian.py` + `pmedian.test1.dat` with CBC and cplex - everything works
* A long-running MIP based on the QAP with a 300-second timeout to test the polling timeout
  * with cplex everything worked well and terminated normally, returning a partial solution
  * CBC didn't recognize the timelimit option (and ignored it).  I eventually killed the job through kestrel and verified that our interface gracefully returned no solution

As for testing, I thing that we might be able to perform limited testing with NEOS (2-3 targeted tests with short timeouts) to verify functionality from Travis.  We cannot test it from Jenkins as the proxy blocks ports 3333 & 3332.  I will work on putting together something now.

PR #149, opened 2017-05-26
IndexedConstraint index_set() bug

michaelbynum:
When adding or deleting a constraint to/from an IndexedConstraint, the _index attribute was not updated. Thus the index_set() method returned an incorrect set. I updated the add() and __delitem__ methods to updated the _index attribute. My changes appear to fix the problem. However, please review my changes to ensure I did not miss anything.

P.s. - There are many unrelated commits where I was testing a persistent gurobipy interface. However, I deleted all of those changes - I am moving them to another branch.

Michael


ghackebeil:
I don't think this will work Michael.

The biggest problem is that the `._index` attribute points to a `Set` object that is likely shared among many indexed components. So by updating it for one of them, you do it for all of them, which is probably not the intention.

The second problem is that sometimes this points to a _virtual_ `Set` object (e.g., a cross product between two or more sets), and you are not allowed to modify those. I think this one, in particular, is why this branch has a test failure on travis.

FYI, there are other scenarios in which the `._index` set can deviate from the keys in the `._data` dictionary. For example, when you return `Constraint.Skip` from a constraint rule, that index will appear in the `._index` set but not in the `._data` dictionary.

michaelbynum:
Okay, thanks Gabe. Should the index_set() method be public then?

ghackebeil:
Good question. I suppose someone might have some use for it. @jsiirola probably knows more.

PR #151, opened 2017-06-01
Add exponential function support to symbolic module and add documentation

qtothec:
[Edit]: Also found out that for some reason, exp() wasn't in the operator map yet. This should fix that omission.

Documentation on differentiate() because I've had to explain to several people now what the function does.


jsiirola:
I discussed this PR with @ghackebeil, and we both agree it is fine.  The submitted test (much appreciated, @qtothec!) is failing due to sympy converting integers to floats; however, it is easier for me to accept the PR and correct the test than to iterate with Qi.

PR #152, opened 2017-06-01
Updating IDAES branch of Pyomo

qtothec:
My code is going to fail tests if the fix to symbolic.py isn't merged in, because I switched to using the Pyomo differentiate() function rather than going through CasADi.

Attn:
@eslickj
@andrewlee94
@tony121psu
@dangunter


jsiirola:
@qtothec:  PRs to the IDAES branch should come from Pyomo/pyomo:master.  The testing infrastructure is currently getting confused because you have 2 PR's that point to the same branch on your clone.  I am closing this (and will be merging #151 in soon).  Then we can open a new PR fro Pyomo's master.

Issue #153, opened 2017-06-02
"list index out of range" error when using inverse of a variable


andrewlee94:
We have been running into this error in some of our models when writing the using the inverse of a variable in our models. If we write the inverse as 1/model.var, we get the error and traceback shown at the end. However, if we replace the inverse with model.var**-1 the problem is fixed. One common thread I have noticed is that the offending variable has always (so far) appeared in a Block.

I have a simple example model which demonstrates the bug, which can be found at:
https://github.com/IDAES/models/tree/unit_models_v2/idaes_models/unit/standard/pyomo_bug
Running the bug_demo.py file as a script resulted in the following traceback for me.

Traceback (most recent call last):
  File "bug_demo.py", line 62, in <module>
    main()
  File "bug_demo.py", line 53, in main
    fs_obj.solve(tee=stee,options={'tol':1e-10})
  File "/home/alee/IDAES Repository/models-master/idaes_models/core/process_base.py", line 923, in solve
    self.pyomo_results = self.pyomo_solver.solve(m, *args, **kwargs)
  File "/home/alee/anaconda2/lib/python2.7/site-packages/pyomo/opt/base/solvers.py", line 581, in solve
    self._presolve(*args, **kwds)
  File "/home/alee/anaconda2/lib/python2.7/site-packages/pyomo/solvers/plugins/solvers/ASL.py", line 188, in _presolve
    SystemCallSolver._presolve(self, *args, **kwds)
  File "/home/alee/anaconda2/lib/python2.7/site-packages/pyomo/opt/solver/shellcmd.py", line 195, in _presolve
    OptSolver._presolve(self, *args, **kwds)
  File "/home/alee/anaconda2/lib/python2.7/site-packages/pyomo/opt/base/solvers.py", line 660, in _presolve
    **kwds)
  File "/home/alee/anaconda2/lib/python2.7/site-packages/pyomo/opt/base/solvers.py", line 728, in _convert_problem
    **kwds)
  File "/home/alee/anaconda2/lib/python2.7/site-packages/pyomo/opt/base/convert.py", line 109, in convert_problem
    problem_files, symbol_map = converter.apply(*tmp, **tmpkw)
  File "/home/alee/anaconda2/lib/python2.7/site-packages/pyomo/solvers/plugins/converter/model.py", line 163, in apply
    io_options=io_options)
  File "/home/alee/anaconda2/lib/python2.7/site-packages/pyomo/core/base/block.py", line 1623, in write
    io_options)
[...]

Issue #73, opened 2016-11-21
AMPLFUNC Environment Variable


eslickj:
It seems that when you call an ASL solver multiple times from a Python script, and you use ExternalFucntion, the AMPLFUNC environment variable accumulates a bunch of repeated copies of the .so file.  I haven't determined if this causes a real problem yet, but it causes the ASL to spew a bunch of warnings about already defined functions.  I printed the environment variables during an initialization consisting of several solves, so you can get an idea of what's happening

'PYOMO_AMPLFUNC': '../../../properties/physical/MEA_Simple/phys_prop.so'

'AMPLFUNC': '/imports/users/jeslick/git/eslickj/models/idaes_models/properties/physical/MEA_Simple/phys_prop.so\n../../../properties/physical/MEA_Simple/phys_prop.so\n../../../properties/physical/MEA_Simple/phys_prop.so\n../../../properties/physical/MEA_Simple/phys_prop.so\n../../../properties/physical/MEA_Simple/phys_prop.so\n../../../properties/physical/MEA_Simple/phys_prop.so\n../../../properties/physical/MEA_Simple/phys_prop.so\n../../../properties/physical/MEA_Simple/phys_prop.so'

whart222:
Can you add a ticket for this?  This looks like a bug.

eslickj:
Sorry, I'm pretty new to how github works.  I made an issue.  How do I go about adding a ticket?

whart222:
Ah.  I missed that this email was generated from an issue.  Thanks!

eslickj:
Thanks for looking into it.  I've been having some trouble with the solver failing mysteriously, and after a lot of searching this seems to be the culprit.  I did some more testing, and after a certain number of times importing the .so file the ipopt solver segfaults before even getting started.  I have no idea why that would happen. It may be something weird in my user function code.  Anyway this bug does seem to potentially cause more problems than just annoying warning messages.

eslickj:
To follow up on this issue, a couple other people working on IDAES have had problems with ipopt crashing when a library gets included in AMPLFUNC too many times, so this does seem to be an important issue.  Previously I wasn't totally sure whether or not this was a serious issue.

jsiirola:
@eslickj: can you take a look at Pyomo/IDAES-premerge (240d3f9) and see if that resolves this issue?

qtothec:
@eslickj Has this been resolved?

eslickj:
[...]

PR #154, opened 2017-06-05
Update IDAES branch from pyomo/master

jsiirola:
I am recreating @qtothec's PR (#152) now that #151 has been merged with master.

As @qtothec pointed out: IDAES/models will begin to fail tests if the fix to symbolic.py isn't merged in, because of a switch to using the Pyomo differentiate() function rather than going through CasADi.

Attn:
@eslickj
@andrewlee94
@tony121psu
@dangunter


eslickj:
I'm okay with this.  I think we will have a minor issue with ExternalFunctions related to the AMPLFUNC environment variable, but we need to work that out anyway, and it won't break any tests.  Otherwise I think this works fine for my stuff.

andrewlee94:
My only concern is the same one John has. Otherwise, I would like to update to the new master for the fix to solving indexed blocks.

jsiirola:
I am looking at #153 right now.  @andrewlee94: are you referring to #73?  I will take a look at that next.

dangunter:
I am fine with this if John and Andrew are.

I think @ksbeattie was going to be "gatekeeper", at least initially.

andrewlee94:
@jsiirola The issue we are referring to is the follow on from #73 we mentioned at the meeting. To recap, the AMPLFUNC environment variable does not seem to be getting registered for those people using newer versions of Pyomo.

ksbeattie:
I believe I did agree to be the "gatekeeper" for these PRs.  As in, I was supposed to be the one making the request.  I appear to have fallen down in that regard for this request.

As such, I'm ok with merging this PR as clearly everyone is ok with it.

I'll make a point of driving these requests in the future from the IDAES team.

PR #157, opened 2017-06-09
BARON interface test for log10

qtothec:
I didn't know how to test if BARON was available in the context of your testing infrastructure, so there's a dummy there.


pyomo/solvers/tests/checks/test_BARON.py
```diff
@@ -0,0 +1,32 @@
+"""Tests the BARON interface."""
+import pyutilib.th as unittest
+from pyomo.environ import (ConcreteModel, Constraint, Objective, Var, log10,
+                           minimize)
+from pyomo.opt import SolverFactory, TerminationCondition
+
+# I don't know how you figure out if BARON is available
+baron_available = True
```
jsiirola: You should be able to use SolverFactory("baron").available()

jsiirola:
General question: we already have several models that we run against BARON. I wonder if this should be a model in the test model library as opposed to a test specific to BARON. For example, it would be good to test all NL solvers using log10... Alternatively, if we are only interested in testing the BAR writer, maybe this should be added to the writer tests?

pyomo/solvers/tests/checks/test_BARON.py
ghackebeil: On the Baron and Gurobi plugins, in particular, it's also necessary to check if the licenses are valid. You should call both the `available()` and `license_is_valid()` methods. Another approach, that I use in many places, is to piggyback off of the list of solver test cases in `pyomo.solvers.tests.solvers`, where all of this logic is already implemented and the result is cached onto the `available` property on each test case. E.g.,

```python
from pyomo.solvers.tests.solvers import test_solver_cases
baron_available = test_solver_cases('baron', 'bar').available
```

jsiirola: @ghackebeil: for this test, it should be well within what the unlicensed version of BARON allows - so do we really need to check the license?

qtothec:
I wasn't sure where in the testing architecture this belonged, so that's why it's where it is. If it belongs somewhere else, feel free to move it, or let me know which directory it should be moved into. As long as it covers my use case, short-term me is indifferent.

pyomo/solvers/tests/checks/test_BARON.py
ghackebeil: This check will not distinguish between those two cases. Although I don't remember precisely, I believe I added this because Baron will error if it finds that your license has expired, rather than just falling back to running in an "unlicensed" state. I believe was the case for a while on Jenkins.

qtothec:
Closing and re-opening to trigger travis rebuild.

whart222:
I strongly vote against accepting this PR as-is.

Either we should (a) add a test model like this in pyomo.solvers.tests.models that includes intrinsic function tests, or (b) we should add a test in pyomo.repn.tests.baron to confirm that we're writing out intrinsic functions properly.

qtothec:
That's fine. This PR was mostly intended as supporting documentation for #158, giving a test case that demonstrates the failure.

whart222:
I have added the BAR writer tests on the expr_dev branch, so I'm going to close this PR.  I'll close #158 when we merge the expr_dev branch.

whart222:
Re-opening. per my discussion with @jsiirola .  We agreed to merge this PR onto master to support the Pyomo 5.4 release.  But per my comments above, a new solution to this issue will be added when the expr_dev branch is merged.

PR #164, opened 2017-06-13
Implementation of GAMS solver interface

qtothec:
This may be of interest to those who have GAMS solver licenses but would like to use Pyomo. Need to resolve the conflicts with master.


pyomo/core/base/label.py
```diff
@@ -9,13 +9,15 @@
...
+           'AlphaNumericTextLabeler','NameLabeler', 'CuidLabeler']
 
 import six
 if six.PY3:
-    _string = str
+    _translate = str.translate
 else:
-    import string as _string
+    import string
+    _translate = string.translate
```
qtothec: Just curious, what's the rationale behind these changes in the imports?

jsiirola: Just being more specific about the thing we were really trying to import: what we really want is the `translate()` method.  I was actually surprised this wasn't handled by `six.moves`

pyomo/core/kernel/expr_coopr3.py
```diff
@@ -638,7 +638,11 @@ def to_string(self, ostream=None, verbose=None, precedence=0):
             ostream.write("( ")
         first = True
         if self._coef != 1:
+            if self._coef < 0:
+                ostream.write("(")
             ostream.write(str(self._coef))
+            if self._coef < 0:
+                ostream.write(")")
```
jsiirola: Why do we need the extra parens?  Does GAMS not bind the negation strongly enough?

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -0,0 +1,792 @@
...
+
+        self.options:
+            output_filename=None:
+                GAMS model file. If None, will use model name. Kept if given.
+            symbolic_solver_labels=False:
+                Use full Pyomo component names rather than
+                shortened symbols (slower, but useful for debugging).
+            labeler=None:
+                Custom labeler option. Incompatible with symbolic_solver_labels.
+            solve=True:
```
jsiirola: Why do you have this argument?  If you are calling a solver's solve method, it should solve.  If you want to just write the GAMS file, you can use `model.write("foo.gms")` or `pyomo convert [...] --output foo.gms`

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -0,0 +1,792 @@
...
+
+        def var_recorder(obj):
+            ans = var_labeler(obj)
+            var_list.append(ans)
+            return ans
+
+        def var_label(obj):
+            return symbolMap.getSymbol(obj, var_recorder)
+
+        with PauseGC() as pgc:
```
jsiirola: The solver should *call* the GAMS writer and not block-copy the writer code here.

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -0,0 +1,792 @@
...
+                Custom labeler option. Incompatible with symbolic_solver_labels.
+            solve=True:
+                If False, GAMSSolver will only create and keep GAMS model file.
+            solver=None:
+                If None, GAMS will use default solver for model type.
+            mtype=None:
+                Model type. If None, will chose from lp, nlp, mip, and minlp.
+            load_model=True:
+                Load results back into pyomo model.
+            print_result=False:
```
jsiirola: This is a non-standard option...  why is this here?

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -0,0 +1,792 @@
...
+                Print summary of solution to stdout.
+            keep_files=False:
+                Keep temporary files in current directory.
+                Equivalent of DebugLevel.KeepFiles.
+                If False, GAMS model file will be deleted as well.
+                Summary of temp files can be found in _gams_py_gjo0.pf
+            holdfixed=False:
```
jsiirola: This should be incorporated into an `options` argument (that is, options that are passed to the solver)

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -0,0 +1,792 @@
...
+    def solve(self, model, tee=False):
+        """
+        Solver for pyomo models using GAMS. Converts pyomo model
+        to GAMS input file, solves in GAMS, then returns solution
+        values back into pyomo model.
+
+        Uses GAMS Python API. For installation help visit:
+        https://www.gams.com/latest/docs/apis/examples_python/index.html
+
+        self.options:
```
jsiirola: For consistency with other solvers, several (most) of these options should be explicit keyword arguments to the `solve()` method.

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -0,0 +1,792 @@
...
+#
+#  Pyomo: Python Optimization Modeling Objects
+#  Copyright 2017 National Technology and Engineering Solutions of Sandia, LLC
+#  Under the terms of Contract DE-NA0003525 with National Technology and 
+#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain 
+#  rights in this software.
+#  This software is distributed under the 3-clause BSD License.
+#  ___________________________________________________________________________
+
+from __future__ import print_function
```
jsiirola: Not necessary.  We only support 2.6+

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -0,0 +1,792 @@
...
+pyutilib.services.register_executable(name="gams")
+ 
+class GAMSSolver(pyomo.util.plugin.Plugin):
+    """A generic interface to GAMS solvers"""
+    pyomo.util.plugin.implements(IOptSolver)
+    pyomo.util.plugin.alias('gams', doc='The GAMS modeling language')
+
+    options = Options()
+
+    def available(self, exception_flag=True):
```
jsiirola: This implementation assumes that you are calling the "gams" executable.  You implementation, however, uses the Python API.  Your `available()` implementation therefore needs to return whether the GAMS Python API is importable, not if the gams executable is on the PATH.  In the future, it might be useful to support both the "direct"" (python) interface, and running GAMS through a subprocess.

pyomo/core/kernel/expr_coopr3.py
qtothec: There was a possibility that two negatives could be written out in a row. GAMS doesn't like that.

pyomo/solvers/plugins/solvers/GAMS.py
qtothec: Makes sense. One question that I have: the `_write_model` code in `pyomo/solvers/plugins/solvers/GAMS.py` makes the code in `pyomo/repn/plugins/gams_writer.py` redundant, right? Should these somehow be consolidated?

pyomo/solvers/plugins/solvers/GAMS.py
qtothec: Ignore comment above. This is what happens when I read down the list of comments one-by-one.

pyomo/solvers/plugins/solvers/GAMS.py
qtothec: Subprocess is now also supported to a degree.

@gseastream seems that it would be a good idea to build out support for some of the other features (e.g. marginals) with the subprocess method.

@jsiirola do you recommend putting the "direct" interface in a different file? Also, it doesn't quite seem 100% direct, because it still requires the *.gams file to be written.

pyomo/core/kernel/expr_coopr3.py
gseastream: Correct. For product expressions in pyomo, negative coefficients were simply written at the front of their expression, regardless of the coefficient related to this expression coming from the "higher up" expression (like a sum expression that contains a product expression, might write "blahblah - -5 * this * that == 0"). I made the change here since I figured this wasn't intended behavior, and this was the fix I thought of.

pyomo/solvers/plugins/solvers/GAMS.py
gseastream: The subprocess method already supports marginals and pretty much everything the API method does.

Questions: what do the "register_executable" and "registered_executable" lines do/check? Also, the API supports calling GAMS on a model from a string, so would it be preferred for the API version if I used a StringIO for writing instead of a file?

qtothec: @gseastream You can do it in-memory? Unless @jsiirola can think of any objections, my thought is why not?

pyomo/solvers/plugins/solvers/GAMS.py
gseastream: I suppose a few of my options arose while writing and debugging the code, and I didn't know which made sense to keep or not. I'm not attached to any in particular so I can get rid of ones that don't belong.

pyomo/solvers/plugins/solvers/GAMS.py
gseastream: Although I'd have to figure out how to do that if I'm using the model.write method, since that inherently creates a file...

qtothec: Check the options on model.write(). You ought to be able to redirect output somehow. If not, maybe add the feature in.

pyomo/solvers/plugins/solvers/GAMS.py
qtothec: No worries. It's hard to know all the conventions at the outset. If you're not using it anymore, you could probably remove this option.

pyomo/solvers/plugins/solvers/GAMS.py
jsiirola: In theory you can use the same file, however the two solver interfaces should be separate classes so that they can be registered separately with the `SolverFactory`.  One option is to make both interfaces "private" and create a third solver that redirects to the correct solver based on the `solver_io` option (see `pyomo/solvers/plugins/solvers/CPLEX.py` for an example).

That said, I don't think that we should spend too much time worrying about it: we are probably going to do a massive overhaul of the solver class hierarchy "soon" to fix some of this weirdness (which would make much of this discussion moot).

@gseastream: `register_executable` is a PyUtilib method that tells an "excutable factory" that you would like it to try and find the listed executable somewhere on the PATH.  We use it because it standardizes all the neurotic platform-specific weirdness.  `registered_executable` will return an object that among other things has the full path to the located executable, or `None` if PyUtilib couldn't find the executable.

jsiirola: @gseastream: and yes: if the Python API supports getting the model as a string, then we should use a StringIO and avoid the mess with a temporary file.

pyomo/core/kernel/expr_coopr3.py
jsiirola: OK.  This is a subtle but significant change (it is breaking some core tests of string generation).  We should think about this a little more...  It will require developing some unittests and adding them to the coopr3 and pyomo4 expression testing systems, then we should decide how to best implement this so that it is as least disruptive as possible.

pyomo/solvers/plugins/solvers/GAMS.py
gseastream: Another question about `available()`: so this should return if the relevant utility is available, which is either the gams executable or the python api? And what should be done with the result of this method? Should the `solve()` method be checking the `available()` method or something?

qtothec: @gseastream correct. It should just return True if the utility is available or False if it is not. It's used elsewhere. You can assume in `solve()` that the utility is available.

gseastream: Ok. As a side note though, it appears that I can still call `solve()` even when `available()` returns False, so I'm not sure what's checking it.

qtothec: @gseastream It's up to the user to check that a solver is available before calling solve. I wouldn't get too caught up on this matter.

pyomo/core/kernel/expr_coopr3.py
qtothec: @jsiirola: is this something that @gseastream can help out with?

pyomo/solvers/plugins/solvers/GAMS.py
gseastream: To clarify, you're suggesting an options argument where the user can include any options they would like to in the gms file? Essentially a string of any additional lines they would like?

pyomo/repn/plugins/gams_writer.py
```diff
@@ -143,56 +231,79 @@ def _print_model( self, model, output_file, solver_capability, var_list,
...
+            # operations at current values, which are None (0) by default
+            # Used to handle log and log10 violations, for example
+            try:
+                value(con.body)
+            except:
+                raise RuntimeError("While evaluating: %s\n\tProblemWriter_gams"
+                                   " encountered an error. Ensure initial"
+                                   " variable values\n\tdo not violate"
+                                   " any domains (are you using log/log10?)"
+                                 % con.name)
```
qtothec: Maybe also print out the constraint expression.

pyomo/repn/plugins/gams_writer.py
```diff
@@ -143,56 +231,79 @@ def _print_model( self, model, output_file, solver_capability, var_list,
...
-        obj = list( model.component_data_objects(Objective, active=True) )
+        obj = list(model.component_data_objects(Objective, active=True))
         if len(obj) != 1:
             raise RuntimeError(
                 "GAMS writer requires exactly one active objective (found %s)"
-                % (len(obj),))
+                % (len(obj)))
         obj = obj[0]
+
+        # Same domain violation check as above
```
qtothec: I'm confused. Why does this need to be here twice, again?

pyomo/repn/plugins/gams_writer.py
```diff
@@ -201,46 +312,62 @@ def _print_model( self, model, output_file, solver_capability, var_list,
...
+            else:
+                # Same domain violation check as above
+                try:
+                    value(v.expr)
+                except:
+                    raise RuntimeError(
+                        "While evaluating: %s\n\tProblemWriter_gams"
+                        " encountered an error. Ensure initial"
+                        " variable values\n\tdo not violate"
+                        " any domains (are you using log/log10?)"
```
qtothec: Is the log/log10 comment appropriate for variables? Also, Pyomo might already check if the variables are within their bounds... @jsiirola?

pyomo/repn/plugins/gams_writer.py
```diff
@@ -249,23 +376,155 @@ def _print_model( self, model, output_file, solver_capability, var_list,
...
+            elif varName in posInts:
+                if var.lb != 0:
+                    output_file.write("%s.lo = %s;\n" % (varName, var.lb))
+                if var.ub is not None:
+                    output_file.write("%s.up = %s;\n" % (varName, var.ub))
+            elif varName in otherInts:
+                if var.lb is None:
+                    # GAMS doesn't allow -INF lower bound for ints
+                    # Set bound to lowest possible bound in Gams
+                    output_file.write("%s.lo = -1.0E+10;\n" % (varName))
```
qtothec: Should probably warn user if this is happening.

pyomo/repn/plugins/gams_writer.py
gseastream: The logger of the value function tracks the full trace of exactly which expression caused the error and goes all the way up to the full constraint expression, so that is already handled there.

pyomo/repn/plugins/gams_writer.py
gseastream: The 3 places the domain check occur are for the constraints, the objective, and Expression objects. They are written in 3 different places because those are encountered and prepared for the output file in different places, so I needed to write checks for those groups individually. Perhaps I'll make a helper function since they're all the same.

pyomo/repn/plugins/gams_writer.py
gseastream: This is the domain check for Expressions, which also exhibit the initial value domain restriction in GAMS, so it was necessary to check them too. Pyomo does make sure variables are set within their bounds, but expressions may have a log function that it's constituent variables violate even if they're inside their bounds. As for the log/log10 comment, those are still the only functions I know of that have this restriction.

pyomo/core/kernel/expr_coopr3.py
jsiirola: I'd be happy for some additional eyes on it.  we may not be able to get around the "simple" fix that @gseastream put in.  However, I think a better option will be to look at the the way we determine precedence when generating the string representations.  The cases that I'd like us to look at are:
  - `sub(y, mul(-1,x))`
  - `sub(y, mul(-1,mul(x,x)))`
  - `sub(y, mul(-2,x))`
  - `sub(y, mul(-2,mul(x,x)))`
  - `sub(y, mul(2,x))`
  - `sub(y, mul(2,mul(x,x)))`
  - `sub(y, neg(x))`
  - `add(y, mul(-1,x))`
  - `add(y, mul(-1,mul(x,x)))`
  - `add(y, mul(-2,x))`
  - `add(y, mul(-2,mul(x,x)))`
  - `add(y, mul(2,x))`
  - `add(y, mul(2,mul(x,x)))`
  - `add(y, neg(x))`

pyomo/solvers/plugins/solvers/GAMS.py
jsiirola: @gseastream: For the record, I am not faulting you -- the way the current solver hierarchy is designed, it is almost impossible to know the "expected" options.

pyomo/solvers/plugins/solvers/GAMS.py
jsiirola: @gseastream: exactly.  Look at how the NL (ASL) and CPLEX/Gurobi interfaces work.  The idea is that "solver_options" are passed directly to the solver with no checking or validation.  That way a (knowledgeable) user can tell the solver to do whatever they need.

pyomo/repn/plugins/gams_writer.py
jsiirola: As a general rule, the following is bad practice:

```
try:
   # .....
except:
  raise <something else>

```
for two reasons:
  1. "blind" `except` clauses will catch all manner of exceptions, not just the one you are expecting.  In this case, you should probably only trap a `ValueError`
  2. Swallowing the exception and raising something new destroys the original message that was generated when throwing the exception, along with the entire stack trace below this point (which makes debugging *hard*).

If you are going to do this, you should use a logger to emit your message and reraise the original exception.

pyomo/repn/plugins/gams_writer.py
jsiirola: There are a couple things going on here.  The first is a hack because of how I elected to generate the expression strings for the GAMS writer.  Currently, `Expression` components show up as if they were proper variables in GAMS, which requires them to have valid initial values.

@gseastream: two more things: 
   1. there are other ways to get bogus initial values.  For example: `(-1)**(0.5)`.
   2. why are you doing these checks at all?  Elsewhere in Pyomo, we do not force the initial conditions to be consistent or valid when emitting the model (if AMPL does because of a bad point, that counts as a solver error and not as a writer error).

pyomo/repn/plugins/gams_writer.py
gseastream: You're right, I forgot to include the ValueError in the except statement.

As for the rest of the exception raising, most information is already included in the logger errors despite my current code raising a new exception. Here is an example of the (long) output if I just do nothing but raise the original exception:

`ERROR: evaluating object as numeric value: log10( x8 )
            (object: <class 'pyomo.core.kernel.expr_coopr3._IntrinsicFunctionExpression'>)
        math domain error
ERROR: evaluating expression: math domain error
            (expression: x10 + log10( x8 ))
ERROR: evaluating object as numeric value: x10 + log10( x8 )
            (object: <class 'pyomo.core.kernel.expr_coopr3._SumExpression'>)
        math domain error
ERROR: evaluating expression: math domain error
            (expression: log10( x10 + log10( x8 ) ))
ERROR: evaluating object as numeric value: log10( x10 + log10( x8 ) )
            (object: <class 'pyomo.core.kernel.expr_coopr3._IntrinsicFunctionExpression'>)
        math domain error
ERROR: evaluating expression: math domain error
            (expression: 2 * x6 * log10( x10 + log10( x8 ) ))
ERROR: evaluating object as numeric value: 2 * x6 * log10( x10 + log10( x8 ) )
            (object: <class 'pyomo.core.kernel.expr_coopr3._ProductExpression'>)
        math domain error
ERROR: evaluating expression: math domain error
            (expression: 2 * x6 * log10( x10 + log10( x8 ) ) + x12)
ERROR: evaluating object as numeric value: 2 * x6 * log10( x10 + log10( x8 ) ) + x12
            (object: <class 'pyomo.core.kernel.expr_coopr3._SumExpression'>)
        math domain error
  File "water_demo.py", line 141, in <module>
    m = main()
  File "water_demo.py", line 107, in main
    m.write('wawa.gms')
  File "/home/gseastre/pyomo/src/pyomo/pyomo/core/base/block.py", line 1647, in write
    io_options)
  File "/home/gseastre/pyomo/src/pyomo/pyomo/repn/plugins/gams_writer.py", line 187, in __call__
    holdfixed=holdfixed
  File "/home/gseastre/pyomo/src/pyomo/pyomo/repn/plugins/gams_writer.py", line 243, in _write_model
    value(con.body)
  File "/home/gseastre/pyomo/src/pyomo/pyomo/core/kernel/numvalue.py", line 161, in value
    tmp = numeric(exception=exception)
  File "/home/gseastre/pyomo/src/pyomo/pyomo/core/kernel/expr_coopr3.py", line 192, in __call__
    exception=exception))
  File "/home/gseastre/pyomo/src/pyomo/pyomo/core/kernel/expr_coopr3.py", line 791, in _apply_operation
    return sum(c*next(values) for c in self._coef) + self._const
  File "/home/gseastre/pyomo/src/pyomo/pyomo/core/kernel/expr_coopr3.py", line 791, in <genexpr>
    return sum(c*next(values) for c in self._coef) + self._const
  File "/home/gseastre/pyomo/src/pyomo/pyomo/core/kernel/expr_coopr3.py", line 201, in _evaluate_arglist
    yield value(arg, exception=exception)
  File "/home/gseastre/pyomo/src/pyomo/pyomo/core/kernel/numvalue.py", line 161, in value
    tmp = numeric(exception=exception)
  File "/home/gseastre/pyomo/src/pyomo/pyomo/core/kernel/expr_coopr3.py", line 693, in __call__
    exception=exception):
  File "/home/gseastre/pyomo/src/pyomo/pyomo/core/kernel/expr_coopr3.py", line 201, in _evaluate_arglist
    yield value(arg, exception=exception)
  File "/home/gseastre/pyomo/src/pyomo/pyomo/core/kernel/numvalue.py", line 161, in value
    tmp = numeric(exception=exception)
  File "/home/gseastre/pyomo/src/pyomo/pyomo/core/kernel/expr_coopr3.py", line 192, in __call__
    exception=exception))
  File "/home/gseastre/pyomo/src/pyomo/pyomo/core/kernel/expr_coopr3.py", line 295, in _apply_operation
    return self._operator(*tuple(values))
  File "/home/gseastre/pyomo/src/pyomo/pyomo/core/kernel/expr_coopr3.py", line 201, in _evaluate_arglist
    yield value(arg, exception=exception)
  File "/home/gseastre/pyomo/src/pyomo/pyomo/core/kernel/numvalue.py", line 161, in value
    tmp = numeric(exception=exception)
  File "/home/gseastre/pyomo/src/pyomo/pyomo/core/kernel/expr_coopr3.py", line 192, in __call__
    exception=exception))
  File "/home/gseastre/pyomo/src/pyomo/pyomo/core/kernel/expr_coopr3.py", line 791, in _apply_operation
    return sum(c*next(values) for c in self._coef) + self._const
  File "/home/gseastre/pyomo/src/pyomo/pyomo/core/kernel/expr_coopr3.py", line 791, in <genexpr>
    return sum(c*next(values) for c in self._coef) + self._const
  File "/home/gseastre/pyomo/src/pyomo/pyomo/core/kernel/expr_coopr3.py", line 201, in _evaluate_arglist
    yield value(arg, exception=exception)
  File "/home/gseastre/pyomo/src/pyomo/pyomo/core/kernel/numvalue.py", line 161, in value
    tmp = numeric(exception=exception)
  File "/home/gseastre/pyomo/src/pyomo/pyomo/core/kernel/expr_coopr3.py", line 192, in __call__
    exception=exception))
  File "/home/gseastre/pyomo/src/pyomo/pyomo/core/kernel/expr_coopr3.py", line 295, in _apply_operation
    return self._operator(*tuple(values))
ValueError: math domain error`

As you can see, the logger error lines already capture the original exception message. Also, the traceback here is quite large, depending on the depth of the violating expression. That being said, I don't personally know which course here is preferred. Is it preferred to keep this long traceback and reraise the original exception (after throwing my own logger error)?

pyomo/repn/plugins/gams_writer.py
gseastream: GAMS has presolver operations, part of which evaluate the constraints at their initial values. If a variable's level has not been explicitly set before the solve command, GAMS assumes its initial value is 0 or the closest thing to 0. So, if there is a constraint written into the GAMS model file with, for example, a log function, when GAMS is told to solve it will evaluate that constraint at initial conditions, and if that log function is called on something that evaluates to 0 or less, GAMS will throw an execution error. I don't know of any way around that behavior from GAMS, so it became necessary for the user to set their variable values such that they don't violate any domains before calling the GAMS solver.

pyomo/repn/plugins/gams_writer.py
jsiirola: That (reraising the original exception) would certainly be my preference.

pyomo/repn/plugins/gams_writer.py
jsiirola: That makes sense.  However, is it the *writer's* responsibility to check this, or is it an error from the solver?  After all, you can (I assume) disable the presolve operations in GAMS.  If you really want to provide additional information to the user, I think I would prefer you to do that as a post-processing operation *if* the solver (i.e., GAMS) returns an error.

pyomo/repn/plugins/gams_writer.py
jsiirola: Agreed.  We got burned a while back when a solver "guessed" a bound when one wasn't provided.

pyomo/repn/plugins/gams_writer.py
qtothec: @gseastream Since you've written the logic, might as well add it as a potential post-processing step. =)

gseastream: I can't find a way to disable the presolver, which means I also can't verify that that would even solve the issue. But I can change it into a post-processing step and warn the user that the domain violation might be why GAMS failed.

qtothec:
@gseastream - this is the TODO list from our conversation with @jsiirola earlier. Could you help fill in the details?

1. Fix the issue with gratuitous parentheses so that regression tests pass again.
2. Clean up code: repeated block in the writer, [don't remember what else exactly]
3. Renaming of various things to follow conventions
4. fix float to int cast to either directly set value or to round first
5. align temporary file/directory behavior with other interfaces
6. Add in solver tests based on response from @ghackebeil

gseastream:
@qtothec  - yeah, that pretty much sums up the To Do list. Here is everything I noted:

1. Alias the solver_io options to "python" and "gms"
2. Call available in solve
3. Fix the filename and file object stuff
4. Fix tmp file naming and location
5. Return a results object (first I have to figure out how to do that)
6. Optionally iterate through symbolMap instead of using var_list
7. Stop supporting load_model=False
8. Make shell command a list of arguments
9. Call path returned by pyutilib.services
10. Put results.dat file in tmp directory
11. Directly set .value (or ._value)
12. Reduce code block depth where possible

pyomo/repn/plugins/gams_writer.py
```diff
@@ -216,8 +229,18 @@ def _write_model(self,
         valid_ctypes = set([
             Block, Constraint, Expression, Objective, Param, 
             Set, RangeSet, Var, Suffix ])
+        valid_kernel_ctypes = set([
```
ghackebeil: The ctypes on the kernel objects match the ctypes for the AML objects. So you can just use the valid_ctypes from above, and do something like

```python
valid_ctypes = set([Block, ...])
if isinstance(model, IBlockStorage):
    if not model.collect_ctypes(active=True, descend_into=True).issubset(valid_ctypes):
        raise RuntimeError(...)
else:
    for c in model.component_objects(active=True):
        if c.type no in valid_ctypes:
            ...
```

pyomo/repn/plugins/gams_writer.py
```diff
@@ -26,6 +26,11 @@
 from pyomo.opt.base import AbstractProblemWriter
 import pyomo.util.plugin
 
+from pyomo.core.kernel import (block, constraint, expression, objective,
```
ghackebeil: Replace all of these kernel imports with:

```python
from pyomo.core.kernel.component_block import IBlockStorage
```

pyomo/repn/plugins/gams_writer.py
```diff
@@ -171,7 +176,11 @@ def var_label(obj):
                     # Support passing of stream such as a StringIO
                     # on which to write the model file
                     output_file = output_filename
-                ComponentData.labeler.append(var_label)
+                if type(model) is block:
```
ghackebeil: use `isinstance(model, IBlockStorage)` to check for kernel models

pyomo/repn/plugins/gams_writer.py
```diff
@@ -189,7 +198,11 @@ def var_label(obj):
             finally:
                 if isinstance(output_filename, string_types):
                     output_file.close()
-                ComponentData.labeler.pop()
+                if type(model) is block:
```
ghackebeil: use isinstance(model, IBlockStorage) to check for kernel models

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -188,19 +188,16 @@ def solve(self, *args, **kwds):
...
         output_file = StringIO()
-        if type(model) is pyomo.core.kernel.component_block.block:
-            # Relevant to pyomo.solvers nosetests
-            # But actually I don't know how to get kernels to work
-            # Or if I even should
-            (_, smap_id) = model.write(filename=output_file,
-                                       format=ProblemFormat.gams,
-                                       _called_by_solver=True,
-                                       **io_options)
+        if type(model) is block:
```
ghackebeil: use `isinstance(model, IBlockStorage)` to check for kernel models

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -238,20 +235,31 @@ def solve(self, *args, **kwds):
 
         has_dual = has_rc = False
         for suf in model.component_data_objects(Suffix, active=True):
-            if (suf.name == 'dual' and suf.import_enabled()):
-                has_dual = True
-            elif (suf.name == 'rc' and suf.import_enabled()):
-                has_rc = True
+            if type(model) is block:
```
ghackebeil: use `isinstance(model, IBlockStorage)` to check for kernel models

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -439,17 +447,16 @@ def solve(self, *args, **kwds):
 
         io_options['put_results'] = results_filename
 
-        if type(model) is pyomo.core.kernel.component_block.block:
-            # Relevant to pyomo.solvers nosetests
-            (_, smap_id) = model.write(filename=output_filename,
-                                       format=ProblemFormat.gams,
-                                       _called_by_solver=True,
-                                       **io_options)
+        if type(model) is block:
```
ghackebeil: use `isinstance(model, IBlockStorage)` to check for kernel models

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -492,20 +499,31 @@ def solve(self, *args, **kwds):
...
+            else:
+                if suf.name == 'dual' and suf.import_enabled():
+                    has_dual = True
+                elif suf.name == 'rc' and suf.import_enabled():
+                    has_rc = True
 
         for sym, ref in symbolMap.bySymbol.iteritems():
             obj = ref()
-            if not obj.parent_component().type() is Var:
+            if type(model) is block:
```
ghackebeil: use `isinstance(model, IBlockStorage)` to check for kernel models

pyomo/core/kernel/component_interface.py
```diff
@@ -87,6 +87,8 @@ class ICategorizedObject(object):
     _is_component = False
     _is_container = False
 
+    labeler = []
```
ghackebeil: Maybe a comment here about what this is used for.

pyomo/core/kernel/component_interface.py
```diff
@@ -210,6 +212,8 @@ def __str__(self):
             for a way to generate a static set of component
             names.
         """
+        if ICategorizedObject.labeler:
```
ghackebeil: So far, I've managed to keep these newer files in Kernel 100% covered with smoke unit tests (no big external dependencies). This change also makes the doc string incorrect. I'm being pedantic here, but I don't want these kind of low-level hacks to start creeping into Kernel without testing and documentation.

An update to the doc string and a unit test for the variable would probably do the trick (pyomo/core/tests/unit/test_component_variable.py). Perhaps @jsirrola can chime in here.

gseastream: That makes total sense, and as a relative newbie to pyomo I'm all for more documentation.

pyomo/repn/plugins/gams_writer.py
```diff
@@ -0,0 +1,512 @@
...
+        # immediately anyway.
+        with PauseGC() as pgc:
+            try:
+                if isinstance(output_filename, string_types):
+                    output_file = open(output_filename, "w")
+                else:
+                    # Support passing of stream such as a StringIO
+                    # on which to write the model file
+                    output_file = output_filename
+                if isinstance(model, IBlockStorage):
```
ghackebeil: @jsiirola This is not thread safe. I think you guys should find a different way to inject a labeler into the expression output (e.g., with a keyword to expr.to_string()).

Attaching a script that exposes the problem. You could add a thread lock to the GAMS writer (and wherever else you are doing this kind of thing), but then you eliminate the potential to use threads for something that we could probably easily work around.

Yes, I know the GIL prevents the thread-based block in this example (multiprocessing.dummy) from showing much speedup, but that is just a characteristic of this example. It would be very easy to change the example so that the thread tasks involved a lot of blocking I/O in addition to building and writing a Pyomo model. Note that I ran this example with the LP writer and just appended to a class level attribute inside the writer (removing it after the write completed), and saw that it grew above size 1 when the thread-based pool was running. I'm guessing you would see the same with the GAMS writer and run into some kind of error from the labelers getting mixed up.

[thread_test.py.txt](https://github.com/Pyomo/pyomo/files/1178565/thread_test.py.txt)

gseastream:
@ghackebeil I got the syntax right and got travis to properly install GAMS, and on python 2 the tests all run properly. A few things:

1. It still errors on the pyomo.solvers.tests.checks.test_writers tests because my solver throws an error when load_solutions=False, both because I still haven't gotten around to creating a results object and because @jsiirola recommended I not support load_solutions=False. What should be done about this? Should these tests be skipped?
2. The GAMS installation log seems to have a massive amount of text in the travis log. Should this text be piped somewhere, and if so, where and how?
3. My solver errors on all of the python3 tests, in different ways. In python 3.4 and 3.5 it gives:
```
test_gams_gms_nonsymbolic_labels (pyomo.solvers.tests.checks.test_pickle.Test_LP_block) ... ERROR
test_gams_gms_symbolic_labels (pyomo.solvers.tests.checks.test_pickle.Test_LP_block) ... ERROR
test_gams_python_nonsymbolic_labels (pyomo.solvers.tests.checks.test_pickle.Test_LP_block) ... /home/travis/.travis/job_stages: line 57:  3800 Segmentation fault      (core dumped) test.pyomo -v --cat=$CATEGORY --cov

travis_time:end:2ef90d70:start=1501124201214749289,finish=1501124383940572302,duration=182725823013
 [0K
 [31;1mThe command "test.pyomo -v --cat=$CATEGORY --cov" exited with 139. [0m

Done. Your build exited with 1.
```
On the first invocation of the GAMS solver, and then the build stops prematurely.
For python 3.6, it finishes the build, but all the tests give the same error (a result of trying to run a python2 script in python3):
```
  File "/home/travis/build/Pyomo/pyomo/pyomo/solvers/tests/checks/test_pickle.py", line 54, in pickle_test
    load_solutions)
  File "/home/travis/build/Pyomo/pyomo/pyomo/solvers/tests/models/base.py", line 107, in solve
    **io_options)
  File "/home/travis/build/Pyomo/pyomo/pyomo/solvers/plugins/solvers/GAMS.py", line 514, in solve
    for sym, ref in symbolMap.bySymbol.iteritems():
AttributeError: 'dict' object has no attribute 'iteritems'
```
Is my solver supposed to work in python3? And if so, how do I make it so that it does?

ghackebeil:
Yeah, why don't you pipe that one command's output to /dev/null.

As for not supporting load_solutions=False, I'm fine with this, but it will require some rewrites to test_writers.py and test_pickle.py, and I don't have to time to look at that right now (and I don't know if it will be easy or hard). It might be quicker to just have this solver go through the extra indirection of populating a results object like all of the other solvers do. You need to do this anyway to provide status information. I doubt you would see a big change in performance and right now it's the fastest way to get this solver plugin to pass tests.

pyomo/repn/plugins/gams_writer.py
gseastream: I'm wondering if I'm understanding this problem correctly: is the issue that if there are multiple processes running at the same time, if they both use this ComponentData.labeler list, then they will confuse each other by trampling on the other's storage location for their labeler? And thus there needs to be a way to pass a labeler that is unique to each process?

If so, I see two options. One is another sort of workaround, where ComponentData.labeler is a dictionary instead of a list, and the labeler is added to this dictionary with a key that is unique to each process. Of course then you have to decide how to generate keys, and me not knowing many options here I can think of making a big random number.

The other option I see, as you said, is to pass the labeler as a keyword to to_string. This would require updating all the different definitions of to_string, for which there are many when it comes to all the different types of expressions. They all need to be updated so that the labeler keyword can be passed along until it reaches a Var, who will act accordingly if it is passed a labeler. Is this the option you prefer?

gseastream: I've started implementing the to_string() keyword choice, but I am a little concerned. I started by updating all the to_string definitions that I knew of, both adding a labeler keyword to their definition and passing this labeler as a keyword to all calls it makes, but of course that was not all of the to_string methods spread around pyomo. Now I'm relying on the regression tests to tell me when I missed one. For example, I started in expr_coopr3 and did all of those, but then a baron test errored because it tried passing labeler to a SimpleExpression's to_string, so then I updated that one. On the other hand, I got an error because I was trying to pass labeler to a NumericConstant's to_string (part of the constant objective tests), but I decided that should just return a string of the value, not a label, so I changed my code to not pass a labeler if objective.is_constant(). With all of this, I worry that either I or the regression tests will miss a to_string() to change. Thoughts?

The nice thing about the ComponentData.labeler strategy was that nothing had to be passed, and the appropriate __str__ or to_string functions knew whether or not to call this labeler. But now it needs to be passed quite a bit.

As a second-hand thought, should every single to_string method definition in pyomo be updated to include the labeler keyword, and only the appropriate (ComponentData) objects be told to use it?

jsiirola: @ghackebeil's point is that any use of globals to change behavior is inherently not thread safe.  We are "saved" because CPython isn't really multithreaded -- but other (GIL-less) interpreters are [Jython, IronPython].  Now, Pyomo doesn't support GIL-less interpreters at the moment; but, @ghackebeil's point is that we shouldn't exacerbate the problem.

Now, what is the correct solution?  I am thinking about some form of modified visitor pattern, where the walker (in this case, `to_string`) would take a "visitor dictionary".  If the type appears in the visitor dictionary, then the walker would call the dict's handler instead of the built-in type's method.  There are additional questions, though... we probably don't want a "simple" dict, but rather one that looks up with something like `isinstance`.  That way the caller doesn't have to divine the number of derived classes that may appear in the expression.  Depending on how this is implemented, we can even use it to (temporarily) descend past the `Expression` objects, so that we get the "real" full expression (and don't have to add the fake variables for the Expressions).  The only problem is that `to_string` is a complicated walker, due to prefix / postfix/ and infix calls that are required.

BTW: implementing this will also help with the BARON writer.

ghackebeil: > We are "saved" because CPython isn't really multithreaded 

I don't think this is correct, John. In CPython, the GIL just prevents most opportunities for CPU-bound threads to run at the same time. However, even in CPython, the GIL is released during blocking I/O (e.g., file.write), which is exactly what is happening in a problem writer. So this is not even safe in CPython.

gseastream: I finished implementing the more-or-less simple strategy of just passing labeler to to_string. I used grep to find all "def to_string" definitions in pyomo, added a labeler keyword to them, and passed labeler to all further calls of to_string within those definitions. ComponentData.to_string() and a new ICategorizedObject.to_string() are the only ones that apply the labeler if provided one. The GAMS solver now passes everything, but speaking of Baron, the Baron solver now errors on the following 4 tests:
```
ERROR: test_baron_bar_nonsymbolic_labels (pyomo.solvers.tests.checks.test_pickle.Test_LP_simple_kernel)
ERROR: test_baron_bar_symbolic_labels (pyomo.solvers.tests.checks.test_pickle.Test_LP_simple_kernel)
ERROR: test_baron_bar_nonsymbolic_labels (pyomo.solvers.tests.checks.test_pickle.Test_QP_simple_kernel)
ERROR: test_baron_bar_symbolic_labels (pyomo.solvers.tests.checks.test_pickle.Test_QP_simple_kernel)
```
I know very little about how Baron works, so I have no idea how to address these errors. The full error messages all look similar to this:
```
ERROR: test_baron_bar_nonsymbolic_labels (pyomo.solvers.tests.checks.test_pickle.Test_LP_simple_kernel)
  File "/home/gseastre/pyomo/src/pyomo/pyomo/solvers/tests/checks/test_pickle.py", line 54, in pickle_test
    load_solutions)
  File "/home/gseastre/pyomo/src/pyomo/pyomo/solvers/tests/models/base.py", line 113, in solve
    **io_options)
  File "/home/gseastre/pyomo/src/pyomo/pyomo/opt/base/solvers.py", line 623, in solve
    "Solver (%s) did not exit normally" % self.name)
ApplicationError: Solver (baron) did not exit normally
-------------------- >> begin captured stdout << ---------------------
ERROR: "[base]/pyomo/pyomo/opt/base/solvers.py", 616, solve
        Solver (baron) returned non-zero return code (3)
ERROR: "[base]/pyomo/pyomo/opt/base/solvers.py", 621, solve
        Solver log:
        ===========================================================================
         BARON version 17.4.1. Built: LNX-64 Sat Apr 1 11:24:13 EDT 2017 

         If you use this software, please cite:
         Tawarmalani, M. and N. V. Sahinidis, A polyhedral
         branch-and-cut approach to global optimization,
         Mathematical Programming, 103(2), 225-249, 2005.

         BARON is a product of The Optimization Firm, LLC. http://www.minlp.com/
         Parts of the BARON software were created at the
         University of Illinois at Urbana-Champaign.
        ===========================================================================
        ERROR: Illegal variable in line 19
        BARON: Syntax error.  Execution will stop

--------------------- >> end captured stdout << ----------------------
-------------------- >> begin captured logging << --------------------
pyomo.opt: ERROR: Solver (baron) returned non-zero return code (3)
pyomo.opt: ERROR: Solver log:
===========================================================================
 BARON version 17.4.1. Built: LNX-64 Sat Apr 1 11:24:13 EDT 2017 

 If you use this software, please cite:
 Tawarmalani, M. and N. V. Sahinidis, A polyhedral
 branch-and-cut approach to global optimization,
 Mathematical Programming, 103(2), 225-249, 2005.

 BARON is a product of The Optimization Firm, LLC. http://www.minlp.com/
 Parts of the BARON software were created at the
 University of Illinois at Urbana-Champaign.
===========================================================================
ERROR: Illegal variable in line 19
BARON: Syntax error.  Execution will stop 

--------------------- >> end captured logging << ---------------------
```

ghackebeil: Try injecting a `model.write('junk.bar')` call somewhere at the beginning of the main test in test_pickle.py, then execute one of those tests before and after your change. The diff of the output .bar files will probably show something obvious that needs to be corrected.

gseastream:
@qtothec @jsiirola @ghackebeil This branch now passes all of the main travis tests for python 2.6, 2.7, and 3.6. The API solver is seg-faulting on python 3.4 and 3.5, as well as the "KEY_JOB" set of python 2.7 tests. I'm not really sure what is causing that, and while I haven't looked into it too hard yet, I was able to run and pass the tests on a python 3.5 installation of pyomo, so if anyone has any insight on that I would much appreciate.

You'll also note in my latest commit the change I made to to_string() so that it passes the unit tests. The original problem with string generation was that a _SumExpression containing a _ProductExpression could sometimes print a double negative, since there are 2 coefficients for a _ProductExpression, one being part of the _ProductExpression and the other stored in the _SumExpression's list of constituent of coefficients (`self._coef`). Thus, an expression could be printed as `x - -3 * y`. I only ever saw this problem arise when a _SumExpression was defined with a _ProductExpression that contains a variable and a negative coefficient on the right side of an equals or inequality sign, so that when pyomo generated the _SumExpression and it moved all the variable expressions to the left-hand side, it negated the _ProductExpression by giving it a -1 coefficient in the _SumExpression's _coef list, but leaving it with it's own negative coefficient. Example:
```python
from pyomo.environ import *
m = ConcreteModel()
m.x = Var()
m.y = Var()
m.z = Var()
m.c = Constraint(expr=m.x == -5 * m.y * m.z)
m.c.body.to_string()
```

I don't know if this is the only case where this arises, however, although I can't think of any more myself.

Anyway, to solve this problem with double negative string generation, I initially bounded all negative _ProductExpression coefficients in parentheses, but that caused it to fail unit tests that assumed there would be no parentheses, because no unit tests checked cases with a double negative. So, I decided to change my fix so that it only applies when needed: a _ProductExpression.to_string with a negative coefficient being called by a _SumExpression.to_string where it is not the first term in the _SumExpression. To achieve this, there weren't many options I could think of in terms of communicating whether something is the first term between the _SumExpression's to_string and the _ProductExpression's to_string that it calls. The choice I went with was a bit of hack: I made a new attribute of _ProductExpression's that indicates if it is the first term in a _SumExpression, called _is_first_of_sum. I only initialize it when to_string() is called on the parent _SumExpression, since I didn't really know where to put it in terms of initialization/generation functions.

So, I'm sure you all might have opinions on this. If this is totally a bad choice, let me know, and of course any potential alternatives would be helpful. Or I guess another option is change the expression generation so that there won't be 2 negative coefficients stored for a _ProductExpression, if this is a bug of sorts (in which case this would be me reporting it now). Also, I realized a _SumExpression passes _ProductExpression.PRECEDENCE in these cases, so I changed it accordingly; would this cause any issues? For example does a _ProductExpression ever call to_string on an inner _ProductExpression?

gseastream:
@jsiirola I am unsure what to do about the segfaulting issue I described above that is affecting python 3.4 and 3.5 and part of 2.7. As I said above, I am able to run the tests fine on a python 3.4 and 3.5 installation of pyomo locally, so it must be an issue with travis. But since this is only an issue on travis, is it an issue I should worry about/fix? And if so, would you know how to approach it?

gseastream:
So I changed my mind on how to solve the parentheses with product expressions problem. I got rid of the whole adding an attribute/slot thing. Now I added a keyword arg to the product expression's to_string() definition called bind_neg_coef. This is defaulted to False, and is only set True when a sum expression is calling to_string() on a product expression when first=False. So, it's the same effect but less disruptive to the structure of a product expression, while changing the definition of to_string() a little.

ghackebeil:
Can you try temporarily adding a bunch of print statements to the GAMS writer / solver plugin to pinpoint exactly where the segault happens. I know it's sloppy, but I can't think of another way to debug it right now.

In the meantime, I will follow the Travis setup and try to run the GAMS tests on my laptop.

gseastream:
Ok. After adding a whole lot of print statements and switching to the nosetests command in .travis.yml so I could see the stdout, I found that the following line is where the segfault occurs:
```python
t1.run(output=sys.stdout if tee else None)
```
This is the command that invokes the GAMS Python API to essentially run GAMS on the model file previously provided in `t1 = ws.add_job_from_string(output_file.getvalue())`. My only idea was to try writing the model onto a file instead of a StringIO (which is what the shell solver does) and do add_job_from_file, but this did not change anything and the segfault still occurs in the same place.

I do not have access to the api code, so I don't think I have any way to figure out what part of `t1.run()` is causing a segfault, just that this command is where it is happening. Again, this is only happening on travis, not locally (for me), and only on python 3.4, 3.5, and the 2.7 test with KEY_JOB="yes", and it is only a problem with the API solver; the shell solver (as seen by the test that is run right before the segfaulting test) runs fine.

So what do we do about this?

qtothec:
It seems to me that the problem is that GAMS or the solver is doing something to segfault. Now that you mention it, I do remember getting random segmentation failures during my master's work when I was running things in GAMS.

1. Is there something that can be done to insulate testing of this interface from that possibility?
2. Which solver(s) are you using for the tests?

gseastream:
I don't know how to insulate this possibility aside from just skipping these tests, but here is the list of default solvers that GAMS uses, for both the models it gets passed and any subproblems it creates and solves. Note that I changed the default for NLP from CONOPT to BARON because CONOPT was getting an answer that was not quite the correct answer while BARON did it correctly (and included marginals), however I think this problem no longer exists for CONOPT after I added the power() function replacement (this was on the QCP problems that CONOPT wasn't getting it right, and I think it had to do with using ** instead of power). Either way I just left NLP using BARON since it succeeded where CONOPT failed at least in one place. Anyway here's the list:
```
DEFAULTS
LP CPLEX
MIP CPLEX
RMIP CPLEX
NLP BARON
MCP PATH
MPEC NLPEC
RMPEC NLPEC
CNS CONOPT
DNLP CONOPT
RMINLP CONOPT
MINLP DICOPT
QCP CONOPT
MIQCP SBB
RMIQCP CONOPT
EMP JAMS
```

jsiirola:
Is that model more than a handful of variables?  I wonder if this is because the GAMS instance you use locally is a fully-licensed version and the GAMS instance on Travis is in demo mode (and subject to the demo size restrictions)?

gseastream:
@jsiirola The issue doesn't seem to be with the licensing or lack thereof. I am able to run all the tests fine locally with a fresh install of GAMS with no license. Also, all the rest of the travis tests that don't segfault are able to run the same models just fine. And the travis tests that do segfault, are able to run the same tests on the shell solver method first before trying the API solver. So model size and licensing do not appear to be the issue.

It is weird though that this segfaulting issue is only occurring on some versions of python. In trying to google for an answer, other people in the past have encountered segfaulting issues on travis where it runs fine locally, but the issues I saw and the solutions given did not seem relevant to this problem. Should I perhaps open an issue on the travis repo? Otherwise I still don't know how to solve this problem.

ghackebeil:
I have two suggestions about the segfault. In the travis file, first try moving all of the gams installation lines before the line where pyomo.extras is installed. If that doesn't work, then just try only installing gams when EXTRAS is _not_ set (e.g., inside of an `if [ -z "${EXTRAS}" ]` block), and maybe add a comment about the segfault.

It looks like all tests pass on Python2.7 and Python3.6 (when EXTRAS is not set), and that should be sufficient to merge this branch.

It might also be worth moving all of the gams installation lines inside of their own bash file and just calling that from the travis file. That is, make a file called `.travis_install_gams.sh` and then call `bash ./.travis_install_gams.sh` from the main travis file to install and setup gams.

gseastream:
I tried moving the installation to before pyomo.extras is installed, to no effect.

I just noticed something very strange, however. After my commit "Fixed elif's and objective value setting", the python 3.4 tests started passing. Previously those tests (which do have EXTRAS installed), segfaulted just the same, but after that commit they all of sudden stopped having that problem. All that commit did was change some if's to elif's and set the objective value in the solutions object for non-kernel models. This really should not have changed anything since the segfault occurs when the GamsJob is run, before any of these changes...

ghackebeil:
Maybe the separate bash file was a bad idea. Not sure why things are going wrong.

gseastream:
I think it has to do with Windows carriage returns, because I'm on a Windows computer. I'm seeing now if there's a quick way to solve that.

.travis.yml
```diff
@@ -70,25 +70,14 @@ install:
...
-#
 # Install free version of GAMS: executable and API bindings
 # No explicit API version for python 3.5, use 3.4
 #
 - wget https://d37drm4t2jghv5.cloudfront.net/distributions/24.8.5/linux/linux_x64_64_sfx.exe
 - chmod u+x linux_x64_64_sfx.exe
 - ./linux_x64_64_sfx.exe > /dev/null
 - PATH=${PATH}:$PWD/gams24.8_linux_x64_64_sfx
-# Change default NLP solver to baron for testing
+# Change default NLP solver to baron for testing, performs better than CONOPT
```
qtothec: Nick will be happy, but just so that you know, BARON uses either CONOPT, IPOPT, FilterSQP, or some other NLP solver for NLP problems. Maybe it's the pre-solve that they have.

gseastream:
Travis is doing a weird thing. I opened an issue: #199

PR #166, opened 2017-06-14
Preliminary implementation of outer approximation and partial surrogate cuts in Pyomo

qtothec:
I accidentally merged in db7b142 earlier on; otherwise, this shouldn't have any conflicts with master.

Provides a beta implementation of outer approximation in Pyomo.


qtothec:
Replaced by #255

PR #168, opened 2017-06-15
Fix global name not defined in _get_linear_collector

qtothec:
Exactly what the title suggests, plus my text editor removed some whitespace at the end of lines.


jsiirola:
@qtothec: the fix is of course fine, but as a matter of principle, can you add a test that exercises this?  I am a bit bothered by the fact that this slipped through all unit tests...

jsiirola:
#169 was made from the same branch, so it supersedes this PR (an in particular, I don't think you can add tests to this PR without affecting #169.  I am going to close this PR and move discussion over to #169.

PR #170, opened 2017-06-16
Fix miscategorization of infeasible problems as unbounded in Gurobi

qtothec:
Gurobi will sometimes report a termination condition of unbounded for problems that are actually infeasible. I had trouble getting a small test case to exhibit this behavior, but below is example output from a slightly larger problem. As you can see, `termination_condition` is `unbounded`, but `lower_bound` is `1e+100`. For comparison, output for an actual unbounded case is also included below. The `lower_bound` in that case is `-inf`. I use this distinction to write a small correction.

CPLEX also exhibits the same behavior, but I couldn't find an easy way to distinguish the infeasible case based on the solver output.

**Infeasible miscategorized as unbounded**
```
section:problem
name: x34
sense:minimize
lower_bound: 1e+100
number_of_objectives: 1
number_of_constraints: 89
number_of_variables: 34
number_of_binary_variables: 8
number_of_integer_variables: 8
number_of_continuous_variables: 26
number_of_nonzeros: 661
section:solver
status: warning
return_code: 0
message: Problem proven to be infeasible or unbounded.
wall_time: 0.00188493728638
termination_condition: unbounded
termination_message: Problem proven to be infeasible or unbounded.
```

**Actually unbounded**
```
section:problem
name: x2
sense:minimize
lower_bound: -inf
number_of_objectives: 1
number_of_constraints: 1
number_of_variables: 2
number_of_binary_variables: 0
number_of_integer_variables: 0
number_of_continuous_variables: 2
number_of_nonzeros: 1
section:solver
status: warning
return_code: 0
message: Problem proven to be infeasible or unbounded.
wall_time: 0.000164985656738
termination_condition: unbounded
termination_message: Problem proven to be infeasible or unbounded.
```

### DIFF
--- a/pyomo/solvers/plugins/solvers/GUROBI.py
+++ b/pyomo/solvers/plugins/solvers/GUROBI.py
@@ -2,8 +2,8 @@
 #
 #  Pyomo: Python Optimization Modeling Objects
 #  Copyright 2017 National Technology and Engineering Solutions of Sandia, LLC
-#  Under the terms of Contract DE-NA0003525 with National Technology and 
-#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain 
+#  Under the terms of Contract DE-NA0003525 with National Technology and
+#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain
 #  rights in this software.
 #  This software is distributed under the 3-clause BSD License.
 #  ___________________________________________________________________________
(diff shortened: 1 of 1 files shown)


ghackebeil:
I don't think this fix is valid. The way the LP-file interface to Gurobi works is that we write an LP file, and then call commands within the gurobi.sh shell to solve the problem and write a solution. This all takes place in the GUROBI_RUN.py file, so if anything, the fix should be to output something like None for the objective from that file. The `termination_message` is coming straight from Gurobi. The fact that the `termination_condition` is set to `Unbounded` is more due to a deficiency in the `TerminationCondition` Enum. It does not have anything like `InfeasibleOrUnbounded`, so someone (possibly me) just arbitrarily picked `Unbounded` for this case. That should probably be fixed as well, by introducing a new member to the `TerminationCondition` Enum.

qtothec:
Oh, I looked in the wrong place to apply the fix, then. I thought that the *.gurobi.txt output files were actually coming from Gurobi. If you think that the proper response is to add a member to the TerminationCondition enum for InfeasibleOrUnbounded, I can do that and put the logic for discriminating between the two in my own code.

ghackebeil:
I think the addition to that Enum sounds like a good idea. Any of the other developers want to chime in?

Note that one of the points I was making is that you can't rely on the objective or bound that is being output to distinguish between the two cases (it's something that is hard coded by us in GUROBI_RUN.py), so I don't think the above code is safe to use even in your own code. Gurobi is not even distinguishing between the two cases, it is just reporting that one of theses cases is true. There are likely some Gurobi options you can use to tell it to try and distinguish between the two cases. Just something to keep in mind.

qtothec:
Yeah, not too long after your post, I stumbled into a case where that fragility was exposed. I'm going to close this and open a separate PR with the Enum addition later.

PR #174, opened 2017-06-23
Add docstring to apply function in Transformation

qtothec:
...so that users with IDEs can see the deprecation warning without trying to use the function or digging through the Pyomo codebase.


whart222:
The test failed due to an intermittent PySP test failure, which I'm ignoring.

PR #176, opened 2017-06-27
Fix so that generate_cuid_names descends into Disjunct objects as well

qtothec:
This is a quick fix to allow `generate_cuid_names` to descend into "Block-like" components. I am not sure why `generate_cuid_names` doesn't use the `_tree_iterator` approach that things like `block_data_objects` uses, but I wasn't comfortable enough with the iteration methods to go ahead and refactor the function with that approach.


jsiirola:
@qtothec: can you take a look at the changes and see if they work for you?

qtothec:
@jsiirola Short answer: no. Trying to figure out why right now. Relevant trace:
```
  File "/home/qichen/.solvers/pyomo/src/pyomo/pyomo/solvers/plugins/solvers/GDPDT.py", line 155, in solve
    model, ctype=(Var, Disjunct), descend_into=(Block, Disjunct))
  File "/home/qichen/.solvers/pyomo/src/pyomo/pyomo/core/base/block.py", line 1930, in generate_cuid_names
    cuid_names_=cuid_names_)
  File "/home/qichen/.solvers/pyomo/src/pyomo/pyomo/core/base/block.py", line 1907, in generate_cuid_names
    for key, obj in block.component_map(ctype=ctypes).items():
AttributeError: '_GeneralVarData' object has no attribute 'component_map'
```

qtothec:
Closing and re-opening to trigger Travis rebuild.

qtothec:
My two cents:
1) My understanding is that this was done for performance reasons.
2) It checks for things like `descend_into=Block`, which is neither True nor a tuple/list/set
3) I'll let @jsiirola answer this one.
4) Ok.
5) I named the PR poorly. It should say "so that `generate_cuid_names` is able to descend into Disjuncts". Default behavior is not to. Passing `descend_into=(Block, Disjunct)` is what I have in my code.

michaelbynum:
2.) I still think descend_ctype is not defined if descend_into is False.
6.) My mistake. I did not realize that component_data_objects worked with descend_into=(Block, Disjunct)

qtothec:
2.) You are correct. `descend_ctype` is never defined if `descend_into` is `False`. This turns out not to be a problem because `descend_ctype` is never used if `descend_into` is `False`. If you'd like, I could make it more explicit?

michaelbynum:
Ah, good point. Would you mind making it more explicit? I worry that someone else will try to update this function in the future and mistakenly try to use descend_ctype when it is not defined.

jsiirola:
To respond to @michaelbynum's initial question: "Should we make iterators of pyomo components (primarily _BlockData.PseudoMap._ctypewalker) use isinstance to check for subclasses?"

No.  One of the design intents of ctype is that it is a classification system that is orthogonal to the class hierarchy.  There are two key benefits of this:
  1) This allows for both classes (like `Disjunct`) that want all the functionality of a base component (something like `Block`), but do not make sense to blindly treat them as if they were the base class.  For example, the LP/NL/direct writers should complain loudly if a model has any active Disjuncts.  If the iterator used isinstance, then the writers would need special code to check for Disjuncts.  By having the iterators group by ctype and not isinstance, the writer should only check for any ctypes that are not in the list of ctypes that it knows how to deal with.
  2) This also allows for "duck-typing".  Someone can implement their own class that is independent of the "normal" class hierarchy, and Pyomo will treat it like it is the ctype it declares.

michaelbynum:
Those are good reasons. Thanks John.

michaelbynum:
Okay, my only remaining concerns are items (3) and (5) from my original review. Then, I will vote that this should be merged.

qtothec:
My concern with implementing the suggestions in (3) and (5) is that we would see the wrong behavior for `descend_into=Block` or `ctype=Block` (and other similar components). `Block` is iterable and you'll end up with a tuple `(None,)`.

michaelbynum:
I suggest something like the following:

if descend_into is True:
    descend_ctype = (Block,)
elif descend_into is False:
    descend_ctype = False
elif isinstance(descend_into, collections.Iterable):
    for i in descend_into:
        assert isinstance(i, Component)
    descend_ctype = tuple(descend_into)
elif isinstance(descend_into, Component):
    descend_ctype = (descend_into,)
else:
    raise ValueError('useful error message')

Then on line 1933, use

if descend_ctype is not False:

rather than

if descend_into:

Do you think that will work Qi? My main concern is that we don't allow just any object that can return True from bool(object) to be passed in for descend_into.

Thanks,
Michael

qtothec:
@michaelbynum I see your concern about `bool(object)`. Is the idiom that we want to use for detection of "is a component" `isinstance`, given the kind of discussion that I see in #276?

jsiirola:
There are 2 separate issues related to @michaelbynum's (3):
  1. Should we accept general iterables for `descend_into=`?  I don't see a problem with it.  @qtothec: FYI: `descend_into=Block` isn't a problem: while a Block instance is iterable, `Block` is a type and is not iterable.
  2. Should we verify the type of the things provided to `descend_into=`?  Maybe?  If we check it, should we enforce `isinstance(Component)`, or just confirm that it is a `type`?  Thinking about how things are written currently in the AML classes, the _object_ that is referenced in the Block (PseudoMap) data structures must be a derivative of Component, but we never require that the `ctype` by a Component type...

michaelbynum:
@qtothec @jsiirola I clearly need more coffee today. John is right. The isinstance function would not work at all. We are working with class types here, not instances of classes. How about this:

if descend_into is True:
    descend_ctype = (Block,)
elif descend_into is False:
    descend_ctype = False
elif type(descend_into) == type:
    descend_ctype = (descend_into,)
elif isinstance(descend_into, collections.Iterable):
    for i in descend_into:
        assert type(i) == type
    descend_ctype = tuple(descend_into)
else:
    raise ValueError('useful error message')

qtothec:
Made that change. Going to close and reopen because Travis got caught up on a setup-related thing in 2.6.

michaelbynum:
Thanks for making that update Qi. 

@jsiirola I am happy with this pull request.

qtothec:
@jsiirola Yeah, I was a bit worried about that. I'll try to keep the major PEP8 changes separate from now on.

PR #180, opened 2017-06-30
DAE simulator

blnicho:
This branch includes the new ODE/DAE simulator interface for pyomo.DAE. It currently interfaces with the integrators/simulators in both scipy (for ODEs) and CasADi (for ODEs and DAEs). It relies heavily on TemplatedExpressions and expression substitution to convert our expression trees into something compatible with scipy or CasADi. The Simulator also includes a function for initializing a discretized Pyomo model using the simulated profile.


blnicho:
This is a different test failure specific to the simulator but I'm having a similar issue as John in that I can't replicate it locally. I thought it was being caused by using a different version of Scipy than Travis but I'm still unable to replicate it. The next most likely theory is problems comparing floating point numbers, either floats not being identified as such or the comparison tolerance not working exactly like I would expect.

blnicho:
The issue with the simulator test was caused by numpy being too clever and pyutilib not ignoring whitespace like I expected. My tests originally used `numpy.set_printoptions(precision=4)` to print numpy arrays to 4 decimal places. However, when this truncation results in floats with extra zeros on the right hand side, they are dropped and not printed. Numpy is smart enough to still print these values in nicely formatted columns:

    [ 4.5252    3.9042]
    [-3.283     4.59  ]

The problem arises when different operating systems end up with slightly different floats (which is expected and handled correctly by the tolerance value set in pyutilib comparisons) that lead to a different number of digits when formatted by numpy. In the second row, the white-space after `4.59` causes that line to be broken into a different number of chunks as the previous line which causes comparison errors in pyutilib. The solution is to change how numpy prints arrays in order to guarantee that 4 decimal places are printed even if there are trailing zeros

    numpy.set_printoptions(formatter={'float':'{: 0.4f}'.format})

PR #190, opened 2017-07-21
Transformation of explicit constraints to variable bounds

qtothec:
Adding a new transformation.


pyomo/core/plugins/transform/bounds_to_vars.py
```diff
@@ -0,0 +1,77 @@
...
+        """Apply the transformation to the given model."""
+        m = model
+
+        for constr in m.component_data_objects(ctype=Constraint,
+                                               active=True,
+                                               descend_into=True):
+            # Check if the constraint is k * x + c1 <= c2 or c2 <= k * x + c1
+            if constr.body.polynomial_degree() == 1:
+                repn = generate_canonical_repn(constr.body)
+                if repn.variables is not None and len(repn.variables) == 1:
```
jsiirola: Again, you should filter out the variables with a 0 coefficient first before you decide if you can update the bounds.

pyomo/core/plugins/transform/bounds_to_vars.py
```diff
@@ -0,0 +1,77 @@
...
+                    var = repn.variables[0]
+                    const = repn.constant if repn.constant is not None else 0
+                    coef = repn.linear[0]
```
jsiirola: For safety (and Python2 compatibility), you should cast the coef to a float to avoid integer division: `coef = float(repn.linear[0])`

pyomo/core/plugins/transform/bounds_to_vars.py
qtothec: I assume that using the `>` and `<` operators implicitly do this filtering. Is that not sufficient?

qtothec:
Close and reopen to trigger rebuild due to a network-connectivity-related issue.

qtothec:
Close and reopen to work around `test_farmer_quadratic_ipopt_with_phpyro` intermittency

pyomo/core/plugins/transform/bounds_to_vars.py
jsiirola: DOH.  I missed that.  Yes - the `__future__` will cover it.  [So you are seeing a style difference ... instead of remembering to start every script with the division import, I just trained myself to explicitly convert something to float]

pyomo/core/plugins/transform/bounds_to_vars.py
jsiirola: No: edge cases like in your other transformation (e.g., `m.x + m.y - m.x <= 0` will end up with a zero coefficient in the canonical repn - and two variables in the `repn.variables`.  That said, the correct answer may be to have generate_canonical_repn filter those out as part of it's final postprocessing step...

qtothec: I think it might be compelling to do that post-processing, since I tend to assume at least for linear systems that the variable terms are aggregated. I don't think that the edge case is relevant to this transformation, though, because of the `len(repn.variables) == 1` requirement. It will just fail to transform that constraint.

qtothec: Actually, it appears that for your simple example, this kind of processing already takes place, both for coopr3 and pyomo4 expression systems.

pyomo/core/plugins/transform/bounds_to_vars.py
```diff
@@ -0,0 +1,84 @@
...
+                            elif coef < 0:
+                                var.setub(min(var.ub, newbound)
+                                          if var.ub is not None
+                                          else newbound)
+                    constr.deactivate()
+                    # Sometimes deactivating the constraint will remove the
+                    # variable from all active constraints, so that it won't be
+                    # updated during the optimization. Therefore, we need to
+                    # shift the value of var as necessary in order to keep it
+                    # within the constraints.
```
carldlaird: Several comments:
1) Unclear exactly what this comment is saying. Are you indicating that it is possible that the variable is only in constr, therefore, when we deactivate it, the variable does not appear in any constraints at all?

2) Also, why is it necessary to shift the value? It is possible to have variable values that are outside the bounds. Is this just "cleanup"?

3) If I understand correctly, then "keep it within the constraints" should read "keep it within the bounds".

qtothec: 1. Yep.
2. Yes, this is a form of cleanup.
3. Yeah, I can change that.

qtothec: On further thought, with 2: It makes sense to shift the value because we are removing a constraint from the system and shifting its "effect" directly onto the variable. In order for post-solve variable values to remain as they would be had we not performed the transformation, we need to move the variable value if it falls outside of the new bound.

PR #192, opened 2017-07-24
Adding transformation to propagate equality-linked fixed variables.

qtothec:
Identifies constraints of form `x - y = 0`. Checks to see if either `x` or `y` are `fixed`. If so, fix the other variable.


pyomo/core/plugins/transform/equality_propagate.py
```diff
@@ -0,0 +1,221 @@
...
+
+        Returns:
+            None
+
+        Raises:
+            ValueError: if two fixed variables x = y have different values.
+
+        """
+        if tmp:
+            self._transformed_instance = instance
```
jsiirola: In general, I have been gravitating to stateless transformations.  The "undo" information should prrobably be put on the instance you are transforming.

pyomo/core/plugins/transform/equality_propagate.py
```diff
@@ -0,0 +1,221 @@
...
+                                           descend_into=True):
+        # Check to make sure the constraint is of form v1 - v2 == 0
+        if constr.lower == 0 and constr.upper == 0 \
+                and isinstance(constr.body, _SumExpression) \
+                and len(constr.body._args) == 2 \
+                and 1 in constr.body._coef \
+                and -1 in constr.body._coef \
+                and isinstance(constr.body._args[0], _GeneralVarData) \
+                and isinstance(constr.body._args[1], _GeneralVarData) \
+                and constr.body._const == 0:
```
jsiirola: Wouldn't it be simpler to just use:

```
if value(constr.lower) == 0 and value(constr.upper) == 0 and constr.body.polynominal_degree() == 1:
    repn = generate_canonical_repn(constr.body)
    var = [v for i,v in enumerate(repn.variables) if repn.linear[i]]
    if len(var) == 2 and sorted(l for l in repn.linear if l) == [-1,1]:
        # This is an a == b constraint.
```

pyomo/core/plugins/transform/equality_propagate.py
```diff
@@ -0,0 +1,221 @@
...
+
+        Raises:
+            ValueError: if two fixed variables x = y have different values.
+
+        """
+        if tmp:
+            self._transformed_instance = instance
+        #: dict: Mapping of variable to its UID
+        var_to_id = self.var_to_id = generate_cuid_names(
+            instance.model(), ctype=Var, descend_into=True)
```
jsiirola: I am not thrilled with the implementation of `ComponentUID` at the moment (and I wrote it!).  For this application, I think that you should use the `ComponentSet` from kernel.

pyomo/core/plugins/transform/equality_propagate.py
qtothec: What happens if you do the same transformation twice at different times and want to revert one? This isn't currently something that I'm worried about, so I'll go ahead and stick the the state information on the instance, but it's something to consider.

qtothec:
Close and reopen to trigger a rebuilt because it failed in Python 2.6 on some slicer test that doesn't seem related to this PR.

pyomo/core/plugins/transform/equality_propagate.py
jsiirola: You could do an out-of-place transformation for the inner problem and then throw that model away.  Alternatively, the transformation could store a list of state data if you think this is something you would like to support.  That said, you have hit on an open debate on transformations (basically, whether the model holds all information needed to use/interpret it, or if that information can be scattered around in other locations (like writers or transformations).

qtothec:
#205 Pyro failure. Non-fragile tests appear to pass.

pyomo/contrib/preprocessing/plugins/equality_propagate.py
```diff
@@ -0,0 +1,226 @@
...
+        Raises:
+            ValueError: if two fixed variables x = y have different values.
+
+        """
+        if tmp and not hasattr(instance, '_tmp_propagate_fixed'):
+            instance._tmp_propagate_fixed = ComponentSet()
+        eq_var_map, relevant_vars = _build_equality_set(instance)
+        #: ComponentSet: The set of all fixed variables
+        fixed_vars = ComponentSet((v for v in relevant_vars if v.fixed))
+        newly_fixed = _detect_fixed_variables(instance)
```
blnicho: I think you should remove this line from this transformation since you have a separate transformation for detecting fixed variables and right now it seems like an undocumented side-effect for this transformation. I think the user should have to manually apply these transformations in a sequence.

pyomo/contrib/preprocessing/tests/test_equality_propagate.py
```diff
@@ -0,0 +1,138 @@
+"""Tests the equality set propagation module."""
+import pyutilib.th as unittest
+from pyomo.environ import (ConcreteModel, Constraint, TransformationFactory,
+                           Var, value)
+
+__author__ = "Qi Chen <https://github.com/qtothec>"
+
+
+class TestEqualityPropagate(unittest.TestCase):
```
blnicho: There are a few tests that I think need to be added:
1. A test to verify the exception throwing logic in your equality propagation transformation
2. Tests for both transformations that include more than one equality set just to be sure all the looping logic is correct (i.e. v1 == v2 and v3 == v4)
3. Tests for both transformations on indexed constraints and variables
4. Tests both transformations on models with constraints that don't fit the format you're looking for to verify that your `_build_equality_set` function will ignore them.

pyomo/contrib/preprocessing/plugins/equality_propagate.py
```diff
@@ -0,0 +1,226 @@
...
+                set2 = eq_var_map.get(v2, ComponentSet([v2]))
+                relevant_vars.update([v1, v2])
+                set1.update(set2)  # set1 is now the union
+                for v in set1:
+                    eq_var_map[v] = set1
+
+    return eq_var_map, relevant_vars
+
+
+def _detect_fixed_variables(m):
```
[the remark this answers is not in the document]
qtothec: @jsiirola had actually mentioned that some of the transformations that I've written may be better captured in the form of a variable aggregator transformation. This transformation is one of them.

I can probably use the other transformation here.

pyomo/contrib/preprocessing/plugins/equality_propagate.py
qtothec: This transformation would not do anything without the other one. The moment that you have `x=y` and fix `x=4`, the original constraint becomes `y=4`, and you aren't able to detect the `x=y` equality link.

pyomo/contrib/preprocessing/plugins/equality_propagate.py
```diff
@@ -0,0 +1,226 @@
...
+            processed.update(eq_set)
+
+    def revert(self, instance):
+        """Revert variables fixed by the transformation."""
+        for var in instance._tmp_propagate_fixed:
+            var.unfix()
+        del instance._tmp_propagate_fixed
+
+
+class VarBoundPropagator(IsomorphicTransformation):
```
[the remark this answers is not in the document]
qtothec: Done.

pyomo/contrib/preprocessing/plugins/equality_propagate.py
```diff
@@ -0,0 +1,226 @@
...
+                var = repn.variables[0]
+                coef = float(repn.linear[0])
+                const = repn.constant if repn.constant is not None else 0
+                var_val = (value(constr.lower) - value(const)) / coef
+                var.fix(var_val)
+                new_fixed_vars.add(var)
+    return new_fixed_vars
+
+
+class FixedVarPropagator(IsomorphicTransformation):
```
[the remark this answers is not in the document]
qtothec: Done.

pyomo/contrib/preprocessing/tests/test_equality_propagate.py
carldlaird: I agree with Bethany's test requests.

blnicho:
@jsiirola or @carldlaird is this PR ready to merge?

PR #196, opened 2017-08-01
Preprocessing transformations for pyomo.contrib

### LINKED ISSUE #2: Trivial Constraints and Empty Rows in NL file
This issue is a follow up from a [Pyomo forum discussion](https://groups.google.com/forum/#!topic/pyomo-forum/ufzyCfPKy-k).

CONOPT complains about empty rows being sent to it in an .nl file. In a larger model, this may happen to rows corresponding to logic constraints when a subset of discrete variables are fixed. While one remedy would be to deactivate the affected equations, it could be difficult to keep track of them. According to the CONOPT developer, AMPL will filter empty rows out on the way into CONOPT and add the proper row information (status and duals) for the empty rows on the way back. Is this something that we could do as well?

Code files for a small test case are included in a Gist: https://gist.github.com/qtothec/30b5edd277e74171088dbb91b11ea77e

qtothec:
1. Deactivating trivial constraints
2. Detection of de-facto fixed variables (based on their bound values)
3. Midpoint initialization of model variables
4. Removal of zero terms from constraints
  
Fixes #2


pyomo/core/plugins/transform/deactivate_trivial_constraints.py
```diff
@@ -0,0 +1,51 @@
...
+
+class TrivialConstraintDeactivator(IsomorphicTransformation):
+    """Deactivates trivial constraints of form constant = constant."""
+
+    alias('core.deactivate_trivial_constraints', doc=__doc__)
+
+    def __init__(self):
+        """Initialize the transformation."""
+        super(TrivialConstraintDeactivator, self).__init__()
+        self._deactivated_constrs = set()
```
jsiirola: In general I have been advocating that transformations should be stateless.  State needed for either undoing the transformation or backmapping should be stored on the model.

pyomo/core/plugins/transform/remove_zero_terms.py
```diff
@@ -0,0 +1,51 @@
...
+                                     for i in nonzero_vars_indx) + const ==
+                                 constr.upper)
+            elif constr.lower is not None:
+                constr.set_value(sum(repn.linear[i] * repn.variables[i]
+                                     for i in nonzero_vars_indx) + const >=
+                                 constr.lower)
+            elif constr.upper is not None:
+                constr.set_value(sum(repn.linear[i] * repn.variables[i]
+                                     for i in nonzero_vars_indx) + const <=
+                                 constr.upper)
```
jsiirola: This logic will fail for bounded inequality constraints (inequalities with both upper and lower bounds).

pyomo/core/plugins/transform/init_vars.py
```diff
@@ -0,0 +1,63 @@
...
+                # If LB and UB do not exist, set variable value to 0
+                var.set_value(0)
+            elif var.lb is None:
+                # if one bound does not exist, set variable value to the other
+                var.set_value(value(var.ub))
+            elif var.ub is None:
+                # if one bound does not exist, set variable value to the other
+                var.set_value(value(var.lb))
+            else:
+                var.set_value((value(var.lb) + value(var.ub)) / 2)
```
jsiirola: `2.` and not 2 to handle int division in py2.

pyomo/core/plugins/transform/deactivate_trivial_constraints.py
```diff
@@ -0,0 +1,51 @@
+# -*- coding: UTF-8 -*-
+"""Transformation to deactivate trivial constraints."""
+from pyomo.core.base.constraint import Constraint
+from pyomo.core.kernel.numvalue import value
+from pyomo.core.plugins.transform.hierarchy import IsomorphicTransformation
+from pyomo.util.plugin import alias
+
+__author__ = "Qi Chen <qichen at andrew.cmu.edu>"
```
jsiirola: We are currently working out a standard for `__authors__`, however, i suspect you don't want to include your andrew email (one hopes that is not a long-term email!).  If anything, I would expect something like a github id.

qtothec:
Close and reopen because of #205

pyomo/core/plugins/transform/deactivate_trivial_constraints.py
```diff
@@ -0,0 +1,75 @@
...
+import textwrap
+from math import fabs
+
+from pyomo.core.base.constraint import Constraint
+from pyomo.core.kernel.component_set import ComponentSet
+from pyomo.core.kernel.numvalue import value
+from pyomo.core.plugins.transform.hierarchy import IsomorphicTransformation
+from pyomo.util.plugin import alias
+
+__author__ = "Qi Chen <https://github.com/qtothec>"
```
whart222: I recommend omitting the __author__ data.  The commit record captures this information.

pyomo/core/plugins/transform/deactivate_trivial_constraints.py
```diff
@@ -0,0 +1,75 @@
...
+    """Deactivates trivial constraints of form constant = constant."""
+
+    alias('core.deactivate_trivial_constraints',
+          doc=textwrap.fill(textwrap.dedent(__doc__.strip())))
+
+    def __init__(self):
+        """Initialize the transformation."""
+        super(TrivialConstraintDeactivator, self).__init__()
+        self.tolerance = 1E-13
+
```
whart222: I never like seeing tolerance constants in code like this, even though I do it myself.  Would it make sense to have a global tolerance in something like pyomo.common (which @jsiirola is drafting to store common information/methods)?

pyomo/core/plugins/transform/deactivate_trivial_constraints.py
```diff
@@ -0,0 +1,75 @@
...
+    def _apply_to(self, instance, tmp=False, ignore_infeasible=False):
+        """Apply the transformation."""
+        if tmp and not hasattr(instance, '_tmp_trivial_deactivated_constrs'):
+            instance._tmp_trivial_deactivated_constrs = ComponentSet()
```
whart222: What is the design goal/requirement for including local data about deactivated components?  When would a user want to revert?  This logic isn't generally supported in Pyomo, so why do this for this transformation and not others?

pyomo/core/plugins/transform/deactivate_trivial_constraints.py
jsiirola: @whart222: In general, I agree.  That said, I am not sure it should hold up this PR.

A long time ago we started thinking about moving to a ConfigBlock-based approach for handling embedded constants like this.  I'd love to set up an issue to start reviving that design idea.

pyomo/core/plugins/transform/deactivate_trivial_constraints.py
```diff
@@ -0,0 +1,75 @@
...
+
+        for constr in instance.component_data_objects(
+                ctype=Constraint, active=True, descend_into=True):
+            if constr.body.polynomial_degree() == 0:
+                # Check to make sure constraint not violated.
+                if (constr.has_lb() and
+                        value(constr.body) < value(constr.lower)):
```
whart222: Shouldn't you apply the tolerance to the comparison on line 36 as well?  You check for values close to zero in the next few lines, but why aren't you checking that the value of the body is <= the value of the lower bound plus some tolerance?

pyomo/core/plugins/transform/deactivate_trivial_constraints.py
```diff
@@ -0,0 +1,75 @@
...
+                        # do nothing, move on to next constraint
+                        continue
+                    else:
+                        raise ValueError(
+                            'Trivial constraint {} violates {} ≤ {}.'
+                            .format(constr.name,
+                                    value(constr.lower),
+                                    value(constr.body)))
+                if (constr.has_ub() and
+                        value(constr.body) > value(constr.upper)):
```
whart222: Same concern as earlier.  Why isn't the tolerance used in this comparison?

pyomo/core/plugins/transform/deactivate_trivial_constraints.py
```diff
@@ -0,0 +1,75 @@
...
+
+    alias('core.deactivate_trivial_constraints',
+          doc=textwrap.fill(textwrap.dedent(__doc__.strip())))
+
+    def __init__(self):
+        """Initialize the transformation."""
+        super(TrivialConstraintDeactivator, self).__init__()
+        self.tolerance = 1E-13
+
+    def _apply_to(self, instance, tmp=False, ignore_infeasible=False):
```
whart222: What does the "tmp" argument do?

pyomo/core/plugins/transform/deactivate_trivial_constraints.py
```diff
@@ -0,0 +1,75 @@
...
+                # Constraint is fine. Deactivate it.
+                if tmp:
+                    instance._tmp_trivial_deactivated_constrs.add(constr)
+                constr.deactivate()
+
+    def revert(self, instance):
+        """Revert constraints deactivated by the transformation."""
+        for constr in instance._tmp_trivial_deactivated_constrs:
+            constr.activate()
+        del instance._tmp_trivial_deactivated_constrs
```
whart222: I recommend replacing replacing _tmp_trivial_deactivated_constrs with a ComponentSet() rather than deleting it.  If the user calls revert() twice in a row, it will cause an error to have an 'empty' value for this attribute.

pyomo/core/plugins/transform/detect_fixed_vars.py
```diff
@@ -0,0 +1,56 @@
+"""Transformation to detect variables fixed by bounds and fix them."""
+import textwrap
+from pyomo.core.base.block import generate_cuid_names
+from pyomo.core.base.var import Var
+from pyomo.core.kernel.numvalue import value
+from pyomo.core.plugins.transform.hierarchy import IsomorphicTransformation
+from pyomo.util.plugin import alias
+from six import iteritems
+
+__author__ = "Qi Chen <https://github.com/qtothec>"
```
whart222: As before, I think this should be omitted.

pyomo/core/plugins/transform/detect_fixed_vars.py
```diff
@@ -0,0 +1,56 @@
...
+
+        for var in instance.component_data_objects(
+                ctype=Var, descend_into=True):
+            if var.fixed or var.lb is None or var.ub is None:
+                continue
+            if value(var.lb) == value(var.ub):
+                self._old_var_values[var_to_id[var]] = var.value
+                var.fix(var.lb)
+
+    def revert(self):
```
whart222: As before, what is the design goal for supporting the revert() method?  This is inconsistent with other model transformations.

pyomo/core/plugins/transform/detect_fixed_vars.py
```diff
@@ -0,0 +1,56 @@
...
+    alias('core.detect_fixed_vars',
+          doc=textwrap.fill(textwrap.dedent(__doc__.strip())))
+
+    def __init__(self):
+        """Initialize the transformation."""
+        super(FixedVarDetector, self).__init__()
+        self._old_var_values = {}
+        self._transformed_instance = None
+
+    def _apply_to(self, instance, tmp=False):
```
whart222: What is "tmp" used for?

pyomo/core/plugins/transform/detect_fixed_vars.py
```diff
@@ -0,0 +1,56 @@
...
+            instance.model(), ctype=Var, descend_into=True)
+        #: dict: Mapping of UIDs to variables
+        self.id_to_var = dict(
+            (cuid, obj) for obj, cuid in iteritems(var_to_id))
+
+        for var in instance.component_data_objects(
+                ctype=Var, descend_into=True):
+            if var.fixed or var.lb is None or var.ub is None:
+                continue
+            if value(var.lb) == value(var.ub):
```
whart222: You need to use a tolerance in this equality comparison

pyomo/core/plugins/transform/init_vars.py
```diff
@@ -0,0 +1,68 @@
...
+from pyomo.core.base.var import Var
+from pyomo.core.kernel.numvalue import value
+from pyomo.core.plugins.transform.hierarchy import IsomorphicTransformation
+from pyomo.util.plugin import alias
+
+__author__ = "Qi Chen <https://github.com/qtothec>"
+
+
+class InitMidpoint(IsomorphicTransformation):
+    """Initializes variables to the midpoint of their bounds."""
```
whart222: This documentation does not completely reflect the intent of this transformation.  The checks below consider other cases than just the midpoint computation.

pyomo/core/plugins/transform/init_vars.py
```diff
@@ -0,0 +1,68 @@
...
+                var.set_value(value(var.ub))
+            elif var.ub is None:
+                # if one bound does not exist, set variable value to the other
+                var.set_value(value(var.lb))
+            else:
+                var.set_value((value(var.lb) + value(var.ub)) / 2.)
+
+
+class InitZero(IsomorphicTransformation):
+    """Initializes variables to zeros."""
```
whart222: This documentation does not completely reflect the intent of this transformation.  The checks below consider other cases than just assigning variables to the value zero.

pyomo/core/plugins/transform/remove_zero_terms.py
```diff
@@ -0,0 +1,64 @@
...
+from pyomo.core.base.constraint import Constraint
+from pyomo.core.plugins.transform.hierarchy import IsomorphicTransformation
+from pyomo.repn.canonical_repn import generate_canonical_repn
+from pyomo.util.plugin import alias
+
+__author__ = "Qi Chen <https://github.com/qtothec>"
+
+
+class RemoveZeroTerms(IsomorphicTransformation):
+    """Looks for 0 * var in a constraint and removes it.
```
whart222: This transformation will need to be reworked to support Pyomo5 expressions.  Hence, I'm not going to provide a detailed review here.

pyomo/core/plugins/transform/deactivate_trivial_constraints.py
whart222: I agree that this shouldn't hold up the PR.

pyomo/core/plugins/transform/deactivate_trivial_constraints.py
qtothec: Sounds good. I think I originally put these in with the thought that my colleagues here know that they can bug me about support for them.

pyomo/core/plugins/transform/deactivate_trivial_constraints.py
qtothec: A constraint may become trivial as a result of some variables (e.g. discrete ones) being fixed. In some cases, we do not want to send these constraints to an NLP solver (like IPOPT), because they will return nonsensical duals that foul up cutting plane generation, as @bernalde found. However, we may want to restore these constraints later when the variables are unfixed.

pyomo/core/plugins/transform/deactivate_trivial_constraints.py
qtothec: Specifies if the user has any intention of reversing the transformation in the future. I'm going to go clean up the documentation.

pyomo/core/plugins/transform/deactivate_trivial_constraints.py
qtothec: Good question

pyomo/core/plugins/transform/deactivate_trivial_constraints.py
qtothec: I would expect to get an error for trying to revert a transformation twice, since it would be a repetition of a state change.

qtothec:
I started a design discussion about the reversions in #277.

Since these are four largely independent transformations, should I break this up into 4 different PRs?

Related to documentation, are you looking for improved docstrings, or is there something else you had in mind?

whart222:
I don't think it's necessary to break up this PR.  That's more work without any specific up-side.

For now, let's update the docstrings to provide more detail about the transformations.  Ultimately, these transformations should be documented in the Sphinx pages.  We're going to start requiring that for new PRs, but for an old one like this I propose that we (a) update the docstrings and (b) create a ticket to add Sphinx documentation.

qtothec:
I cleaned up the docstrings some more and opened #279. Let me know if there's more changes that you would like to see.

PR #198, opened 2017-08-09
Add transformation to fix linear elements of 0 == x1 + x2, where x's are non-negative, to zero

qtothec:
This transformation detects when all elements of a linear constraint must be equal to zero by checking their bounds and signs.

### DIFF
--- a/pyomo/core/plugins/transform/zero_sum_propagator.py
+++ b/pyomo/core/plugins/transform/zero_sum_propagator.py
@@ -1,0 +1,81 @@
+"""Transformation to propagate a zero value to terms of a sum."""
+from pyomo.core.base.constraint import Constraint
+from pyomo.core.kernel.numvalue import value
+from pyomo.core.plugins.transform.hierarchy import IsomorphicTransformation
+from pyomo.repn.canonical_repn import generate_canonical_repn
+from pyomo.util.plugin import alias
+
+from six.moves import range
+
+__author__ = "Qi Chen <qichen at andrew.cmu.edu>"
+
+
+class ZeroSumPropagator(IsomorphicTransformation):
+    """Propagates fixed-to-zero for sums of only positive (or negative) vars.
+
+    If x is fixed to zero and x == x1 + x2 + x3 and x1, x2, x3 are all
+    non-negative or all non-positive, then x1, x2, and x3 will be fixed to
+    zero.
+
+    """
+
+    alias('core.propagate_zero_sum', doc=__doc__)
+
+    def __init__(self):
+        """Initialize the transformation."""
+        super(ZeroSumPropagator, self).__init__()
+
+    def _apply_to(self, instance):
+        """Apply the transformation.
+
+        Args:
+            instance (Block): the block on which to search for x == sum(var)
+                constraints. Note that variables may be located anywhere in
+                the model.
+
+        Returns:
+            None
+
+        """
+        for constr in instance.component_data_objects(ctype=Constraint,
+                                                      active=True,
+                                                      descend_into=True):
+            if not constr.body.polynomial_degree() == 1:
+                continue  # constraint not linear. Skip.
+
+            repn = generate_canonical_repn(constr.body)
+            if (constr.has_ub() and (
+                (repn.constant is None and value(constr.upper) == 0) or
+                    repn.constant == value(constr.upper))):
+                # vars <= 0
+                # all var terms need to be non-negative
+                if all(
+                    # variable has 0 coefficient
+                    repn.linear[i] == 0 or
+                    # variable is non-negative and has non-negative coefficient
+                    (repn.variables[i].has_lb() and
+                     value(repn.variables[i].lb) >= 0 and
+                     repn.linear[i] >= 0) or
+                    # variable is non-positive and has non-positive coefficient
+                    (repn.variables[i].has_ub() and
+                     value(repn.variables[i].ub) <= 0 and
+                     repn.linear[i] <= 0) for i in range(len(repn.linear))):
+                    for var in repn.variables:
+                        var.fix(0)
+                    continue
+            if (constr.has_lb() and (
+                (repn.constant is None and value(constr.lower) == 0)) or
+                    (repn.constant == value(constr.lower))):
+                # vars >= 0
+                # all var terms need to be non-positive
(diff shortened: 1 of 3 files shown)


pyomo/core/plugins/transform/zero_sum_propagator.py
```diff
@@ -0,0 +1,81 @@
...
+                    # variable is non-negative and has non-negative coefficient
+                    (repn.variables[i].has_lb() and
+                     value(repn.variables[i].lb) >= 0 and
+                     repn.linear[i] >= 0) or
+                    # variable is non-positive and has non-positive coefficient
+                    (repn.variables[i].has_ub() and
+                     value(repn.variables[i].ub) <= 0 and
+                     repn.linear[i] <= 0) for i in range(len(repn.linear))):
+                    for var in repn.variables:
+                        var.fix(0)
```
jsiirola: This is too aggressive: consider the following expression `m.x + m.y + m.z - m.y == m.w`.  This will result in a linear coefficient matrix of `[m.x, m.y, m.z, m.w] <-> [ 1, 0, 1, -1]`.  This transformation should NOT result in fixing `y` to zero.

pyomo/core/plugins/transform/zero_sum_propagator.py
```diff
@@ -0,0 +1,81 @@
...
+                                                      descend_into=True):
+            if not constr.body.polynomial_degree() == 1:
+                continue  # constraint not linear. Skip.
+
+            repn = generate_canonical_repn(constr.body)
+            if (constr.has_ub() and (
+                (repn.constant is None and value(constr.upper) == 0) or
+                    repn.constant == value(constr.upper))):
+                # vars <= 0
+                # all var terms need to be non-negative
```
jsiirola: These two comments appear to be contradictory.  I understand what you are going after, but a little more description would be helpful

pyomo/core/tests/transform/test_zero_sum_propagate.py
```diff
@@ -0,0 +1,73 @@
...
+    """Tests zero sum propagation."""
+
+    def test_non_negative_propagate(self):
+        """Test zero sum propagation for non-negative vars."""
+        m = ConcreteModel()
+        m.v1 = Var(initialize=0)
+        m.v2 = Var(initialize=2, domain=NonNegativeReals)
+        m.v3 = Var(initialize=3, domain=NonNegativeReals)
+        m.v4 = Var(initialize=4, domain=NonNegativeReals)
+        m.c1 = Constraint(expr=m.v1 == m.v2 + m.v3 + m.v4)
```
jsiirola: For the sake of completeness, can you also test `m.v2 + m.v3 + m.v4 == m.v1`?  Mathematically it is the same, but results in opposite signs for the linear coefficients.

pyomo/core/plugins/transform/zero_sum_propagator.py
```diff
@@ -0,0 +1,81 @@
...
+from pyomo.core.plugins.transform.hierarchy import IsomorphicTransformation
+from pyomo.repn.canonical_repn import generate_canonical_repn
+from pyomo.util.plugin import alias
+
+from six.moves import range
+
+__author__ = "Qi Chen <qichen at andrew.cmu.edu>"
+
+
+class ZeroSumPropagator(IsomorphicTransformation):
```
jsiirola: I am OK with this transformation, but it is really a special case of a bounds propagator.  At some point, we should make it more general and do proper bounds propagation.

pyomo/core/plugins/transform/zero_sum_propagator.py
```diff
@@ -0,0 +1,81 @@
...
+                    # variable has 0 coefficient
+                    repn.linear[i] == 0 or
+                    (repn.variables[i].has_lb() and
+                     value(repn.variables[i].lb) >= 0 and
+                     repn.linear[i] <= 0) or
+                    (repn.variables[i].has_ub() and
+                     value(repn.variables[i].ub) <= 0 and
+                     repn.linear[i] >= 0) for i in range(len(repn.linear))):
+                    for var in repn.variables:
+                        var.fix(0)
```
jsiirola: ...same problem as above...

pyomo/core/plugins/transform/zero_sum_propagator.py
qtothec: Agreed. Unless I can find a MSc student or an undergrad to do it, I am only likely to hit the cases that are immediately obvious with my models.

pyomo/core/plugins/transform/zero_sum_propagator.py
qtothec: Oops.

qtothec:
Close and reopen in order to trigger rebuild due to network connectivity-related failure.

pyomo/core/plugins/transform/zero_sum_propagator.py
```diff
@@ -0,0 +1,83 @@
...
+                    repn.linear[i] == 0 or
+                    # variable is non-negative and has non-negative coefficient
+                    (repn.variables[i].has_lb() and
+                     value(repn.variables[i].lb) >= 0 and
+                     repn.linear[i] >= 0) or
+                    # variable is non-positive and has non-positive coefficient
+                    (repn.variables[i].has_ub() and
+                     value(repn.variables[i].ub) <= 0 and
+                     repn.linear[i] <= 0) for i in range(len(repn.linear))):
+                    for i in range(len(repn.linear)):
```
jsiirola: I generally prefer `for i,l in enumerate(repn.linear)` over `for i in range(len(repn.linear))` for two reasons: 1) it is cleaner code (and what the `enumerate` method is for) and 2) because `range` behaves differently in Python2 and 3: in python2 it returns a full list and not a generator (which can be costly for large lists).

jsiirola: (Also note that this idiom appears several times in this PR)

qtothec: I was overwriting `range` with the `six.moves` import, but yes, `enumerate` was a missing part of my python vocabulary.

qtothec:
Farmer failure rebuild.

PR #203, opened 2017-08-21
Add documentation to public functions, even the deprecated ones. That…

qtothec:
… way, they show up when you do help(Suffix). Otherwise, they just look blank, and I have to go digging.

Also: applied autoPEP8

### DIFF
--- a/pyomo/core/base/suffix.py
+++ b/pyomo/core/base/suffix.py
@@ -35,7 +35,8 @@ logger = logging.getLogger('pyomo.core')
 #   - active_suffix_generator
 #   - suffix_generator
 
-def active_export_suffix_generator(a_block,datatype=False):
+
+def active_export_suffix_generator(a_block, datatype=False):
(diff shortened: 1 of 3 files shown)


qtothec:
Triggering rebuild to see if it will help with #204

qtothec:
You would need to pass it the logger somehow:
```
@deprecated("method 'foo' has been replaced by method 'bar'", logger)
def foo(self, *args):
    # the function...
    pass
```

jsiirola:
You could just grab & use the "global" `pyomo.core` logger:
```
import logging
logging.getLogger('pyomo.core').warn('DEPRECATED: ...')
```

qtothec:
I'll piece together something simple and we can look at it.

qtothec:
@jsiirola How's it look now?

jsiirola:
This is great.  I just added/expanded some of the tests.

Issue #116, opened 2017-02-20
DeveloperError when using a singleton set as a index for an IndexedConstraint


qtothec:
See small example below:
```
qichen@QC-CMU-Tower:~$ pyothon
Python 2.7.11+ (default, Apr 17 2016, 14:00:29) 
[GCC 5.3.1 20160413] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from pyomo.environ import *
>>> m = ConcreteModel()
>>> m.s = Set(initialize=['one'])
>>> m.c = Constraint(m.s)
>>> m.c[m.s].deactivate()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/qichen/.solvers/pyomo/src/pyomo/pyomo/core/base/indexed_component.py", line 492, in __getitem__
    return self._default(ndx)
  File "/home/qichen/.solvers/pyomo/src/pyomo/pyomo/core/base/indexed_component.py", line 655, in _default
    % (self.__class__.__name__,))
pyomo.util._config.DeveloperError: Internal Pyomo implementation error:
	'Derived component IndexedConstraint failed to define _default().'
	Please report this to the Pyomo Developers.
>>> 
```
On the other hand, nothing breaks when I try to do the same thing with IndexedVar
```
>>> from pyomo.environ import *
>>> m = ConcreteModel()
>>> m.s = Set(initialize=['one'])
>>> m.x = Var(m.s, initialize=1)
>>> m.x.display()
x : Size=1, Index=s
    Key : Lower : Value : Upper : Fixed : Stale : Domain
    one :  None :     1 :  None : False : False :  Reals
>>> m.x[m.s].display()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: '_GeneralVarData' object has no attribute 'display'
>>> m.x[m.s].deactivate()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: '_GeneralVarData' object has no attribute 'deactivate'
>>> 
```

jsiirola:
#133 "fixed" the Constraint example, although it will still generate an error:
```
>>> from pyomo.environ import *
>>> m = ConcreteModel()
>>> m.s = Set(initialize=['one'])
>>> m.c = Constraint(m.s)
>>> m.c[m.s].deactivate()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #212, opened 2017-08-31
[WIP] Rewrite of the GDP core components and BigM/CHull transformations

jsiirola:
I believe that this is ready to merge (modulo merge conflicts).

@emma58, @qtothec do you agree?


qtothec:
Seems that there's a number of Travis failures

emma58:
About the failing tests from Travis:
1) It looks like the ones in pyomo.mpec are all just because they are diffing files that don't know how a Disjunction pprints now (they don't know about the XOR indicator var or the list of Disjuncts).
2) The chull jobshop ones were my fault--I just pushed updated LP files for those
3) The cutting planes ones are all failing because I have the solver for the separation problem hard-coded as Gurobi... It doesn't matter which solver it is as long as it can handle the quadratic objective, and it shouldn't be hard-coded in the long run anyway. So should I change this to something else @jsiirola?
4) I'm testing create_using by pprinting the apply_to model and the create_using one, capturing stdout from each, and then diffing the two strings. That doesn't seem to work in the Travis run. I'm not sure why?

qtothec:
I just ran full speed into https://github.com/emma58/pyomo/blob/8529b17eca3467849c569120696d72d62a4d0d5b/pyomo/gdp/plugins/bigm.py#L485.

`newConstraint.add((i, 'lb'), c.lower <= c.body - M_expr)` falls flat on its face when `i` is a tuple, i.e. the original constraint was indexed over multiple sets.

The TODO comment mentions #116, which I don't quite understand, because it's a similar but separate issue that's going on.

qtothec:
https://github.com/emma58/pyomo/pull/4 should resolve this issue. Took me a while to figure out what was going on, because of #9

qtothec:
The commits from 5 days ago (96cde3b through 73b7f95) changed something that's causing my case studies to behave oddly.

jsiirola:
@qtothec: Can you add any details or share an example?

qtothec:
@jsiirola I don't have any self-contained examples. Trying to figure out what's going on right now.

qtothec:
It looks like 7bfb0ab disagrees in some way with how I have implemented my set covering initialization in GDPopt.

qtothec:
@jsiirola If you activate an active Disjunct, it will unnecessarily unfix the indicator var. I'm not sure if this is what's causing my issues, but it's a concern.

jsiirola:
@qtothec: OK - that is interesting.  I am curious, though: what does it mean to have an active disjunct with a fixed indicator variable?  More importantly, I would expect the following behavior:

1. If I deactivate a disjunct, then that disjunct is effectively "not in" the model.  It's indicator_var should be fixed to 0 so that it is effectively removed as a option from any global logical constraints or disjunctions.
2. If I activate a disjunct, the indicator_var going in should be fixed to 0.  An active disjunct with a fixed indicator_var (equal to 0) is still deactivated, so that makes no sense, and the indicator_var should be freed.

Is the error you are seeing because you are activating a disjunct that had not previously been deactivated?

qtothec:
@jsiirola well, the var_mover hack currently calls `.activate()`. I think that might have something to do with it.

jsiirola:
Ahh - I will look at that.

qtothec:
Also, the behavior on `activate()` does not change after reclassification

jsiirola:
I think the fix in 0925986 is correct: as the reclassifier is a hack (and marked as such), the weird processing of the `indicator_var` is captured there.  I also considered making it so that Disjunct's `activate` (and `deactivate`) only unfixed/fixed the indicator_var when the Disjunct's ctype is a derivative of Disjunct, but I don't think that is needed or wise.

qtothec:
`AttributeError: 'IndexedDisjunct' object has no attribute '_activate_without_unfixing_indicator'`

qtothec:
Ok, now I'm happy again.

pyomo/gdp/disjunct.py
```diff
@@ -2,83 +2,118 @@
...
+    ActiveComponent, ActiveComponentData, ComponentData
+)
+from pyomo.core.base.numvalue import native_types
+#=======
+#from pyomo.core import *
+#from pyomo.core.base.plugin import register_component
+#from pyomo.core.base.constraint import (SimpleConstraint,
+#                                        IndexedConstraint,
+#                                        _GeneralConstraintData)
+#>>>>>>> master
```
qtothec: I don't think that you mean to have this in here.

jsiirola: Yup - that is completely developer notes (it is to remind me where we were importing things from while I remove `from x import *` from the codebase).

jsiirola:
Closing this PR.  I will re-open it against Pyomo/pyomo:gdp-rewrite when we are ready to merge.

Issue #211, opened 2017-08-31
Figure out location to put model introspection tools


qtothec:
There is general demand for model introspection utilities like determining the number of degrees of freedom. Where should we put these utilities? People don't like `pyomo.debug`, apparently, so what is it going to be?
`pyomo.core.introspect`?

blnicho:
How about `pyomo.diagnostics`?

qtothec:
That sounds fine to me. Do we want to specify at all that it's model diagnostics rather than solver diagnostics?

blnicho:
I don't think it's necessary to distinguish them. I think it would be fine to have both in the same module/folder.

qtothec:
So, can I assume that everyone is happy with this and move forward with it? =D

whart222:
I would assume that we'll have different introspection tools for different types of models.  Yes?

How would they be distinguished in this common directory?

blnicho:
Discussion from the developer's meeting today:
@DLWoodruff likes 'diagnostics' instead of 'introspection'. 'Introspection' is a loaded word in Python 

Could distinguish tools for different types of models using function naming or subfolders (i.e. pyomo.diagnostics.lp or pyomo.diagnostics.nlp)

Diagnostic capabilities for special modeling components dae/gdp/etc could live in those directories (pyomo.dae.diagnostics)

qtothec:
Sounds like a good idea to me. Where would something more general like `degrees_of_freedom(m)` go?

blnicho:
I would say in the main diagnostics folder such that you could import it like:

    from pyomo.diagnostics import calculate_degrees_of_freedom

whart222:
PR #217 adds diagnostic tools in pyomo.util.diagnostics.  I think we can add additional tools there, so I'm closing this issue.

jsiirola:
I am comfortable closing this issue, but need to reference #267 and the newly-created #321 that will track the final disposition of `pyomo.util`

PR #217, opened 2017-09-07
First pass at diagnostic functions for Pyomo models.

qtothec:
This relates to #211 and is a first-pass at including helpful diagnostic functions.

Naming-related suggestions and comments appreciated.


whart222:
I agree with the general concern expressed by @carldlaird.  It's not obvious whether these sort of utility functions necessarily should be added to Pyomo.  But for now, I'm also comfortable adding these to the pyomo.util sub-package.

PR #220, opened 2017-09-08
Transformation to strip bounds from model

### LINKED ISSUE #218: Strip/Restore Bounds Feature
It looks like at least in solving square problems with IPOPT it is useful to remove the bounds on variables before solving.  There are likely other situations where this would be helpful.  When solving an optimization problem with IPOPT you usually need to bring the bounds back.  A simple addition to the solve function that could send problems to the solver unbounded may suffice here. There are also strip_bounds and restore_bounds functions in the IDAES stuff that save the bounds then remove them in the Pyomo problem.

Anyway I think this would be a good addition to Pyomo.

qtothec:
This fixes #218


eslickj:
I haven't dug too far into this so this could be my fault and I don't have time to provide a good example right now, but I had this error on my first try.
AttributeError                            Traceback (most recent call last)
<ipython-input-11-22f0dbc30b9c> in <module>()
----> 1 stp.apply_to(model)

/home/jeslick/git/pyomo/pyomo/core/base/plugin.pyc in apply_to(self, model, **kwds)
    333         if not hasattr(model, '_transformation_data'):
    334             model._transformation_data = TransformationData()
--> 335         self._apply_to(model, **kwds)
    336 
    337     def create_using(self, model, **kwds):

/home/jeslick/git/pyomo/pyomo/core/plugins/transform/strip_bounds.pyc in _apply_to(self, instance, tmp)
     40                 if tmp:
     41                     instance._tmp_var_bound_strip_lb[var] = var.lb
---> 42                 var.set_lb(None)
     43             if var.has_ub():
     44                 if tmp:

AttributeError: '_GeneralVarData' object has no attribute 'set_lb'

qtothec:
Syntax error on my part. One second.

eslickj:
Thanks.  I didn't test extensively but strip and revert seem to work.

qtothec:
Excellent. Once someone finds the time to write some simple unit tests for this, it can be merged.

qtothec:
@eslickj I just added the ability to strip domains as well. It's on by default.

pyomo/core/plugins/transform/strip_bounds.py
```diff
@@ -0,0 +1,68 @@
...
+                if tmp:
+                    instance._tmp_var_bound_strip_lb[var] = var.lb
+                var.setlb(None)
+            if var.has_ub():
+                if tmp:
+                    instance._tmp_var_bound_strip_ub[var] = var.ub
+                var.setub(None)
+
+    def revert(self, instance):
+        """Revert variables fixed by the transformation."""
```
blnicho: Update the docstring, this transformation is not fixing variables

pyomo/core/tests/transform/test_strip_bounds.py
```diff
@@ -0,0 +1,93 @@
+"""Tests stripping of variable bounds."""
+import pyutilib.th as unittest
+
+from pyomo.environ import (Binary, ConcreteModel, Integers, NonNegativeReals,
+                           PositiveReals, Reals, TransformationFactory, Var)
+
+__author__ = "Qi Chen <https://github.com/qtothec>"
```
blnicho: Remove author information, the commit log captures this information

pyomo/core/tests/transform/test_strip_bounds.py
```diff
@@ -0,0 +1,93 @@
...
+    def test_strip_bounds(self):
+        """Test bound stripping and restoration."""
+        m = ConcreteModel()
+        m.v0 = Var(bounds=(2, 4))
+        m.v1 = Var(domain=NonNegativeReals)
+        m.v2 = Var(domain=PositiveReals)
+        m.v3 = Var(bounds=(-1, 1))
+        m.v4 = Var(domain=Binary)
+        m.v5 = Var(domain=Integers)
+
```
blnicho: Recommend adding a test case where you have both bounds and a domain specified.

pyomo/core/tests/transform/test_strip_bounds.py
```diff
@@ -0,0 +1,93 @@
...
+    def test_no_strip_domain(self):
+        """Test bounds stripping without domain change."""
+        m = ConcreteModel()
+        m.v0 = Var(bounds=(2, 4))
+        m.v1 = Var(domain=NonNegativeReals)
+        m.v2 = Var(domain=PositiveReals)
+        m.v3 = Var(bounds=(-1, 1))
+        m.v4 = Var(domain=Binary)
+        m.v5 = Var(domain=Integers)
+
```
blnicho: Recommend adding a test case where you have both bounds and a domain specified.

pyomo/core/plugins/transform/strip_bounds.py
```diff
@@ -0,0 +1,68 @@
...
+            strip_domains (bool, optional): strip the domain for discrete
+                variables as well
+            tmp (bool, optional): Whether the bound stripping will be
+                temporary. If so, store information for reversion.
+
+        Returns:
+            None
+
+        """
+        if tmp and not hasattr(instance, '_tmp_var_bound_strip_lb'):
```
blnicho: I'm not sure I agree with this logic. If `instance` already has one of these component maps then I would expect the transformation to throw an error saying that the transformation has already been applied and must be reverted before it can be applied again. I can't think of a case where you would want to strip the bounds multiple times before reverting unless you are adding variables to the block after the transformation. I would be worried about over-writing information by reusing an existing component map.

pyomo/core/plugins/transform/strip_bounds.py
```diff
@@ -0,0 +1,68 @@
...
+                    instance._tmp_var_bound_strip_lb[var] = var.lb
+                var.setlb(None)
+            if var.has_ub():
+                if tmp:
+                    instance._tmp_var_bound_strip_ub[var] = var.ub
+                var.setub(None)
+
+    def revert(self, instance):
+        """Revert variables fixed by the transformation."""
+        for var, lb in iteritems(instance._tmp_var_bound_strip_lb):
```
blnicho: Will you ever run into a case where a variable has been completely removed from the instance between applying this transformation and reverting it? Would it be safer to iterate over all the Var Data objects again and look-up that object in the component maps?

pyomo/core/tests/transform/test_strip_bounds.py
```diff
@@ -0,0 +1,93 @@
+"""Tests stripping of variable bounds."""
```
blnicho: I think you should add a test for the logic at the beginning of the transformation where you check if the instance already has the component maps.

pyomo/core/plugins/transform/strip_bounds.py
```diff
@@ -0,0 +1,68 @@
...
+        super(VariableBoundStripper, self).__init__()
+
+    def _apply_to(self, instance, strip_domains=True, tmp=False):
+        """Apply the transformation.
+
+        Args:
+            instance (Block): the block on which to strip variable bounds
+            strip_domains (bool, optional): strip the domain for discrete
+                variables as well
+            tmp (bool, optional): Whether the bound stripping will be
```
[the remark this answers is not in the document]
qtothec: I'm going to wait on this for resolution of #277, since it would affect some other existing transformations as well.

qtothec:
@eslickj Do you think this might be something that we could move back into IDAES?

eslickj:
@qtothec Yeah.  That makes sense to me.  I'm eager to start using it.

pyomo/core/plugins/transform/strip_bounds.py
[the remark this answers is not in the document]
carldlaird: I don't see any need to wait for resolution of #277 (It will take longer than the 5.4 release deadline). If we want this in the 5.4 release, we should proceed with the requested changes as is.

pyomo/core/plugins/transform/strip_bounds.py
carldlaird: I agree with Bethany. Qi, any comment?

carldlaird:
Qi, would you like to see this in the 5.4 release? If so, let's not wait for #277, and refactor this when that issue is resolved.

pyomo/core/plugins/transform/strip_bounds.py
qtothec: Completely removing a variable from an instance sounds like asking for trouble to me

qtothec:
Sure. I've made the relevant changes.

pyomo/core/plugins/transform/strip_bounds.py
carldlaird: Alas, this is why we need the design discussion for transformations - should do it soon. I think it is reasonable to add a condition (assumption) that the model structure has not changed between these calls. If a user wants that behavior, they can write a specialized "revert" using the data.

qtothec: I'll switch to doing the lookups, since I don't feel strongly enough about this either way right now.

pyomo/core/tests/transform/test_strip_bounds.py
```diff
@@ -0,0 +1,107 @@
...
+        m.v5 = Var(domain=Integers, bounds=(15, 16))
+
+        xfrm = TransformationFactory('core.strip_var_bounds')
+        xfrm.apply_to(m, reversible=True)
+        # At this point, component maps for reversion already exist.
+        with self.assertRaises(RuntimeError):
+            xfrm.apply_to(m, reversible=True)
+
+    def test_strip_bounds(self):
+        """Test bound stripping and restoration."""
```
jsiirola: This is nit-picking, but I prefer that test methods NOT use docstrings (use general comments instead).  The problem is that when the test fails (or when nose is running in verbose mode), nose will report the docstring instead of the fully-qualified test name (`module.module.class.method`), making it tough to actually *find* the test in question.

pyomo/core/tests/transform/test_strip_bounds.py
```diff
@@ -0,0 +1,107 @@
...
+        self.assertEqual(m.v5.bounds, (15, 16))
+        self.assertEqual(m.v0.domain, Reals)
+        self.assertEqual(m.v1.domain, NonNegativeReals)
+        self.assertEqual(m.v2.domain, PositiveReals)
+        self.assertEqual(m.v3.domain, Reals)
+        self.assertEqual(m.v4.domain, Binary)
+        self.assertEqual(m.v5.domain, Integers)
+
+    def test_no_strip_domain(self):
+        """Test bounds stripping without domain change."""
```
jsiirola: Same as above: I prefer comments over docstrings.

pyomo/core/tests/transform/test_strip_bounds.py
qtothec: This feels like it ought to be more of a nose option setting than a docstring vs. comment issue, because removing the docstrings makes pyflakes angry at me for not having a docstring on a public method.

carldlaird: Here is a workaround that allows docstrings on tests, but also reports the test name instead of the docstring. Would this work?

https://www.saltycrane.com/blog/2012/07/how-prevent-nose-unittest-using-docstring-when-verbosity-2/

jsiirola: Yes.  I think that might work.  I think we can put into pyutilib.th -- which would fix all of Pyomo.

jsiirola: @whart222: do you see any reason not to implement a default `shortDescription` method for `pyutilib.th.pyunit.TestCase`?

jsiirola:
@qtothec:  What do you think of moving this plugin to `pyomo.contrib.preprocessing`?

qtothec:
@jsiirola Seems reasonable to me for the time being.

jsiirola:
I spoke with @blnicho, and she agrees with the move to contrib, so this is good to merge.

PR #221, opened 2017-09-09
Add GAMS Solver documentation

gseastream:
In my last pull request I removed the `load_solutions` documentation because it said the solver didn't support `load_solutions=False` which was incorrect, but I overlooked the fact that I should have replaced it with correct documentation.

Also I changed the line endings for GAMS.py because I didn't realize they were still Windows carriage returns.


jsiirola:
This looks good.

PR #223, opened 2017-09-11
Persistent interfaces

michaelbynum:
I think the persistent interfaces is about ready to merge into master. The CPLEXPersistent interface could still use some performance improvements, but the current performance is reasonable. I also still need to add some tests. However, I have been using the GurobiPersistent interface extensively, and it seems to be working pretty well.

Please let me know if anyone has any comments or concerns.

Michael


coroa:
We just integrated the persistent solver interfaces into [PyPSA](https://github.com/FRESNA/PyPSA/tree/persistent_solver) and I have to congratulate you, they were easy to start working with, and finally enable the interactiveness necessary for iterated optimizations of large-scale problems! Three questions:

1. When is it available in `master`?
2. Did you test, whether renaming the variables to x<n> provides any benefits, or whether it would be better to enable symbolic_solver_labels by default?
3. Any chances for a persistent solver interface to ipopt?

michaelbynum:
Thanks! I am glad they are working well for you. That means a lot to me.

1. I am hoping the persistent interfaces will be available on master within
the next couple weeks. In the mean time, I am trying to keep the
persistent_interfaces branch up to date with master.
2. I did not do any performance testing with symbolic_solver_labels -
sorry. My guess is that symbolic_solver_labels=False is faster, but it is
just a guess.
3. I think many of us would love a persistent interface to ipopt. I
honestly do not know if we will have support for such an interface any time
soon. Maybe @jsiirola or @carldlaird can comment more on this.

Other than myself, I think you may be the first to be using the persistent
interfaces. If you run into any problems, please let me know so that I can
fix them!

qtothec:
I'm confused. Why is this being closed without being merged?

michaelbynum:
It is not ready to be merged. I will recreate the oil request when it is
ready.

Michael

coroa:
@michaelbynum that's unfortunate. We're quite happy users of it. Can you explain in what cases it is fine to continue using it and where it fails? Will the persistent_interfaces branch persist?

Thanks, Jonas

michaelbynum:
I only closed the pull request because it is not ready to be merged. I do
intend to get the persistent interfaces merged into master, I just need to
fix a couple things:

1.) The load_solutions=False functionality is not backwards compatible. In
fact, for the direct interfaces, there is no way to load solutions at a
later time if load_solutions is set to False.
2.) There is a problem with the Cplex direct/persistent interface with
quadratic constraints.

Michael

carldlaird:
Michael, can you comment on whether or not people can/should still use your branch code while these changes are being made?

michaelbynum:
I am confident the gurobi_persistent interface is ready to be used. The
cplex_persistent interface needs more testing.

Michael

PR #225, opened 2017-09-25
convert to and complete the entire Iterating

sgjkkx:


blnicho:
Could you add a description of what this is? Why do you want to merge it into the dae-element-initializer branch instead of master? 

If this is just an example extension using Pyomo the more appropriate place for it might be the PyomoGallery repo.

jsiirola:
This appears to be an accidental PR.  There is only one commit relative to master (8bc64ba), and that appears to be @sgjkkx's development code for LU factorization and does not have a specific tie to Pyomom.  I am closing this PR.

PR #235, opened 2017-10-10
[WIP] Scaling transformation

carldlaird:
This code is for a new scaling transformation that uses the values in 'scaling_factor' suffix to scale variables, objectives, and constraints
- includes support for create_using and apply_to
- includes a new method 'propagate_solution' to send the primal (and duals) from the scaled model back to the original model. This method should likely be added to the transformation base class as I suspect there are others that require this functionality
- added a test that compares the solution of the unscaled with the solution of the scaled that is then backmapped to the original
- added new 'Transformations' section in the library_reference of the online documentation
- added autosummary and autoclass to create online docs from the docstrings
- added a transformations directory in examples/pyomo
- added a short example in examples/pyomo/transformations/scaling.py


pyomo/core/plugins/transform/scaling.py
```diff
@@ -0,0 +1,269 @@
...
+                    stack.append(child)
+        elif hasattr(e, '_args'):
+            for i, child in enumerate(e._args):
+                if isinstance(child, _VarData):
+                    if child in sub_component_map:
+                        new_expr = clone_expression(sub_component_map[child])
+                        e._args = _replace_entry_in_list_or_tuple(e._args, new_expr, i)
+                else:
+                    stack.append(child)
+
```
jsiirola: What about `Expression` objects?  Should you clone the subexpression before substituting?  It would break the "shared subexpression", but would prevent double substitutions.

pyomo/core/plugins/transform/scaling.py
```diff
@@ -0,0 +1,269 @@
...
+                if isinstance(child, _VarData):
+                    if child in sub_component_map:
+                        e._denominator[i] = sub_component_map[i]
+                else:
+                    stack.append(child)
+        elif hasattr(e, '_args'):
+            for i, child in enumerate(e._args):
+                if isinstance(child, _VarData):
+                    if child in sub_component_map:
+                        new_expr = clone_expression(sub_component_map[child])
```
jsiirola: Same as above (lines 36-37)

pyomo/core/plugins/transform/scaling.py
```diff
@@ -0,0 +1,269 @@
...
+                ## TODO: don't need the isinstance... just check in sub_component_map
+                if isinstance(child, _VarData):
+                    if child in sub_component_map:
+                        e._numerator[i] = clone_expression(sub_component_map[i])
+                else:
+                    stack.append(child)
+            for i, child in enumerate(e._denominator):
+                if isinstance(child, _VarData):
+                    if child in sub_component_map:
+                        e._denominator[i] = sub_component_map[i]
```
jsiirola: Same as above (lines 36-37)

pyomo/core/plugins/transform/scaling.py
```diff
@@ -0,0 +1,269 @@
...
+    while stack:
+        e = stack.pop()
+        e_type = type(e)
+        if e_type is expr_coopr3._ProductExpression:
+            # _ProductExpression is fundamentally different in _Coopr3
+            for i, child in enumerate(e._numerator):
+                ## TODO: don't need the isinstance... just check in sub_component_map
+                if isinstance(child, _VarData):
+                    if child in sub_component_map:
+                        e._numerator[i] = clone_expression(sub_component_map[i])
```
jsiirola: 1) Would `e._numerator[i] = sub_component_map.get(child, child)` be better than the `if ... then` clause?
2) I think it needs to be `sub_component_map[child]` not `[i]`
3) Why are you using clone_expression?  That shouldn't do anything for leaf nodes.

pyomo/core/plugins/transform/scaling.py
```diff
@@ -0,0 +1,269 @@
...
+                component_scaling_factor_map[c] = self._get_float_scaling_factor(model, c)
+
+        else:
+            raise ValueError("ScaleModel transformation: unknown scaling_method found"
+                             "-- supported values: 'user' ")
+
+        # scale the variable bounds and values and build the
+        # variable substitution map for scaling vars in constraints
+        variable_substitution_map = ComponentMap()
+        for variable in [var for var in model.component_objects(ctype=Var, descend_into=True)]:
```
jsiirola: This is weird syntax.  Why the intermediate list?  If you are worried about the tree changing while you iterate over the Vars, I would prefer:
```
var_objects = list( model.component_objects(Var, descend_into=True) )
for var in var_objects:
```

pyomo/core/plugins/transform/scaling.py
```diff
@@ -0,0 +1,269 @@
...
+                    if c._lower is not None:
+                        c._lower = c._lower * scaling_factor
+                    if c._upper is not None:
+                        c._upper = c._upper * scaling_factor
+
+                    if scaling_factor < 0:
+                        # swap inequalities
+                        temp = c._lower
+                        c._lower = c._upper
+                        c._upper = temp
```
jsiirola: The Pythonic Swap is:
```
c._lower, c._upper = c._upper, c._lower
```
(it is also faster than using the intermediate variable)

pyomo/core/plugins/transform/scaling.py
```diff
@@ -0,0 +1,269 @@
...
+                                 'have a component_scaling_factor_map. It is possible this method was called '
+                                 'using a model that was not scaled with the ScaleModel transformation')
+        component_scaling_factor_map = scaled_model.component_scaling_factor_map
+
+        # get the objective scaling factor
+        objectives = list()
+        for obj in scaled_model.component_data_objects(ctype=Objective, active=True, descend_into=True):
+            objectives.append(obj)
+        if len(objectives) != 1:
+            raise NotImplementedError('ScaleModel.propagate_solution requires a single active objective function, but %d objectives found.' % (len(objectives)))
```
jsiirola: This check / exception should be done in `_apply_to` and not here.

whart222:
I recommend that we accept this PR as-is, and then merge this change into the expr_dev branch where we rework this transformation.  I've already drafted a scaling transformation using a visitor pattern, and I think it'll be simpler to address some of the concerns mentioned by @jsiirola with that visitor pattern.

pyomo/core/plugins/transform/scaling.py
whart222: This will be replaced by the visitor logic.

pyomo/core/plugins/transform/scaling.py
whart222: This will be replaced by the visitor logic.

pyomo/core/plugins/transform/scaling.py
whart222: This will be replaced by the visitor logic.

pyomo/core/plugins/transform/scaling.py
whart222: @jsiirola I don't see why this exception should be moved.  This is done *after* the model is transformed.  This is about the solution transformation process.

jsiirola: My thought was that if the solution cannot be propagated back to the user (and the transformation will know this upfront), then the transformation should fail early, instead of waiting until after the solver spent all the time getting a solution only to tell the user "and I won't tell you what the solution is!"

carldlaird:
I suggest we hold this PR until after the new expressions are merged in. It will change with the new expression system. I can address all the comments along with that re-write.

PR #240, opened 2017-10-15
Use more efficient category checking in gams writer

daviskirk:
## Handle GAMS writer performance bottleneck

- In the gams interface, the model writing portion was taking far longer than the solution part.  The hotspot identified was a "search in list" operation for each variable to evaluate the variable category.  Iterating over the variable names AND categories allows this to become a simple comparison operation pretty much eliminating the time spend here.
- For us, this decreases model write time in our test cases from 400 seconds to about 45 seconds.
- No new tests since no functionality was added

p.s. This is my first contribution to pyomo, I couldn't find a "contributing" document or style guide so please let me know if there's anything I missed.

### DIFF
--- a/pyomo/repn/plugins/gams_writer.py
+++ b/pyomo/repn/plugins/gams_writer.py
@@ -319,42 +319,25 @@ class ProblemWriter_gams(AbstractProblemWriter):
         ))
 
         # Categorize the variables that we found
-        binary = []
-        ints = []
-        positive = []
-        reals = []
-        for var in var_list:
-            v = symbolMap.getObject(var)
-            if v.is_binary():
-                binary.append(var)
-            elif v.is_integer():
-                if (v.has_lb() and (value(v.lb) >= 0)) and \
-                   (v.has_ub() and (value(v.ub) <= 1)):
-                    binary.append(var)
-                else:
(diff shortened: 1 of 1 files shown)


pyomo/repn/plugins/gams_writer.py
```diff
@@ -508,6 +494,47 @@ def _write_model(self,
...
+            v = symbol_map.getObject(var)
+            if v.is_binary():
+                self.binary.append(var)
+            elif v.is_integer():
+                if (v.has_lb() and (value(v.lb) >= 0)) and \
+                   (v.has_ub() and (value(v.ub) <= 1)):
+                    self.binary.append(var)
+                else:
+                    self.ints.append(var)
+            elif value(v.lb) == 0:
```
qtothec: Greater than or equal to?

gseastream: Currently, we use the positive classification to set those variables' lower bound to 0. As in we do not check (on the new line 358) for a lower bound to write for positive variables. If the lower bound is not exactly 0 they are declared as free and have a lower bound checked for and written.

jsiirola:
This looks good to me.

PR #245, opened 2017-10-31
Logging overhaul and support for timing concrete models

jsiirola:
This PR has two main parts:
  - overhaul of the Pyomo logging system.  This creates a single centralized LogHandler at the `pyomo` namespace.  Filtering is all handles through the (sub)namespace log levels.  This greatly simplifies logging for (new) subpackages, although it also strongly motivates the elimination of the `pyomo` namespace package in the near future.
  - moving the construction timing reports from "`print`" statements in PyomoModel class and distributes them as log messages generated by the individual components.  This supports two key things:
    1) timing for sub-blocks
    2) timing for concrete models (that never call `create_instance()`


pyomo/core/base/block.py
```diff
@@ -1718,6 +1719,7 @@ def construct(self, data=None):
                           self.__class__.__name__, self.name, str(data) )
         if self._constructed:
             return
+        timer = ConstructionTimer(self)
```
ghackebeil: There is another return statement in this method (inside the if rule is None block) that needs a `timer.report()` just prior to it.

PR #246, opened 2017-11-03
Constraint Bound Tightening

sjkale:
Added Isomorphic Transformation that tightens the bounds on a constraint based on the variable bounds


pyomo/core/plugins/transform/bounds_tightener.py
```diff
@@ -0,0 +1,61 @@
...
+            UB = 0
+            if repn.constant:
+                UB = -1 * repn.constant
+            LB = UB
+            for i, coef in enumerate(repn.linear):
+                #TODO: handle unbounded vars
+                #TODO: ROunding issues
+
+                if coef >= 0:
+                    if repn.variables[i].ub and repn.variables[i].lb:
```
jsiirola: Why do you only update the UB/LB if the variable has both?  You should be able to tighten one side even if the other side is still unbounded.

pyomo/core/plugins/transform/bounds_tightener.py
```diff
@@ -0,0 +1,61 @@
...
+                    else:
+                        if not repn.variables[i].ub:
+                            LB = -float('Inf')
+                        if not repn.variables[i].lb:
+                            UB = float('Inf')
+
+            if(constr.has_ub() and value(constr.upper) > UB) or (not constr.has_ub()):
+                constr._upper = float(UB)
+            if(constr.has_lb() and value(constr.lower) < LB) or (not constr.has_lb()):
+                constr._lower = float(LB)
```
jsiirola: I think this will convert unbounded constraints to having bounds of `+/-Inf` in the case of constraints with unbounded variables.  This is probably not good...

pyomo/core/plugins/transform/bounds_tightener.py
```diff
@@ -0,0 +1,61 @@
...
+            if UB < LB:
+                print "Infeasible variable bounds"
```
jsiirola: This is not valid Python 3.x syntax.  You should emit info/warning/errors using the Python logging functionality and not embed `print` statements.

pyomo/core/plugins/transform/bounds_tightener.py
```diff
@@ -0,0 +1,61 @@
+from pyomo.core.base.constraint import Constraint
+from pyomo.core.kernel.numvalue import value
+from pyomo.core.plugins.transform.hierarchy import IsomorphicTransformation
+from pyomo.repn.canonical_repn import generate_canonical_repn
+from pyomo.util.plugin import alias
+
+import textwrap
+
+class BoundsTightener(IsomorphicTransformation):
```
jsiirola: We should also consider a more descriptive name: I expect us to eventually have a number of bounds tighteners (FBBT, OBBT, etc.)

qtothec:
@sjkale I don't think you meant to push those changes into this pull request. I would open up a separate one.

PR #248, opened 2017-11-06
PyMP

cristianallara:
Nested Benders Decomposition algorithm applied to small multi-period LP problem, and Nested Decomposition algorithm for IDAES capacity expansion model (multi-period MILP)


pyomo/PyMP/MPLP.py
```diff
@@ -0,0 +1,100 @@
+
+################################################################################
+##### Model taken from: Ind.Eng.Chem.Res (2003), 42, 3045, 3055 ################
```
qtothec: Slightly more reference data would be nice =)

PR #249, opened 2017-11-07
Update README.txt

XinNickyHuang:
Typo correction on line 11
Modify command for new version of Pyomo on line 9
But I am also wondering, why in models folder there is no "ScenarioStructure.py"??
Thanks for answering
@whart222

### DIFF
--- a/examples/pysp/farmer/README.txt
+++ b/examples/pysp/farmer/README.txt
@@ -6,9 +6,9 @@ This problem serves as a simple, quick smoke-screen test for a functioning stoch
 
 The deterministic, expected-value model can be solved simply by the following command-line: 
 
-pyomo models/ReferenceModel.py scenariodata/AverageScenario.dat
+pyomo solve --solver=clpex models/ReferenceModel.py scenariodata/AverageScenario.dat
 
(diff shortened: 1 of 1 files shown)


jsiirola:
The pyomo "model" that `ScenarioStructure.dat` populates to define the scenario tree is built into the Pyomo distribution, so the user does not have to specify it.

XinNickyHuang:
Thanks for answering @jsiirola , I see.

If I remember right, the orignal paper of Hart et al. (2012) on PySP emphasized the file ScenarioStructure.py at that time. But now after I check Pyomo Online Documentation 5.1.1, I find that the process of using latest version of Pyomo and PySP simplifies to merely create ReferenceModel.py with specification of ScenarioStructure.dat and scenario data.

Following the above idea, I successfully run the farmer example of PySP by CPLEX 12.7, although the answers is slight different from readme.txt in that instance. And I also find that when using command runph, we need to specify rho=1 (or other values) explicitly nowadays, which is quite different from old version of PySP in Hart et al. (2012), and I modify the readme.txt again to pull request in the repository.

Thanks for your help and I can continue my own research on portfolio selection after confirming right installation and successfully trying examples here.

PR #250, opened 2017-11-07
Update README.txt

XinNickyHuang:
Correct previous typo and update new requirement on rho when using runph command

### DIFF
--- a/examples/pysp/farmer/README.txt
+++ b/examples/pysp/farmer/README.txt
@@ -20,7 +20,7 @@ The optimal solution for the stochastic model plants 170 acres of wheat, 80 acre
 
 The stochasic farmers problem can be solved via the following command-line:
 
-runph --model-directory=models --instance-directory=nodedata -r=1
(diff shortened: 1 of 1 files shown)


examples/pysp/farmer/README.txt
```diff
@@ -20,7 +20,7 @@ The optimal solution for the stochastic model plants 170 acres of wheat, 80 acre
 
 The stochasic farmers problem can be solved via the following command-line:
 
-runph --model-directory=models --instance-directory=nodedata 
+runph --model-directory=models --instance-directory=nodedata -r=1
```
jsiirola: I would prefer `--default-rho`  instead of `-r` for instructional reasons.

XinNickyHuang: Sure, I agree with you! Thanks for correction. We should notcie that --default-rho setting is required nowadays in the runph command, as claimed in the "runph --help". Thanks again

PR #253, opened 2017-11-13
[HACK] Fix to allow GAMS writer to handle models with deactivated connectors

qtothec:


pyomo/repn/plugins/gams_writer.py
```diff
@@ -243,7 +243,7 @@ def _write_model(self,
         # how to deal with, plus Suffix if solving
         valid_ctypes = set([
             Block, Constraint, Expression, Objective, Param,
-            Set, RangeSet, Var, Suffix ])
+            Set, RangeSet, Var, Suffix, Connector ])
```
gseastream: Asking as someone who has never used Connectors before, should we check to make sure that any Connectors in the model are deactivated? So as to not appear as though the writer supports active Connectors?

qtothec: Yes, that seems like a good idea.

qtothec: I'm a bit confused now, because I thought that all of the connectors should have been disabled in the model where I ran into this problem. I'll check again tomorrow.

gseastream: I am also confused because line 247 should only be collecting active ctypes in the model, so the writer supposedly already supports any deactivated ctypes

jsiirola: The correct solution (in the New World Order) is for Transformations to deactivate non-standard components after they have been transformed "out" of the model.  So, I believe that "option 1" is the correct way to go.

qtothec: When I get a chance, I'll poke at the ConnectorExpander.

whart222: I agree with John.  So we can close this PR?

qtothec: Sounds reasonable to me.

jsiirola: I am officially changing my mind on this PR.  When I wrote the above, I forgot that Connectors are *not* `ActiveComponent`s, and therefore cannot be deactivated.  I do not think that Connectors should be made Active, either (for the same reason that neither Var or Param are Active).  I believe that while deactivating the component is the correct solution when the component is an `ActiveComponent`, in this case, the correct solution should be for the writer to just ignore the Connectors (see PR #280, design discussion #309).

PR #260, opened 2017-11-21
Add error message to model.write if guess_format returns None

### LINKED ISSUE #222: Unclear error from model.write with invalid file extension
The error given by `model.write` when the user provides an invalid file extension with the intention of relying on `guess_format` is unclear. See the following:

```
Python 2.7.12 (default, Nov 19 2016, 06:48:10) 
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> m = ConcreteModel()
>>> m.write('file.something')
  File "/home/gseastre/pyomo/src/pyomo/pyomo/core/base/block.py", line 1647, in write
    io_options)
TypeError: __call__() takes at most 3 arguments (5 given)
```

This appears to be due to the fact that `guess_format` returns `None` when the file extension does not match any know problem format. Thus, `model.write` is calling the return of `WriterFactory(None)`, which is a generic `PluginFactoryFunctor object`.

My feeling is that the error should indicate that no valid format could be determined from `guess_format`. I discovered this issue when I tried to input a .xlsx file without realizing I was not on my excel branch so .xlsx was not a supported format.

gseastream:
Resolves #222

### DIFF
--- a/pyomo/core/base/block.py
+++ b/pyomo/core/base/block.py
@@ -1633,6 +1633,10 @@ Components must now specify their rules explicitly using 'rule=' keywords.""" %
             format = ProblemFormat.cpxlp
         if (filename is not None) and (format is None):
             format = guess_format(filename)
+            if format is None:
+                raise ValueError(
+                    "guess_format could not find a valid format from "
+                    "filename '%s'" % filename)
         problem_writer = WriterFactory(format)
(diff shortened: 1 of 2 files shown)


jsiirola:
I ended up cleaning up the error message and adding a test.  This is good to merge.

PR #262, opened 2017-11-21
Persistent interfaces

michaelbynum:
This pull request refactors the direct interfaces to cplex and gurobi. It also implements persistent interfaces to allow efficient resolves with the addition and removal of constraints, variables, blocks, etc. I have also added documentation for these interfaces to the online docs. The cplex_persistent interface still needs a bit more profiling, but the performance is reasonable.


jwatsonnm:
I have tested the persistent interfaces for both gurobi and cplex rather extensively in the context of runph, and they are working quite well. And, more importantly, correctly. I've also tested the --solver-io option variants with both runph and the pyomo command line, to ensure "=python" picks up the new direct interfaces. I have also looked at the design and code, and all looks very clean and straightforward to me. It's time to get this into broader circulation, so I will merge once I submit this comment.

PR #263, opened 2017-11-22
Fix GAMS writer replace_power to support function-wrapped powers

### LINKED ISSUE #244: GAMS writer fails for power nested in a log
The functionality to selectively replace `**` with `power()` fails in an edge case when `**` is nested within a `log` function, e.g. `log(x ** 2)`. In `replace_power()`, the `assert` that `arg` is bound by `(` and `)` fails because the arg is bound by `log(` and `)`. This is possibly also true with other similar functions, e.g. `exp()`.

@gseastream

gseastream:
Fixes #244

### DIFF
--- a/pyomo/repn/plugins/gams_writer.py
+++ b/pyomo/repn/plugins/gams_writer.py
@@ -345,7 +345,7 @@ class ProblemWriter_gams(AbstractProblemWriter):
                 # Investigate power functions for an integer exponent, in which
                 # case replace with power(x, int) function to improve domain
                 # issues.
-                line = replace_power(line) + ';'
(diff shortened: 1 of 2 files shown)


qtothec:
May want to add a quick test in there for this case.

whart222:
I'm closing this pull request after confirming that the changes in the expr_dev branch account for the error considered here.

whart222:
John wanted this PR re-opened to support an upcoming release on the master branch.

jsiirola:
(Spoke with @carldlaird and he was also OK merging this PR.  Going ahead and merging it)

PR #265, opened 2017-11-27
Changes to correct the settimelimit option

manish-byteprophecy:
The timelimit is set to 1 sec lower than provided option to ensure CBC process gets time to exit properly before the pyutilib.subprocess.processmngr checks the exit code. In my test runs the process manager checked the exit status at T+0.02 seconds and CBC process typically exited after T+0.1 seconds.
Also changed the CBC code to use the wallclock seconds from the default of CPU seconds as the process manager is also checking the same.

### DIFF
--- a/pyomo/opt/solver/shellcmd.py
+++ b/pyomo/opt/solver/shellcmd.py
@@ -300,7 +300,7 @@ class SystemCallSolver(OptSolver):
             [rc, log] = run(
                 command.cmd,
                 stdin = _input,
-                timelimit = self._timelimit,
+                timelimit = self._timelimit + max(1, 0.01*self._timelimit),
                 env   = command.env,
                 tee   = self._tee
              )
--- a/pyomo/opt/solver/shellcmd.py
+++ b/pyomo/opt/solver/shellcmd.py
@@ -300,7 +300,7 @@ class SystemCallSolver(OptSolver):
             [rc, log] = run(
                 command.cmd,
                 stdin = _input,
-                timelimit = self._timelimit + max(1, 0.01*self._timelimit),
+                timelimit = self._timelimit if self._timelimit is None else self._timelimit + max(1, 0.01*self._timelimit),
                 env   = command.env,
                 tee   = self._tee
              )
--- a/pyomo/solvers/plugins/solvers/CBCplugin.py
+++ b/pyomo/solvers/plugins/solvers/CBCplugin.py
@@ -266,7 +266,6 @@ class CBCSHELL(SystemCallSolver):
             if self._timelimit is not None and self._timelimit > 0.0:
                 cmd.extend(['-sec', str(self._timelimit)])
                 cmd.extend(['-timeMode', "elapsed"])
-                self._timelimit+=1
             if "debug" in self.options:
                 cmd.extend(["-log","5"])
             for key, val in _check_and_escape_options(self.options):
@@ -280,7 +279,6 @@ class CBCSHELL(SystemCallSolver):
             if self._timelimit is not None and self._timelimit > 0.0:
                 cmd.extend(['-sec', str(self._timelimit)])
                 cmd.extend(['-timeMode', "elapsed"])
-                self._timelimit+=1
             if "debug" in self.options:
                 cmd.extend(["-log","5"])
             # these must go after options that take a value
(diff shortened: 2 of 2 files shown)


manish-byteprophecy:
I agree with your suggestion @jsiirola . I have made the required change and committed the same.
After setting the -secs option for CBC I have increased the self._timelimit by 1 second.

carldlaird:
A couple of points.

1. This is not only an issue with CBC, but likely will all of our solvers (implemented in the base class?).
2. I think we should set the subprocess timeout to be the minimum of 1 sec or x% of the solver timeout (maybe 5%)

Thoughts?

jsiirola:
I agree with @carldlaird's points (although 2 should be `max()` not `min()`).  I think the correct place to make the adjustment to the timilimit is in the `_execute_command` method in `pyomo/opt/solver/shellcmd.py`.

My real question for @whart222 and @jwatsonnm is what is the expected behavior for a user when they set a timeout?  Is that the time you expect the solver to quit, or the time you expect Pyomo's "solve" to return?  I believe that the correct answer is the former (as is currently implemented in the PR)

jwatsonnm:
My (strong) feeling is that when a user sets a timeout on a solver, they believe that the time limit of the solver process - in the loose sense, but whenever and however the real solver starts - is being set. For example, if they set the time limit to 60 seconds for really difficult MIPs, they would expect the results object to report 60 seconds (plus or minus epsilon). Any extra time that is required due to waiting for processes to die and such is something that should be IMHO handled / augmented by the solver plugin hierarchy. The latter can be variable, might change based on OS, and who knows what else. But independent of pyomo's solver plugin implementation, I still would want to see my 60 seconds.

jsiirola:
@manish-byteprophecy: I agree: wall time is the metric we should be discussing, and I fully agree with your change of the CBC plugin to switch to wall time.

@jwatsonnm: to confirm, you believe that the timelimit should be the total time between the call to `solver.solve(model)` and when that method returns?  I think I disagree: we have had cases in the past where users inquired about differences between the solve time reported by the solver and the timelimit, so I think that the intent of the "timelimit" option is to regulate how much time the solver has to run (for the purposes of this discussion, that means passing the timelimit option into the solver unaltered).

I think the question about what to set the subprocess timeout at is a little more complicated, but boils down to what is the *intent* of the timeout for `pyutilib.subprocess.run`?  I believe that that timeout is there as a "safety" switch to terminate 'hung' processes, and in the normal course of business should never actually be triggered.  This suggests to me that we should be OK with relaxing the subprocess timeout (because under normal circumstances the solver will honor the timelimit and terminate cleanly).  There is also a penalty for not allowing the subprocess to terminate cleanly: it won't be given an opportunity to report any "final/last solution".  In addition, implicit in @carldlaird's comment is that many solvers only check the time limit periodically (once per node, etc), so the *actual* time that the solver exits can be `timelimit + time_for_single_iteration`.  For large / hard problems, `time_for_single_iteration` can be significant.  This is which I think we should make the subprocess timeout be `timelimit + max(1, timelimit*0.05)`.

jwatsonnm:
No - I believe the time limit that a user specifies is the time that is ultimately allocated to the *solver* itself. Not the time allocated to the solve() method invocation. You can control the former to high tolerance, but not the latter. This is the present interpretation for all of the gurobi and cplex plugins (for example), as the option is handled by the actual solver - not the plugin.

carldlaird:
I agree with JP and John (who I think agree with each other). The solver time limit is for the solver. Our subprocess time limit should have a buffer to allow the solver (some, but not forever) to cleanup.

manish-byteprophecy:
Made the changes as suggested by @jsiirola.

Used 1% buffer instead of 5%.

manish-byteprophecy:
I am not sure why the tests are failing. In my environment "Pyomo 5.3 (CPython 3.6.1 on Linux 3.10.0-514.10.2.el7.x86_64)" the solver is running properly without any issues, but the online build log shows an error.
timelimit = self._timelimit + max(1, 0.01*self._timelimit),
TypeError: unsupported operand type(s) for *: 'float' and 'NoneType'
Since 0.01 is a float the self._timelimit is potentially being detected as NoneType. Can someone tell me what change I might be needed to make to ensure compatibility with all python versions.

qtothec:
If no time limit is set, it may be that the expected behavior is for `self.timelimit` to be zero.

qtothec:
Sorry, I meant `None`

PR #272, opened 2017-12-05
Merge Expression Branch

whart222:
This PR can be referenced to discuss issues that need to be resolved to merge the expressions branch:  expr_dev.


pyomo/solvers/plugins/solvers/gurobi_direct.py
```diff
@@ -131,43 +131,29 @@ def _apply_solver(self):
...
 
-        degree = canonical_degree(repn)
+        degree = repn.polynomial_degree()
         if (degree is None) or (degree > max_degree):
             raise DegreeError('GurobiDirect does not support expressions of degree {0}.'.format(degree))
 
-        if isinstance(repn, LinearCanonicalRepn):
-            if (repn.linear is not None) and (len(repn.linear) > 0):
-                list(map(referenced_vars.add, repn.variables))
-                new_expr = self._gurobipy.LinExpr(repn.linear, [self._pyomo_var_to_solver_var_map[i] for i in repn.variables])
```
michaelbynum: If we know the expression is linear, this is a much faster way to create the gurobi expression. I may take a stab at updating this method.

pyomo/solvers/plugins/solvers/cplex_direct.py
```diff
@@ -160,7 +160,7 @@ def _process_stream(arg):
     def _get_expr_from_pyomo_repn(self, repn, max_degree=2):
         referenced_vars = ComponentSet()
 
-        degree = canonical_degree(repn)
+        degree = repn.polynomial_degree()
         if (degree is None) or (degree > max_degree):
```
michaelbynum: This method needs updated (similar to gurobi_direct.py). I may take a stab at this too.

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3356 @@
...
+    PRECEDENCE = 3.5
+
+    def nargs(self):
+        return 1
+
+    def _precedence(self):
+        return ReciprocalExpression.PRECEDENCE
+
+    def _compute_polynomial_degree(self, result):
+        if result[0] is 0:
```
ghackebeil: `is 0` to `== 0`

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3356 @@
...
+            return _l < _b and _b <= _r
+        else:
+            return _l < _b and _b < _r
+
+    def _to_string(self, values, verbose, smap, compute_values):
+        return "{0}  {1}  {2}  {3}  {4}".format(values[0], '<' if self._strict[0] else '<=', values[1], '<' if self._strict[1] else '<=', values[2])
+        
+    def is_constant(self):
+        return (self._args_[0].__class__ in native_numeric_types or self._args_[0].is_constant()) and \
+
```
ghackebeil: `is 0` to `== 0`

pyomo/repn/repn_ideas.py
```diff
@@ -0,0 +1,860 @@
...
+                            nonl.append(res[None])
+                if len(nonl) > 0:
+                    nonl = Sum(x for x in nonl)
+                    if not (nonl.__class__ in native_numeric_types and isclose(nonl,0)):
+                        ans[None] = nonl
+                        linear = False
+            # Add constant terms
+            cons = 0
+            cons = 0 + sum(res[0] for res in _result if 0 in res)
+            if not cons is 0:
```
ghackebeil: `is 0` to `== 0`

pyomo/repn/repn_ideas.py
```diff
@@ -0,0 +1,860 @@
...
+    keys = list(ans.linear.keys())
+    repn.linear_vars = tuple(idMap[key] for key in keys)
+    repn.linear_coefs = tuple(ans.linear[key] for key in keys)
+
+    if quadratic:
+        keys = list(ans.quadratic.keys())
+        repn.quadratic_vars = tuple((idMap[key[0]],idMap[key[1]]) for key in keys)
+        repn.quadratic_coefs = tuple(ans.quadratic[key] for key in keys)
+
+    if not ans.nonl is 0:
```
ghackebeil: `is 0` to `== 0`

pyomo/core/expr/expr_pyomo5.py
whart222: Done

pyomo/solvers/plugins/solvers/gurobi_direct.py
whart222: I took a stab at making this change.  Can you double-check this for me?

pyomo/solvers/plugins/solvers/cplex_direct.py
whart222: I'm not sure what you mean.  Can you suggest a change?

pyomo/solvers/plugins/solvers/gurobi_direct.py
```diff
@@ -167,18 +167,23 @@ def _get_expr_from_pyomo_repn(self, repn, max_degree=2):
         if (degree is None) or (degree > max_degree):
             raise DegreeError('GurobiDirect does not support expressions of degree {0}.'.format(degree))
 
-        new_expr = repn.constant
-
-        for i,v in enumerate(repn.linear_vars):
-            new_expr += repn.linear_coefs[i] * self._pyomo_var_to_solver_var_map[v]
-            referenced_vars.add(v)
+        if len(repn.linear_vars) > 0:
+            list(map(referenced_vars.add, repn.linear_vars))
```
ghackebeil: `referenced_vars.update(repn.linear_vars)` seems like it would be equivalent, faster, and a bit easier to read

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3413 @@
...
+        elif not self._strict[1]:
+            return _l < _b and _b <= _r
+        else:
+            return _l < _b and _b < _r
+
+    def _to_string(self, values, verbose, smap, compute_values):
+        return "{0}  {1}  {2}  {3}  {4}".format(values[0], '<' if self._strict[0] else '<=', values[1], '<' if self._strict[1] else '<=', values[2])
+
+    def is_constant(self):
+        return (self._args_[0].__class__ in native_numeric_types o
```
ghackebeil: Can we eliminate this implicit conversion to an equality expression? It doesn't add any real functionality, it's confusing, and it violates the documentation for this function.

ghackebeil:
I was looking over the `generate_standard_repn` code today, and I noticed the `isclose` function is being used in a number of places. Implicitly, this means we have built a form of scaling into canonical representation generation, and there is currently no way for a user to override these settings. In general, I think it is a bad idea for us to have tolerances built into any of the solver interfaces (unless we expose them via a keyword or something). Anyone have thoughts on this?

jwatsonnm:
I concur with Gabe - the solver interfaces should pass through whatever they are given. Even if it is ridiculously small. If one wants to nix "too-small" values, then one should write a transform.

jsiirola:
Agreed.  Neither the expression system nor the "compiler" (generate_standard_repn) should have any tolerances.

ghackebeil:
Failing "AMPL comparison" tests related to the comments above are found in:
 1. https://github.com/Pyomo/pyomo-model-libraries/blob/expr_dev/cute/testCUTE_asl.py
 2. https://github.com/Pyomo/pyomo/blob/expr_dev/pyomo/repn/tests/ampl/test_ampl_comparison.py

These tests use the [gjh_asl_json](https://github.com/ghackebeil/gjh_asl_json) executable.

ghackebeil:
Here is a minimal example that shows how the `not expr.potentially_variable()` checks in `generate_standard_repn` fail to move a fixed piece of a nonlinear expression into the constant:
```python
>>> from pyomo.repn.standard_repn import generate_standard_repn
>>> import pyomo.kernel as pmo
>>> m = pmo.block()
>>> m.x = pmo.variable()
>>> m.y = pmo.variable()
>>> m.y.fix(1.0)
>>> print(generate_standard_repn(m.x - (1-m.y**2), quadratic=False))

constant:       -1
linear vars:    ['x']
linear var ids: [4551289560]
linear coef:    [1]
quadratic vars:    []
quadratic var ids: []
quadratic coef:    []
nonlinear expr:
y**2
nonlinear vars: []
```

whart222:
@ghackebeil The example you showed above has been resolved.  This was not due to differences between is_fixed() and is_potentially_variable().  Rather, there was a bug in the formation of standard repns when exponential terms were used.

As we discussed, is_potentially_variable() is preferred because it is much much faster in Pyomo5.

pyomo/core/util.py
```diff
@@ -0,0 +1,278 @@
...
+            # term is not a native numeric type.  Otherwise, large summation
+            # objects are created for the constant term.
+            #
+            if linear:
+                nvar=0
+                for term in terms:
+                    c,v = term
+                    if not v is None:
+                        nvar += 1
+                    elif not c.__class__ in native_numeric_types:
```
ghackebeil: Should this check be changed to `elif c.is_potentially_variable():`? The example I'm using is
```python
import pyomo.kernel as pmo
from pyomo.environ import quicksum

m = pmo.block()
m.d = pmo.data_expression(1.0)

print(quicksum(m.d for i in range(10)).__class__.__name__)
```
Which seems to generate the general "nonlinear" `ViewSumExpression`.

ghackebeil: Actually, I suppose it needs to be `and`-ed with the current check so that simple numeric types don't result in an error:

`(c.__class__ not in native_numeric_types) and c.is_potentially_variable():`

examples/doc/pyomobook/pyomo-components-ch/param_misc.txt
```diff
@@ -43,6 +43,6 @@
 1 Objective Declarations
     o : Size=1, Index=None, Active=True
         Key  : Active : Sense    : Expression
-        None :   True : minimize : p[1] * x[1] + p[2] * x[2] + p[3] * x[3]
+        None :   True : minimize : x[1] + p[2]*x[2] + p[3]*x[3]
```
jsiirola: This is a bug (a bad baseline update)!  p[1] is a *mutable* Param and should not be eliminated from the expression.

pyomo/contrib/preprocessing/plugins/remove_zero_terms.py
```diff
@@ -36,18 +39,18 @@ def _apply_to(self, model):
...
                 continue  # we currently only process linear constraints
-            repn = generate_canonical_repn(constr.body)
+            repn = generate_standard_repn(constr.body)
 
             # get the index of all nonzero coefficient variables
             nonzero_vars_indx = [
-                i for i, _ in enumerate(repn.variables)
-                if not repn.linear[i] == 0
+                i for i, _ in enumerate(repn.linear_vars)
+                if not isclose(repn.linear_coefs[i],0)
```
jsiirola: This is a change in the behavior of the transformation and shouldn't be part of expr_dev.  If you think the transformation should use `isclose()` then there should be a ticket and the contrib package owner should make that call.

pyomo/core/base/component.py
```diff
@@ -366,6 +366,14 @@ def reconstruct(self, data=None):
         self._constructed = False
         self.construct(data=data)
 
+    def is_named_expression_type(self):
```
jsiirola: Why do *components* need to duck type the expression API?

pyomo/core/base/constraint.py
```diff
@@ -469,14 +466,14 @@ def set_value(self, expr):
                         "Constraint '%s' does not have a proper "
                         "value. Found '%s'\nExpecting a tuple or "
                         "equation. Examples:"
-                        "\n   summation(model.costs) == model.income"
+                        "\n   sum_product(model.costs) == model.income"
```
jsiirola: This is not a good example, if for no other reason than there is no product being formed.  I would rather examples like:
```
sum_product(model.unit_price, model.sales) == model.income
sum(model.costs[i] for i in model.costs) == model.expenses  # or quicksum
```

pyomo/core/base/constraint.py
```diff
@@ -469,14 +466,14 @@ def set_value(self, expr):
...
                         "\n   (0, model.price[item], 50)"
                         % (self.name, str(expr)))
             except AttributeError:
                 msg = ("Constraint '%s' does not have a proper "
                        "value. Found '%s'\nExpecting a tuple or "
                        "equation. Examples:"
-                       "\n   summation(model.costs) == model.income"
+                       "\n   sum_product(model.costs) == model.income"
```
jsiirola: Same as above.

pyomo/core/base/objective.py
```diff
@@ -547,7 +545,7 @@ def set_value(self, expr):
         if self._check_skip_add(None, expr) is None:
             del self[None]
             return None
-        return _GeneralObjectiveData.set_value(self, expr)
+        return _GeneralObjectiveData.set_value(self, EXPR.compress_expression(expr))
```
jsiirola: What is compress_expression and why do we need to call it?

pyomo/core/expr/current.py
```diff
@@ -0,0 +1,128 @@
...
+import math
+import copy
+
+#
+# Data and methods that are exposed when importing pyomo.core.expr
+#
+__public__ = ['log', 'log10', 'sin', 'cos', 'tan', 'cosh', 'sinh', 'tanh',
+           'asin', 'acos', 'atan', 'exp', 'sqrt', 'asinh', 'acosh',
+           'atanh', 'ceil', 'floor']
+
```
jsiirola: Why are we inventing a nonstandard `__public__` module attribute?  This seems undesirable, as "`__`" methods are generally reserved for the interpreter...

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3434 @@
...
+'_MutableSumExpression',    # This should not be referenced, except perhaps while testing code
+'_MutableLinearExpression',     # This should not be referenced, except perhaps while testing code
+'_decompose_linear_terms',      # This should not be referenced, except perhaps while testing code
+'_chainedInequality',           # This should not be referenced, except perhaps while testing code
+'_using_chained_inequality',           # This should not be referenced, except perhaps while testing code
+'_generate_sum_expression',                 # Only used within pyomo.core.expr
+'_generate_mul_expression',                 # Only used within pyomo.core.expr
+'_generate_other_expression',               # Only used within pyomo.core.expr
+'_generate_intrinsic_function_expression',  # Only used within pyomo.core.expr
+'_generate_relational_expression',          # Only used within pyomo.core.expr
```
jsiirola: Why are "`_`" methods in `__all__`?

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3434 @@
...
+from pyomo.util.deprecation import deprecation_warning
+from pyomo.core.expr.symbol_map import SymbolMap
+from pyomo.core.expr.numvalue import \
+    (NumericValue,
+     NumericConstant,
+     native_types,
+     nonpyomo_leaf_types,
+     native_numeric_types,
+     as_numeric,
+     value)
```
jsiirola: Style:
```
from pyomo.core.expr.numvalue import (
    NumericValue,
    NumericConstant,
    native_types,
    nonpyomo_leaf_types,
    native_numeric_types,
    as_numeric,
    value
)
```

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3434 @@
...
+     _pow, _neg, _abs, _inplace,
+     _unary, _radd, _rsub, _rmul,
+     _rdiv, _rpow, _iadd, _isub,
+     _imul, _idiv, _ipow, _lt, _le,
+     _eq)
+from pyomo.core.expr import expr_common as common
+
+
+if _using_chained_inequality:               #pragma: no cover
+    class _chainedInequality(object):
```
jsiirola: Why is this class definition in an if block?  That seems wrong.

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3434 @@
...
+    from pyomo.core.base.param import _ParamData, SimpleParam
+    from pyomo.core.base.template_expr import TemplateExpressionError
+
+
+def compress_expression(expr):
+    """
+    Deprecated function that was used to compress deep Pyomo5
+    expression trees.
+    """
+    return expr
```
jsiirola: If this is deprecated, then it should be removed before the merge.

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3434 @@
...
+        return self.e
+
+    def __exit__(self, *args):
+        if self.e.__class__ == _MutableSumExpression:
+            self.e.__class__ = SumExpression
+
+#: A context manager object for nonlinear expressions.
+#: This is an instance of the :class:`mutable_sum_contex <pyomo.core.expr.current.mutable_sum_context>` context manager.
+#: Different instances of this context manager are not necessary, unless these context managers are nested.
+nonlinear_expression = mutable_sum_context()
```
jsiirola: These should not be global, as they will cause problems with nested usage (like nested quicksum)

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3434 @@
...
+        class to transform the context into a nonmutable
+        form.
+        """
+        if self.e.__class__ == _MutableLinearExpression:
+            self.e.__class__ = LinearExpression
+
+#: A context manager object for linear expressions.
+#: This is an instance of the :class:`mutable_linear_contex <pyomo.core.expr.current.mutable_lienar_context>` context manager.
+#: Different instances of this context manager are not necessary unless these context managers are nested.
+linear_expression = mutable_linear_context()
```
jsiirola: Same.  This should not be a global object.

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3434 @@
...
+
+
+#-------------------------------------------------------
+#
+# Visitor Logic
+#
+#-------------------------------------------------------
+
+class SimpleExpressionVisitor(object):
```
jsiirola: This file is getting unwieldy.  Should we move the visitor logic into a separate `visitor.py` module?

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3434 @@
...
+#
+# Functions used to process expression trees
+#
+#-------------------------------------------------------
+
+# =====================================================
+#  clone_expression
+# =====================================================
+
+class _CloneVisitor(ExpressionValueVisitor):
```
jsiirola: I do not like using a visitor to implement cloning.  If the point is to avoid deepcopy it won't work, as deepcopying a model (or block) will end up copying (cloning) the expressions in the model using deepcopy and not this method.  That leads to having two ways of duplicating expressions depending on the execution path -- and that strikes me as asking for trouble.

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3434 @@
...
+
+    def visit(self, node):
+        if node.__class__ in self.types:
+            if id(node) in self.seen:
+                return
+            self.seen.add(id(node))
+            return node
+
+
+def identify_components(expr, component_types):
```
jsiirola: Is this used by anything?  This seems bad to identify components using `__class__` and not ctype.  Either it should use ctype or not be called "identify_components".

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3434 @@
...
+    visitor = _MutableParamVisitor()
+    for v in visitor.xbfs_yield_leaves(expr):
+        yield v
+
+
+# =====================================================
+#  _polynomial_degree
+# =====================================================
+
+class _PolyDegreeVisitor(ExpressionValueVisitor):
```
jsiirola: For consistency, this should be `_PolynomialDegreeVisitor`

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3434 @@
...
+        tmp = []
+        for i,val in enumerate(values):
+            arg = node._args_[i]
+
+            if arg is None:
+                tmp.append('Undefined')                 # TODO: coverage
+            elif arg.__class__ in native_numeric_types:
+                tmp.append(val)
+            elif arg.__class__ in nonpyomo_leaf_types:
+                tmp.append("'{0}'".format(val))
```
jsiirola: Why use `.format()`?  It is measurably slower than `"'%s'" % (val,)`

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3434 @@
...
+            The i-th child node.
+        """
+        if i >= self.nargs():
+            raise KeyError("Invalid index for expression argument: %d" % i)
+        if i < 0:
+            return self._args_[self.nargs()+i]
+        return self._args_[i]
+
+    @property
+    def args(self):
```
jsiirola: Why is this defined this way in the base class?  Since Sum is the only thing where nargs and len(_args_) aren't the same, you could make this more efficient by making this implementation just iterate over _args_ and move the extra function calls and complexity to just Sum.

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3434 @@
...
+        Args:
+            memo (dict): a dictionary that maps object ids to clone
+                objects generated earlier during the cloning process.
+
+        Returns:
+            A new expression tree.
+        """
+        return clone_expression(self, memo=memo, clone_leaves=True)
+
+    def construct_node(self, args, memo):
```
jsiirola: Why do we need this?  Is this only for replicating deepcopy?

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3434 @@
...
+        within this expression.  This method returns :const:`True`
+        even if there the variables in the expression are currently
+        fixed.
+
+        Returns:
+            A boolean.  Defaults to :const:`True` for expressions.
+        """
+        return True
+
+    def is_named_expression_type(self):
```
jsiirola: This seems like a *very* specific method.  Why do we need it?  In particular, since Expressions are components, could/should we not use ctype?

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3434 @@
...
+            while tmp[i] == ' ':
+                i += 1
+            return tmp[i:]
+        return "- "+tmp
+
+    def _apply_operation(self, result):
+        return -result[0]
+
+
+class NPV_NegationExpression(NegationExpression):
```
jsiirola: I still don't understand why we have Net Present Value in our expression system.  We need a better name.

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3434 @@
...
+            if not self._strict[1]:
+                return _l <= _b and _b <= _r
+            else:
+                return _l <= _b and _b < _r
+        elif not self._strict[1]:
+            return _l < _b and _b <= _r
+        else:
+            return _l < _b and _b < _r
+
+    def _to_string(sel
```
jsiirola: I disagree with this note.  There is no reason for this to be any different from any other Unary Function

jsiirola: Is this used anywhere?  I don't see any references outside the tests for this function.

examples/dae/disease_DAE.py
```diff
@@ -227,7 +227,7 @@ def _scaled_beta(model, i):
 model.con_city_varying_beta = Constraint(model.S_BETA, rule=_scaled_beta)
 
 def _mean_patt(model):
-    return (1.0, summation(model.beta_patt)/len(model.S_BETA))
+    return (1.0, sum_product(model.beta_patt)/len(model.S_BETA))
```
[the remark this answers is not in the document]
blnicho: This model is old and appeared to use `summation` instead of `sum` for no good reason.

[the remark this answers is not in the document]
jsiirola: FWIW, `sum()` won't work, as it will sum over the indices of `model.beta_patt` and not the values.  You would need `sum(model.beta_patt[i] for i in model.beta_patt)`

ghackebeil:
@whart222 @jwatsonnm: I believe I've knocked out the remaining glaring PySP issues on this branch (with the exception of the issue in my next comment).

ghackebeil:
It's becoming more clear that adding the in-place operators to the named expression classes blurs the line between their _usage_ and their _modification_. In particular, the order in which I sum over a list such as `[1.0, 0.0, <named_expression>]` in a for-loop, should not effect whether that loop modifies the named expression or creates a new expression. I think this change results in too many subtle differences that could changes the semantic meaning of existing code.

I've confirmed that removing them fixes all of the examples I can come up with. No changes are required for how expressions are simplified.

jsiirola:
@ghackebeil, @whart222: FWIW, we have had the discussion around in-place operations on Expression objects before, and reached the conclusion that Expression objects should NOT support in-place operators.  (I had implemented them before and had been forced to remove them because of the same issues Gabe raises here).

ghackebeil:
@whart222, @jsiirola: I've pushed a single commit that reverts the change to named expressions in AML and Kernel, and adds tests similar to my example above. It can easily be reverted if we change our minds, but hopefully this helps clears up time to discuss other issues tomorrow.

pyomo/core/base/component.py
```diff
@@ -732,17 +743,22 @@ def __str__(self):
...
+            except:
+                pass
+        if smap:
+            return smap.getSymbol(self, labeler)
         if labeler is not None:
-            ostream.write(labeler(self))
+            return labeler(self)
         else:
-            ostream.write(self.__str__())
+            return self.__str__()
```
jsiirola: This change in the API needs to be prominently documented.  I agree with it, but it is a change.

pyomo/core/base/constraint.py
```diff
@@ -498,111 +495,86 @@ def set_value(self, expr):
         # user did ( var < 1 > 0 ) (which also results in a non-None
         # chainedInequality value)
         #
-        if EXPR.generate_relational_expression.chainedInequality is not None:
-            raise TypeError(EXPR.chainedInequalityErrorMessage())
+        if EXPR._using_chained_inequality and EXPR._chainedInequality.prev is not None:
+            raise TypeError(EXPR._chainedInequality.error_message())
```
jsiirola: Why the first half of the test?  Is this because you only conditionally include the class definition?  I still think that is not a good style.

pyomo/core/base/constraint.py
```diff
@@ -498,111 +495,86 @@ def set_value(self, expr):
...
-                            self._upper = ZeroConstant
-                        #self._body  = EXPR.generate_expression_bypassCloneCheck(
-                        #    _sub, _args[0], _args[1])
+                    self._lower = None
+                    self._body = expr.arg(0)
+                    self._body -= expr.arg(1)
+                    self._upper = ZeroConstant
+
+            
+            else:   # RangedExpression
```
jsiirola: This is a different type.  Do you ever validate that you get the correct type?

pyomo/core/base/constraint.py
```diff
@@ -641,6 +613,12 @@ def set_value(self, expr):
                     "non-finite term." % (self.name))
             assert self._lower is self._upper
 
+        #
+        # Compress the body expression
+        #
+        self._body = EXPR.compress_expression(self._body)
```
jsiirola: What is compress expression, why do we need it, and even if we need it, why should it be done here?  I think this should be deleted.

pyomo/core/base/external.py
```diff
@@ -54,18 +56,43 @@ def __init__(self, *args, **kwds):
...
+                args_.extend(val for val in args[i])
+            else:
+                args_.append(args[i])
+        #
+        # Loop and do two thing:
+        #   1. Wrap non-numeric arguments
+        #   2. See if we have a potentially variable argument
+        #
+        pv = False
+        for i in range(len(args_)):
```
jsiirola: this should be `for i,arg in enumerate(args_)`

That will avoid a list (in Python 2 -- since you are not using `six`), plus avoid a bunch of `__getitem__` calls below.

pyomo/solvers/plugins/solvers/gurobi_direct.py
michaelbynum: This looks good now. Thanks.

pyomo/solvers/plugins/solvers/gurobi_direct.py
```diff
@@ -163,43 +163,31 @@ def _apply_solver(self):
...
+        new_expr += repn.constant
 
         return new_expr, referenced_vars
 
     def _get_expr_from_pyomo_expr(self, expr, max_degree=2):
-        repn = generate_canonical_repn(expr)
+        if max_degree == 2:
+            repn = generate_standard_repn(expr, quadratic=True)
+        else:
+            repn = generate_standard_repn(expr, quadratic=False)
```
michaelbynum: The max_degree is determined by the version of Gurobi. Older versions of Gurobi do not support quadratic constraints. If an older version of Gurobi is being used and max_degree is 1, will "quadratic=False" cause an error to be raised if a user tries to use quadratic constraints, or will this quietly ignore the quadratic parts of the expression?

michaelbynum: Bill and I discussed this in person. This is handled correctly.

pyomo/core/base/external.py
whart222: In fact, the variable `i` is only used to index `args_`, so it's even better to replace this with `for arg in args:`

whart222: Hmmm... but the exception needs the variable `i`.

pyomo/core/base/constraint.py
whart222: Agreed.  This function is no longer needed.

pyomo/core/expr/expr_pyomo5.py
whart222: Named expressions get treated differently during cloning.  This method is needed to support duck typing for named expressions.

pyomo/core/expr/expr_pyomo5.py
whart222: That's probably true.  I haven't worried about performance of string operations in expr_pyomo5.py, since this only impacts debugging info right now.

pyomo/core/expr/expr_pyomo5.py
whart222: Agreed

pyomo/core/expr/expr_pyomo5.py
whart222: This is used in connector expansion.  I'm not sure I agree with your argument about the name.   I think it's reasonable to find a component by its class type.

pyomo/core/expr/expr_pyomo5.py
whart222: The semantics for deepcopy() and clone_expression() are different, which justifies a separate function.

pyomo/core/expr/expr_pyomo5.py
whart222: That would make swapping a different expression system more complicated in the future.

pyomo/core/expr/expr_pyomo5.py
whart222: Agreed

pyomo/core/expr/expr_pyomo5.py
whart222: This is deprecated functionality.  Hence, this captures the logic that will be deleted.

pyomo/core/expr/expr_pyomo5.py
whart222: As the comments note, these shouldn't be exposed.  But I needed to expose them for testing.  And there were some instances where we were using the this functionality unexpectedly.  Also, this includes deprecated functionality.

pyomo/core/expr/current.py
whart222: Agreed

pyomo/core/base/component.py
whart222: I confirmed that this method isn't needed here.

CHANGELOG.txt
```diff
@@ -24,6 +24,7 @@ Pyomo 5.4.1
 Pyomo 5.4
 -------------------------------------------------------------------------------
 
+=======
```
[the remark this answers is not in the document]
whart222: What should be reverted?

examples/dae/disease_DAE.py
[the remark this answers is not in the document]
whart222: It's not clear whether a change is necessary.  Or perhaps the code should be changed back to summation()?

examples/doc/pyomobook/pyomo-components-ch/param_misc.txt
whart222: The baseline wasn't up-to-date.

pyomo/contrib/preprocessing/plugins/remove_zero_terms.py
whart222: Reverted...

.travis.yml
```diff
@@ -153,12 +153,13 @@ script:
   - SITE_PACKAGES=`python -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())"`
   - echo "import coverage; coverage.process_startup()" > ${SITE_PACKAGES}/run_coverage_at_startup.pth
 
+  # Run documentation tests
+  - cd doc/OnlineDocs; make doctests -d; cd ../..
+  # nosetests -v --with-doctest --doctest-extension=.rst doc/OnlineDocs
+
```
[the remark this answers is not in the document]
whart222: I'm not sure why/how this changed.  I agree that we should re-order.

.travis.yml
```diff
@@ -124,7 +124,7 @@ install:
   #
   # Clone but don't install pyomo-model-libraries
   #
-  - git clone --quiet https://github.com/Pyomo/pyomo-model-libraries.git
+  - git clone -b expr_dev --quiet https://github.com/Pyomo/pyomo-model-libraries.git
```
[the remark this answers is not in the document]
whart222: Yep!

pyomo/core/expr/expr_pyomo5.py
whart222: Done

pyomo/repn/repn_ideas.py
whart222: Done

pyomo/repn/repn_ideas.py
whart222: Done

pyomo/pysp/tests/unit/test_instancefactory.py
```diff
@@ -289,9 +289,10 @@ def test_init5_default(self):
         self.assertEqual(len(factory._archives), 0)
         self.assertTrue("ReferenceModel" in sys.modules)
 
-    # model: name of .py file with model
-    # scenario_tree: Pyomo scenario tree model
+    @unittest.category('pyomo5_expr_failures')
```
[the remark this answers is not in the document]
whart222: I get a test failure on my Mac.

pyomo/solvers/plugins/solvers/gurobi_direct.py
whart222: Agreed.

pyomo/core/expr/expr_pyomo5.py
whart222: Done

examples/doc/pyomobook/pyomo-components-ch/param_misc.txt
jsiirola: I don't understand.  That example creates 2 models and pprints them both.  In the first model, 
```
model.p = Param(model.A, initialize=p)
```
`model.p` is immutable, so the `p`'s should (and ARE) removed from the expression.  That pprint ends at line 22 in the baseline.  In the second model, 
```
model.p = Param(model.A, initialize=p, mutable=True)
```
`model.p` is *mutable*.  Therefore,
```
model.o = Objective(expr=summation(model.p, model.x))
```
Should generate an expression `p[1]*x[1] + p[2]*x[2] + p[3]*x[3]` and **_NOT_** `x[1] + p[2]*x[2] + p[3]*x[3]`.  This baseline is *wrong*.

pyomo/core/util.py
whart222: @ghackebeil We need to move this example into a separate ticket.  I think the answer is "no", but then I don't really understand what a data_expression(1.0) object is.

whart222:
@ghackebeil @jsiirola @blnicho I believe that I have resolved all coding issues identified here.  There's one issue above that I suggested that Gabe move to a separate ticket.  I think the design of quicksum() is confusing, but intentional.

pyomo/core/util.py
ghackebeil: @whart222: The `data_expression` component is just an example of something that is NPV and mutable. You could replace it with a mutable parameter in the above and it would capture the same behavior. I've opened #431 and modified the example there.

pyomo/core/expr/expr_pyomo5.py
jsiirola: If we get rid of the expression-specific clone operation (and rely entirely on deepcopy), then named `Expression` objects will be handled correctly automatically.  I think that would allow us to remove this method.

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3407 @@
...
+        if node.__class__ in nonpyomo_leaf_types or not node.is_potentially_variable():
+            return True, 0
+
+        if not node.is_expression_type():
+            return True, 0 if node.is_fixed() else 1
+
+        return False, None
+
+
+def _polynomial_degree(node):
```
jsiirola: This "helper" function isn't used anywhere in Pyomo except one place in this file.  We should remove it.

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3407 @@
...
+    def polynomial_degree(self):
+        """
+        Return the polynomial degree of the expression.
+
+        Returns:
+            A nonnegative integer that is the polynomial degree of
+            the expression, if the expression is polynomial.  And
+            :const:`None` otherwise.
+        """
+        return _polynomial_degree(self)
```
jsiirola: This should just
```
 return _PolynomialDegreeVisitor().dfs_postorder_stack(node)
```

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3407 @@
...
+
+class NPV_ProductExpression(ProductExpression):
+    __slots__ = ()
+
+    def is_potentially_variable(self):
+        return False
+
+
+class MonomialTermExpression(ProductExpression):
+    __slots__ = ()
```
jsiirola: What happens when Expression replacement replaces one of the arguments with something that is not a MonomialExpression?  I am thinking of something like where we replace  `m.x` with a perspective function like `y * h(nu / y)`.

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3407 @@
...
+
+
+class ReciprocalExpression(ExpressionBase):
```
jsiirola: I think this should be returned to DivisionExpression.  I am not convinced that adding to the number of nodes in a tree is a good thing.

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -0,0 +1,3407 @@
...
+        return (self._args_[0].__class__ in native_numeric_types or self._args_[0].is_constant()) and \
+               (self._args_[1].__class__ in native_numeric_types or self._args_[1].is_constant()) and \
+               (self._args_[2].__class__ in native_numeric_types or self._args_[2].is_constant())
+
+    def is_potentially_variable(self):
+        return (self._args_[1].__class__ not in native_numeric_types and \
+                self._args_[1].is_potentially_variable()) or \
+               (self._args_[0].__class__ not in native_numeric_types and \
+                self._args_[0].is_potentially_variable()) or \
+               (self._args_[2]._
```
jsiirola: This seems specific to binary sums.  Should it not be moved to the derived class?

jsiirola: This is also specific to binary sums and should be moved to the derived class.

jsiirola: This is a very specific class to only define the name "sum" and the precedence, and makes me wonder why we have it, especially given the comments above about needing to check for the derived classes anyways.

jsiirola: Why does this not derive from SumExpression?  I think as declared, this appears to be an incomplete class (it has no data slots?).  Is it used anywhere?  Can it be removed?

pyomo/core/expr/numvalue.py
```diff
@@ -218,16 +267,35 @@ def potentially_variable(obj):
...
+    return as_numeric(obj).polynomial_degree()
 
 # It is very common to have only a few constants in a model, but those
 # constants get repeated many times.  KnownConstants lets us re-use /
 # share constants we have seen before.
 KnownConstants = {}
 
+def update_KnownConstants(obj, val):
+    if len(KnownConstants) < 100:
+        KnownConstants[obj] = val
```
jsiirola: What's the point of the "100" magic number?

jsiirola: Agreed, but just caching the first 100 is also probably bad. Plus that is an unlabeled embedded constant. 

Separately, @whart222 and I discussed whether we needed the KnownConstants at all...

pyomo/dae/tests/test_simulator.py
```diff
@@ -20,18 +21,18 @@
...
+)
 from pyomo.core.base.template_expr import (
     IndexTemplate, 
     _GetItemIndexer,
-    substitute_template_expression, 
-    substitute_getitem_with_param,
-    substitute_template_with_value,
+    #substitute_template_expression, 
+    #substitute_getitem_with_param,
+    #substitute_template_with_value,
```
blnicho: These commented imports should be removed. Also, I think you added an extra expression checker to the Simulator for NegationExpressions. We should add a test for that, I'll commit one shortly.

ghackebeil:
@whart222: Either before or after you merge, can you remove the `'pyomo5_expected_failures` test categorization stuff from PySP. I've verified that all of those tests pass locally, and I imagine they haven't been running on Travis because of this categorization.

whart222:
Thanks.  I'm going to leave these in.  I'm about at the point where I have clean tests, and I want to be able to merge soon.

PR #280, opened 2017-12-07
Deactivate connectors after they have been expanded

### LINKED ISSUE #274: GAMS writer does not accept models with Connector objects
The GAMS writer does not accept models with Connector objects. See discussion in https://github.com/Pyomo/pyomo/pull/253#discussion_r152372039 as well as #310 and #280 and #309

qtothec:
This also required making connectors deactivatable.

Fixes #274.

### DIFF
--- a/pyomo/core/base/connector.py
+++ b/pyomo/core/base/connector.py
@@ -2,8 +2,8 @@
 #
 #  Pyomo: Python Optimization Modeling Objects
 #  Copyright 2017 National Technology and Engineering Solutions of Sandia, LLC
-#  Under the terms of Contract DE-NA0003525 with National Technology and 
-#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain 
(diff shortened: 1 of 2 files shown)


whart222:
FWIW, I have a hard time evaluating this PR because we do not have very good documentation on connectors.  Specifically, I'd love to see some design documentation for developers in Sphinx.

qtothec:
Sounds to me like an invitation for a new issue.

whart222:
Good point.  I've added #297 to document the needed for this design documentation.  IMHO, that ticket should be a pre-requisite for completing this PR.

jsiirola:
I do not thing it makes sense to convert Connectors over to ActiveComponents, as it really doesn't make sense to "deactivate" a connector (in the same way that it doesn't make any sense to deactivate a Var or Param.  I recommend closing this PR.

PR #282, opened 2017-12-11
Some minor bug fixes in direct/persistent solver interfaces.

michaelbynum:
These should fix some (but not all) of the failing tests on Jenkins.


pyomo/solvers/plugins/solvers/cplex_direct.py
```diff
@@ -66,6 +66,7 @@ def _init(self):
             while len(self._version) < 4:
                 self._version += (0,)
             self._version = self._version[:4]
+            self._version = tuple([int(i) for i in self._version])
```
jsiirola: Can you combine this with the line before:
```
self._version = tuple(int(i) for i in self._version[:4])
```

pyomo/solvers/plugins/solvers/cplex_direct.py
```diff
@@ -356,7 +357,11 @@ def _add_sos_constraint(self, con):
 
         self._vars_referenced_by_con[con] = ComponentSet()
 
-        for v, w in con.get_items():
+        if hasattr(con, 'get_items'):
+            sos_items = con.get_items()
+        else:
+            sos_items = con.items()
```
jsiirola: What is `get_items`?  Is there a reason we don't just use `.items()`, or better yet, `six.iteritems(con)`?

pyomo/solvers/plugins/solvers/gurobi_direct.py
```diff
@@ -288,7 +292,11 @@ def _add_sos_constraint(self, con):
 
         self._vars_referenced_by_con[con] = ComponentSet()
 
-        for v, w in con.get_items():
+        if hasattr(con, 'get_items'):
+            sos_items = con.get_items()
+        else:
+            sos_items = con.items()
+        for v, w in sos_items:
```
jsiirola: Same question as above: why not use `six.iteritems(con)`?

ghackebeil:
I jumped the gun on this. I had noticed failures related to this when I was getting ready for a big merge yesterday, so I went ahead and made some of the obvious fixes. I didn't realize there was a PR started for this at the time.

To summarize, I made _some_ of these fixes plus a few others not here and already pushed them to master. I think there are still some fixes in this PR I did not catch, so please merge master into this branch and continue with the PR.

michaelbynum:
Thanks Gabe! Will do!

Michael

pyomo/solvers/plugins/solvers/cplex_direct.py
ghackebeil: get_items() applies when it is an SOSConstraint coming from the ASL. items() applies when it is an sos object from the kernel (where I've attempted to clean up the interface for various components, which sometimes requires little hacks like these in the writers until (hopefully) the AML adopts it)

`six.iteritems(con)` (for kernel sos): It is not implemented as a fully dict-like thing, though it certainly could be. I just went ahead and kept the implementation as simple as possible. So this would not work.

ghackebeil:
A few other things to try and fix:

 - The direct interfaces are now reporting the solver time as `results.solver.wallclock_time`. This one is new. When I collect solve time statistics in PySP I now need to check for `wallclock_time`, `user_time`, and `time`. It would be great if we could get this more standardized.
- The direct interfaces are no longer reporting the absolute optimality gap on `results.solution.gap`. We make heavy use of this value in PySP.

pyomo/solvers/plugins/solvers/gurobi_direct.py
jwatsonnm: See Gabe's response to the question above.

pyomo/solvers/plugins/solvers/cplex_direct.py
jwatsonnm: Done.

PR #284, opened 2017-12-16
Allow file objects for Data Input

### LINKED ISSUE #115: Documentation unavailable
The documentation for Pyomo 5.x is unavailable on the [Pyomo documentation website](http://www.pyomo.org/documentation) which only provides a link to v4.1. Also, the link to [PySP documentation](https://software.sandia.gov/trac/pyomo/export/HEAD/pyomo.doc/trunk/pysp/pyspdoc.pdf) is broken.

### LINKED ISSUE #283: Allow file objects as a valid DataPortal
This is to discuss the feature proposed [here](https://groups.google.com/forum/#!topic/pyomo-forum/5jscHaCwJSs) so that
``` python
input_dat = io.StringIO("""
param n := 672 ;
param days_horizon := 7 ;
param delta_t := 0.25 ;
""")
instance = model.create_instance(input_dat)
```

does not result in
```
ValueError: Cannot load model model data from with object of type '<class '_io.StringIO'>'
```

Initially, I was looking for the place where the [file object](https://docs.python.org/3/glossary.html#term-file-object) API would be used.  Then I found the DataPortal machinery and figured that implementing a file object DataPortal would be the way to go.  See also #284.  WDYT?

fleimgruber:
* pyomo/core/base/PyomoModel.py (Model.load): DataPortal init for file
  object.

* pyomo/core/base/DataPortal.py (DataPortal.__init__): Discriminate on
  file object in kwargs and pass to DataManagerFactory.

* pyomo/core/data/process_data.py (_process_include): Pass data directly
  from file object to parse_data_commands.

* pyomo/core/plugins/data/datacommands.py: Explicit new file object to
  distinguish between .dat files and file objects.

* pyomo/core/tests/unit/test_file_object.py: Test if io.StringIO file object is
  valid for creating an abstract model instance.

Closes #283

### DIFF
--- a/pyomo/core/base/DataPortal.py
+++ b/pyomo/core/base/DataPortal.py
@@ -61,6 +61,12 @@ class DataPortal(object):
             self.connect(filename=filename, **kwds)
             self.load()
             self.disconnect()
+        # Or load data from a file object (e.g. StringIO)
+        if 'file_object' in kwds:
+            file_object = kwds.pop('file_object')
+            self.connect(file_object=file_object, **kwds)
+            self.load()
+            self.disconnect()
         # Or load data from a dictionary
         elif 'data_dict' in kwds:
             data = kwds.pop('data_dict')
@@ -87,6 +93,8 @@ class DataPortal(object):
             data = kwds.get('filename',None)
         if data is None:
             data = kwds.get('server',None)
+        if data is None:
+            data = 'file_object'
         if '.' in data:
             tmp = data.split(".")[-1]
         else:
--- a/pyomo/core/base/PyomoModel.py
+++ b/pyomo/core/base/PyomoModel.py
@@ -785,6 +785,8 @@ arguments (which have been ignored):"""
             dp = arg
         elif type(arg) is dict:
             dp = DataPortal(data_dict=arg, model=self)
+        elif hasattr(arg, 'read'):
+            dp = DataPortal(file_object=arg, model=self)
         elif isinstance(arg, SolverResults):
             if len(arg.solution):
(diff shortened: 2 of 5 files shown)


fleimgruber:
I tried my best to grok the DataPortal API.  Now any object that implements the `read` method of a [file object](https://docs.python.org/3/glossary.html#term-file-object) can be passed to `AbstractModel().create_instance`.

whart222:
This is a good idea, but I'm not sure about this implementation.  Let me look into this and see if there's a simpler way to support this functionality.

fleimgruber:
Yeah, given my limited knowledge about the DataPortal API and my impression that this part of the codebase grew organically layer by layer, I also felt this implementation to be awkward.  I was also thinking about a refactoring of some of the DataPortal code since supported `.dat` files would also fit into the file object abstraction.  Also FWIW, the commit links to the wrong issue, the correct one would be #283.

whart222:
I've looked at this a bit, and I think it would make sense for all of the DataManager plug-ins to accept a Python file object.  This wouldn't be limited to being a StringIO, but could be a file opened by the user.  

That would allow, for example, a JSON object to be loaded from a StringIO object as well as a Python data commands file.

This sort of generalization would require changes to the TableData class, but it wouldn't involve a new plug-in class but rather a rework of the APIs of existing plug-ins.

I've started prototyping this idea, with the goal of adding a new PR.  Let me know what you think.

mrmundt:
Because this draft implementation was meant as a discussion rather than an permanent implementation of #283, we are going to close this PR.

PR #299, opened 2018-01-03
Adding pyomo.contrib.example

whart222:
This PR creates the pyomo.contrib.example package, which illustrates how to directly extend pyomo.contrib to include third-party contributes.

This PR also includes Sphinx documentation that discusses two strategies for including third-party contributions:

1. pyomo.contrib.simplemodel - Including symbols from an external third-party package

1. pyomo.contrib.example - Including a third-party package within the Pyomo source tree

I manually confirmed that (a) the plugins are included from pyomo.contrib.example when this package is imported, and (b) that tests in this package are only run when the TEST_PYOMO_CONTRIB environment variable is set.

This PR was motivated by discussions with the Pyomo developer team about how to manage contributions like PR #248.  We agreed that these should be treated as optional capabilities, and pyomo.contrib seems like a natural way to include them in the Pyomo source tree without forcing a contributor to develop a third-party package (e.g. as was done with pyomocontrib_simplemodel).


doc/OnlineDocs/pyomo_contrib.rst
```diff
@@ -0,0 +1,74 @@
+Third-Party Contributions
+=========================
+
+Pyomo uses the ``pyomo.contrib`` package to facilitate the integrate
```
blnicho: Typo, remove "the integrate"

doc/OnlineDocs/pyomo_contrib.rst
```diff
@@ -0,0 +1,74 @@
...
+
+* Tests in this package are skipped unless the ``TEST_PYOMO_CONTRIB`` environment variable is set to ``1``.
+
+Including contrib packages in the Pyomo source tree provides a
+convenient mechanism for defining new functionality that can be
+optionally deployed by users.  We expect this mechanism to include
+Pyomo extensions and experimental modeling capabilities.  However,
+contrib packages are treated as optional packages until the Pyomo
+developers decide that they should be fully supported in future
+releases.
```
blnicho: I think you should add another sentence that explicitly states that source code included in pyomo.contrib will not be maintained by the Pyomo developer team and it is the responsibility of the code contributor to keep it up-to-date.

pyomo/contrib/example/tests/test_example.py
```diff
@@ -0,0 +1,25 @@
+#
+# Only run the tests in this package if the TEST_PYOMO_CONTRIB environment
+# variable has been set to '1'.  Otherwise, skip the tests.
+#
+import os
+import sys
+if os.environ.get('TEST_PYOMO_CONTRIB', 0):
```
blnicho: It looks like you're putting the responsibility of checking this environment variable on the code contributor. Wouldn't it make more sense to include this check in the main pyomo testing script to determine if the testing folders in pyomo.contrib are even descended into? I would be worried about a malicious/novice contributor not including this logic in their own code and a core developer forgetting about this requirement before merging a PR.

doc/OnlineDocs/pyomo_contrib.rst
```diff
@@ -0,0 +1,74 @@
...
+Pyomo uses the ``pyomo.contrib`` package to facilitate the integrate
+the inclusion of third-party contributions that enhance Pyomo's
+core functionaliy.  The following two examples illustrate two ways
+that ``pyomo.contrib`` can be used to integrate third-party
+contributions.
+
+Including External Packages
+---------------------------
+
+The `pyomocontrib_simplemodel
```
jsiirola: I would rather see this be documentation on how to implement a contrib package.  I think the actual documentation for simplemodel and example (even the brief overview) should be in their respective directories and linked from here.

doc/OnlineDocs/pyomo_contrib.rst
```diff
@@ -0,0 +1,74 @@
...
+    from pyomo.environ import *
+    from pyomo.contrib.example import a
+
+    # Print the value of 'a' defined by this package
+    print(a)
+
+Although ``pyomo.contrib.example`` is included in the Pyomo source tree, it is
+treated as an optional package.  Consequently,
+
+* Plugin classes defined in this package are not loaded when `pyomo.environ` is loaded.  Plugins are loaded when the ``pyomo.contrib.example`` package is imported.
```
jsiirola: Why not?  I would think that we would want all contrib plugins to be loaded by `pyomo.environ`.  The "optional" part should be indicated by prefacing the plugin names with "`contrib.`"  Plugins that shouldn't be auto-loaded shouldn't be put in a "plugins" directory.

doc/OnlineDocs/pyomo_contrib.rst
whart222: I'm going to add this commentary at the top of the section, to provide broad context for pyomo.contrib.

pyomo/contrib/example/tests/test_example.py
whart222: @blnicho Yes, there is a risk that TEST_PYOMO_CONTRIB is not appropriately checked.  But contributions in pyomo.contrib will be reviewed by Pyomo developers, which mitigates that risk.  This process is compatible with the Python ``nose`` tool;  I couldn't think of a strategy that would work with ``nose`` in the manner you suggest.

@jsiirola I think we should not run the contrib tests by default.  That would impede normal Pyomo development.  But, rather, we want to be able to enable these tests as part of our Travis/appveyor tests to check whether the packages are being maintained.

doc/OnlineDocs/pyomo_contrib.rst
```diff
@@ -0,0 +1,74 @@
+Third-Party Contributions
```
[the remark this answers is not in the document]
whart222: @jsiirola I think we should have the links for contrib package documentation elsewhere.  This documentation pertains to whether/how/when we support contrib packages.  That's really developer documentation.  The links to contrib packages are really end-user documentation.

doc/OnlineDocs/pyomo_contrib.rst
whart222: I disagree.  These packages should not be loaded automatically.  They should be optional, so that when/if they fail then a user can safely ignore them.  For example, if I change pyomo.core and break all of the contrib packages, then I and users should be able to import pyomo.environ to test just the main part of pyomo.  That is, errors in pyomo.contrib.* packages should never break the normal operation of pyomo.  Consequently, these packages should always be optional in every sense.

Thus, a pyomo.contrib package is only loaded when explicitly imported, including its plugins.

whart222:
NOTE:  There isn't a top-level "developer" documentation area.  I've started this on another branch (for expressions).  When that get's merged, then I plan to re-organize this documentation.

doc/OnlineDocs/pyomo_contrib.rst
jsiirola: How you envision users cause contributed plugins to be imported?  The "standard" pyomo packages (core, solvers, etc) (intentionally) do not load the plugins when the packages are loaded (the plugin imports are encapsulated in the "load()" method.

It really sounds like `pyomo.contrib.*.plugins` should all be loaded by default, but be marked (in `pyomo.environ`) as "optional", so that failed imports do not cause environ to fail to import.

whart222:
I think I've addressed the concerns raised by @jsiirola and @blnicho .  pyomo.contrib packages are now imported and tested by default.  In fact, testing by default didn't require any changes to the test.pyomo script.  Adding support to ignore pyomo.contrib tests will require changes to pyutilib, which should not hold up approval of this PR.

PR #301, opened 2018-01-04
Allow None to be a valid value for Params

### LINKED ISSUE #300: Default mutable Param value
The following used to work:

```python
m = ConcreteModel()
model.p = Param(mutable=True)
assert model.p.value == None
```

This now results in an error when `model.p.value` is accessed with the message "The Param value is undefined and no default value is specified”. I’m not here to fight that behavior, but I do often make use of the ability to count on mutable parameters having a default value of `None`. It is a very natural starting state. There is also some critical behavior in PySP that relies on this functionality, and this issue is currently holding up a merge of the PySP dev branch into master. 

My naive attempt to get around this issue was to change the Param declarations to look something like

```python
model.p = Param(mutable=True, default=None)
# or
model.p = Param(mutable=True, initialize=None)
```

But this actually doesn’t fix anything because the value of `None` just gets interpreted as “not set”.

What do folks think about changing the semantics of Param so that explicitly setting keywords like "default" and "initialize" to `None` gets interpreted differently from not using the keywords at all? There are various ways this can be implemented. E.g.,

```python
_noarg = object()
def f(x=_noarg):
    if x is not _noarg:
        ...
```
or
```python
def f(**kwds):
    if 'x' in kwds:
        ...
```

jsiirola:
This fixes #300 by allowing `None` to be a valid value for Params, including using it as a valid initialization and as a default value.

Note that as part of reviewing this PR, we should have a deeper discussion as to if we want "None" to be a valid value for Params.


qtothec:
This might be a good opportunity to update the operator overload warnings to be more descriptive when you try to do operations against `None`.
`TypeError: Cannot convert object of type 'NoneType' (value = None) to a numeric value.` makes sense to me, but may be confusing for someone new to Pyomo.

jsiirola:
@qtothec: I am open to making any exception / error / warning more descriptive.  However, I am trying to get away from "monster" PRs when we can (they take too long to review and merge in).  I would strongly encourage us to separate PRs that improve the user experience (i.e., errors, etc.) from functionality changes (like this PR).

pyomo/core/base/param.py
```diff
@@ -878,11 +870,16 @@ def _pprint(self):
         """
         Return data that will be printed for this component.
         """
+        if self._default_val is _NotValid:
+            default = "None" # for backwards compatibility in reporting
```
jsiirola: I debated this.  By returning "None", Param's pprint is consistent with previous behavior (i.e., no baselines had to be updated); however, now pprint() will not distinguish between "No default value" and a default value of None.  The downside of doing the right thing and changing pprint() to return something like "{no default}" or even "{none}" is that just about *every* pprint() baseline would change.

pyomo/core/base/indexed_component.py
```diff
@@ -40,6 +40,9 @@ def normalize_index(index):
     return idx
 normalize_index.flatten = True
 
+class _NotValid(object):
+    pass
```
jsiirola: Having written this, I am wondering if this flag should be renamed `_NotPresent` (or `_NotFound`)?  I worry that it will become confusing having `param._NotValid`, which indicated that an invalid value is present (either as the Default or as the value of a scalar) along with `indexed_component._NotValid`.

pyomo/core/base/param.py
```diff
@@ -40,6 +40,10 @@ def _raise_modifying_immutable_error(obj, index):
         "declare the parameter as mutable [i.e., Param(mutable=True)]"
         % (name,))
 
+class _NoArgument(object):
```
ghackebeil: This class is defined again a few lines down. Delete

ghackebeil:
`default=None` seems to work as expected on this branch, but `initialize=None` does not. Is this intentional?

ghackebeil:
Oops. Nevermind. I was testing on the wrong branch.

PR #302, opened 2018-01-06
add logfile option to GAMS solver

daviskirk:
This would add an option for specifying a custom logfile for the GAMS command line solver.  The other solvers seem to support such an option and it is quite useful if one want more control over the logfiles destination.

### DIFF
--- a/pyomo/solvers/plugins/solvers/GAMS.py
+++ b/pyomo/solvers/plugins/solvers/GAMS.py
@@ -741,7 +741,10 @@ class GAMSShell(pyomo.util.plugin.Plugin):
         exe = self.executable()
         command = [exe, output_filename, 'o=' + lst_filename]
         if tee and not logfile:
-            pass  # default behaviour of gams is to print to console
+            # default behaviour of gams is to print to console, for
+            # compatability with windows and *nix we want to explicitly log to
+            # stdout (see https://www.gams.com/latest/docs/UG_GamsCall.html)
+            command.append("lo=3")
         elif not tee and not logfile:
             command.append("lo=0")
         elif not tee and logfile:
--- a/pyomo/solvers/plugins/solvers/GAMS.py
+++ b/pyomo/solvers/plugins/solvers/GAMS.py
@@ -138,6 +138,8 @@ class GAMSDirect(pyomo.util.plugin.Plugin):
 
         tee=False:
             Output GAMS log to stdout.
+        logfile=None:
+            Optionally a logfile can be written.
         load_solutions=True:
             Optionally skip loading solution into model, in which case
             the results object will contain the solution data.
@@ -194,6 +196,7 @@ class GAMSDirect(pyomo.util.plugin.Plugin):
 
         load_solutions = kwds.pop("load_solutions", True)
         tee            = kwds.pop("tee", False)
+        logfile        = kwds.pop("logfile", None)
         keepfiles      = kwds.pop("keepfiles", False)
         tmpdir         = kwds.pop("tmpdir", None)
         report_timing  = kwds.pop("report_timing", False)
@@ -252,7 +255,8 @@ class GAMSDirect(pyomo.util.plugin.Plugin):
         t1 = ws.add_job_from_string(output_file.getvalue())
 
         try:
-            t1.run(output=sys.stdout if tee else None)
+            with OutputStream(tee=tee, logfile=logfile) as output_stream:
+                t1.run(output=output_stream)
         except GamsExceptionExecution as e:
             try:
                 if e.rc == 3:
@@ -1046,6 +1050,49 @@ class GAMSShell(pyomo.util.plugin.Plugin):
                   (postsolve_completion_time - initial_time))
 
         return results
+
+
+class OutputStream:
+    """Output stream object for simultaneously writing to multiple streams.
+
+    tee=False:
+        If set writing to this stream will write to stdout.
+    logfile=None:
+        Optionally a logfile can be written.
+
+    """
+
+    def __init__(self, tee=False, logfile=None):
+        """Initialize output stream object."""
+        if tee:
+            self.tee = sys.stdout
+        else:
+            self.tee = None
+        self.logfile = logfile
+        self.logfile_buffer = None
+
+    def __enter__(self):
+        """Enter context of output stream and open logfile if given."""
+        if self.logfile:
+            self.logfile_buffer = open(self.logfile, 'a')
+        return self
+
+    def __exit__(self, *args, **kwargs):
+        """Enter context of output stream and close logfile if necessary."""
+        if self.logfile_buffer:
+            self.logfile_buffer.close()
+        self.logfile_buffer = None
+
+    def write(self, message):
+        """Write messages to all streams."""
+        if self.tee:
+            self.tee.write(message)
+        if self.logfile_buffer:
+            self.logfile_buffer.write(message)
+
+    def flush(self):
+        """Needed for python3 compatibility."""
+        pass
 
 
 def check_expr_evaluation(model, symbolMap, solver_io):
--- a/pyomo/solvers/plugins/solvers/GAMS.py
+++ b/pyomo/solvers/plugins/solvers/GAMS.py
@@ -10,7 +10,7 @@
 
 from six import StringIO, iteritems, itervalues
 from tempfile import mkdtemp
-import os, sys, subprocess, math, logging, shutil, time
+import os, sys, math, logging, shutil, time
 
 from pyomo.core.base import (Constraint, Suffix, Var, value,
                              Expression, Objective)
@@ -752,7 +752,7 @@ class GAMSShell(pyomo.util.plugin.Plugin):
             command.append("lf=" + str(logfile))
 
         try:
-            rc = subprocess.call(command)
+            rc, _ = pyutilib.subprocess.run(command)
 
             if keepfiles:
                 print("\nGAMS WORKING DIRECTORY: %s\n" % tmpdir)
--- a/pyomo/solvers/plugins/solvers/GAMS.py
+++ b/pyomo/solvers/plugins/solvers/GAMS.py
@@ -619,6 +619,8 @@ class GAMSShell(pyomo.util.plugin.Plugin):
 
         tee=False:
             Output GAMS log to stdout.
+        logfile=None:
+            Optionally a logfile can be written.
         load_solutions=True:
             Optionally skip loading solution into model, in which case
             the results object will contain the solution data.
@@ -669,6 +671,7 @@ class GAMSShell(pyomo.util.plugin.Plugin):
 
         load_solutions = kwds.pop("load_solutions", True)
         tee            = kwds.pop("tee", False)
+        logfile        = kwds.pop("logfile", None)
         keepfiles      = kwds.pop("keepfiles", False)
         tmpdir         = kwds.pop("tmpdir", None)
         report_timing  = kwds.pop("report_timing", False)
@@ -733,8 +736,14 @@ class GAMSShell(pyomo.util.plugin.Plugin):
 
         exe = self.executable()
         command = [exe, output_filename, 'o=' + lst_filename]
-        if not tee:
+        if not tee and not logfile:
             command.append("lo=0")
+        elif not tee and logfile:
+            command.append("lo=2")
+        elif tee and logfile:
+            command.append("lo=4")
+        if logfile:
+            command.append("lf=" + str(logfile))
 
         try:
             rc = subprocess.call(command)
--- a/pyomo/solvers/plugins/solvers/GAMS.py
+++ b/pyomo/solvers/plugins/solvers/GAMS.py
@@ -1075,26 +1075,29 @@ class OutputStream:
 
     def __enter__(self):
         """Enter context of output stream and open logfile if given."""
-        if self.logfile:
+        if self.logfile is not None:
             self.logfile_buffer = open(self.logfile, 'a')
         return self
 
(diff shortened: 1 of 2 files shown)


jsiirola:
While there is nothing wrong with this PR, I would like to see tests added for the 4 permutations of `{tee|not tee}` and `{logfile|logfile=None}`, in part to ensure that they all work correctly and to be defensive against future changes in the GAMS API.

They can be added to `solvers/tests/checks/test_GAMS.py`.

daviskirk:
Sure thing, I'll add the tests

pyomo/solvers/tests/checks/test_GAMS.py
```diff
@@ -162,5 +162,53 @@ def test_keepfiles_gms(self):
 
             shutil.rmtree(tmpdir)
 
+
+class GAMSTests(unittest.TestCase):
```
gseastream: The additional tests should be added to the GAMSTest class above instead of creating a new class. This new class actually overwrites the tests above, which I verified in the travis log file (https://travis-ci.org/Pyomo/pyomo/jobs/329755594). Alternatively the class name can be changed for the new class.

daviskirk: Yeah noticed that too, sorry.  Got the tests running locally now so next time will be smoother.

daviskirk:
The test failing here didn't fail in the previous build's but did fail in appveyor.  However, I don't see how it can be related to these changes.

```
ERROR: test_clone_unclonable_attribute (pyomo.core.tests.unit.test_block.TestBlock)
  File "/home/travis/build/Pyomo/pyomo/pyomo/core/tests/unit/test_block.py", line 1852, in test_clone_unclonable_attribute
    sorted(id(x) for x in identify_variables(n.c.body)),
  File "/home/travis/build/Pyomo/pyomo/pyomo/core/tests/unit/test_block.py", line 1852, in <genexpr>
    sorted(id(x) for x in identify_variables(n.c.body)),
  File "/home/travis/build/Pyomo/pyomo/pyomo/core/kernel/expr_coopr3.py", line 91, in identify_variables
    _argList = _sub._args
AttributeError: _args
```

blnicho:
Have you pulled from Pyomo master recently?

ghackebeil:
That test fails on that specific Python version about 1 out of every 100 builds or so. Nothing to worry about (at least for this PR).

pyomo/solvers/tests/checks/test_GAMS.py
```diff
@@ -162,5 +162,48 @@ def test_keepfiles_gms(self):
...
+        m.o = Objective(expr= m.x)
+        self.m = m
+        self.tmpdir = mkdtemp()
+        self.logfile = os.path.join(self.tmpdir, 'logfile.log')
+
+    def tearDown(self):
+        """Clean up temporary directory after tests are over."""
+        shutil.rmtree(self.tmpdir)
+
+    def test_tee_gms(self):
```
gseastream: It seems the name of this test got mixed up with the one below, should be `test_no_tee_gms`.

daviskirk:
I'll look at the api solver and add the tests for that too

gseastream:
There is still the comment I made above about the apparent mix up of test names, where it looks like test_tee_gms is mixed up with test_no_tee_gms

daviskirk:
@michaelbynum I'm not sure how to redirect stdout in the gams command line solver as I think that the output is produced by the subprocess and not within pyomo or python if I understand correctly.  I'm not sure how the gams api works but it might work similarly.  Is there an example of some other tests (for other solvers for example) where this is done?  I couldn't find any grepping around

daviskirk:
@gseastream I added functionality and for using the GAMS python api in #0654fbf2

jsiirola:
You can grab the stdout/stderr using `setup_redirect` / `reset_redirect`:

```
from pyutilib.misc import setup_redirect, reset_redirect
from six import StringIO
#...
    def test_tee_gms(self):
        with SolverFactory("gams", solver_io="gms") as opt:
            OUTPUT = StringIO()
            setup_redirect(OUTPUT)
            try:
                opt.solve(self.m, tee=True)
            finally:
                reset_redirect()
            self.assertFalse(os.path.exists(self.logfile))
            # Then check that whatever output you wanted is in OUTPUT.getvalue()
```

daviskirk:
@jsiirola The gams api seems to work fine (it uses the python stdout) but the gams command somehow bypasses it.  I can always see the output (in the console) but I can't capture it.

gseastream:
Not sure if it helps, but `pyutilib.subprocess.run` returns what is printed to stdout. That's how `GAMSShell._get_version` works.

michaelbynum:
@daviskirk I don't think tee=True and logfile is None works correctly. I don't think there is a case for this in the if-else statements on lines 743-750. It looks to me like there will not be any output to the console in this case. All the test checks is that the logfile does not exist.

daviskirk:
@gseastream thanks for the tip!
@michaelbynum there's no case in the if/else statement since no option is added to the command.  Without any "lo" command the default is to print to console, which is what we want to do when using tee.  I added an explicit "pass" and a comment to make this clearer for future readers.

jsiirola:
@michaelbynum: this seems to be a problem with the redirection not working on Windows.  

It appears that on Windows, `subprocess.call()` does *not* send its output to `sys.stdout`.  I have no idea why this is - or really how to "fix" the behavior of `subprocess.call()`.  The best solution I can come up with is that `pyomo/solvers/plugins/solvers/GAMS.py` line 755 should use `pyutilib.subprocess.run(command)` instead of `subprocess.call`.  Pyutilib's `subprocess.run` relies on `subprocess.Popen` and not `call`, which seems to work correctly.  This change would also simplify the testing, as both the GmsTests and PyTests ca use `redirected_stdout` (and `redirected_subprocess_call` can be deleted).

pyomo/solvers/tests/checks/test_GAMS.py
```diff
@@ -162,5 +165,146 @@ def test_keepfiles_gms(self):
...
+
+@contextlib.contextmanager
+def redirected_stdout():
+    """Temporarily redirect stdout into a string buffer."""
+    output = StringIO()
+    try:
+        setup_redirect(output)
+        yield output
+    finally:
+        reset_redirect()
```
jsiirola: I like that this is implemented as a context manager.  Given that, the try-except is unnecessary.  We should be fine with:
```
def redirected_stdout():
    """Temporarily redirect stdout into a string buffer."""
    output = StringIO()
    setup_redirect(output)
    yield output
    reset_redirect()
```
This is so generally useful that we should promote it to pyomo.misc.redirect_io.

pyomo/solvers/tests/checks/test_GAMS.py
```diff
@@ -162,5 +165,146 @@ def test_keepfiles_gms(self):
...
+                opt.solve(self.m, logfile=self.logfile)
+        self.assertFalse(output.getvalue())
+        self._check_logfile(exists=True)
+
+    def test_tee_and_logfile(self):
+        with SolverFactory("gams", solver_io="gms") as opt:
+            with redirected_subprocess_call() as output:
+                opt.solve(self.m, logfile=self.logfile, tee=True)
+        self.assertTrue(output.getvalue())
+        self._check_logfile(exists=True, expected=output.getvalue())
```
jsiirola: Given the suggested change from `subprocess.call` to `pyutilib.subprocess.run` in `GAMSShell.solve()`, we can convert all these from `redirected_subprocess_call` to `redirected_stdout`

pyomo/solvers/tests/checks/test_GAMS.py
```diff
@@ -162,5 +165,146 @@ def test_keepfiles_gms(self):
...
+    output = StringIO()
+    try:
+        setup_redirect(output)
+        yield output
+    finally:
+        reset_redirect()
+
+
+@contextlib.contextmanager
+def redirected_subprocess_call():
```
jsiirola: We can delete this method.

pyomo/solvers/tests/checks/test_GAMS.py
```diff
@@ -162,5 +165,146 @@ def test_keepfiles_gms(self):
...
+
+        """
+        if not exists:
+            self.assertFalse(os.path.exists(self.logfile))
+            return
+
+        self.assertTrue(os.path.exists(self.logfile))
+        with open(self.logfile) as f:
+            logfile_contents = f.read()
+        self.assertTrue(logfile_contents)
```
jsiirola: In general, I do not like `assertTrue` / `assertFalse` for testing for non-empty / empty strings.  A much stronger test is to decide on a characteristic string that should be there ("GAMS" is probably a good guess), or to `assertEqual(logfile_contents, "")`.

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -1039,6 +1054,49 @@ def solve(self, *args, **kwds):
...
+    def write(self, message):
+        """Write messages to all streams."""
+        if self.tee:
+            self.tee.write(message)
+        if self.logfile_buffer:
+            self.logfile_buffer.write(message)
+
+    def flush(self):
+        """Needed for python3 compatibility."""
+        pass
```
jsiirola: You should at least pass the `flush` to the underlying file, if applicable.

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -1039,6 +1054,49 @@ def solve(self, *args, **kwds):
...
 
+class OutputStream:
+    """Output stream object for simultaneously writing to multiple streams.
+
+    tee=False:
+        If set writing to this stream will write to stdout.
+    logfile=None:
+        Optionally a logfile can be written.
+
+    """
```
jsiirola: I am torn about introducing an entirely new class to manage this, however, I see the merits of having this as a context manager.

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -1039,6 +1054,49 @@ def solve(self, *args, **kwds):
...
+        if self.logfile_buffer:
+            self.logfile_buffer.close()
+        self.logfile_buffer = None
+
+    def write(self, message):
+        """Write messages to all streams."""
+        if self.tee:
+            self.tee.write(message)
+        if self.logfile_buffer:
+            self.logfile_buffer.write(message)
```
jsiirola: I would rather these tests be against None and not rely on implicit casts to bool.  That is:
```
if self.tee is not None:
    self.tee.write(message)
if self.logfile_buffer is not None:
    self.logfile_buffer.write(message)
```
...the same applies to the other methods in this class.

pyomo/solvers/tests/checks/test_GAMS.py
jsiirola: I stand corrected.  I am a bit surprised with this, as I thought the point of a context manager was to ensure that the `__exit__` is always called.  Given this behavior, I am actually a bit surprised that the `contextlib.contextmanager` decorator actually works...

pyomo/solvers/tests/checks/test_GAMS.py
daviskirk: This can't be deleted as far as I can see since what it does is use the output it gathers and write it into the stream buffer it returns.  Using redirect_stdout unfortunately results in the same empty string we had before.

pyomo/solvers/plugins/solvers/GAMS.py
daviskirk: Done in dbcff25a

pyomo/solvers/plugins/solvers/GAMS.py
daviskirk: Done in dbcff25a

pyomo/solvers/tests/checks/test_GAMS.py
daviskirk: converted `subprocess.call` to `pyutilib.subprocess.run` but left the test calls as they were (see comment below)

daviskirk:
I don't know why the windows output isn't captured.  I don't have a windows system available so I can't tell if this is just the capturing part or the actual output on windows that is the problem.

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -733,11 +740,19 @@ def solve(self, *args, **kwds):
 
         exe = self.executable()
         command = [exe, output_filename, 'o=' + lst_filename]
-        if not tee:
+        if tee and not logfile:
+            pass  # default behaviour of gams is to print to console
```
jsiirola: I think we can fix Windows tests by changing this path from an implicit "`lo=1`" to an explicit "`lo=3`".  From the GAMS documentation (https://www.gams.com/latest/docs/UG_GamsCall.html):

Value | Meaning
-- | --
0 | No log output
1 | Log output to screen (console)
2 | Log output to logfile
3 | Log output to standard output
4 | Log output to logfile and standard output

Note that on *nix, the console and stdout are usually synonymous; however, on Windows they are not.  Making the output explicitly go to stdout should get redirection working.

daviskirk: Thanks!

jsiirola:
BTW: Did you happen to test if we can get away without the separate `redirected_subprocess_run` after the move to using `lo=3`?

daviskirk:
@jsiirola yes, but unfortunatly it still doesn't work.  Both the `lo=3` and the `lo=4` for tee and logfile together give a `""` output if used with the `redirected_stdout` instead of the `redirected_subprocess_run` so this wasn't a pure `lo=3` issue.

PR #310, opened 2018-01-10
Fix so that GAMS writer can safely ignore Connector objects

### LINKED ISSUE #274: GAMS writer does not accept models with Connector objects

qtothec:
This is the same as #253, just updated.

Fixes #274


jsiirola:
@blnicho: good point.  The LP / NL writers should produce errors because they explicitly walk the tree should fail if a Connector is found in an active expression.  The problem you are seeing is that the way the GAMS (and I suppose Baron) interfaces emit Constraint expressions is to use the normal "`to_string()`" method and then edit the resulting string.  `to_string` is supported by Connector, so the string is cleanly generated.  Note that simply reclassifying the Connector as something else isn't sufficient: the original Constraint that contains a Connector is *never* valid.  Upon expansion, the original Constraint is deactivated and a set of new constraints are emitted, where the Connector is substituted for one of the contained objects (Var/Param//Expression).

The problem with the GAMS/Baron writers should go away after the move to pyomo5 expressions (not because of a Connector rewrite), as we will use custom walkers for both Baron and GAMS that should throw exceptions if an unrecognized ctype is encountered when walking the tree.

For what it's worth, the same principle applies to DAE: it is not the existence of a DerivativeVar that should cause an error, but rather the presence of a DerivativeVar *in an active Constraint*.

I agree with needing a FIXME / TODO for the GAMS/Baron writers after the pyomo5 expression merge.  Actually, it should happen as *part* of merging #272.

qtothec:
@jsiirola It turns out that the current implementation does not detect the presence of `Connector` in active constraints. I'm trying to figure out the best way to implement that.

Is there a cheap way to check whether a constraint body contains a connector? I was trying to add a check to the variable labeler (`var_label` function), but as it turns out, `SimpleConnector` uses the `Component.to_string` function, which completely ignores any passed labeler.

What would be the best way to deal with this?

jsiirola:
This is tricky.  The problem is that we use `to_string` to generate the raw GAMS string.  The new expression system (#272) will render this moot, as the GAMS writer will be updated to use it's own custom walker to generate the expression string representations. That walker will easily be able to detect the existence of unexpected node types (like Connectors).

For the time being, I think we can do one of two things:
  1. Play ostrich and ignore errors due to models with Connectors still embedded
  2. Put together a quick-and-dirty routine to walk the expression and check for Connectors (and throw errors if they are encountered)

2 would be a performance hit, but would help ensure that the models are always correct -- so I am leaning that direction.

qtothec:
@jsiirola By "walk the expression," do you mean making a loop descending into all the `_args`? Is there a cleaner way of doing it?

whart222:
@qtothec In the new expression system, there are pre-defined expression walkers that can be easily customized.  A developer should never need to touch the `_args` data in an expression!

qtothec:
@whart222 Right. Per @jsiirola's comment above, this issue goes away with the advent of the new expression system. I was wondering in terms of a nearer-term solution. Is the new expression system going to be merged in within the next month? If so, we might just hold off on this until afterwards.

jsiirola:
@qtothec:  Yes: The near-term solution is to walk the _args (+numerator & denominator for coopr3 _ProductExpression) and check for any Connector leaf nodes.  It's an ugly bit of ~25 lines, but will at least produce reasonable errors for the 5.4 release.  That code will be ripped out as soon as the expr_pyomo5 system is merged in (hopefully by the end of February).

PR #311, opened 2018-01-11
Better CL options and faster soln file parsing

rsmith54:
This PR makes some improvements to the ```CBCplugin.py```.

First, it fixes the command line parsing to work with some additional options (CBC is quite finicky).  In particular, one can now run using the barrier method, which did not work for me on the master branch.

Second, it makes significant speedups to the soln parsing code, by avoiding unnecessarily running many string comparisons at the beginning of the file lines loop, as we only need to parse the header once.  We don't see the header again, so these string checks are wrapped inside the ```if not header_processed``` statement.  I think additional improvements could also be made here, but this is already significantly faster from my testing ( > 50% on a 10 MB soln file).  

Please let me know if there are additional checks, tests, or code I can add.  Thanks.

### DIFF
--- a/pyomo/solvers/plugins/solvers/CBCplugin.py
+++ b/pyomo/solvers/plugins/solvers/CBCplugin.py
@@ -546,8 +546,16 @@ class CBCSHELL(SystemCallSolver):
                 elif constraint[:2] == 'r_':
                     # For the range constraints, supply only the dual with the largest
                     # magnitude (at least one should always be numerically zero)
-                    if abs(constraint_dual) > 0:
-                        #use same key for r_u_ and r_l_
+                    existing_constraint_dual_dict = solution.constraint.get( 'r_l_' + constraint[4:], None )
+                    if existing_constraint_dual_dict:
+                        # if a constraint dual is already saved, then update it if its
+                        # magnitude is larger than existing; this avoids checking vs
+                        # zero (avoiding problems with solver tolerances)
+                        existing_constraint_dual = existing_constraint_dual_dict["Dual"]
+                        if abs( constraint_dual) > abs(existing_constraint_dual):
+                            solution.constraint[ 'r_l_' + constraint[4:] ] = {"Dual": constraint_dual}
+                    else:
+                        # if no constraint with that name yet, just save it in the solution constraint dictionary
                         solution.constraint[ 'r_l_' + constraint[4:] ] = {"Dual": constraint_dual}
 
             elif processing_constraints is False:
--- a/pyomo/solvers/plugins/solvers/CBCplugin.py
+++ b/pyomo/solvers/plugins/solvers/CBCplugin.py
@@ -2,8 +2,8 @@
 #
 #  Pyomo: Python Optimization Modeling Objects
 #  Copyright 2017 National Technology and Engineering Solutions of Sandia, LLC
-#  Under the terms of Contract DE-NA0003525 with National Technology and 
-#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain 
+#  Under the terms of Contract DE-NA0003525 with National Technology and
+#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain
 #  rights in this software.
 #  This software is distributed under the 3-clause BSD License.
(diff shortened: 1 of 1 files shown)


ghackebeil:
I'm seeing a number of test failures pop up on this branch. The failures do not appear on Travis because we are not currently installing CBC there (I don't know if this is because we can't or if we just haven't yet). You can reproduce these failures locally by running the test file `pyomo/solvers/tests/checks/test_writers.py` on a machine that has CBC installed.

The way that test suite sets up tests is a little convoluted, but all of the test models as well as the solution baselines that we test against are in `pyomo/solvers/tests/models/`. You should be able to map any failures you see to a model located in that directory and investigate more closely my manually solving the model with CBC in a simple Pyomo script and checking the solution against the baseline.

rsmith54:
@ghackebeil thanks for the information, I'll look into it and see if I can figure out what's causing those failures.

rsmith54:
@ghackebeil I have fixed the failures in test_writers, while preserving the speed increase.  Please share any additional suggestions or changes.

ghackebeil:
Great! I can confirm tests are passing on my machine as well. I'll review the changes.

pyomo/solvers/plugins/solvers/CBCplugin.py
```diff
@@ -361,8 +361,11 @@ def process_logfile(self):
                 soln.objective['__default_objective__']['Value'] = float(tokens[2])
             if len(tokens) > 4 and tokens[0] == "Optimal" and tokens[2] == "objective":
                 # parser for log file generetated without discrete variable
-                soln.status = SolutionStatus.optimal
-                soln.objective['__default_objective__']['Value'] = float(tokens[4])
+                try:
```
ghackebeil: Is this try block needed or is it a remnant of debugging. If it's needed, can you add a bit of documentation about it, try to make the except statement more specific, and remove the print statement.

jsiirola:
FYI: Travis builds should have CBC (lp interface only) now.  We are using the conda-forge build, which at this moment, does not build the ASL interface.  Appveyor builds continue to not have CBC, as conda-forge only builds CBC for linux/osx.

pyomo/solvers/plugins/solvers/CBCplugin.py
rsmith54: Good catch, it is not needed.  I have removed that line.

ghackebeil:
I don't believe so. I'm fine with merging this as soon as another developer completes a review.

@jsiirola @whart222 @jwatsonnm  @blnicho @michaelbynum @carldlaird ?

pyomo/solvers/plugins/solvers/CBCplugin.py
```diff
@@ -533,14 +539,16 @@ def process_soln_file(self, results):
...
+                if constraint[:2] == 'c_':
                     solution.constraint[constraint] = {"Dual" : constraint_dual}
-                elif constraint.startswith('r_l_'):
-                    range_duals.setdefault(constraint[4:],[0,0])[0] = constraint_dual
-                elif constraint.startswith('r_u_'):
-                    range_duals.setdefault(constraint[4:],[0,0])[1] = constraint_dual
+                elif constraint[:2] == 'r_':
+                    # For the range constraints, supply only the dual with the largest
+                    # magnitude (at least one should always be numerically zero)
+                    if abs(constraint_dual) > 0:
```
jsiirola: I worry about this causing problems with solver tolerances.  My understanding of the previous code is that by deferring this until the end of processing, we wouldn't "accidentally" report very small duals if, say the solver returned `r_l_` of 500 and `r_u_` as 1e-8.  The code here could return 1e-8 as the dual value.  Maybe only update the dual on line 551 if the current constraint_dual value is larger?

ghackebeil: That's a good catch. I agree that this needs to be fixed.

rsmith54: @ghackebeil @jsiirola I have pushed an additional

whart222:
I'm happy to merge this PR after confirming that tests pass.

PR #314, opened 2018-01-18
Remove API url from GAMS solver docstring

gseastream:
The GAMS online docs changed the appropriate url to lead to installation help for the Python API, so I figured it would be better to just make the docstring generic in case they change it again in the future. Very minor change.

### DIFF
--- a/pyomo/solvers/plugins/solvers/GAMS.py
+++ b/pyomo/solvers/plugins/solvers/GAMS.py
@@ -43,7 +43,7 @@ class GAMSSolver(pyomo.util.plugin.Plugin):
 
     Pass solver_io keyword arg to SolverFactory to choose solver mode:
         solver_io='direct' or 'python' to use GAMS Python API
-            Requires installation, visit online docs for help.
+            Requires installation, visit https://www.gams.com for help.
         solver_io='shell' or 'gms' to use command line to call gams
             Requires the gams executable be on your system PATH.
     """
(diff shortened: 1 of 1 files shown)


pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -43,8 +43,7 @@ class GAMSSolver(pyomo.util.plugin.Plugin):
 
     Pass solver_io keyword arg to SolverFactory to choose solver mode:
         solver_io='direct' or 'python' to use GAMS Python API
-            Requires installation, visit this url for help:
-            https://www.gams.com/latest/docs/apis/examples_python/index.html
+            Requires installation, visit online docs for help.
```
qtothec: visit online docs at `http://www.gams.com`? Otherwise they might look at Pyomo online docs. Same with the docstring under `solve`.

gseastream: The correct url at this point would be `https://www.gams.com/latest/docs/API_PY_TUTORIAL.html`, so perhaps `https://www.gams.com/latest/docs/` would be a little more desciptive but still defend against changes. I'm also just not sure how much to expect them to change it again.

qtothec: Ultimately, that's up to the GAMS folks, but as long as our users get a good general sense of where to go, I think that it's sufficient.

PR #337, opened 2018-02-02
GDPopt solver initial implementation

qtothec:
Moving from https://github.com/emma58/pyomo/pull/7

Initial version of the GDPopt solver into contrib. Code is still a bit messy.


jsiirola:
@qtothec: two quick notes: You use non-2.6 syntax throughout this PR:
  - `{a:b for a,b in ...}` --> `dict((a,b) for a,b in ...)` or `dict(x for x in ...)` (where x is a tuple)
  - `" ... {} ...".format(x)` is not legal.  I would use `" ... %s ..." % (x,)`: is it more portable, and is generally faser than `.format()`

We are probably dropping 2.6 support *after* this release, but if you want the PR in for this release, we need to fix this.

pyomo/contrib/gdpopt/GDPopt.py
```diff
@@ -0,0 +1,1467 @@
...
+        return solnFeasible
+
+    def _is_feasible(self, m, constr_tol=1E-6, var_tol=1E-8):
+        for constr in m.component_data_objects(
+                ctype=Constraint, active=True, descend_into=(Block, Disjunct)):
+            # constraint is an equality
+            if constr.equality:
+                if fabs(value(constr.lower) -
+                        value(constr.body)) >= constr_tol:
+                    logger.info('%s: %s ≠ %s' % (
```
jsiirola: It would be significantly better if you did not include unicode characters (not equal to) in the code base.

pyomo/contrib/gdpopt/GDPopt.py
```diff
@@ -0,0 +1,1467 @@
+# -*- coding: UTF-8 -*-
```
jsiirola: emacs complains when "UTF" is in all-caps.

pyomo/contrib/gdpopt/GDPopt.py
qtothec: Ruins the one-character symmetry with < and > to use != or the like, though

jsiirola: You could always use `!` if you really want a single-character operator.

qtothec:
Well, the only thing failing now is diff coverage. How aggressively do you want me to pursue that?

jsiirola:
OK - the test coverage is bad becasue we don't have commercial solvers out on Travis or Appveyor.  We *do* have CBC on Travis, though.  I just spot-checked it, and it appears to work.  We'll see how that improves coverage.

jsiirola:
To answer your original question, we should strive for good code coverage (I have a dream to get things above 90%).  That said, I don't see a reason to hold up a contrib package solely on the basis of coverage, *except* that the package must have *some* coverage so that we know when it starts failing.  Assuming that 7cfc5fe gets the coverage to some reasonable level, then I will merge it.

PR #339, opened 2018-02-07
adding additional check to skip complementarity gap lines

rsmith54:
## Fixes # .

When running cbc with the barrier method, you can get lines like : 
```Optimal - objective gap and complementarity gap```
which are tripped by the condition below in the PR.

## Summary/Motivation:

Fix for this input, as it causes a ValueError (```float("and")```).

Adds a check on "and" to skip these lines.

### Legal Acknowledgement

### DIFF
--- a/pyomo/solvers/plugins/solvers/CBCplugin.py
+++ b/pyomo/solvers/plugins/solvers/CBCplugin.py
@@ -363,7 +363,7 @@ class CBCSHELL(SystemCallSolver):
                 soln.objective['__default_objective__']['Value'] = float(tokens[2])
             if len(tokens) > 4 and tokens[0] == "Optimal" and tokens[2] == "objective" and tokens[4] != "and":
                 # parser for log file generetated without discrete variable
-                # last check avoids lines like "Optimal - objective gap and complementarity gap both smallish and small steps"
+                # see pull request #339: last check avoids lines like "Optimal - objective gap and complementarity gap both smallish and small steps"
                 soln.status = SolutionStatus.optimal
                 soln.objective['__default_objective__']['Value'] = float(tokens[4])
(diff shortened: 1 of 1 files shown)


qtothec:
Could you also provide a small test file to demonstrate this in action?

rsmith54:
A small LP problem? I actually found this is in a large problem I was working on, unfortunately.  I could attach the log file that is being parsed if that's useful.

rsmith54:
I'm not sure how to create a file demonstrating this behavior.

jsiirola:
On the surface, this is a pretty innocuous PR, and we should be able to merge it pretty quickly.  That said, there are two things I would like to see addressed:
1. that chunk of code reads the objective value back from CBC.  When CBC issues lines like "`Optimal - objective gap and complementarity gap`", is there another line that contains the objective value?  If not, how is the objective value returned?
2. as @qtothec hinted at: Can you add a test that exercises this path in the code?  Minimal is of course most welcome, but *any* test is better than no test.  Maybe someone from CBC has a smallish example that reproduces that output?  @tkralphs?

rsmith54:
@jsiirola 

1. The objective value will show up in a later line, following the same format as before (so this will just skip a line with that pattern).
2. Sure, I'll try and add a test ASAP.

jsiirola:
@rsmith54: great! thanks!  Re (1), I am trying to get into a better habit of documenting the logic in the code.  Would you mind adding to your comment indicating that the objective value will be picked up on a subsequent line?

whart222:
@rsmith54 What is your timeline for adding a test?  Our resolution is waiting on that.

rsmith54:
Hi @whart222, I'll try and look into it later this week.  Thanks for the reminder.

PR #342, opened 2018-02-09
[WIP] New pysp_instance_creation_callback signature

ghackebeil:
## Summary/Motivation:

This commit changes the signature for the pysp_instance_creation_callback function from
`pysp_instance_creation_callback(scenario_name, node_name_list)`
to
`pysp_instance_creation_callback(caller_data, scenario_tree, scenario_name, node_name_list)`

We are currently proposing that `caller_data` be the scenario tree argument passed to `ScenarioTreeInstanceFactory`, which can be a .dat filename string, a Pyomo scenario tree model, a networkx tree, or a PySP ScenarioTree object, and that the `scenario_tree` argument be the current PySP ScenarioTree object that instances are being requested for.

This will provide users with more control during scenario instance creation by allowing them to send data into this function by placing it on the scenario tree argument given to the `ScenarioTreeInstanceFactory` (e.g., when it is not simply a string).

### Legal Acknowledgement


jsiirola:
Is there a rationale for the argument order?  If you were to make it instead
```
pysp_instance_creation_callback(scenario_name, node_name_list, caller_data, scenario_tree)
```
it might be easier to both support deprecation / graceful fallback to previous behavior, and examples that work for both old- and new-style APIs

ghackebeil:
@jsiirola: I think that makes sense

PR #354, opened 2018-02-16
GDP rewrite

### LINKED ISSUE #318: gdp.chull cannot handle named expressions
The gdp.chull plugin fails to handle named expressions (class Expression) in disjuncts, at least if the expressions are indexed. `expr.type()` in gdp/plugins/chull.py:411 raises AttributeError because _GeneralExpressionData does not have a type attribute. The whole branch seems superfluous because the following else branch also throws, albeit a different exception. However, I suppose really fixing this would require creating disaggregated expressions.

jsiirola:
## Summary/Motivation:
This is a rewrite of the `pyomo.gdp` package.

- Rewrite of the `Disjunct` and `Disjunction` components to bring them in line with current Component conventions.
- Adding more examples
- Rewrites of the bigm and chull transformations
- Adding a cuttingplane reformulation
- Adding GDPopt solver (to `pyomo.contrib`)

### Legal Acknowledgement


pyomo/gdp/plugins/bigm.py
```diff
@@ -2,307 +2,660 @@
...
+        suffix_list = []
+        while block is not None:
+            bigm = block.component('BigM')
+            if type(bigm) is Suffix:
+                suffix_list.append(bigm)
+            block = block.parent_block()
+        return suffix_list
+
+    def _apply_to(self, instance, targets=None, **kwds):
+        config = self.CONFIG().set_value(kwds.pop('options', {}))
```
qtothec: I'm very confused how this is passing tests, because on my system, `set_value()` has no return value (so it's always `None`), which causes `config = None`, which causes issues when you try to do `config.set_value(kwds)` below.

pyomo/gdp/plugins/bigm.py
```diff
@@ -2,307 +2,660 @@
...
+        'src': <source disjunct>
+        'srcConstraints': ComponentMap(relaxed_constraint: constraint)
+    """
+
+    alias('gdp.bigm', doc=textwrap.fill(textwrap.dedent(__doc__.strip())))
+
+    CONFIG = ConfigBlock("gdp.bigm")
+    CONFIG.declare('bigM', ConfigValue(
+        default=None,
+        domain=_to_dict,
```
qtothec: What does `domain` mean with respect to a `ConfigBlock`?

pyomo/gdp/plugins/bigm.py
jsiirola: You need the current pyutilib master.

pyomo/gdp/plugins/bigm.py
jsiirola: Note that the `domain` is passed to the `ConfigValue` constructor.  `domain` is a callable object (could be a type, function, or functor).  When a value is set, the incoming value is passed as the single argument to `__call__` and the return value is what is stored in the `ConfigValue`.  At it's simplest, it can be a type like `int` so that all incoming values are automatically cast to `int`.  In this case it is a little more complicated, as it accepts dicts or "values" (int/float/tuple).  The former is preserved and the latter is cast to `{None: val}`.

To answer your question, `ConfigBlocks` can not have domains, although they accept an `implicit_domain`, which is used when the `ConfigBlock` accepts undeclared (implicit) keys - in which case, any implicit values are passed through the `implicit_domain` callable.

pyomo/gdp/plugins/bigm.py
qtothec: Ah, yes. That seems to have done the trick.

examples/gdp/ConstrainedLayout.py
```diff
@@ -0,0 +1,151 @@
...
+
+# generate the list of possible pairs of boxes
+def box_pairs_filter(model, i, j):
+    return i < j
+model.BoxPairs = Set(initialize=model.BOXES * model.BOXES, dimen=2,
+                     filter=box_pairs_filter)
+
+# bounds for box coordinates
+# TODO: What are they trying to accomplish here? Why is it always index
+# 2 for the first two??
```
qtothec: This is an error in the original GAMS file. The intent is to find the min and max x and y coordinates that the boxes can have, based on the circle locations and radii. The code as written works because circle 2 has the higher x and y coordinate values.

examples/gdp/ConstrainedLayout.py
```diff
@@ -0,0 +1,151 @@
...
+                                model.yCoord[j] - model.Height[j]/2)
+    else:
+        raise RuntimeError("Unrecognized box relationship: %s" % boxRelation)
+model.no_overlap_disjunct = Disjunct(model.BoxPairs, model.BoxRelations,
+    rule=no_overlap_disjunct_rule)
+
+def no_overlap_rule(model, i, j):
+    return [model.no_overlap_disjunct[i, j, direction] for direction in \
+            model.BoxRelations]
+model.no_overlap_disjunction = Disjunction(model.BoxPairs, rule=no_overlap_rule)
```
qtothec: Are we going to support the shortcut notation?
```
@model.Disjunction(model.BoxPairs)
def no_overlap_rule(model, i, j):
    ...
```

Similarly with Disjunct.

examples/gdp/batchProcessing.py
```diff
@@ -0,0 +1,230 @@
...
+in between. We decide the number and volume of units, and the volume and location of the storage
+tanks. The problem is convexified and has a nonlinear objective and global constraints
+
+NOTE: When I refer to 'gams' in the comments, that is Batch101006_BM.gms for now. It's confusing
+because the _opt file is different (It has hard-coded bigM parameters so that each constraint 
+has the "optimal" bigM).'''
+
+model = AbstractModel()
+
+# TODO: it looks like they set a bigM for each j. Which I need to look up how to do...
```
qtothec: `j`'s are stages?

examples/gdp/batchProcessing.py
```diff
@@ -0,0 +1,230 @@
...
+StorageTankSizeFactorByProd = 3
+MinFlow = -log(10000)
+VolumeLB = log(300)
+VolumeUB = log(3500)
+StorageTankSizeLB = log(100)
+StorageTankSizeUB = log(15000)
+UnitsInPhaseUB = log(6)
+UnitsOutOfPhaseUB = log(6)
+# TODO: YOU ARE HERE. YOU HAVEN'T ACTUALLY MADE THESE THE BOUNDS YET, NOR HAVE YOU FIGURED OUT WHOSE
+# BOUNDS THEY ARE. AND THERE ARE MORE IN GAMS.
```
qtothec: Seems unfinished?

examples/gdp/batchProcessing.py
```diff
@@ -0,0 +1,230 @@
...
+
+##########
+# Sets
+##########
+
+model.PRODUCTS = Set()
+model.STAGES = Set(ordered=True)
+model.PARALLELUNITS = Set(ordered=True)
+
+# TODO: this seems like an over-complicated way to accomplish this task...
```
qtothec: Given what's written, the filter seems reasonable.

examples/gdp/batchProcessing.py
```diff
@@ -0,0 +1,230 @@
...
+                                    for i in model.PRODUCTS)
+model.finish_in_time = Constraint(rule=finish_in_time_rule)
+
+
+###############
+# Disjunctions
+###############
+
+def storage_tank_selection_disjunct_rule(disjunct, selectStorageTank, j):
+    model = disjunct.model()
```
qtothec: this line is not strictly necessary? :0)

examples/gdp/batchProcessing.py
```diff
@@ -0,0 +1,230 @@
...
+        # disjunct.no_volume = Constraint(expr=model.storageTankSize_log[j] == MinFlow)
+        disjunct.no_batch = Constraint(model.PRODUCTS, rule=no_batch_rule)
+model.storage_tank_selection_disjunct = Disjunct([0,1], model.STAGESExceptLast, 
+                                       rule=storage_tank_selection_disjunct_rule)
+
+def select_storage_tanks_rule(model, j):
+    return [model.storage_tank_selection_disjunct[selectTank, j] for selectTank in [0,1]]
+model.select_storage_tanks = Disjunction(model.STAGESExceptLast, rule=select_storage_tanks_rule)
+
+# though this is a disjunction in the GAMs model, it is more efficiently formulated this way:
```
qtothec: why?

pyomo/core/plugins/transform/add_slack_vars.py
```diff
@@ -0,0 +1,93 @@
...
+from nose.tools import set_trace
+
+class AddSlackVariables(NonIsomorphicTransformation):
+    """
+    This plugin adds slack variables to every constraint or to the constraints
+    specified in targets
+    """
+
+    alias('core.add_slack_variables', \
+          doc="Create a model where we had slack variables to every constraint "
```
qtothec: typo. "where we add"

pyomo/gdp/disjunct.py
```diff
@@ -2,83 +2,111 @@
...
+        if type(arg) in native_types:
+            return (_Initializer.value, bool(arg))
+        elif type(arg) is types.FunctionType:
+            return (_Initializer.function, arg)
+        elif isinstance(arg, ComponentData):
+            return (_Initializer.deferred_value, arg)
+        elif hasattr(arg, '__getitem__'):
+            return (_Initializer.dict_like, arg)
+        else:
+            # Hopefully this thing is castable to teh type that is desired
```
qtothec: typo: "the"

pyomo/gdp/disjunct.py
```diff
@@ -2,83 +2,111 @@
...
+your rule.
+"""
 
 class GDP_Error(Exception):
     """Exception raised while processing GDP Models"""
 
 
+# THe following should eventually be promoted so that all
+# IndexedComponents can use it
+class _Initializer(object):
```
qtothec: Some short description?

pyomo/core/base/block.py
```diff
@@ -673,8 +673,8 @@ def __delattr__(self, name):
             super(_BlockData, self).__delattr__(name)
 
     def set_value(self, val):
-        for k,v in iteritems(getattr(self, '_decl', {})):
-            super(_BlockData, self).__delattr__(k)
+        for k in list(getattr(self, '_decl', {})):
+            self.del_component(k)
```
carldlaird: What was the reason for this change?

pyomo/core/plugins/transform/__init__.py
```diff
@@ -9,15 +9,16 @@
 #  ___________________________________________________________________________
 
 import pyomo.core.plugins.transform.relax_integrality
-#import pyomo.core.plugins.transform.eliminate_fixed_vars
-#import pyomo.core.plugins.transform.standard_form
+# import pyomo.core.plugins.transform.eliminate_fixed_vars
```
carldlaird: What is the reason for these commented out imports?

pyomo/environ/__init__.py
```diff
@@ -38,12 +38,13 @@ def _do_import(pkg_name):
...
 #
 # These packages also contain plugins that need to be loaded, but
-# we silently ignore any import errors because these 
+# we silently ignore any import errors because these
 # packages are optional and/or under development.
 #
 _optional_packages = set([
     'pyomo.contrib.example',
-    'pyomo.contrib.preprocessing'])
+    'pyomo.contrib.preprocessing',
```
carldlaird: What are these _optional_packages used for?

pyomo/gdp/plugins/bilinear.py
```diff
@@ -20,7 +20,7 @@
 from pyomo.gdp import *
 from pyomo.repn import generate_canonical_repn
 
-logger = logging.getLogger('pyomo.core')
+logger = logging.getLogger('pyomo.gdp')
```
carldlaird: Great.

pyomo/mpec/plugins/solver2.py
```diff
@@ -39,7 +39,7 @@ def _apply_solver(self):
         xfrm.apply_to(self._instance)
 
         xfrm = TransformationFactory('gdp.bigm')
-        xfrm.apply_to(self._instance, default_bigM=self.options.get('bigM',10**6))
+        xfrm.apply_to(self._instance, bigM=self.options.get('bigM',10**6))
```
carldlaird: Totally minor, but why 10**6 vs 100000 used in an earlier file as the default? (1) I think it was 1e6 vs 1e5, but I am not sure. Also, what is wrong with 1e5? or 1e6?

pyomo/mpec/tests/t11_mpec.simple_disjunction.txt
```diff
@@ -50,7 +50,7 @@ cc : Size=1, Index=None, Active=True
 
     1 Disjunction Declarations
         complements : Size=1, Index=None, Active=True
-            Key  : Lower : Body                                                                     : Upper : Active
-            None :   1.0 : cc.expr1.indicator_var + cc.expr2.indicator_var + cc.expr3.indicator_var :   1.0 :   True
+            Key  : Disjuncts                            : Active : XOR
+            None : ['cc.expr1', 'cc.expr2', 'cc.expr3'] :   True : True
```
carldlaird: This move to list from a sum is expected?

pyomo/solvers/plugins/solvers/CPLEX.py
```diff
@@ -159,7 +159,7 @@ def _warm_start(self, instance):
             mst_file.write("<header/>\n")
             mst_file.write("<quality/>\n")
             mst_file.write("<variables>\n")
-            for var in instance.component_data_objects(Var, active=True):
+            for var in instance.component_data_objects(Var):
```
carldlaird: Why do we need to loop through variables that are not active?

pyomo/solvers/plugins/solvers/GUROBI.py
```diff
@@ -179,7 +179,7 @@ def _warm_start(self, instance):
             smap = instance.solutions.symbol_map[self._smap_id]
         byObject = smap.byObject
         with open(self._warm_start_file_name, 'w') as mst_file:
-            for vdata in instance.component_data_objects(Var, active=True):
+            for vdata in instance.component_data_objects(Var):
```
carldlaird: See above.

pyomo/solvers/plugins/solvers/glpk_direct.py
```diff
@@ -139,7 +139,7 @@ def _populate_glpk_instance ( self, model ):
         if objective.is_minimizing(): sense = GLP_MIN
 
         constraint_list = model.component_map(Constraint, active=True)
-        variable_list   = model.component_map(Var, active=True)
+        variable_list   = model.component_map(Var)
```
carldlaird: Same as above.

pyomo/core/plugins/transform/__init__.py
qtothec: I think at some point some of the transformations were deprecated and thus commented out. This PR doesn't change that fact, other than a cosmetic addition of a space after the `#`.

pyomo/environ/__init__.py
qtothec: The `_optional_packages` import the necessary contrib modules, such as the transformation plugins that I have under `pyomo.contrib.preprocessing`. The comment lines above describe its functionality.

pyomo/mpec/tests/t11_mpec.simple_disjunction.txt
qtothec: It's a move from a `Constraint-like` to a list, with a flag for whether it's an XOR disjunction vs. an OR disjunction. It introduces support for OR disjunctions, whereas the old approach only supported XOR disjunctions.

pyomo/core/plugins/transform/__init__.py
carldlaird: OK, I see. In my view it looked like they were added and then commented out.

pyomo/environ/__init__.py
carldlaird: OK, the comment is confusing since it says they "need to be loaded", but also says that they "are optional". No biggie. I am fine with it.

examples/gdp/batchProcessing.py
emma58: Yes, j's are stages. I think the gams file uses 3 different Big-M values, one for each disjunction. They are indexed by stage, but I don't think they actually depend on it. I never got as far as making my Big-M values match those ones.

examples/gdp/batchProcessing.py
emma58: Yes, that is likely... I ran into some mistakes in the gams file for this problem, and so I know I wasn't ever able to match their results for this one.

examples/gdp/batchProcessing.py
emma58: @jsiirola's thought was that in both these disjunctions you are just choosing a value for unitsInPhase and unitsOutOfPhase, so you could just formulate it that way rather than having the disjunctions. But I guess that might make it less useful as a GDP test case?

examples/gdp/ConstrainedLayout.py
jsiirola: Yes.  That notation should (already) be supported.

pyomo/core/plugins/transform/__init__.py
jsiirola: I believe that @whart222 "deprecated" them by removing their registration and seeing if anyone complained.

pyomo/mpec/plugins/solver2.py
jsiirola: I believe that @whart222 wrote this code and that was his style.  I agree that 1e6 would be better.

pyomo/solvers/plugins/solvers/CPLEX.py
jsiirola: It turns out that Vars aren't active (they are not ActiveComponents and cannot be deactivated).  The active flag only controlled if the generator descended into all blocks or only active blocks.  This change allows the generator to descend into all blocks - not just the active ones.

carldlaird: Right. (We just had this conversation :)

PR #357, opened 2018-02-22
cplex 12.6 raises an error if you call objective.get_quadratic() when…

michaelbynum:
… the objective is linear

## Summary/Motivation:
Tests were failing for Cplex 12.6

- Use pyomo to check the degree of the objective instead of cplex
-

### Legal Acknowledgement

### DIFF
--- a/pyomo/solvers/plugins/solvers/cplex_direct.py
+++ b/pyomo/solvers/plugins/solvers/cplex_direct.py
@@ -108,7 +108,11 @@ class CPLEXDirect(DirectSolver):
         if self._keepfiles:
             print("Solver log file: "+self._log_file)
 
-        if len(self._solver_model.objective.get_quadratic()) != 0:
+        obj_repn = generate_canonical_repn(self._objective.expr)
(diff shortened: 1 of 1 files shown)


pyomo/solvers/plugins/solvers/cplex_direct.py
```diff
@@ -108,7 +108,11 @@ def _process_stream(arg):
         if self._keepfiles:
             print("Solver log file: "+self._log_file)
 
-        if len(self._solver_model.objective.get_quadratic()) != 0:
+        obj_repn = generate_canonical_repn(self._objective.expr)
+        obj_degree = canonical_degree(obj_repn)
```
jsiirola: This seems like an awful lot of work.  Why not just:
```
obj_degree = self._objective.expr.polynomial_degree()
if obj_degree is None or obj_degree > 2:
    raise #...
```
(Note the check for None - in Python 3, comparing None with an `int` raises an exception)

michaelbynum: Good suggestion! Thanks John!

PR #361, opened 2018-02-23
Persistent solvers

michaelbynum:
## Summary/Motivation:
get_stats method does not work with cplex12.6 and python2.7.14

-
-

### Legal Acknowledgement

### DIFF
--- a/pyomo/solvers/plugins/solvers/cplex_direct.py
+++ b/pyomo/solvers/plugins/solvers/cplex_direct.py
@@ -547,23 +547,11 @@ class CPLEXDirect(DirectSolver):
             soln.gap = None
 
         self.results.problem.name = gprob.get_problem_name()
-        stats = gprob.get_stats()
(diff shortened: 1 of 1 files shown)


jsiirola:
Sorry - I missed that it was the `get_stats()` method that was failing and not the transfer of fields.  I still think that a try-except block would be a better solution so that "most" users would be able to get the stats information.

Maybe also include a warning that `get_stats` was not available?

michaelbynum:
@jsiirola The information being added to the results object (which I just deleted) is actually new functionality that, most likely, no one even knows about. This information could be valuable to users. However, the results object may not be the right place to put it anyway. I propose we create methods on the persistent solvers to query information like this.

jsiirola:
@michaelbynum That seems reasonable.  The idea is that sometime in the future we would add a "getter" method similar to what exists for the Gurobi interface?

PR #363, opened 2018-02-26
Fix to gurobi_direct initialization when gurobipy is not available.

ghackebeil:
In this case, the solver version was None and it was being compared to an integer, leading to a confusing warning message (probably for Python 3+ only) when calling SolverFactory that looked something like:

WARNING: Gurobi direct solver plugin could not be created: '<' is invalid for 'Nonetype' and int

-
-

### Legal Acknowledgement


michaelbynum:
Just for good practice, I am going to wait until the tests finish to merge this.

PR #366, opened 2018-02-27
Remove checks for is_indexed from PersistentSolver methods

ghackebeil:
Change methods such as {add,remove,update}_{var,constraint,block} on PersistentSolver to no longer check for indexed objects. That is, `solver.add_var(<singleton_variable>)` continues to work, but `solver.add_var(<indexed_variable>)` no longer works.

## Fixes # .

Allows all methods to work with objects from pyomo.kernel that do no support the same indexed API. In particular, indexed objects in the kernel do not have an is_indexed method, and only singleton objects from the kernel that inherit from NumericValue have an is_indexed method that returns False (i.e., single and indexed constraints do not have is_indexed methods).

Additionally, the current implementations of these PersistentSolver methods use .values() to iterate over indexed objects, but this would not work with all container implementations in kernel (e.g., list-like containers).

## Summary/Motivation:

This removes trivial functionality from this API before it appears in its first release, while expanding the set of modeling objects this API is compatible with.

-
-

### Legal Acknowledgement

### DIFF
--- a/pyomo/solvers/plugins/solvers/cplex_persistent.py
+++ b/pyomo/solvers/plugins/solvers/cplex_persistent.py
@@ -84,6 +84,9 @@ class CPLEXPersistent(PersistentSolver, CPLEXDirect):
         ----------
         var: Var
         """
+        # see PR #366 for discussion about handling indexed
+        # objects and keeping compatibility with the
+        # pyomo.kernel objects
         #if var.is_indexed():
         #    for child_var in var.values():
         #        self.compile_var(child_var)
--- a/pyomo/solvers/plugins/solvers/cplex_persistent.py
+++ b/pyomo/solvers/plugins/solvers/cplex_persistent.py
@@ -84,10 +84,10 @@ class CPLEXPersistent(PersistentSolver, CPLEXDirect):
         ----------
         var: Var
         """
-        if var.is_indexed():
-            for child_var in var.values():
-                self.compile_var(child_var)
-            return
+        #if var.is_indexed():
+        #    for child_var in var.values():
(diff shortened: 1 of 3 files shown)


jsiirola:
Thanks, @ghackebeil!  This got me thinking about documentation... @michaelbynum: do you think we should just update the docstrings to indicate that the `add_*` methods add *data* objects and not components (as they imply now), or should we go as far as renaming the methods to `add_*_data`?

Separately (and probably either in an issue or a separate PR), the `_add_*` methods should almost certainly be renamed `_add_*_data`.

ghackebeil:
I am against using the name `data` in methods or documentation to mean anything other than the data we use to instantiate a model.

jsiirola:
What alternative name would you use to distinguish between the container and the thing contained?  I know in Kernel you favored "FooDict" and "Foo" -- the problem is that for most of our users, a "variable" can be scalar, vector, or matrix - so I feel that we are somewhat stuck with things like "Var()" being the delegator (@carldlaird does raise an interesting question as to if it needs to be a class, or if a function would be more appropriate), and then needing alternative names for the container, the thing contained, and potentially the scalar thing.

Core has been using "Data" for a long time, and that usage has crept into the public methods (e.g., `component_data_objects`).

carldlaird:
I agree with both of you. I was confused by "Data" initially. I would have thought VarItem or VarEntry or VarElement or something would have been better (if you want to keep Var as the container name). Having said that, Data has permeated through the code.

ghackebeil:
@jsiirola or @michaelbynum: Mind refreshing your review so I can merge this?

Issue #371, opened 2018-02-27
Fix confusing semantics of model.solutions.load_from


ghackebeil:
See #365 and #370

ghackebeil:
Options to consider:
 - Tag the results object after the solution was taken out of it by `model.solutions.load_from`, so that if a user attempts to pass it to `model.solutions.load_from` again, we can provide a helpful exception.
 - Change `model.solutions.load_from` to leave the solution in the results object after loading it.

michaelbynum:
This should be resolved in `contrib.solver`.

PR #372, opened 2018-02-28
Do not silently fail to load a solution for kernel models

ghackebeil:
### Fixes ###

The following should not run without _some_ exception or warning (but it does).

```python
import pyomo.kernel as pmo
model = pmo.block()
model.x = pmo.variable()
model.o = pmo.objective(model.x)
model.c = pmo.constraint(model.x >= 1)

opt = pmo.SolverFactory("cplex")
opt.options['sadfasdfasdfkl;'] = 'totally_not_an_option'
results = opt.solve(model)
```
If you run this with tee=True, you will see that Cplex just pukes.

This PR changes the behavior (for Kernel models only), so that when `opt.solve(...)` is called with load_solutions=True (the default), an error message will be logged when no solution is found in the results object.

I strongly believe this change should be made for AML models as well, but this change would cascade into how the semantics of `model.solutions.load_from` are defined. I think Kernel is a good testing ground for new behavior such as this, and can provide a good working example for discussions about changes to the behavior of `model.solutions.load_from` (#371).

I also think this change affects a small enough audience that it can make it into the Pyomo 5.4 release.

### Legal Acknowledgement


pyomo/opt/base/solvers.py
```diff
@@ -649,6 +648,18 @@ def solve(self, *args, **kwds):
...
+                       (len(result.solution) == 0):
+                        raise ValueError(
+                            "No solution is available. Try calling "
+                            "solve with the 'load_solutions' keyword "
+                            "set to False. Then, check the statuses "
+                            "on the results object before attempting "
+                            "to load the solution (if one is available). "
+                            "Additionally, one can call solve with the "
+                            "'tee' keyword set to True to view the solver "
+                            "output.")
```
jsiirola: Should this be an exception or just an Error logged to the logger?  I could see users just wanting an error (after all, they could be about to check the results object for the solver status anyway).  As we move forward, we should be actively *discouraging* users from requesting the solution in the results object: that is a very slow data structure to generate, and for most situations, a pointless step.  The way this is written, if the user *ever* suspected that the model might fail, then they will be forced to specify `load_solutions=False`...

ghackebeil:
In every advanced scripting application I’ve written recently, I always use load_solutions=False because I need to check statuses before I let values on the model get potentially changed to garbage (which can happen with the NL interface) and I also want to minimize the output I don’t have control over (e.g., Pyomo logging warnings). But I agree that we should discuss this more as we fix this for the AML and make behavior consistent across the two modeling frameworks. I’m happy to stick with an exception until we begin that discussion.

pyomo/opt/base/solvers.py
```diff
@@ -640,7 +640,6 @@ def solve(self, *args, **kwds):
                             self._default_variable_value
                         if self._load_solutions:
                             _model.load_solution(result.solution(0))
-                            result.solution.clear()
```
michaelbynum: @ghackebeil If you delete this line, the solution will never be removed from the results object, right? Maybe the ValueError you added below should fit inside this if statement before "result.solution.clear()"?

ghackebeil: My intention was to leave the solution in the results object either way. I think we are confusing users when they see a results object with no solution on it, but I agree that we should revisit this as part of a larger design discussion very soon.

I also did not want raise the ValueError before executing the next few lines below this one, where I clean the symbol map off of the model (placed there by the solver interface or writer).

This is all a mess write now (for both AML and Kernel), I'm just trying to provide a working alternative to base future discussions on.

I am okay with someone re-adding `result.solution.clear()` in an `else` statement after the block where I raise the exception if they have a strong opinion about this.

michaelbynum:
I think I am in favor or raising an error unless the user specifies load_solutions=False.

pyomo/opt/base/solvers.py
michaelbynum: Good points. Maybe it does make sense to leave the solutions on the results object. A larger design discussion is definitely warranted - which is really related to refactoring the solver interfaces. I am fine with this pull request as is for now.

jsiirola:
@michaelbynum: to clarify, are you in favor of an error or an exception?  I think I prefer the former.

pyomo/opt/base/solvers.py
jsiirola: For the time being, I am OK with leaving the solution on the results object - as that has burned users in the past ("Why am I not getting the solution back from the solver?").  But @ghackebeil is right: this needs to be a larger design discussion as part of the writer/solver API rework that I hope will happen this summer: putting the solution into the results object is pretty slow and probably shouldn't be done unless a user specifically asks for it (it should be much more efficient to directly parse the solution into the model).  The "number of solutions" value in the results object should be the number returned from the solver and not the number stored in the results object.  Finally, we should re-think how the solutions are stored in the results object and use something more efficient than the current mechanism.

michaelbynum:
@jsiirola Sorry - I prefer an exception - primarily because logging is not turned on by default, and I think users should be notified when they try to load a solution after a solver does not converge.

ghackebeil:
@jsiirola: I think the language you are using is easy to confuse. Perhaps 'error message' would make it more clear what you mean. I commonly use the terms 'error' and 'exception' interchangeably (both in the context of raising something that stops the program). I admit, though, that my usage my be incorrect.

jsiirola:
Logging *is* turned on by default.  Also note that the problem with the exception is that the `solve()` method exits prematurely.  In this case, it's not too significant (I think that the only thing that is missed is setting / reporting the `postsolve_completion_time` ... but note that the results object is not returned, so there is nothing that the user can do to diagnose the solve.  Also not that this effectively *requires* the user to turn `load_solutions` off if there is the slightest chance that the solve fails, which it not the long-term direction that I think we should push users to.

pyomo/opt/base/solvers.py
ghackebeil: I agree on all counts. I like what Michael is doing in the persistent interface where you can tell the solver to load the solution (possibly partially) from the most recent solve in a separate step. Perhaps this could be generalized for non-persistent interfaces as a way to avoid a temporary solutions object, while still allow one to delay loading until after statuses are check. Another idea would be to allow users to provide a callback to the solve method that indicates whether or not they want the solution loaded based on statuses that are placed on the results object.

jsiirola:
Also, as a matter of practice: @ghackebeil can you add a test that exercises the exception?  (That would improve the code coverage)

ghackebeil:
I've started this locally, but I've reached my quota for the day. I'll push this later in the week when I can find time to finish it. I'd still like to get this PR into the release though.

michaelbynum:
My mistake. You are right. I agree. This should be an error rather than an exception.

Michael

pyomo/solvers/tests/checks/test_no_solution_behavior.py
```diff
@@ -0,0 +1,110 @@
...
+
+# The test directory
+thisDir = os.path.dirname(os.path.abspath( __file__ ))
+
+# Cleanup Expected Failure Results Files
+_cleanup_expected_failures = True
+
+#
+# A function that function that returns a function that gets
+# added to a test class.
```
jsiirola: Can you clarify this?  I feel that there is a typo here, but the logic is sufficiently complex that I'm not sure...

pyomo/solvers/tests/checks/test_no_solution_behavior.py
```diff
@@ -0,0 +1,110 @@
...
+        # solve
+        load_solutions = True
+        symbolic_labels = False
+        opt, results = model_class.solve(
+            solver,
+            io,
+            test_case.testcase.io_options,
+            test_case.testcase.options,
+            symbolic_labels,
+            load_solutions)
```
jsiirola: This "test" worries me, as it does not assert anything.  Can you wrap this in a LoggingIntercept context manager to ensure that the expected error is emitted?

pyomo/solvers/tests/checks/test_pickle.py
```diff
@@ -116,6 +116,7 @@ def failing_writer_test(self):
 
     return pickle_test
 
+cls = None
```
jsiirola: Is this line needed? ...or is this just defensive programming due to a potential exception?

pyomo/solvers/tests/checks/test_writers.py
```diff
@@ -132,6 +132,7 @@ def failing_writer_test(self):
     # Return a normal test
     return writer_test
 
+cls = None
```
jsiirola: Is this line needed? ...or is this just defensive programming due to a potential exception?

pyomo/solvers/tests/checks/test_pickle.py
ghackebeil: Defensive programming

pyomo/solvers/tests/checks/test_writers.py
ghackebeil: Defensive programming

pyomo/solvers/tests/checks/test_no_solution_behavior.py
ghackebeil: The problem is that I've only guaranteed logging when no solution exists for Kernel models. The AML logging output happens in `model.solutions.load_from` based on the status.

If someone want to expand the scope of the PR to make changes to the solver for AML models, I am okay with that.

pyomo/solvers/tests/checks/test_no_solution_behavior.py
ghackebeil: It was a typo

pyomo/solvers/tests/checks/test_no_solution_behavior.py
jsiirola: Given that this is an error and not an exception, I am OK with making the change for the AML side, too.

jsiirola: (Actually, what I meant to say is I am strongly in favor of adding it to the AML side)

PR #373, opened 2018-02-28
Clean up the pyomo.kernel namespace

ghackebeil:
### Changes ###
 - exposes `pyomo.version.version_info` as `pyomo.kernel.version_info`
 - adds a `__version__` string with the form "x.x.x" (this should probably be migrated back into pyomo.version as it seems pretty common to most libraries)
 - removes references to lower-level module files not meant to be used for general modeling (some of them have ugly confusing names previously used to distinguish them from similar files in pyomo.core.base)

I promise this is the last PR I am requesting before the 5.4 release ;)

### Legal Acknowledgement


jsiirola:
We should eventually have a discussion as to the role of the `pyomo.kernel` namespace before advertising it.  For example, since `pyomo.kernel` imports pyomo.environ, is there a reason this wasn't done inside `pyomo.environ`?  I could have easily seen the following syntax being preferable:

```
from pyomo.environ import kernel as pmo
```

ghackebeil:
I agree that the import syntax should likely change in the long run. For now, this enforces the idea that if you import `pyomo.kernel`, you should not be importing anything else under `pyomo` as the two likely will not mix.

jsiirola:
I went ahead and cleaned up the `__version__` calculation (and while I was at it, cleaned up the `pyomo.version` namespace (which I figure is within the scope of this PR).  @ghackebeil, I have approved your changes ... if my changes to pyomo.version look good to you, feel free to merge.

PR #383, opened 2018-03-11
Fix to kernel linear_constraint when assigning an empty list of terms

ghackebeil:
## Fixes # 

Allows assignment of an empty list to the `terms` keyword or property of linear_constraint in the Kernel. Before this fix, this kind of assignment would result in a `ValueError` caused by an attempt to unpack an empty return value from `zip` into two items.

### Legal Acknowledgement


ghackebeil:
They are 1 because I'm simply reusing the constraint from a few lines above which was initialized that way. Updating the terms only affects what is in the body. You are correct that this creates an infeasible constraint, but I'm simply doing low-level unit testing of the public API (which just happens to allow one to create infeasible constraints).

PR #386, opened 2018-03-14
Mixed-integer nonlinear decomposition toolbox in Pyomo (MindtPy)

bernalde:
## Summary/Motivation:
This pull request is to include the Mixed-Integer Decomposition Toolbox in Pyomo (MindtPy) as part of the contrib packages.

- Adding the MindtPy solver which includes Outer Approximation algorithm for MINLP.
- To come in future PRs: Partial Surrogate Cuts, Generalized Benders Decomposition, and Extended Cutting Plane 

### Legal Acknowledgement


qtothec:
I opened #594 to help resolve one of the errors I ran into. You might also try reading through http://pyomo.readthedocs.io/en/latest/developer_reference/expressions/design.html#expression-classes, but I still couldn't quite tell how to go about doing checking for SumExpression, or if there were hidden "gotchas" involved.

The others I saw are related to incorrect/too many/too few arguments to function calls. You should be able to address those, but let me know if you run into one that doesn't make any sense.

qtothec:
@bernalde We are running into this problem. Is this a nonlinear objective issue?

```
ERROR: Test the outer approximation decomposition algorithm.
  File "/home/travis/build/Pyomo/pyomo/pyomo/contrib/mindtpy/tests/test_mindtpy.py", line 100, in test_OA_MINLP_simple
    nlp_solver=required_solvers[0])
  File "/home/travis/build/Pyomo/pyomo/pyomo/contrib/mindtpy/MindtPy.py", line 342, in solve
    MindtPy_iteration_loop(solve_data, config)
  File "/home/travis/build/Pyomo/pyomo/pyomo/contrib/mindtpy/iterate.py", line 24, in MindtPy_iteration_loop
    solve_OA_master(solve_data, config)
  File "/home/travis/build/Pyomo/pyomo/pyomo/contrib/mindtpy/mip_solve.py", line 146, in solve_OA_master
    (master_terminate_cond, results.solver.message))
ValueError: MindtPy unable to handle MILP master termination condition of unbounded. Solver message: <undefined>
```

qtothec:
I think that this is now ready for review

pyomo/contrib/mindtpy/tests/alan.py
```diff
@@ -0,0 +1,49 @@
+#  MINLP written by GAMS Convert at 11/10/17 15:35:19
```
jsiirola: Please add a reference to the original model.  Since it is a direct translation of another model written in GAMS, we need to know the license under which the original model was released.

pyomo/contrib/mindtpy/tests/batchdes.py
```diff
@@ -0,0 +1,84 @@
+#  MINLP written by GAMS Convert at 11/10/17 15:35:19
```
jsiirola: Please add a reference to the original model.  Since it is a direct translation of another model written in GAMS, we need to know the license under which the original model was released.

pyomo/contrib/mindtpy/tests/flay03m.py
```diff
@@ -0,0 +1,100 @@
+#  MINLP written by GAMS Convert at 11/10/17 15:35:21
```
jsiirola: Please add a reference to the original model.  Since it is a direct translation of another model written in GAMS, we need to know the license under which the original model was released.

jsiirola (changes requested):
Overall, this looks good.  We need to address the licensing of some of the examples and then I think we will be good to go.

pyomo/contrib/mindtpy/tests/alan.py
bernalde: Solved in the last commit.

pyomo/contrib/mindtpy/tests/batchdes.py
bernalde: Solved in the last commit.

pyomo/contrib/mindtpy/tests/flay03m.py
bernalde: Solved in the last commit.

jsiirola (approved):

pyomo/contrib/mindtpy/cut_generation.py
```diff
@@ -0,0 +1,99 @@
...
+            var.value = val
+
+    # generate new constraints
+    # TODO some kind of special handling if the dual is phenomenally small?
+    jacs = solve_data.jacobians
+    for constr, dual_value in zip(MindtPy.constraint_list, duals):
+        if constr.body.polynomial_degree() in (1, 0):
+            continue
+        rhs = ((0 if constr.upper is None else constr.upper) +
+               (0 if constr.lower is None else constr.lower))
```
qtothec: You may want to consider the case of `constr.upper = constr.lower = some number`.

pyomo/contrib/mindtpy/tests/eight_process_problem.py
```diff
@@ -0,0 +1,149 @@
...
+"""
+from __future__ import division
+
+from six import iteritems
+
+from pyomo.environ import (Binary, ConcreteModel, Constraint, NonNegativeReals,
+                           Objective, Param, RangeSet, Var, exp, minimize)
+
+
+class EightProcessFlowsheet(ConcreteModel):
```
qtothec: Should this be replaced by just importing the GDP model and calling `bigM` on it, or is it worth keeping around the MINLP reformulation separately?

qtothec (changes requested):
Glancing over things again. Noticed something we ought to address, and have a question.

pyomo/contrib/mindtpy/cut_generation.py
bernalde: After some thought, I have decided to subtract the lower bound if both exist. In case of an equality constraint, we will result in an RHS equal to the original RHS and in the next lines relaxed to be generated in whichever direction the dual says. If it has both bounds and are different, it uses as the RHS the upper bound, which will be looser and therefore less risky to cut stuff that we wouldn't like to cut (UB>=LB).
Solved in the last commit.

pyomo/contrib/mindtpy/tests/eight_process_problem.py
bernalde: I would go for the independent files for the moment, but I'm open to just use the Big-M transformation of the GDP.

qtothec (approved):
Looks good, assuming that I didn't mess up and the tests pass.

PR #389, opened 2018-03-15
[WIP] Nested decomp

cristianallara:
-
-

### Legal Acknowledgement


whart222:
@jsiirola I seem to recall discussing with you that this example would sit better as a contrib package.  Do you recall?

carldlaird:
Not sure if you discussed this with John or not, but I agree. I asked for this to be moved from core Pyomo to examples, however, this was before contrib really existed. I think we should add an examples directory to contrib, and put this there.

We have asked Cristiana to move this already (and she did), so let's make sure we know where we want it before asking again (or move it ourselves).

whart222:
Do we really want a pyomo.contrib.examples sub-package?  I hadn't thought of that, and I'm inclined to use a simpler directory structure.

mrmundt:
This has been added to the [Archived design discussions wiki](https://github.com/Pyomo/pyomo/wiki/Archived-Design-Discussions); will close for now.

PR #392, opened 2018-03-16
Fix handling unbounded constraints

### LINKED ISSUE #388: Unbounded constraints for AMPL solvers
Hello,

I have been using Pyomo 5.2 until now, and yesterday decided to update to 5.4.3. In the version 5.2 it was possible to have unbounded constraints, namely, I could have something like this `sum(x) < float('inf')`. Now it is not possible to have this kind of constraints and I get `AssertionError` from the `ampl_.py` when the method `has_ub()` is invoked. I read in the changelog that methods `has_ub()` and `has_lb()` are present starting from 5.3 version. Is it possible to have the unbounded constraints in the current/next version? 

Best,
Pavlo

ghackebeil:
## Fixes # .

Makes the behavior of solvers interfaces consistent in how unbounded constraints are handled. Some interfaces would raise an exception and others would skip them. All solver interfaces now skip over them.

Fixes #388.

### Legal Acknowledgement

### DIFF
--- a/pyomo/solvers/plugins/solvers/cplex_direct.py
+++ b/pyomo/solvers/plugins/solvers/cplex_direct.py
@@ -341,18 +341,18 @@ class CPLEXDirect(DirectSolver):
             my_sense = 'E'
             my_rhs = [value(con.lower) - cplex_expr.offset]
             my_range = []
-        elif con.has_lb() and (value(con.lower) > -float('inf')) and con.has_ub() and (value(con.upper) < float('inf')):
+        elif con.has_lb() and con.has_ub():
             my_sense = 'R'
             lb = value(con.lower)
             ub = value(con.upper)
             my_rhs = [ub - cplex_expr.offset]
             my_range = [lb - ub]
             self._range_constraints.add(con)
-        elif con.has_lb() and (value(con.lower) > -float('inf')):
+        elif con.has_lb():
             my_sense = 'G'
             my_rhs = [value(con.lower) - cplex_expr.offset]
             my_range = []
-        elif con.has_ub() and (value(con.upper) < float('inf')):
+        elif con.has_ub():
             my_sense = 'L'
             my_rhs = [value(con.upper) - cplex_expr.offset]
             my_range = []
--- a/pyomo/solvers/plugins/solvers/cplex_direct.py
+++ b/pyomo/solvers/plugins/solvers/cplex_direct.py
@@ -283,8 +283,10 @@ class CPLEXDirect(DirectSolver):
             self._solver_model = self._cplex.Cplex()
         except Exception:
             e = sys.exc_info()[1]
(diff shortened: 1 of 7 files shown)


pyomo/solvers/plugins/solvers/cplex_direct.py
```diff
@@ -311,55 +316,72 @@ def _add_constraint(self, con):
...
             my_range = []
-        elif con.has_ub() and (value(con.upper) < float('inf')):
+        elif con.has_ub():
             my_sense = 'L'
             my_rhs = [value(con.upper) - cplex_expr.offset]
             my_range = []
         else:
-            raise ValueError('Constraint does not have a lower or an upper bound: {0} \n'.format(con))
+            raise ValueError("Constraint does not have a lower "
+                             "or an upper bound: {0} \n".format(con))
```
jsiirola: This seems to have a different behavior from AMPL above: AMPL ignores trivial (unbounded) constraints, but here we are throwing an error.

pyomo/solvers/plugins/solvers/gurobi_direct.py
```diff
@@ -278,34 +283,51 @@ def _add_constraint(self, con):
...
+                                                        name=conname)
+        elif con.has_ub():
+            gurobipy_con = self._solver_model.addConstr(lhs=gurobi_expr,
+                                                        sense=self._gurobipy.GRB.LESS_EQUAL,
+                                                        rhs=value(con.upper),
+                                                        name=conname)
         else:
-            raise ValueError('Constraint does not have a lower or an upper bound: {0} \n'.format(con))
+            raise ValueError("Constraint does not have a lower "
+                             "or an upper bound: {0} \n".format(con))
```
jsiirola: As with CPLEX above, this seems counter to this PR: exceptions will be generated for trivial constraints (instead of omitting them?)

pyomo/solvers/plugins/solvers/cplex_direct.py
ghackebeil: I made the change to skip unbounded constraints at the block level (`direct_or_persistent_solver._add_block`). I figured that at the level of granularity of calling `add_constraint`, raising an exception made more sense than doing nothing, so I left it the way @michaelbynum wrote it.

So to be clear, there is now a test that the solver plugins handle skip over unbounded constraints, and the Cplex and Gurobi direct interfaces pass this test in their current form.

pyomo/solvers/plugins/solvers/gurobi_direct.py
ghackebeil: see other comment

pyomo/solvers/plugins/solvers/cplex_direct.py
jsiirola: That makes sense to me.

PR #393, opened 2018-03-19
Allow retrieving a mutable Param even when no initial or default value is specified

jsiirola:
## Summary/Motivation:
In Pyomo 5.4 we became more stringent when retrieving Param objects when they do not have an initial value or a default value.  IDAES developers pointed out that this strict behavior is a regression from previous Pyomo releases.  They have a use case where it is completely reasonable to create an empty mutable Param, use it for generating expressions / constraints, and then fill in values before calling the solver. 

- This PR makes it allowable to fetch the (`_ParamData`) for a mutable Param without an initial or default value.  Retrieving the *value* of the `_ParamData` will still generate an exception.
- This PR also makes it so that we can `pprint` Params with non-numeric data (while seemingly unrelated, needed for testing the above change).

### Legal Acknowledgement


blnicho:
@jsiirola looks like a Python 3 incompatibility snuck into this PR related to calling `pprint` on Params with non-numeric data

jsiirola:
@blnicho: It turns out that the bug with `pprint` has been in there for at least a couple years.  I just had the misfortune of finally adding a test that exercised it.  That bug has been resolved (and for pprint-ing all components, not just Param).

Tests and coverage are now all passing.

pyomo/core/base/misc.py
```diff
@@ -97,6 +97,69 @@ def apply_parameterized_indexed_rule(obj, rule, model, param, index):
...
+                    # (run-to-run), but at least is consistent within
+                    # this run.
+                    self._typemap[_type] = lambda x:id(x)
+        return _type.__name__, self._typemap[_type](val)
+
+
+def sorted_robust(arg):
+    """Utility to sort an arbitrary iterable.
+
+    This returns the sorted(arg) in a consistent order by first tring
```
blnicho: Typo "tring"

Issue #22, opened 2016-08-20
Incorrect behavior with the relax_integrality solver option 


ghackebeil:
The way this option is handled in the CPLEX solver plugin is to simply change the problem type to an LP. However, this produces incorrect behavior when the problem has any quadratic constraints or objective terms (they get ignored).

Also, this is technically not a solver option. It should probably be managed using a keyword to the solve command.

mrmundt:
@ghackebeil - I'm going through old bug reports. Do you know if this bug still exists in the current version of Pyomo?

ghackebeil:
Yes. From CPLEX.py:
```python
        if 'relax_integrality' in self.options:
            script += 'change problem lp\n'
```

michaelbynum:
Don't we recommend doing this through a transformation now?

emma58:
Yes, `core.relax_integer_vars`.

PR #408, opened 2018-03-28
allowing relax_integrality option for gurobi direct and gurobi persis…

michaelbynum:
…tent

## Summary/Motivation:
The lp interface to gurobi handles a "relax_integrality" option. This PR adds the same functionality to the gurobi direct and persistent interfaces.

- 
-

### Legal Acknowledgement

### DIFF
--- a/pyomo/solvers/plugins/solvers/gurobi_direct.py
+++ b/pyomo/solvers/plugins/solvers/gurobi_direct.py
@@ -148,12 +148,30 @@ class GurobiDirect(DirectSolver):
                     raise
                 self._solver_model.setParam(key, float(option))
 
+        binary_vars = ComponentSet()
+        integer_vars = ComponentSet()
+        if 'relax_integrality' in self.options:
+            if self.options['relax_integrality']:
+                for v in self._solver_model.getVars():
+                    if v.vtype == self._gurobipy.GRB.BINARY:
+                        binary_vars.add(v)
+                        v.vtype = self._gurobipy.GRB.CONTINUOUS
+                    elif v.vtype == self._gurobipy.GRB.INTEGER:
+                        integer_vars.add(v)
+                        v.vtype = self._gurobipy.GRB.CONTINUOUS
+                self._solver_model.update()
+
         if self._version_major >= 5:
             for suffix in self._suffixes:
                 if re.match(suffix, "dual"):
                     self._solver_model.setParam(self._gurobipy.GRB.Param.QCPDual, 1)
 
         self._solver_model.optimize()
+
+        for v in binary_vars:
+            v.vtype = self._gurobipy.GRB.BINARY
+        for v in integer_vars:
+            v.vtype = self._gurobipy.GRB.INTEGER
 
         self._solver_model.setParam('LogFile', 'default')
 


ghackebeil:
@michaelbynum: Could you also look into fixing #22 with this PR. In particular, the behavior with the CPLEX LP-interface is broken. It just sets the problem type to an LP in the CPLEX run file, which will cause it to ignore quadratic portions of the model if it is a QP. It it doesn't look like there is an easy way to fix it, then don't worry about it.

pyomo/solvers/plugins/solvers/gurobi_direct.py
```diff
@@ -148,13 +148,31 @@ def _apply_solver(self):
...
                 if re.match(suffix, "dual"):
                     self._solver_model.setParam(self._gurobipy.GRB.Param.QCPDual, 1)
 
         self._solver_model.optimize()
 
+        for v in binary_vars:
+            v.vtype = self._gurobipy.GRB.BINARY
+        for v in integer_vars:
+            v.vtype = self._gurobipy.GRB.INTEGER
+
```
ghackebeil: Do you need to call self._solver_model.update() again at this point? Or does that clear the current solution?

pyomo/solvers/plugins/solvers/gurobi_direct.py
```diff
@@ -148,13 +148,31 @@ def _apply_solver(self):
                     raise
                 self._solver_model.setParam(key, float(option))
 
+        binary_vars = ComponentSet()
+        integer_vars = ComponentSet()
+        if 'relax_integrality' in self.options:
+            if self.options['relax_integrality']:
```
ghackebeil: I don't really like that we handle this kind of functionality through the solver options dictionary. That seems like it should be used for solver specific options only. Could this be implemented through a keyword to the solve method? Again though, I don't want to prevent you from getting other work done, so I will approve this PR either way.

pyomo/solvers/plugins/solvers/gurobi_direct.py
michaelbynum: It clears the current solution...

pyomo/solvers/plugins/solvers/gurobi_direct.py
michaelbynum: I completely agree. I only did it this way to match the behavior of the lp interface... If we want another solution, such as the one you suggested, that is perfectly fine with me.

michaelbynum:
@ghackebeil I should be able to take a look at the cplex issue late next week.

pyomo/solvers/plugins/solvers/gurobi_direct.py
jsiirola: Silly style question:  Why are the above two "if"s not combined into:
```
if self.options.get('relax_integrality', False):
```

pyomo/solvers/plugins/solvers/gurobi_direct.py
```diff
@@ -148,13 +148,31 @@ def _apply_solver(self):
...
+                for v in self._solver_model.getVars():
+                    if v.vtype == self._gurobipy.GRB.BINARY:
+                        binary_vars.add(v)
+                        v.vtype = self._gurobipy.GRB.CONTINUOUS
+                    elif v.vtype == self._gurobipy.GRB.INTEGER:
+                        integer_vars.add(v)
+                        v.vtype = self._gurobipy.GRB.CONTINUOUS
+                self._solver_model.update()
```
jsiirola: Question... are the bounds correctly set when changing the vtype to `GRB.CONTINUOUS`?  I only ask because I was once burned by this in Pyomo where simply changing the domain "forgot" the original bounds for Binary variables...

michaelbynum:
I am going to close this until I get a chance to look at the cplex issue. I will reopen it then.

PR #418, opened 2018-04-05
Use ctype instead of isinstance for Disjunct detection

qtothec:
This change allows for duck-typing of Disjuncts for the IDAES conceptual design framework.

### Legal Acknowledgement

### DIFF
--- a/pyomo/gdp/disjunct.py
+++ b/pyomo/gdp/disjunct.py
@@ -212,7 +212,7 @@ class _DisjunctionData(ActiveComponentData):
     def set_value(self, expr):
         for e in expr:
             # The user gave us a proper Disjunct block
-            if isinstance(e, _DisjunctData):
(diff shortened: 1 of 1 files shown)


pyomo/gdp/disjunct.py
```diff
@@ -212,7 +212,7 @@ def __getstate__(self):
     def set_value(self, expr):
         for e in expr:
             # The user gave us a proper Disjunct block
-            if isinstance(e, _DisjunctData):
+            if e.type() == Disjunct:
```
jsiirola: This should result in test failures.  For the case where the user provides implicit Disjuncts with
```
m.d = Disjunction(expr=[m.x >= 5, m.x <= 5])
``
the `e` will not be a (derived) Component and will not have a `.type()` method.

PR #422, opened 2018-04-09
Fixing some cut+paste issues when using atypical methods to create a …

jwatsonnm:
## Fixes # 
- Corrects some cut+paste issues that are incurred when using atypical methods (and untested methods) to create a scenario tree and all associated instances.

## Summary/Motivation:
- To make Dave Woodruff happy.

## Changes proposed in this PR:
- See above - minor syntax error fixes.

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #424, opened 2018-04-10
More cleanup related to what was fixed in #422. Also adding a few uni…

ghackebeil:
## Fixes # .

The code prior to #422 had intended to prevent doing a forced module import twice, but the typo broke that execution path. The change in #422 fixed the typo, but did so in a way that introduced the possibility of doing a forceful import twice (which might have bad side effects).

These changes fix that problem. They also add some additional unit tests to cover these cases.

## Summary/Motivation:

- to keep Dave Woodruff happy

### Legal Acknowledgement


DLWoodruff:
The motivating use-case is fixed.

PR #437, opened 2018-04-20
Gurobi persistent update

michaelbynum:
## Summary/Motivation:
This PR changes when Gurobi Model.update() gets called in the direct and persistent interfaces. This is both more efficient and accounts for older versions of Gurobi. In particular, the direct and persistent interfaces did not work with Gurobi versions less than 7 if any variables were fixed.

- See summary/motivation
-

### Legal Acknowledgement

### DIFF
--- a/pyomo/solvers/plugins/solvers/gurobi_direct.py
+++ b/pyomo/solvers/plugins/solvers/gurobi_direct.py
@@ -222,15 +222,15 @@ class GurobiDirect(DirectSolver):
         else:
             ub = self._gurobipy.GRB.INFINITY
 
+        if var.is_fixed():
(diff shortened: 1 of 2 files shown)


michaelbynum:
The call to optimize() actually does the update. The problem occurs when you add a var or constraint and then try to modify or remove it before calling optimize(). I was originally handling this by calling update after any var/constraint was added. However, that turns out to be slow for large models. This way, update() just gets called when it is really needed.

The try/except blocks are not covered by any tests yet. Good catch. I will add a couple tests real quick and then merge.

michaelbynum:
This is not ready. I will reopen this pull request at a later date.

PR #442, opened 2018-04-25
Fixing External Functions with fixed arguments in NL writer.

jsiirola:
This resolves issues where AMPL can SEGFAULT on expressions with external functions that have strictly constant arguments by simplifying the arguments and (if possible) eliminating the ExternalFunction call entirely.

## Summary/Motivation:

This issue was reported as part of IDAES:

> I’ve got a problem with the NL writer that I need to solve fast, so I was hoping you guy may know something that could help.  I have some external functions for steam property calculations, and in some cases using them can cause a segmentation fault in IPOPT.  I know you are thinking that I did something stupid in my external functions, but I really don’t think that is the case.  I also don’t think it is a problem with IPOPT.  The problem seems to be either the NL writer or the ASL or both.  I used a debugger, a lot of trial and error, and experimentation with the NL files to determine what I think is happening.  I should also say that using these external functions usually works fine.
>  
> So here is what’s happening:
>  
> When all the arguments to a function are fixed, usually the NL writer seems to turn that function into a constant and the function calls don’t show up in the NL file.  In that case everything works fine. In other cases for whatever reason the NL writer doesn’t replace those functions calls with a constant and the function shows up like this in the nl file
>  
> f1
> o2
> n0.1234
> n4.5678
> o3
>  
> In that case for whatever reason (an ASL bug?) there is a segfault (details at end of email).  This happens before IPOPT starts any calculation of anything. The problem seems to be in the ASL parsing the NL file. I really don’t know enough about the NL file format to know a malformed NL file. If I manually edit the NL file and make one of the constants a variable (like n4.5678 -> v16) it runs fine.
>  
> Do you guys know why this is happening?  Since fixing the ASL seems out of the question, maybe we could figure why the NL writer is putting these constant functions through?  I tried to reproduce this with a simple example, but I can’t come up with on that actually outputs a function like this without removing it. I can send you an example of this, but it requires compiling my functions and setting up environment variables and stuff.  It’s a bit of a pain.
> 

  
```
SEGFAULT Details:

starting program: /home/jeslick/local/bin/ipopt error.nl

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff754de71 in ewalk.constprop () from /home/jeslick/local/lib/libcoinasl.so.1
(gdb) bt
#0  0x00007ffff754de71 in ewalk.constprop () from /home/jeslick/local/lib/libcoinasl.so.1
#1  0x00007ffff754e2ea in ewalk.constprop () from /home/jeslick/local/lib/libcoinasl.so.1
#2  0x00007ffff754e2ea in ewalk.constprop () from /home/jeslick/local/lib/libcoinasl.so.1
#3  0x00007ffff754e2de in ewalk.constprop () from /home/jeslick/local/lib/libcoinasl.so.1
#4  0x00007ffff754e2de in ewalk.constprop () from /home/jeslick/local/lib/libcoinasl.so.1
#5  0x00007ffff754e2de in ewalk.constprop () from /home/jeslick/local/lib/libcoinasl.so.1
#6  0x00007ffff754e2de in ewalk.constprop () from /home/jeslick/local/lib/libcoinasl.so.1
#7  0x00007ffff754e2de in ewalk.constprop () from /home/jeslick/local/lib/libcoinasl.so.1
#8  0x00007ffff754e281 in ewalk.constprop () from /home/jeslick/local/lib/libcoinasl.so.1
#9  0x00007ffff754e2ea in ewalk.constprop () from /home/jeslick/local/lib/libcoinasl.so.1
#10 0x00007ffff754e2de in ewalk.constprop () from /home/jeslick/local/lib/libcoinasl.so.1
#11 0x00007ffff754ec7b in co_walkloop () from /home/jeslick/local/lib/libcoinasl.so.1
#12 0x00007ffff7550c09 in pfgh_read_ASL () from /home/jeslick/local/lib/libcoinasl.so.1
#13 0x00007ffff7bd10dd in Ipopt::AmplTNLP::AmplTNLP(Ipopt::SmartPtr<Ipopt::Journalist const> const&, Ipopt::SmartPtr<Ipopt::OptionsList>, char**&, Ipopt::SmartPtr<Ipopt::AmplSuffixHandler>, bool, Ipopt::SmartPtr<Ipopt::AmplOptionsList>, char const*, char const*, char const*, std::string*) () from /home/jeslick/local/lib/libipoptamplinterface.so.1
#14 0x0000000000401425 in main ()
(gdb) 
```

After a bit of debugging … Some notes:
1)	The current problem appears to be a SEGFAULT thrown by the ASL in pfg_read:ewalk.  I looked at all the NL we generate, and it is all valid.  Unfortunately the ASL itself is impenetrable, so I can’t really diagnose where the segfault is coming from.
2)	If  I hack the Pyomo NL writer to collapse the fixed arguments of an external function call down to scalar numbers, then we still die in a SEGFAULT, but this time it is coming from iapws95.cpp:cp_with_derivs (754: grad[0] = cv_grad[0] +…)
3)	If I further hack the NL writer to completely remove calls to the external function when the arguments are constant, then ipopt will run.  BUT, it fails in restoration.  There are tons of other issues, too.  The least of which is the number of AMPL evaluation errors thrown by ipopt.  If you set halt_on_ampl_errors, then you get mixes of divide-by-zero errors and “Error evaluating constraint turbine_inlet_stage.blade_velocity_correlation[0.0]: can't evaluate sqrt(-6.35059).”  There is also an invalid read of size 1 in Ipopt::AmplTNLP::call_hesset() at the very beginning, but that is nonfatal (and could be a valgrind issue).
4)	If I fix (2) above by only calculating the grad / hes when the result is non-null, we avoid the SEGFAULT and ipopt runs, but gets lost in restoration with a metric ton of “Error in an AMPL evaluation” warnings.
5)	Just fixing iapws95 and not also at least computing / simplifying the function arguments does not appear to resolve the SEGFAULT from (1) above.  We will probably have to make the NL writer hack permanent.

- Simplify fixed arguments to external functions in the NL writer.
- Simplify the entire external function call out if all arguments are fixed.

### Legal Acknowledgement

### DIFF
--- a/pyomo/repn/plugins/ampl/ampl_.py
+++ b/pyomo/repn/plugins/ampl/ampl_.py
@@ -574,6 +574,15 @@ class ProblemWriter_nl(AbstractProblemWriter):
                 self._print_nonlinear_terms_NL(exp._args[0])
 
             elif exp_type is expr._ExternalFunctionExpression:
+                # We have found models where external functions with
+                # strictly fixed/constant arguments causes AMPL to
+                # SEGFAULT.  To be safe, we will collapse fixed
+                # arguments to scalars and if the entire expression is
+                # constant, we will eliminate the external function
+                # call entirely.
+                if exp.is_fixed():
+                    self._print_nonlinear_terms_NL(exp())
+                    return
                 fun_str, string_arg_str = \
                     self._op_string[expr._ExternalFunctionExpression]
                 if not self._symbolic_solver_labels:
@@ -589,6 +598,8 @@ class ProblemWriter_nl(AbstractProblemWriter):
                 for arg in exp._args:
                     if isinstance(arg, basestring):
                         OUTPUT.write(string_arg_str % (len(arg), arg))
+                    elif arg.is_fixed():
+                        self._print_nonlinear_terms_NL(arg())
                     else:
                         self._print_nonlinear_terms_NL(arg)
             elif (exp_type is expr._PowExpression) or \


jsiirola:
@eslickj: here is the fix to the NL writer

whart222:
@jsiirola Can you add a test to confirm this behavior?  Otherwise, I fear this fix will be lost when we merge expr_dev.

eslickj:
@jsiirola, thanks for getting to this so quickly.

PR #444, opened 2018-04-25
Gams writer fix for expr_dev

qtothec:
## Summary/Motivation:
The GAMS `power(x, y)` function is intended only for integer `y` exponent values. This change uses the GAMS `x ** y` operator for real number `y` exponents. 

### Legal Acknowledgement

### DIFF
--- a/pyomo/repn/plugins/gams_writer.py
+++ b/pyomo/repn/plugins/gams_writer.py
@@ -68,8 +68,9 @@ class ToGamsVisitor(EXPR.ExpressionValueVisitor):
         if node.__class__ is EXPR.PowExpression:
             # If the exponent is a positive integer, use the power() function.
             # Otherwise, use the ** operator.
-            if (tmp[1].__class__ in native_numeric_types and
-                    tmp[1] == int(tmp[1])):
+            exponent = node.arg(1)
+            if (exponent.__class__ in native_numeric_types and
+                    exponent == int(exponent)):
                 return "power({0}, {1})".format(tmp[0], tmp[1])
             else:
                 return "{0} ** {1}".format(tmp[0], tmp[1])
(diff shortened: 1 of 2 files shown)


whart222:
@qtothec Please (1) add a test to confirm the desired behavior, and (2) add a note to the CHANGELOG about this PR.  Thx

jsiirola:
CHANGELOG is not necessary/appropriate, as this is fixing an error on expr_dev and not changing behavior from a previous release.

whart222:
If we're starting to see PRs applied to branches, then I think it's reasonable to document them in the CHANGELOG.

blnicho:
Why not just use '**' always (even in the case of integer exponents)?

whart222:
@qtothec I don't understand your comment about the 'string'.  Internally, none of the elements of an expression tree are strings.

qtothec:
@blnicho There's an advantage to using `power()` for integer exponents, because GAMS has slightly specialized behavior for it. Specifically, `0 ** 3` yields an error but `power(0, 3)` does not. `x ** y` results in `exp(y * log(x))`.

@whart222 I was trying to interrogate the type of `tmp[1]` in the changed code, and it appeared to be `str`.

whart222:
@qtothec Note that `tmp` is defined at the beginning of this function, and that it's a function of the values of the node arguments (see line 53).  So you can query the value of `node._args_[1]`  (or, better yet, `node.arg(1)`).

qtothec:
Thanks. That makes a lot more sense. Aside from baseline test change failures, I hope this should work now.

whart222:
I'm closing this, after migrating these changes to PR #449

whart222:
@qtothec Do you need help with these test failures?  I'm not sure I understand your intent.

qtothec:
I'm hoping that my most recent commit on #454 resolved the failures.

PR #462, opened 2018-05-02
Doctest fix

ghackebeil:
## Fixes # .

Fixes execution of doctests

## Summary/Motivation:

 - `make doctests` was failing due to unknown target `doctests` (just needed to remove "s").
 - failure return code was being overwritten by a later command (replaced ; with &&)
 - needed to `pip install sphinx`

### Legal Acknowledgement

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,7 +79,7 @@ install:
     fi
   - pip install --quiet xlrd
   - pip install --quiet openpyxl
-  - pip install sphinx_rtd_theme
+  - pip install sphinx sphinx_rtd_theme
   - pip install dill
   - conda install -q -y -c anaconda pandas networkx
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -163,7 +163,7 @@ script:
   - test.pyomo -v --cat=$CATEGORY pyomo `pwd`/pyomo-model-libraries
 
   # Run documentation tests
-  - cd doc/OnlineDocs; make doctests -d; cd ../..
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -163,7 +163,7 @@ script:
   - test.pyomo -v --cat=$CATEGORY pyomo `pwd`/pyomo-model-libraries
 
   # Run documentation tests
-  - cd doc/OnlineDocs; make doctests -d; cd ../..
+  - cd doc/OnlineDocs && make doctest -d && cd ../..
```
jsiirola: This worries me.  If doctest fails, then the CWD is not restored back to outside of the doc directory.  I have frequently used:
```
pushd doc/OnlineDocs
make doctest -d # possible also '|| exit 1'
popd
```
when I was working with scripts that had to change/restore the CWD.

ghackebeil:
maybe just just delete the `&& cd ..` and add another line in the travis file immediately after with something like `cd $TRAVIS_BUILD_DIR`

ghackebeil:
Also, if doctests fails, would travis stop executing commands anyway? If so, what's the problem.

jsiirola:
I am not sure if Travis aborts on nonzero return codes.  For example, nose / test.pyomo return nonzero return codes any time a test fails, yet the build continues...

ghackebeil:
Ah good point. What about just using `make -C doc/OnlineDocs …`?

PR #463, opened 2018-05-03
Preemptive Python-3.7 fixes in PySP

ghackebeil:
In Python-3.7, `async` becomes a directive, so it can not be used as a variable name.

 - In phinit.py, the --async argparse option will now store its result into `async_mode`
 - In the `ScenarioTreeManager` class, the `async` keyword used in various methods has been renamed `async_call`.
 - For consistency with the above change, in the same methods, the `oneway` keyword has been renamed to `oneway_call`.

### Legal Acknowledgement


jwatsonnm:
I performed a visual check of the diff, and all appears well / benign.

PR #469, opened 2018-05-05
Optimizations to minimize use of NumericConstant objects

whart222:
## Summary/Motivation:
Clean-up the use of as_numeric and re-define it with a more restrictive semantics.

The as_numeric() function is used to create NumericConstant objects, which are used to wrap numeric values in Pyomo components.

This PR also changes the caching mechanism.  Values are not coerced to floats, but instead they are cached separately for each type.

### Legal Acknowledgement


whart222:
Most of the changes in this PR reflect baseline updates, since numeric values are cached without coercion.

whart222:
This PR eliminates the unnecessary use of as_numeric() within the value() function (and other functions).  Additionally, it eliminates calls to value() when we can reliably expect expressions.  Altogether, this results in a modest improvement in Pyomo performance (~3% in many test problems).

pyomo/core/expr/numvalue.py
```diff
@@ -280,78 +257,65 @@ def polynomial_degree(obj):
...
 # share constants we have seen before.
-KnownConstants = {}
-
-def update_KnownConstants(obj, val):
-    if len(KnownConstants) < 100:
-        KnownConstants[obj] = val
+_KnownConstants = {}
 
 
 def as_numeric(obj):
```
jsiirola: This PR fundamentally changes the meaning of `as_numeric()`.  Since pre-coopr 3.x, `as_numeric` returned its argument cast into a class *derived* from `NumericValue`.  The other key thing that `as_numeric` did was maintain / update the `native_numeric_types` set when users used types that Pyomo doesn't know about but behave like numeric types.  This PR changes that so that it is not clear what / how the `native_numeric_types` set gets updated.

pyomo/core/expr/numvalue.py
```diff
@@ -280,78 +257,65 @@ def polynomial_degree(obj):
...
-        % (type(obj).__name__, obj, ))
+        #
+        # Because INT, FLOAT, and sometimes LONG hash the same, we
+        # index _KnownConstants with both the class and value.  This ensures
+        # consistent results.
+        #
+        try:
+            return _KnownConstants[obj.__class__, obj]
+        except KeyError:
+            pass
```
jsiirola: Exceptions are MUCH slower than testing for existence and returning if found.

pyomo/core/expr/numvalue.py
```diff
@@ -280,78 +257,65 @@ def polynomial_degree(obj):
...
+        # is worth the extra cost.
+        #
+        if len(_KnownConstants) < 100:
+            _KnownConstants[obj.__class__,obj] = retval
+            return retval
+        #
+        return retval
+    elif obj.__class__ in native_types:
+        raise TypeError("Cannot treat the value '%s' as a constant" % str(obj))
+    return obj
```
jsiirola: There is no error checking for non-numeric types that are not in native_types.

pyomo/core/tests/unit/test_numvalue.py
```diff
@@ -135,7 +136,7 @@ def test_const2(self):
         val = NumericConstant('foo')
         self.assertTrue(is_constant(val))
 
-    def test_error(self):
+    def Xtest_error(self):
```
jsiirola: What's an Xtest?  If you are deleting this test, please delete it.

whart222:
@jsiirola I need a motivating example to illustrate the type of functionality that you think is being omitted.  I don't get what your comments are pointing to, which I suspect reflects a weakness in our testing.

jsiirola:
In the original design of the expression system, we wanted to be able to coexist with other non-built-in numeric type systems.  The original design did not attempt to list *every* possible type / numeric type; instead prepopulating the `native_types` and `native_numeric_types` sets with *Python* data types and allowing others (notably numpy types) to be added when they were first observed.  This allowed Pyomo to be able to be seamlessly used with other numeric data systems.

That management was all handled through `as_numeric`.  This was attractive as it was a *single* place where we could write an efficient and robust approach for automatically identifying numeric types and adding them to the sets.

The specific use case is the following:
```
import numpy
model.c = Constraint(expr= model.x**2 <= numpy.double(3.14))
```
For the current system, this works only because we brute-force in the numpy types through the kernel (`core/kernel/register_numpy_types.py`), which bypasses the issue by explicitly adding every numpy type to the various sets.  The original design avoided this.  Without the automatic set management, we now need to add a `register_foo_types` for every library `foo` that a user would like to mix with Pyomo -- and make sure that that gets triggered before any expression generation.  If we don't, then several of the methods will generate "AttributeErrors" for numeric types that are not in the `native_numeric_type` set.

pyomo/core/expr/numvalue.py
```diff
@@ -180,18 +180,12 @@ def value(obj, exception=True):
...
-        try:
-            numeric = as_numeric(obj)
-        except ValueError:
-            if isinstance( obj, string_types + (text_type, binary_type) ):
-                native_types.add(type(obj))
-                return obj
-            raise
-    try:
-        tmp = numeric(exception=exception)
+        tmp = obj(exception=exception)
```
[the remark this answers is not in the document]
whart222: Can you give an example?

pyomo/core/expr/numvalue.py
whart222: That's the design I want.

pyomo/core/expr/numvalue.py
```diff
@@ -280,78 +257,74 @@ def polynomial_degree(obj):
...
+        #
+        if len(_KnownConstants) < 100:
+            _KnownConstants[obj.__class__,obj] = retval
+            return retval
+        #
+        return retval
+    elif obj.__class__ in native_types:
+        raise TypeError("Cannot treat the value '%s' as a constant" % str(obj))
+    elif not isinstance_NumericValue(obj):
+        raise TypeError("Cannot treat the value '%s' as a constant because it has unknown type '%s'" % (str(obj), obj.__class__))
```
jsiirola: Why isinstance?  In the old design, we did the equivalent of the following:
```
try:
    return obj.as_numeric()
except AttributeError:
    # try to figure out what to do with this type.  If it is a "new" numeric type add it to the native_numeric_types set and return it.  If it isn't, add it to the native_types set and raise an exception.
```
That is, if the object contained an `as_numeric` attribute, then we took whatever that method returned as the NumericValue.  That allowed both for duck-typing, and for non-NumericValue classes to be used in expressions as long as they implemented an as_numeric method that returned a valid NumericValue-like thing.

jsiirola:
Given that we still haven't resolved issues from expr_dev, I am not personally in favor of "merge it now and fix it later".

whart222:
@jsiirola This PR is independent of issues within expr_dev.  Hence, I don't see the logic for holding up this PR.

whart222:
The recent changes re-added the logic for auto-registering numeric types.  However, I added a warning to users that the Pyomo registration functions are recommended.

ghackebeil:
@wehart: I just noticed that the changes on this branch now result in the kernel.constraint lb and ub attributes converting simple types into NumericConstants:
```python
>>> import pyomo.kernel as pmo
>>> c = pmo.constraint()
>>> c.lb = 1
>>> type(c.lb)
<class 'pyomo.core.expr.numvalue.NumericConstant'>
```
Did you mean to change the API in Kernel?. I would prefer those setters simply assert that what is being assigned is not potentially variable. I had done this by creating a temporary NumericConstant and checking the is_potentially_variable method on that, but if you know of a cheaper way to do that I would be okay with it.

Also, note that I thought I was testing the behavior that I expect, albeit implicitly, by assuming that Pyomo relational expressions eventually triggered an exception when cast to bool (and there are plenty of relational expressions being created by the unit tests I wrote now that this change has been made). However, Pyomo relational expressions seem to be evaluating to what a user might expect. E.g.,
```python
>>> import pyomo.kernel as pmo
>>> c = pmo.constraint()
>>> x = pmo.variable()
>>> c.lb = 1
>>> x.value = 1
>>> bool(c.lb == x)
True
>>> x.value = 0
>>> bool(c.lb == x)
False
```
Are the other developers aware of this behavior?

ghackebeil:
I've reverted this change in behavior in the kernel and expanded unit tests to ensure the type does not change when assigning a value to constraint and variable bounds.

I've also added a new function to numvalue.py named `is_numeric_data` to ensure valid assignments to these attributes without calling as_numeric. Calling this function is almost equivalent to `not potentially_variable(...)`, except it is not true for strings and other non-numeric native types. E.g.,
```python
is_numeric_data(1.0)           # -> True
is_numeric_data(parameter())   # -> True
is_numeric_data("a string")    # -> False
is_numeric_data(variable())    # -> False
is_numeric_data(expression())  # -> False
```

whart222:
@ghackebeil Thanks for checking this out and making this change!  Given your feedback this week, I actually started looking at reverting the change as you suggested.  But then I was confused why tests were passing, so I didn't.

whart222:
@ghackebeil I'll follow-up with a separate issue about relational expressions.  I think there has been a change there.

whart222:
@jsiirola AFAIK I've addressed all concerns.  Also, our tests demonstrate a fairly consistent improvement in performance from this change.

pyomo/core/expr/numvalue.py
```diff
@@ -180,18 +180,12 @@ def value(obj, exception=True):
     """
     if obj.__class__ in native_types:
         return obj
+    if obj.__class__ is NumericConstant:
+        if obj.value is None:
+            raise ValueError("No value for uninitialized NumericConstant object %s" % (obj.name,))
```
jsiirola: Bug: this raises an exception regardless of the value of the "exception" flag.

Editorial: separately, why do we have the exception flag?

pyomo/core/expr/numvalue.py
```diff
@@ -266,92 +247,139 @@ def potentially_variable(obj):
...
-                    return tmp
+                #
+                # Create the numeric constant and add to the 
+                # list of known constants.
+                #
+                # Note: we don't worry about the size of the
+                # cache here, since we need to confirm that the
+                # object is hashable.
+                #
+                _KnownConstants[obj.__class__,obj] = retval
```
jsiirola: Editorial, but strongly recommended (and easy to implement):  In an attempt to reduce the amount of repeated code, instead of creating the constant here, what about just updating the sets, toss the warning, and then returning a recursive call?  That also has the benefit of avoiding needing to explain why we are ignoring the cache size here.

pyomo/core/tests/unit/test_component_constraint.py
```diff
@@ -381,7 +398,7 @@ def test_equality(self):
         self.assertEqual(c.rhs, 3)
         self.assertEqual(c.equality, True)
 
-        with self.assertRaises(TypeError):
+        with self.assertRaises(ValueError):
             c.rhs = 'a'
```
jsiirola: Why is this a ValueError now? I would think that the former TypeError was more appropriate.

pyomo/core/expr/numvalue.py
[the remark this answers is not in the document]
jsiirola:
```
m.p = my_numeric_like_class_that_pyomo_has_never_seen_before()
#...
def some_rule(m, i):
    if i == value(m.p):  # because we train users to wrap things in value() calls...
        return Constraint.Skip
    #...
```
In this case, the first time Pyomo sees this class is in the `value()` call -- not in as_numeric()

pyomo/core/expr/numvalue.py
```diff
@@ -223,11 +216,7 @@ def is_constant(obj):
     #
     if obj.__class__ in native_types:
         return True
-    try:
-        return obj.is_constant()
-    except AttributeError:
-        pass
-    return as_numeric(obj).is_constant()
+    return obj.is_constant()
```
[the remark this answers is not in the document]
jsiirola: Again, as with value, there is no guarantee that the first time Pyomo sees a new class object is in the expression system.  It could be through any of the "global scope" information methods: `value()`, `is_constant()`, etc...

The advantage of the:
```
try:
    return obj.is_constant()
except AttributeError:
    return as_numeric(obj).is_constant()
```
idiom is that it costs NO time in the nominal case, but works correctly in the exceptional case where this is the first time that Pyomo sees a new class.

pyomo/core/expr/numvalue.py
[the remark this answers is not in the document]
jsiirola: In addition, as with `value()` below, the reason to use the `.as_numeric()` attribute is so that the nominal case works quickly (Pyomo numeric types all have `as_numeric` attributes)  The way this PR is written, you will get a generic exception for anything that is not callable with an `exception` argument - either because it was a Pyomo numeric type that failed to evaluate, or if it was a non-numeric type.

jsiirola:
Separately, I finally remembered why we mapped integers to floating point.  Consider the following it the mapping does not take place:
```
m.x = Var()
m.c = Constraint(expr= 1 <= m.x)
m.c = m.c.lower/2 <= m.x
```
In Python 2.7 users could get a nasty surprise if the division occurred as integer division.  Now, I think we are safe here as I don't think that would happen: the `lower` attribute would always return a NumericValue, so the division would actually happen in the expression module where we `from __future__ import division`.  However, things are nastier with Param.  Consider the following:
```
m.p = Param([1], initialize={1:1})
m.q = Param(initialize=1)
m.p[1] / 2 # == 0
m.q / 2 # == 0.5
```
This is the current behavior on master, and is almost certainly a BUG.

pyomo/core/expr/numvalue.py
whart222: This code is simpler because the code is repeated.  The recursive logic is less clear, and the duplicated code is < 10 lines.

whart222:
I'm inclined to treat value() different from is_*().  In fact, I'm not even sure why we have the is_*() functions.  Hence, it's not obvious whether they are used within our library or by users.  This is an important distinction w.r.t. whether we need to test for unseen data types.  (It relates to the frequency with which this will happen.)  Thoughts?

jsiirola:
I believe the historical context for the `is_*` functions fell under the umbrella of "convenience functions".  I used to agree that they should have been removed.  However, the introduction of immutable Params raised a sticky issue around the programming API:
```
m.p = Param([1], initialize={1:1}, mutable=True)
m.q = Param([1], initialize={1:1})
m.p[1].is_constant() # == False
m.q[1].is_constant() # AttributeError !
```
The module-level "convenience" functions provide a standard API that users can use that is independent of the storage implementations in components like Param.  This is also why we advocate `value(m.x)` and not `m.x.value()`.

whart222:
@jsiirola I reverted to coerce all NumercConstant values to floats.  I also added tests that cover most of the value() and as_numeric() code.  There are a few exceptional conditions in as_numeric that aren't covered, for which it wasn't immediately obvious how to create a test.

whart222:
@jsiirola This is still blocked awaiting your review.  I think I've addressed the concerns we discussed.

whart222:
@jsiirola Checkout the test failure in test_chull.py.  I've been seeing this for a while on my Mac, and I thought this was a Mac-specific issue.  But I think this is a weak test.  AFAIK, this test failure is not specific to this branch.

whart222:
@jsiirola What's the ETA on finalizing this PR?  I'm inclined to mark the CHULL test as fragile and then finalize this PR.

whart222:
I'm merging this, per my conversation with John.

PR #474, opened 2018-05-10
[WIP] Pynumero

santiagoropb:
## Summary/Motivation:

Add PyNumero to pyomo.contrib. PyNumero is a python package for numerical optimization that provides a high-level programming framework for rapid development of nonlinear optimization algorithms. The package gives access to all high-level features of the python programming language without making large sacrifices on computational performance. It combines the capabilities of Pyomo with libraries like the AMPL Solver Library (ASL), The Harwell Subroutine (HSL) (optional), and NumPy/SciPy. PyNumero performs all linear algebra operations in compiled code, and is designed to avoid marshalling of data between the C and the Python environments. 

-
-

### Legal Acknowledgement


jsiirola:
@santiagoropb: Overall this looks good (and test are passing everywhere).  However, coverage is a bit low.  Can you take a look at some of the uncovered code and add "simple" tests?

qtothec:
I keep forgetting exactly what this package will give us. Does it allow us to avoid writing `*.nl` files for IPOPT?

jsiirola:
No - You still have to write the NL file (although we can avoid writing it to disk).  The big wins are that you can repeatedly interrogate the model (compute GJH) at various points without generating another model and that the interrogation doesn't spawn a separate process (like GHJ does).

jsiirola:
Superseded by #695.

santiagoropb:
> I keep forgetting exactly what this package will give us. Does it allow us to avoid writing `*.nl` files for IPOPT?

@qtothec The package does rely on the NL to compute derivatives via ASL. However, the main advantage is that it only requires to write that file once and query the NLP information several times. This has a couple of benefits:

1. It is a step closer towards a persistent interface. The NLP class in pynumero allows for some modifications to the NLP problem without rewriting the NL file. One example is the implementation of ADMM/PH. I have written a basic implementation of PH with Pynumero where the complicating variables and multipliers of linking constraints are updated at every PH iteration without rewriting the NL file.

2. As it is completely based on Numpy and Scipy, Pynumero can be used to write algorithms that can fully integrate with Numpy or Scipy subroutines. For instance, one can easily write a Newton solver where the Jacobian and the right hand side are evaluated via pynumero in C and the linear system is solve with one of Scipy sparse solvers. The data remains in C and Numpy and Scipy will only manipulate the pointers populated from pynumero.

3. The package was design for prototyping nonlinear decomposition algorithms from python while doing expensive operations in C. Pynumero.sparse has functionality that facilitates building block algebra components. An easy example is the BlockMatrix. Some of the Work from Larry has focused on manipulating submatrices from the KKT system to decompose the solution of the augmented system in the interior-point (e.g. Schur-complement, Cyclic reduction). Such manipulations can be done fairly easily with pynumero and given that we are based on numpy parallelization can be supported via mpi4py

PR #487, opened 2018-05-15
Allow targets to be specified for fix_disjuncts transformation

qtothec:
## Summary/Motivation:
I learned today that the first argument to `apply_to` needs to be Block-like. This change allows specification of custom targets for the fix_disjuncts transformation.

### Legal Acknowledgement


qtothec:
Appropriate behavior for `create_using` is contingent on resolution of #497.

PR #495, opened 2018-05-17
Adding relative_to argument for getname()

whart222:
## Summary/Motivation:
Adding option to specify a relative block when generating a fully qualified name.

- Changes to the getname() method

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/component.py
+++ b/pyomo/core/base/component.py
@@ -451,9 +451,10 @@ class Component(_ComponentBase):
         """
         if fully_qualified:
             pb = self.parent_block()
-            if pb is not None and pb is not self.model() and (relative_to is None or relative_to is not pb):
-                ans = pb.getname(fully_qualified, name_buffer, relative_to) \
-                      + "." + self._name
+            if relative_to is None:
+                relative_to = self.model()
+            if pb is not None and pb is not relative_to:
+                ans = pb.getname(fully_qualified, name_buffer, relative_to) + "." + self._name
             else:
                 ans = self._name
         else:
--- a/pyomo/core/base/component.py
+++ b/pyomo/core/base/component.py
@@ -455,6 +455,8 @@ class Component(_ComponentBase):
                 relative_to = self.model()
             if pb is not None and pb is not relative_to:
                 ans = pb.getname(fully_qualified, name_buffer, relative_to) + "." + self._name
+            elif pb is None and relative_to != self.model():
+                raise RuntimeError("The relative_to argument was specified but not found in the block hierarchy: %s" % str(relative_to))
             else:
                 ans = self._name
         else:
--- a/pyomo/core/base/component.py
+++ b/pyomo/core/base/component.py
@@ -46,12 +46,12 @@ def _name_index_generator(idx):
         return "[" + _escape(str(idx)) + "]"
 
 
-def name(component, index=None, fully_qualified=False):
+def name(component, index=None, fully_qualified=False, relative_to=None):
     """
     Return a string representation of component for a specific
     index value.
     """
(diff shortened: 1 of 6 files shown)


pyomo/core/base/component.py
```diff
@@ -438,19 +438,21 @@ def to_string(self, verbose=None, labeler=None, smap=None, compute_values=False)
...
                                     generation (using a dictionary)
+            relative_to         When generating a fully qualified name,
+                                    stop at this block.
         """
         if fully_qualified:
             pb = self.parent_block()
-            if pb is not None and pb is not self.model():
-                ans = pb.getname(fully_qualified, name_buffer) \
+            if pb is not None and pb is not self.model() and (relative_to is None or relative_to is not pb):
+                ans = pb.getname(fully_qualified, name_buffer, relative_to) \
```
jsiirola: This is technically redundant.
```
if pb is not None and pb is not self.model() and pb is not relative_to:
```
is sufficient as the first clause becomes redundant when relative_to is None.  Now that I think about it, I think a better (clearer) approach would be to make this code:
```
pb = self.parent_block()
if relative_to is None:
    relative_to = self.model()
if pb is not None and pb is not relative_to:
    ans = pb.getname(fully_qualified, name_buffer, relative_to) + "." + self._name
```
This approach explains that the name is relative to the model unless explicitly told otherwise.  It also eliminates a redundant calls to `model()` in deeply-nested models.

jsiirola:
I was just talking to @carldlaird about this PR, and we are mutually unsure what the correct behavior should be for the following case:
```
m = ConcreteModel()
m.b = Block()
m.b.x = x = Var()
m.c = Block()
x.getname(fully_qualified=True, relative_to=m.c)
```
Naively, I would want `../b/x` ... but we don't use `/` for the object separator, and have no obvious concept of "up one level".  In light of that, should this example generate an error (or exception?).  The current behavior is to return the fully-qualified name relative to the model, but then:
```
rqn = x.getname(fully_qualified=True, relative_to=m.c)  # == 'b.x'
find_x = m.c.find_component(rqn)  # == None
```
which seems bad.  @whart222: any thoughts?

whart222:
@jsiirola If this PR is acceptable for merging, then why was the review marked as requiring revisions?

whart222:
@jsiirola @carldlaird I think your example should generate an error, since the relative block isn't in the hierarchy containing the variable.  I'll make that change.

jsiirola:
@whart222: my review is not marked as "requiring revisions".  I submitted it as a comment, which is neither an approval (so merging is still waiting for an approved review), but is also not *blocked* by me (revisions required).

whart222:
@jsiirola  I guess I don't understand the logic of adding a comment rather than an approval.  Given GitHub's review logic, the effect is the same as requiring revisions since a merge isn't possible without an approving review.

jsiirola:
A comment doesn't prevent *someone else* from approving, whereas requiring revisions blocks the merge regardless of if someone else approves.

whart222:
@jsiirola  Yes, I understand the mechanical distinction.  I just didn't understand why you would add a comment saying you're ok with the merge without actually approving the merge.

jsiirola:
For the record, I said that "I didn't see anything that would prevent merging", not that I was OK with merging it.  A subtle distinction perhaps, but an important one.  I am trying to provide the "strong editorial fix" versus "changes required to fix bugs" distinction that you have been advocating.

While I am thinking about it (and providing editorial comments), I had to think for a long time why a22520c was correct in all the edge cases (like when relative_to was and was not provided, crossed with the crazy situation where the component in question was and was not attached to a block).  We should consider adding some code documentation to save both of us time the next time we revisit this corner of the codebase.

PR #497, opened 2018-05-17
Storing transformation arguments on model before cloning in create_using

jsiirola:
## Summary/Motivation:
There has always been a level of complexity when passing arguments to transformations through the `create_using` API.  The default implementation of `_create_using` simply clones the model and calls `apply_to` on the resulting instance.  The problem arises when passing model components or model references as arguments.  In this case, the arguments will refer to the original model and make no sense on the resulting model.  We have previously worked around this using ComponentUIDs.

This PR simplifies things by attacking the arguments onto the model before cloning the model, and then pull the arguments for `apply_to` from the generated instance.  This ensures that any references to the model will be automatically be updated to point to the new model.  Arguments are removed from both models before returning.

The only downsides are:
 - users cannot receive side effects through arguments to the default implementation of `_create_using`
 - there could be performance implications when passing very large data structures to `create_using`.

Both could be avoided by either cloning the model yourself and calling `apply_to`, or for transformations implementing custom `_create_using` methods.

- Attach arguments to `create_using` to the model before cloning

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/plugin.py
+++ b/pyomo/core/base/plugin.py
@@ -36,9 +36,9 @@ __all__ = ['pyomo_callback',
 
 import logging
 import pyutilib.misc
-from pyomo.util.deprecation import deprecated
-from pyomo.util.modeling import unique_component_name
-from pyomo.util.plugin import (
+from pyomo.common.deprecation import deprecated
+from pyomo.common.modeling import unique_component_name
+from pyomo.common.plugin import (
     alias, implements, Interface, Plugin, PluginFactory, CreatePluginFactory,
     PluginError, ExtensionPoint )
 from pyomo.common.timing import TransformationTimer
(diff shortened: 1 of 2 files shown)


pyomo/core/base/plugin.py
```diff
@@ -350,7 +351,17 @@ def _apply_to(self, model, **kwds):
             "The Transformation.apply_to method is not implemented.")
 
     def _create_using(self, model, **kwds):
+        # Put all the kwds onto the model so that when we clone the
+        # model any references to things on the model are correctly
+        # updated to point to the new instance.  Note that users &
+        # transformation developers cannot rely on things happening by
+        # argument side effect.
```
qtothec: Should we add in a warning/disclaimer about functions being passed in to kwds?

jsiirola: @qtothec: I don't think that is an issue.  Functions generally don't store state and therefore shouldn't care which instance they may eventually be called on, and functors (callable objects), which was my original worry turn out to work fine.  The only case that might burn users are functions that refer to the model directly without passing it in as an argument.  That will fail, but I also don't think there is any way that we could catch that edge case.

jsiirola:
@qtothec: good catch. CHANGELOG is updated.

PR #498, opened 2018-05-17
Remove GDP var mover false positive warning

### LINKED ISSUE #496: Appropriate behavior for activate() and deactivate() for nested disjunctions?
```
from pyomo.environ import *
m = ConcreteModel()
m.d1 = Disjunct()
m.d2 = Disjunct()
m.d1.sub1 = Disjunct()
m.d1.sub2 = Disjunct()
m.d1.disj = Disjunction(expr=[m.d1.sub1, m.d1.sub2])
```

What should happen when disjuncts containing nested disjunctions are deactivated? `m.d1.deactivate()`.
The reclassifier hack complains about `m.d1.sub1` and `m.d1.sub2` not being expanded.

qtothec:
## Summary/Motivation:
Var mover complains about active disjuncts wherever they appear in the model, even if they are underneath a deactivated `Block` or deactivated `Disjunct` with `indicator_var` fixed to zero. This performs a check to make sure the disjunct in question is on an active path of the model before complaining.

Fixes #496 

### Legal Acknowledgement


pyomo/gdp/plugins/gdp_var_mover.py
```diff
@@ -91,3 +91,18 @@ def _apply_to(self, instance, **kwds):
...
+    def _disjunct_on_active_block(self, disjunct):
+        # Check first to make sure that the disjunct is not a
+        # descendent of an inactive Block or fixed and deactivated
+        # Disjunct, before raising a warning.
+        parent_block = disjunct.parent_block()
+        while parent_block is not None:
+            if parent_block.type() is Block and not parent_block.active:
+                return False
+            elif (parent_block.type() is Disjunct and not parent_block.active
+                  and parent_block.indicator_var.value == 0):
```
jsiirola: I think you also need to test that `not parent_block.indicator_var.active`.  The block could have been deactivated by a transformation, and the indicator_var could have a current (initial) value of 0, but the `Disjunct` is still technically "active".

qtothec: Ah, I see your point. There really needs to be a better way of indicating that a disjunct has been relaxed. I think that I should check:
```
if parent_block.type() is Block and not parent_block.active and not hasattr(parent_block, 'indicator_var'):
    return False
elif (parent_block.type() is Disjunct and not parent_block.active
       and parent_block.indicator_var.value == 0 and parent_block.indicator_var.fixed):
```
Does that seem right?

qtothec: Should we be using that in the reclassify transformation, then?

qtothec: Note to self: @jsiirola said yes to this question walking by.

jsiirola:
This should be fine to merge pending successful test result.

qtothec:
@jsiirola fyi, I changed the `gdp.reclassify` transformation to use `GDP_Error`, and now other tests are failing.

qtothec:
@jsiirola Is it wrong to be using `GDP_Error` rather than `gdp.reclassify` in this case? If so, I need to figure out how to test for the existence of a `logger.error()` call.

qtothec:
@jsiirola Ok, this is ready to look at again.

pyomo/gdp/tests/test_reclassify.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def test_deactivated_parent_disjunct(self):
+        m = ConcreteModel()
+        m.d1 = Disjunct()
+        m.d1.sub1 = Disjunct()
+        m.d1.sub2 = Disjunct()
+        m.d1.disj = Disjunction(expr=[m.d1.sub1, m.d1.sub2])
+        m.d1.deactivate()
+        TransformationFactory('gdp.reclassify').apply_to(m)
+        # This should not raise an error
```
jsiirola: I somewhat loathe "tests" that have no assertions.  Can you check that the appropriate disjunctions were all ended up with the correct ctypes?  I believe you want:
```
self.assertIs(m.d1.type(), Disjunct)
self.assertIs(m.d1.sub1.type(), Disjunct)
self.assertIs(m.d1.sub2.type(), Disjunct)
```

pyomo/gdp/tests/test_reclassify.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def test_deactivated_parent_block(self):
+        m = ConcreteModel()
+        m.d1 = Block()
+        m.d1.sub1 = Disjunct()
+        m.d1.sub2 = Disjunct()
+        m.d1.disj = Disjunction(expr=[m.d1.sub1, m.d1.sub2])
+        m.d1.deactivate()
+        TransformationFactory('gdp.reclassify').apply_to(m)
+        # This should not raise an error
```
jsiirola: I somewhat loathe "tests" that have no assertions.  Can you check that the appropriate disjunctions were all ended up with the correct ctypes?  I believe you want:
```
self.assertIs(m.d1.type(), Block)
self.assertIs(m.d1.sub1.type(), Disjunct)
self.assertIs(m.d1.sub2.type(), Disjunct)
```

pyomo/gdp/tests/test_reclassify.py
```diff
@@ -0,0 +1,52 @@
...
+            TransformationFactory('gdp.reclassify').apply_to(m)
+
+    def test_active_parent_block(self):
+        m = ConcreteModel()
+        m.d1 = Block()
+        m.d1.sub1 = Disjunct()
+        m.d1.sub2 = Disjunct()
+        m.d1.disj = Disjunction(expr=[m.d1.sub1, m.d1.sub2])
+        with self.assertRaises(GDP_Error):
+            TransformationFactory('gdp.reclassify').apply_to(m)
```
jsiirola: Can we add a few extra tests that show that for this model transforming the Disjunction (with the explicit `targets=` flag will relax the Disjuncts and that reclassify will correctly (and silently) convert them to Blocks?

qtothec:
@jsiirola The behavior is not quite what you expect, since reclassify is really a transformation post-process that acts on the whole model.

PR #510, opened 2018-05-24
GAMS writer: wrap fixed vars in parentheses

### LINKED ISSUE #306: GAMS writer: invalid syntax caused by -0.0
@gseastream I have a model in which constraints like `x - -0.0 == 0.0` started to appear. This is invalid GAMS syntax, and so the model failed to solve. I don't have the Pyomo model handy right now, but can send it to you later.

gseastream:
Fixes #306 .


gseastream:
Oh you're right, it should probably be `float(val) == 1`. As for expected behavior, I don't think this overlaps with any other expression system in Pyomo since it's part of the gams writer, so we should be able to decide how gams should print fixed vars without affecting anything else, but we can wait for @jsiirola to confirm.

qtothec:
I think we are probably ok erring on the side of verbosity in generated code, so I would say go ahead and make the change to baseline to have `1.0`.

qtothec:
@gseastream Fyi, if you see that all tests on travis pass except for 1 stubborn one that failed because it couldn't access the right webserver at the right time, you can ask me to restart just that job.

gseastream:
All tests pass now, and I fixed the python API installation. The python API also no longer segfaults when run with extras (which it did last summer). Looks like someone needs to review it.

gseastream:
@jsiirola I don't think that issue is ever encountered. Consider the following model:

```python
from pyomo.environ import *
m = ConcreteModel()
m.x = Var()
m.y = Var()
m.z = Var()

m.c = Constraint(expr=m.x + m.y - -1 == m.z)
m.z.fix(-0.0)
m.o = Objective(expr=m.x)
```

With the current changes, the output file will contain:

```
c1.. x1 + x2 + 1 - (-0.0) =e= 0.0 ;
```

If we make the same parentheses change to line 58 as you referenced, it produces:

```
c1.. x1 + x2 + (1) + (-1)*(-0.0) =e= 0.0 ;
```

I don't think multiple signs are ever associated with terms anymore. They aren't really for fixed variables either, the issue is that the tree walker to_string functionality writes a minus sign because `m.z` comes after a minus sign, and then it also replaces `m.z` with `-0.0` resulting in two negatives. As far I know this is the only instance of an issue like this, and numeric constants seem to be handled comprehensively.

jsiirola:
@gseastream: A couple weird edge cases come to mind (that I have not checked):
  - Immutable scalar Params
  - Mutable Params
  - `m.x - log(m.p)` where p is (either mutable, immutable, or a fixed var) with a value < 1 (and tested with both + and -).

If those pass, then we should be good.

jsiirola:
@gseastream: the think I was wondering about was `log(0.5)`.  That said, I didn't realize that we weren't simplifying fixed expressions in the GAMS writer.  I won'er a bit about what that might do to GAMS, but as that appears to be a performance question and not a correctness question, fixing it can be deferred to a separate issue.

PR #512, opened 2018-05-24
New ShortNameLabeler, used to limit GAMS symbol names

### LINKED ISSUE #488: Symbolic solver labels causing GAMS writer to produce element names too long for GAMS
GAMS apparently has a limit on label lengths of 63 characters. We need to make sure what is generated is shorter than that.

http://yetanothermathprogrammingconsultant.blogspot.com/2015/06/gams-limits-on-labels.html

gseastream:
Resolves #488 .

Create a new labeler called `ShortNameLabeler` which can take a limit size, a prefix, a start, and even a custom labeler. If no custom labeler is provided, `AlphaNumericTextLabeler` will be used. The final labels are shortened from the output from this first labeler. This was applied to the GAMS writer in order to enforce a symbol name limit. Once caveat I though of so far was the rare possibility of a name overlap/conflict, which would only happen if the user has a 63-character component name that might look like `mycomponent_17`, and another component over 63 characters similarly named `mycomponentbutdifferent` which happens to be the 17th component with a long name so it gets shortened to `mycomponent_17`. This feels very rare to me, and even if it happened to a user, they could create their own version of a `ShortNameLabeler` with a different prefix and pass that to the writer via the `labeler` keyword.

Also let me know if the tests I added are appropriate or if something else should be done to test the new functionality. I ran it as a solver test to make sure that, by default, the GAMS writer produces output that can be successfully run by GAMS.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,7 @@ services:
 - docker
 env:
   global:
-  - secure: dC9Ihi6+5pLmJD3EtJNa/UMDRAUE+BK2QLOhD/igYvuCOVzA/blfDozDUXUKhBj4151pMRmT5HJW2tKoktAGvPMeV9N7kWEwSl5/tmOy+ATzAYcdQFylW2bYshPfyBRlK3CW4XrsF1/e0J4+L6FRIlvpAEFTtkPQQUDl/4cBE5ftR69gOKV1cEcbVzQ7zfA28TZN+J6DAR0rE11OaL1d0wRUzrxXbGVGNJ2W50R1LV+GZbRHu20qTwIGfLUVegriUXCa/xuq95upzN5/2mOIgDaLrwkpFEQDj6gUb60QDIPPi2GgNX05uaAszRWLnbroChCoK7niK2t616HvOgXRSvkcr+HRMQlwSCA1NAf3Qn5S1EMR5cekNTver9f13p8YbXzm0Hil/n8qCAAhMsKc9SAn1S6vDIdnMIUYaQ/eHPWJgGTqRkWGiaWzgb5/5QIHHE/WjJAleeV3p/pOD6myJeIqDByLCnLOhIgQeZ7fZ0d/S9+VavSoHTKfp5+ZBp/Xh7utl7wKDsSNak7Cw8miV7CntT8zTBtSiHVFWd0jvd6TEqHsFHw0AloUJu/0998kv3vk6Xj7yqpd/grZqL9DJSoBlJnHmuf/lqIYknGEUOX3ZczGnnmstc9c9TNwp8g6MD4f5cRXSLuqvW94BM7MoFh3eyx/mmCX1+uJTEpxHZs=
-  - secure: Sofev3DLx/rWCgAmNxprKOEue4AjPq6i6z8o7eHigPEx6SO3IMH36EENsuvEpE/HyNn+PHqGWudEz1ArkL+2pxiVo1FVRL94cUoo3FwachUMWMK7BnIJleJ7nW1pUwsANs0bWpw8qNkRMVRG3ycCY28MQzfx71U+/AQd+xS0EkVqDg3WWcbjVnVDRpmFH1pdm/syOHzW3jGVKTRsr1u5v9LJ+QKwEQHeKIrntP17mzyWQhZ+SbgqHcEQrK0o0sCSv4g3b/cJdazaR12H7nfxmtNjKETiLCx37OPIE0DnMKWYJVFlhbxeUfX+fPvGdjFkfu7rGnanZgsHEPy0eiXQAPfXLCwSsrJ8/RTL5RVUzQ7aK/bIQbkUr1fvqwCcS0UQOj/ajTHZYKLe19s17lgOhxtMA8CIIC8rY828uEb8hZKKSvfz37lx0yz4OePxDddZJAm1jNsfqNeofW333yjE2jQx4ASIcm29arAv+izv0S3+2RZTPSa5XdWnqw9/Ko2whgbb/Ccts/hhbCYCLqED0XjZ2OSb0cLDScd/qXJRfkW8YBraE6KMaJou10S3NCyXBYQVB9PCmT7mPH32f/pPO98JiBC0sW1Ghe285pYUJMxrhSCScw80b2Mb8847KyviH9aQt3yqtkwAl87W0YICZu0AH5hRUSO3kpaW0WbmaBA=
+  - DOCKERHUB_USERNAME=qkjqvazeqfjqftchxlgw
   matrix:
   - IMAGE_NAME=test-builds:python_3.7-rc CATEGORY="nightly"
   - IMAGE_NAME=test-builds:python_3.6    CATEGORY="nightly"
@@ -20,7 +19,6 @@ env:
   - IMAGE_NAME=test-builds:python_2.7    CATEGORY="parallel"
 
 before_install:
-  - echo "$DOCKERHUB_PASSWORD" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
   - docker pull ${DOCKERHUB_USERNAME}/${IMAGE_NAME}
  # (1) We are mounting the current directory onto the
  #     docker container and giving it a mount location with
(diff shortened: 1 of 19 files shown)


qtothec:
Are you not able to have a set of used names and detect when a collision would happen?

gseastream:
I thought about that briefly too. I got caught up in the fact that, referencing my example above, if `mycomponentbutdifferent` is encountered first, then `mycomponent_17` will be registered before the real `mycomponent_17` is labeled, and then you will have to relabel a properly short component name to something it is not. I wasn't sure, but I thought the other behavior was preferable. Alternatively, I can just send a warning if component names conflict and instruct the user to either change some names (or somehow change the order in which their components are encountered such that `mycomponentbutdifferent` is not 17th but that seems weird), or to pass their own labeler. Although now that I think of it, the warning should probably come from the labeler in which case the warning has to be generic and can't just say to pass a different labeler to the GAMS solver. Then again it could say "Do you want to try a different prefix?" or something of the sort...

qtothec:
Are there any special characters that are valid in GAMS entity names?

gseastream:
Not that I could find. GAMS is pretty strict. [Here's](https://www.gams.com/latest/docs/UG_GAMSPrograms.html#UG_GAMSPrograms_Characters) their list of legal characters. I tried all the special ones.

qtothec:
Yeah, it looks like identifiers have to conform to `[A-Za-z][A-Za-z_]{0,62}`. You could add `x_` as the separator or something like that, so that it's `mycomponentx_1`

ghackebeil:
Just out of curiosity, other than adding a warning about exceeding max name length for the GAMS writer (which could be done without a new labeler), what additional functionality does this actually provide?

If I understand correctly, it doesn't really provide some simple rule for mapping between labels and components (because the most distinguishing part of a name could be mapped to x_\<int\>). That is, you still need to know how to use the SymbolMap if you plan on parsing the GAMS file. And it also may not be particularly useful for debugging if you have long names, for the same reason.

I'm not trying to get in the way of IDEAS work. Really just curious.

I would like to raise a concern about testing. With pyomo.core.kernel and pyomo.core.expr, we are actually getting very close to nearly complete coverage of pyomo.core with simple unit tests (i.e., that do not require a thirdparty solver). It we be great if we could continue to unit test new code added to pyomo.core in this way. Would it be possible to add some simple unit tests for this new labeler?

qtothec:
@ghackebeil No problem. You raised a fair question earlier that had slipped my mind. My only objective in this was to be able to easily debug a misbehaving model. This solution came to mind because I am not familiar enough with how `SymbolMap` works. If you had a utility function that could post-process an errored `*lst` file to turn all of the `x1, x2, x3` into intelligible symbols, I would be just as happy with the result.

ghackebeil:
@qtothec: I think I see what you are getting at now. You want to turn this labeler on by default in your solver, having a guarantee that it will always produce a valid GAMS file, so that in the event of a failure you still have a file that is mostly human readable (and you don't have to re-run everything to produce it). I can see the utility in that.

I can also see how writing a function to post-process a GAMS file to convert it to one with symbolic labels could be a pain.

gseastream:
@qtothec I'm not sure making the prefix `x_` versus just `_` improves much, since you could still have the same name conflict issue if you have `mycomponentx_17` and `mycomponent_butdifferent`.

@ghackebeil to answer your question, I think this may have some use in that you can still have somewhat or mostly descriptive names for debugging / inspection, and the file will run properly. One case I can think of is if your really long component names have nothing to do with the part of the model that is causing an error, but since they're too long the model won't run. With this name shortener, you could run the file and not care if you can't determine the identity of certain symbols because that part of the model is fine. Obviously this is the not always the case, but again it will always at least provide a somewhat descriptive symbol (which may prove better than a numeric `x1084` symbol) and ensure the model file can always be run.

But I agree, to fully inspect and debug a model with long component names, it would be good to have easy and well-known access to the `SymbolMap` which I suppose people could essentially use as a translation card when looking at the model file, manually seeing an unknown symbol and looking it up in the map. Either that, or as @qtothec said a utility to parse the lst file and rewrite symbols to their real names. I wonder if this might mess up some of the references the lst errors use to point to where exactly in a particular line an error occurred, since rewriting the names (i.e. making them longer) would change where those errors point. I think the lst file uses astericks below the line to indicate this sort of thing?

A different idea I just thought of. I don't know if we want to make this the recommended method, but one could supposedly run the model with short numeric names in order to see where the errors are, and then also write a model file with full and long symbolic component names (one that will not run because the names are too long) and compare those two files to see which names refer to which components.

jsiirola:
I would like to propose a different approach to the `ShortNameLabaler`.  The current implementation has too many edge cases where the Labeler will generate label conflicts.  This also keeps the "most significant" part of a fully-qualified name (the end) and truncates the beginning.  I think with a relatively simple change we can get a label that does what we need without running an undue risk of having a label collision:
```
class ShortNameLabeler(object):
    def __init__(self, limit, prefix, start=0, labeler=None, reserve=8):
        self.id = start
        self.prefix = prefix
        self.maxID = 10**reserve
        self.limit = limit - reserve - 1 # limit + "_" + number
        if labeler is not None:
            self.labeler = labeler
        else:
            self.labeler = AlphaNumericTextLabeler()

    def __call__(self, obj=None):
        lbl = self.labeler(obj)
        if len(lbl) > self.limit:
            self.id += 1
            if self.id >= self.maxID:
                # Not good, but we will adjust the limit so the max length is honored and hope 
                # for the best.  Label conflicts will be detected and announced by the SymbolMap.
                # logger.warning("More than %s long labels; potential for label conflict exists")
                self.limit -= 1
                self.maxID *= 10
            suffix = self.prefix + str(self.id)
            return lbl[-self.limit:] + "_" + str(suffix)
        else:
            return lbl
```

gseastream:
I agree that keeping the end of the name is a much better idea than the beginning. I wonder if it might be even more useful to keep some combination of the beginning _and_ the end? Like take half from the front and half from the back, and perhaps place some characters in the middle to indicate where the text was split? Unless that's pointless.

So the other labelers don't do any checking to see if there are label conflicts, which is why I didn't implement anything like that at first. For example, `AlphaNumericTextLabeler` will create the same label for objects named `indexedvar[5]` and `indexedvar_5_` since they would both be labeled the latter. I figured this kind of overlap was just as likely as the `mycomponent_17` example I gave above. Furthermore, the same `NumericLabeler` called on the same object twice will produce different labels, so none of them are perfect mappings.

I see how your alternative with a reserve avoid label conflicts (until you need to extend the `maxID` number, but then you're making names shorter than they need to be and shortening names that shouldn't need shortening at all. Is that preferable, in the name of avoiding a super rare label conflict?

jsiirola:
@gseastream: I agree.  How about a compromise:
```
def __call__(self, obj=None):
    lbl = self.labeler(obj)
    if len(lbl) < self.limit or (len(lbl) == self.limit and lbl[-len(self.prefix):] != self.prefix):
        return lbl
    else:
        self.id += 1
        suffix = self.prefix + str(self.id) + self.prefix
        return lbl[:self.limit - len(suffix)] + suffix
```
This causes all truncated strings to be heralded (only truncated strings will end with `prefix` and be exactly `limit` characters long).  It will also truncate the minimum number of strings, and has the nice feature that we aren't preallocating an arbitrary amount of space.  We can still get name collisions, but at least the truncation process will never *create* a new collision.

gseastream:
First, perhaps it was a typo in your example, but would we still slice the `lbl` by keeping the end? Second, this seems to guard against any name conflicts I can think of. How can we still get name collisions?

jsiirola:
Doh. Yes.  I was copying from the PR and missed that it was still using the beginning of the string.  I would much prefer keeping the end of the string over the beginning.  As to your second question, we can still get collisions in the mapping from `cname` to legal GAMS characters (just like in the TextLabeler).

gseastream:
As a side comment, currently `NameLabeler` and `CNameLabeler` do the exact same thing. I imagine this is an artifact from a time when they were different, but do we still want to keep both of them?

gseastream:
While we casually wait for AppVeyor, I'll bump my earlier question about some of the other labelers. `NameLabeler` and `CNameLabeler` do the exact same thing. Is that intended? Is there a reason to have both of them?

jsiirola:
I believe that it is historical.  `Component.name` used to be `Component.cname()`.  I suspect both are there for backwards compatibility.  We should probably consider deprecating the CNameLabeler.

@ghackebeil: what do you think?

PR #479, opened 2018-05-11
Travis docker

ghackebeil:
This PR configures Travis CI to run Pyomo tests using custom DockerHub images that have been built with software we are currently re-installing for each job (e.g., with conda). This new configuration replaces most of that setup time (~5-8 minutes) with the time it takes to pull an image from DockerHub (~1 minute). It also adds working PyPy 2 and 3 builds (previously we could not set these up because of incompatibilities with conda) as well as Anaconda 2 and 3 builds.

It also includes a CPython 3.7 release candidate build. I recently discovered changes that were required in PySP for 3.7 support because of this testing (async was being used as a keyword, but it becomes a reserved word in 3.7).

Builds included:
 - CPython:
   - 3.7-rc [nightly]
   - 3.6.5   [nightly]
   - 3.6.5   [nightly, no-extras]
   - 3.6.5   [parallel]
   - 3.5.5   [nightly]
   - 2.7.15  [nightly, key_job]
   - 2.7.15  [parallel]
 - PyPy:
   - 3.5.3 (6.0.0) [nightly]
   - 2.7.13 (6.0.0) [nightly]
 - Anaconda:
    - 3.6.4 [nightly]
    - 2.7.14 [nightly]

Solvers Included:
 - ~~cplex 12.8 (full)~~
 - ~~cplexamp 12.6.3 (student)~~
 - ~~gurobi_ampl 7.5.0 (student)~~
 - cbc 2.9.9
 - ipopt 3.12.9
 - glpk 4.65
 - gams 25.0.3 (trial, python bindings available on non-PyPy 2.7 and 3.6)
 - ~~baron 18.5.9 (trial)~~

Additional software / libraries:
 - gsl 2.1
 - gjh
 - glj_asl_json

Additional notes:
 - PyPy builds are *much* slower than CPython when coverage is activated, so I've disabled coverage on those builds.
 - I have working install scripts that could insert additional commercial solvers (e.g., gurobi, cplex, baron) into the docker images, but there are licensing issues that still need to be worked out, so these are excluded for now.
 - I still own the public GitHub [repo](https://github.com/ghackebeil/build_docker) that uses Travis CI to build the Docker images as well as the private DockerHub repo that stores the images. This can be reconfigured at any point in the future if we want to place the two repositories under some official Pyomo accounts.

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
[...]

PR #514, opened 2018-05-26
Resolve GDP errors with Pyomo5 expressions

jsiirola:
## Summary/Motivation:
This fixes GDP so that expression replacement happens correctly and all the tests pass.  This should return Pyomo to a fully-functional master branch.

This PR has two significant changes outside of `pyomo.gdp`:
1. A rework of the Expression Replacement walkers.  This was motivated by an observation that in every case I could think of, the user would want named expressions to be removed from the new expression tree (that it, the new tree should not be entangled with other trees.  That option has now been propagated up the the base replacement walker, although the default is still to leave the named expressions in place.  We can revisit changing the default in a future issue / PR.
2. Remove the use of custom walkers when cloning expressions.  This significantly simplified the cloning logic and made expression duplication more robust (in that, barring stack overflows, the code relies on core Python to correctly duplicate complicated data structures).  This however changes a design assertion in Pyomo5 expressions.  If examples crop up where expression manipulation leads to stack overflows, then we can revisit this decision.

- Update the GDP tests to positively test the resulting relaxed expressions
- Rework the expression replacement walkers across Pyomo
- Remove the use of custom walkers when cloning expressions.  

### Legal Acknowledgement


qtothec:
Merged in master because of travis configuration changes from #479, to get tests to pass.

PR #513, opened 2018-05-25
Refactor of GDPopt to use config blocks

qtothec:
## Summary/Motivation:
Use Config Blocks

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #519, opened 2018-05-28
Address comments raised for GDPopt in #513

qtothec:
- Update some ConfigValue domains
- Tweak logging behavior
- Update GDP model to use new syntax

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/gdpopt/GDPopt.py
+++ b/pyomo/contrib/gdpopt/GDPopt.py
@@ -126,7 +126,7 @@ class GDPoptSolver(pyomo.common.plugin.Plugin):
     CONFIG.declare("tee", ConfigValue(
         default=False,
         description="Stream output to terminal.",
-        domain=In([True, False])
+        domain=bool
     ))
     CONFIG.declare("logger", ConfigValue(
         default=logging.getLogger('pyomo.contrib.gdpopt'),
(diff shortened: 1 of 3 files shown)


pyomo/contrib/gdpopt/GDPopt.py
```diff
@@ -287,7 +287,7 @@ def solve(self, model, **kwds):
         solve_data.results.problem.lower_bound = solve_data.LB
         solve_data.results.problem.upper_bound = solve_data.UB
 
-        if config.tee:
+        if config.tee and old_logger_level > logging.INFO:
             config.logger.setLevel(old_logger_level)
```
jsiirola: There is a subtle issue here where an exception raised by GDPopt will cause the logger level to not be restored.  A slightly more robust approach would be:
```
old_logger_level = config.logger.getEffectiveLevel()
try:
    if config.tee and old_logger_level > logging.INFO:
        # If the logger does not already include INFO, include it.
        config.logger.setLevel(logging.INFO)
    # ........
finally:
    # No need to test tee ... just unconditionally restore the log level.
    config.logger.setLevel(old_logger_level)
```

jsiirola:
Looks good.  We can merge pending successful tests.

PR #529, opened 2018-05-31
GAMS Solver: Allow check_expr to warn user about zero division too

### LINKED ISSUE #527: GAMS solver initial expression evaluation checking
When GAMS is run on a model, it requires that all expressions evaluate at their initial values. So, for example, if you don't warmstart a variable `x` and then you have a constraint that contains `log(x)`, then GAMS will have an `Execution error` because `x` is initially 0 by default and GAMS can't evaluate `log(0)`. The same is true for division by zero. To aid in this type of error, if GAMS fails a solve, then the GAMS solver will proceed to try to evaluate every expression in the model at their initial or default values, and throw a helpful warning if it finds an expression it thinks GAMS failed on. The current code to do this is as follows:

```python
try:
    value(expr)
except ValueError:
    logger.warning("While evaluating model.%s's expression, GAMS solver "
                   "encountered an error.\nGAMS requires that all "
                   "equations and expressions evaluate at initial values.\n"
                   "Ensure variable values do not violate any domains, "
                   "and use the warmstart=True keyword to solve()." % name)
    if solver_io == 'shell':
        # For shell, there is no previous exception to worry about
        # overwriting, so raise the ValueError.
        # But for direct, the GamsExceptionExecution will be raised.
        raise
```

This works for `log` and `log10`, since errors in evaluating those raise `ValueError`, but I just found out for expressions that try to divide by 0, it raises `ZeroDivisionError`, so it was not caught and the helpful warning was not sent. This will be a simple fix, but I wanted to ask, should we change the `except` to catch any error that arises from `value(expr)` (i.e. put nothing after `except`)? Or change it to include exactly `ValueError` and `ZeroDivisionError`? I wasn't sure if we wanted to anticipate more situations like this where a different kind of error we weren't expecting showed up, but I also didn't want to stomp out some other error that actually means something else went wrong not having to do with domain issues.

gseastream:
## Summary/Motivation:
See issue.

- Allow GAMS interface to also warn about zero division issues in initial variable value evaluation.

### Legal Acknowledgement


qtothec:
This begs the question: should `warmstart=True` by default?

gseastream:
Actually `warmstart=True` is already the default (see [line](https://github.com/Pyomo/pyomo/blob/dab6c8d0213ec6bd42667ed96b3801c3b321d522/pyomo/repn/plugins/gams_writer.py#L206) in `gams_writer.py`). This was updated at some point and the docstring in the writer was updated but not in the solver interfaces, where the writer section of the docstring is repeated. Now this begs the question, should the solver docstrings not include the writer `io_options`? If we took it out of the solver docstrings then it should show up somewhere else in the sphinx documentation, although I recognize that whole thing is getting overhauled anyway.

qtothec:
Even though things are due for an overhaul, I think you have enough of an audience with this tool to make the intermediate changes.

jsiirola:
Agreed.

gseastream:
So would you guys like me to remove the `io_options` section of the solver docstrings and just rely on the writer's docstring (and then add a section in the sphinx docs to show the writer doc)? Or leave it as is (and update the solver docstrings since they're wrong at the moment)?

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -153,36 +153,8 @@ def solve(self, *args, **kwds):
         report_timing=False: bool
             Print timing reports for presolve, solver, postsolve, etc.
         io_options: dict
-            Options that get passed to the writer.
+            Options that get passed to the writer. See writer for details.
```
qtothec: See writer in `pyomo.repn.plugins.gams_writer` for details?

gseastream: Well if looking at the sphinx page, the writer documentation is right below it which is why I put that referring sentence in there. Would the full path the writer be more helpful?

qtothec: I was thinking in the context of someone doing `help(SolverFactory('gams'))`

gseastream: Fair enough

PR #534, opened 2018-06-02
Updating the README.md File

whart222:
## Summary/Motivation:
We need to add a link for performance plots somewhere...

- Removed reference to CPython 2.6 tests
- Added a link for the performance plots.

### Legal Acknowledgement


whart222:
It's not obvious that we're ready to make these performance plots public, but this PR will force that discussion about what needs to happen next.

gseastream:
I think more documentation on the performance plots page would be useful. I think I understand how it all works now, but I feel the page itself is currently not very intuitive, so describing why things are labeled slow vs fast and what the graph is showing would be helpful. Also, would it make more sense to flip the graph so that it starts with the latest release and nightlies are to the right?

I am also still confused as to why the GAMS solver seems to be much slower on most of its tests than the latest release timings, but maybe this has nothing to do with the plots...

whart222:
@gseastream So it would be more intuitive to have old data points on the right?

I'm also surprised by the GAMS differences.  This is a consequence of the release update.  I'm not sure what is the root cause for that, and I haven't dug into that code recently.

jsiirola:
It might be more intuitive to have the last release on the left; however, it is more useful to have the last release next to the most recent result: that is the comparison we are most interested in and helps avoid slow drift away from the last release.

I would like to see the text (OK, SLOW, FAST, etc) on the main page replaced with the actual percentage differences.  Since the color coding is synonymous with the text, we could include useful (non-redundant) information by including the number.

jsiirola:
More germane to this PR, why are you putting the performance plots links on the main github page?  I would think it should be posted in a "developer information" section...

whart222:
I routinely look at the websites of codes I use to understand whether they are doing performance testing.  This information is not just for developers.

jsiirola:
I like the idea of the line for the most recent release.

whart222:
I thought about using a line for the latest release, and I thought it wouldn't work.  But I think that was for a different plot.  I'll update the plots.

PR #540, opened 2018-06-05
Obtain column names from `cursor.description()` in sql query

### LINKED ISSUE #537: Obtain column names from `cursor.description()` in sql query
Dear Pyomo team:

I came across legacy issue [#4480](https://software.sandia.gov/trac/pyomo/ticket/4480) and I would like to ask whether it is possible to "revive" it. I have tested and had success with the suggestion proposed in the legacy ticket of removing the code that extracts column names from the data file query, or fetches them from the database with a `SELECT *` type statement. I have removed that code and replaced it thus with a statement that obtains column names from the cursor. The portion of the code in question in `core/plugins/data/db_read.py` (for Pyomo 5.5.0) reads as follows:
```python
def read(self):
        #
        # Get the table from the database
        #
        if self.db is None:
            return
        cursor = self.db.cursor()
        tmp = []
        if self.options.query is None:
            if self.options.table is None:
                raise IOError("Must specify 'query' or 'table' option!")
            self.options.query = 'SELECT * FROM %s' % self.options.table
        elif self.options.query[0] in ("'", '"'):
            self.options.query = self.options.query[1:-1]
        cursor.execute(self.options.query)
        try:
            rows = cursor.fetchall()
            for col in cursor.description:
                tmp.append(col[0])
            tmp=[tmp]
        except: # subsequent code remains the same in db_read.py
```

The reason for this request is that I do not have control over the name of the columns in the database I need to access, and renaming sets and variables in pyomo to match results in code that is less readable.

In fact when using pyomo with the modified functionality as above, 
I am able to simply put an alias on all the columns I need. 
For example: a query to query for the value of a parameter `P` that is indexed over 3 sets `A`. `B` and `C` would read:
```sql
SELECT database_column_name1 as A, database_column_name2 as B, 
database_column_name3 as C,database_column_name4 as P
FROM dabatase_table
WHERE ... 
```
 
Best regards

jotaigna:
## Summary/Motivation:
To implement legacy request  [#4480](https://software.sandia.gov/trac/pyomo/ticket/4480) which proposes to obtain column names in an SQL query from `cursor.description()` as opposed to parsing the query request from the .dat file. This enables the use of alias names for columns in databases whose names cannot match model declarations.

- Eliminate the parsing of column names from query in `pyomo.core.data.db_table.read`
- Execute the query inside an existing try statement to catch SQL malform expressions non-silently
- If successful, obtain returned column names from `cursor.description` and use subsequently to populate model
- Raise an explicit exception in `pyomo.core.data.TableData._set_data` method to catch when returned column names do not match model declarations (e.g. when an alias is required)

## Alias example
For example: a query to obtain the value of a parameter `P` that is indexed over 3 sets `A`. `B` and `C` would read:
```sql
SELECT database_column_name1 as A, database_column_name2 as B, 
database_column_name3 as C,database_column_name4 as P
FROM dabatase_table
WHERE /*more sql code here*/
```

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/data/TableData.py
+++ b/pyomo/core/data/TableData.py
@@ -2,8 +2,8 @@
 #
 #  Pyomo: Python Optimization Modeling Objects
 #  Copyright 2017 National Technology and Engineering Solutions of Sandia, LLC
-#  Under the terms of Contract DE-NA0003525 with National Technology and 
-#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain 
+#  Under the terms of Contract DE-NA0003525 with National Technology and
+#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain
 #  rights in this software.
 #  This software is distributed under the 3-clause BSD License.
 #  ___________________________________________________________________________
@@ -120,7 +120,10 @@ class TableData(Plugin):
                 header_index.append(i)
         else:
             for i in self.options.select:
-                header_index.append(headers.index(str(i)))
+                try:
+                    header_index.append(headers.index(str(i)))
+                except ValueError as e:
+                    raise Exception("Model declaration '%s' not found in returned query columns" %str(i)) from e
         self.options.ncolumns = len(headers)
 
         if not self.options.param is None:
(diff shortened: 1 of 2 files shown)


whart222:
The appveyor test failure is a known issue with Python 3.5, so I'm resolving this PR.

PR #545, opened 2018-06-05
Fix for simulating discretized models

### LINKED ISSUE #531: Simulator fails when passed a discretized model
This is caused by the Simulator identifying DerivativeVar components based on their ctype which is reclassified to Var after applying a discretization.

@jsimkoff is working on a bug fix

jsimkoff:
## Summary/Motivation:
Fixes bug which causes simulation failure if simulator is passed a discretized model

- Check Vars to see if their python type is DerivativeVar and add to derivs list if true.

### Legal Acknowledgement


blnicho:
@jsimkoff I talked with @jsiirola about this fix last week and he made some good points about trying to avoid using `isinstance` calls and relying instead on ctypes and book-keeping in the discretization. For that reason, I'm going to close this PR and we can talk later today about the alternative fix for this.

PR #547, opened 2018-06-06
[WIP] - Adding check for valid component names

### LINKED ISSUE #525: Rethinking Pyomo component string representations
Yesterday, I discussed a preliminary extension of VarList to include per-variable labels.  This would allow me to label variables as I add them to the list.

This conversation raised some concerns that the users would not be able to associate labels with the location of the variable in a model.  I noted that this extension really isn't logically different from Pyomo capabilities that already exist, where we can assign variables on a model with an arbitrary name.  For example:
```
from pyomo.environ import *

m = ConcreteModel()
setattr(m, 'some wierd.name', Var())
setattr(m, ' ', Var())

m.pprint()
```
Which produces output like:
```
2 Var Declarations
      : Size=1, Index=None
        None :  None :  None :  None : False :  True :  Reals
    some wierd.name : Size=1, Index=None
        None :  None :  None :  None : False :  True :  Reals

2 Declarations: some wierd.name  
```

Note that the setattr() method is being used in a manner that makes it impossible to identify the original variable component.  In fact, I didn't even know that I could use a space as an attribute name until I tried  this.  But more generally, we can add arbitrary symbols in attribute names.

I think this is generally bad, but I'd like to assess everyone's thoughts on this.  Should we change setattr() semantics?  Should we require that string representations provide exact correspondence to the location of components in models?  If not, then what names are reasonable and appropriate?  Or are any allowable?

whart222:
Changed the logic for adding slacks to use a Var(Any) component.

## Summary/Motivation:
Adding a check to ensure that component attribute names are valid

- Checking that component attribute names are valid
- Reworking transformations that used '.'s in attribute names

### Legal Acknowledgement


gseastream:
Would you mind pulling from master to try out the new testing on branches restriction? And then if that works we ideally pull the update into all the main repo's branches.

whart222:
@gseastream I've merged from master.

whart222:
Per the comment in #525, I'm closing this PR due to inactivity.

PR #551, opened 2018-06-07
Fix pyomo help -s

gseastream:
## Summary/Motivation:
See issue. Simple fix so that the `pyomo help -s` command runs properly again.

Note that the GAMS solver breaks when you try to use the `pyomo solve model.py --solver=gams` command, because it doesn't have a `config_block` method, which it would if it inherited from `OptSolver`. I got confused quickly when I tried to understand how those worked for the solvers.

## Changes proposed in this PR:
- Add `_metasolver=False` attribute to GAMS and GDPopt solvers
- Fix `logging` vs `logger` typo causing a bug
- Make some of the doc strings longer

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #558, opened 2018-06-11
Integrate options attribute into GAMS solver

gseastream:
## Summary/Motivation:
This was initially motivated by metasolver and subsolver discussion in #551, since GAMS was unable to support the `:` notation when creating a solver instance. Then I found out about the `py` solver, and figured I should fix GAMS so the API solver can be created with `"py:gams"`, and then I decided to fully implement the `options` attribute that was only put there as a placeholder before. Anyway, now one can set options on the solver instance and they'll persist across runs, and will be overwritten by keywords passed to the `solve` method. Additionally, as a somewhat intended side effect, if you call `SolverFactory("gams:ipopt")` or some other subsolver, it will return a GAMS solver whose `solver` option is already set to `"ipopt"`.

- Integrate `options` attribute into GAMS interface
- Aggregate common GAMS methods into superclass
- Don't set `"solver=subsolver"` option on the solver if coming from `py` solver

### Legal Acknowledgement


qtothec:
As discussed at the meeting today, would be nice to get a couple test cases with the `gams:solver` notation.

gseastream:
Looks like the AppVeyor problem has propagated to Python 2.7 and Python 3.6 now, and something on the Pyro parallel tests failed on travis.

gseastream:
@blnicho @whart222 @jsiirola Does anyone want to address this pull request? I believe it's ready to go, and I think some people are interested in this functionality.

qtothec:
@ghackebeil I'm not sure exactly what you mean by
> conflates a solver option with a solver interface
Could you explain? In my mind, the GAMS "solver" is properly an interface.

gseastream:
@ghackebeil Do we support `SolverFactory("ipopt", solver_io="gams")`? I'm not aware of that functionality.

As for the `:` notation, I don't know much about the history of it, but from what I can tell it basically does the same thing as other solver interfaces that support subsolvers. If we decided that we didn't want to support the `:` notation for the GAMS solver, then we would need to hardcode some new things in now, because with this PR, the GAMS interface now supports the `options` attribute, which means if we want to prevent the `:` notation we cannot allow the solver constructor to set the `solver=subsolver` attribute on the new solver, since this is a meaningful option to the GAMS writer (the other solver interfaces end up silently ignoring this option if the `:` notation is used because they don't use an option named `solver`, but GAMS does). Previously the `:` notation was prevented on the GAMS interface only because it did not have a `set_options` method, but now in this PR it does because it fully supports the options attribute.

ghackebeil:
I think that due to the way the solver API is structured it is easy for us get crossed definitions for a Pyomo solver and solver interface. I hope that can be fixed this summer.

Similar to our ASL solver plugin, I don't think of the GAMS solver plugin as a "solver", I think of it as an interface to solvers. The solver plugins that are there are just thin wrappers that around the NL writer and GAMS writer.

I do agree that the GAMS solver plugin (whatever you want to think of it as) should have an options dictionary, but I don't think the SolverFactory should be structured to hijack something in the options dictionary for internal Pyomo usage. The ASL plugin does the same thing in order to support the syntax "asl:ipopt". As a result, you have this strange looking code in ASL.py that pulls these entries back out of the options dictionary:
```python
        for key in self.options:
            if key == 'solver':
                continue
            if ...
                ...
            elif key == 'subsolver':
                opt.append("solver="+str(self.options[key]))
                cmd.append(str(key)+"="+str(self.options[key]))
            else:
                ...
```
Now what happens if a real solver that uses the ASL (i.e., uses NL files as input) has a real option named "solver" or "subsolver". We silently pulling that out the options dictionary, even though a user may have put them there on purpose.

I expect that hack will go away when we rewrite the Pyomo solver API, mainly because there is no need to use

`SolverFactory("asl:<solver_name>")`

You can simply use

`SolverFactory("<solver_name>",solver_io="nl")`

(side note: perhaps it would have been better to use `solver_io="asl"`).

So I guess what I'm saying is, if you can make

`SolverFactory("<solver_name>", solver_io="gams")`
`SolverFactory("<solver_name>", solver_io="gams_py")`

work (and not hijack the options dictionary to do it), I think that would be a better long term decision.

Thoughts?

qtothec:
@ghackebeil I see your point.

For what it's worth, I don't have a strong preference for one convention versus another, so long as it's well-documented.

gseastream:
I see your points. And yes, in this case `solver` is a GAMS option, which is why it makes sense to put the option in the options dictionary. But there are currently several ways to specify the `solver` option for the GAMS writer to use:

```python
opt = SolverFactory("gams")
opt.solve(m, solver="baron")
opt.solve(m, io_options={"solver": "baron"})
opt.options["solver"] = "baron" # this is possible with this PR
```

Like I alluded to in the PR description, being compatible with the `:` syntax is a side effect of introducing compatibility with an options dictionary, and this side effect is meaningful because `solver` is a meaningful option. But doing `SolverFactory("gams:baron")` or `SolverFactory("gams", solver="baron")` is not at all the only way to set the `solver` option to the writer, nor does it need to be the recommended way to do so. And since this is a side effect I didn't do anything to purposely support, we would have to change the SolverFactory instantiation code to not make use of the `:` notation, which may inadvertently affect some functionality unless we know when and when not to support that syntax, but to me that feels like extra effort to not support functionality that would have worked kinda as expected anyway. As Qi said, this can be documented too if desired so that people know exactly what they're doing. And while I want to say that no one would probably even think to do this, I got asked a question today on how to specify the `solver` to use and their first guess was to do `SolverFactory("gams", subsolver="baron")`.

As for doing `SolverFactory("ipopt", solver_io="gams")`, this is very different functionality from how SolverFactory currently works, since currently, to get the Python API GAMS solver, you do:

`SolverFactory("gams", solver_io="python")`

And you can replace `python` with `direct` for the same result, or use `shell` or `gms` to specify the shell solver (which is the default). Thus, to support the use you mentioned, we'd have to change the way everyone interprets the `solver_io` argument, since currently the code above would create an instance of an ipopt solver and probably (hopefully) complain because "gams" is not a recognized version of the ipopt solver. I mean I suppose you could hardcode in to check if `solver_io` is "gams" or "gams_py" and then proceed from there but that doesn't feel like code that supports this general use case.

whart222:
Since I'm responsible for introducing most of the APIs being discussed, let me admit that I've never been particularly happy with notations like "solver:subsolver" or SolverFactory("gams", solver_io="python").  These were hacks to get something working, and I always expected to revisit this later when we had more experience with hybrid solvers, etc.

I'm having trouble following this conversation in this PR, but I think this is a case where a design discussion separate from a PR would make sense.  I'm not sure if there's an immediate need for changes, but if not there are some deeper issues to be considered here regarding how solvers are declared, solver options are specified, intermediate file formats are defined, etc.

gseastream:
I had hoped that I wasn't getting too off track, but to summarize, my understanding of this discussion is that we aren't sure what we think about the `solver:subsolver` notation but I agree that appears to be a topic for a separate design discussion since the real change proposed in this PR is to integrate the options dictionary, and the whole subsolver thing was just along for the ride. That being said, I did add tests for using the subsolver notation since it felt like behavior I was happy to support but if we don't want anything to do with the subsolver notation anymore I can take those tests out. This still wouldn't change the fact that doing `"gams:baron"` would set `options["solver"] = "baron"`; we can only stop that by changing the solver api.

PR #562, opened 2018-06-13
Initial stab at resolving #31

### LINKED ISSUE #31: Bug using numpy float values in Pyomo expressions.
Consider the following script:

```
# pyomo 4.4.1
# python 3.5
from pyomo.environ import *
import numpy

m = ConcreteModel()

m.T = RangeSet(10)
m.v = Var(initialize=1, bounds=(0,None))
m.c = Var(m.T, initialize=20)

def rule(m, t):
    h = numpy.float32(1.0)
    # This works
    return m.c[t]  == h * m.c[t]
m.X = Constraint(m.T, rule=rule)

def rule(m, t):
    h = numpy.float32(1.0)
    # This works
    return m.c[t]  == h * m.c[t] * m.v
m.X = Constraint(m.T, rule=rule)

def rule(m, t):
    h = numpy.float32(1.0)
    # This fails
    return m.c[t]  == h * m.v
m.x = Constraint(m.T, rule=rule)
```

The first two constraints are constructed fine, but the last one fails, generating the following error:

```
ERROR: Rule failed when generating expression for constraint x with index 1:
    KeyError: "Error accessing indexed component: Cannot treat the scalar component 'v' as an array"
ERROR: Constructing component 'x' from data=None failed:
    KeyError: "Error accessing indexed component: Cannot treat the scalar component 'v' as an array"
  File "tmp.py", line 26, in <module>
    m.x = Constraint(m.T, rule=rule)
  File "/Users/wehart/home/src/python35/src/pyomo/pyomo/core/base/block.py", line 483, in __setattr__
    self.add_component(name, val)
  File "/Users/wehart/home/src/python35/src/pyomo/pyomo/core/base/block.py", line 849, in add_component
    val.construct(data)
  File "/Users/wehart/home/src/python35/src/pyomo/pyomo/core/base/constraint.py", line 736, in construct
    ndx)
  File "/Users/wehart/home/src/python35/src/pyomo/pyomo/core/base/misc.py", line 59, in apply_indexed_rule
    return rule(model, index)
  File "tmp.py", line 25, in rule
    return m.c[t]  == h * m.v
  File "/Users/wehart/home/src/python35/src/pyomo/pyomo/core/base/indexed_component.py", line 499, in __getitem__
    raise KeyError(msg)
KeyError: "Error accessing indexed component: Cannot treat the scalar component 'v' as an array"
```

whart222:
## Summary/Motivation:
Removing the use of len() for non-indexed components.

- Adding tests for the use of numpy data in expressions
- Various changes because len(component) returns TypeError unless the component is indexed.

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/PyomoModel.py
+++ b/pyomo/core/base/PyomoModel.py
@@ -609,11 +609,11 @@ use the AbstractModel or ConcreteModel class instead.""")
         self.statistics.number_of_objectives = 0
         for block in self.block_data_objects(active=active):
             for data in self.component_map(Var, active=active).itervalues():
-                self.statistics.number_of_variables += len(data)
+                self.statistics.number_of_variables += len(data) if data.is_indexed() else 1
             for data in self.component_map(Objective, active=active).itervalues():
-                self.statistics.number_of_objectives += len(data)
+                self.statistics.number_of_objectives += len(data._data)
             for data in self.component_map(Constraint, active=active).itervalues():
-                self.statistics.number_of_constraints += len(data)
+                self.statistics.number_of_constraints += len(data._data)
 
     def nvariables(self):
         self.compute_statistics()
--- a/pyomo/core/base/block.py
+++ b/pyomo/core/base/block.py
@@ -1233,7 +1233,7 @@ Components must now specify their rules explicitly using 'rule=' keywords.""" %
             if comp.is_indexed():
                 _items = comp.iteritems()
             # This is a hack (see _NOTE_ above).
-            elif len(comp) or not hasattr(comp, '_data'):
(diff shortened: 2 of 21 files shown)


whart222:
This is a work-in-progress.  I think that removing the use of len() for non-indexed components is the right thing to do in Pyomo.  But the design of many non-indexed components allows for 'uninitialized' singletons (which is a bad idea).  Hence, I've introduced too many references to len(component._data) throughout.  Perhaps we need a different method to return the _data value?

whart222:
Note:  I haven't yet added tests to confirm that numpy data types work with this PR.  But I've confirmed that this works in practice.

whart222:
Closed per my comments in #31

PR #568, opened 2018-06-15
Implicitly ordered sets

whart222:
## Summary/Motivation:
If we order sets in Pyomo based on insertion order, then the set doesn't appear ordered and comparable to the user, but we can avoid sorting and perform other optimizations within Pyomo.

- This just makes ordering the default for Set() objects.
- There's more to be done here to exploit this feature.  For example, we can avoid sorting within the problem writers.
- I think we should change the semantics of 'ordered' such that ordered=True implies that the set members are sorted.

### Legal Acknowledgement


whart222:
I'm frankly surprised that this small change still passes all of the tests.  This suggests that are tests w.r.t. unordered sets in Pyomo are weak.

carldlaird:
From the PR:

> I think we should change the semantics of 'ordered' such that ordered=True implies that the set members are sorted.

I don't agree. If we want this behavior, a keyword like "sorted=True" seems to make more sense to me. However, I don't think this behavior is implemented in the PR, so I don't have any complaints.

carldlaird:
Regarding weak tests, I am not sure what a test for an unordered set would look like. If the code works with them unordered, it should also work with them ordered. The fact that the tests pass doesn't surprise me too much.

jwatsonnm:
I agree with Carl's disagreement. The semantics for ordered=True are different from sorted=True. And various users and libraries (including PySP) rely on ordered=True enforcing that sets are maintained in a user-defined order - they act as lists. For example, time stages in a stochastic program are an ordered set, e.g., "TheFirstStage" and "NextStage". If they were ordered, the semantics would be broken.

jsiirola:
And for the record, I am not at all surprised that the tests all pass. Since we expect the tests to pass on all versions of Python, Pyomo does a lot of sorting to ensure determinism. Plus, since he ordering generally only changes when you change Python versions, how would we test for unorderedness?

whart222:
After thinking about this and the response to my thoughts about 'ordered=True', I realized that we shouldn't change the default value of 'ordered'.  Instead, we should change the storage mechanism when 'ordered=False'.  I've commit these changes, which should be reviewed.

The basic idea is that even when the user doesn't specify 'ordered=True', the insertion order is preserved and used when iterating over the set.  When specifying 'ordered=True', not much more happens;  there's an API change allowing methods like next() and prev(), since the user has indicated that these are valid operations on their set.

PR #569, opened 2018-06-15
Kernel Update 1: Simplify handling of active status on containers

ghackebeil:
This PR simplifies how the active status is managed on the simple kernel containers (dict, list, tuple). 

Previously, the implementation used complex logic to make the active status of the container depend on whether or not active children where contained within it, while also giving the container an extra degree of freedom to be active or inactive when empty. When `activate()` or `deactivate()` was called on the container, this status change was propagated down to all children of the container. This differed from the default behavior of blocks, where the active status is managed independently from the children, and calling `activate()` or `deactivate()` does not change the status of anything under the block. The behavior of blocks can be changed by changing either call to look like `activate(shallow=False)` or `deactivate(shallow=False)`, which propagates the status change to everything stored below the block. It may be worth discussing changing this keyword to something like `depth={0,1,...}` to allow controlling how far down the status change is propagated.

All containers in kernel now behave the same as blocks.

Many lines of code were deleted in this PR, including a mixin class, which should make it easier for developers to begin understanding the kernel base class logic.

A simple example of how this would change the behavior of (likely rarely used) kernel code:
```python
d = constraint_dict()
print(d.active) # -> True
d.deactivate()
print(d.active) # -> False
c = constraint()
print(c.active) # -> True
d[0] = c
```
Before this PR, after the final line, `d.active` would become `True` (since an active child was added to it). After this PR, `d.active` would remain `False`.

### Legal Acknowledgement


qtothec:
Does it make any sense for a variable_dict to be inactive?

ghackebeil:
@qtothec: This PR doesn't make any changes related to that discussion. In kernel, `variable`, `variable_dict`, etc. still do not have an active status. However, I can see some utility in allowing groups of variables to be deactivated without forcing them to live on an inactive block (for performance / debugging reasons). E.g., I've deactivated a set of constraints and I think these variables are no longer used, so if I deactivated them hopefully I get an error or something if they are used. 

It is hard to go into this discussion without going down the rabbit hole on how to collect "active" variables and what should be viewed as "the model", which we've had in other threads. We should probably keep that discussion out of this PR.

ghackebeil:
@qtothec: My bad. I see why you asked that question now. I used a bad example in the PR description. I'll fix it.

PR #579, opened 2018-06-20
Model size report utility

qtothec:
## Summary/Motivation:
Adds a model statistics reporting utility, along with a working definition of what it means to be "active" in the context of Pyomo.GDP.

### Legal Acknowledgement


pyomo/util/model_size.py
```diff
@@ -0,0 +1,155 @@
...
+            Disjunct, descend_into=block_like))
+    report.overall.constraints = sum(
+        1 for c in model.component_data_objects(
+            Constraint, descend_into=block_like))
+    report.overall.nonlinear_constraints = sum(
+        1 for c in model.component_data_objects(
+            Constraint, descend_into=block_like)
+        if c.body.polynomial_degree() not in (1, 0))
+
+    report.warn = Container()
```
jsiirola: I think I like "warnings" as opposed to "warn"

pyomo/util/model_size.py
```diff
@@ -0,0 +1,155 @@
...
+        active_constraints.update(new_active_constraints)
+
+    active_vars.update(
+        var for constr in new_active_constraints
+        for var in EXPR.identify_variables(
+            constr.body, include_fixed=False))
+    active_vars.update(
+        disj.indicator_var for disj in active_disjuncts)
+
+    report.active = Container()
```
jsiirola: I worry about using "active" because of the potential of confusion between "active constraints" (where value(body) is at a bound) and "activated constraints" (i.e., not deactivated).  What do you think of renaming this "activated"?

pyomo/util/tests/test_model_size.py
```diff
@@ -0,0 +1,147 @@
...
+        m.z.fix(3)
+        m.c = Constraint(expr=m.x ** 2 == 4)
+        m.c2 = Constraint(expr=m.x / m.y == 3)
+        m.c3 = Constraint(expr=m.x * m.z == 5)
+        m.c4 = Constraint(expr=m.x * m.y == 5)
+        m.c4.deactivate()
+        model_size = build_model_size_report(m)
+        self.assertEqual(model_size.active.nonlinear_constraints, 2)
+        self.assertEqual(model_size.overall.nonlinear_constraints, 3)
+
```
blnicho: Could you add a test with indexed variables and constraints?

pyomo/util/model_size.py
qtothec: That makes sense.

pyomo/util/tests/test_model_size.py
qtothec: Yes, in the process of activating the previously skipped eight-process-problem test.

PR #582, opened 2018-06-20
Improper basic steps implementation

qtothec:
## Summary/Motivation:
This PR introduces the ability to handle improper basic steps to the basic steps implementation in Pyomo.GDP. Tests are also added to exercise this ability.

Depends on #581 because of testing example.

### Legal Acknowledgement


pyomo/gdp/basic_step.py
```diff
@@ -26,10 +30,23 @@ def _pseudo_clone(self):
...
 
 
-def apply_basic_step( disjunctions ):
+def _squish_singletons(tuple_iter):
+    """Squish all singleton tuples into their non-tuple values."""
+    for tup in tuple_iter:
+        if len(tup) == 1:
+            yield tup[0]
+        else:
+            yield tup
```
jsiirola: Why not use `flatten_tuple`?

pyomo/gdp/basic_step.py
```diff
@@ -57,12 +74,20 @@ def apply_basic_step( disjunctions ):
...
                     continue
                 tmp.del_component(k)
                 ans.disjuncts[idx].src[i].add_component(k,v)
+        # Copy in the constraints corresponding to the improper disjunctions
+        ans.disjuncts[idx].improper_constraints = ConstraintList()
+        for constr in constraints:
+            for indx in constr:
+                ans.disjuncts[idx].improper_constraints.add(
+                    (constr[indx].lower, constr[indx].body, constr[indx].upper)
+                )
```
jsiirola: Does it make sense to give the user an option to deactivate the original constraint after applying the basic step?

qtothec: Do I not deactivate the original constraint? If so, that's an oversight on my part.

pyomo/gdp/basic_step.py
qtothec: Actually, looking at what `flatten_tuple` does, we only want to flatten it if it's a singleton. That is, `((0, 0), (0, 1), (0, 2))` is fine, but when only one disjunction is participating, it becomes `((0,), (1,), (2,))`, which is problematic as input to `initialize=` for `Set(dimen=1)`.

PR #485, opened 2018-05-15
Add quotes around paths in cplex script file

ghackebeil:
Avoids issues with spaces in path names.

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #586, opened 2018-06-22
Additional fixes for spaces in cplex file names

### LINKED ISSUE #480: Blank space in project path leads to wrong file type in CPLEX
When path to python file is '/Users/XXX/Desktop/…/Pyomo Spam/eggs.py' 
(blank space after …/Pyomo)

CPLEX will throw an error containing:

`Logfile '/Users/XXX/Desktop/…/Pyomo' open.
CPLEX> Cannot interpret file type.`

when path is  '/Users/XXX/Desktop/…/Pyomo_Spam/eggs.py'  it works.

But when path to python file is '/Users/XXX/Desktop/…/Pyomo Spam/e g g s.py' so when the file itself contains the spacing, no problems occur.

jsiirola:
## Summary/Motivation:
This resolves a regression introduced in #485 that broke our CPLEX interface for versions prior to 12.8.  Prior to 12.8, cplex did not allow quote characters in file names.

- Only quote file names if there are spaces in the name *and* the CPLEX version is >= 12.8
- Raise an (informative) exception when spaces are detected in file names with older versions of cplex. 

### Legal Acknowledgement


qtothec:
We may want to consider emitting a warning when there's non-alphanumeric characters in the path name for CPLEX: https://stackoverflow.com/questions/50939777/cant-use-cplex-as-solver-when-using-pyomo/50941197

pyomo/solvers/plugins/solvers/CPLEX.py
```diff
@@ -185,6 +185,17 @@ def _presolve(self, *args, **kwds):
...
+                if self.version()[:2] >= (12,8):
+                    if '"' not in self._warm_start_file_name:
+                        self._warm_start_file_name \
+                            = '"'+self._warm_start_file_name+'"'
+                else:
+                    raise ValueError(
+                        "Space detected in CPLEX warm start file path/name and "
+                        "CPLEX older than version 12.8.  Please either upgrade "
+                        "CPLEX or remove the space from the warm start path.")
+
```
carldlaird: Are there not standard library methods to do this correct quoting? (A simple googling resulted in me not seeing anything, but I find that pretty strange.) Thoughts?

carldlaird: Do quotes work in Linux, or is this just windows? Linux needs escapes, right?

carldlaird: It seems that the assumption is, if there is any quote in the path, then we are going to trust the user. Is that correct?

carldlaird: Each of these cases appears to be almost the same code pasted 4 times. Can we make a function that returns a "quoted_path_if_necessary" instead of copy/paste code?

jsiirola:
@qtothec - yes: it is probably a good idea to trap any Unicode characters, too. I am not sure if CPLEX supports them (even in 12.8). What is worse is that I don’t have any way to test it.

pyomo/solvers/plugins/solvers/CPLEX.py
jsiirola: I don’t know of one

jsiirola: The original fix was developed by @ghackebeil, I assume in Linux. I don’t have access to 12.8, so cannot test the quotes.

jsiirola: Yes. Given @qtothec’s comment, I am going to rewrite the test to disallow anything other than `a-zA-Z0-9.-_`, plus spaces in 12.8+. I think I will only allow quotes as the first and last character.

qtothec:
Also, is there an easier way of telling people to temporarily change the `tmpdir` without having them play with `pyutilib`?

jsiirola:
@carldlaird: I believe everything in your review has been addressed.  Please re-review. The helper function is fully-tested.  The remainder is not tested due to cplex not being available on Travis/Appveyor

jsiirola:
Merging - the only check failure is the diff coverage, which is due to unavailable licensed solvers (cplex).

PR #587, opened 2018-06-22
Adds a missing import to core.base.symbolic

qtothec:
## Summary/Motivation:
Adds a missing import. Without it, this line doesn't work: https://github.com/qtothec/pyomo/blob/f19bb51225d8a0597ac93d177a33a581d48ca063/pyomo/core/base/symbolic.py#L200

### Legal Acknowledgement


qtothec:
I would if I knew how to construct something that would trigger it instead of running into it with a large model.

PR #590, opened 2018-06-23
Fix to solver tempfiles

ghackebeil:
Recently, I noticed that a massive amount of temporary .row and .col solver files were left around in the working directory after running solver tests. The cause of this was twofold:

 1. When an NL-file based solver was used, we were actually never adding the .row and .col files to the TempfileManager.
 2. The recently added `pyomo.contrib.trustregion.PyomoInterface.PyomoInterface` class definition (which gets imported with `pyomo.environ`) includes the line `TempfileManager.tempdir = "."` at the class scope. As a result, the default tempdir for anyone who imports `pyomo.environ` is now `.`.

Issue (1) has existed since the beginning of time, and issue (2) made it easily observable.

This PR fixes (1) and also include a small improvement to GUROBI.py that avoids creating a file to extract the solver version, which was also not being cleaned up in most cases.

I think someone should commit a fix for (2) as part of this PR.

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/trustregion/PyomoInterface.py
+++ b/pyomo/contrib/trustregion/PyomoInterface.py
@@ -1,7 +1,6 @@
 import os
 import numpy as np
 from pyutilib.math import infinity
-from pyutilib.services import TempfileManager
 from pyomo.common.modeling import randint, unique_component_name
 from pyomo.core import Block, Var, Param, Set, VarList, ConstraintList, Constraint, Objective, RangeSet, value, ConcreteModel, Reals, sqrt, minimize, maximize
 from pyomo.core.expr import current as EXPR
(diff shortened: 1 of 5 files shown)


qtothec:
I'm not as familiar with the solver system as @gseastream. Quick question, though: what's the idea with `pyomo.contrib.PyomoInterface.PyomoInterface`? I don't see it in contrib.

ghackebeil:
Sorry, I left out "trustregion".

ghackebeil:
There seems to be a py2-style print statement breaking py3 builds.

jsiirola:
@ghackebeil: DOH.  That said, this is why I make debugging statements be 2.7-only (so that Travis reminds me to remove them...).  We should be good now.

PR #593, opened 2018-06-27
Change GAMS writer ctype checking

### LINKED ISSUE #340: GAMS writer problem: Unallowable component ContinuousSet
1.- When using expressions and exporting the model to GAMS, there are some issues with the final expressions. @jsiirola 
GAMS error: "More than one operator in a row. You need to use parenthesis"
Two examples: 
Y = 4000 + - T + 2000
(2*x)** -0.5
pyomo returns "*** Status: Compilation error(s)" errors from GAMS could be useful to understand the issue.

2.- Importing a model using DAE tools is not working (@blnicho):
RuntimeError: Unallowable component(s) ContinuousSet.
The GAMS writer cannot export models with this component type

Regards,
Miguel

gseastream:
## Summary/Motivation:
We want the writers to be able to write models whenever possible and only throw an error when they get a model that doesn't make sense to them. This implements the checks that the GAMS writer needs in order to know if it can write the provided model: are there strange things in active constraints, or are there strange active components that are active on the model.

- Check the model at the beginning for any active components it isn't expecting
- Check everything the constraint walker sees to see if it is a `Var`, `Param`, `Expression`, or `Objective`
- Check that everything in active constraints are on the same model as the model passed to the writer
- Add `is_component_type` method to NumericValues so we know to expect a `type()` method
- Add standard sets of components that writers in general can use to reference allowable components
- Other minor cleanup and improvements
- Tests

### Legal Acknowledgement


gseastream:
There is only one failing test, and it's a bit of a dilemma. It's [this test](https://github.com/Pyomo/pyomo/blob/67a12983ad2ba9babd67627ea4a3f41c4ff01095/pyomo/repn/tests/gams/test_gams.py#L124), which uses the following model:

```python
def test_var_on_nonblock(self):
    class Foo(Block().__class__):
        def __init__(self, *args, **kwds):
            kwds.setdefault('ctype',Foo)
            super(Foo,self).__init__(*args, **kwds)

    model = ConcreteModel()
    model.x = Var()
    model.other = Foo()
    model.other.a = Var()
    model.c = Constraint(expr=model.other.a + 2*model.x <= 0)
    model.obj = Objective(expr=model.x)
    self._check_baseline(model)
```

The error comes when the writer checks all active components for things it expects, and the ctype `Foo` is not something it expects. Going by our discussion, I don't think we should expect the writer to be ok with an active `Foo` block, since this is basically the same issue as a Disjunct that hasn't been transformed. I see that @whart222 added this test, but I don't think this should pass as is. Thus, how should we proceed? Two options I see are we either remove the test or deactivate the Foo.

qtothec:
Are there bad side-effects to deactivating the Foo?

gseastream:
In this example, the only thing on it is the variable and the constraint should still be able to reference variables on deactivated blocks. To me it feels like if this were some component in need of a transformation, Foo would either be reclassified or deactivated, and since the name of the test implies being able to find variables on non blocks, we should deactivate it instead of reclassifying it. I'm just wondering if that's what we want out of this test.

gseastream:
I found out there were some deactivated tests lower in that file that used to assert that some errors were raised. Two of those tests, `test_var_on_other_model` and `test_var_on_deactivated_block`, were made to expect `KeyErrors`, but they were also located above in the activated tests and did not expect those errors. Similarly, the `test_var_on_nonblock` test was also in those deactivated tests and expected a `RuntimeError`. Since that is back to being the normal behavior, I elected to remove the Foo test from above and the failing KeyError tests from below and reactivate the RuntimeError Foo test. Let me know if this was the wrong move.

Another question comes from this though: are writers supposed to error when they find vars on deactivated blocks, non-blocks, or other models? Or does this fall under the issue that our writers all behave differently?

qtothec:
@gseastream That last question is encompassed within the "active" discussion that we had.

jsiirola:
(Catching up here) @gseastream: Removing the old "`Foo`" ctype test is the right thing: It doesn't matter what "Foo" has on it, by declaring itself to have a different `ctype`, Foo is explicitly stating that it is *not* a normal Block.  The GAMS writer should not assume that it knows how to handle it and should (correctly) raise an exception.

As for the question of Vars on deactivated things / not on the model...  I would propose that we move to the "@carldlaird worldview" and allow it, with the possible exception of Vars not in the same storage tree as the model (that is, `var.model() is instance.model()` should be True, whether or not `instance is instance.model()` is True.

gseastream:
@jsiirola I added an error to be thrown if it finds vars that don't have the same parent model as the model that was passed to write. Unfortunately this had to go pretty late in the function where it is looping through all the variables it already found and building the bounds statements for them. Due to the current api of the expression walker, there is no reference to the model that was passed to write in the visiting functions, so (if I'm not wrong) there is no way to do this check while we are walking the expressions. This might be fine, but there are 2 issues:

1. We could save a lot of time if we were able to throw the error sooner since we could throw it as soon as we came across the bad variable, instead of waiting to compute all the constraints.
2. This still allows you to have an Expression defined on a different model and used within a constraint on the correct model, as long as that outside Expression only contains vars from the correct model. Meaning, the following example will work:

```python
from pyomo.environ import *
from six import StringIO

m = ConcreteModel()
other = ConcreteModel()
m.x = Var()
other.e = Expression(expr=m.x + 2)
m.c = Constraint(expr=other.e == 5)
m.o = Objective(expr=m.x)
os = StringIO()
m.write(os, format="gams", io_options=dict(symbolic_solver_labels=True))
print(os.getvalue())
```

Which contains:
```
EQUATIONS
	c
	o;

VARIABLES
	GAMS_OBJECTIVE
	x;

c.. (x + 2) =e= 5.0 ;
o.. GAMS_OBJECTIVE =e= x ;
```

jsiirola:
@gseastream - this makes sense.  Things like the `Expression` issue you point out highlight why I think I am in favor of not worrying about *where* non-activatable components are stored (including Expressions / Vars / Params).  However, in the neat term, I think that keeping the check in is probably a good thing -- at the very least, it is an example of what would be required for the other writers.

As for where we check the Vars, I actually prefer checking after they are collected: most variables appear in multiple expressions, so by deferring the check we will only check each var once.  Also, we can work on making the check more efficient (i.e., passing as soon as we encounter a block that we have seen before and thereby avoiding walking all the way to the top of the hierarchy every time).

We will need to check Expressions (and Params) in the expression walker.  I think the best thing there is to just create a derived walker that stores the model on the walker as part of the `__init__()`.

pyomo/repn/plugins/gams_writer.py
```diff
@@ -104,15 +119,122 @@ def visiting_potential_leaf(self, node):
...
+                raise RuntimeError(
+                    "Found an 80,000+ character string with no spaces")
+            i -= 1
+        new_lines += line[:i] + '\n'
+        line = line[i + 1:]
+    new_lines += line
+    return new_lines
+
+
+def check_on_model(model, comp):
```
jsiirola: Why not make this a class?  Then it can store a ComponentSet of known blocks... it can initialize the ComponentSet with the model and all the model's parents.  Then `check_on_model` only has to check the object's parent_block().  If it is not in the set, walk up the hierarchy an collect "new" blocks.  If you encounter one that is in the ComponentSet, then add everything you've run across to the Set and return True; else if you get to the root without finding anything in the Set, throw the list of collected blocks away and return False.

pyomo/repn/plugins/gams_writer.py
```diff
@@ -327,29 +448,20 @@ def _write_model(self,
...
-        # HACK: Temporary check for Connectors in active constriants.
-        # This should be removed after the writer is moved to an
-        # explicit GAMS-specific expression walker for generating the
-        # constraint strings.
-        has_Connectors = Connector in model_ctypes
+        valid_ctypes = set([Block, Constraint, Objective, Suffix])
+        invalids = set()
+        for t in (model_ctypes - valid_ctypes):
+            if issubclass(t, ActiveComponent):
+                invalids.add(t)
```
jsiirola: It is not invalid to have a Component derived from an ActiveComponent on the model, as long as it is deactivated.  I think you need to go through your list of ctypes here and collect the list of active components.  If that list is empty, all is good -- otherwise throw the error.

gseastream: I think this is already doing what you want. I use `model_ctypes = model.collect_ctypes(active=True)` to collect all the active ctypes on the model, and I subtract `valid_ctypes` which is the set of active ctypes I am ok with. Then I go through what remains and if anything in there is an `ActiveComponent`, that means there is an active component on the model that is not deactivated and should be, so I throw an error.

pyomo/repn/plugins/gams_writer.py
gseastream: Good idea. Blocks are hashable though, so is there any reason not to just use a normal set?

pyomo/repn/tests/gams/test_gams.py
```diff
@@ -195,26 +172,10 @@ def test_var_on_other_model(self):
...
-        model.other.deactivate()
-        model.c = Constraint(expr=model.other.a + 2*model.x <= 0)
-        model.obj = Objective(expr=model.x)
-
-        baseline_fname, test_fname = self._get_fnames()
-        self._cleanup(test_fname)
-        self.assertRaises(
-            KeyError,
-            model.write, test_fname, format='gams' )
-        self._cleanup(test_fname)
```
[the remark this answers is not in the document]
gseastream: The test is still active, it is higher in the file and we test that it works.

pyomo/repn/plugins/gams_writer.py
```diff
@@ -92,6 +92,21 @@ def visiting_potential_leaf(self, node):
         if node.__class__ in native_types:
             return True, str(node)
 
+        # Make sure all components in active constraints are basic ctypes we
+        # know how to deal with. This means anything with a type() method must
+        # be one of the following allowable types.
+        if hasattr(node, "type"):
```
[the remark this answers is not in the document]
gseastream: Where you say "At this point, this had better be a Pyomo type", we can still get `NumericConstant` types here, and those do not have a `type()` method. Is this contrary to what you expected? And what do you then think should be done about that?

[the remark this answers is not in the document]
jsiirola: Hmm - that is contrary to what I expected.  I think I would be in favor of adding `type()` to `NumericConstant`.  Thoughts, @whart222?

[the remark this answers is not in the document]
whart222: I think that adding type() would be confusing, since this is a component type ... and NumericConstants are not components.  I think it's sensible to check for is_expression_type() as John suggests, and avoid calling this function.  (More generally, I've seen several posts about Python that suggest that developers avoid the use of hasattr(), since it's behavior is not as reliable as you might guess.)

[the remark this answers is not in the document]
gseastream: So currently, the following code will not work:

```python
def visiting_potential_leaf(self, node):
    """
    Visiting a potential leaf.

    Return True if the node is not expanded.
    """
    if node is None:
        return True, None

    if node.__class__ in native_types:
        return True, str(node)

    if node.is_expression_type():
        # we will descend into this, so type checking will happen later
        return False, None

    # At this point, this should be a Pyomo type because native types and 
    # expressios have already been filtered, so make sure all components in
    # active constraints are basic ctypes we know how to deal with.
    if self.ctype(node) not in (Var, Param, Expression, Objective):
        raise RuntimeError(
            "Unallowable component '%s' of type %s found in an active "
            "constraint or objective.\nThe GAMS writer cannot export "
            "expressions with this component type."
            % (node.name, self.ctype(node).__name__))
    if self.ctype(node) in (Param, Expression, Objective):
        # For these, make sure it's on the right model. We can check
        # Vars later since they don't disappear from the expressions
        self.treechecker.on_tree(node)

    if node.is_variable_type():
        if node.fixed:
            return True, str(value(node))
        label = self.smap.getSymbol(node)
        return True, label

    return True, str(value(node))

def ctype(self, comp):
    if isinstance(comp, IComponent):
        return comp.ctype
    else:
        return comp.type()
```

Are you saying we should just change the `ctype` helper function to do this, and also allow `NumericConstant` in the list above?

```python
def ctype(self, comp):
    if isinstance(comp, IComponent):
        return comp.ctype
    elif isinstance(comp, Component):
        return comp.type()
    else:
        return type(comp)
```

gseastream:
@jsiirola @whart222 This should be ready for review (again). I've updated the PR description to include more of the changes I have added, the more recent of which are the new `is_component_type` method on NumericValues and having a standard `valid_ctypes_minlp` set in `problem_utils.py`, both of which came from my conversation with Bill about proper standard practice the writers should exhibit.

pyomo/repn/plugins/gams_writer.py
```diff
@@ -92,27 +94,155 @@ def visiting_potential_leaf(self, node):
...
-        if not node.is_expression_type():
-            return True, str(value(node))
+        return True, str(value(node))
 
-        return False, None
+    def ctype(self, comp):
+        if isinstance(comp, IComponent):
+            return comp.ctype
+        else:
+            return comp.type()
```
jsiirola: What is the point of the `ctype()` helper function?  What case are you catching here?

pyomo/repn/plugins/problem_utils.py
```diff
@@ -8,284 +8,8 @@
...
-                lt = len(t)
-                if lt > 0:
-                    Cnz.append(cndx)
-                    nt += 1
-            if nt > 0:
-                Con1.append(key)
-            C._Cnz = Cnz
-        model.Cnontriv = Con1
+valid_expr_ctypes_minlp = {Var, Param, Expression, Objective}
+valid_active_ctypes_minlp = {Block, Constraint, Objective, Suffix}
```
jsiirola: This file doesn't really belong here.  The "plugins" directory should hold plugin implementations.  Common utilities like this should be elsewhere.

pyomo/repn/plugins/gams_writer.py
gseastream: For kernel components, it's a property called `ctype`. For regular components, it's a method called `type()`.

jsiirola: OK.  @ghackebeil: why is there this inconsistency?  Should we move `core.base` to also define a `ctype` property?  Move both to have a `ctype()` method?  I for one never liked the move from a `ctype` attribute to a `type()` method because it doesn't return the `type`, but rather the effective component type.  This is highlighted by the developers referring to it as `ctype` in conversation.

Regardless, I think we are good to merge and should move this discussion to an Issue / subsequent PR.

PR #598, opened 2018-06-30
[WIP] Misc. Cleanup in Kernel

ghackebeil:
TODO...

### Legal Acknowledgement


whart222:
@ghackebeil I've reviewed this briefly.  When this is ready for a review, I recommend that we chat.  I think it'll be more effective for me to simply talk through the changes you're making and why.

ghackebeil:
In effort to reduce the number of base class methods, I removed `parent_block` from kernel objects. It's both unclear what `.parent_block` is supposed to be and trivial enough to implement using calls to `.parent`.

If you'd like, I could re-implement the tree checker class to not use `parent_block` at all, and look the same for both AML and Kernel.

gseastream:
I was just curious, I don't think that'll be too necessary. Although I know we discussed the fact that we should minimize api differences between aml and kernel, so if that's relevant then we might want to consider converging the logic here.

ghackebeil:
@wehart: I believe this PR is ready for a review. I'd be happy to discuss these changes whenever you are free.

gseastream:
Everything on master is failing after this merge ([test](https://travis-ci.org/Pyomo/pyomo/builds/407739115)) because the import name for `ComponentMap` and `ComponentSet` has changed. Looks like we should have updated this branch before merging it to master.

PR #606, opened 2018-07-12
Creating pyomo.dualize

whart222:
## Summary/Motivation:
This change consolidates elements of Pyomo that
are used to create dual problems.  Currently, this is only used
by pyomo.bilevel.  But this transformation has broader utility.

- Creating pyomo.dualize
- Moving the linear_dual transform out of pyomo.core.plugins into pyomo.dualize.plugins
- Moving the collection routines out of pyomo.repn into pyomo.dualize
- Moving associated tests out of pyomo.core

### Legal Acknowledgement


pyomo/duality/tests/test_linear_dual.py
```diff
@@ -15,7 +15,7 @@
 from os.path import abspath, dirname, normpath, join
 
 currdir = dirname(abspath(__file__))
-exdir = normpath(join(currdir,'..', '..','..','..','examples','pyomo','core'))
+exdir = normpath(join(currdir,'..','..','..','examples','pyomo','core'))
```
jsiirola: For the record: this is a bad test structure - the test will only work on source checkouts and not if someone installs the code.

blnicho: @jsiirola could you point to a better way to do this? I know I've copied this approach for some of my Simulator tests not knowing how else to reach scripts in the Pyomo examples directory. This might be a good topic for the developer meeting.

jsiirola: The fundamental problem is that we do not distribute the examples directory with anything other than source/git installs.  Personally, I think we should move to a model where the examples are alongside the source in `pyomo/<subpackage>/examples` -- that is, the `examples` are a peer directory to the `tests` directory.  We could then add a utility (to the pyomo script) to extract the examples to wherever the user wants them.  Regardless, this should be discussed on an issue and not hold up this PR.

whart222: Can we run tests on an installed Pyomo package?  What would that look like?

PR #607, opened 2018-07-12
Add pyomo.dataportal subpackage

whart222:
## Summary/Motivation:
Move DataPortal logic out of pyomo.core to better modularize the Pyomo code base.

- Moved code from of pyomo.core.data into pyomo.dataportal
- Moved plugins in pyomo.core.plugins into pyomo.dataporta.plugins
- Moved some tests in pyomo.core.tests into pyomo.dataportal.tests

### Legal Acknowledgement


whart222:
NOTE:  Some of the logic in pyomo.dataportal depends on knowing whether data is being loaded into a Set or Param.  However, this code is not strictly dependent on that functionality.  It just helps in some cases.  Hence, I think this move is reasonable.

Imports from pyomo.core.base were moved into the specific functions/methods that needed to import a Set or Param object.  These should be called very infrequently in Pyomo scripts, so there should be minimal impact on that move.

pyomo/dataportal/parse_table_datacmds.py
```diff
@@ -0,0 +1,133 @@
+
```
jsiirola: This file is autogenerated by PLY and I believe should not be committed to the repo.  I believe there are issues if the table is out of date (wrt the version of PLY that the user has installed) and the user cannot edit the file (e.g., Pyomo was installed into the system python).

jsiirola:
@whart222: I think that we should merge (pending tests) and then add an issue to replace the reference to `DataPortal` from `PyomoModel.py` and replace it with a function registry.  This will break the circular dependence between `pyomo.core.base.PyomoModel` and `pyomo.dataportal` (so that dataportal can import Set/Param at the global scope).

PR #610, opened 2018-07-16
More efficient string manipulation in LP writer

whart222:
## Summary/Motivation:
Follow best practices for string manipulation in Python

- Collect strings and join them together instead of writing each string separately
- Avoid using '+' to concatenate strings
- Cache strings used repeatedly

### Legal Acknowledgement

### DIFF
--- a/pyomo/repn/plugins/cpxlp.py
+++ b/pyomo/repn/plugins/cpxlp.py
@@ -56,15 +56,13 @@ class ProblemWriter_cpxlp(AbstractProblemWriter):
 
         AbstractProblemWriter.__init__(self, ProblemFormat.cpxlp)
 
-        # the LP writer is responsible for tracking which variables are
-        # referenced in constraints, so that one doesn't end up with a
-        # zillion "unreferenced variables" warning messages. stored at
-        # the object level to avoid additional method arguments.
-        # dictionary of id(_VarData)->_VarData.
+        # The LP writer tracks which variables are
+        # referenced in constraints, so that a user does not end up with a
+        # zillion "unreferenced variables" warning messages.
+        # This dictionary maps id(_VarData) -> _VarData.
         self._referenced_variable_ids = {}
 
-        # Keven Hunter made a nice point about using %.16g in his attachment
-        # to ticket #4319. I am adjusting this to %.17g as this mocks the
+        # Per ticket #4319, we are using %.17g, which mocks the
         # behavior of using %r (i.e., float('%r'%<number>) == <number>) with
         # the added benefit of outputting (+/-). The only case where this
         # fails to mock the behavior of %r is for large (long) integers (L),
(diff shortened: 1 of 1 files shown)


whart222:
On p-median8, this reduced runtime on my Mac by about 1 second, which is a 3% improvement.

whart222:
According to 'top', on p-median8, the master branch requires a maximum of 1.05G, which this PR requires a maximum of 1.19G.  That's just shy of a 15% increase in peak memory.

If that seems excessive, then we could revise the writer to flush the list of strings if it gets too long...

jsiirola:
I see a similar performance improvement in Python 2.7 under Linux (I am assuming that @whart222 and @carldlaird were testing 3.6).  I also see similar memory differentials (although the numbers are worse): on Master it takes 1GB to create the model and a peak of 1.6GB during the write process.  The PR branch uses a peak of 1.8GB.  (This is all according to @whart222's `memmon`)

I see nothing wrong with this approach; however, I am also concerned that we are building up the entire file in memory before emitting it.  I good compromise would be to at least dump things constraint-at-a-time.  Alternatively, I could see an approach that dumped things when the list exceeded N entries.

As a side benefit, capping the size of the list can also have performance improvements, too.

P.S.: Thank you for looking at this.  Right now, it takes twice as long to write the LP file as it does to generate the model...  and our overall performance on pmedian8 is noticeably worse on master... model construction is 7% slower than Pyomo 5.5 (with a roughly equivalent write time).  Beating down the write time helps to offset performance losses in construction.

whart222:
I'll look to see if there's a simple strategy for flushing the line cache on a "regular" basis.

whart222:
@jsiirola Your pmedian performance statistics are inconsistent with our nightly tests on snotra.  Let's look offline to see how you're testing.

jwatsonnm:
While a 15% increase in memory may not seem bad, we're already IMHO very memory-intensive - it's often a bottleneck when running on laptops. And relative to other AMLs. If there is potentially a quick fix that mitigates the increase, it would be worth looking in to.

whart222:
I tried John's suggestion of flushing the cache of lines occasionally.  It worked so well that I couldn't detect any changes in the max memory in this branch.  :)

qtothec:
@gseastream You think the GAMS writer would also benefit from this?

gseastream:
@qtothec perhaps. The GAMS writer does a combination of storing things in memory first and writing everything individually. All the constraints are written to a separate StringIO first, and then when it's time to write those to the file it goes through line by line and checks if it's too long first (in which case it breaks it up so GAMS won't complain) and then writes each line to the file one by one. For every other line that gets put in the model file, they're each immediately written the output file.

This discussion would indicate that this isn't as fast as it could be, especially since the constraints get written one at a time despite being stored in memory together. Are there performance issues with the GAMS writer that would warrant applying these changes to it too?

whart222:
To clarify, my recent changes were both fast and did not increase the memory during the writer.  Hence, we can merge.

whart222:
@gseastream I think it's worth revisiting the string writing for other writers.  I've seen performance comparisons that suggest that writing to StringIO and then to a file is slower than collecting strings, joining them, and then writing.

PR #619, opened 2018-07-20
Lp fixed bounds

whart222:
## Summary/Motivation:
We need to fix the generation of bounds for fixed variables in the LP writer.  @DLWoodruff noted that we were getting erroneous LP files when (a) a variable was fixed and (b) one of the variable bounds was None.  The bounds should be equal to the fixed variable value, but an infinite value was being reported.

- Restructured the processing of fixed variables to print these variable bounds.
-

### Legal Acknowledgement


whart222:
Note that I do not include a test here.  This is intentional.  I've confirmed this fix with the following test problem:
```
from pyomo.environ import *

model = ConcreteModel()

model.x1 = Var(initialize=1.0, bounds=(0.0, 2.0))
model.x2 = Var(initialize=1.0, bounds=(None, 2.0))
model.x3 = Var(initialize=1.0, bounds=(0.0, None))

model.y1 = Var(initialize=1, bounds=(0, 2), within=Integers)
model.y2 = Var(initialize=1, bounds=(None, 2), within=Integers)
model.y3 = Var(initialize=1, bounds=(0, None), within=Integers)

model.z = Var()

model.OBJ = Objective(expr=model.x1 + model.x2 + model.x3 + model.y1 + model.y2 + model.y3 + model.z)
model.CON = Constraint(expr=model.x1 + model.x2 + model.x3 + model.y1 + model.y2 + model.y3 + model.z == 0)

from pyomo.repn.standard_repn import preprocess_block_objectives

preprocess_block_objectives(model)
model._gen_obj_repn = False

model.x1.fixed = True
model.x2.fixed = True
model.x3.fixed = True

model.y1.fixed = True
model.y2.fixed = True
model.y3.fixed = True
```
HOWEVER, I needed to test this problem with the option "output_fixed_variable_bounds" to True.  

PySP appears to be exercising the LP writer in a fairly non-standard manner to exercise the code that caused this bug.  I'm going to submit a ticket to prompt a design discussion w.r.t. the generation of the StandardRepn objects for writers, and to assess whether/why options like "output_fixed_variable_bounds" are necessary.  (More generally, I'm wondering why we shouldn't always omit fixed variables...)

jwatsonnm:
A quick comment on why the option "output_fixed_variable_bounds" is necessary / desired, and why we should **not** always omit fixed variables. It has to do with the cost of preprocessing and model re-construction, more than anything. When implementing modify+re-solve type algorithms (including PH, but many others), we are often selecting a variable or a handful of variables, fixing them to a new value, re-solving, and analyzing. Without the existing option to output fixed variable bounds, we would have to re-generate the entire standard representation based on what was fixed. Pyomo currently doesn't have an efficient back-map from variables to constraints they appear in, and even if we did, re-generating representations would be major bottleneck. It is far more efficient to simply change the variable bounds, and re-run. This is even a bigger issue for persistent interfaces, where both the representations and the model would have to be re-built at every re-solve step.

jwatsonnm:
The fix to cpxlp.py looks good. However, there are numerous other files that are unrelated to this fix / PR, e.g., having to do with parallel branch and bound and something about a BARON test. Can you remove all but the cpxlp.py file from this PR? It might help forensic code anthropologists in the far future, and perhaps even us in the next year or so.

whart222:
Closing this PR.  Somehow it got polluted with other files that I can't figure out how to delete.  :(  I'll re-open this PR with another branch.

PR #634, opened 2018-07-26
Silencing invalid error messages being logged with Template Expressions

blnicho:
## Summary/Motivation:
One unintended side effect of the new expression system is that error messages are now being logged every time Template Expressions are used. This PR defers error handling and error message logging to the Template Expression system.

### Legal Acknowledgement


pyomo/core/expr/numvalue.py
```diff
@@ -217,13 +217,23 @@ def value(obj, exception=True):
         #
         # Here, we try to catch the exception
         #
+
+        # This import needs to be deferred to here due to circular
+        # imports
+        from pyomo.core.expr.current import TemplateExpressionError
```
jsiirola: Does it break the circular reference to move the TemplateExpressionError definition either into this file, or better yet into a separate file?

michaelbynum:
Oops. I did not realize John reviewed this already...

pyomo/core/expr/numvalue.py
blnicho: Yes, moving TemplateExpressionError to pyomo/core/expr/numvalue.py breaks the circular reference. I also tried moving it to pyomo/core/base/template_expr.py and pyomo/core/base/misc.py but ran into the same circular reference problem. My only hesitation with making this change is that numvalue.py seems like a weird place for this error definition.

qtothec: Yeah, I agree that `numvalue.py` seems like a really odd place to put it. Why not put it in an `expr_errors.py` file? Or an `expr_util.py` file?

blnicho: @qtothec yeah, I like that idea better

PR #662, opened 2018-08-20
[WIP] Trustregion changes

dmvigi:
## Summary/Motivation:
This is a first attempt for adding support for fixed variables and for setting options using config blocks

- Changed parameters to use Python ConfigBlock parameters. Example1.py solution matches after implementing new code.
- Allow trust regions external functions to be called with fixed variables. Example2 files show the difference between trust region solver and direct solve in Pyomo. See README in examples file.
TODO: Need to implement python logging to print debug statements correctly.

I am also requesting a code review from @jsiirola since this is my first pull request.

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/trustregion/TRF.py
+++ b/pyomo/contrib/trustregion/TRF.py
@@ -11,14 +11,18 @@ from pyomo.contrib.trustregion.Logger import (IterLog, Logger)
 from pyomo.contrib.trustregion.PyomoInterface import (
     PyomoInterface, ROMType)
 
-def TRF(m,eflist):
+def TRF(m, eflist, config):
     """The main function of the Trust Region Filter algorithm
+
     m is a PyomoModel containing ExternalFunction() objects Model
     requirements: m is a nonlinear program, with exactly one active
     objective function.
 
     eflist is a list of ExternalFunction objects that should be
     treated with the trust region
+
+    config is the persistent set of variables defined 
+    in the ConfigBlock class object
 
     Return: 
     model is solved, variables are at optimal solution or
@@ -32,13 +36,9 @@ def TRF(m,eflist):
     problem = PyomoInterface(m, eflist)
     x, y, z = problem.getInitialValue()
 
-    trustRadius = CONFIG.trust_radius
-    sampleRadius = CONFIG.sample_radius
-    sampleregion_yn = CONFIG.sample_region
-
     iteration = -1
 
-    romParam, yr = problem.buildROM(x, sampleRadius)
+    romParam, yr = problem.buildROM(x, config.sample_radius)
     #y = yr
     rebuildROM = False
     xk, yk, zk = cloneXYZ(x, y, z)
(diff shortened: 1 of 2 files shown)


pyomo/contrib/trustregion/PyomoInterface.py
```diff
@@ -27,13 +30,22 @@ def __init__(self, trf_block, efSet):
         self.efSet = efSet
 
     def visit(self, node, values):
+        
+        sys.stdout.write('visit: \n  node = %s\n' % node)
+        for i in range(len(values)):
+            sys.stdout.write('  values[%d] = %s\n' % (i,values[i]))
```
jsiirola: This looks like debugging?  If it is, it should be removed before merging.  If it is intentional output, it should come out through a `logger`

pyomo/contrib/trustregion/PyomoInterface.py
```diff
@@ -48,12 +60,15 @@ def visit(self, node, values):
         # things later; i.e., accessing the private _fcn attribute
         # below).
         for arg in list(values)[1:]:
-            if type(arg) in nonpyomo_leaf_types or arg.is_fixed():
-                # We currently do not allow constants or parameters for
+            
+            sys.stdout.write('    %s:  node is a PythonCallbackFuction\n' % arg)
+            if type(arg) in nonpyomo_leaf_types:
```
jsiirola: You should treat `nonpyomo_leaf_nodes` as if they were fixed variables.  These are any leaf nodes (values) in an expression tree that are not Pyomo NumericValue types -- to it includes things like literal constants.

pyomo/contrib/trustregion/PyomoInterface.py
```diff
@@ -48,12 +60,15 @@ def visit(self, node, values):
         # things later; i.e., accessing the private _fcn attribute
         # below).
         for arg in list(values)[1:]:
```
jsiirola: See the note above, but if you start treating nonpyomo_leaf_nodes as fixed arguments and ignoring them from the TR models, then we can remove the `[1:]` restriction here, an probably relax the set of allowable ExternalFunction types to "all".

pyomo/contrib/trustregion/PyomoInterface.py
```diff
@@ -94,6 +123,7 @@ class PyomoInterface:
 
     '''
 
+    TempfileManager.tempdir = "."
```
jsiirola: We do NOT want to be mucking with the tempfile directory....  I think I removed this line from master -- and you do not want to re-add it. (it breaks other tests and does evil things with polluting a user's current working directory.

pyomo/contrib/trustregion/PyomoInterface.py
```diff
@@ -471,9 +538,12 @@ def criticalityCheck(self, x, y, z, rom_params, worstcase=False, M=[0.0]):
...
 
-        optGJH = SolverFactory('contrib.gjh')
-        optGJH.solve(model, tee=False, symbolic_solver_labels=True)
-        g, J, varlist, conlist = model._gjh_info
+        optGJH = SolverFactory('gjh', solver_io='nl')
+
+        optGJH.solve(
+            model, keepfiles=True, tee=False, symbolic_solver_labels=True)
+
+        g, J, varlist, conlist = readgjh()
```
jsiirola: This is reverting changes to the trustregion added to master.

pyomo/contrib/trustregion/TRF.py
```diff
@@ -1,45 +1,41 @@
...
+    FilterElement, Filter)
+from pyomo.contrib.trustregion.helper import (cloneXYZ, packXYZ,
+    minIgnoreNone, maxIgnoreNone)
+from pyomo.contrib.trustregion.Logger import (IterLog, Logger)
+from pyomo.contrib.trustregion.PyomoInterface import (
+    PyomoInterface, ROMType)
 
 def TRF(m,eflist):
+    """The main function of the Trust Region Filter algorithm
+    m is a PyomoModel containing ExternalFunction() objects Model
```
jsiirola: There should be a blank line between the 1-line "short description" and the remainder of the docstring

pyomo/contrib/trustregion/filterMethod.py
```diff
@@ -1,5 +1,5 @@
 from pyomo.contrib.trustregion.param import *
-
+from pyomo.util.config import ConfigBlock
```
jsiirola: This import doesn't appear to be needed.

pyomo/contrib/trustregion/TRF.py
```diff
@@ -1,45 +1,41 @@
...
-    """
-
 
     logger = Logger()
     filteR = Filter()
-    problem = PyomoInterface(m,eflist)
+    problem = PyomoInterface(m, eflist)
     x, y, z = problem.getInitialValue()
 
+    trustRadius = CONFIG.trust_radius
```
jsiirola: There is a slight misunderstanding of how to best use the ConfigBlock system.  The main `CONFIG` object should be thought of as a "template" that *defines* the acceptable options, and I usually store this as a static class member on the solver (or transformation) class.  Then, the solver in its `__init__()` creates an instance object by passing the `**kwds` sent to the solver.  This allows a user to specify the options as keyword arguments to the solver, without overriding the defaults written in by the solver developer (think about creating 2 copies of the same solver: by making a *copy* of the Config, both solvers start from the same default template).  For example:
```
class MySolver(plugin.Plugin):
    plugin.implements(IOptSolver)
    plugin.alias('mysolver')

    CONFIG = pyomo.config.trustregion.param.CONFIG
    # Actually, I would usually declare everything here -- to keep the related code 
    # together, but since you already have a param.py...

    def __init__(self, **kwds):
        super(MySolver, self).__init__(**kwds)
        self.config = MySolver.CONFIG(kwds)
```

Issue #355, opened 2018-02-19
Pyomo closes any open Excel windows


jsiirola:
As reported on StackOverflow (https://stackoverflow.com/questions/48864193/pyomo-closes-excel), running `from pyomo.environ import *` closes any open Excel windows.

This appears to be due to `pyomo.core.plugins.data.sheet.py`, and in particular:
```
if win32com_available:
    from pyutilib.excel.spreadsheet_win32com import ExcelSpreadsheet_win32com
    tmp = ExcelSpreadsheet_win32com()
    try:
        tmp._excel_dispatch()
        tmp._excel_quit()
        _excel_available = True
    except:
        pass
```

At a minimum, this shouldn't be happening during import.  Better yet, we should figure out a way to detect if we should / should not call `._excel_quit()`

blnicho:
I copied this import check to fix a fragile test in https://github.com/Pyomo/pyomo/blob/master/pyomo/core/tests/examples/test_tutorials.py as discussed in #669. When this issue is resolved the fix should be propagated to this file.

goroderickgo:
Hi, glad I just stumbled upon this issue, as we are still having this issue with pyomo 5.3+. 

After testing, 5.2 is the last release in which this bug doesn't cause Excel to quit. Is there a better fix than us modifying `pyomo.core.plugins.data.sheet.py`, perhaps an update to the `_excel_quit()` method?

goroderickgo:
@jsiirola and @whart222, my organization is in a slight paradox: we need to (1) update the version of pyomo to >=5.5 to get the solver interface updates but also (2) don't want to roll out a product to users that causes the related Excel files to close every time we try to run a model. Is a fix in the works?

jsiirola:
@goroderickgo: to my knowledge, no one is digging into this at the moment (I would very much like to fix this as I suspect that it burns most Windows users, but as the saying goes, "there are only so many hours in a day").

A PR would be most welcome if you have time to look into it.

blnicho:
@jsiirola I think I found a simple fix for this and have opened a PR in PyUtilib.

blnicho:
@goroderickgo could you install the master branch of PyUtilib and see if this is resolved for you? It will likely be a few days before we can get this fix into a release.

jsiirola:
@goroderickgo: PyUtilib 5.7.1 and Pyomo 5.6.6 are now released and contain the fix for this issue.

goroderickgo:
[...]

PR #669, opened 2018-08-27
Adding import check to dataportal test

blnicho:
This adds an import check to a dataportal test that has been failing intermittently on Appveyor due to `pyodbc` not being installed.

- Skip one of the dataportal tests if pyodbc is unavailable

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/tests/examples/test_tutorials.py
+++ b/pyomo/core/tests/examples/test_tutorials.py
@@ -27,6 +27,17 @@ try:
     _win32com=True
 except:
     _win32com=False #pragma:nocover
+
+if _win32com:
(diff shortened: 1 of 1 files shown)


blnicho:
Looks like this is actually running afoul of some weirdness in this file: https://github.com/Pyomo/pyomo/blob/master/pyomo/dataportal/plugins/sheet.py

The classes defined in this file depend on which Python packages are available and pyodbc is used as the default without checking if it is available first. There is a comment in this file saying that the pyodbc interface "doesn't work right now" so this file desperately needs to be reworked to either fix the interface with pyodbc or remove it as the default. I copied an extra import check to fix the test but this isn't a good permanent solution as discussed in #355. I'm also going to open an issue to fix the dataportal dependency on pyodbc.

PR #671, opened 2018-08-29
Implement a general-purpose expression visitor

jsiirola:
## Summary/Motivation:
This implements a general-purpose event-driven expression walker.  The idea is to implement a depth-first visitor that generates a full event stream similar to other tree visitors (like expat).  This class can be customized for particular purposes by either subclassing it or parameterizing it through the constructor.  It provides a strict superset of the functionality in the current set of visitors, and appears to be slightly more performant (due to a change in the stack management implementation).

- Add a "GenericExpressionVisitor" class

### Legal Acknowledgement


jsiirola:
@blnicho: That is TBD, and really up to @whart222.  It should be pretty simple to migrate them over.  However, I would argue for the conversion if we can point to either simpler implementations or (documented) faster performance.

pyomo/core/tests/unit/test_expr_pyomo5.py
```diff
@@ -6379,6 +6379,355 @@ def test_replace_nested(self):
...
+
+    def test_bad_args(self):
+        with self.assertRaisesRegexp(
+                RuntimeError, "Unrecognized keyword arguments: {'foo': None}"):
+            EXPR.GenericExpressionVisitor(foo=None)
+
+    def test_default(self):
+        walker = EXPR.GenericExpressionVisitor()
+        ans = walker.walk_expression(self.e)
+        ref = [
```
carldlaird: Hmmm... to make this more understandable, how about
```
ref = [
            [
               [],
               []
            ],
            [],
            [
               [],
               [
                  [],
                  []
               ]
            ]
        ]
```
(I am not convinced this is better...)

pyomo/core/tests/unit/test_expr_pyomo5.py
```diff
@@ -6379,6 +6379,355 @@ def test_replace_nested(self):
...
+                return (), node
+            return node.args, []
+        def finalize(result):
+            if type(result) is list:
+                return result
+            else:
+                return [result]
+        walker = EXPR.GenericExpressionVisitor(
+            enterNode=enter, finalizeResult=finalize)
+
```
carldlaird: I think the test is missing?

carldlaird: Or it is tested above and can be removed.

pyomo/core/tests/unit/test_expr_pyomo5.py
```diff
@@ -6379,6 +6379,355 @@ def test_replace_nested(self):
...
+    def test_beforeChild_acceptData_afterChild(self):
+        counts = [0,0,0]
+        def before(node, child):
+            counts[0] += 1
+            if type(child) in nonpyomo_leaf_types \
+               or not child.is_expression_type():
+                return False, child
```
carldlaird: Maybe return False, None since "child" is not really used anywhere.

pyomo/core/tests/unit/test_expr_pyomo5.py
```diff
@@ -6379,6 +6379,355 @@ def test_replace_nested(self):
...
+            counts[2] += 1
+        walker = EXPR.GenericExpressionVisitor(
+            beforeChild=before, acceptData=accept, afterChild=after)
+        ans = walker.walk_expression(self.e)
+        m = self.m
+        ref = []
+        self.assertEqual(ans, ref)
+        self.assertEquals(counts, [9,9,9])
+
+    def test_enterNode(self):
```
carldlaird: This test is a little confusing. If before is saying to skip all children that are leaves, why do we check for leaves in enter?

pyomo/core/tests/unit/test_expr_pyomo5.py
```diff
@@ -6379,6 +6379,355 @@ def test_replace_nested(self):
...
+    def test_finalize(self):
+        ans = []
+        def before(node, child):
+            if type(child) in nonpyomo_leaf_types \
+               or not child.is_expression_type():
+                return False, child
+        def accept(node, data, child_result):
+            if data is not child_result:
+                data.append(child_result)
+        def enter(node):
```
carldlaird: Same is true here as for above test... before should make this first if not ever true.

qtothec:
I'm asking @johnnybatesjr to chime in, because we found a couple of minor things in using it for the MC++ work.

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -224,7 +225,247 @@ class to transform the context into a nonmutable
...
+                args, data = tmp
+        else:
+            args = None
+            data = []
+        if args is None:
+            if type(expr) in nonpyomo_leaf_types \
+                    or not expr.is_expression_type():
+                args = ()
+            else:
+                args = expr.args
```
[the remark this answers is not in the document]
jsiirola: Yes.  By design, the `args` returned by `enterNode` needs to support `__len__` and `__getitem__`.  If `enterNode` returns `args == None`, then the expression's `node.args` needs to support `__len__` and `__getitem__`.

If this is an issue, what API changes would you suggest?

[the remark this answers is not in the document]
qtothec: @johnnybatesjr Do you remember what `args` was getting passed?

whart222:
I like the API.  It's clearly more intuitive.  Some thoughts for changes/revisions:

- I think we should avoid using 'extensionpoint' in the code, as that has a very different semantics w.r.t. PyUtilib plugins
- The code coverage for the expr_pyomo5.py decreases after this change.  Can this be fixed?
- I really think the code needs more documentation.  Right now, this seems out of place in the expr_pyomo5.py file.  Specifically, the high-level motivation provided in this PR is missing in the code.
- I'm pretty sure we should adopt the performance enhancements in the other walker class.  This might have a noticable impact on Pyomo's high-level performance.  But perhaps that's a separate PR?

jsiirola:
Re coverage: I am not sure what is going on, as 100% of the touched lines (the diff) are covered by tests and I did not remove any other tests.  This PR is adding functionality, so I am willing to chalk it up as a weirdness in nose/coverage.  There was one obsoleted class (with ~4 lines of code) that I just removed -- maybe that will fix what you are seeing?

Re documentation: what specifically do you want?  Currently the docstring is 33% of the diff (81 of the 242 lines, with code comments accounting for another 24%).  The PR motivation was taken from the docstring (except for the comment on relative performance).

My intent is to eventually break apart the expr_pyomo5 file apart -- and one obvious thing would be to put the walkers in a single file.  Reworking the other walkers to use the generic base is also a TODO, but is dependent on measurable test cases.

Re performance: migrating Pyomo walkers to use ideas from the generic walker should indeed be a separate PR.  I thought about doing it as part of this PR, but did not have a convenient example where the existing walkers showed up on the performance profiles.  Any change to improve performance should be able to document the performance improvement.  If you have an example where walking is a measurable amount of time, I would be interested in playing with it to document what (if any) performance benefits adopting the logic in the new walker would offer, as well as if there is a measurable difference between re-implementing specialized walkers versus using the new general walker.

PR #674, opened 2018-08-31
Pyomoext

whart222:
### Overview

This PR is a proof-of-concept for creating a dynamic pyomo.ext module.  The idea is that when a user/developer executes
```
import pyomo.ext.foo
```
then the symbols from the package "pyomoext_foo" are imported below "ext".  If pyomo.ext had an __init__.py file, then this would be equivalent to the following line being added to the that file:
```
import pyomoext_foo as foo
```

This is similar to pyomo.contrib, in that users/developers can pull-in external packages that look like they are part of pyomo.  The key difference is that the pyomo.ext module does this dynamically.  Thus, no previous registration is required.  All packages that begin with "pyomoext_" can be imported dynamically.

Additionally, the pyomo/__init__.py defines a dictionary that is used to initialize pyomo.ext, which can specify additional extension packages that do not follow this naming convention.  This allows Pyomo to optionally standardize the use of external packages through pyomo.ext.

### Motivation

While Pyomo has been developed as a monolithic package, increasingly developers have made the distinction between (1) the "core" packages that define key elements of Pyomo's modeling environment and (2) "extension" packages that build on these "core" capabilities.  For example, Pyomo's expression system and file writers are core elements, which the GDP/PYSP/DAE modeling packages extend to represent different classes of optimization applications.

Further, developers are increasingly developing extensions of Pyomo that require a separate development repository.  For example, Pyomo extensions are being developed in private repositories to support and protect unpublished research activities.  Similarly, Pyomo extensions are being developed in separate repositories for low-TRL research activities;  such projects are often ill-suited for direct inclusion in Pyomo since they involve immature, unstable code.  (FUrther, I'd argue that Pyomo's peer-review development process is ill-suited for such projects.)

While Pyomo development can proceed without coordination amongst such activities, this PR supports the definition of a common namespace that allows Pyomo and related packages to develop capabilities in a synergistic manner.  For example, if John is developing package foo, and he has a cool new function bar(), then my package can reliably import:
```
from pyomo.ext.foo import bar
```
Sure, I don't need to do that, since
```
from foo import bar
```
also works.  But the pyomo.ext module defines a namespace that minimizes the logical overhead for users/developers.  We can all act like we're working within a namespace, even if packages have different names.

- This PR creates the pyomo.ext object, which is a module that dynamically translates the name of the package being imported.
- The module object is created using the _ImportRedirect class, which defined in pyomo.common.ext.py

### Legal Acknowledgement


whart222:
Closing and re-opening to avoid duplicate commits.

PR #678, opened 2018-09-03
Allow user, password and database to be passed as options for pymysql

jotaigna:
## Summary/Motivation:
`user` and `password` options were not being passed to `db_table` objects and are thus unavailable to  create a connection with pymysql. Also, added the option field database to specify database in MySQL server (as opposed to overloading any other keyword)

- Added the keyword `database` to the list of options processed by the `_process_load` module
- included `user`. `password` and `database` to the options loaded into the data structure during intialization of that data
- append database to the list of arguments to launch a connection in the `connect()` method of the `db_Table` class

### Legal Acknowledgement

### DIFF
--- a/pyomo/dataportal/plugins/db_table.py
+++ b/pyomo/dataportal/plugins/db_table.py
@@ -167,6 +167,8 @@ or that there is a bug in the ODBC connector.
                 args.append(options.user)
             if not options.password is None:
                 args.append(options.password)
+            if not options.database is None:
+                args.append(options.database)
             return mod.connect(*args, **kwds)
         except ImportError:
             return None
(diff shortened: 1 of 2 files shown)


jotaigna:
Good point. I can write a mock object that simulates a database connection using the `mock` object from unittest. The mock object can be programmed to respond with data when presented with a correct `load` statement and errors when either the credentials or the url are incorrect. Would that work?

blnicho:
Given the scope and limited changes in this PR, I think we can go ahead and merge this without waiting for a test.

jotaigna:
Thank you much appreciated.
For what is worth this feature is being used with an AWS RDS database that is currently live.

PR #698, opened 2018-09-26
Fix double-operator due to negative constant in GAMS Writer

### LINKED ISSUE #663: Issue exporting model to gams with power operator
Hi,
I am exporting an expression with a negative power operator to gams:
pyomo exports: m.Eq1=Constraint(expr=m.B+ m.A * m.T1**-1.0==0)
gams receives:   Eq1..        B+ A* T1**- 1.0 =e= 0.0 

And I am getting the following error in gams:

445  More than one operator in a row. You need to use parenthesis
       if you want to use such statements. For example
           x := x + - not y;  ->  x := x + (- (not y));
           sum(i$not x, ..)   ->  sum(i$(not x),..)
Best, 
Paz

### LINKED ISSUE #697: GAMS writer not wrapping negative products in parentheses
GAMS writer is producing `56*-7` rather than `56*(-7)`.

I am coming across this in the following expression:
```
c4770.. x3100 - 8.314*((x4989 - 298.15)*19.85894 + (power(x4989, 2) - 88893.42249999999)*-13.066840000000001*0.01*0.5 + (power(x4989, 3) - 26503573.918374993)*3.36356*0.0001*0.3333333333333333)*0.001 =e= 0.0 ;
```

qtothec:
## Summary/Motivation:
Negative constants were causing a GAMS syntax error, because `2*-2` is not valid GAMS input. See example in #697.

### Legal Acknowledgement


gseastream:
Looks like there are some failing tests that still need to be updated. However, the fix you wrote also changes much more of the expression writing than it needs to, for example it turned `x2 - x3` into `x2 + (-1)*x3` in one of the baseline tests. Do you think you should alter it so it only wraps in parens when it needs to (or at least only a little more than it needs to)?

Also could you write a test to capture the fixed issue?

blnicho:
We discussed this at the Pyomo developers meeting today and decided that because this is a fix for a writer and the expressions are not meant to be human-readable, this fix is fine.

PR #705, opened 2018-10-04
Fixing NumPy MKL issues in Appveyor

jsiirola:
## Summary/Motivation:
This is once again trying to fix NumPy / Intel MKL issues that result in
```
Intel MKL FATAL ERROR: Cannot load mkl_intel_thread.dll.
```

- Switch back to using Appveyor's Miniconda
- Switch to using 32-bit Python
- Disabling the DataPortals Excel ODBC interface.  This class does not appear to work and was causing failures when `pyodbc` is present but not `xlrd` (which overrides the ODBC interface).
- add testing for Python 3.7
- switch `codecov` installation from `conda` back to `pip`

### Legal Acknowledgement

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -62,6 +62,12 @@ install:
   - "SET ANACONDA=%CONDA_INSTALL% -c anaconda"
   - "SET CONDAFORGE=%CONDA_INSTALL% -c conda-forge --no-update-dependencies"
   #
+  # Determine if we will use Appveyor's Miniconda or install Anaconda
+  # (intermittently one or the other suffers from NumPy failing to load the
+  # MKL DLL; See #542, #577
+  #
+  - SET USING_MINICONDA=1
+  #
   # Update conda, then force it to NOT update itself again
   #
   # Somehow, the update from anaconda stalls for Python 3.4. So we're not specifying the channel here.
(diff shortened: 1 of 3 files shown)


jsiirola:
(Additional notes... https://github.com/ogrisel/python-appveyor-demo may offer a solution, but would require abandoning conda)

blnicho:
Notes from @carldlaird from the last time he looked at this issue:

OK, first, some facts:

1)	I can get miniconda tests running by “pinning” the version of mkl that is used
2)	The first install takes longer on AppVeyor than subsequent jobs – I suspect things are cached in the VM
3)	The full Anaconda tests take approximately 25 minutes, while the miniconda-based tests take 17 minutes (diff: 8 minutes per job)


Tradeoffs:
1)	By using miniconda and pinning the version of mkl, we get faster tests (although not as much faster as we may have thought – see above)
2)	By using miniconda and pinning the version of mkl, we do NOT get the environment that most users will have installed. For windows users: If users are using miniconda, they will likely see this same mkl error. If they are using full anaconda, they should not see any error.

jsiirola:
I am willing to call this PR ready to review (the most recent Appveyor failure in Python 3.6 is due to networking issues with anaconda.org and not out code).

@whart222: Note, this PR will further disable the use of the pyodbc Excel interface in DataPortal.  I went ahead and did that as there are comments in the code indicating that the pyodbc interface does not work yet (and I could not see how to make it work: there is a requirement that either a `query` or `table` option had to be set, but I could not identify any code that ever set those options).  From what I could glean, the Excel interface works through `xlrd`, but the `pyodbc` fallback interface does not work on any platform (I tested on both Linux and Windows) - so disabling the pyodbc interface does not change any functionality from the user's perspective.

DLWoodruff:
I am probably not qualified to review this PR, but I am qualified to say that I need it.

PR #710, opened 2018-10-15
Bugfix: overriding set values (setitem) for IndexedSets in Python2.7

jsiirola:
## Summary/Motivation:
@DLWoodruff found a (longstanding) problem with IndexedSet.

This fixes a problem where overriding values for IndexedSet sets doesn't
work in Python 2.7 (for unordered sets, the old values were not removed,
and for ordered sets an exception is raised).

- 6-charcter fix for IndexedSet
- Tests that verify this functionality works (and fail without this fix)

### Legal Acknowledgement


DLWoodruff:
I have independently tested this an it fixes the problem.

PR #707, opened 2018-10-09
[WIP] Using MC++ for GLOA implementation

qtothec:
## Summary/Motivation:
Preliminary implementation of Global Logic-based Outer Approximation using MC++ interface.

Making a clean PR that does not include the MindtPy changes, but still incorporates the work on both MC++ and GLOA by @johnnybatesjr.

Notes from #647:
- [ ] make installation guide to document needed steps(see below)
  - run getMCpp.py, run cmake, etc...
- [ ] check licensing for required third party libraries
- [ ] create MakeFile (or even better, CMakeFile)
- [ ] organize into three directories: C/C++ code, Python code, python/C++ interfacing files (`*.so` and associated python file)
- [ ] `*.so` file needs to include all of the dependent libraries (LAPACK, BLAS, MC++). Look into flag to statically link dependent libraries.
- [ ] create a "third_party" directory and `getMCpp.py` file
- [ ] create unittest cases (use reference binaries for testing)

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #647, opened 2018-08-07
MC++ Interface Implementation

johnnybatesjr:
## Summary/Motivation:
This provides an implementation of the McCormick relaxation technique along with capabilities to query subgradients of the convex and concave relaxations at specified variable values.

MC++ library: https://omega-icl.github.io/mcpp/index.html

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #717, opened 2018-10-17
GLOA implementation and GDPopt updates

qtothec:
## Summary/Motivation:
Preliminary implementation of Global Logic-based Outer Approximation using MC++ interface.

Making a clean PR that does not include the MindtPy changes, but still incorporates the work on both MC++ and GLOA by @johnnybatesjr. Also does not include the compiled shared object file because of licensing concerns. This PR replaces #707.

Notes from #647:
- [ ] make installation guide to document needed steps(see below)
  - run getMCpp.py, run cmake, etc...
- [ ] check licensing for required third party libraries
- [ ] create MakeFile (or even better, CMakeFile)
- [ ] organize into three directories: C/C++ code, Python code, python/C++ interfacing files (`*.so` and associated python file)
- [ ] `*.so` file needs to include all of the dependent libraries (LAPACK, BLAS, MC++). Look into flag to statically link dependent libraries.
- [ ] create a "third_party" directory and `getMCpp.py` file
- [ ] create unittest cases (use reference binaries for testing)

### Legal Acknowledgement


pyomo/contrib/mcpp/pyomo_mcpp.py
```diff
@@ -0,0 +1,361 @@
...
+from pyomo.core.kernel.component_map import ComponentMap
+
+logger = logging.getLogger('pyomo.contrib.mcpp')
+
+path = os.path.dirname(__file__)
+
+
+def mcpp_available():
+    """True if the MC++ shared object file exists. False otherwise."""
+    return os.path.isfile(path + '/mcppInterface.so')
```
jsiirola: This is fine for development, but not for end-users, who may not have write access to this directory in order to insert the mcppInterface.so file.  Eventually we will want to make this work for Windows / OSX platforms, too.  I am starting to collect some useful utilities for managing compiled extensions in the [common-downloader](https://github.com/jsiirola/pyomo/tree/common-downloader) branch.  If you have use-cases, please pass them on.

jsiirola:
@michaelbynum: I am OK with the `.gitignore`, at least for the time being: it prevents developers from accidentally committing the built MC++ interface.  The compiled binary is not licensed under BSD and should not be added to the Pyomo repo.

PR #723, opened 2018-10-19
Add `SubclassOf` class to support finding components based on issubclass

jsiirola:
## Summary/Motivation:
This allows users to iterate over model hierarchies based on `issubclass(obj.type(), ctype)` instead of just `obj.type() is ctype`.  This way, for example, users can get Var objects on any "Block-like" object in the model with:
```
model.component_data_objects(Var, descend_into=SubclassOf(Block))
```

- Add `pyomo.core.base.block.SubclassOf` class
- Add tests

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/block.py
+++ b/pyomo/core/base/block.py
@@ -131,9 +131,10 @@ class SubclassOf(object):
 
         model.component_data_objects(Var, descend_into=SubclassOf(Block))
     """
-    def __init__(self, ctype):
+    def __init__(self, *ctype):
         self.ctype = ctype
-        self.__name__ = 'SubclassOf(%s)' % (ctype.__name__,)
+        self.__name__ = 'SubclassOf(%s)' % (
+            ','.join(x.__name__ for x in ctype),)
 
     def __contains__(self, item):
         return issubclass(item, self.ctype)
--- a/pyomo/core/base/block.py
+++ b/pyomo/core/base/block.py
@@ -118,7 +118,34 @@ class _component_decorator(object):
             self._component, self._block, *args, **kwds)
 
 
+class SubclassOf(object):
+    """This mocks up a tuple-like interface based on subclass relationship.
+
+    Instances of this class present a somewhat tuple-like interface for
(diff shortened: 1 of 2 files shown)


ghackebeil:
The approach I've adopted in Kernel is to allow passing a function that indicates whether or not an object should be descended into. That would seem to encapsulate the functionality you are providing here (and then some), but without a new class. So instead of

```python
model.component_data_objects(Var, descend_into=SubclassOf(Block))
```

it could simply be

```python
model.component_data_objects(Var, descend_into=lambda x: issubclass(x, Block))
```

This seems like a more lightweight and flexible solution. Do you agree?

ghackebeil:
Maybe I am misunderstanding though. If `descend_into` can already act as a filter function in the AML Block, then maybe this PR is more about the addition of a helper class to shorten common filters.

jsiirola:
Both `ctype=` and `descend_into=` can take either a single ctype or a tuple of ctypes that specify which ctypes to return or recurse into, respectively.  It was never originally designed to be a filter, as the ctypes are used at the very beginning to set up the iterators within each block (it is not checked for each component or component_data, but instead relies on the grouping that Block already does).

This PR adds the ability to specify *base class* ctypes instead of explicit ctypes.  My original idea was to add something like an "isinstance" argument, but that was messy.  The `SubclassOf` was a cleaner approach that both made for a simpler/more intuitive API and required touching almost nothing in the current component iteration implementation.

You could implement it as a lambda, but that would require changing more of the Component iteration implementation (the implementation is based on `__contains__` and not `__call__` -- because the original use was to specify a type, and `__call__` would return an object instead of a bool).  The current approach also relies on a few other things (like `__name__` to give a more informative error message when necessary) that would be harder to implement with a bare lambda.  Of course, you could also write a trivial wrapper around a lambda (very similar to `SubclassOf`) that would work in the current infrastructure -- and the Block API could even automatically add that wrapper to a bare lambda if necessary.  Adding that could be a separate PR.

ghackebeil:
Agreed.

PR #760, opened 2018-11-27
Minor Fixes to Avoid Deprecation Warnings

ghackebeil:
## Fixes # .

- In Python 3+, use `inspect.getfullargspec` in place of `inspect.getargspec`.
- In Python 3+, import abstract base classes from `collections.abc` instead of `collections` 

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/reference.py
+++ b/pyomo/core/base/reference.py
@@ -8,7 +8,14 @@
 #  This software is distributed under the 3-clause BSD License.
 #  ___________________________________________________________________________
 
-import collections
+try:
+    # python3
+    from collections.abc import MutableMapping as collections_MutableMapping
+    from collections.abc import Set as collections_Set
+except:                                           #pragma:nocover
+    from collections import MutableMapping as collections_MutableMapping
+    from collections import Set as collections_Set
+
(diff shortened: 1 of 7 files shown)


pyomo/core/base/reference.py
```diff
@@ -8,7 +8,14 @@
 #  This software is distributed under the 3-clause BSD License.
 #  ___________________________________________________________________________
 
-import collections
+try:
+    # python3
+    from collections.abc import MutableMapping as collections_MutableMapping
+    from collections.abc import Set as collections_Set
+except:                                           #pragma:nocover
```
jsiirola: We shouldn't need this `#pragma:nocover`, as the 2.7 tests should pick up this branch.  More generally, why are we using a try-except instead of `if six.PY3`?

ghackebeil: Initially, I wasn't sure if it was a Python 3 change or a late Python 3 change. I suspect using the six check would be okay.

jsiirola: I just checked, and `MutableMapping` is in `collections.abc` at least as far back as Python 3.4.

PR #761, opened 2018-11-27
Kernel Updates

ghackebeil:
- Changing default behavior of general iterators in kernel to use `active=True`.
- Move the `preorder_traversal` method from base class hierarchy into a helper utility function.
- Generalizing the `descend_into` keyword for the general iterators to allow users to pass a function. This is similar to the `descend` keyword for `preorder_traversal`, except that it is only called for heterogeneous containers (e.g., blocks).
- Adding reserved attributes check to block (prevents users from overwriting important block methods)
- Adding `functional_value` component, which behaves like a `parameter` but instead of storing a value it stores a function that is called on demand (may require using dill for serialization).
- Use bisection to speed up piecewise function evaluation
- Miscellaneous fixes to documentation
- Fixing import in admin performance script

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/_task.py
+++ b/pyomo/common/_task.py
@@ -16,6 +16,7 @@ __all__ = ['pyomo_api', 'IPyomoTask', 'PyomoAPIFactory', 'PyomoAPIData']
 
 import inspect
 import logging
+import six
 from six import iteritems, with_metaclass
 
 import pyutilib.workflow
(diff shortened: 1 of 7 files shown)


pyomo/core/kernel/block.py
```diff
@@ -76,6 +76,21 @@ class block(IBlock):
...
+
+    @staticmethod
+    def _refresh_block_reserved_words():
+        block._block_reserved_words = set(dir(block()))
+        # these private attributes are frequently modified
+        block._block_reserved_words.remove('_parent')
+        block._block_reserved_words.remove('_storage_key')
+        block._block_reserved_words.remove('_active')
+        block._block_reserved_words.remove('_byctype')
+        block._block_reserved_words.remove('_order')
```
jsiirola: By removing these from `_block_reserved_words`, you are removing any protections for these (important) attributes.  Wouldn't us be better to leave these in the protected list and just use `object.__setattr__(x)` where you use it?  Or if they really do get used that often, make a helper method that wraps the call to `object.__setattr__()`?

PR #763, opened 2018-11-28
Adding a utility for calculating variable values

jsiirola:
## Summary/Motivation:
This is a generalization of a capability requested by @andrewlee94 for IDAES.  This function will take an equality constraint and a variable, and will calculate the variable value necessary to make the constraint feasible.  It uses a quick calculation if the variable appears linearly, otherwise it falls back on Newton's method.

- adding `calculate_variable_from_constraint()`

### Legal Acknowledgement

### DIFF
--- a/pyomo/util/tests/test_calc_var_value.py
+++ b/pyomo/util/tests/test_calc_var_value.py
@@ -18,6 +18,7 @@ class Test_calc_var(unittest.TestCase):
     def test_initialize_value(self):
         m = ConcreteModel()
         m.x = Var()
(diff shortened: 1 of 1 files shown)


jsiirola:
@andrewlee94: there is actually a test for that (currently line 79).  If the calculated slope is zero, it falls through to the Newton's method routine.  There is a trap there to check to see if the symbolic derivative is actually zero, at which point it raises a `ValueError`.

andrewlee94:
@jsiirola Missed that - it was later in the method than I expected .

PR #772, opened 2018-12-06
Fixing the unicode handling in the Jenkins driver

jsiirola:
## Summary/Motivation:
This should fix the setup for the Book tests in Jenkins to get the build running again.

- change the way we handle subprocess output to run cleanly on all Pythons

### Legal Acknowledgement

### DIFF
--- a/admin/jenkins.py
+++ b/admin/jenkins.py
@@ -87,9 +87,12 @@ elif config == "core":
         _run_cmd("python/bin/pyomo install-extras", shell=True)
     elif _run_cmd is subprocess.check_output:
         output = _run_cmd("python/bin/pyomo install-extras", shell=True)
-        if hasattr(output, 'encode'):
-            output = output.encode('utf-8','replace')
-        print(output.decode('utf-8'))
+        try:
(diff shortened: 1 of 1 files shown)


admin/jenkins.py
```diff
@@ -87,9 +87,12 @@
         _run_cmd("python/bin/pyomo install-extras", shell=True)
     elif _run_cmd is subprocess.check_output:
         output = _run_cmd("python/bin/pyomo install-extras", shell=True)
-        if hasattr(output, 'encode'):
-            output = output.encode('utf-8','replace')
-        print(output.decode('utf-8'))
+        try:
+            print(output)
```
qtothec: Why does it fix the unicode behavior to try to print the output before the `encode` and `decode`?

jsiirola: The encode / decode part is needed for Python3 support.  I could never get it to cleanly back-port to 2.7.  This change (adding the raw `print()`) is just throwing up my hands.  This is a test driver and not core code, so it is more important that it just works than it being clean/efficient.

PR #775, opened 2018-12-07
Rework pynumero's block algebra

santiagoropb:
## Summary/Motivation:

Rework block vector to inherit from np.ndarray. This allows users to directly call numpy ufuncs on block vectors. Drops support special support for symmetric matrices. PyNumero no longer subclasses from scipy sparse matrices but directly populate scipy sparse matrices. This changes make the maintenance of the code much easier and simpler to use from a user standpoint. 

- NLP interfaces provide direct interface to scipy sparse matrices
- Better support for BlockVector that now inherits from np.ndarray
- Simplifications to BlockMatrix
- Direct interface of pynumero NLP objects to cyipopt

### Legal Acknowledgement


pyomo/contrib/pynumero/algorithms/solvers/cyipopt_solver.py
```diff
@@ -0,0 +1,299 @@
...
+    raise ImportError('ipopt solver relies on cyipopt. Install cyipopt'+
+                      ' https://github.com/matthias-k/cyipopt.git')
+import numpy as np
+from pyomo.contrib.pynumero.interfaces import PyomoNLP
+import pyomo.environ as aml
+import sys
+import os
+
+
+def redirect_stdout():
```
jsiirola: This is a general utility, and is very close to an implementation in `pyutilib.misc.redirect_io`.  Is there a reason that you need this implementation?

pyomo/contrib/pynumero/algorithms/solvers/cyipopt_solver.py
```diff
@@ -0,0 +1,299 @@
...
+            d_norm,
+            regularization_size,
+            alpha_du,
+            alpha_pr,
+            ls_trials
+    ):
+        pass
+
+
+class IpoptSolver(object):
```
jsiirola: Several questions about this class:
- Why is this not named `CyIpoptSolver`?  It is specific to that solver interface...
- Why is this not a Pyomo solver?  Or would it be better to make a formal Pyomo Solver that takes a Pyomo model, creates the PyomoNLP, passes it to an instance of this class and calls `solve()`?

pyomo/contrib/pynumero/algorithms/solvers/cyipopt_solver.py
```diff
@@ -0,0 +1,299 @@
...
+
+        x, info = cyipopt_solver.solve(xstart)
+
+        if not tee:
+            os.dup2(newstdout, 1)
+
+        return x, info
+
+
+def create_model1():
```
jsiirola: This method and the ones below it looks like testing functions.  Should they be moved into `pyomo.contrib.pynumero.algorithms.tests`?

pyomo/contrib/pynumero/cmake/asl_interface/src/AmplInterface.cpp
```diff
@@ -524,8 +529,13 @@ extern "C"
                               const_x, nx, const_lam, nc);
    }
 
-   void EXTERNAL_AmplInterface_free_memory(AmplInterface *p_ai) { 
+   void EXTERNAL_AmplInterface_free_memory(AmplInterface *p_ai) {
       p_ai->~AmplInterface();
    }
+
+   void EXTERNAL_AmplInterface_dummy(AmplInterface *p_ai) {
```
jsiirola: Is this a debugging method that was left in?

pyomo/contrib/pynumero/cmake/asl_interface/src/AmplInterface.cpp
```diff
@@ -163,6 +160,7 @@ AmplInterface::~AmplInterface() {
     pi0 = NULL;
     delete[] havepi0;
     havepi0 = NULL;
+    delete oi;
```
jsiirola: Because this is now a class member, it might be good/safe to follow this with `oi = NULL;`

pyomo/contrib/pynumero/cmake/sparse_utils/src/SparseUtils.cpp
```diff
@@ -257,6 +258,58 @@ void sym_csc_diagonal(int* col_pointer,
...
+    {
+        row = i;
+        for(int jj = row_pointer[row]; jj < row_pointer[row+1]; ++jj)
+        {
+            col = col_indices[jj];
+            if(col==row){
+                ++on_diagonals;
+            }
+            else{
+                ++off_diagonals;
```
jsiirola: As written, this method has a computational complexity of O(NNZ).

Not a huge problem, but it might be more performant to only count the diagonals and then subtract to get the off-diagonals.  That would allow you to break out if the inner loop as soon as you see a on-diagonal entry.

I forget: are the columns sorted? If so, you could leverage that and make this more efficient...

pyomo/contrib/pynumero/cmake/sparse_utils/src/SparseUtils.cpp
```diff
@@ -257,6 +258,58 @@ void sym_csc_diagonal(int* col_pointer,
...
+            }
+        }
+    }
+    return 2*off_diagonals + on_diagonals;
+}
+
+int sym_csc_allnnz(int* col_pointer,
+                   int* row_indices,
+                   int ncols)
```
jsiirola: Technically, this method and `sym_csr_allnnz` are identical.  Why not remove this one and rename `sym_csr_allnnz` to `sym_allnnz`?

pyomo/contrib/pynumero/sparse/__init__.py
```diff
@@ -21,13 +21,9 @@
     scipy_available = False
     
 if numpy_available and scipy_available:
```
jsiirola: As a matter of Python style, we are (slowly) removing all use of `from foo import *` from Pyomo, in favor of explicit imports (`from foo import bar, baz`)

pyomo/contrib/pynumero/interfaces/nlp.py
```diff
@@ -1285,36 +1293,41 @@ def hessian_lag(self, x, y, out=None, **kwargs):
...
             assert out.shape[0] == self.nx, "hessian has {} rows".format(self.nx)
             assert out.shape[1] == self.nx, "hessian has {} columns".format(self.nx)
             assert out.nnz == self.nnz_hessian_lag, "hessian has {} nnz".format(self.nnz_hessian_lag)
 
-            data = out.data
-            self._asl.eval_hes_lag(x, y, data)
+            data = np.zeros(self._nnz_hess_lag_lower, np.double)
+            self._asl.eval_hes_lag(x, y, data, obj_factor=obj_factor)
+            values = np.concatenate((data, data[self._lower_hess_mask]))
+            values += 1e-16 # this is to deal with scipy bug temporarily
```
jsiirola: Is there any way to test if that bug is fixed and only apply this hack if it hasn't been fixed?

pyomo/contrib/pynumero/interfaces/nlp.py
```diff
@@ -1881,3 +1722,59 @@ def constraint_idx(self, constraint):
             raise RuntimeError("Constraint must be not indexed")
         return self._conToIndex[constraint]
 
+
+import pyomo.environ as aml
```
jsiirola: This looks like testing code and should be moved to a `tests` directory.  In particular, we should be cautious about "import pyomo.environ" in any Pyomo code outside of examples and tests (it leads to circular imports)

pyomo/contrib/pynumero/interfaces/nlp_compositions.py
```diff
@@ -528,34 +456,38 @@ def evaluate_g(self, x, out=None, **kwargs):
...
+
+                # evaluate coupling Ax-z
+                A = self._AB_csr[sid, sid]
+                res[sid + self.nblocks] = A * x[sid] - x[self.nblocks]
+                # scenario_vids = self._zid_to_vid[sid]
+                # diff = []
+                # for zid in range(self.nz):
+                #     vid = scenario_vids[zid]
+                #     diff.append(x[sid][vid] - x[self.nblocks][zid])
+                # res[sid + self.nblocks] = np.array(diff, dtype=np.double)
```
jsiirola: Should this commented code be removed?  Otherwise, please leave a comment as to why it's here.

pyomo/contrib/pynumero/interfaces/nlp_transformations.py
```diff
@@ -376,45 +368,330 @@ def hessian_lag(self, x, y, out=None, **kwargs):
...
+
+    def expansion_matrix_xl(self):
+        """
+        Returns expansion matrix for lower bounds on primal variables
+        """
+        return self._base_nlp.expansion_matrix_xl()
 
 
+    #jac = nlp.jacobian_c(x)
+    #print(jac.todense())
```
jsiirola: Should these be deleted or moved into `if __name__ == "__main__"`?

pyomo/contrib/pynumero/algorithms/solvers/cyipopt_solver.py
santiagoropb: Thanks for the suggestion @jsiirola . To be honest I was not familiar with pyutilib.misc.redirect_io. I will change cyipopt_solver.py to work with pyutilib and drop the general utilities I coded for output redirection.

pyomo/contrib/pynumero/algorithms/solvers/cyipopt_solver.py
santiagoropb: Right, I will rename this to CyIpoptSolver. 

I agree that at some point CyIpoptSolver should be a formal pyomo solver. For now I have been mostly experimenting with it. I suggest moving CyIpoptSolver to a formal solver in a future pull request. However, I don't think there will be much benefit in creating the PyomoNLP at the solve( ) call. The main advantage of this approach is that users may modified attributes of Pynumero NLPs without rewriting the NL-file. For instance, I have implemented ADMM with this and \rho and \lambda_i can be updated directly in the NLP interface avoiding rewriting the NL file at every NLP iteration. Another good example is solving a TwoStageStochasticNLP that is a composite NLP built pyomo models.

santiagoropb: I will be happy to discuss this further and improve the design of CyIpoptSolver. In particular because it helps a lot for external-decomposition approaches. It can also help towards a persistent interface to NLPs

pyomo/contrib/pynumero/algorithms/solvers/cyipopt_solver.py
santiagoropb: I agree. As I said before this was more of a experimentation face of CyIpoptSolver. I will write couple of tests and move them to algorithms.tests

pyomo/contrib/pynumero/cmake/asl_interface/src/AmplInterface.cpp
santiagoropb: This is left there on purpose. The latest distribution of pynumero_libraries from conda-forge was not allowing for scaling the hessian matrix from the ASL interface. The dummy function was a mechanism to identified if scaling was enabled in the binaries or not. I believed I have updated the conda-forge recipe to now have this functionality as well. For the release I believe this can be kept and I will remove it in a future pull request, once is verified the binaries are working as expected. I will also remove the build rules for pynumero_SPARSE since now all sparse matrices are directly from scipy.

pyomo/contrib/pynumero/interfaces/nlp.py
santiagoropb: I opened an issue on their github project a couple of months ago to get this fixed (#9318). However, so far no news on the fix yet. This, however, is not a problem unless there are explicit zeros in the Hessian, and that the Hessian is added to other sparse matrices. It is a 'soft bug' in the sense that performing addition or subtraction on scipy matrices with explicit zeros  destroys explicit zeros. While the result is correct, the structure of the resulting sparse matrix is not consistent with the explicit zeros of the original matrices. Adding 1e-16 to the values of our hessian ensures those zeros are not destroyed. We can remove line 1329 of our nlp.py when scipy developers fix this. They have agreed that explicit zeros must be preserved and that their original design was supposed to do that.

pyomo/contrib/pynumero/algorithms/solvers/cyipopt_solver.py
```diff
@@ -0,0 +1,299 @@
...
+        for k, v in self._options.items():
+            cyipopt_solver.addOption(k, v)
+
+        if not tee:
+            newstdout = redirect_stdout()
+
+        x, info = cyipopt_solver.solve(xstart)
+
+        if not tee:
+            os.dup2(newstdout, 1)
```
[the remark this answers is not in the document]
santiagoropb: @jsiirola I'm I wrong or capture_output() only redirects stdout and stderr from python calls? If the print statement has been written in an underlying c code capture_output does not redirect to a string. I ran into this when I was trying to redirect the output of cyipopt because the print statement is actually happening at the share library (.so, .dylib) level and not at the python level. For this reason I wrote my own redirect_stdout().

pyomo/contrib/pynumero/interfaces/nlp_transformations.py
santiagoropb: Done

pyomo/contrib/pynumero/interfaces/nlp_compositions.py
santiagoropb: Removed

pyomo/contrib/pynumero/interfaces/nlp.py
santiagoropb: Removed

pyomo/contrib/pynumero/sparse/__init__.py
santiagoropb: Done. No longer importing `*` in pynumero

pyomo/contrib/pynumero/cmake/sparse_utils/src/SparseUtils.cpp
santiagoropb: Removed SparseUtils from `pynumero.cmake` because is no longer used.

pyomo/contrib/pynumero/cmake/sparse_utils/src/SparseUtils.cpp
santiagoropb: Removed SparseUtils from `pynumero.cmake` because is no longer used.

pyomo/contrib/pynumero/algorithms/solvers/cyipopt_solver.py
santiagoropb: Done. Added `pyomo.contrib.pynumero.algorithms.tests.test_cyipopt_solver.py`

pyomo/contrib/pynumero/algorithms/solvers/cyipopt_solver.py
santiagoropb: Renamed IpoptSolver to CyIpoptSolver

pyomo/contrib/pynumero/algorithms/solvers/cyipopt_solver.py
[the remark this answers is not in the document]
santiagoropb: Leaving output redirection as-is for now.

blnicho:
@jsiirola, I disagree. The Pyomo.DAE tests that are failing all have a safety net that checks if scipy is availble before running the test. If the test failures were unrelated to this PR then we would see Appveyor tests failing on other PRs. I'm going to re-run the Appveyor tests to see if it was an intermittent problem or something subtle going on in this PR.

blnicho:
I retract my previous comment, looks like the test failures were an intermittent Appveyor issue

PR #779, opened 2018-12-10
Fix windows stdout issue in gams solver

gseastream:
## Summary/Motivation:
Windows has a compatibility issue when it comes to streaming to stdout vs the console (screen). We came across this in #302, but the GAMS shell solver's `version` method still made a call to `pyutilib.subprocess.run` without specifying `lo=3` to explicitly log to stdout instead of the console. This led to both:

- The `version` method did not work because the `run` call was expecting to capture output from stdout, not console
- The output of the gams call in the `version` method would show on the screen, and since `version` is called in the `solve` method, it would show during a solve as well

- Explicitly log to stdout
- Add a test to make sure `version` does not return `None`

### Legal Acknowledgement

### DIFF
--- a/pyomo/solvers/plugins/solvers/GAMS.py
+++ b/pyomo/solvers/plugins/solvers/GAMS.py
@@ -163,10 +163,9 @@ class GAMSDirect(_GAMSSolver):
             return _extract_version('')
         from gams import GamsWorkspace
         ws = GamsWorkspace()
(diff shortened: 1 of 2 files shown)


gseastream:
Running on my own windows machine, master fails on the shell solver's test. The api solver's version method does not call the executable (it just gets the version number from the api itself), so it was already fine.

PR #795, opened 2019-01-03
Update gdpopt.rst

natnij:
## Summary/Motivation:
keyword for the mip_solver is I believe changed to 'mip' in contrib.gdpopt.GDPopt. With the original sample code, 'mip_solver' is ignored and for 'mip' the default value of 'gurobi' is used, causing a 'ApplicationError: No executable found for solver 'gurobi''.

- change the keyword from 'mip_solver' to 'mip' in sample code for using gdpopt.

### Legal Acknowledgement


qtothec:
Thanks for noticing. `mip` is in fact the old keyword name from the previous release. The Pyomo 5.6 version of GDPopt should use `mip_solver` as the keyword name.

PR #804, opened 2019-01-16
[WIP] Adding DynoptNLP interface, interior-point solver and linear solver i…

santiagoropb:
…nterfaces

## Fixes # .
Fixes bugs in block vector division methods

## Summary/Motivation:
Add interior-point code for research with decomposition approaches e.g. Schur-complement decomposition. This code forms a basis to start testing ideas with decomposition algorithms that exploit structure of optimization problems at the linear algebra level. The current state of pynumero has interfaces to composite nlps that can be decomposed within the interior-point. Examples of these composite nlps are DynopNLP and TwoStageStochaticNLP.

-
-

### Legal Acknowledgement


blnicho:
@santiagoropb this biggest file in this PR at over 8000 lines is `pyomo/contrib/pynumero/algorithms/solvers/tests/cute_models.py`. Are these just standard test models from the CUTE library? If so, I think we already have many (all?) of them implemented in the pyomo-model-libraries repo. Can you rework your tests to use those instead of re-implementing them here?

santiagoropb:
@blnicho Most of the tests in 
`pyomo/contrib/pynumero/algorithms/solvers/tests/cute_models.py` were actually not being used. Only 20 models are used in the tests. I have removed the models that not were used. I have also renamed it to `pyomo/contrib/pynumero/algorithms/solvers/tests/ip_test_models.py` since not all models are from cutertest.

PR #806, opened 2019-01-16
Updating scaling and fixing an error in the DAE Jacobian calculation

eslickj:
## Summary/Motivation:

This brings user scaling to the PETSc solver.  It should use the same suffixes as ipopt, so it should be compatible.  Both equation and variable scaling seem to work for DAE solver, but DAE solver needs more testing in general.   

- Error in DAE Jacobian 
- Error in ASL to PETSc mapping, which was previously unused
- Removed and Added scaling code

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/petsc/examples/dae_pyomo/example01.py
+++ b/pyomo/contrib/petsc/examples/dae_pyomo/example01.py
@@ -28,10 +28,10 @@ if __name__ == "__main__":
     #    (dy6/dt is not explicitly in the equations, so only 5 ydots)
     model.t = Var(initialize=0) #time, but not used anywhere could delete
                                 #using time explicilty in equations is possible
-    model.y = Var([1,2,3,4,5,6])  #
-    model.ydot = Var([1,2,3,4,5]) # dy/dt
-    model.r = Var([1,2,3,4,5])
-    model.Fin = Var()
+    model.y = Var([1,2,3,4,5,6], initialize=1.0)  #
+    model.ydot = Var([1,2,3,4,5], initialize=1.0) # dy/dt
+    model.r = Var([1,2,3,4,5], initialize=1.0)
+    model.Fin = Var(initialize=1.0)
 
     # Equations
     model.eq_ydot1 = Constraint(expr=model.ydot[1] == -2.0*model.r[1] +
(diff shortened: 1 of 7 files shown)


blnicho:
@eslickj is there a reason you didn't use @carldlaird's scaling transformation to scale the model before passing it to the solver? See:  https://github.com/Pyomo/pyomo/blob/master/examples/pyomo/transform/scaling_ex.py

carldlaird:
(Not sure why I cannot reply directly to Bethany's comment above)...

> @eslickj is there a reason you didn't use @carldlaird's scaling transformation to scale the model before passing it to the solver? See: https://github.com/Pyomo/pyomo/blob/master/examples/pyomo/transform/scaling_ex.py

In my opinion, John is doing this correctly. I don't think we should every use the transformation when the solver supports the scaling directly (this is also true with IPOPT). It should be much faster to set the scaling factors in the solver instead of transforming all the expressions and maintaining two instances of the model. The scaling transformation is really for cases when the solver doesn't already support it.

eslickj:
Yeah my goal was to emulate the scaling options in ipopt.  Having those doesn't prevent using Pyomo's scaling transformations, and the ASL handles most of the work.  The only weird thing is that since the DAE Jacobin contains some linear combinations of differential variables and derivative variables, there is probably a chance you could mess up the scaling and not get a solution.  The DAE stuff in here just applies the differential variable scaling to the derivative too regardless of what the user sets the derivative scale factor to.

@blnicho, do you have any test problems you've been using with the DAE stuff?  I guess I should write some tests.

blnicho:
@eslickj you can look in this folder for several Pyomo.DAE examples: https://github.com/Pyomo/pyomo/tree/master/examples/dae

I think `car_example.py` or `Path_Constraint.py` would be good ones to look at. You could easily change constants and coefficients in the models to vary the scaling.

eslickj:
Thanks @blnicho.  I added those examples as best I could since I can't do optimization.  I'll use those to do some more testing.  Unfortunately they showed two new problems.  1)  Sometimes differential variables don't appear in the equations, so they are not in the nl file.  I had to force them in by adding dummy constraints.  I'll have to see if there is a better way to handle that.  2) With the car example I found the variables in the saved trajectory were not in the order I expected, so I'll have to look into finding a good way to match up the saved trajectories to the Pyomo vars.  I thought they would be in the order they were created, but they don't seem to be.

blnicho:
@eslickj is this ready to review or should I wait until the two problems you mentioned are resolved?

eslickj:
@blnicho, I guess its ready to review.  I'm looking for fixes to the two problems, but I'm afraid those will require changes to the NL writer, so we may as well review this, and try to fix the problems later.

eslickj:
Also, I forgot to mention here, but the trajectory information for variables is in the order of the variables in the col-file in case you need to look at those before we have a total solution there.

eslickj:
@blnicho, I went ahead and added something to address the trajectory variable matching problem.  I added a petsc solver plug-in to copy the col file before its deleted (and force symbolic_solver_labels=True for DAE problems), so you don't need keepfiles and you don't need to find the file.  I also have the solver generate a file I called a typ file that has information about whether the variables are algebraic, differential, derivative, or time.  I reworked the vis examples to use those two files and properly line up the variables in the trajectory.  Sorry to make this change before you finish reviewing it.  I think the last problem will require an nl writer change to fix, so I'll definitely stop changing things now.  Also, I did have to make one change outside contrib to have pyomo search for plugins in petsc's location.

blnicho:
I think this is just waiting on a second review. @jsiirola could you take a look?

PR #812, opened 2019-01-18
5.6.1 patch release

blnicho:
This PR updates the CHANGELOG in preparation for the 5.6.1 patch release and provides a workspace for completing the release.

### Legal Acknowledgement


blnicho:
@qtothec this patch release is motivated by the NEOS fix. We think it impacts enough users to warrant a patch release. I think the plan is to not wait as long as we did between the 5.5.1 and 5.6 releases but you also shouldn't expect monthly releases. This is a special case.

PR #819, opened 2019-01-23
Provide a common utility for downloading files

jsiirola:
## Summary/Motivation:
This PR provides a common infrastructure for downloading files into a user's Pyomo installation.  It represents a generalization of the `getGSL.py` and `getGJH.py` scripts.  Other changes include:
 - Adding a `DownloadFactory` for registering downloadable files
 - Adding `pyomo download-extensions` command that loops over the `DownloadFactory` and downloads all files
 - Adding `pyomo.common.fileutils` with convenient utilities for manipulating / finding files.  In particular, adds:
   - `this_file()` which robustly returns the file from where the function was called
   - `this_file_dir()`, which is equivalent to `dirname(this_file())`
   - `find_library()`, which locates a file (library) by looking in common library locations (e.g., `PYOMO_CONFIG_DIR/lib`, `LD_LIBRARY_PATH`, and optionally `PATH`)
   - `find_executable()`, which locates a file (executable) by looking in common locations (e.g., `PYOMO_CONFIG_DIR/bin`, and `PATH`)
   - `ExecutableManager` class, and the `Executable` singleton, which provides a more customizable registry for managing paths to executables (replaces the pyutilib `register_executable` / `registered_executable` infrastructure)
 - Adding `pyomo.common.fileutils.PYOMO_ROOT_DIR`, which points to the root directory of the Pyomo installation
 - Adding `pyomo.common.config.PYOMO_CONFIG_DIR`, which points to a Pyomo configuration directory (nominally `~/.pyomo`, although it can be overridden by the `PYOMO_CONFIG_DIR` environment variable
 - Convert all of Pyomo to use the `pyomo.common.Executable` registry (instead of the pyutilib `register_executable` registry)
 - Update the Travis and Appveyor builds to specify a custom config dir location (within the build directory).

- (see description above)

### Legal Acknowledgement


pyomo/common/fileutils.py
```diff
@@ -8,10 +8,28 @@
 #  This software is distributed under the 3-clause BSD License.
 #  ___________________________________________________________________________
 
+import inspect
 import os
+import platform
 import six
 
 
+def thisFile():
```
ghackebeil: Why camelCase?

jsiirola: Good question.  Random choice? I have removed it.

jsiirola:
This is ready for review.

pyomo/common/fileutils.py
```diff
@@ -0,0 +1,492 @@
...
+    'windows': '.exe',
+    'cygwin':  '.exe',
+    'darwin':  None,
+}
+
+_libExt = {
+    'linux':   ('.so', '.so.*'),
+    'windows': ('.dll',),
+    'cygwin':  ('.dll', '.so', '.so.*'),
+    'darwin':  ('.dynlib',),
```
ghackebeil: change `dynlib` to `dylib`

pyomo/common/plugins.py
```diff
@@ -0,0 +1,14 @@
+#  _________________________________________________________________________
+#
+#  PyUtilib: A Python utility library.
```
ghackebeil: Is this meant to be a PyUtilib header?

ghackebeil:
@jsiirola: Can you comment on the efficiency of your changes to the _default_executable methods on many of the solvers. To summarize, you have changed:
```python
def _default_executable(self):
    executable = registered_executable("ipopt")
    ...
    return executable.get_path()
...
# happens during module import
register_executable("ipopt")
```
to
```python
def _default_executable(self):
    executable = pyomo.common.Executable("ipopt")
    ...
    return executable.path()
```
Note that `_default_executable` is called each time `solve()` is called. Is there any caching going on to make sure that all of the expensive directory walking (especially on HPC) to find an executable does not get repeated? Should the `Executable` declaration perhaps be moved to global module scope?

pyomo/common/fileutils.py
jsiirola: Fixed.

pyomo/common/plugins.py
jsiirola: Fixed.

jsiirola:
@ghackebeil: Performance should be the same, if not better.  `Executable()` caches the resolved path in a similar way that `registered_executable()` did.  The biggest difference is that the search is deferred to the first call to `Executable('foo').available()` or `Executable('foo').path()`, and *not* during module import.  That means that only the executables actually *used* by the Pyomo script are ever registered or resolved.

There are other benefits as well: a user has control over when the search is performed, and can force it to be repeated through 
```
Executable('foo').rehash()  # for a single executable
Executable.rehash()  # to rehash all registered executables
```
and can override the default executable found by the search with 
```
Executable('foo').executable = '/path/to/new/executable'
```
which will persist, even through subsequent calls to `rehash()`.

ghackebeil:
So are you saying there is some caching going on between separate instantiations `Executable(<solver_name>)`?

jsiirola:
Yes. Executable is an instance of ExecutableManager, and its __call__ creates, caches, and returns instances of _ExecutableData.

jsiirola:
Thanks, @ghackebeil.  This is a big enough / low-level enough change that I would appreciate if one or two others (@whart222, @michaelbynum, @blnicho, @carldlaird), put eyes on it; particularly to review the design and naming.  For example, given @ghackebeil's confusion over caching, I wonder if there are better names we can use in the user API.

carldlaird:
@santiagoropb Can you have a look at this?

pyomo/common/config.py
```diff
@@ -9,10 +9,20 @@
...
 
+if 'PYOMO_CONFIG_DIR' in os.environ:
+    PYOMO_CONFIG_DIR = os.path.abspath(os.environ['PYOMO_CONFIG_DIR'])
+elif platform.system().lower().startswith(('windows','cygwin')):
+    PYOMO_CONFIG_DIR = os.path.abspath(
+        os.path.join(os.environ.get('LOCALAPPDATA', ''), 'Pyomo'))
+else:
+    PYOMO_CONFIG_DIR = os.path.abspath(
+        os.path.join(os.environ.get('HOME', ''), '.pyomo'))
+
```
carldlaird: Is there a good reason why the default is ~/.pyomo instead of PYOMO_ROOT_DIR/.pyomo? I feel like putting the default in the pyomo root directory would be better for the many of us who run multiple virtual environments with different versions of Pyomo.

pyomo/common/__init__.py
```diff
@@ -10,11 +10,18 @@
 
 # The log should be imported first so that the Pyomo LogHandler can be
 # set up as soon as possible
-import pyomo.common.log
+from . import log
```
carldlaird: Can you "record" the reason for changing from absolute imports to relative? Is this a practice we should be doing everywhere? If so, why?

pyomo/common/fileutils.py
```diff
@@ -0,0 +1,492 @@
+#  ___________________________________________________________________________
```
carldlaird: These functions look really useful - good addition. Is there an easy way for a Pyomo user to add a directory to the search path for executables / libraries, etc.? Specifically without needing to edit the PATH environment variable, I mean. There are many users who are not familiar with environment variables and maybe this would simplify the fix when Pyomo is unable to find their solver.

pyomo/core/expr/expr_pyomo5.py
```diff
@@ -1178,6 +1178,13 @@ def evaluate_expression(exp, exception=True, constant=False):
             raise
         return None
 
+    except TypeError:
```
carldlaird: Snuck this one into another PR, eh? :)

pyomo/opt/base/opt_config.py
```diff
@@ -9,17 +9,3 @@
 #  ___________________________________________________________________________
 
 import pyomo.common
-
```
carldlaird: Where are these done now? Are they registered in the modules that use them?

pyomo/common/__init__.py
jsiirola: Relative imports used to be discouraged back when the syntax was ambiguous (pre PEP 328).  Now they are unambiguous, they are permitted and used as a matter of personal style (see comment by GvR: https://mail.python.org/pipermail/python-dev/2010-October/104476.html).

I moved to it because @santiagoropb had started using them and I liked how it allowed gently package reorganization without editing a ton of old imports.

pyomo/opt/base/opt_config.py
jsiirola: Explicit registration is not needed / recommended. They get “registered” the first time they are requested. By not registering executables, we avoid the file system search for programs we likely will never need.

pyomo/common/config.py
jsiirola: Yes. First, `$HOME/.app_name` on *NIX and `%APPDATA%\app_name` on Windows is the standard location for applications to store user-specific data/configuration. Second, PYOMO_ROOT_DIR is not guaranteed to be writable, for example when Pyomo was installis into the system Python (and that site-packages directory should only be writable by root/administrator. 

I see no problem with a follow-on PR that adds calls to `expanduser()` and `expandvars()` to the PYOMO_CONFIG_DIR processing, and injecting PYOMO_ROOT_DIR into the environment. However, there is a circular dependency we will have to work through: currently the CONFIG Dir is set before the module that computes/sets the ROOT dir.

pyomo/core/expr/expr_pyomo5.py
jsiirola: Not really.  Accidentally committed it here when I thought I was on another branch, but had already pushed it up by the time I caught the mistake.

pyomo/common/fileutils.py
jsiirola: Yes. `find_library()` and `find_executable()` both take a `pathlist` argument where users can specify / override the list of search paths.  The `Executable` manager has a similar attribute that a user can set:
```python
Executable.pathlist.append(new_dir)
Executable.rehash() # causes the Executable manager to 
#     re-do the search for all known executables
```
Users can also always just *set* the path to the executable of choice:
```python
Executable('my_solver').executable = path_to_my_solver
```
and that will persist despite changes to the `pathlist` or calls to `rehash()`

Having said all of that, there is currently no global "registry" for libraries akin to the `Executable` manager.  Perhaps we should add one later?

pyomo/common/fileutils.py
```diff
@@ -0,0 +1,492 @@
...
+    """Returns the file name for the module that calls this function.
+
+    This function is more reliable than __file__ on platforms like
+    WIndows and in situations where the program has called
+    os.chdir().
+
+    """
+    # __file__ fails if script is called in different ways on Windows
+    # __file__ fails if someone does os.chdir() before
+    # sys.argv[0] also fails because it doesn't not always contains the path
```
blnicho: Typo(?): "doesn't not always"

pyomo/common/fileutils.py
```diff
@@ -0,0 +1,492 @@
...
+
+    .. doctest::
+
+        >>> if Executable('demo_exec_file'):
+        ...     cmd = "%s --help" % Executable('demo_exec_file')
+
+    The :py:class:`ExecutableManager` caches the location / existence of
+    the target executable.  If something in the environment changes
+    (e.g., the PATH) or the file is created or removed after the first
+    time something querried the location or availability, the
```
blnicho: Typo: querried

pyomo/common/fileutils.py
```diff
@@ -0,0 +1,492 @@
...
+    Users may also override the normal file resolution by explicitly
+    setting the files ``executable`` attribute:
+
+    .. doctest::
+
+        >>> Executable('demo_exec_file').executable = os.path.join(
+        ...     pyomo.common.config.PYOMO_CONFIG_DIR, 'bin', 'demo_exec_file')
+
+    Explicitly setting the executable is an absolute operation and will
+    set the location wether or not that location points to an actual
```
blnicho: Typo: wether

pyomo/common/tests/test_fileutils.py
```diff
@@ -0,0 +1,436 @@
...
+        self.assertNotIn('-', _system())
+        self.assertNotIn('_', _system())
+
+    def test_path(self):
+        orig_path = os.environ.get('PATH', None)
+        if orig_path:
+            self.assertEqual(os.pathsep.join(_path()), os.environ['PATH'])
+        os.environ.pop('PATH', None)
+        self.assertEqual(os.pathsep.join(_path()), os.defpath)
+        # PATH restored by teadDown()
```
blnicho: Typo: `tearDown`

pyomo/common/tests/test_fileutils.py
```diff
@@ -0,0 +1,436 @@
...
+
+    def _make_exec(self, fname):
+        open(fname,'w').close()
+        mode = os.stat(fname).st_mode
+        os.chmod( fname, mode | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH )
+
+    def test_this_file(self):
+        self.assertEquals(_this_file, __file__.replace('.pyc','.py'))
+        # Note that in some versions of PyPy, this can return <module>
+        # instead if the normal <string>
```
blnicho: Typo: instead if

jsiirola:
8e0fc66 should address the typos found by @blnicho.

PR #833, opened 2019-01-29
Initial pull request for units

carldlaird:
## Summary/Motivation:
This PR is stage 1 for units handling in Pyomo. The proposed sequence of pull requests is:
- stage 1: implement units creation and the ability to include and validate units within expressions
- stage 2: implement conversion routines to allow for converting units implicitly within expressions and converting numerical values explicitly within (or outside of) expressions
- stage 3: add support to specify units on the components Var and Param and implement model-wide validation functions.

The PR for stage 1 (this PR) implements the PyomoUnitsContainer that can be used to create _PyomoUnit objects. It also includes a visitor implementation _UnitExtractionVisitor that is used to obtain and validate units within expressions.

The units implementation is based on the Python package "pint", however, the details of pint are hidden from the user (in case we ever wanted to change the underlying units engine).

The online documentation has been updated to include the sphinx docs for the module and for the PyomoUnitsContainer.

There is a module level instance of a PyomoUnitsContainer called "units" that should be used for most units interaction. There is no need to create a new instance of a PyomoUnitsContainer. This should simplify the users' interaction with the units package.

-
-

### Legal Acknowledgement


carldlaird:
This PR is failing tests because pint was not installed. Good catch. I will need to add conditional import of pint (so it is not required if it is not used), add skipping to the tests if it is not present, and add pint to the testing infrastructure. No need to review this for a couple of days.

michaelbynum:
@carldlaird, this is awesome. Nice work. The code/implementation looks good to me. I did run into a couple potential issues when trying out units. Whether or not these are issues depends on the expected behavior.

1. An error is raised when calling ```get_units``` on an expression with units which cancel (dimensionless). For example, ```units.get_units(units.m/units.m)```.

2. Some simplification appears to be done in ```get_units```. For example, ```m*s/s``` simplifies to ```m```. However, ```J/N``` does not simplify to ```m```.

Also, it would be really nice if there were a list of units somewhere. I checked the Pint docs briefly, but did not see anything. For a user just beginning to use units, I think a perfectly normal question is something like "does it recognize N, newton, or both?". I realize that an error is raised if the unit is not recognized, so maybe this is not important.

ghackebeil:
@carldlaird, @jsiirola: Mind adding this to kernel?
``` diff
diff --git a/pyomo/kernel/__init__.py b/pyomo/kernel/__init__.py
index 133aa2d7b..f55ba12fd 100644
--- a/pyomo/kernel/__init__.py
+++ b/pyomo/kernel/__init__.py
@@ -23,6 +23,7 @@ from pyomo.opt import (SolverFactory,
 # Define the modeling namespace
 #
 from pyomo.core.expr import *
+from pyomo.core.base.units_container import units
 import pyomo.core.kernel
 from pyomo.kernel.util import (generate_names,
                                preorder_traversal,
```

ReinboldV:
Hi, 
Is this branch available for testing ? Sorry for the beginner's question. I am really interested in trying this out. Also interested in the "online documentation".
Greetings !

ghackebeil:
Yes. You can always test out what is on a branch. One way to do it is to clone the source repo (`git clone https://github.com/carldlaird/pyomo.git`) and then checkout the branch identified in this PR  (`git checkout units-pr-1`). I'm sure there are other ways to do it if you already have a clone of the Pyomo repo.

jsiirola:
To add to what @ghackebeil said: yes, you can play with this branch, but working on a [WIP] (work in progress) branch is about as bleeding edge as it gets. That said, I think this is getting close to being ready to merge.  Please remember that per the PR description, this is only Stage 1 of 3. Full support for units in Pyomo (including Vars and Params) will not work until stage 3 is complete.

pyomo/core/tests/unit/test_units.py
```diff
@@ -0,0 +1,446 @@
...
+
+def python_callback_function(arg1, arg2):
+    return 42.0
+
+@unittest.skipIf(pint_available is False, 'Testing units requires pint')
+class TestPyomoUnit(unittest.TestCase):
+
+    def test_PyomoUnit_NumericValueMethods(self):
+        m = ConcreteModel()
+        uc = units
```
jsiirola: Is `units` coming in through `pyomo.environ`?

pyomo/core/tests/unit/test_units.py
```diff
@@ -0,0 +1,446 @@
...
+        # check __mul__
+        self.assertTrue(str(uc.get_units(kg*3)), 'kg')
+        # check __rmul__
+        self.assertTrue(str(uc.get_units(3*kg)), 'kg')
+        # check div / truediv
+        self.assertTrue(str(uc.get_units(kg/3.0)), 'kg')
+        # check rdiv / rtruediv
+        self.assertTrue(str(uc.get_units(3.0/kg)), '1 / kg)')
+        # check pow
+        self.assertTrue(str(uc.get_units(kg**2)), 'kg**2')
```
jsiirola: Shouldn't this be `assertEqual`?  `assertTrue(x, 'kg')` onlt checks that `x.__bool__()` returns True and if it doesn't, then prints `'kg'` as the error message.

pyomo/core/base/units_container.py
```diff
@@ -0,0 +1,1335 @@
+#  _________________________________________________________________________
+#
+#  Pyomo: Python Optimization Modeling Objects
+#  Copyright (c) 2014 Sandia Corporation.
```
jsiirola: This is the old copyright block

pyomo/core/base/units_container.py
```diff
@@ -0,0 +1,1335 @@
...
+        if pint_module is None:
+            # pint was not imported for some reason
+            raise RuntimeError("The PyomoUnitsContainer in the units_container module requires"
+                              " the package 'pint', but this package could not be imported."
+                              " Please make sure you have 'pint' installed.")
+
+        if self.__pint_registry is None:
+            self.__pint_registry = pint_module.UnitRegistry()
+
+        return self.__pint_registry
```
jsiirola: This has two checks against None every time someone gets the _pint_registry, and could be very expensive (plus the overhead for the property call).  What do you think of the following alternative implementation:
  - Just have the PyomoUnitsContainer directly and unconditionally access the module-level pint_module
  - Swap out the global registry if pint is unavailable.
```python
class PyomoUnitsContainer(object):
    def __init__(self):
        if pint_module is None:
            raise RuntimeError("The PyomoUnitsContainer in the units_container module requires"
                              " the package 'pint', but this package could not be imported."
                              " Please make sure you have 'pint' installed.")
        self._pint_registry = pint_module.UnitRegistry()
    # ....

class PyomoUnitsContainer_NoPint(object):
    def __getattr__(self, attr):
        raise RuntimeError("The PyomoUnitsContainer in the units_container module requires"
                              " the package 'pint', but this package could not be imported."
                              " Please make sure you have 'pint' installed.")
    def __setattr__(self, attr, value):
        raise RuntimeError("The PyomoUnitsContainer in the units_container module requires"
                              " the package 'pint', but this package could not be imported."
                              " Please make sure you have 'pint' installed.")

if pint_module is None:
    units = PyomoUnitsContainer_NoPint()
else:
    units = PyomoUnitsContainer()
```

pyomo/core/base/units_container.py
```diff
@@ -0,0 +1,1335 @@
...
+        except pint_module.errors.UndefinedUnitError as exc:
+            pint_unit = None
+
+        if pint_unit is None:
+            raise AttributeError('Attribute {0} not found.'.format(str(item)))
+
+    def create_PyomoUnit(self, pint_unit):
+        return _PyomoUnit(pint_unit, self._pint_registry)
+
+    # ToDo: Add support to specify a units definition file instead of this programatic interface
```
jsiirola: "TODO"  (todo's in all caps are actually picked up by the Jenkins builds and catalogued)

pyomo/core/base/units_container.py
```diff
@@ -0,0 +1,1335 @@
...
+        if ostream is None: #pragma:nocover
+            ostream = sys.stdout
+        ostream.write(str(self))
+        # There is also a long form, but the verbose flag is not really the correct indicator
+        # if verbose:
+        #     ostream.write('{:!s}'.format(self._pint_unit))
+        # else:
+        #     ostream.write('{:!~s}'.format(self._pint_unit))
+
+def _is_leaf(expr):
```
jsiirola: This is a trivial function actively used in only one place in the code (and that location is inside a tight loop).  [It also appears in an assertion].  Consider in-lining.

pyomo/core/base/units_container.py
```diff
@@ -0,0 +1,1335 @@
...
+        lhsq = (1.0 * lhs).to_base_units()
+        rhsq = (1.0 * rhs).to_base_units()
+
+        if lhsq.dimensionality == rhsq.dimensionality and \
+               abs(lhsq.magnitude/rhsq.magnitude - 1.0) < self._units_equivalence_tolerance:
+            return True
+
+        return False
+
+    def _get_units_relational(self, node, list_of_unit_tuples):
```
jsiirola: This is an exact copy of `_get_unit_for_equivalent_children()`.  DRY: consider removing this method and only using that one.

pyomo/core/base/units_container.py
```diff
@@ -0,0 +1,1335 @@
...
+
+            # ToDo: Check for Var or Param and return their units...
+            # I have a leaf, but this is not a PyomoUnit - (treat as dimensionless)
+            return (None, None)
+
+        # not a leaf - get the appropriate function for type of the node
+        node_type = type(node)
+        if node_type in self.node_type_method_map:
+            node_func = self.node_type_method_map[node_type]
+            if node_func is not None:
```
jsiirola: This seems more complicated than it needs to be.  Why not:
```python
node_func = self.node_type_method_map.get(type(node), None)
if node_func is not None:
    #...
```

pyomo/core/base/units_container.py
```diff
@@ -0,0 +1,1335 @@
...
+            # if we make it here, then no exceptions were thrown
+            return True
+
+        # allow_exceptions is not True, therefore swallow any exceptions in a try-except
+        try:
+            pyomo_unit, pint_unit = self._get_units_tuple(expr=expr)
+        except:
+            return False
+
+        return True
```
jsiirola: This method is a bit strange (and repeats the call to _get_units_tuple).  What about:
```python
try:
    pyomo_unit, pint_unit = self._get_units_tuple(expr=expr)
    return True
except: # Personally, I would also only trap specific exceptions here
    if allow_exceptions:
        raise
    return False
```

pyomo/core/tests/unit/test_units.py
jsiirola: I should have gone back and edited this comment... it is, of course.  I am wondering about the capitalization.  I think I like the lower case, but it does highlight an inconsistency in Pyomo, as the other classes / singleton symbols coming in from `pyomo.environ` are upper CamelCase.  The lower case things are functions (like `value()`) and the weird constants `maximize` and `minimize`,

pyomo/core/tests/unit/test_units.py
carldlaird: Good catch - copy/paste bug - this was in several places.

pyomo/core/tests/unit/test_units.py
carldlaird: Before everything is finalized we should have a broader discussion about the singleton approach in general. There might be a better way.

pyomo/core/base/units_container.py
carldlaird: fixed.

pyomo/core/base/units_container.py
carldlaird: See issue #899

pyomo/core/base/units_container.py
carldlaird: done.

pyomo/core/base/units_container.py
carldlaird: Good call. I frequently forget about the .get with a default.

pyomo/core/base/units_container.py
carldlaird: I purposely did not inline this because it was not at all obvious to me when first looking at the expression code that those checks were, in fact, checking for a leaf. I will change it as you indicate, however, we might want to work on the naming to make this more clear.

pyomo/core/base/units_container.py
carldlaird: I like it. done.

pyomo/core/base/units_container.py
carldlaird: I fixed this, but there appears to be some auto-switching of TODO to .. todo:: that is messing up the docstrings. Therefore, I also moved my long todo list outside the docstring.

carldlaird:
> Hi,
> Is this branch available for testing ? Sorry for the beginner's question. I am really interested in trying this out. Also interested in the "online documentation".
> Greetings !

@ReinboldV: I am still working on the integration with the Pyomo components. It should not take much longer (fingers-crossed).

jonmather:
Hi @carldlaird, I was just wondering if any progress has been made on stages 2 or 3 per your initial post? Thanks for the work so far!

PR #836, opened 2019-01-30
Fix for Appveyor build failure for Python3.7

michaelbynum:
## Fixes # .
Appveyor build failure for Python 3.7

-
-

### Legal Acknowledgement


carldlaird:
Sorry, this was my fault. I meant to indicate --no-update-deps. But I agree with John that I would try the new alias (but I believe that is it --freeze-installed, and not --freeze-install (based on conda documentation online - I did not test or verify).

jsiirola:
@carldlaird: I dug through the conda code and the "portable" argument is `--no-update-deps`.  That is, in it existed before `--no-update-dependencies` was removed and still is valid now.  The new `--freeze-installed` option was only recently added, so switching it would break the Appveyor builds that are currently working.

carldlaird:
Fair enough.

PR #837, opened 2019-01-30
Changing conda --no-update-dependencies option to --no-update-deps.

### LINKED ISSUE #829: Appveyor Failure on Python 3.7
Looks like Appveyor started failing on Python 3.7 with the error message:

    conda-script.py: error: unrecognized arguments: --no-update-dependencies

I think this might be caused by the new conda version (4.6.1) that was just released but I haven't been able to figure out why or how to fix it. I can't find any evidence that `--no-update-dependencies` was deprecated or removed.

jsiirola:
## Summary/Motivation:
This tracks a change in the conda CLI.

- Changing conda `--no-update-dependencies` option to `--no-update-deps`.

### Legal Acknowledgement

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -65,7 +65,7 @@ install:
   #
   - SET CONDA_INSTALL=conda install -q -y
   - "SET ANACONDA=%CONDA_INSTALL% -c anaconda"
-  - "SET CONDAFORGE=%CONDA_INSTALL% -c conda-forge --no-update-dependencies"
+  - "SET CONDAFORGE=%CONDA_INSTALL% -c conda-forge --no-update-deps"
   #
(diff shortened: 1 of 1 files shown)


carldlaird:
@jsiirola Can we briefly document (either here or with a comment in the appveyor files) why we need this option? I have never used it in standard development. Does it speedup some update operations?

jsiirola:
@carldlaird: If I recall correctly, `--no-update-deps` was added to speed up the Appveyor builds.  Because we are using both anaconda and conda-forge, installation steps were updating packages provided by one source with versions provided by the other.  This both caused a slowdown (some large packages were being installed multiple times), and occasionally caused the installation to get itself into a bind where updating a package led to dependency violations and a failed environment setup.

carldlaird:
That is what I suspected.

PR #841, opened 2019-01-31
Update param doc

### LINKED ISSUE #838: Clarify documentation for updating a Param value
The [online documentation](https://pyomo.readthedocs.io/en/latest/working_models.html#changing-the-model-or-data-and-re-solving) recommends that we do repeated solutions with different parameter values by updating instance data like `instance.ParamName[idx].value`. However, at least in Pyomo 5.1.1 or 5.6.1, `instance.ParamName[idx]` is a simple value with no `.value` attribute, so attempting this gives an `AttributeError`. Here is some example code:

    import pyomo.environ as pyo
    i = pyo.ConcreteModel()
    i.param = pyo.Param(['a', 'b'], initialize=lambda m, k: 1)
    i.param.display()
    i.param['a'].value = 2  # AttributeError: 'int' object has no attribute 'value'
    type(i.param['a']) # int

I don't know if this worked in some earlier version of Pyomo, but it doesn't seem to now. So I assume the online documentation needs to be updated.

DLWoodruff:
## Summary/Motivation:

Change documentation concerning assignment to a Param to use direct (i.e., overloaded) assignment
rather than .value; also, start to reduce use of spy4scripts.spy by moving code to be inline.

- Change doc of Param assignment to use direct assignment
- Move some of the code from spy4scripts.spy to be inline (but more work is needed)

### Legal Acknowledgement

### DIFF
--- a/doc/OnlineDocs/working_models.rst
+++ b/doc/OnlineDocs/working_models.rst
@@ -202,7 +202,7 @@ changes to the concrete model instances.  This is particularly important
 for ``AbstractModel`` users, as this implies working with the
 ``instance`` object rather than the ``model`` object, which allows us to
 avoid creating a new ``model`` object for each solve. Here is the basic
-idea:
+idea for users of an ``AbstractModel``:
 
 #. Create an ``AbstractModel`` (suppose it is called ``model``)
 #. Call ``model.create_instance()`` to create an instance (suppose it is called ``instance``)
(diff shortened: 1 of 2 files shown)


doc/OnlineDocs/working_models.rst
```diff
@@ -202,28 +202,62 @@ changes to the concrete model instances.  This is particularly important
...
+    NewVal = 1134
+    
+    instance.x = pyo.Var([1,2,3], initialize=0)
+    instance.y = pyo.Var()
+    instance.iVar = pyo.Var([1,2,3], initialize=1, domain=pyo.Boolean)
+    instance.sVar = pyo.Var(initialize=1, domain=pyo.Boolean)
+
+If ``instance`` has a parameter whose name is ``Theta`` that was
+declared to be ``mutable`` (i.e., ``mutable=True``) with an
 index that contains ``idx``, the the value in ``NewVal`` can be assigned to
```
blnicho: Typo: the the

doc/OnlineDocs/working_models.rst
```diff
@@ -247,23 +281,35 @@ solving the model again. The main lines of interest are:
...
+Note that
 
-.. literalinclude:: script_spy_files/spy4scripts_Equivalent_form_of_instance.x.fix(2).spy
-   :language: python
+   >>> instance.x.fix(2)
 
-and ``instance.x.fix()`` is equivalent to ``instance.x.fixed = True``
+is equivalent to
+
+   >>> instance.y.value = 2
```
blnicho: Why did you use `.value` here instead of directly assigning the value?

doc/OnlineDocs/working_models.rst
```diff
@@ -399,16 +455,21 @@ Access to paramaters is completely analgous to access to variables. For
...
 
 .. note::
 
-   The value of a ``Param`` can be returned as None+ if no data was
+   The value of a ``Param`` can be returned as ``None`` if no data was
    specified for it. This will be true even if a default value was
-   given. To inspect the default value of a ``Param``, replace
-   ``.value`` with ``.default()`` but note that the default might be a
-   function.
+   given. To inspect the default value of a ``Param``, use
```
blnicho: I'm not sure that this note is still true. I did a quick test with mutable and immutable parameters and couldn't find a case where `None` was returned when a default value was specified.

blnicho:
@DLWoodruff I went ahead and made the minor edits that John and I suggested. I'll wait to let you merge it in case you want to take another look.

PR #847, opened 2019-02-04
GDP Branch and Bound Solver

sjkale:
## Summary/Motivation:
Addition of a GDP Branch and Bound solver.  Solver constructs optimal disjunction configuration by relaxing all disjunctions and successively branching on activating each disjunction with a different active disjunct. Branches are explored in order of best solution to the relaxed sub-problem.

### Legal Acknowledgement


qtothec:
@sjkale Is there anything that you can do about improving coverage?

qtothec:
Is `baron` not available on the travis instance? I'm seeing 91% coverage locally.

jsiirola:
BARON is a commercial code, and as such is not available on Travis or Appveyor.  It is available on Jenkins.

qtothec:
Oh, I thought for some reason that we had the non-licensed size-limited version for testing.

jsiirola:
Are the test cases used here small enough to fit under the size limitations for the unlicensed version of BARON?  If they are, then I think it might sense to see if we can include the unlicensed version of BARON in the Travis images.  @ghackebeil?

ghackebeil:
I’ll build some new images and see what breaks.

ghackebeil:
New images are ready. I suspect there are a few models in the solvers/tests/checks/test_writers.py test suite that are slightly too large for the free license. If so, I'll put together a PR that breaks them down into smaller models.

doc/OnlineDocs/contributed_packages/gdpbb.rst
```diff
@@ -0,0 +1,41 @@
+GDP Branch and Bound Solver
+===========================
+
+The GDP Branch and Bound solver is used to solve Generalized Disjunctive
+Programming (GDP) Problems. It branches through relaxed subproblems with
+inactive disjunctions. It explores the possibilities based on best lower bound,
+eventually activating all disjunctions and presenting the global optimal.
+
+
+Using GDB Branch and Bound Solver
```
blnicho: Typo: GDB

doc/OnlineDocs/contributed_packages/gdpbb.rst
```diff
@@ -0,0 +1,41 @@
...
+  >>> m.obj = Objective(expr=m.x1 + m.x2,sense = minimize)
+  >>> m.y1 = Disjunct()
+  >>> m.y2 = Disjunct()
+  >>> m.y1.c1 = Constraint(expr = m.x1 >= 2)
+  >>> m.y1.c2 = Constraint(expr = m.x2 >= 2)
+  >>> m.y2.c1 = Constraint(expr = m.x1 >= 3)
+  >>> m.y2.c2 = Constraint(expr = m.x2 >= 3)
+  >>> m.djn = Disjunction(expr=[m.y1,m.y2])
+
+  Invoke the GDBbb solver
```
blnicho: Typo: GDB

pyomo/contrib/gdpbb/GDPbb.py
```diff
@@ -0,0 +1,252 @@
...
+        with time_code(solve_data.timing, 'total'),\
+                restore_logger_level(config.logger),\
+                create_utility_block(model, 'GDPbb_utils', solve_data):
+            if config.tee and old_logger_level > logging.INFO:
+                # If the logger does not already include INFO, include it.
+                config.logger.setLevel(logging.INFO)
+
+            #Setup results
+            self.setup_results_object(solve_data,model,config)
+            # Initialize ist containing indicator vars for reupdating model after solving
```
blnicho: Typo: ist

pyomo/contrib/gdpbb/GDPbb.py
```diff
@@ -0,0 +1,252 @@
...
+                    heapq.heappush(heap, (ordering_tuple, mnew, result, vars))
+                config.logger.info("Added %s new nodes with %s relaxed disjunctions to the heap. Size now %s." % (
+                    len(next_disjunction.disjuncts), djn_left, len(heap)))
+
+    def validate_model(self, model):
+        # Validates that model has only exclusive disjunctions
+        for d in model.component_data_objects(
+                ctype=Disjunction, active=True):
+            if (not d.xor):
+                raise ValueError('GDPlbb unable to handle '
```
blnicho: What is the 'L' in GDPlbb'?

pyomo/contrib/gdpbb/GDPbb.py
```diff
@@ -0,0 +1,252 @@
...
+                                 'non-exclusive disjunctions')
+        objectives = model.component_data_objects(Objective, active=True)
+        obj = next(objectives, None)
+        if next(objectives, None) is not None:
+            raise RuntimeError(
+                "GDP LBB solver is unable to handle model with multiple active objectives.")
```
blnicho: What is the 'L' in GDP LBB?

pyomo/contrib/gdpbb/GDPbb.py
```diff
@@ -0,0 +1,252 @@
...
+        if obj is None:
+            raise RuntimeError(
+                "GDP LBB solver is unable to handle model with no active objective.")
```
blnicho: Same as previous comment

pyomo/contrib/gdpbb/GDPbb.py
sjkale: Holdover from previous name of solver. Has been resolved

qtothec:
I am confused by the fact that appveyor seems to pass but didn't turn green.

blnicho:
Yeah, looks like a broken connection but I'm not sure if it's on the GitHub or Appveyor side. I haven't noticed this on any other PRs so I'm assuming it was a random fluke. I can try rerunning the Appveyor tests (which will take over an hour to finish) or we can leave it and I can attest that all of the Appveyor tests are passing.

pyomo/contrib/gdpbb/GDPbb.py
```diff
@@ -0,0 +1,252 @@
...
+            root.GDPbb_utils.curr_active_disjunctions = []
+
+            # deactivate all disjunctions in the model
+            # self.indicate(root)
+            for djn in root.GDPbb_utils.unenforced_disjunctions:
+                djn.deactivate()
+            # Deactivate all disjuncts in model. To be reactivated when disjunction
+            # is reactivated.
+            for disj in root.component_data_objects(Disjunct, active=True):
+                disj._deactivate_without_fixing_indicator()
```
jsiirola: Design question for @sjkale, @qtothec, and @emma58: now that we have algorithms relying on this method, should we move to make it a public, officially supported method?

pyomo/contrib/gdpbb/GDPbb.py
```diff
@@ -0,0 +1,252 @@
...
+    """
+    CONFIG = ConfigBlock("gdpbb")
+    CONFIG.declare("solver", ConfigValue(
+        default="baron",
+        description="Subproblem solver to use, defaults to baron"
+    ))
+    CONFIG.declare("solver_args", ConfigValue(
+        default={},
+        description="Dictionary of keyword arguments to pass to the solver."
+    ))
```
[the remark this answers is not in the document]
sjkale: I'm not too familiar with the Config system. What does this do?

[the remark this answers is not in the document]
jsiirola: The config system provides objects that extend the standard Python types: ConfigValue is for holding "plain old data", ConfigList is a list container, and ConfigBlock is a dict container (in hindsight, this should have been ConfigDict, and that is now a valid alias).  The ConfigBlock holds key-value pairs like a normal dict, however, unlike a normal dict, by default a user must "declare" the valid keys.  This is very useful, as it automatically provides error checking for unknown/unallowed options.  However, there are cases where you want a "free form" block of key/value pairs - for example, collecting parameters for an unknown sub-solver.  In this case, you want a ConfigBlock that will accept any keys (just like a normal dict).  You can get that with `implicit=True`.  In the end, using an implicit ConfigBlock is functionally equivalent to a ConfigValue holding a dict, but you get slightly better error and usage reporting with the ConfigBlock.

pyomo/contrib/gdpbb/GDPbb.py
qtothec: I think so. I was never not entirely comfortable with the current `deactivate()` fixes indicator variable to zero scheme, since deactivation feels like it ought to be a problem relaxation, but it can often end up being a problem restriction, as deactivating all the disjuncts in a disjunction yields an infeasible problem.

Maybe we can call it `.relax()`?

jsiirola: I could be OK with `.relax()` (although `relax_disjunct()` might be better).  I think it makes a lot of sense for `.deactivate()` to fix the indicator to 0: if you are removing a disjunct from a model, then the solver should NOT be allowed to select it's indicator variable as a valid solution to the disjunction.  Can you remind me the situation where you want the disjunct removed, but the indicator variable still in the model?  Maybe there is a different API that we should be considering (a `relax` method on a Disjunction?)?

qtothec:
I am going to merge this to let @sjkale work on integration with z3, and open a new PR to address some of the comments above.

pyomo/contrib/gdpbb/GDPbb.py
qtothec: Because I had originally seen `deactivate()` as implying relaxation. By that line of argument, deactivating a `Block` should somehow remove all of its variables from the model? (this thought experiment again)

jsiirola: That's a tough question.  It depends on if a model is defined by it's active constraints, or by everything reachable through an active block path.  The problem is that there is correlated information: the "activeness" of a variable and the use of a variable in an active constraint.  For general variables, one can make an argument that deactivating the block shouldn't effect (restrict) it's use in other constraints.  Now, @carldlaird has pointed out that this problem goes away if variables are not allowed to be referenced outside of the block in which it is attached (and any sub-blocks).  In that case, deactivating the block with the variable will (by definition) deactivate all related constraints.  The downside is that it becomes challenging both to enforce, and to create models (users would have a lot a=b constraints, plus all the linking constraints would be at the sub-block level, which violates encapsulation ... making building up larger models from components more challenging).

For disjuncts it's even bit trickier: the variable *indicates* if the associated constraints are enforced.  Deactivating the block with the constraints, but allowing the indicator variable to be selected violate that semantic "promise".  But, as the indicator variable is typically used in a Disjunction, that relationship is always expressed *outside* the scope of any of the disjuncts.  As the constraint is formed as a transformation, it is easy to exclude the indicator variables from deactivated disjuncts (effectively implying that they are fixed to 0 in that constraint).  However, that does not help other logical constraints that may reference the deactivated disjunct (by leaving the indicator var free, you could see a solver setting it to 1 to satisfy a logical side-constraint (from `Omega(Y)==True`), but not have selected that disjunct in the disjunction.

PR #856, opened 2019-02-06
Adding logic to control signals in shellcmd solvers

whart222:
By default, Pyomo calls PyUtilib to launch solvers using signals.
This causes problems in some multithreaded applications, where Pyomo
is not running in the main thread.  We added control of signal
management to PyUtilib last year, and this branch propagates this
change to Pyomo, using the flag 'use_signal_handling' (which defaults
to True).

## Summary/Motivation:

This adds a 'use_signal_handling' flag, which is passed to the shellcmd solver.

- See above.
- NOTE:  No tests are added with this change.  This change adds the ability to remove error checks from Pyomo solver execution.  It's not clear how to test this, without having tests that fail.

### Legal Acknowledgement

### DIFF
--- a/pyomo/opt/solver/shellcmd.py
+++ b/pyomo/opt/solver/shellcmd.py
@@ -45,6 +45,8 @@ class SystemCallSolver(OptSolver):
         # broadly useful for reporting, and in cases where
         # a solver plugin may not report execution time.
         self._last_solve_time = None
+        self._last_solve_time = None
+        self._define_signal_handlers = True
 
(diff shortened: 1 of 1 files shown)


whart222:
@jsiirola I asked myself would an optimization modeler understand what "define_signal_handlers" meant when calling opt.solve(), and I couldn't convince myself that the answer was "yes".  I think that "use" has a different semantics than "define", especially for a modeling environment like Pyomo.

PR #858, opened 2019-02-07
Updated set_value for indexed components

### LINKED ISSUE #743: Mutable Params should allow should allow updating values through set_value
The Param object should support the "normal" assignment of values through `set_value()`.  That is, a user should expect the following to work:

```
from pyomo.environ import *
m = ConcreteModel()
m.p = Param([1,2],[3,4],mutable=True)
m.p = {(1,3):2, (1,4):4, (2,3):6, (2,4):8}
```
or
```
m.p.set_value({(1,3):2, (1,4):4, (2,3):6, (2,4):8})
```
Unfortunately, the above raises an exception:
```
  File "/home/jdsiiro/Research/pyomo/pyomo/core/base/block.py", line 621, in __setattr__
    _set_value(val)
  File "/home/jdsiiro/Research/pyomo/pyomo/core/base/indexed_component.py", line 676, in set_value
    % (self.name, self.name))
ValueError: Cannot set the value for the indexed component 'p' without specifying an index value.
	For example, model.p[i] = value
```
This issue was originally raised by Juan Esteban Moreno Agudelo on the [forum](https://groups.google.com/forum/#!topic/pyomo-forum/7a96gjYUWYY).

kaklise:
## Summary/Motivation:
Updated set_value for indexed components. Values can now be set using a dictionary.  This includes tests for indexed Var and mutable Params.  If the updated values include an index that is not in the set, validate_index raises an error.

- 
-

### Legal Acknowledgement


pyomo/core/tests/unit/test_indexed.py
```diff
@@ -135,7 +135,40 @@ def test_index_by_unhashable_type(self):
...
+        m.x = vals
+        self.assertEqual(m.x.extract_values(), vals)
+        m.p = vals
+        self.assertEqual(m.p.extract_values(), vals)
+        
+        vals = {1:4, 2:8}
+        m.x.set_value(vals)
+        self.assertEqual(m.x.extract_values(), vals)
+        m.p.set_value(vals)
+        self.assertEqual(m.p.extract_values(), vals)
```
jsiirola: Can you add an 
```
with self.assertRaisesRegexp(...):
    m.x = [2, 4]
with self.assertRaisesRegexp(...):
    m.p = [2, 4]
```
to exercise the exception branch?

pyomo/core/tests/unit/test_indexed.py
```diff
@@ -135,7 +135,40 @@ def test_index_by_unhashable_type(self):
...
+        m.x = vals
+        self.assertEqual(m.x.extract_values(), vals)
+        m.p = vals
+        self.assertEqual(m.p.extract_values(), vals)
+        
+        vals = {(1,3):4, (1,4):8, (2,3):12, (2,4):16}
+        m.x.set_value(vals)
+        self.assertEqual(m.x.extract_values(), vals)
+        m.p.set_value(vals)
+        self.assertEqual(m.p.extract_values(), vals)
```
jsiirola: Can you add a defensive test for scalars:
```
m.x = Var()
m.assertEqual(m.x.extract_values(), {None:None})
m.x.set_value(5)
m.assertEqual(m.x.extract_values(), {None:5})

m.p = Param(mutable=True)
m.assertEqual(m.p.extract_values(), {})
m.p.set_value(6)
m.assertEqual(m.p.extract_values(), {None:6})
```

pyomo/core/base/indexed_component.py
```diff
@@ -673,11 +673,15 @@ def _setitem_when_not_present(self, index, value):
     def set_value(self, value):
         """Set the value of a scalar component."""
         if self.is_indexed():
-            raise ValueError(
+            if type(value) is dict:
```
jsiirola: Should we specifically look for `dict`, or would it be better / more flexible to do:
```
if hasattr(value, '__getitem__') and hasattr(value, '__iter__'):
    for key in value:
        self[key] = value[key]
```
...or maybe just put the thing in a try-catch block and any AttributeError would get mapped to the the ValueError below (I am less enamoured with that, as we have been bitten in the past for suppressing / replacing an exception?

blnicho:
> What should happen to `m.x[2]`?

In response to @jsiirola, I think the value of `m.x[2]` should be left unchanged in that case

jsiirola:
@blnicho: what about the following:
```
m.p = Param([1,2,3], default=0, initialize={2:20}, mutable=True)
m.p = {1:10, 3:30}
```
should `m.p[2]` revert back to it's default value?

...I tend to agree with your answer (even in this case), but I could see situations where a user might expect a different behavior and am curious as to the behavior others expect.

ghackebeil:
Ignoring the fact that I don't support the implicit calls to `set_value` in the first place, my opinion is that `m.p = {1:10, 3:30}` should not be interpreted as anything other than a shorthand for
```python
for key, val in {1:10, 3:30}.items():
    m.p[key] = val
```
no matter how `m.p` was declared. I can't see how a user would wrap their mind around any other behavior.

ghackebeil:
"I can't see how a user would wrap their mind around any other behavior."

I take this part back. I see what you are getting at now. I'm not happy that it's not clear what the correct answer is, which I think highlights a larger issue.

michaelbynum:
I agree with @blnicho that the value of ```m.x[2]``` should be unchanged in that case.

michaelbynum:
Sorry, I missed the second question. I think the value of m.x[2] should be unchanged in either case. I do understand that this is personal preference and different users would expect different behavior.

kaklise:
I just checked in additional tests that set values on a subset of indexed Var or Params (as noted above m.x[2] remains unchanged), checks for expected errors, and sets values on scalars.

qtothec:
If the desired behavior is "leave unchanged indices alone", I would advocate for calling it `update()` or something rather than `set_value()`.

blnicho:
After discussing this during the dev meeting and taking another look at the issue behind this PR, I'm not convinced that changing the `set_value` behavior for every `IndexedComponent` is the best approach for resolving it. As @jsiirola pointed out in the Pyomo forum post that spawned this issue, `IndexedParam` components already have a [`store_values`](https://github.com/Pyomo/pyomo/blob/4997726dd1f11bdb86589ff1c2f4badc654a69ad/pyomo/core/base/param.py#L361) method that can be used to update the values in an indexed mutable param using a dictionary. In addition, this method has a number of extra safe-guards that aren't included in the changes in this PR. I think a better solution to support this assignment syntax for `IndexedParam` would be to override `set_value` and have it call `store_values`. 

Whether this syntax should be supported in general for components that inherit `IndexedComponent` is  a separate design issue.

blnicho:
This PR was discussed at the Pyomo dev meeting today and we decided that the best approach for resolving #743 is to rename `store_values` to `set_value` in `IndexedParam` and deprecate the `store_values` function. The `store_values` function offers the same capability requested in #743 but was added as a performance enhancement for PySP. It doesn't make sense to replicate this functionality in two different places so renaming the function is the best approach for resolving the issue and maintaining the performance enhancements used by PySP and invoked by the optional `check` keyword argument. We are going to close this PR and implement the fix in a new PR.

PR #862, opened 2019-02-13
Satisfiability Solver using z3

sjkale:
## Summary/Motivation:
Implementation of Satisfiability Solver interface for Pyomo.
Converts pyomo variables and constraints to standard SMT-LIB form and uses z3 prover to check for satisfiability.

[More about SMT-LIB form](http://smtlib.cs.uiowa.edu/)
[z3 Github](https://github.com/Z3Prover/z3)
- Added SatSolver object that can take a Pyomo model and check satisfiability

### Legal Acknowledgement


sjkale:
Would anyone know why the appveyor test is failing to install the z3 solver?

jsiirola:
It looks like pip is trying to build Z3 using MSVS and cannot find it.  The relevant part of the build log is:
```
  Makefile was successfully generated.
    compilation mode: Release
    platform: x86
  To build Z3, open a [Visual Studio Command Prompt], then
  type 'cd c:\users\appveyor\appdata\local\temp\1\pip-install-4qsfgi\z3-solver\core\build && nmake'
  
  Remark: to open a Visual Studio Command Prompt, go to: "Start > All Programs > Visual Studio > Visual Studio Tools"
  Building Z3
  error: [Error 2] The system cannot find the file specified
```
We generally haven't been trying to build C/C++ code as part of the Pyomo build, so it doesn't surprise me that MSVS is not configured.  It might be possible to install a pre-built version of Z3 using conda (anaconda cloud indicated that there is a z3-solver package through the `asmeurer` channel.  We generally only use conda-forge, but it might work.

pyomo/contrib/satsolver/satsolver.py
```diff
@@ -0,0 +1,273 @@
...
+        self.walker = SMT_visitor(self.variable_label_map)
+        self.solver = z3.Solver()
+
+        if model is not None:
+            self._process_model(model)
+
+    # Set up functions to be added to beginning of string
+    def _get_default_functions(self):
+        default = list()
+        default.append("(define-fun exp ((x Real)) Real (^ 2.718281828459045 x))")
```
jsiirola: Instead of embedding this constant, can you compute it:
```
default.append("(define-fun exp ((x Real)) Real (^ %0.15f x))" % (exp(1),) )
```

pyomo/contrib/satsolver/satsolver.py
```diff
@@ -0,0 +1,273 @@
...
+        return default
+
+    # processes pyomo model into SMT model
+    def _process_model(self, model):
+        for v in model.component_data_objects(ctype=Var, descend_into=False):
```
jsiirola: This looks like it only handles "flat" models with Disjunctions.  Sort of.  You will only pick Vars on the top-level model, but you will pick up Constraints in any active Block.  Is that the desired behavior?

pyomo/contrib/satsolver/satsolver.py
```diff
@@ -0,0 +1,273 @@
...
+            self.variable_list.append("(declare-fun " + label + "() Real)\n")
+            self._add_bound(var)
+        elif domain is IntegerSet:
+            self.variable_list.append("(declare-fun " + label + "() Int)\n")
+            self._add_bound(var)
+        elif domain is BooleanSet:
+            self.variable_list.append("(declare-fun " + label + "() Int)\n")
+            self._add_bound(var)
+        else:
+            raise NotImplementedError("SMT cannot handle" + str(domain) + "variables")
```
jsiirola: It is good that this is defensive, but there are a number of other domains that are commonly used that you will not catch.  In particular, NonNegativeReals and PositiveIntegers.  See pyomo/core/kernel/set_types.py

pyomo/contrib/satsolver/satsolver.py
```diff
@@ -0,0 +1,273 @@
...
+        for disj in djn.disjuncts:
+            constraints = []
+            iv = disj.indicator_var
+            label = self.add_var(iv)
+            xor_expr = "(+ " + xor_expr + " " + label + ")"
+            for c in disj.component_data_objects(ctype=Constraint, active=True):
+                constraints.append(self.walker.walk_expression(c.expr))
+            disjuncts.append((label, constraints))
+        xor_expr = "(assert (= 1 " + xor_expr + "))\n"
+        self.disjunctions_list.append((xor_expr, disjuncts))
```
jsiirola: Note that the disjunction is only an XOR if the `djn.xor` is True.  Otherwise it is a simple OR.

.travis.yml
```diff
@@ -58,6 +58,8 @@ install:
   - ${DOC} git clone --quiet https://github.com/Pyomo/pyomo-model-libraries.git
  # Install PyUtilib (master branch)
   - ${DOC} pip install --quiet git+https://github.com/PyUtilib/pyutilib
+ # Install z3 Satisfiability solver
+  - ${DOC} pip install --quiet z3-solver
```
jsiirola: We should not install z3 here.  Instead, this should be added to the Docker -- and REMOVED for SLIM builds.  I will go ahead and put in the PR to get z3 added to the Docker image.  What is the plan for Appveyor?  Does it not (cleanly) support z3?

qtothec: Thanks John. I was having a bit of difficulty getting Z3 to properly install on appveyor, either with pip or conda, as suggested

jsiirola: For the record, I am OK only testing this on Travis/Jenkins -  was just wondering what the final verdict for Appveyor was.

pyomo/contrib/satsolver/satsolver.py
sjkale: I was under the impression that those would still be handled, since they are all instances of either of those 3 classes. I will add some tests that confirm this

pyomo/contrib/satsolver/satsolver.py
sjkale: Thanks for the catch, believe it was left in from debugging.

pyomo/contrib/satsolver/satsolver.py
jsiirola: While it is true that things like NonNegativeReals are *instances* of RealSet, the test in your code is `is`, and `Reals is RealSet == False`.  I think what you want is:
```
if isinstance(domain, RealSet):
```
(and so on)

jsiirola:
@qtothec: I don't see any problem merging this.  If you will be the second reviewer, I think we can be good to go.  Note that the functionality will be somewhat limited until relax the tests around satsolver.py line ~134.

pyomo/contrib/satsolver/satsolver.py
sjkale: Okay thanks! That makes sense

qtothec: @jsiirola What you say would be true if he had just done `var.domain is RealSet`. But for some reason, Sunjeev's logic works because of `domain = type(var.domain)`, which is a bit confusing offhand to me.

jsiirola: Ahh - my mistake: I had missed that detail in his implementation.

PR #863, opened 2019-02-14
Improvements for FBBT

michaelbynum:

1) Improved support for binary variables in FBBT. For example, if the lower bound is computed to be 0.5, then it is actually 1.
2) Add an option to control whether or not variable bounds are updated. Also return a ComponentMap with the new variable bounds.
3) Add an option to deactivate constraints which are always satisfied.
4) Don't raise an error when encountering an unknown expression type. Instead, just issue a warning and don't improve the bounds.

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/fbbt/fbbt.py
+++ b/pyomo/contrib/fbbt/fbbt.py
@@ -663,15 +663,15 @@ class _FBBTVisitorLeafToRoot(ExpressionValueVisitor):
     This walker propagates bounds from the variables to each node in
     the expression tree (all the way to the root node).
     """
-    def __init__(self, bnds_dict, binary_tol=1e-4):
+    def __init__(self, bnds_dict, integer_tol=1e-4):
         """
         Parameters
         ----------
         bnds_dict: ComponentMap
-        binary_tol: float
+        integer_tol: float
         """
         self.bnds_dict = bnds_dict
-        self.binary_tol = binary_tol
+        self.integer_tol = integer_tol
 
     def visit(self, node, values):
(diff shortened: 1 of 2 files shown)


jsiirola:
...that said, @michaelbynum: can you add a test to exercise this code?  Ideally, you could create a "fake" expression node type and include it in a simple constraint and verify that the warning is issued.

michaelbynum:
@jsiirola Good point. That is an easy test to add.

pyomo/contrib/fbbt/fbbt.py
```diff
@@ -729,10 +735,26 @@ def visiting_potential_leaf(self, node):
 
         if node.is_variable_type():
             lb, ub = self.bnds_dict[node]
-            if lb != -math.inf:
-                node.setlb(lb)
-            if ub != math.inf:
-                node.setub(ub)
+            if node.is_binary():
```
jsiirola: Isn't this a special case of general integers?  Would it be difficult to change this that if the variable is discrete, then:
```
if node.is_binary() or node.is_integer():
    lb = max(lb, math.ceil(lb - self.binary_tol))
    ub = min(ub, math.floor(up + self.binary_tol))
    self.bnds_dict[node] = (lb, ub)
```
(The max / min is probably not necessary, but I include it for safety)

...and then it would make sense to rename `binary_tol` to `integer_tol`.

michaelbynum: @jsiirola Good suggestion. I just pushed this change. I like your approach a lot more than my if-else statements. Thanks.

PR #865, opened 2019-02-15
Fixing typo in Var docstring

datapythonista:


pyomo/core/base/var.py
```diff
@@ -471,7 +471,7 @@ class Var(IndexedComponent):
             Defaults to (None, None).
         initialize (float or function, optional): The initial value for
             the variable, or a rule that returns initial values.
-        rule (function, optional): An alias for `rule`
+        rule (function, optional): An alias for `domain`
```
jsiirola: This is indeed a typo, but this is not the correct fix: per the code below, rule is an alias for `initialize`.  `within` is an alias for `domain`.

datapythonista: You're right, sorry about that. Should be ok now. Thanks!

PR #877, opened 2019-02-22
Disjunctive variable bound tightening with GLOA

qtothec:
## Summary/Motivation:
Variables often have improved bounds within a disjunctive context (e.g. a Disjunct) than in the global model scope. This PR allows the GLOA algorithm to exploit this fact when generating cuts based on the generalized McCormick envelopes for a GDP model.

- Introduce small toy problem
- modify introduce additional function for accessing disjunctive variable bounds
- Clean up previously contributed code
- Modify MC++ code to allow optional specification of improved variable bounds
- Modify GDPopt code to allow optional calculation of disjunctive variable bounds
- Enable use of FBBT to calculate disjunctive variable bounds.

### Legal Acknowledgement


qtothec:
Looks like poor codecov is because I touched MC++, which we don't have a way of testing yet.

pyomo/contrib/gdp_bounds/plugins/compute_bounds.py
```diff
@@ -13,15 +13,28 @@
...
+
+linear_degrees = {0, 1}
+
+
+def disjunctive_bounds(scope):
+    """Return all of the variable bounds defined at a disjunctive scope."""
+    possible_disjunct = scope
+    while possible_disjunct is not None:
+        try:
+            return possible_disjunct._disj_var_bounds
```
michaelbynum: @qtothec Is it possible (and not too much trouble) to do this without accessing private attributes?

qtothec: Well, I actually stick `_disj_var_bounds` onto the Disjunct object if it doesn't exist, so I guess I could rename it to `disj_var_bounds`?

michaelbynum: @qtothec Oh, I misunderstood what was going on there. Okay. That is perfectly reasonable. I assume it is removed from the disjunct before the algorithm exits?

michaelbynum:
@qtothec One more question before this gets merged. Did the new example come from anything that needs a reference (e.g., a paper)?

pyomo/contrib/gdp_bounds/plugins/compute_bounds.py
qtothec: No, because it is intended to be used elsewhere as well (e.g. in GDPopt)

qtothec:
@michaelbynum This PR is still a bit WIP because I intend to clean up the implementation of the disjunctive bounds calculation. This is actually the work I was talking about when I wanted FBBT to be able to give me the updated variable bounds without actually changing the variable bound values.

In terms of a paper, it's in preparation.

michaelbynum:
@qtothec Sorry... I did not notice the WIP...

qtothec:
@michaelbynum No worries. I was surprised by the attention that my humble WIP PR was getting from you :joy:.

qtothec:
@michaelbynum I had to poke a bit at FBBT to make it work with disjunctive scopes. Can you make sure that I didn't mess anything up?

michaelbynum:
@qtothec I would like to understand the motivation for the initial bounds dictionary. What prompted this?

qtothec:
@michaelbynum I'll refer you to this test problem: https://github.com/Pyomo/pyomo/pull/877/files#diff-dc2031eff720806b77360ab55f26d54cR68

Say you have a structure `[x >= 2, [y >= x + 3] V [y <= x - 4]] V [something else]` with `0 <= x, y <= 8`. You would want to have within the scope of the nested Disjunct `y >= x + 3` that the bounds on `y` are `5 <= y <= 8` rather than just `3 <= y <= 8`.

michaelbynum:
@qtothec Thanks. I understand now. I think this can all be done with the bnds_dict created in fbbt_con. Then the walkers would not have to be changed at all. The bnds_dict can be an optional argument and if it is None, then create a new component_map. Do you agree?

michaelbynum:
@qtothec By the way, I don't think your changes would break anything. I just want to keep the walkers as simple as possible...

qtothec:
@michaelbynum I don't remember if I said this already, but the change you have in the PR towards this branch doesn't pass the nested Disjuncts test.

michaelbynum:
@qtothec Okay, I will take a look.

qtothec:
@michaelbynum your bounds manager worked. =)

michaelbynum:
@qtothec Great! I'm glad to hear it. I'm sorry for being picky about this. I'll finish reviewing this PR first thing in the morning.

PR #886, opened 2019-02-28
Incorporating z3 Sat Solver into GDPbb

sjkale:
## Summary/Motivation:
Allows option to check satisfiability at each node in Branch and Bound. The goal is to prune unsatisfiable nodes without attempting to solve.  Currently slows down the tests by a lot, but is functional. 

-Add check_sat option that checks satisfiability at each node in tree. Defaults to false

### Legal Acknowledgement


qtothec:
You should include a @skip on the test if z3 isn't available.

qtothec:
@sjkale Why not test variants both with and without the sat solver?

PR #885, opened 2019-02-27
Clean up z3satsolver imports

sjkale:
## Summary/Motivation:
z3 imports not working correctly when used in solvers

## Changes proposed in this PR:
- cleaned up imports

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

Issue #483, opened 2018-05-15
Discussion: Refactoring the new expr folder in Pyomo


carldlaird:
Discussion: Maybe we should consider changing the name of current.py. Should we also consider partitioning expr_pyomo5.py into smaller parts?

whart222:
Yep to both.  @jsiirola suggested that we eliminate 'current' altogether, so I think it's worth doing that.

blnicho:
This was completed in #888

Issue #439, opened 2018-04-23
Pyomo expression design issues


whart222:
The following design issues should be addressed after we merge the expr_dev branch:

- [x] Evaluate whether division expressions would be better than reciprocal expressions
- [ ] Review whether identify_components should use ctypes
- [x] Reorganize pyomo.core.expr and eliminate support for swapping in/out a new expression system.  (We can simply develop alternatives in a branch, so the 'current' module isn't needed.)  This reorganization should eliminate the inclusion of '_*' symbols from all modules.
- [ ] Review clone_expression, which is only used to expand connectors.  Is this logic needed?
- [x] Add a logical expression that is separate from non-logical expressions.
- [ ] Evaluate the known constants and as_numeric() logic.  These may not be needed now that we include numbers in expression trees.

qtothec:
`clone_expression` has uses beyond just connector expansion, because of its `substitute` functionality (e.g. for a variable aggregator)

carldlaird:
I suspect that this functionality ('substitute') is handled through a new API in the new expression system.

blnicho:
The design issues above were related to the Pyomo5 expression system. We are now using the Pyomo6 expression system so this list of design issues is out-dated. I'm going to close this and we can open new issues for these design discussions as-needed.

PR #888, opened 2019-02-28
Split expr_pyomo5.py into three files

jsiirola:
## Summary/Motivation:
This is a (giant) nonfunctional change.  It simply splits the existing Pyomo5 expression into 3 separate files to manage the size / complexity of the expression system.  I propose this change because we are working on expanding the support for Logical expressions, which will eventually touch the main expression system, and keeping everything in a single file is becoming unwieldy.

- split `expr_pyomo5.py` into `logical_expr.py`, `numeric_expr.py` and `visitor.py`.

### Legal Acknowledgement


pyomo/contrib/derivatives/differentiate.py
```diff
@@ -1,6 +1,7 @@
 from pyomo.core.kernel.component_map import ComponentMap
-import pyomo.core.expr.expr_pyomo5 as _expr
-from pyomo.core.expr.expr_pyomo5 import ExpressionValueVisitor, nonpyomo_leaf_types, value
+import pyomo.core.expr.current as _expr
```
qtothec: why not just import from numeric_expr?

jsiirola: No real reason.  I was trying for the minimal set of changes.  Eventually, I would like it so that users outside of `pyomo.core` never have to import anything other than `pyomo.core.expr`.  But that will be a more involved change, with possible problems with backwards compatibility.

pyomo/contrib/satsolver/satsolver.py
```diff
@@ -2,18 +2,27 @@
...
+    AbsExpression,
+    UnaryFunctionExpression,
+)
+from pyomo.core.expr.numvalue import (
+    nonpyomo_leaf_types,
+)
+from pyomo.core.expr.visitor import (
+    StreamBasedExpressionVisitor,
+)
+from pyomo.environ import SymbolMap, NumericLabeler, Var, Constraint
```
qtothec: I think we want to import these from `pyomo.core` instead. See circular reference PR #885.

blnicho:
This PR is also related to #439 and #483

qtothec:
I too enjoy smaller files, and more and more I am having people who have an interest in modifying these files.

PR #891, opened 2019-03-05
add .coin-or/projDesc.xml

svigerske:
## Fixes # .

Nothing.

## Summary/Motivation:

This adds the project description file that is used by the COIN-OR server to the assemble list of projects for the COIN-OR website.
Because the svn repositories will go away eventually, we move these files into the projects source trees.

Add copy of https://projects.coin-or.org/Pyomo/browser/conf/projDesc.xml into Pyomo git repository.

### Legal Acknowledgement

### DIFF
--- a/.coin-or/projDesc.xml
+++ b/.coin-or/projDesc.xml
@@ -1,0 +1,337 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?xml-stylesheet type = "text/xsl" href = "http://www.coin-or.org/projects/autoGen.xsl"?>
+<projectData xmlns="coin-or.org" 
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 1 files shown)


blnicho:
@svigerske when will the COIN-OR server begin using this file instead of the old one? I noticed that the Pyomo version number and supported Python versions are out of date. Is fixing them here sufficient or should we also fix them in the old file location?

svigerske:
It may be best to fix them at both locations. It's rather a matter of weeks or months than days until the old location will not be used anymore.

jsiirola:
I think this is OK to merge.  I tried to push the updated projDesc.xml to the current COIN SVN repo, but it appears that only @whart222, Matt, and two "synchronization" entity accounts have write access to that repo (and I can't find the credentials for the entity accounts).

PR #897, opened 2019-03-13
Pyomo View Exception Handling

eslickj:
## Summary/Motivation:
If there is an exception that is not caught coming from GUI code it would terminate the the application immediately.  These types of exceptions are probably indicative of a bug, and should be rare, but you definitely never want the GUI application to suddenly terminate without warning. So I added a thing to catch all exceptions that are not specifically caught elsewhere and write a log message and show a dialog box with traceback.  The GUI will keep on going after the exception.

- Added unhandled exception catch in the stand alone pyomo viewer GUI 

### Legal Acknowledgement


pyomo/contrib/viewer/pyomo_viewer.py
```diff
@@ -12,9 +12,29 @@
 Run a Pyomo model viewer with embeded Jupyter QtConsole
 """
 
+import sys
 import logging
+import traceback
 _log = logging.getLogger(__name__)
 from pyomo.contrib.viewer.ui import *
+import pyomo.environ as pyo
```
qtothec: This import may cause issues if you intend any other auto-registered plugins to be importing this file.

eslickj: This file shouldn't be imported by anything else.  Its just a script to run the viewer. Anything you may want to import would be elsewhere.  To be safe though, I can move that back inside the main function.

qtothec: That's fine, then. Passing comment since I ran into something similar with one of my modules.

pyomo/contrib/viewer/pyomo_viewer.py
```diff
@@ -12,9 +12,29 @@
...
+    """
+    This will log an exception and show a message box about what went wrong, but
+    let the application keep running.  It could be frustrating if an unhandled
+    exception caused the program to immediatly terminate.
+    """
+    try:
+        _log.critical("unhandled exception", exc_info=(etype, evalue, etrace))
+        msgBox = QMessageBox()
+        msgBox.setWindowTitle("Error")
+        msgBox.setText("Unhandled Exception")
```
jsiirola: Should there also be a disclaimer that the error should be reported to the developers and that the GUI should be exited immediately?  After an unhandled exception like this, I am not sure there is much we can say about the validity / stability of the underlying object model, right?

eslickj: You're right.  I'll make that update as soon as I get a chance.

eslickj: I added to the message,. While I was at it I added handling of the KeyboardInterrupt exception and bailed on the ui.py import if pyqt isn't available.   I have some dummy classes so I could test the data model, but it was causing the ui module import to get too far before failing in a weird way, so this make thing a bit easier to understand.

pyomo/contrib/viewer/ui.py
```diff
@@ -46,6 +46,10 @@ class _MainWindowUI(object):
     class _MainWindow(object):
         pass
 
+if not qt_available:
+    _log.error("Qt is not available. Cannot create UI classes.")
+    sys.exit(1)
```
jsiirola: I missed this the first time around, but I think that this "`if`" needs to be re-thought.  The current implementation will forceably kill the Python interpreter if `not qt_available`.  That will wreak havoc in our slim builds and for any users that do not have Qt.

eslickj: Yeah.  You're right I'll raise an import error.

eslickj: Replaced the with raise import error and tested without PyQt.  The result is a pretty easy to understand exception.  Everything seems good.

PR #904, opened 2019-03-19
Fix bug in active variable detection for model size report

qtothec:
## Summary/Motivation:
Fixes logic error due to out-of-scope variable usage not caught earlier because I wrote things in too much of a hurry.

Adds test to exercise detection.

### Legal Acknowledgement

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -100,7 +100,10 @@ install:
   #   that usually come with the full Anaconda distribution
   #
   - SET MINICONDA_EXTRAS=""
-  - IF DEFINED USING_MINICONDA (SET MINICONDA_EXTRAS=numpy scipy ipython openpyxl sympy pyodbc pyyaml networkx xlrd pandas matplotlib dill seaborn)
+  - IF DEFINED USING_MINICONDA (SET MINICONDA_EXTRAS=numpy scipy ipython openpyxl sympy pyodbc pyyaml<=3.12 networkx xlrd pandas matplotlib dill seaborn)
+  #
+  #    Note: We are fixing pyyaml to 3.12 because newer versions break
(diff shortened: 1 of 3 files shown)


qtothec:
I'm having a hard time figuring out why this PR would cause dataportal failures in Appveypor Python 3.7.

jsiirola:
It looks like this build was picking up a different YAML implementation, causing different but equivalent YAML files to be generated.  I was seeing diffs like:
```
-  index: [1, 2]
+  index:
+     - 1
+     - 2
```
This is likely pointing to a fragility in the YAML comparison.  I am re-running the 3.7 build now to see if it "just goes away", and if it doesn't, then we can dig into why the YAML tests are fragile.

qtothec:
@jsiirola fwiw, I re-ran the test earlier and it gave the same result.

jsiirola:
OK - I just looked:  The 2.7 - 3.6 builds are installing
```
    pyyaml-3.12                |           py36_1         121 KB  conda-forge
```
whereas the 3.7 build is installing
```
    pyyaml-5.1                 |   py37he774522_0         147 KB
```
(note that that is not coming from conda-forge.  I think the short-term solution is to fix the pyyaml version on appveyor to 3.12 until we have a chance to fix the DataPortal tests.  I propose making that chance on this PR.

qtothec:
Are you going to push that change onto this branch? I'm not sure how to go about doing that.

jsiirola:
Sure.  I can do that.

jsiirola:
@qtothec: OK - the build is passing.  Have you verified (by hand?) that the updated baselines are actually correct?

qtothec:
@jsiirola Yes, it makes sense, because none of those models should start with deactivated components.

PR #911, opened 2019-03-22
Pyomo Viewer, Improved error handling for Expression and Constraint calculation

eslickj:
## Summary/Motivation:
There was a problem when calculation of expressions values ended in an error in the Pyomo Viewer.  I improved the error handling for both expression and constraint calculation and added a bunch of tests for different kinds of errors.  I use the ```exception=False``` arg for ```value()``` and so far the only error it doesn't seem to catch is division by zero.  I assume there is a good reason for that, so I catch division by zero myself.  Any error not handled by those two things will still fall through and  1) halt the calculations and 2) trigger the unhanded exception message.

I'm a little hesitant to just ignore all exceptions when doing calculations in case I make a coding mistake, so I'm just going to try to catch all the expected cases.

- Improved error trapping when calculating constraint and expression values
- tests for above
- add information about installing (ok not installing exactly) to readme

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/viewer/model_browser.py
+++ b/pyomo/contrib/viewer/model_browser.py
@@ -179,10 +179,14 @@ class ComponentDataItem(object):
                 self._cache_value = value(self.data.body, exception=False)
             except ZeroDivisionError:
                 self._cache_value = "Divide_by_0"
-            # I think these are constants, so don't think I need try/except
-            self._cache_lb = value(self.data.lower, exception=False)
-            self._cache_ub = value(self.data.upper, exception=False)
-
+            try:
+                self._cache_lb = value(self.data.lower, exception=False)
+            except ZeroDivisionError:
+                self._cache_lb = "Divide_by_0"
+            try:
+                self._cache_ub = value(self.data.upper, exception=False)
+            except ZeroDivisionError:
+                self._cache_ub = "Divide_by_0"
 
     def get(self, a):
         """Get an attribute"""
--- a/pyomo/contrib/viewer/model_browser.py
+++ b/pyomo/contrib/viewer/model_browser.py
@@ -170,14 +170,19 @@ class ComponentDataItem(object):
     def calculate(self):
         """Calculate items, applies to expressions and constraints"""
         if isinstance(self.data, _ExpressionData):
-            self._cache_value = value(self.data)
+            try:
+                self._cache_value = value(self.data, exception=False)
+            except ZeroDivisionError:
+                self._cache_value = "Divide_by_0"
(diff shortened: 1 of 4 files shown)


pyomo/contrib/viewer/tests/test_data_model_item.py
```diff
@@ -38,18 +44,62 @@ def blackbox(a,b):
...
 
     def test_expr_calc(self):
         cdi = ComponentDataItem(parent=None, ui_setup=None, o=self.m.b1.e1)
         cdi.calculate()
         assert(abs(cdi.get("value")-3) < 0.0001)
 
+    def test_expr_calc_div0(self):
+        cdi = ComponentDataItem(parent=None, ui_setup=None, o=self.m.b1.e2)
+        cdi.calculate()
+        assert(cdi.get("value") == "Divide_by_0")
```
jsiirola: I am not going to block this PR (as the tests are still valid), but in the future, the "unittest way" to assert equality is:
```
self.assertEqual(cdi.get("value"), "Divide_by_0")
```

pyomo/contrib/viewer/model_browser.py
```diff
@@ -170,14 +170,19 @@ def calculate_children(self):
...
                 self._cache_value = value(self.data.body, exception=False)
-                self._cache_lb = value(self.data.lower, exception=False)
-                self._cache_ub = value(self.data.upper, exception=False)
-            except:
-                pass
+            except ZeroDivisionError:
+                self._cache_value = "Divide_by_0"
+            # I think these are constants, so don't think I need try/except
+            self._cache_lb = value(self.data.lower, exception=False)
+            self._cache_ub = value(self.data.upper, exception=False)
```
jsiirola: These are constants, but that doesn't mean that they will not raise divide-by-zero.  Consider:
```
m = ConcreteModel()
m.x = Var()
m.p = Param(mutable=True, initialize=1)
m.c = Constraint(expr = m.x <= 1/m.p)
m.p = 0
```

pyomo/contrib/viewer/tests/test_data_model_item.py
eslickj: I may as well update these, while I'm fixing things related to the other comment.

pyomo/contrib/viewer/model_browser.py
eslickj: I see how this could be a problem.  I'll add this check, and try to add a test for it also to be sure it works right.

eslickj:
While addressing @jsiirola's comments a new test found a problem in the degrees of freedom calculation, so I fixed that too.

PR #917, opened 2019-03-26
[WIP] LinearExpression fixed variable handling for GAMS writer

### LINKED ISSUE #924: GAMS writer fails to identify fixed variables in LinearExpression
The following two tests both fail because the GAMS writer is not able to properly handle the fixed variables in `LinearExpression`.

```
    def test_quicksum(self):
        m = ConcreteModel()
        m.y = Var(domain=Binary)
        m.c = Constraint(expr=quicksum([m.y, m.y], linear=True) == 1)
        m.y.fix(1)
        lbl = NumericLabeler('x')
        smap = SymbolMap(lbl)
        tc = StorageTreeChecker(m)
        self.assertEqual(expression_to_string(m.c.body, tc, smap=smap), "2")
```
expected: `1 + 1`; output: `x1 + x1`

```
    def test_quicksum_integer_var_fixed(self):
        m = ConcreteModel()
        m.x = Var()
        m.y = Var(domain=Binary)
        m.c = Constraint(expr=quicksum([m.y, m.y], linear=True) == 1)
        m.o = Objective(expr=m.x ** 2)
        m.y.fix(1)
        os = StringIO()
        m.write(os, format='gams')
        self.assertIn("USING nlp", os.getvalue())
```

I suspect that this is related in some way to #919, but I am still working to understand the GAMS visitor enough again.

@gseastream Any quick thoughts?

qtothec:
## Summary/Motivation:
**Update 3:**: Instead of changing the `to_string` of `LinearExpression`, this PR now adds in special string handling for `LinearExpression` in the GAMS writer expression tree visitor.

---

**Update 2:** I found out that the GAMS writer is not able to detect fixed variables inside of `LinearExpression` nodes due to similar issues as #919. This PR changes the string representation in `LinearExpression` to check if the variables are fixed and emit their fixed value if that is the case. It also adds two tests that verify the expected output.

Due to the changes from the previous update below, this PR requires merging of #921 in order pass tests.

Fixes #924 

---

**Update:** The original symptoms were found to be caused by a separate issue. This PR has therefore changed to simply add a small test for the GAMS writer.

---

*Original:*
GAMS writer will sometimes misidentify an NLP as MINLP if the model type change is due to the presence of fixed variables. This PR improves identification of such models and adds support for fixed variables.

### Legal Acknowledgement


qtothec:
@gseastream Want to take a quick glance?

gseastream:
The example model in the test you wrote generates the same gams file both on current pyomo master and on your PR code. Also, pending a test that highlights the behavior you described, I was pretty sure that fixed variables never made it anywhere into the labeler or the model file at all. I don't believe they get assigned a symbol, and in the gams file they are only represented by their constant value, not a symbol, so there shouldn't be a need for the code you added where you define fixed variables in the model file.

qtothec:
Thanks, Grant. I'll have to take a closer look. I was working with a larger model and seeing behavior in which variables were left in.

jsiirola:
@qtothec Does this mean that this is back to being a WIP PR for the time being?

qtothec:
@gseastream You were right. The error that I was observing was related to a separate issue, #919. The GAMS writer was working as intended. I'm changing this PR to simply add another test to the GAMS writer library, concerning nested Disjuncts with partial deactivation.

(pending changes to fix_disjuncts)

pyomo/core/expr/numeric_expr.py
```diff
@@ -1303,8 +1303,10 @@ def _to_string(self, values, verbose, smap, compute_values):
             tmp = [self.constant.to_string(compute_values=False)]
         if verbose:
             for c,v in zip(self.linear_coefs, self.linear_vars):
-                if smap:                        # TODO: coverage
+                if smap and not v.fixed:
                     v_ = smap.getSymbol(v)
+                elif v.fixed:
+                    v_ = str(v.value)
```
jsiirola: I think we need to re-think this.  This changes the string representation of *all* Pyomo LinearExpressions to replace Vars with their values when they are fixed.  We probably don't want to do this.  Unfortunately, I am not sure how to (easily) fix this without changing the `_to_string` API.

qtothec:
If this isn't the fix that we want, I can think of two other ways to obtain the desired behavior in GAMS:
1. not assume that fixed variables are automatically converted to constants in the GAMS writer, or
2. use an expression replacement visitor to map all fixed variables to constant values.

Which do you think would make the most sense?

jsiirola:
For the time being, I was thinking more along the lines of putting a trap for LinearExtession in the GAMS writer's visitor.  Longer term, I think we could/should revisit the entire to_string system for performance and extensibility (for example, we keep running into paren issues both here and in the baron writer).

pyomo/core/expr/numeric_expr.py
qtothec: I put a fix in the GAMS writer visitor instead.

qtothec:
Hm, this is causing some other issues.

pyomo/repn/plugins/gams_writer.py
```diff
@@ -78,6 +78,16 @@ def visit(self, node, values):
...
 
+        if node.__class__ is EXPR.LinearExpression:
+            argstr = []
+            for coef, var in zip(node.linear_coefs, node.linear_vars):
+                if var.fixed:
+                    argstr.append("{}".format(coef * var.value))
+                else:
+                    argstr.append("{} * {}".format(
+                        coef, var.to_string(verbose=False, smap=self.smap, compute_values=True)))
+            return " + ".join(argstr) + (" + {}".format(node.constant) if node.constant != 0 else "")
```
jsiirola, line 89: We need to wrap formatting of constant values in `ftoa()` to better handle output precision.

qtothec:
Closing in favor of #1463.

Issue #709, opened 2018-10-12
Quicksum fails to identify variables


qtothec:
```
from pyomo.environ import *
m = ConcreteModel()
m.x = Var()
expr = quicksum([m.x, m.x], linear=True)
from pyomo.core.expr.current import identify_variables
vars = list(identify_variables(expr, include_fixed=False))
print(vars)
```
Result:
```
[]
```

blnicho:
Fixed in #920

PR #920, opened 2019-03-26
Allow variable identification for LinearExpression

### LINKED ISSUE #919: LinearExpression incompatible with identify_variables()
`LinearExpression` always returns `expr.nargs() = 0` and has no `args`. This causes the walker for `identify_variables()` to see it as a leaf node. Unfortunately, it is not a leaf node.

This issue probably also causes the effect observed in #709.

qtothec:
## Summary/Motivation:
`identify_vars()` sees `LinearExpression` as a leaf node and does not properly yield its constituent variables. These changes allow proper behavior for `identify_vars()` without making changes to `LinearExpression`.

Fixes #919 

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/expr/expr_pyomo5.py
+++ b/pyomo/core/expr/expr_pyomo5.py
@@ -1249,6 +1249,19 @@ class _VariableVisitor(SimpleExpressionVisitor):
             self.seen.add(id(node))
             return node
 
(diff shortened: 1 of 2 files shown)


qtothec:
@blnicho It appears that it does likely fix #709, since the test code in that issue passes locally.

PR #921, opened 2019-03-27
Simplify the fix_disjuncts transformation

qtothec:
## Summary/Motivation:
The inability to have ComponentData-level granularity in reclassification makes effective use of specifying targets for this transformation difficult. As a result, use cases for this have not meaningfully materialized.
This PR simplifies the transformation to no longer accept the `targets` flag.

### Legal Acknowledgement

### DIFF
--- a/pyomo/gdp/plugins/fix_disjuncts.py
+++ b/pyomo/gdp/plugins/fix_disjuncts.py
@@ -1,119 +1,68 @@
 # -*- coding: UTF-8 -*-
-"""Transformation to fix disjuncts.
-
-This transformation looks for active disjunctions in the passed model instance
-and fixes the participating disjuncts based on their current indicator_var
-values. Active disjuncts are transformed to Block and inactive disjuncts
-(indicator_var = 0) are transformed to Block with all of their constituent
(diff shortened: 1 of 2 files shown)


jsiirola:
I am torn by this change.  Is it fixing a bug? That is, is there a known limitation or problem with the transformation that is being fixed by removing functionality?

Given that (eventually) we will fix the writers to not require variables be on active blocks, it seems to be a shame to lose this functionality.

ghackebeil:
> Given that (eventually) we will fix the writers to not require variables be on active blocks, it seems to be a shame to lose this functionality.

@jsiirola: Are you sure that this is an eventuality?

qtothec:
@jsiirola As it stands, the transformation is invalid for partial targeting of an indexed Disjunct. I have several of those models, and as a result I get very "interesting" interactions with solvers and other transformations as a result.

I don't think that the functionality would be difficult to add in properly once the variable ownership issue is resolved.

jsiirola:
@ghackebeil: good point.  In the discussions I have been in over the last couple months, the consensus appeared to be yes, but no formal PEP has been filed, discussed, or approved.

PR #927, opened 2019-04-03
Fixing bug in this_file_dir() of common/fileutils.py

carldlaird:
The function this_file_dir was returning location of "fileutils.py" instead of the location of the file that called this_file_dir (it was adding another call onto the stack which resulted in this_file using an incorrect stack entry). this_file now accepts a stack_offset which defaults to 1.

## Summary/Motivation:
- fixes a bug in this_file_dir (see description above).

-
-

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/tests/test_fileutils.py
+++ b/pyomo/common/tests/test_fileutils.py
@@ -24,11 +24,12 @@ from pyutilib.subprocess import run
 import pyomo.common.config as config
 from pyomo.common.log import LoggingIntercept
 from pyomo.common.fileutils import (
-    this_file, find_file, find_library, find_executable, ExecutableManager,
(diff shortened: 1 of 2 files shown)


pyomo/common/tests/test_fileutils.py
```diff
@@ -72,6 +73,11 @@ def test_this_file(self):
             'print(this_file())'
         )[1].strip(), '<stdin>')
 
+    def test_this_file_dir(self):
+        print(_this_file_dir)
```
blnicho: Is this a debugging line that should be removed?

carldlaird: Yes. fixed.

PR #936, opened 2019-04-12
Fix sets not iterating in correct order

### LINKED ISSUE #835: Non-deterministic behavior with Set (shouldn't the default be ordered)
I thought that pull request https://github.com/Pyomo/pyomo/pull/568 was going to make Sets be ordered by default. However, the following behavior illustrates that this is not the case.

Please see the following code:
```
from pyomo.environ import *
a = ["a", "x", "c", "b"]

model = ConcreteModel()
model.S = Set(initialize=a)
model.OS = Set(initialize=a, ordered=True)

model.S.pprint()
model.OS.pprint()

print('set not ordered: {}'.format([i for i in model.S]))
print('set ordered: {}'.format([i for i in model.OS]))
```

The first time I run this, I get:
```
S : Dim=0, Dimen=1, Size=4, Domain=None, Ordered=False, Bounds=None
    ['a', 'b', 'c', 'x']
OS : Dim=0, Dimen=1, Size=4, Domain=None, Ordered=Insertion, Bounds=None
    ['a', 'x', 'c', 'b']
set not ordered: ['c', 'a', 'x', 'b']
set ordered: ['a', 'x', 'c', 'b']
```

and the second time, I get:
```
S : Dim=0, Dimen=1, Size=4, Domain=None, Ordered=False, Bounds=None
    ['a', 'b', 'c', 'x']
OS : Dim=0, Dimen=1, Size=4, Domain=None, Ordered=Insertion, Bounds=None
    ['a', 'x', 'c', 'b']
set not ordered: ['c', 'a', 'b', 'x']
set ordered: ['a', 'x', 'c', 'b']
```

To me, this implies the following:
1) ordered does not appear to be the default (see the Set pprint)
2) If you mark a set as ordered, then the order in the set appears to be the same as the initialize list and the order is maintained during iteration.
3) If you do not mark a set as ordered, then the order in the set is not the same as the initialize list (although it appears to be consistent), but the order during iteration is not consistent (and does not match the order in the Set).

jsiirola:
## Summary/Motivation:
This resolves issues where iterating over implicitly ordered sets does not honor the set order.  This standardizes the list() of set members as the `value_list` attribute.  

- standardize on the use of `value_list` as the attribute that stores the ordered set data.

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/sets.py
+++ b/pyomo/core/base/sets.py
@@ -121,7 +121,7 @@ def _value_sorter(self, obj):
 
     """
     if self.ordered:
-        return obj.value
+        return obj.value_list
     else:
         return sorted_robust(obj)
 
@@ -279,7 +279,7 @@ class _OrderedSetData(_SetDataBase):
                                 if the discard method is used.
     """
 
-    __slots__ = ('value', 'order_dict', '_bounds', '_is_sorted')
+    __slots__ = ('value', 'value_list', 'order_dict', '_bounds', '_is_sorted')
 
     def __init__(self, owner, bounds):
         #
@@ -323,7 +323,7 @@ class _OrderedSetData(_SetDataBase):
         'value' attribute.  The underlying set values are not be stored
         as a Python set() object.
         """
-        return set(self.value)
+        return self.value
 
     def _sort(self):
         """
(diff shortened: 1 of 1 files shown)


ZedongPeng:
In the latest version of Pyomo, it seems that `value_list` is no longer supported. Error is reported.
`AttributeError: 'OrderedSimpleSet' object has no attribute 'value_list'`
Is there anything else that serves as the same function of `value_list`. 
Thanks. @jsiirola @michaelbynum

PR #943, opened 2019-04-17
Deprecate pyomo.bilevel

whart222:
## Summary/Motivation:
Deprecate the use of pyomo.bilevel, which has known errors

- This generates a deprecation warning when a pyomo.bilevel.SubModel component is created.
- Note that I'd prefer to simply create a deprecation warning when pyomo.bilevel is imported. But doing so creates a warning whenever pyomo.environ is imported, which is undesirable. If we edit pyomo.environ to omit pyomo.bilevel in its initialization of plugins, then pyomo.bilevel tests fail. That might be an OK alternative, especially if we disable those tests. Is that preferable???

### Legal Acknowledgement


carldlaird:
To answer your question in the PR - I prefer to deprecate as you have - including the deprecation during Pyomo import would impact all users - and many will not be familiar with bilevel.

pyomo/bilevel/components.py
```diff
@@ -13,13 +13,17 @@
...
 
 # TODO: Do we need to have SimpleSubModel and IndexedSubModel classes?
 
 
 @ModelComponentFactory.register("A submodel in a bilevel program")
 class SubModel(SimpleBlock):
 
+    @deprecated("Use of the pyomo.bilevel package is deprecated. Development of bilevel \
+optimization capabilities has been shifted to the Pyomo Adversarial Optimization (PAO) library. \
+Please contact William Hart for further details (wehart@sandia.gov).")
```
jsiirola: I believe that backslashes for line continuation within strings is not recommended.  Can we change this to either:
```python
    @deprecated("Use of the pyomo.bilevel package is deprecated. Development of bilevel "
                "optimization capabilities has been shifted to the Pyomo Adversarial "
                "Optimization (PAO) library. Please contact William Hart for further "
                "details (wehart@sandia.gov).")
```
or 
```python
    @deprecated("""Use of the pyomo.bilevel package is deprecated. Development of bilevel
optimization capabilities has been shifted to the Pyomo Adversarial Optimization (PAO) library.
Please contact William Hart for further details (wehart@sandia.gov).""")
```

carldlaird:
I don't have any concerns about deprecating. In my opinion, deprecation is exactly the right approach for api changes that were documented in the book. It provides an indicator to the user that this functionality is not being developed further and will eventually be replaced, giving them time to find a suitable replacement.

But I agree, we should move this to a PEP.

whart222:
See #950 for the PEP associated with this PR.

blnicho:
This was discussed at the PMC meeting and the decision was to move ahead with deprecating pyomo.bilevel and to expand this PR to include deprecation of pyomo.duality.

PR #946, opened 2019-04-18
Add warm start capabilities to CBC solver plugin.

### LINKED ISSUE #881: Passing initial solution to CBC solver
Hello!
While working on a personal project I needed to pass an initial solution for a MIP problem (for example a solution generated with an heuristic) to the CBC solver. CBC supports that with the 'mipstart' option (passing a soln file with the variables' values). I couldn't find support for this in pyomo. Is there a way to achieve that?
I implemented a workaround:
I'd first set up the initial values in the concrete instance and then call this function:
```python
def write_solution(instance):
    with open('solution.soln', 'w') as solution:
        for v in instance.component_objects(Var, active=True):
            for index in v:
                if str(v[index].domain) == 'Binary' and v[index].value:
                    trans = str.maketrans('[,]', '(_)')
                    print(
                        '0 {} {}'.format(
                            v[index].name.translate(trans), v[index].value
                        ),
                        file=solution
                    )
```
If you think this could be useful, I could clean it up a little bit and submit a PR, what do you think?

lapseofreason:
## Summary/Motivation:

Unlike other solver plugins, the plugin for CBC lacks warm start capabilities. This pull request adds this functionality.

It is based on similar functionality for the GUROBI and CPLEX solver plugins.

- Add functionality to write SOLN files from a model instance.
- Add support for warmstart parameter to CBC plugin.

## Comments:

- Writing SOLN files might be of interest for different use cases and could later be moved somewhere else.
- This fixes #881.

### Legal Acknowledgement


lapseofreason:
Thank you for the review. I was trying to stick to the original style and the style from the other plugins, but I agree to your suggestions, I think they are better / more common style.

pyomo/solvers/plugins/solvers/CBCplugin.py
```diff
@@ -178,10 +189,98 @@ def _default_results_format(self, prob_format):
...
+
+        if isinstance(instance, IBlock):
+            smap = getattr(instance, "._symbol_maps")[self._smap_id]
+        else:
+            smap = instance.solutions.symbol_map[self._smap_id]
+        byObject = smap.byObject
+
+        with open(filename, 'w') as solnfile:
+            for var in instance.component_data_objects(Var):
+                if var.value and (id(var) in byObject):
```
jsiirola: This appears to only write out nonzero values. Is that sufficient for CBC to accept a warmstart?

lapseofreason: Yes, in fact the CBC documentation for mipstart explicitly states that only non-zero integer/boolean variables are considered. I made some changes to clarify this and not write continuous variables either.

lapseofreason:
About the test: Where do you think the test should go? Can you point me towards any similar existing tests (CPLEX/GUROBI maybe) as examples / to extend?

I could write a test that checks for one of those lines in the cbc log, I'm just not sure where to put it:

```
opening mipstart file /tmp/tmpADj1trP.cbc.soln.
MIPStart values read for 2 variables.
...
Cbc0045I MIPStart provided solution with cost 123
```

jsiirola:
@lapseofreason: our solver interface tests are not the easiest to follow (and could stand to undergo a significant overhaul).  I would suggest creating a `test_cbc.pt` in `pyomo/solvers/tests/checks`.  You can follow some of the logic in the `GAMSLogfileGmsTests.test_tee()` test in `test_GAMS.py`.  In particular, the use of `with capture_output() as output:` to capture the solver output into a `StringIO` buffer that you can then interrogate.

ghackebeil:
@lapseofreason: Would you mind making the following additional change on this branch? I recently noticed the CBC developers added a fix on trunk for a longstanding issue that we have an expected failure for. This will avoid a failing test when we start testing the next CBC release (after 2.10.1).

Change [this](https://github.com/lapseofreason/pyomo/blob/a966de896804a4a1a5826513b21cfd09151361af/pyomo/solvers/tests/testcases.py#L81) from
```python
(lambda v: v <= _trunk_version,
```
to
```python
(lambda v: v <= (2,10,1,0)
```

lapseofreason:
@jsiirola Thanks for the pointers, I now added the test.

@ghackebeil I made the change, although it is unrelated to the warm start functionality.

I had to make some corrections because I looked at the wrong base class for variables. There seems to be some inconsistency between pyomo.core.kernel.variable and pyomo.core.base.var.Var, because the former considers IntegerSet as is_binary, whereas the latter doesn't.

PR #954, opened 2019-04-18
Fbbt updates

michaelbynum:
## Summary/Motivation:
1) This improves FBBT for powers.
2) This includes a bug fix for symbolic differentiation.

### Legal Acknowledgement


qtothec:
I see a lot of removal of `update_variable_bounds`. Is there a change in the defaults?

michaelbynum:
@qtothec I removed the ```update_variable_bounds``` option in favor of using the BoundsManager.

michaelbynum:
@qtothec Is this problematic? Were you using this feature? I thought you had already switch to the ```BoundsManager```.

qtothec:
@michaelbynum Ah, you're right. We had switched to using the `BoundsManager`. Was worried for a moment. At some point I'll need to go through and understand how the BoundsManager works.

qtothec:
Are you testing for some of the more exotic cases, e.g. fractional power not fixed, etc.?

michaelbynum:
@qtothec Yes. There are at least some tests for this. See ```test_pow1``` and ```test_pow2``` in test_fbbt.py. It is certainly possible that there are edge cases I did not think of though.

pyomo/contrib/fbbt/interval.py
```diff
@@ -34,7 +38,11 @@ def power(xl, xu, yl, yu):
         elif yu <= 0:
             lb = xu ** yl
             ub = xl ** yu
+        else:
+            raise RuntimeError('This was unexpected.')
```
jsiirola: I would replace this with a DeveloperError -- that is, an exception that you would never expect to be able to be hit, and if someone does, they should report it to the developers.

pyomo/contrib/fbbt/fbbt.py
```diff
@@ -912,11 +899,17 @@ def fbbt_block(m, tol=1e-4, deactivate_satisfied_constraints=False, update_varia
                 var_ubs[v] = value(v.ub)
             var_to_con_map[v].append(c)
 
+    for _v in m.component_data_objects(ctype=Var, active=True, descend_into=True, sort=True):
+        if _v.is_fixed():
+            _v.setlb(_v.value)
+            _v.setub(_v.value)
```
jsiirola: Why are you changing the actual variable bounds here?

pyomo/contrib/fbbt/interval.py
```diff
@@ -92,13 +101,111 @@ def power(xl, xu, yl, yu):
...
+                        _xl = -xu
+                    else:
+                        _xl = xl
+                    if orig_xu < xl:
+                        _xu = -xl
+                    else:
+                        _xu = xu
+                xl = _xl
+                xu = _xu
+        elif y % 2 == 1:
```
jsiirola: This should probably be:
```
else: # y % 2 == 1
```
(so folks digging through the code understand that there aren't unhandled cases in this if tree)

pyomo/contrib/fbbt/interval.py
```diff
@@ -92,13 +101,111 @@ def power(xl, xu, yl, yu):
...
+    """
+    z = x**y => compute bounds on x
+    x = exp(ln(z) / y)
+    """
+    xl, xu = log(zl, zu)
+    xl, xu = div(xl, xu, yl, yu)
+    xl, xu = exp(xl, xu)
+
+    # if y is an integer, then x can be negative
+    if yl == yu and yl == round(yl):
```
jsiirola: This is a pretty complicated if tree .. could you provide (terse, inline) comments giving hints at the conditions that each branch is trying to cover?

pyomo/contrib/fbbt/fbbt.py
michaelbynum: @jsiirola This is just an implementation detail. First, the fbbt functions are expected to change variable bounds. Second, suppose we have a constraint such as ```x + y == 1``` and ```x``` is fixed to ```1``` but does not have any bounds. In this case, we should be able to infer that ```y``` is ```0```. However, because of the way I have implemented fbbt, no bounds on y will be inferred unless the bounds on x get updated. It is certainly possible to do this correctly without changing the bounds on fixed variables, but I don't see a good reason not to change the bounds on fixed variables. If you can think of a reason, please let me know.

pyomo/contrib/fbbt/interval.py
michaelbynum: Good suggestion. I will fix this.

pyomo/contrib/fbbt/interval.py
michaelbynum: Good suggestion. I will fix this.

pyomo/contrib/fbbt/interval.py
michaelbynum: Yes. I meant to add some docs but forgot.

qtothec:
Looks good, Michael. This might be a separate exercise, but it would be good to have a test for the BoundsManager.

michaelbynum:
@qtothec Good point.

PR #955, opened 2019-04-19
Model viewer install/testing

eslickj:
This PR dose a few things
1. Makes the model viewer work on AbstractModels that haven't been constructed.
2. Adds a bit to setup.py to also install ui files.
3. Adds a pyomo subcommand to launch viewer (```pyomo model-viewer```)
4. A start on running tests for qt parts of the model viewer.  Getting these to work in the automated tests may take some doing.

### Legal Acknowledgement


eslickj:
The added tests need to be run with pytest, so I just skipped them for now.  We may be able to sort it out later.  The other thing in this PR is that the model viewer won't break down for AbstractModels that haven't been constructed.

eslickj:
I'm going to continue working here, but it looks like the only PyQt testing framework I could find that seems to be actively maintained is pytest-qt.  This requires that the test be run with pytest.  I think I can write a test using the pyomo's standard unit test framework and running a pytest subprocess including a timeout to prevent any rouge UI tests from causing tests to hang.

@jsiirola is investigating making pytest and PyQt available on some test platforms, so we'll either work out how to run the tests here or put it on hold if it proves to difficult, in which case I'll add and issue for viewer testing.

jsiirola:
I have PyQt4 available on the Jenkins environment.  pytest and pytest-qt are in the queue to be distributed (hopefully will show up tonight).  Compiling PyQt is challenging, so it may be a while before we get it on Jenkins.

I will start working to get these installed into the Travis docker images as well.

eslickj:
@jsiirola, I'm probably missing something here, but why do you need to compile PyQt? I think now days, you can get it precompiled with pip.

eslickj:
Changed back to WIP.  I'll try to write the test wrapper before finishing this off.

eslickj:
@jsiirola I added a test wrapper that will run the pytests when you run nosetests.  I suspect it needs some work, but that basic idea is all there.

blnicho:
@eslickj is this still a WIP?

eslickj:
I'm going to say this isn't a work in progress anymore and its ready to go.  The tests pass, but I don't think the ones that require pyqt are running, from what I can see of the testing output.  My guess would be that even though it appears that pyqt is installed there is some problem importing it.  There are a few things that can go wrong there, and I can't tell what's happening.  The tests do seem to work okay locally.

The critical things to check here are:
1) The changes to setup.py to add ```pyomo model-viewer``` command.
2) pyqt tests that run pytest as a subprocess.

ghackebeil:
My vote would be that we stop adding new scripts that get placed in the user bin directory, and start adding to sub-commands to the pyomo command.

eslickj:
@ghackebeil, that was my intention, but it is possible I didn't execute right.  I does work as expected, but I may have added something extra that I didn't need.

blnicho:
I agree with @ghackebeil but I don't know enough about the pyomo script infrastructure to check if this PR is creating a new script or adding a sub-command. @ghackebeil or @jsiirola could you check if this is adding a sub-command to the `pyomo` script? If it isn't, could you offer some guidance on how it should be implemented? Maybe the recently added `pyomo download-extensions` command would be a good example to follow?

eslickj:
@blnicho, I wasn't sure either so I just checked it again.  The subcommand works, and I do not see any additional scripts being created.

ghackebeil:
I can confirm @eslickj comment. This works as expected.

blnicho:
@ghackebeil thanks for checking that. I think this is ready to merge, just waiting on a second review.

PR #940, opened 2019-04-16
Adding a general infrastructure for building extensions

jsiirola:
## Summary/Motivation:
This provides a general approach for extensions (primarily in contrib) to register functions for building compiled extensions, and makes that functionality available through the `pyomo build-extensions` command.

## Changes proposed in this PR:
- refactor `common.fileutils` to add a `Library` singleton and a `find_dir()` method
- adding a utilite for downloading and extracting ZIP archives
- Adding a build harness for `contrib.mcpp` based on distutils

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #958, opened 2019-04-22
Return the subcommand error code from the pyomo script

jsiirola:
## Summary/Motivation:
The `pyomo` script always returns a return code of 0 (indicating no failure), even if the underlying subcommand encountered an error.  This PR propagates the subcommand error code out to the calling shell so that those who script around pyomo can more easily detect if the pyomo script encountered an error.

- Return the underlying subcommand `errorcode` value from the pyomo script to the calling process
- Eliminate the `get_result` flag from `pyomo_main.main()`.  

### Legal Acknowledgement


jsiirola:
This PR also resolves several diffs that crept into the Pyomo Book tests through a previous PR (#940)

PR #963, opened 2019-04-23
Preliminary multisolve implementation

qtothec:
## Summary/Motivation:
"Simple" parallel multisolve capability using multiple processes on the same node.
If there's a better way of going about this, please do let me know.

### Legal Acknowledgement


blnicho:
@qtothec could you provide more context on how this compares to the multistart solver already in contrib, how this is different, and why you decided not to integrate them?

qtothec:
@blnicho Multistart is trying with different initial starting points. Multisolve is trying the same starting point with different solvers, and taking the first or best solution obtained.

I didn't integrate them because I was working towards a minimum functional product that I could start using. I think it does make sense to fold them into each other at some point, since multistart is often parallelized as well.

qtothec:
This is waiting on pathos availability in the build harness. See #1122.

pyomo/contrib/multisolve/multisolve.py
```diff
@@ -0,0 +1,146 @@
...
+        config.set_value(kwds)
+
+        if not _pathos_available:
+            raise ImportError("Multisolve requires the `pathos` package.")
+
+        solve_data = GDPoptSolveData()
+        with create_utility_block(model, 'multisolve_utils', solve_data):
+            results = []
+
+            with Pool(processes=2, maxtasksperchild=1) as pool:
```
jsiirola: It looks like the use of a multiprocessing pool as a context manager is only supported in Python 3.  I am not sure why you need to use it in a `with` clause, either.  It seems if you just replace this with
```
pool = Pool(processes=2, maxtasksperchild=1)
```
then things work in Python 2.7 and 3.x.

pyomo/contrib/multisolve/multisolve.py
```diff
@@ -0,0 +1,146 @@
...
+import logging
+import time
+import timeit
+
+_pathos_available = True
+try:
+    # Pathos is required because a dill-friendly multiprocessing alternative was necessary.
+    from pathos.pools import _ProcessPool as Pool
+except ImportError:
+    _pathos_available = False
```
jsiirola: Why do you need pathos?  It looks like the only thing you are using from pathos is `pathos.pools._ProcessPool`, and that class isn't actually from pathos:
```python
>>> import pathos
>>> pathos.pools._ProcessPool
<class 'multiprocess.pool.Pool'>
```
If you dig into the pathos code, `_ProcessPool` is from multiprocessing if multiprocessing is available (since 2.6), otherwise it falls back on the older `processing` package.

qtothec: I thought that pathos was a reimplementation of the multiprocess pool using `dill` instead of `pickle`, but I could be wrong.

pyomo/contrib/multisolve/multisolve.py
qtothec: Makes sense. I think the context manager might automatically `terminate` the pool at the end? I can't remember exactly.

jsiirola: Fair enough.  If there is something that happens on `__exit__`, then we should be able to implement it in a portable way with an explicit:
```python
try:
    # use the pool
finally:
   # terminate the pool
```

pyomo/contrib/multisolve/multisolve.py
qtothec: Locally, at least, replacing the pathos import with a `Pool` import from `multiprocessing` causes a failure (I'm running cpython 3.6 on Ubuntu).

jsiirola:
FYI: The Jenkins test failure is the following (observed intermittently on python3.6 and python3.7, but always with the same test):
```
03:44:17 test_8PP (pyomo.contrib.multisolve.test_multi.MultistartTests) ... Build was aborted
05:43:21 [TASKS] Skipping publisher since build result is ABORTED
```
...indicating some form lock...

qtothec:
Closing to not pollute the testing infrastructure while I revisit the system.

PR #966, opened 2019-04-24
Fixing expression infix representation

### LINKED ISSUE #902: BARON writer parentheses issue (arithmetic operator after unary operator) causing "NonLinearity Error in POW expression" error
I believe that one reason for causing the "NonLinearity Error in POW expression" error message is the omission of (for BARON necessary) parentheses by sympy/pyomo. If e.g. a constraint in pyomo is defined the following way:
model.con = Constraint(expr=model.x1 - 60*model.x2**(-1) * 50*model.x3 == 0)
the constraint's expression will be written (by baron_writer.py):
con: x1 - 60*x2^-1 * 50*x3 == 0;
The expression, however, is (unitentionally) misinterpreted by BARON the following way:
x1 - 60*x2^(-1 * 50*x3) == 0; (see attached screenshot of BARON manual)
page in baron manual on arithmetic operators in exponents

In this simplified case it is of course easy to fix by defining the constraint differently. However, if constraint expressions are generated by e.g. pyomo.core.base.symbolic.differentiate(), I believe the best option to fix this problem, is in constraint section of the baron_writer.py

### LINKED ISSUE #916: Incorrect string representation of nested powers
When I assessed an expression containing nested powers from a pyomo model pprint() I was confused to see that it apparently did not uphold right associativity that is assumed by python, thaty is, a ** b ** c should be a ** (b ** c).

```python
from pyomo.environ import *
m = ConcreteModel()
m.x = Var(initialize=2)
x = m.x
exprs =    [2 ** x ** 3, (2 ** x) ** 3, 2 ** (x ** 3)]
expected = [2 ** 2 ** 3, (2 ** 2) ** 3, 2 ** (2 ** 3)]

# This gives the expected numerical results, but the string representation of all expr is identical
for expr, expe in zip(exprs, expected):
    print(f'{expr} gives {expr()}, should be {expe}')
```
gives me
```text
2**x**3 gives 256, should be 256
2**x**3 gives 64, should be 64
2**x**3 gives 256, should be 256
```

So it turned out that the operator precedence acually agrees with python but the string representation is wrong:

```python
# Here's the expected output:
def expect(expr):
    try:
        expr.args
        return f'({expr})'
    except AttributeError:
        return str(expr)
for expr in exprs:
    base, exponent = expr.args
    print(f'{expr}, should be printed as {expect(base)} ** {expect(exponent)}')
```

gives me
```text
2**x**3, should be printed as 2 ** (x**3)
2**x**3, should be printed as (2**x) ** 3
2**x**3, should be printed as 2 ** (x**3)
```

---

- pyomo '5.6.1' 
- python '3.6.7 |Anaconda, Inc.| (default, Dec 10 2018, 20:35:02) [MSC v.1915 64 bit (AMD64)]'

jsiirola:
## Summary/Motivation:
Pyomo expressions currently generate invalid infix string representations of expression in certain circumstances, notably when subexpressions appear on the RHS of a binary operation with the same precedence as the binary operator.  The "simple" solution is to add an "associativity" flag to expressions so that the string generator can know when an expression tree would violate associativity rules.  This PR also improves the infix string generation in the GAMS and BARON writers by adding the associativity rules and a more robust treatment of tokens beginning with a `-` operator.

- Adding an `_associativity()` method for expressions so that we can correctly wrap complex expressions in parens (fixes #916)
- More consistently parenthesize tokens beginning with "`-`" (fixes #902)
- Make BARON writer error on unrecognized components/expression objects

### Legal Acknowledgement


pyomo/core/expr/numeric_expr.py
```diff
@@ -263,6 +263,17 @@ def to_string(self, verbose=None, labeler=None, smap=None, compute_values=False)
     def _precedence(self):
         return ExpressionBase.PRECEDENCE
 
+    def _associativity(self):
+        """Return the associativity of this operator.
+
+        Returns 1 if this operator is left-to-right associative or -1 if
+        it is right-to-left associative.  ANy other return value will be
```
blnicho: Typo: ANy

pyomo/core/expr/numeric_expr.py
```diff
@@ -263,6 +263,17 @@ def to_string(self, verbose=None, labeler=None, smap=None, compute_values=False)
...
+        interpresed as "not associative" (implying any arguments that
```
blnicho: Typo: interpreted

PR #970, opened 2019-04-26
Standardize preliminary time limit support across MindtPy and GDPopt

RomeoV:
When solving models, now a `time_limit=[secs]` argument can be passed to
the solve method, signaling that the algorithm should terminate after that time.

## Fixes
No issue assigned

## Summary/Motivation:
Most solvers support a time limit. This was missing so far from the **mindtpy** and **gdpopt** solvers, although timing capability was already implemented. This PR adds a new config block `time_limit` and an appropriate abort condition in the existing methods `algorithm_should_terminate` that already look for an iteration limit. ~Note that the name `timelimit` was chosen in accordance to other solvers.~

- modify `time_code` method (in gdpopt util) to store start time
- add `time_limit` block to **MindtPy** and **gdpopt**
- add abort condition for time right after abort condition for iterations

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/gdpopt/GDPopt.py
+++ b/pyomo/contrib/gdpopt/GDPopt.py
@@ -65,7 +65,7 @@ class GDPoptSolver(object):
         description="Iteration limit."
     ))
     CONFIG.declare("time_limit", ConfigValue(
-        default=10,
+        default=600,
         domain=PositiveInt,
         description="Time limit (seconds)",
         doc="Seconds allowed until terminated"
--- a/pyomo/contrib/gdpopt/GDPopt.py
+++ b/pyomo/contrib/gdpopt/GDPopt.py
@@ -68,9 +68,9 @@ class GDPoptSolver(object):
         default=600,
         domain=PositiveInt,
         description="Time limit (seconds)",
-        doc="Seconds allowed until terminated. Note that currently only the"
-            "master problem can be checked and terminated, so you might need"
-            "to set a subsolver timelimit too."
+        doc="Seconds allowed until terminated. Note that the time limit can"
+            "currently only be enforced between subsolver invocations. You may"
+            "need to set subsolver time limits as well."
     ))
     CONFIG.declare("strategy", ConfigValue(
         default="LOA", domain=In(["LOA", "GLOA"]),
--- a/pyomo/contrib/gdpopt/GDPopt.py
+++ b/pyomo/contrib/gdpopt/GDPopt.py
@@ -68,7 +68,9 @@ class GDPoptSolver(object):
         default=600,
         domain=PositiveInt,
         description="Time limit (seconds)",
-        doc="Seconds allowed until terminated"
+        doc="Seconds allowed until terminated. Note that currently only the"
+            "master problem can be checked and terminated, so you might need"
+            "to set a subsolver timelimit too."
     ))
     CONFIG.declare("strategy", ConfigValue(
         default="LOA", domain=In(["LOA", "GLOA"]),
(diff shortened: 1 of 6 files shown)


qtothec:
Huh. Some of the runs are getting a different objective value for the constrained layout problem. I'll have to check to see if that's due to hitting an iteration limit.

pyomo/contrib/gdpopt/GDPopt.py
```diff
@@ -64,6 +64,12 @@ class GDPoptSolver(object):
         default=30, domain=NonNegativeInt,
         description="Iteration limit."
     ))
+    CONFIG.declare("time_limit", ConfigValue(
+        default=600,
+        domain=PositiveInt,
+        description="Time limit (seconds)",
+        doc="Seconds allowed until terminated"
```
qtothec: We should note that this time limit is "loose". That is, if we get stuck solving a subproblem for a long time, we can exceed the time limit.

pyomo/contrib/gdpopt/iterate.py
```diff
@@ -89,6 +90,18 @@ def algorithm_should_terminate(solve_data, config):
         solve_data.results.solver.termination_condition = tc.maxIterations
         return True
 
+    # Check time limit
+    total_time_elapsed = default_timer() - solve_data.timing.total.start
+    if total_time_elapsed > config.time_limit:
+        config.logger.info(
+            'MindtPy unable to converge bounds '
```
blnicho: Is the solver used here always going to be MindtPy?

RomeoV: Good catch! I will fix the message.

RomeoV: Fixed in f0a20454e3f4a83e61d8cebec060662def0debc9

pyomo/contrib/gdpopt/GDPopt.py
RomeoV: Fixed in b3d04a30011ee833153708cf87cee68f09a96b73

pyomo/contrib/gdpopt/GDPopt.py
```diff
@@ -68,7 +68,9 @@ class GDPoptSolver(object):
         default=600,
         domain=PositiveInt,
         description="Time limit (seconds)",
-        doc="Seconds allowed until terminated"
+        doc="Seconds allowed until terminated. Note that currently only the"
+            "master problem can be checked and terminated, so you might need"
+            "to set a subsolver timelimit too."
```
qtothec: I would prefer the language below, because if the algorithm stalls on the MILP master, we would still have the same issue. (That is, unclear what "master" refers to):

> Seconds allowed until terminated. Note that the time limit can currently only be enforced between subsolver invocations. You may need to set subsolver time limits as well.

RomeoV: Fixed in 1aa6815f9d10ba88f7de5ca36cd1bc101b95d0ed

pyomo/contrib/mindtpy/MindtPy.py
```diff
@@ -363,6 +372,14 @@ def solve(self, model, **kwds):
             solve_data.results.problem.lower_bound = solve_data.LB
             solve_data.results.problem.upper_bound = solve_data.UB
 
+        solve_data.results.solver.timing = solve_data.timing
+        solve_data.results.solver.user_time = solve_data.timing.total
+        solve_data.results.solver.wallclock_time = solve_data.timing.total
+
+        solve_data.results.solver.iterations = solve_data.mip_iter  # TODO @David this makes sense right?
```
qtothec: Yeah, we usually report the master iterations. You'll want to double-check when the mip_iter advancement happens to make sure that `mip_iter += 1` doesn't happen at an inconvenient location; otherwise, you'll always report one more iteration than it actually required.

RomeoV: Fixed in 0afb2f58e8f91bd681aeeb5213bcc069392e80a7

qtothec:
This looks good to me, pending test results.

PR #971, opened 2019-04-26
Mosek

harunkivril:
## Fixes # .
mosek_direct plugin is inserted inside the solver plugins.
__init__ is updated to import mosek_direct.
A test script for mosek_direct is added to solver/tests/check.
mosek_direct is registered inside the testcases.py.

## Summary/Motivation:
Some Pyomo users have approached Mosek asking for a Pyomo interface to the Mosek convex solver (www.mosek.com). This PR introduces a direct interface to Mosek.

-Making Mosek available in Pyomo. The interface supports LP, QP, QCQP and integer variables. Note that Mosek will only solve QP and QCQP problems as long as they are convex.

### Legal Acknowledgement

### DIFF
--- a/pyomo/solvers/plugins/solvers/mosek_direct.py
+++ b/pyomo/solvers/plugins/solvers/mosek_direct.py
@@ -483,7 +483,7 @@ class MosekDirect(DirectSolver):
                                 msk.rescode.trm_mio_num_branches,
                                 msk.rescode.trm_num_max_num_int_solutions]:
             self.results.solver.status = SolverStatus.ok
-            self.results.solver.termination_message = "Optimization terminated because maximum number if relaxations" \
+            self.results.solver.termination_message = "Optimization terminated because maximum number of relaxations" \
                 " / branches / integer solutions exceeded " \
                 "the value specified in the TimeLimit parameter."
             self.results.solver.termination_condition = TerminationCondition.maxEvaluations
--- a/pyomo/solvers/plugins/solvers/mosek_direct.py
+++ b/pyomo/solvers/plugins/solvers/mosek_direct.py
@@ -165,7 +165,7 @@ class MosekDirect(DirectSolver):
             x, y = v
             qsubj.append(self._pyomo_var_to_solver_var_map[x])
             qsubi.append(self._pyomo_var_to_solver_var_map[y])
-            qval.append(repn.quadratic_coefs[i]*2)
+            qval.append(repn.quadratic_coefs[i]*((qsubi==qsubj)+1))
             referenced_vars.add(x)
             referenced_vars.add(y)
         new_expr.extend([qval, qsubi, qsubj])
@@ -233,20 +233,6 @@ class MosekDirect(DirectSolver):
 
         self._add_block(model)
 
-        for var, n_ref in self._referenced_variables.items():
-            if n_ref != 0:
-                if var.fixed:
-                    if not self._output_fixed_variable_bounds:
-                        raise ValueError(
-                            "Encountered a fixed variable (%s) inside "
-                            "an active objective or constraint "
-                            "expression on model %s, which is usually "
-                            "indicative of a preprocessing error. Use "
-                            "the IO-option 'output_fixed_variable_bounds=True' "
-                            "to suppress this error and fix the variable "
-                            "by overwriting its bounds in the Gurobi instance."
-                            % (var.name, self._pyomo_model.name,))
-
     def _add_block(self, block):
         DirectOrPersistentSolver._add_block(self, block)
 
@@ -293,8 +279,6 @@ class MosekDirect(DirectSolver):
         self._vars_referenced_by_con[con] = referenced_vars
         self._pyomo_con_to_solver_con_map[con] = con_index
         self._solver_con_to_pyomo_con_map[con_index] = con
-
-    # def _add_sos_constraint(self, con):
 
     def _mosek_vtype_from_var(self, var):
         """
@@ -405,8 +389,6 @@ class MosekDirect(DirectSolver):
         msk = self._mosek
 
         itr_soltypes = [msk.problemtype.qo, msk.problemtype.qcqo]
-        if self._version_major < 9:
-            itr_soltypes.append(msk.problemtype.geco)
 
         if (msk_task.getnumintvar() >= 1):
             self._whichsol = msk.soltype.itg
@@ -509,7 +491,7 @@ class MosekDirect(DirectSolver):
 
         else:
             self.results.solver.termination_message = " Optimization terminated %s response code." \
-                "Check Mosek response code documentetion for further explanation." % self._termcode
+                "Check Mosek response code documentation for further explanation." % self._termcode
             self.results.solver.termination_condition = TerminationCondition.unknown
 
         if SOLSTA_MAP[sol_status] == 'unknown':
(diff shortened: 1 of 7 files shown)


ghackebeil:
@harunkivril: Thank you for putting this together!

I've tested this branch locally, and I can confirm that, with a few minor fixes, this thing is pretty much ready to merge. I'll comment on the issues shortly.

As for the testing, it's likely that we don't have Mosek installed on any of the servers where our tests are run. I _could_ add it to the docker images that are used by our Travis CI tests, but that may not be appropriate since those images are public and Mosek requires a license (I am currently using an acedemic license to test locally). It would probably be better to work out some sort of testing license situation with the Sandia folks, so that Mosek tests can be run on one of their private Jenkins test servers.

That probably doesn't need to hold up this PR though.

ghackebeil:
For consistency with the other solvers, I think it would make the most sense for users to be able to instantiate this solver using
```python
SolverFactory('mosek')
# or
SolverFactory('mosek', solver_io='python')
```
which should both result in the same interface to Mosek (because there is only one that we are including). If you agree, all instances of "mosek_direct" should be changed to "mosek".

ghackebeil:
With the correction to the test case declaration in `pyomo/solvers/tests/solvers.py`, running the test suit in `pyomo/solvers/tests/checks/test_writers.py` results in four minor failures (plus some others that I'll mention next). The minor failures are that the dual solution returned by Mosek for one of the test QPs is off from the baseline by just above the diff tolerance.

I think its okay to just relax the diff tolerance a bit more. You can do this by changing `self.diff_tol = 1e-4` in `pyomo/solvers/tests/models/base.py` to `1e-3`.

ghackebeil:
The other failures I see in that suite have to do with Mosek not seeing some of the test cases as convex. For instance,
```python
import pyomo.kernel as pmo
model = pmo.block()
model.x = pmo.variable(domain=Binary)
model.y = pmo.variable(domain=Binary)
model.z = pmo.variable(domain=Binary)

model.obj = pmo.objective(model.x,sense=maximize)
model.c0 = pmo.constraint(model.x+model.y+model.z == 1)
model.qc0 = pmo.constraint(model.x**2 + model.y**2 <= model.z**2)
model.qc1 = pmo.constraint(model.x**2 <= model.y*model.z)
```
I would have thought is was possibly due to `qc1` being in the nonconvex-looking rotated form, but I removed that and it still complains about `qc0`.

Let me know where you get with this and we can modify the tests if possible. E.g., if there certain convex forms that Mosek doesn't recognize, then we can have the test suite skip over them.

aszekMosek:
Hi, Michael from Mosek here.

qc0 as such does not define a convex set over all real numbers - you would have to assume ``model.z>=0`` for instance. 

Regarding license - we could provide a license for testing as long as it would not be visible to the public. Send an email to support@mosek.com if you have an idea for how that can be done. (The same if someone from the pyomo team would need a license for testing locally).

ghackebeil:
@aszekMosek:

Hi Michael,

I assume what you are saying is that these conic constraints need to be specified via the "putcone" Mosek interface rather than "putqcon".

We may need to introduce some new constraint types into Pyomo that will allow the Mosek interface to easily identify these. Another option would be to add some mechanism into this interface to detect quadratic constraints that are conic.

In the meantime, go ahead and tag these as expected failures by adding the following somewhere in `pyomo/solvers/tests/testcases.py`:
```python
#                                                                                                                                              
# MOSEK                                                                                                                                        
#                                                                                                                                              

ExpectedFailures['mosek_direct', 'python', 'QCP_simple'] = \
    (lambda v: True,
     "Conic constraints not yet handled by this interface")

ExpectedFailures['mosek_direct', 'python', 'QCP_simple_nosuffixes'] = \
    (lambda v: True,
     "Conic constraints not yet handled by this interface")

ExpectedFailures['mosek_direct', 'python', 'MIQCP_simple'] = \
    (lambda v: True,
     "Conic constraints not yet handled by this interface")
```

and don't forget to make the 'mosek_direct' -> 'mosek' string substitution if you follow my earlier recommendation.

ghackebeil:
@aszekMosek, @harunkivril: Just a heads up that we are hoping to do a release in a day or two. It would be great to get this in before then. Let us know if you won't have time.

aszekMosek:
@ghackebeil As far as I know @harunkivril wil work on it on Wednesday.

Detecting simple conic constraints is perhaps a good thing to think about for the future.

harunkivril:
Hi @ghackebeil,
Thanks for your comments.
Michael is right, I will be able to make the changes that you mention on Wednesday.

pyomo/solvers/plugins/solvers/mosek_direct.py
```diff
@@ -0,0 +1,854 @@
...
+            self.results.solver.status = SolverStatus.Aborted
+            self.results.solver.termination_message = "Optimization terminated because of the user callback "
+            self.results.solver.termination_condition = TerminationCondition.userInterrupt
+            soln.status = SolutionStatus.unknown
+
+        elif self._termcode in [msk.rescode.trm_mio_num_relaxs,
+                                msk.rescode.trm_mio_num_branches,
+                                msk.rescode.trm_num_max_num_int_solutions]:
+            self.results.solver.status = SolverStatus.ok
+            self.results.solver.termination_message = "Optimization terminated because maximum number if relaxations" \
```
blnicho: Typo: maximum number **of**

ghackebeil:
@blnicho, @jsiirola: Any idea what went wrong on Jenkins after the last build?

blnicho:
I think we need to merge master into this. Looks like Jenkins failed because this branch is missing the `jenkins.sh` testing script

ghackebeil:
Ah okay. Well I've verified again that everything works locally. So I am okay with skipping that step if you are.

pyomo/solvers/tests/checks/test_mosek.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def setUp(self):
+        self.stderr = sys.stderr
+        sys.stderr = None
+
+    def tearDown(self):
+        sys.stderr = self.stderr
+
+    @unittest.skipIf(not mosek_available,
+                     "The 'mosek' python bindings are not available")
```
jsiirola: Generally, I think we should rely on `SolverFactory("mosek", solver_io='python').available()` instead of the import test.

pyomo/solvers/plugins/solvers/mosek_direct.py
```diff
@@ -0,0 +1,854 @@
...
+def _is_numeric(x):
+    try:
+        float(x)
+    except ValueError:
+        return False
+    return True
+
+
+@SolverFactory.register('mosek', doc='Direct python interface to Mosek')
+class MosekDirect(DirectSolver):
```
jsiirola: I believe the interface is missing the `.available()` method.  That would be the standard approach for detecting if the solver interface was available and properly licensed.

pyomo/solvers/tests/models/base.py
```diff
@@ -57,7 +57,7 @@ def __init__(self):
         self.results_file = None
         self.disable_suffix_tests = False
         self.test_suffixes = []
-        self.diff_tol = 1e-4
+        self.diff_tol = 1e-3
```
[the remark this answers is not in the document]
ghackebeil: No.

pyomo/solvers/plugins/solvers/mosek_direct.py
ghackebeil: The base class implements the `available()` method based on the instance variable `self._python_api_exists`, which the mosek plugin sets.

I can confirm that `pyomo help --solvers` works with and without mosek being available. Is there another command you want me to test?

jsiirola: The problem I am seeing is:
```
>>> s = SolverFactory('mosek')
>>> s.available()
  File "[...]/pyomo/solvers/plugins/solvers/direct_or_persistent_solver.py", line 299, in available
    "plugin").format(type(self)))
pyutilib.common._exceptions.ApplicationError: No Python bindings available for <class 'pyomo.solvers.plugins.solvers.mosek_direct.MosekDirect'> solver plugin
```
I would expect `available()` to return False.

ghackebeil: I agree it should return False, but that has been the default behavior of `available()` since it was implemented. I believe you can pass `exception_flag=False` to get the more rational behavior, which we should make the default in a new PR.

ghackebeil:
There were some failures on appveyor that had to do with conda. I'm going to merge this in a few minutes unless there is a strong objection.

ghackebeil:
@harunkivril: Thanks again for putting this together. This is a very well written plugin.

harunkivril:
@ghackebeil You are welcome.  Thanks for your effort to merge it.

PR #975, opened 2019-04-27
Fix pyomo.network decomposition when ports contain References

### LINKED ISSUE #974: Bug in Sequential Decomposition `tear_diff_direct`
Working with @jsiirola, we have found a bug in the sequential decomposition where an mismatch of indices is occurring.

In this case, I have two blocks with Ports with the following structure:

```
Size=1, Index=None
Key  : Name                : Size : Variable
None : flow_mol_phase_comp :    8 : fs.H101._inlet_flow_mol_phase_comp_ref
     :            pressure :    1 :            fs.H101._inlet_pressure_ref
     :         temperature :    1 :         fs.H101._inlet_temperature_ref
```

These Ports are connected via an Arc, which is the tear stream in my model. However, `flow_mol_phase_comp_ref` is a Pyomo `Reference` with three indices (time, phase, component), which references a set of variables in an indexed block (form: `block[time].flow_mol_phase_comp[phase, component]`).

When running the sequential decomposition, I get the following traceback:

  File "module_2.py", line 196, in <module>
    seq.run(m, function)
  File "/home/alee/pyomo/pyomo/pyomo/network/decomposition.py", line 336, in run
    self.solve_tear_direct(**kwds)
  File "/home/alee/pyomo/pyomo/pyomo/network/foqus_graph.py", line 123, in solve_tear_direct
    svals, dvals = self.tear_diff_direct(G, tears)
  File "/home/alee/pyomo/pyomo/pyomo/network/decomposition.py", line 844, in tear_diff_direct
    dvals.append(value(self.source_dest_peer(arc, name, index)))
  File "/home/alee/pyomo/pyomo/pyomo/network/decomposition.py", line 686, in source_dest_peer
    return mem[index]
  File "/home/alee/pyomo/pyomo/pyomo/core/base/indexed_component.py", line 334, in __getitem__
    validated_index = self._validate_index(index)
  File "/home/alee/pyomo/pyomo/pyomo/core/base/indexed_component.py", line 494, in _validate_index
    % ( idx, self.name, ))
KeyError: "Index '('Liq', 'benzene')' is not valid for indexed component 'fs.H101._inlet_flow_mol_phase_comp_ref'"

It appears that the sequential decomposition is taking the indexes from the original `flow_mol_phase_comp` variable (some instance), and is trying to use these as the indices for the `flow_mol_phase_comp_ref` reference.

jsiirola:
## Summary/Motivation:
This fixes a bug in pyomo.network when resolving sorce/destination variables when the variables are defined using references.  The root cause was that `_VarData.index()` returns the index from the point of view of the `_VarData` and not the reference in the Port.

- Change how ports iterate over members and use the index defined by the indexed component in the Port (potentially a Reference)
- Update one of the decomposition tests to include a Port member defined by a Reference to exercise this functionality.

### Legal Acknowledgement

### DIFF
--- a/pyomo/network/decomposition.py
+++ b/pyomo/network/decomposition.py
@@ -598,7 +598,7 @@ class SequentialDecomposition(FOQUSGraph):
 
     def load_values(self, port, default, fixed, use_guesses):
         sources = port.sources()
-        for name, obj in port.iter_vars(fixed=False, names=True):
+        for name, index, obj in port.iter_vars(fixed=False, names=True):
             evars = None
             if port.is_extensive(name):
                 # collect evars if there are any
@@ -610,9 +610,8 @@ class SequentialDecomposition(FOQUSGraph):
                     try:
                         # index into them if necessary, now that
                         # we know they are not None
-                        i = obj.index()
                         for j in range(len(evars)):
-                            evars[j] = evars[j][i]
+                            evars[j] = evars[j][index]
                     except AttributeError:
                         pass
             if evars is not None:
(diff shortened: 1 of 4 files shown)


jsiirola:
The problem comes from the convention that all ComponentData objects hae exactly one `parent_component()`.  When you ask a, e.g., `_VarData` for it's index, it will return it's index within its `parent_component()`'s `_data` dict.

References are a bit strange.  They can best be thought of as "views" into the model hierarchy.  For example:
```python
m = ConcreteModel()
@m.Block([1,2,3])
def b(m, i):
    b.x = Var([4,5,6],[7,8,9])
m.r = Reference(m.b[:].x[5,:])
v = m.r[1,7]
v is m.b[1].x[5,7]  # True
v.parent_component() is m.b[1].x  # True
m.index() is (5,7)  # True
```
Here, the "index" for `v` from the point of view of the reference (`m.r`) is `[1,7]`, but it's index (that is it's index within `m.b[1].x`) is `[5,7]`.

PR #978, opened 2019-04-29
Standard repn bug

michaelbynum:
## Summary/Motivation:
```generate_standard_repn``` raises an exception when a constant is raised to the power of a variable which is fixed to a value of 1 or 2. This PR adds ```float``` and ```int``` to the ```_repn_collectors``` dict, to handle these cases. The motivating problem had a constant raised to an expression of binary variables. When the binary variables were fixed (to solve an NLP), an error was raised when writing the nl file.

### Legal Acknowledgement

### CODE REVIEW

**Comment by ghackebeil** [comment]
@michaelbynum: IMO, the bug is that the second check in _collect_pow is checking
```python
    elif not exp._args_[1].is_potentially_variable():
        ...
```
when it should be checking
```python
    elif exp._args_[1].is_fixed():
        ...
```
The addition of `float` and `int` to the `_repn_collectors` dict doesn't really cover all the cases, and it may mask other spots where we make this mistake.

On the other hand, if the outcome is the same, it probably doesn't matter. Anyone else have thoughts?

michaelbynum:
@ghackebeil Good point. I can make this change, but I will wait to see if others have input first.

jsiirola:
I share @ghackebeil's confusion on the use of `is_potentially_variable()` over `is_fixed()`.  In my mind, `is_potentially_ariable()` is relevant during expression generation, but `generate_standard_repn()` is an expression compiler, and should probably only use `is_fixed()`.  I am similarly confused by the role of the `compute_values` flag, which I believe is only used in the bilevel and duality modules.

In the mean time (before we sort out the potentially variable issue), I think I would rather the following patch, which is more robust to other types (like numpy.float64) sneaking into the expression system:
```diff
diff --git a/pyomo/repn/standard_repn.py b/pyomo/repn/standard_repn.py
index e89c4d1..50714d5 100644
--- a/pyomo/repn/standard_repn.py
+++ b/pyomo/repn/standard_repn.py
@@ -921,6 +921,9 @@ def _collect_standard_repn(exp, multiplier, idMap,
     fn = _repn_collectors.get(exp.__class__, None)
     if fn is not None:
         return fn(exp, multiplier, idMap, compute_values, verbose, quadratic)
+    if exp.__class__ in native_numeric_types:
+        return _collect_const(exp, multiplier, idMap, compute_values,
+                              verbose, quadratic)
     #
     # These are types that might be extended using duck typing.
     #
```

PR #979, opened 2019-04-29
detecting infeasible constraints in FBBT

### LINKED ISSUE #969: FBBT can have math domain error in log()
In interval,py, we have the following:
```
def log(xl, xu):
    if xl > 0:
        return math.log(xl), math.log(xu)
    elif xl == 0:
        if xu > 0:
            return -math.inf, math.log(xu)
        else:
            return -math.inf, -math.inf
    else:
        return -math.inf, math.inf
```

If `xu = 0`, then `math.log(xu)` raises a ValueError.

@michaelbynum did you mean to have:
```
def log(xl, xu):
    if xl > 0 and xu > 0:
        return math.log(xl), math.log(xu)
    elif xl == 0:
        if xu > 0:
            return -math.inf, math.log(xu)
        else:
            return -math.inf, -math.inf
    else:
        return -math.inf, math.inf
```

I'm also a bit confused about the branch that returns `-math.inf, -math.inf`.

michaelbynum:
## Summary/Motivation:
This PR adds code to raise an exception when infeasible constraints are detected during FBBT. If this exception is not raised, a more difficult to interpret exception is often raised at a later point in time.

### Legal Acknowledgement


michaelbynum:
@jsiirola Good questions. I actually meant to choose 1e-5 for the integer tolerance (for two reasons). First, I want to be very sure that, for example, a value of 0 for a binary variable is infeasible before changing the lower bound to 1. The larger this tolerance, the more "sure" I can be. Second, if I solve the the problem

```
min x
s.t. x >= XL
      x \in {0, 1}
```

with Gurobi, the solution will be 0 as long as ```XL <= 1e-5```. As for the feasibility tolerance, 1e-6 is the default feasibility tolerance for Gurobi. I realize some solvers use 1e-8, and I am really fine with either.

PR #980, opened 2019-04-30
Add specialized conic constraints to kernel

ghackebeil:
## Summary/Motivation:

This PR adds six specialized conic constraint implementations into the `pyomo.kernel` namespace. These are
 - `quadratic`
 - `rotated_quadratic`
 - `primal_exponential`
 - `primal_power`
 - `dual_exponential`
 - `dual_power`

These classes implement the standard constraint interface, so they work with the NL writer (and the LP / MPS writer if quadratic). They also work with the existing general constraint containers (`constraint_dict`, `constraint_list`, `constraint_tuple`).

They lazily build and cache the documented constraint expression when the `.body` property is accessed. The new Mosek plugin will be able to identify and directly translate these convex constraint forms without using the expression system.

## Example
```python
import pyomo.kernel as pmo
x1 = pmo.variable()
x2 = pmo.variable()
r = pmo.variable(lb=0)
# x1^2 + x2^2 <= r^2
c = pmo.conic.quadratic(x=[x1,x2], r=r)
assert c.check_convexity_conditions()
```
## TODO:
 - [x] update Mosek solver plugin
 - ~~update some of the models used by `pyomo/solvers/tests/checks/test_writer.py` to use these new classes~~ I'll revisit this when adding support for SDPs

### Legal Acknowledgement


erling-d-andersen:
If you extension allows one to do

A x + b in cone

i.e. a linear expression in the cone and not just

x  in cone 

then the users will be much more happy. Although this is just syntactic sugar but very important sugar.

If you need help Mosek with the Mosek plugin then let us know.

ghackebeil:
@erling-d-andersen: Thank you for the feedback. I'd be happy to explore how we can extend this design to make common use cases more convenient.

I think it would certainly help to see how this would be handled through the Mosek API. Mostly, I am wondering if it would involve creating intermediate variables, or if there is a more direct way of expressing it with Mosek.

@harunkivril, @aszekMosek: Feel free to chime in if you have any feedback on the changes I am making to the Mosek plugin to support this.

Quick question: Does Mosek provide a way to collect the dual solution for conic constraints?

aszekMosek:
@ghackebeil 

Currently if you wanted to make Ax+b in the cone then you will have to make a slack variable z=Ax+b and use z in putcone.

BTW some such slacks are anyways necessary even in more basic cases because in Mosek each variable can appear in at most one cone. So if x appears in two cones in the model you must duplicate it when calling mosek.

The dual solution for conic constraints is called snx in https://docs.mosek.com/9.0/pythonapi/prob-def-conic.html#duality-for-conic-optimization and the method is getsnx.

ghackebeil:
@aszekMosek, @harunkivril:

I have an alternate syntax in mind that we could also support that may be more intuitive for those with a background in conic optimization.
```python
m = pmo.block()
m.x = pmo.variable()
m.y = pmo.variable()
m.z = pmo.variable()
Qr, aux_c, aux_x = pmo.conic.rotated_quadratic.\
    as_domain(x=[math.sqrt(2)],
              r1=2*m.x-m.y
              r2=m.z+1)
```
which would be equivalent to something like:
```python
aux_x = pmo.variable_list([
    pmo.variable() for i in range(3)])
aux_c = pmo.constraint_list([
    pmo.constraint(aux_x[0] == math.sqrt(2)),
    pmo.constraint(aux_x[1] == 2*m.x-m.y),
    pmo.constraint(aux_x[2] == m.z+1)
Qr = pmo.conic.rotated_quadratic(x=[aux_x[0]],
                                 r1=aux_x[1],
                                 r2=aux_x[2])
```
Would something like this do the job?

Another quick question about getsnx(): I was thrown off by the fact that this aspect of the solution has length `getnumvar` instead of `getnumcone`. Is there an easy computation on the snx values for the variables of a cone so that I can express the dual of the main conic constraint (e.g., `x^2 + y^2 <= z^2`) as a scalar?

aszekMosek:
@ghackebeil I have never used Pyomo so I can't really judge but the first snippet looks very nice and clear from a conic perspective.

The only way I can interpret your duality question is: if I have a quadratic cone constraint with ``(x,y,z)`` in QuadCone, what would be the dual of ``x^2-y^2-z^2>=0`` if I reformulated the problem as an QP.  If I did my back-of-the-envelope check correctly now then it is ``snx/(2*x)`` for any of the variables appearing in the cone. This is based on similar calculations as in https://docs.mosek.com/modeling-cookbook/qcqo.html#id1 that relates the dual of QP and dual of its conic reformulation.

ghackebeil:
I've added support and tests for the `as_domain` syntax, I've also included a section in the Kernel docs about direct support for conic constraints with Mosek. I'm going to put off the issue of extracting duals for later. I plan on prototyping some new components in another PR that will add support for PSD constraints through Mosek, so perhaps I will revisit it then.

I think this PR read to merge.

ghackebeil:
I've canceled the appveyor build because the queue is crazy right now, and Mosek is not installed there anyway. But this is tested locally and ready to merge.

doc/OnlineDocs/library_reference/kernel/index.rst
```diff
@@ -91,6 +91,96 @@ A simplified version of this using :python:`pyomo.environ` components might look
 
 The transformer expressed using :python:`pyomo.kernel` components requires roughly 2 KB of memory, whereas the :python:`pyomo.environ` version requires roughly 8.4 KB of memory (an increase of more than 4x). Additionally, the :python:`pyomo.kernel` transformer is fully compatible with all existing :python:`pyomo.kernel` block containers.
 
+Direct Support For Conic Constraints with Mosek
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Pyomo 5.6.2 introduced support into :python:`pyomo.kernel`
```
blnicho: I think this will end up being Pyomo 5.6.3 unless you instead want to reference the first version of Pyomo to include the Mosek interface.

doc/OnlineDocs/library_reference/kernel/index.rst
```diff
@@ -91,6 +91,96 @@ A simplified version of this using :python:`pyomo.environ` components might look
...
+constraint containers (:python:`constraint_tuple`,
+:python:`constraint_list`, :python:`constraint_dict`).
+
+Each conic constraint class supports two methods of
+instantiation. The first method is to directly instantiate
+a conic object, providing all necessary input variables:
+
+.. code-block:: python
+
+   import pyomo.kernel as pmo
```
blnicho: Any objection to including these as doctests instead of code-blocks?

doc/OnlineDocs/library_reference/kernel/index.rst
```diff
@@ -91,6 +91,96 @@ A simplified version of this using :python:`pyomo.environ` components might look
...
+class, but in place of variables, each argument can
+optionally be assigned a constant or expression. The
+:python:`as_domain` method will return a :python:`block`
+object that includes the core conic constraint expressed
+using automatically created auxiliary variables, which are
+linked to the input arguments using auxiliary
+constraints. Example:
+
+.. code-block:: python
+
```
blnicho: Same comment

doc/OnlineDocs/library_reference/kernel/index.rst
ghackebeil: Thanks. I meant to ask that.

doc/OnlineDocs/library_reference/kernel/index.rst
ghackebeil: No objection. I can do that.

ghackebeil:
@jsiirola: Do you have the Mosek license set up for Jenkins yet? Just curious.

PR #983, opened 2019-05-01
Cleaning up oa implementation

RomeoV:
## Fixes 
No issue assigned

## Summary/Motivation:
Soon I will be starting to work on the Level based outer approximation. Since I want to re-use a lot of the code already available in the current OA implementation, I was going through the code. Some methods were really big (especially the nlp_subproblem solve and the master_mip_solve methods).
The main change in this PR is splitting up the solving methods into the actual solving (where not much has changed) and the handling of the results.  
To achieve that, some minor-non breaking changes have been made to `gdpopt/util.py` too, since Mindtpy uses methods from there.

Finally, I wrote a bunch of docstrings for the OA methods and added a couple more tests testing more diverse scenarios. (This could still be expanded though) 

- Add possibility for force-copying floats into binaries (in `gdpopt/util`). This is neccessary when the initial relaxed NLP is solved and an OA-cut is constructed
- Add option in `gdpopt/util/process_objective` to always move objective function to constraints, so that the master problem can be solved by minimizing the auxiliary objective variable
- Seperate `mindtpy/nlp_solve` main solve function into solve and handle results
- Seperate `mindtpy/mip_solve` main solve function into solve and handle results
- Move some of the logic from `nlp_solve` and `mip_solve` to the iteration loop
- Add a bunch of docstrings to the methods

## WIP parts
- [x] correct infeasible handling for NLP
- [x] increasing value handling only for psc
- [x] TransformationFactory('core.fix_discrete') broken? (see #996)
- [x] docstring for outer approximation (@bernalde)
- [x] replace outer approximation with version I implemented for feasibility pump (I'll get to that tomorrow)

### Legal Acknowledgement


pyomo/contrib/preprocessing/plugins/int_to_binary.py
```diff
@@ -103,16 +103,18 @@ def _apply_to(self, model, **kwds):
...
                 idx, expr=int_var == sum(
                     reform_block.new_binary_var[idx, pwr] * (2 ** pwr)
-                    for pwr in range(0, highest_power + 1)))
+                    for pwr in range(0, highest_power + 1))
+                    + int_var.lb)
 
             # Relax the original integer variable
-            int_var.domain = NonNegativeReals
+            # TODO-romeo @qtothec why do we need this?
+            #int_var.domain = NonNegativeReals
```
qtothec: We still need it because otherwise your model still has integer variables in it.

pyomo/contrib/gdpopt/util.py
```diff
@@ -144,24 +142,37 @@ def a_logger(str_or_logger):
...
+            continue  # Skip fixed variables.
         try:
             v_to.set_value(value(v_from, exception=False))
             if skip_stale:
                 v_to.stale = False
         except ValueError as err:
             err_msg = getattr(err, 'message', str(err))
             var_val = value(v_from)
-            rounded_val = int(round(var_val))
+            rounded_val = round(var_val) if six.PY3 else int(round(var_val))
```
qtothec: is there any harm in having int(round(var_val)) in Python 3, or do we need to have the check?

pyomo/contrib/gdpopt/util.py
```diff
@@ -144,24 +142,37 @@ def a_logger(str_or_logger):
...
-            if 'is not in domain Binary' in err_msg and (
+            if ignore_integrality \
+                and ('is not in domain Binary' in err_msg
+                or 'is not in domain Integers' in err_msg):
+               v_to.value = value(v_from, exception=False)
+            elif 'is not in domain Binary' in err_msg and (
                     fabs(var_val - 1) <= config.integer_tolerance or
                     fabs(var_val) <= config.integer_tolerance):
                 v_to.set_value(rounded_val)
+            # TODO What about PositiveIntegers etc?
```
qtothec: If it's in a domain that is not explicitly handled here, then the user gets handed the exception via re-raise. If you would like to add in handling for the other domain groups, be my guest.

pyomo/contrib/mindtpy/tests/MINLP2_simple.py
```diff
@@ -4,7 +4,7 @@
 MINLP test problem in Pyomo
 Author: David Bernal <https://github.com/bernalde>
 
-The expected optimal solution value is 6.0072.
+The expected optimal solution value is 6.00976.
```
qtothec: Huh? Was this simply wrong before?

pyomo/contrib/mindtpy/tests/test_mindtpy.py
```diff
@@ -5,7 +5,10 @@
 import pyutilib.th as unittest
 from pyomo.contrib.mindtpy.tests.eight_process_problem import \
     EightProcessFlowsheet
-from pyomo.contrib.mindtpy.tests.MINLP_simple import SimpleMINLP
+from pyomo.contrib.mindtpy.tests.MINLP_simple import SimpleMINLP as SimpleMINLP
+from pyomo.contrib.mindtpy.tests.MINLP2_simple import SimpleMINLP as SimpleMINLP2
+from pyomo.contrib.mindtpy.tests.MINLP3_simple import SimpleMINLP as SimpleMINLP3
```
qtothec: is there no better name for these?

pyomo/contrib/mindtpy/util.py
```diff
@@ -34,15 +34,6 @@ def model_is_valid(solve_data, config):
...
 
-    # Check for any integer variables
-    if any(True for v in m.component_data_objects(
-            ctype=Var, descend_into=True)
-            if v.is_integer() and not v.fixed):
-        raise ValueError('Model contains unfixed integer variables. '
-                         'MindtPy does not currently support solution of '
-                         'such problems.')
-        # TODO add in the reformulation using base 2
-
```
qtothec: With this deletion, you might be able to just use the `model_is_valid()` function from `gdpopt.util`. Only difference is that I don't set up the `dual` Suffix here. You'll need to change the `gdpopt.util.model_is_valid()` function slightly to use the `util_blk = getattr(m, solve_data.util_block_name)` idiom instead of directly hardcoding `GDPopt_utils`.

pyomo/contrib/preprocessing/plugins/int_to_binary.py
qtothec: You still have this line commented. You need to have an update from `Integers` to `Reals`, otherwise your model still has integer variables in it. You've just now added binary variables to correspond to each of the integer variables. The purpose of this transformation is to reformulate away all the integers.

pyomo/contrib/gdpopt/util.py
```diff
@@ -89,28 +87,28 @@ def model_is_valid(solve_data, config):
...
-def process_objective(solve_data, config):
+def process_objective(solve_data, config, always_move_objective=False):
     m = solve_data.working_model
     util_blk = getattr(m, solve_data.util_block_name)
     # Handle missing or multiple objectives
-    objs = list(m.component_data_objects(
+    active_objectives = list(m.component_data_objects(
         ctype=Objective, active=True, descend_into=True))
-    num_objs = len(objs)
-    solve_data.results.problem.number_of_objectives = num_objs
```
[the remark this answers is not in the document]
RomeoV: I disliked the name `objs` so I changed it to the more verbose `active_objectives`.
I also disliked introducing a new variable (and new name to keep in mind) when calling `len(active_objectives)` does the job.
`num_objs` doesn't tell you whether it's `number_of_objects` or `number_of_objectives` or `number_of_active_objectives`. So instead I changed all occurrences (it's only a few) of `num_objs` to `len(active_objectives)` which is very clear what it is.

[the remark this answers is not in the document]
qtothec: Oh, I agree with the changes you've listed above. I'm asking here specifically about the line `solve_data.results.problem.number_of_objectives = num_objs`, which removes functionality.

pyomo/contrib/gdpopt/util.py
```diff
@@ -89,28 +87,28 @@ def model_is_valid(solve_data, config):
     return True
 
 
-def process_objective(solve_data, config):
+def process_objective(solve_data, config, always_move_objective=False):
```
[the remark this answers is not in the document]
RomeoV: Imo `force_xyz` means that something throws an error/warning or something similar, but then we override that error and still do `xyz` (think for example `bash: rm -f *`).
`always_do_xyz` to me sounds like we override an if-clause, in this case the `is_nonlinear` decision.
Maybe something like `also_move_linear_objective` would do the trick better?

For the current implementation I agree a docstring is neccesary, but  I think `also_move_linear_objective` might be verbose enough by itself. There still should be a docstring though, describing in general what the purpose of the method is (that should probably already have been there when I started to work on it ;P)

pyomo/contrib/gdpopt/util.py
RomeoV: There is no harm in it. When I looked at it I was surprised to see `int(round(var_val))`, because I expected `round(x)` to be always an integer. After looking up and understanding the problem (python2 returns a float there) I decided on this implementation because (at least to me) it points out why we need the `int` cast.

If you would like me to change it back that's okay though.

pyomo/contrib/mindtpy/MindtPy.py
```diff
@@ -239,17 +250,22 @@ def solve(self, model, **kwds):
         solve_data.results = SolverResults()
         solve_data.timing = Container()
 
+        solve_data.original_model = model
+        solve_data.working_model = model.clone()
+        if config.integer_to_binary:
+            TransformationFactory('contrib.integer_to_binary'). \
+                apply_to(solve_data.working_model)
```
RomeoV: I just want to point out the reason it was necessary to move the `working_model` declaration further up (outside of the `create_utility_block` context) is that the `TransformationFactory('contrib.integer_to_binary')` introduces new variables. This way `solve_data.working_model.MindtPy_utils.variable_list` will include those variables.

pyomo/contrib/gdpopt/util.py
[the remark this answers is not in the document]
qtothec: You could shorten that to `move_linear_objective` and get the same meaning, then.

pyomo/contrib/mindtpy/tests/MINLP2_simple.py
RomeoV: Yeah it's literally different in the paper.
See https://www.sciencedirect.com/science/article/pii/009813549587027X Table 1 (page 5)

pyomo/contrib/mindtpy/tests/test_mindtpy.py
RomeoV: It's what they were already named - I didn't implement them myself. I agree that there could be better names

pyomo/contrib/mindtpy/util.py
RomeoV: Uhm I'm not sure if we agree on the function of this code.
The way it was written before, the model was checked if it contained any integers (since it didn't support integer variables in the solver).
Now I used the `TransformationFactory('contrib.integer_to_binary')` so that models with integer variables can be solved too. Thus, we don't need to check for this anymore.

pyomo/contrib/preprocessing/plugins/int_to_binary.py
RomeoV: Oh, I have a misunderstanding of the method then. The way I want to use it is to just introduce the binary variables so that I can do an integer cut. For my use-case, I don't need the variable to be relaxed.

pyomo/contrib/mindtpy/util.py
qtothec: I think there's a misunderstanding here. I was talking about `gdpout.util.model_is_valid()`, which last time I looked does not include any check for integer variables. I am saying that because of the proposed change here, which I agree with, there is a potential opportunity for code deduplication.

pyomo/contrib/mindtpy/tests/MINLP2_simple.py
qtothec: The paper also says that Duran and Grossmann (1986) give a solution of 6.00972. Looks like numerical noise to me, so I suppose it's fine. What solution do you end up getting?

pyomo/contrib/gdpopt/util.py
qtothec: Yeah, may as well just leave it as `int(round(var_val))`. You can pop in a comment if you want to. Leaves the `six` occurrences for when there are larger compatibility concerns.

pyomo/contrib/gdpopt/util.py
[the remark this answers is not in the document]
RomeoV: Oh yeah, you're right. I guess I overlooked that.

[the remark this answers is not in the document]
RomeoV: Wait I just reviewed once more, and I just change `solve_data.results.problem.number_of_objectives = num_objs` to `solve_data.results.problem.number_of_objectives = len(active_objectives)`, so no functionality is lost right?

pyomo/contrib/mindtpy/tests/MINLP2_simple.py
```diff
@@ -70,7 +70,7 @@ def __init__(self, *args, **kwargs):
         m.const7 = Constraint(expr=Y[1] + Y[2] <= 1)
 
         """Cost (objective) function definition"""
-        m.cost = Objective(expr=-5*Y[1] - 6*Y[2] - 8*Y[3] - X[4], sense=maximize)
+        m.cost = Objective(expr=+5*Y[1] + 6*Y[2] + 8*Y[3] + X[4], sense=minimize)
```
RomeoV: Come to think of it, I can change it back so that we check for maximization as well.

pyomo/contrib/gdpopt/util.py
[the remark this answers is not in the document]
RomeoV: Fixed in 4b8d692595588f00b0e58c82144a81ff4fab36cd

pyomo/contrib/gdpopt/util.py
```diff
@@ -89,28 +87,28 @@ def model_is_valid(solve_data, config):
...
-        main_obj = objs[0]
+        main_obj = active_objectives[0]
     solve_data.results.problem.sense = main_obj.sense
     solve_data.objective_sense = main_obj.sense
 
     # Move the objective to the constraints if it is nonlinear
-    if main_obj.expr.polynomial_degree() not in (1, 0):
+    if main_obj.expr.polynomial_degree() not in (1, 0) \
+            or always_move_objective:
         config.logger.info("Objective is nonlinear. Moving it to constraint set.")
```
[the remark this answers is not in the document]
RomeoV: Fixed in 4b8d692595588f00b0e58c82144a81ff4fab36cd

pyomo/contrib/gdpopt/util.py
RomeoV: Fixed in 4b8d692595588f00b0e58c82144a81ff4fab36cd

pyomo/contrib/gdpopt/util.py
RomeoV: I think somebody should have another look at it at some point, so I will keep the TODO tag.
But I won't work on it now.

pyomo/contrib/mindtpy/tests/MINLP2_simple.py
RomeoV: Fixed in  5a9a61c446342e6649022595553a54a2b1e68d3e

pyomo/contrib/mindtpy/tests/MINLP2_simple.py
RomeoV: You're right, both previous versions had a minor mistake.
Fixed in aafd64294b77d34eebc452fd2c4b56e023aadddc

pyomo/contrib/mindtpy/tests/test_mindtpy.py
RomeoV: Someone can rename them but I'm not up for it right now

pyomo/contrib/mindtpy/iterate.py
```diff
@@ -23,54 +25,75 @@ def MindtPy_iteration_loop(solve_data, config):
...
-        if not making_progress and (
-                config.strategy == 'hPSC' and
-                config.strategy == 'PSC'):
-            config.logger.info(
-                'Not making enough progress for {} iterations. '
-                'Switching to OA.'.format(max_nonimprove_iter))
-            config.strategy = 'OA'
+            if not making_progress and (
+                    config.strategy == 'hPSC' and
+                    config.strategy == 'PSC'):
```
qtothec: There's a logical error in the `and` here.

pyomo/contrib/preprocessing/plugins/int_to_binary.py
qtothec: I think this is the only blocking issue that I have with this PR, as it breaks functionality for my use cases. If there's no objections, I will simply revert this line and call it a day with this PR.

pyomo/contrib/mindtpy/iterate.py
blnicho: @qtothec is this logic error you previously pointed out still an issue?

qtothec:
Ok, my small change to `integer_to_binary` seems to have broken the PhD proposal test problem. Offhand, I fail to see how this is possible, since the transformation produces a mathematically equivalent formulation, regardless of if the original integer variable domain is changed to `Reals` or if it remains `Integers`.

pyomo/contrib/mindtpy/iterate.py
qtothec: @blnicho Yes, good catch. I had changed it in my own copy of the code.

qtothec:
There appears to be an implementation error in the addition of integer cuts.

qtothec:
Hopefully I have resolved the outstanding issues.

blnicho:
@jsiirola could you take another quick look at this one?

PR #989, opened 2019-05-02
Replace ReciprocalExpression with DivisionExpression

jsiirola:
## Summary/Motivation:
This is a combined PEP / candidate PR.  Discussion in this PR is expected to cover the merits of the code design as well as the implementation.

One of the parts of #439 (Pyomo5 design issues) was to look at replacing ReciprocalExpression with DivisionExpression.  When I started looking in to it, I could not find any operations that were simpler by only working with reciprocal nodes in the expression trees.  Reintroducing division has several advantages, too:
- general nonlinear expressions became simpler: if you look at the NL baseline differences, in all cases, `* (1/x)` (2 operators) was replaced by a single division operator.
- as there is one fewer operators, it is likely to be subject to less floating point error (although I do not have a case that illustrates this).
- the expression graphs stay closer to what the user actually wrote.  A user can always regain the old behavior by writing their expressions with multiplication and reciprocal operations to recover the old expression graphs.

This PR does not remove ReciprocalExpression - only deprecate it, so dependent packages will not suffer import errors.  However, as generate_expression() now generates DivisionExpression nodes instead of ReciprocalExpression nodes, packages that walk expression trees (notably [cog-imperial/suspect](https://github.com/cog-imperial/suspect), @fracek) will need to be updated.

This PR attempts to update the whole Pyomo code base, including:
- [ ] generate_expression (@whart222)
- [ ] generate_standard_repn (@whart222)
- [x] the NL writer (@carldlaird, @ghackebeil)
- [x] DAE simulator (@blnicho)
- [x] FBBT (@michaelbynum)
- [x] units (@carldlaird)
- [x] differentiate (@michaelbynum)
- [x] MCPP (@qtothec)
- [x] satsolver (@qtothec)
...I would appreciate if the noted developers would review changes to their code.

Finally, this PR is expected to have test failures due to needed baseline updates in pyomo-model-libraries (I cannot update those baselines before this PR is accepted as the master builds / other PRs would start to fail)

- Add DivisionExpression
- Deprecate ReciprocalExpression
- propagate these changes through all walkers.

### Legal Acknowledgement


michaelbynum:
I have not yet looked at the implementation, but I am in favor of this change. It simplifies my work in Coramin to automatically relax MINLP's.

michaelbynum:
@jsiirola FBBT and differentiate both look good.

pyomo/core/expr/numeric_expr.py
```diff
@@ -784,6 +785,45 @@ class MonomialTermExpression(ProductExpression):
...
+    """
+    __slots__ = ()
+    PRECEDENCE = 4
+
+    def nargs(self):
+        return 2
+
+    def _precedence(self):
+        return DivisionExpression.PRECEDENCE
+
```
carldlaird: I suspect that _cpmpute_polynomial_degree is not returning the correct polynomial degree for many expressions.

1) Can you clarify why we only return a result if the denominator is constant? Is it just that it is too difficult to compute this for more complex expressions?

2) Is other code assuming that None means "not known"?

pyomo/core/expr/numeric_expr.py
```diff
@@ -784,6 +785,45 @@ class MonomialTermExpression(ProductExpression):
     def getname(self, *args, **kwds):
         return 'mon'
 
+class DivisionExpression(ExpressionBase):
```
carldlaird: What is the :: notation in the doc string - does this process specially? It would be good to have a note here that indicates args[0] is numerator and args[1] is denominator.

pyomo/core/expr/numeric_expr.py
```diff
@@ -1562,6 +1607,12 @@ def _decompose_linear_terms(expr, multiplier=1):
                 yield term
         else:
             raise LinearDecompositionError("Quadratic terms exist in a product expression.")
+    elif expr.__class__ is DivisionExpression:
+        if expr._args_[1].__class__ in native_numeric_types or not expr._args_[1].is_potentially_variable():
+            for term in _decompose_linear_terms(expr._args_[0], multiplier/expr._args_[1]):
```
carldlaird: I assume that this iterator throws an exception if the argument passed is, in fact, nonlinear?

pyomo/core/expr/numeric_expr.py
jsiirola: `None` means "not polynomial".  Division expressions with non-constant denominators are not polynomial.

pyomo/core/expr/numeric_expr.py
jsiirola: It's a typo.

pyomo/core/expr/numeric_expr.py
jsiirola: Yes. that appears to be correct.

pyomo/core/expr/numeric_expr.py
jsiirola: I take it back.  That seems to indicate a code block.  It is at least consistent with the other docstrings in that file, and were added as part of the original Pyomo5 expression rewrite.

carldlaird:
@jsiirola Do you know about any performance issues with this change? Did this slow things down at all?

jsiirola:
Not that I have measured.  In general, expression trees are simpler, so my expectation is that things will not be any slower.

pyomo/repn/plugins/ampl/ampl_.py
```diff
@@ -87,12 +87,14 @@ def _build_op_template():
...
     _op_comment[EXPR.ProductExpression] = prod_comment
+    _op_template[EXPR.DivisionExpression] = div_template
+    _op_comment[EXPR.DivisionExpression] = div_comment
     _op_template[EXPR.ReciprocalExpression] = div_template
     _op_comment[EXPR.ReciprocalExpression] = div_comment
-    del prod_template
-    del prod_comment
-    del div_template
-    del div_comment
+    #del prod_template
```
blnicho: Why were these commented out?

pyomo/repn/plugins/baron_writer.py
```diff
@@ -70,6 +111,7 @@ def visit(self, node, values):
                 if val and val[0] in '-+':
                     parens = True
                 elif arg.__class__ in native_numeric_types:
+                    #assert type(val) is str
```
blnicho: Is this a debugging statement that can be removed?

jsiirola: I am not even sure that it is debugging (I am pretty sure that assertion would always fail).  It can definitely be deleted.

pyomo/repn/plugins/ampl/ampl_.py
jsiirola: I commented them out because I could not figure out why we were going to the effort of deleting them: I can't see how they would lead to memory leaks, nor why keeping a string template hanging around would impact how our code ran.

qtothec:
Had a look at `satsolver` and `mcpp`. They appear to look fine.

PR #994, opened 2019-05-02
FBBT Improvements

### LINKED ISSUE #984: Request: sqrt support in FBBT
@michaelbynum Would it be very difficult to also support `sqrt()` in FBBT?

Right now it's giving:
```
WARNING: Unsupported expression type for FBBT: <class
    'pyomo.core.expr.numeric_expr.UnaryFunctionExpression'>. Bounds will not
    be improved in this part of the tree.
```

michaelbynum:
- Adds support for ```sqrt``` in FBBT.
- Adds some infeasibility detection to FBBT.

### Legal Acknowledgement


pyomo/common/errors.py
```diff
@@ -25,3 +25,5 @@ def __str__(self):
                  % ( repr(self.parameter), ) )
 
 
+class InfeasibleConstraintException(Exception):
```
qtothec: Should we have documentation to the effect of "This exception should be raised by Pyomo transformations to indicate that an infeasible constraint has been identified."?

qtothec:
Other than the documentation request, I think this looks good, pending successful test results, as long as other `Transformation` developers are fine with the introduction of the `InfeasibleConstraintException`.

pyomo/common/errors.py
```diff
@@ -25,3 +25,10 @@ def __str__(self):
...
```
jsiirola: As a basic design question, should we make a general class that all Pyomo exceptions inherit off of?  Something like:
```python
class PyomoException(Exception):
    pass
```

pyomo/contrib/fbbt/fbbt.py
```diff
@@ -765,7 +803,7 @@ def visiting_potential_leaf(self, node):
         return False, None
 
 
-def fbbt_con(con, deactivate_satisfied_constraints=False, integer_tol=1e-4):
+def fbbt_con(con, deactivate_satisfied_constraints=False, integer_tol=1e-5, infeasible_tol=1e-8):
```
jsiirola: Are fbbt_con / fbbt_block always/only called by fbbt?  If that is the case, would you consider removing the default values for the arguments here as a defensive programming step to guarantee that the caller always remembers to pass in the values from the calling fbbt()?

qtothec: I know that I call `fbbt_block()` from another place.

pyomo/common/errors.py
michaelbynum: @jsiirola I think this is a great suggestion. This would enable packages that wrap Pyomo to catch all pyomo exceptions. I will make this change.

michaelbynum: I am putting this in another PR which will be ready shortly.

pyomo/contrib/fbbt/fbbt.py
michaelbynum: This is a good question/point. I could probably do that and make fbbt_con and fbbt_block private. I'm not sure there is a need for these to be public? @qtothec Is there any reason you could not use ```fbbt()``` instead of ```fbbt_block()```? They do the same thing as long as you are passing them a block and not a constraint. If we do make this change, I will do so in another PR.

qtothec: If they do the same thing when passed a block, then there's no reason, other than that I decided to use the more specific name.

michaelbynum: I will also put this in another PR.

Issue #326, opened 2018-01-30
[PEP] Redesign Pyomo Set component


jsiirola:
## Proposal

This PEP proposes a partial redesign of the Set component.  The goal is to be API-compatible with the current Set components, while resolving a large number of bug reports.  The draft redesign is currently available on [jsiirola/set-rewrite](https://github.com/jsiirola/pyomo/tree/set-rewrite)

## Motivation

We have started accumulating a number of documented issues with the current implementation of `Set` components.  Known `Set` issues:

- [ ] #1050
- [ ] #995
- [x] #938 
- [x] #835
- [x] #758
- [x] #637 
- [x] #358 
- [x] #325
- [x] #305 
- [x] #298 
- [x] #191 
- [x] #165 
- [x] #148 
- [x] #142 
- [x] #134 
- [x] #121 
- [x] #116 
- [x] #43

## Discussion

The current Set design has a number of bugs and limitations.  Some particularly problematic issues include:

- Set does not follow the typical Pyomo Component class hierarchy.
- IndexedSet members cannot by used in the same ways as scalar Set objects, as certain methods were implemented on the Set object and not the _SetData object
- Set operators do not preserve flags like ordered or sorted
- Users are responsible for knowing how to (efficiently, reliably) sort/order set members
- Support for non-discrete sets is ad hoc, and there is no support for semi-continuous ranges

### Range objects

This redesign first proposes adding full support for discrete and continuous ranges.  This is done through four fundamental "range" objects:

- `NumericRange` for representing a contiguous continuous (closed or open) or discrete (closed) range
- `NonNumericRange` for representing a single discrete non-numeric value
- `AnyRange` for representing the special "Any" range
- `RangeProduct` for representing cross products of ranges

Note that range objects are NOT `Set` objects, although they implement a subset of a `Set`-like API:

- `__contains__`: return True if a value is within the range
- `__eq__`: return True if two ranges are equal
- `isdiscrete()`: return True if the range is over discrete values
- `isfinite()`: return True if the range is over a finite set of discrete values (`isfinite() == True` implies `isdiscrete() == True`)
- `isdisjoint(other)`: return True if the range is completely disjoint from the `other` range.
- `issubset(other)`: return True if the range is contained within (equal to or strict subset of) the `other` range.
- `range_difference(other_ranges)`: return a list of ranges resulting from subtracting all of the `other_ranges` from this range
[...]

PR #996, opened 2019-05-02
Replace domain comparison by type(domain) comparison

### LINKED ISSUE #995: TranformationFactory('core.fix_discrete') doesn't work for cloned models
# Summary
I have noticed that often times `TransformationFactory('core.fix_discrete')` doesn't work for a model, i.e. it doesn't actually fix the variables. I have found out that it happens to cloned models (whereas 'normal' models work as expected).

## Minimal working example
```python
from pyomo.environ import *
m = ConcreteModel()
m.y = Var(domain=Integers, bounds=(0,10), initialize=0)
m_clone = m.clone()
TransformationFactory('core.fix_discrete').apply_to(m_clone)
m_clone.pprint()
```
yields
```
1 Var Declarations
    y : Size=1, Index=None
        None :     0 :     0 :    10 : False : False : Integers
1 Suffix Declarations
    _fixed_discrete_vars : Direction=Suffix.LOCAL, Datatype=Suffix.FLOAT
        Key  : Value
        None :    []
2 Declarations: y _fixed_discrete_vars
```
Notice that the variable `y` didn't actually get fixed. If the `TranformationFactory` is intead applied to model `m`, the behaviour is as expected.

## Looking into the code
When diving into the code, the problem ends up being when the domain of the variable `y` is checked against a list of domains [here](https://github.com/Pyomo/pyomo/blob/ea161615a67b93c2540384eae4cce55e998dc8cc/pyomo/core/plugins/transform/discrete_vars.py#L116-L119)

The expression `var.domain in _discrete_relaxation_map` will yield `False`, even though `var` is an Integer.

RomeoV:
Fixes #995
Note that this bug is blocking development for #983

## Summary/Motivation
Variables of a cloned models don't get fixed by the Transformation (due to an ID copy issue). See the assigned issue for more information.

- change `domain` comparison to `type(domain)` comparison

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/plugins/transform/discrete_vars.py
+++ b/pyomo/core/plugins/transform/discrete_vars.py
@@ -95,26 +95,29 @@ class RelaxDiscreteVars(Transformation):
 #
 # This transformation fixes known discrete domains to their current values
 #
-@TransformationFactory.register('core.fix_discrete', 
-           doc="Fix known discrete domains to continuous counterparts" )
+@TransformationFactory.register('core.fix_discrete',
+           doc="Fix known discrete domains to continuous counterparts")
 class FixDiscreteVars(Transformation):
 
     def __init__(self):
         super(FixDiscreteVars, self).__init__()
 
-    def _apply_to(self, model, **kwds): 
+    def _apply_to(self, model, **kwds):
         options = kwds.pop('options', {})
         if kwds.get('undo', options.get('undo', False)):
             for v in model._fixed_discrete_vars[None]:
                 v.unfix()
             model.del_component("_fixed_discrete_vars")
             return
-        
+
         fixed_vars = []
         _base_model_vars = model.component_data_objects(
-            Var, active=True, descend_into=True )
+            Var, active=True, descend_into=True)
         for var in _base_model_vars:
-            if any(type(var.domain) is type(domain) for domain in _discrete_relaxation_map) and not var.is_fixed():
+            # Instead of checking against `_discrete_relaxation_map.keys()` 
+            # we just check the item properties to fix #995 
+            # When #326 has been resolved, we can check against the dict-keys again
+            if not var.is_continuous() and not var.is_fixed():
                 fixed_vars.append(var)
                 var.fix()
         model._fixed_discrete_vars = Suffix(direction=Suffix.LOCAL)
--- a/pyomo/core/plugins/transform/discrete_vars.py
+++ b/pyomo/core/plugins/transform/discrete_vars.py
@@ -114,7 +114,7 @@ class FixDiscreteVars(Transformation):
         _base_model_vars = model.component_data_objects(
             Var, active=True, descend_into=True )
         for var in _base_model_vars:
-            if var.domain in _discrete_relaxation_map and not var.is_fixed():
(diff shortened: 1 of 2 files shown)


pyomo/core/plugins/transform/discrete_vars.py
```diff
@@ -114,7 +114,7 @@ def _apply_to(self, model, **kwds):
         _base_model_vars = model.component_data_objects(
             Var, active=True, descend_into=True )
         for var in _base_model_vars:
-            if var.domain in _discrete_relaxation_map and not var.is_fixed():
+            if any(type(var.domain) is type(domain) for domain in _discrete_relaxation_map) and not var.is_fixed():
                 fixed_vars.append(var)
```
jsiirola: I think I would prefer:
```python
for var in _base_model_vars:
    if var.is_fixed():
        continue
    if var.domain not in _discrete_relaxation_map:
        # The following extra check is needed for cloned models until the management 
        # of Set singletons is resolved by #326.
        if all(type(var.domain) is not type(domain) for domain in _discrete_relaxation_map):
            continue
    fixed_vars.append(var)
    var.fix()
```

RomeoV: Since we don't actually use the `_discrete_relaxation_map.values()` (we just compare against the keys), we could also consider the more simple
```python
for var in _base_model_vars:
    # Instead of checking against `_discrete_relaxation_map.keys()` we just check the item properties to fix #995 
    # When #326 has been resolved, we can check against the dict-keys again
    if not var.is_continuous() and not var.is_fixed():
        fixed_vars.append(var)
        var.fix()
```

RomeoV: One thing to note of course is that neither of the proposed solutions fix the `TransformationFactory('core.relax_discrete')` defined in the same file, because that implementation actually needs to look up values from `_discrete_relaxation_map`.
It would of course be nice to fix that too, but without #326 we will probably need a fix that's not super pretty.

jsiirola: I think moving to `is_continuous()` is a nice fix.

As for #326, it is closer than you might think: the bulk of the work is done, and I am just cleaning up set operations and finally getting to the Set initialization.  That said, the PR / PEP review might take a while.  I am hoping that it might make it on master by mid June, but we will see.

RomeoV: Fixed in the `is_continuous()` in 72dbabaf64dfd7e216f8abd25aafe14b970d9c6c

RomeoV: When looking at the tests, I noticed another thing:
The tests only check for `TransformationFactory('core.relax_integrality')` which just transforms all Variables to domain `Reals` (and sets the bounds) [source](https://github.com/Pyomo/pyomo/blob/master/pyomo/core/plugins/transform/relax_integrality.py) with no regard for the original domain. Note that this works also with cloned models, because no type lookup happens.
The method discussed earlier was `TransformationFactory('core.relax_discrete')`. This seems like redundant functionality.

RomeoV: Assigned issue #998

PR #1002, opened 2019-05-06
FBBT Numerics improvements.

### LINKED ISSUE #988: Equal-to-zero tolerance for FBBT?
I am coming across a machine precision issue of having `xl=0` and `xu=-4.440892098500626e-16` as arguments to the following function:

```
def inv(xl, xu):
    if xl <= 0 and xu >= 0:
        return -math.inf, math.inf
    return 1.0/xu, 1.0/xl
```

Should we have a epsilon tolerance for this? @michaelbynum

michaelbynum:
## Summary/Motivation:
- This PR primarily involves making FBBT more robust to floating point arithmetic. 
- This PR also creates a PyomoException for other exceptions to inherit from.

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/errors.py
+++ b/pyomo/common/errors.py
@@ -10,6 +10,11 @@
 
 
 class PyomoException(Exception):
+    """
+    Exception class for other pyomo exceptions to inherit from,
+    allowing pyomo exceptions to be caught in a general way
+    (e.g., in other applications that use Pyomo).
+    """
     pass
 
 
--- a/pyomo/common/errors.py
+++ b/pyomo/common/errors.py
@@ -9,7 +9,11 @@
 #  ___________________________________________________________________________
 
 
-class DeveloperError(NotImplementedError):
+class PyomoException(Exception):
+    pass
+
+
+class DeveloperError(PyomoException, NotImplementedError):
     """
     Exception class used to throw errors that result from Pyomo
     programming errors, rather than user modeling errors (e.g., a
@@ -25,5 +29,5 @@ class DeveloperError(NotImplementedError):
                  % ( repr(self.parameter), ) )
 
 
-class InfeasibleConstraintException(Exception):
+class InfeasibleConstraintException(PyomoException):
     pass
--- a/pyomo/contrib/fbbt/fbbt.py
+++ b/pyomo/contrib/fbbt/fbbt.py
@@ -11,7 +11,7 @@ from pyomo.core.base.constraint import Constraint
 from pyomo.core.base.var import Var
 from pyomo.gdp import Disjunct
 import logging
-from pyomo.common.errors import InfeasibleConstraintException
+from pyomo.common.errors import InfeasibleConstraintException, PyomoException
 from pyomo.common.config import ConfigBlock, ConfigValue, In, NonNegativeFloat, NonNegativeInt
 
 logger = logging.getLogger(__name__)
@@ -24,7 +24,7 @@ if not hasattr(math, 'inf'):
 The purpose of this file is to perform feasibility based bounds 
 tightening. This is a very basic implementation, but it is done 
 directly with pyomo expressions. The only functions that are meant to 
-be used by users are fbbt, _fbbt_con, and _fbbt_block. The first set of 
+be used by users are fbbt and compute_bounds_on_expr. The first set of 
 functions in this file (those with names starting with 
 _prop_bnds_leaf_to_root) are used for propagating bounds from the  
 variables to each node in the expression tree (all the way to the  
@@ -54,7 +54,7 @@ improved on either x or y.
 """
 
 
-class FBBTException(Exception):
+class FBBTException(PyomoException):
     pass
 
 
@@ -418,7 +418,7 @@ def _prop_bnds_root_to_leaf_PowExpression(node, bnds_dict, feasibility_tol):
         ub1 = _ub1
     bnds_dict[arg1] = (lb1, ub1)
 
-    if is_fixed(arg2) and lb2 == ub2:
+    if is_fixed(arg2) and lb2 == ub2:  # No need to tighten the bounds on arg2 if arg2 is fixed
         pass
     else:
         _lb2, _ub2 = interval._inverse_power2(lb0, ub0, lb1, ub1, feasiblity_tol=feasibility_tol)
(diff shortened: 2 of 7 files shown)


pyomo/common/errors.py
```diff
@@ -9,7 +9,11 @@
 #  ___________________________________________________________________________
 
 
-class DeveloperError(NotImplementedError):
+class PyomoException(Exception):
```
qtothec: Docstring please?

pyomo/contrib/fbbt/fbbt.py
```diff
@@ -727,6 +733,8 @@ def visiting_potential_leaf(self, node):
                     lb = -math.inf
                 if ub is None:
                     ub = math.inf
+                if lb - self.infeasibility_tol > ub:
+                    raise InfeasibleConstraintException('Variable has a lower bound which is larger than its upper bound: {0}'.format(str(node)))
             self.bnds_dict[node] = (lb, ub)
             return True, None
```
qtothec: Out of curiosity, I think `LinearExpression` still presents itself as a leaf node (see https://github.com/Pyomo/pyomo/issues/924). Do we have some kind of handling for that situation?

pyomo/contrib/fbbt/fbbt.py
```diff
@@ -845,10 +871,12 @@ def fbbt_con(con, deactivate_satisfied_constraints=False, integer_tol=1e-5, infe
     if not con.active:
         return ComponentMap()
 
+    interval.feasibility_tol = infeasible_tol
```
qtothec: This isn't going to come and bite us later w.r.t. concurrency issues is it?

pyomo/contrib/fbbt/fbbt.py
```diff
@@ -869,7 +897,7 @@ def fbbt_con(con, deactivate_satisfied_constraints=False, integer_tol=1e-5, infe
 
     # check if the constraint is always satisfied
     if deactivate_satisfied_constraints:
-        if lb >= _lb and ub <= _ub:
+        if lb >= _lb - infeasible_tol and ub <= _ub + infeasible_tol:
```
qtothec: Doesn't this make the `infeasible_tol` effectively twice the value that it is?

pyomo/contrib/fbbt/interval.py
```diff
@@ -1,11 +1,18 @@
...
+from pyomo.common.errors import DeveloperError, InfeasibleConstraintException, PyomoException
 
 logger = logging.getLogger(__name__)
 
 
+class IntervalException(PyomoException):
+    pass
+
+
+feasibility_tol = 1e-8
```
qtothec: Why `feasibility_tol` here and `infeasibility_tol` in the other module?

pyomo/contrib/fbbt/interval.py
```diff
@@ -19,13 +26,22 @@ def mul(xl, xu, yl, yu):
...
+        ub = math.inf
+    else:
+        ub = 1.0 / xl
+        lb = 1.0 / xu
+    return lb, ub
 
 
 def div(xl, xu, yl, yu):
-    return mul(xl, xu, *inv(yl, yu))
+    if yl <= feasibility_tol and yu >= -feasibility_tol:
```
qtothec: Comment would be nice: "If denominator can include 0, then quotient is unbounded."

pyomo/contrib/fbbt/interval.py
```diff
@@ -19,13 +26,22 @@ def mul(xl, xu, yl, yu):
 
 
 def inv(xl, xu):
-    if xl <= 0 and xu >= 0:
-        return -math.inf, math.inf
-    return 1.0/xu, 1.0/xl
+    if xl <= feasibility_tol and xu >= -feasibility_tol:
```
qtothec: Comment would be nice: "If denominator can include 0, then quotient is unbounded."

qtothec:
I also just opened a PR against this branch which enables basic handling for `_GeneralExpressionData` and `SimpleExpression` nodes.

pyomo/contrib/fbbt/fbbt.py
michaelbynum: No, ```LinearExpression``` is not currently supported. I will have to add support in another PR.

pyomo/contrib/fbbt/fbbt.py
michaelbynum: I don't think so. This tests the lower bound of the constraint and the upper bound of the constraint independently. It ensures that, even after the constraint is deactivated, the body of the constraint will not violate the lower bound of the constraint by more than ```feasibility_tol```. Similarly, even after the constraint is deactivated, the body of the constraint will not violate the upper bound of the constraint by more than ```feasibility_tol```.

pyomo/contrib/fbbt/interval.py
michaelbynum: fixed

pyomo/contrib/fbbt/interval.py
michaelbynum: done

pyomo/contrib/fbbt/interval.py
michaelbynum: done

pyomo/common/errors.py
michaelbynum: Done

michaelbynum:
@qtothec Assuming tests pass, I think this is ready. Sorry for the delay.

pyomo/contrib/fbbt/fbbt.py
carldlaird: @michaelbynum Have you added an issue about that already?

pyomo/contrib/fbbt/fbbt.py
```diff
@@ -52,11 +55,11 @@
 """
 
 
-class FBBTException(Exception):
+class FBBTException(PyomoException):
     pass
 
 
-def _prop_bnds_leaf_to_root_ProductExpression(node, bnds_dict):
+def _prop_bnds_leaf_to_root_ProductExpression(node, bnds_dict, feasibility_tol):
```
carldlaird: @michaelbynum Can we document the keyword argument "feasibility_tol" or remove the documentation for the arguments?
This is true for most (all?) of the private methods that have this change.

pyomo/contrib/fbbt/fbbt.py
michaelbynum: @carldlaird Yes, there is an existing issue. I just updated it too.

michaelbynum:
@carldlaird Tests are still running, but I believe I have addressed your concerns.

Issue #17, opened 2016-08-06
Using constant Params as an index


blnicho:
In the following code snippet, building the objective function fails because the get item function on an indexed component does not automatically replace a constant Param with its value. 

```
m.tf = Param(initialize=1)
m.t = ContinuousSet(bounds=(0,m.tf))
m.x = Var(m.t)

m.obj = Objective(expr=m.x[m.tf])
```

We could of course take the value of the parameter explicitly in the objective expression but this isn't as elegant and I think most users would expect the above code to work.

ghackebeil:
The issue here is that an immutable singleton Param cannot replace itself with its own value. I suppose the model could do this when you access it, but then what is the point of keeping a Param object around anyway if you can't access it.

One possibility is to have all immutable Param objects be replaced with simple dictionaries or numeric values on the instance constructed by an AbstractModel. I like this because it leads to a nice trade off for abstract vs concrete modellers: Since immutable Params serve no purpose on ConcreteModel (maybe there is something I am forgetting though), Param declarations there could default to mutable; whereas on AbstractModel, Param declarations could default to immutable.

I'm not sure which type of model Bethany is using in the above example, but if it is AbstractModel, then it should lead to an error because the objective is not using a rule (m.x has not been constructed yet). If it is a ConcreteModel, there is no point in making m.tf a Param because it was not declared as mutable (just do m.tf = 1). With the changes I am proposing, if it was AbstractModel and the objective was fixed to use a rule, then the example would work. With a ConcreteModel, it would still be invalid to make m.tf a Param because it is used as an index (and the declaration would now create a mutable Param).

@jsiirola, @blnicho : What do you think about this? Maybe I am overthinking this and there is a one line "fix" to index component that would make this work.

jsiirola:
I think there is a simple fix for this (I talked to Bethany & JP about it before asking her to submit the ticket):  I think we should change the `__getitem__` implementation in `IndexedComponent` (actually, I think it will be in `flatten_tuple`) so that any constant `_NumericValue` objects are evaluated to literals before generating the index.

I like this for a couple reasons: first, it does "the right [expected] thing"(tm) for immutable Params, but without relying on an `isinstance` call.  Second, in that logic we can generate meaningful errors for the case where non-constant things (mutable Params and Vars [fixed or free]) are passed as indexes. That is,
`model.var1[model.var2]`
[...]

PR #1006, opened 2019-05-07
Fix bug looking up a IndexedComponent member using a nonhashable object.

### LINKED ISSUE #1003: Discretization leads to ampl preprocessing error
Hi,
this small example

```
from pyomo.environ import (ConcreteModel, Constraint,
                           Objective, Param,
                           Var, minimize, SolverFactory)

from pyomo.environ import TransformationFactory
from pyomo.dae import ContinuousSet
from pyomo.dae import DerivativeVar
from pyomo.environ import ConstraintList

# model m
m = ConcreteModel()

# Time definitions
m.tf = Param(initialize=1)
m.t = ContinuousSet(bounds=(0, m.tf))

# Variable definition
m.x = Var(m.t, bounds=(0, 10))

# Derivative definition
m.dx = DerivativeVar(m.x, wrt=m.t)

# initial values
def _init(m):
    yield m.x[m.t.first()] == 1


m.init_conditions = ConstraintList(rule=_init)


# Differential Equation 4
def _xdot(m, t):
    if t == m.t.first():
        return Constraint.Skip
    else:
        return m.dx[t] == - (m.x[t])


m.xdotcon = Constraint(m.t, rule=_xdot)

# objective function
m.obj = Objective(expr=-m.x[m.tf], sense=minimize)

# Discretization
discretizer = TransformationFactory('dae.collocation')
discretizer.apply_to(m, nfe=10, ncp=3, scheme='LAGRANGE-RADAU')

# Solving
solver = SolverFactory('ipopt')
results = solver.solve(m, tee=True)
```
leads to a not informative error (KeyError: 2915115411248 after trying to generate the expression in line 993 of ampl_.py) error while trying to solve it with IPOPT. When trying to use other solvers (e.g. BARON or GAMS) it works well, pointing that it may be a bug in the ampl parser. Interesting enough, the error does not appear in installations of Pyomo 5.5, while the GAMS solver fails to transform the model as in Issue #340 . Maybe while solving that issue, the ampl solver was affected.

jsiirola:
## Summary/Motivation:
This fixes a bug where looking up a IndexedComponent member (i.e., with __getitem__) using an index that contains an unhashable object (like an immutable scalar Param) would return a new ComponentData and overwrite the original one. 

- Fix for `IndexedComponent.__getitem__`

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/indexed_component.py
+++ b/pyomo/core/base/indexed_component.py
@@ -318,10 +318,17 @@ You can silence this warning by one of three ways:
         try:
             obj = self._data.get(index, _NotFound)
         except TypeError:
-            obj = _NotFound
-            index = self._processUnhashableIndex(index)
-            if index.__class__ is _IndexedComponent_slice:
-                return index
+            new_index = self._processUnhashableIndex(index)
+            if new_index.__class__ is _IndexedComponent_slice:
+                return new_index
+            # The index could have contained constant but nonhashable
+            # objects (e.g., scalae immutable Params).
+            # _processUnhashableIndex will evaluate those constants, so
+            # if it made any changes to the index, we need to re-check
+            # the _data dict for menbership.
+            if new_index is not index:
+                index = new_index
+                obj = self._data.get(index, _NotFound)
 
         if obj is _NotFound:
             # Not good: we have to defer this import to now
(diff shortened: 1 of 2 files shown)


blnicho:
Before merging this we should update the tests [here](https://github.com/Pyomo/pyomo/blob/master/pyomo/core/tests/unit/test_indexed.py#L96) to check that the `VarData` objects being returned are the same instead of just checking if the variable values are the same.

jsiirola:
I need to update the tests from #17 to actually catch the case that this PR fixes before we merge.

blnicho:
@jsiirola those tests are linked in my above comment for easy access

PR #1009, opened 2019-05-08
Fix compilation with Cython

### LINKED ISSUE #1008: Cython support is broken.
Running `python setup.py install --with-cython` yield an error about `visitor_expr.pyx` not existing.

It turns out that the list of files-to-Cythonize in `setup.py` is not up-to-date.

basilfx:
## Fixes 
#1008 

## Summary/Motivation:
This re-enabled Cython compilation using latest master. But also the latest PyPI version is broken.

- Remove the file-to-Cythonize that does not exist.
- Silence warnings about language version.

### Legal Acknowledgement

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -90,6 +90,7 @@ if using_cython:
             "pyomo/core/expr/numvalue.pyx",
             "pyomo/core/expr/numeric_expr.pyx",
             "pyomo/core/expr/logical_expr.pyx",
+            #"pyomo/core/expr/visitor.pyx",
             "pyomo/core/util.pyx",
             "pyomo/repn/standard_repn.pyx",
             "pyomo/repn/plugins/cpxlp.pyx",
--- a/setup.py
+++ b/setup.py
@@ -99,7 +99,8 @@ if using_cython:
         ]
         for f in files:
             shutil.copyfile(f[:-1], f)
-        ext_modules = cythonize(files)
+        ext_modules = cythonize(files, compiler_directives={
+            "language_level": 3 if sys.version_info >= (3, ) else 2})
     except:
         if using_cython == CYTHON_REQUIRED:
             print("""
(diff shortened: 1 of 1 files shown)


basilfx:
The missing file was (I guess) renamed from `visitor_expr.py` to `visitor.py`. However, this fails due to missing classes, which cannot be imported due to cyclic imports.

One way to fix this, is to provide `core/expr/visitor.pxd` as follows:

```Cython
from .numberic_expr import (
    clone_counter
    LinearExpression, 
    MonomialTermExpression,
    NPV_expression_types
)
```

But I can understand this makes the code base too opinionated.

setup.py
```diff
@@ -90,7 +90,6 @@ def get_version():
             "pyomo/core/expr/numvalue.pyx",
             "pyomo/core/expr/numeric_expr.pyx",
             "pyomo/core/expr/logical_expr.pyx",
-            "pyomo/core/expr/visitor_expr.pyx",
```
jsiirola: This should not be removed, but instead be changed to `"pyomo/core/expr/visitor.pyx"`

setup.py
```diff
@@ -100,7 +99,8 @@ def get_version():
         ]
         for f in files:
             shutil.copyfile(f[:-1], f)
-        ext_modules = cythonize(files)
+        ext_modules = cythonize(files, compiler_directives={
+            "language_level": 3 if sys.version_info >= (3, ) else 2})
```
jsiirola: This is a good fix.

setup.py
basilfx: Please see my second comment.

jsiirola: Doh.  You are, of course, correct.  We should probably leave it in as a comment, though: the visitor file has all the routines for walking expression trees and a good chunk of the benefit of cython will come from cythonizing that file, so we should try and get the circular dependence resolved soon.

ghackebeil:
Can we add a build to Travis that actually tests this code? I'd be okay with just testing it with Python 3.7. Maybe even running a more lightweight set of tests (e.g., 'smoke').

ghackebeil:
Okay. That's fine with me.

PR #1011, opened 2019-05-08
Add read-only view of kernel matrix_constraint.A

ghackebeil:
Adds an `A` property method to kernel matrix constraint, along with unit tests.

### Legal Acknowledgement


blnicho:
@ghackebeil looks like there is an issue with the test you added and the older version of Scipy used on Appveyor (0.19.1).

PR #1043, opened 2019-05-30
IndexedBlock pprint method

### LINKED ISSUE #402: Allow pprint() for _ComponentData
Right now you can `component.pprint()`, but it is not possible to `component[idx].pprint()`. A lot of users try to do the latter at some point in time, and get the error message that it is not supported.

michaelbynum:
## Summary/Motivation:
Currently, the ```IndexedBlock``` ```pprint``` method uses ```_BlockData.pprint(b)``` for its children instead of ```b.pprint()```. For classes that inherit off of ```_BlockData``` and override the ```pprint``` method, the wrong ```pprint``` method gets called for the children of the ```Indexed``` version of the inherited class (unless the ```Indexed``` version of the inherited class also overrides the ```pprint``` method). 

This PR corrects this through a significant rewrite of the way `pprint()` interacts with derived classes.  In particular, it introduces two new private methods (in `Block` and `_BlockData`) that derived `Block`s can override in lieu of pverriding `pprint()` itself.

This also adds a `pprint()` method to `ComponentData`.

- Overhaul how derived classes interact with `pprint()`.  It is no longer necessary (or advised) for derived classes to override `pprint()`
- Added support for calling `pprint()` on all `ComponentData` objects

### Legal Acknowledgement


ghackebeil:
I probably can't comment on whether or not this change is correct, but I think it should probably come with a new unit test if so.

carldlaird:
I recommend checking with @jsiirola to make sure this isn't hardcoded because of a lacking super() call in the diamond hierarchy (or something)...

michaelbynum:
I agree. This should definitely not be merged without the approval of @jsiirola.

jsiirola:
I have taken a look at this, and along with @michaelbynum, "fixed" it by redesigning how `pprint()` interacts with derived classes through `_pprint()`.  I think this is ready for review.

A follow-on PR should propagate this logic over to the `display()` method.

pyomo/core/base/component.py
```diff
@@ -222,6 +225,30 @@ def cname(self, *args, **kwds):
 context of the immediate parent container.""")
         return self.getname(*args, **kwds)
 
+    def pprint(self, ostream=None, verbose=False, prefix=""):
+        """Print component information
+
+        Note that this method is generally only reachable through
+        ComponentData objects in an IndexedComponent container.
+        Components, including unindexed Component dervatives and both
```
blnicho: Typo: dervatives

blnicho:
After taking a second look at this, I don't think @ghackebeil's initial comment was addressed. As the scope of this PR was expanded to include a more significant overhaul of the pprint infrastructure (including a new feature) I think it's really important that this PR include a few tests of the updated functionality.

michaelbynum:
Good point. I can work on some tests late this week or early next week.

michaelbynum:
I added a couple tests for the improved functionality of `pprint`. If tests pass, this PR should be ready to go.

PR #1058, opened 2019-06-13
enabled nl interface

merraksh:
## Summary/Motivation:
Solving nonquadratic nonlinear problems or quadratic nonconvex problems with Pyomo using the Xpress shell interface calls amplxpress on a .nl file, but amplxpress does not handle the above classes of problems. This PR is to enable a user to define an Xpress with solver_io='nl' and solve these problems using the console in the same way that it is done with .mps files: using readprob (which works with .nl files) and running a solve command.

- switching from 'asl' to '_xpress_shell' when the solver_io is chosen as 'nl'
- enabling ProblemFormat.nl as a valid problem format for the Xpress shell solver.

### Legal Acknowledgement


michaelbynum:
These changes look reasonable to me. Do we have any tests that exercise this change?

ghackebeil:
No. I don’t think we are able to obtain a test license. Or we just haven’t set one up on any test servers.

I’d like to take a closer look at this if you don’t mind holding off on a merge.

michaelbynum:
@ghackebeil No problem at all.

qtothec:
I'm working with Xpress fairly regularly now. You should be able to solve small problems using the community license when downloading standalone Xpress.

https://community.fico.com/s/fico-xpress-optimization-licensing-optio

blnicho:
@ghackebeil any updates on this PR?

merraksh:
Just added a few more changes for correct retrieval of solution values. Now with ProblemFormat.nl it writes the solution in AMPL format as otherwise it's unable to get the right values.

mrmundt:
The preferred route for XPRESS is through the direct interface. As a result, we are closing this PR.

PR #1059, opened 2019-06-17
Taylor series; Common interface for differentiation

### LINKED ISSUE #840: Multiple methods for differentiation in different locations
There are multiple functions for differentiation (pyomo/core/base/symbolic.py and pyomo/contrib/derivatives/differentiate.py). Consider putting these in the same location and possibly adding a wrapper function.

michaelbynum:
## Changes:
1. This PR creates a common interface for the various methods of differentiation we have in Pyomo right now. It also moves some code around. I moved both ```pyomo.contrib.derivatives``` and ```pyomo.core.base.symbolic``` to ```pyomo.core.expr.differentiation```. I left a deprecated ```differentiate``` function in ```pyomo.core.base.symbolic```. These functions do not require any Pyomo components - just the expression system. Therefore, I thought ```pyomo.core.expr``` was the right place to move them. Other ideas are, of course, welcome. Finally, I elevated the new ```differentiate``` method to pyomo.environ.

2. This PR adds a function for generating a taylor series approximation of an expression.

### Legal Acknowledgement


pyomo/core/base/symbolic.py
```diff
@@ -8,123 +8,15 @@
...
-        self.pyomo2sympy[pyomo_object] = sympy_obj
-        self.sympy2pyomo[sympy_obj] = pyomo_object
-        return sympy_obj
-
-    def sympyVars(self):
-        return iterkeys(self.sympy2pyomo)
 
+@deprecated(msg=('The differentiate function in pyomo.core.base.symbolic has been deprecated. Please use the ' +
+                 'differentiate function in pyomo.core.expr.differentiate.differentiate'),
+            version='5.6.6',
```
jsiirola: The version should be "TBD".  we will set it to the next release as part of the Pyomo release process.

pyomo/core/expr/differentiate/diff_with_sympy.py
```diff
@@ -0,0 +1,299 @@
...
+
+    def beforeChild(self, node, child):
+        if not child._args:
+            item = self.object_map.getPyomoSymbol(child, None)
+            if item is None:
+                item = float(child.evalf())
+            return False, item
+        return True, None
+
+def sympify_expression(expr):
```
jsiirola: Not a fault of this PR, but I think I should have called this "sympyify_expression" or "pyomo2sympy".

pyomo/core/expr/differentiate/diff_with_sympy.py
```diff
@@ -0,0 +1,299 @@
...
+    #
+    # Return the answer
+    #
+    return ans if wrt is None else ans[0]
+
+
+# =====================================================
+# sympify_expression
+# =====================================================
+
```
jsiirola: As discussed with @michaelbynum, the Pyomo/SymPy interface is more general than just differentiation and should be moved into `pyomo.core.expr.sympy`.

pyomo/core/expr/differentiate/differentiate.py
```diff
@@ -0,0 +1,80 @@
+from .diff_with_sympy import differentiate as sympy_diff
+from .diff_with_pyomo import reverse_sd, reverse_ad
+from pyutilib.enum import Enum
+from pyomo.core.kernel.component_map import ComponentMap
+
+
+DiffModes = Enum('sympy', 'reverse_symbolic', 'reverse_numeric')
```
jsiirola: I think I would call this just "`Modes`".  To me, `DiffModes` is an awkward abbreviation (I read it as difference) and is redundant (the namespace `differentiate.Modes` makes it clear what the modes refer to).

pyomo/core/expr/taylor_series/taylor.py
```diff
@@ -0,0 +1,22 @@
+from pyomo.core.expr.current import identify_variables, value
```
jsiirola: I would advocate moving `taylor.py` to `pyomo/core/expr/taylor_series.py` and git rid f the extra directory.

pyomo/environ/__init__.py
```diff
@@ -106,4 +106,5 @@ def _import_packages():
     TerminationCondition, SolverStatus,
     )
 from pyomo.core.base.units_container import units
+from pyomo.core.expr.differentiate.differentiate import differentiate, DiffModes
```
jsiirola: What concerns to folks have around "polluting" the `pyomo.environ` namespace?  What about:
```python
import pyomo.core.expr.calculus as calculus
```
Then users would 
```python
pyo.calculus.differentiate(my_expr, wrt=my_var, mode=pyo.calculus.differentiate.Modes.sympy
```

michaelbynum: Maybe we should leave this out of environ for now. We can always add it later. I have added the differentiate function to ```pyomo.core.expr```, so imports should not be bad:

```
from pyomo.core.expr import differentiate
```

michaelbynum:
@jsiirola Tests are still running, but I think I addressed most of your concerns.

pyomo/common/errors.py
```diff
@@ -41,3 +41,8 @@ class InfeasibleConstraintException(PyomoException):
     the course of range reduction).
     """
     pass
+
+
+class NondifferentiableError(ValueError):
```
jsiirola: This should also inherit from `PyomoException`, right?

pyomo/core/expr/taylor_series.py
```diff
@@ -0,0 +1,22 @@
+from pyomo.core.expr.current import identify_variables, value
+from pyomo.core.expr.calculus.derivatives import differentiate
+
+
+def get_taylor_series(expr, diff_mode=differentiate.Modes.reverse_numeric):
```
jsiirola: Why `get_taylor_series`?  We are not really retrieving a Taylor series (I generally think of `get_` as methods on a class).  What do you think about something like `Taylor_series_expansion()`?

Also, this method is only for first-order Taylor series... would you consider adding an `order=n` option for getting higher order expansions?  (Of course, the default order would be 1, and the mode would be restricted to symbolic derivatives for higher orders...)

That would make a user's experience something like:

```python
model.c = Constraint(expr=model.y == Taylor_series_expansion(model.x**3, order=2)
```

pyomo/core/tests/unit/test_taylor_series.py
```diff
@@ -0,0 +1,17 @@
+import pyutilib.th as unittest
+import pyomo.environ as pe
+from pyomo.core.expr.taylor_series import get_taylor_series
+from pyomo.core.expr.current import polynomial_degree
+
+
+class TestDerivs(unittest.TestCase):
```
jsiirola: Should this class name be changed?

pyomo/common/errors.py
michaelbynum: Yes. Good point. ```PyomoException``` did not exist when I created this PR. I'll fix it now.

pyomo/core/expr/taylor_series.py
michaelbynum: Yes. Good suggestion.

michaelbynum:
@jsiirola I think I have addressed your concerns now. Tests are still running though.

pyomo/core/tests/unit/test_taylor_series.py
```diff
@@ -1,17 +1,47 @@
...
+        e = m.x**3 + m.y**3
+        tse = taylor_series_expansion(e, diff_mode=differentiate.Modes.reverse_symbolic, order=3)
+        for _x in [-2, -1.5, -1, -0.5, 0, 0.5, 1, 1.5, 2]:
+            for _y in [-2, -1.5, -1, -0.5, 0, 0.5, 1, 1.5, 2]:
+                m.x.value = _x
+                m.y.value = _y
+                self.assertAlmostEqual(pe.value(e), pe.value(tse))
+
+        e = (m.x*m.y)**2
+        tse = taylor_series_expansion(e, diff_mode=differentiate.Modes.reverse_symbolic, order=4)
```
jsiirola: Does this test make sense?  The 3rd and 4th order terms should be 0...

I think I would be generally more comfortable if these tests ensured that the correct number of terms were being returned and that the individual values in the expansion were correct.

michaelbynum: If the expression were ```(m.x + m.y)**2```, then the 3rd and 4th order terms would be zero. As written, this is a 4th order polynomial. Thus, the 4th order taylor series expansion should be exact, which is essentially what this test checks. In fact, that is what all of these tests are doing. This is why I focused on testing polynomials. With polynomials, we can test that a taylor series of the same order as the polynomial returns the same polynomial. It is not terribly easy to test that the expressions generated are exactly what I expect without writing a custom walker. However, it is very unlikely that the value of the original expression and the value of the taylor series expansion will be identical at all of the points being tested unless the expressions are identical. In my opinion, that makes these very good tests.

Keep in mind that, in the current implemenation, the higher order taylor series expansions of multivariate expressions are very ugly. This is because the implementation is rather naive, and no simplification is done. For example, the 4th order taylor series expansion for 

```(m.x*m.y)**2``` 

is 

```2.25*(x - 0.5) + 0.75*(y - 1.5) + 0.5625 + 0.5*(x - 0.5)*(4.5*(x - 0.5) + 3.0*(y - 1.5)) + 0.5*(y - 1.5)*(3.0*(x - 0.5) + 0.5*(y - 1.5)) + 0.16666666666666666*(x - 0.5)*(x - 0.5)*(6.0*(y - 1.5)) + 0.16666666666666666*(x - 0.5)*(y - 1.5)*(6.0*(x - 0.5) + 2.0*(y - 1.5)) + 0.16666666666666666*(y - 1.5)*(x - 0.5)*(6.0*(x - 0.5) + 2.0*(y - 1.5)) + 0.16666666666666666*(y - 1.5)*(y - 1.5)*(2.0*(x - 0.5)) + 0.041666666666666664*(x - 0.5)*(x - 0.5)*(y - 1.5)*(4*(y - 1.5)) + 0.041666666666666664*(x - 0.5)*(y - 1.5)*(x - 0.5)*(4*(y - 1.5)) + 0.041666666666666664*(x - 0.5)*(y - 1.5)*(y - 1.5)*(4*(x - 0.5)) + 0.041666666666666664*(y - 1.5)*(x - 0.5)*(x - 0.5)*(4*(y - 1.5)) + 0.041666666666666664*(y - 1.5)*(x - 0.5)*(y - 1.5)*(4*(x - 0.5)) + 0.041666666666666664*(y - 1.5)*(y - 1.5)*(x - 0.5)*(4*(x - 0.5))```

 If you simplify it, it does turn out to be ```x**2*y**2```.

**Comment by jsiirola** [comment]
I think this is fine, but as it involves changes to core, I would recommend considering getting an additional set of eyes on it.

File: pyomo/core/base/symbolic.py [hunk 11]
```diff
@@ -8,123 +8,15 @@
...
-        sympy_obj = sympy.Symbol("x%d" % self.i)
-        self.i += 1
-        self.pyomo2sympy[pyomo_object] = sympy_obj
-        self.sympy2pyomo[sympy_obj] = pyomo_object
-        return sympy_obj
-
-    def sympyVars(self):
-        return iterkeys(self.sympy2pyomo)
 
+@deprecated(msg=('The calculus function in pyomo.core.base.symbolic has been deprecated. Please use the ' +
```
blnicho: I think there might be a "find-and-replace" typo in this deprecation message. Should "calculus" be "differentiate"?

michaelbynum: Yes, good catch. Thanks.

PR #1063, opened 2019-06-20
Tweaking the collection of standard_repn

whart222:
There was a line that wasn't being executed.  While working on Coek,
I realized that this line would never be executed.

## Summary/Motivation:
This PR removes a branch that is never executed.  This branch was tested for every variable in every expression of a model, so this will likely improve the runtime performance of Pyomo a modest amount.

- Deleting a branch.

### Legal Acknowledgement


jsiirola:
The proof that the code is unreachable is pretty straightforward: as `ans` is created at the beginning of the function, and the removed `if key in ans.linear` is the first reference to `ans.linear`, it is impossible for `key` to be in `ans.linear`.  As a result, I am comfortable merging this without the customary detailed review that we typically require for changes to core.

PR #1066, opened 2019-06-20
Suppress QT import errors until the UI class is created

jsiirola:
## Summary/Motivation:
This resolves diffs that have appeared in the Book tests.

- Cache Qt import error messages identified during module import
- Log cached import error messages when attempting to instantiate the UI
- Add an additional dummy class so that more modules can be imported without PyQt

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/viewer/qt.py
+++ b/pyomo/contrib/viewer/qt.py
@@ -40,18 +40,24 @@ class DummyQAbstractItemModel(object):
     def __init__(*args, **kwargs):
         pass
 
+class DummyQAbstractTableModel(object):
+    """
+    A dummy QAbstractTableModel class to allow some testing without PyQt
+    """
+    def __init__(*args, **kwargs):
+        pass
+
 qt_available = False
+qt_import_errors = []
 
 try:
     from PyQt5 import QtCore
 except:
-    _log.exception("Cannot import PyQt5.QtCore")
+    qt_import_errors.append("Cannot import PyQt5.QtCore")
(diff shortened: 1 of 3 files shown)


pyomo/contrib/viewer/qt.py
```diff
@@ -40,18 +40,24 @@ class DummyQAbstractItemModel(object):
...
 except:
-    _log.exception("Cannot import PyQt5.QtCore")
+    qt_import_errors.append("Cannot import PyQt5.QtCore")
     try:
         from PyQt4 import QtCore
     except:
-        _log.exception("Cannot import PyQt4.QtCore")
-        QAbstractItemModel = DummyQAbstractItemModel
-        QtCore = DummyQtCore
+        qt_import_errors.append("Cannot import PyQt4.QtCore")
```
carldlaird: Don't quite understand why we capture the errors to report later instead of delaying the import until it is needed? Can you clarify?

jsiirola: I was working to minimally change @eslickj's code.

I think his motivation was to enable testing of a good chunk of the viewer interface without PyQt. So some of the framework (including the Qt wrappers) need to be imported during testing. It became an issue for us when we asked to move the executable that drives the UI out of console_scripts and into a Pyomo subcommand.  This causes Pyomo to have to import more of the viewer than it used to when users run the pyomo command, so pyqt import error messages were cropping up (e.g., in unrelated tests).

PR #1071, opened 2019-06-21
Model Viewer, fix bug in sorting residuals

eslickj:
## Summary/Motivation:

The main purpose of this PR is to fix a bug I came across in the residual table where if you calculate the constraints then add a new constraint to the model, the new constraint won't be calculated yet and the sort function will try to sort a list that contains floats and None, and fail.  I fixed this so if a residual value is None the sort treats it as -inf, making those constraints show up at the end.

Since I had to make some changes, I went ahead and made a few other minor improvements to the stand-alone model viewer.  I added a menu bar to the qtconsole window, and put menu items to run a script, change working directory, exit, show an hide the gui.  I also made only show the qtconsole window on start up.  You need to launch the model viewer by going to view and selecting show model viewer.  That fixed two problems 1) the model viewer no longer obscures the qtconsole window so people won't be confused about how to use it, and 2) I don't need to wait a few seconds to start the model browser UI in the jupyter kernel. 

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/viewer/README.md
+++ b/pyomo/contrib/viewer/README.md
@@ -58,9 +58,11 @@ The viewer will start with an empty Pyomo ConcreteModel called ```model```. The
 advantage of the stand-alone viewer is that it will automatically set up the
 environment and start the UI, saving typing a few lines of code. In the kernel,
 ``pyomo.environ`` is imported as ```pyo```. An empty ConcreteModel is available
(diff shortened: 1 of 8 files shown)


eslickj:
I removed the file menu and exit from the model viewer menu bar, but forgot to remove the test for it.  It should be fixed.

PR #1076, opened 2019-06-28
Fbbt log10 support

michaelbynum:
## Summary/Motivation:
This PR adds support for log10 in FBBT.

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/fbbt/interval.py
+++ b/pyomo/contrib/fbbt/interval.py
@@ -296,6 +296,18 @@ def log(xl, xu):
     return lb, ub
 
 
+def log10(xl, xu):
+    if xl > 0:
+        lb = math.log10(xl)
+    else:
+        lb = -math.inf
+    if xu > 0:
+        ub = math.log10(xu)
+    else:
+        ub = -math.inf
+    return lb, ub
+
+
 def sin(xl, xu):
     """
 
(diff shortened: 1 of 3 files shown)


pyomo/contrib/fbbt/interval.py
```diff
@@ -296,6 +296,18 @@ def log(xl, xu):
...
 
+def log10(xl, xu):
+    if xl > 0:
+        lb = math.log10(xl)
+    else:
+        lb = -math.inf
+    if xu > 0:
+        ub = math.log10(xu)
+    else:
+        ub = -math.inf
```
jsiirola: I am not sure how this got past tests, but `math.inf` was only introduced in Python 3.5.  Prior to that you needed to use `float('inf')`.  Frustratingly, `math.inf is float('inf') == False` in Python 3.x (although `math.inf == float('inf')` is True).

My recommendation would be to standardize on `float('inf')` throughout interval.py by declaring a module-level variable `inf = float('inf')` so that we can easily change it later.

michaelbynum: @jsiirola Good suggestion. This did not get past tests. At the top of ```fbbt.py```, I assign ```float('inf')``` to ```math.inf```  if ```math``` does not have an attribute ```inf```. I am also careful not to compare against ```math.inf``` with ```is```. That being said, your solution makes a lot more sense. I will create an issue for this and open a separate PR since this is not really related to support for ```log10```.

PR #1085, opened 2019-07-16
MPEC descend into disjuncts

emma58:
## Summary/Motivation: 
All four mpec transformations now descend into Disjuncts and transform Complementarity objects on them. The reason for this is that if you have a complementarity constraint on a disjunct, you must transform it before you can transform the disjunct, so it should get picked up by the mpec transformation.

- mpec transformations descend into Disjuncts to find and transform Complementarity components
- mpec.nl searches on the entire model for free variables rather than just on its argument. Otherwise you would get different results if you called the transformation on a model versus on a block.

### Legal Acknowledgement


emma58:
I don't think transformation order matters because all the transformations are done in place. But that is really a question for @whart222.

PR #1097, opened 2019-08-07
Changed constructor of LinearExpression for easier user creation

carldlaird:
## Summary/Motivation:
It is possible to create a LinearExpression object directly in constraints. This might be desirable if this is faster than using sum() (which it appears it is - at least for dense constraints). This PR makes the constructor of LinearExpression a little easier to work with for users.

** still need to add a test - but I want to run Travis first.

-
-

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/expr/numeric_expr.py
+++ b/pyomo/core/expr/numeric_expr.py
@@ -1281,7 +1281,8 @@ class LinearExpression(ExpressionBase):
         followed by the coefficients, followed by the variables.
         
         Alternatively, you can specify the constant, the list of linear_coeffs
-        and the list of linear_vars separately.
+        and the list of linear_vars separately. Note that these lists are NOT
+        copied.
         """
         # I am not sure why LinearExpression allows omitting args, but
         # it does.  If they are provided, they should be the constant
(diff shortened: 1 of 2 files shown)


pyomo/core/expr/numeric_expr.py
```diff
@@ -1282,8 +1293,15 @@ def __init__(self, args=None):
...
         else:
             self.constant = 0
+            if constant is not None:
+                self.constant = constant
             self.linear_coefs = []
+            if linear_coefs is not None:
+                self.linear_coefs = linear_coefs
             self.linear_vars = []
+            if linear_vars is not None:
+                self.linear_vars = linear_vars
```
jsiirola: For efficiency (to avoid making and disposing of an empty list), I would make these all along the lines of:
```
if linear_vars is None:
    self.linear_vars = []
else:
    self.linear_vars = linear_vars
```
Also, are there any concerns around `linear_vars` and `linear_ceofs` arguments being shared with whomever constructed this object, or is the API contract that the caller is responsible for making sure the lists aren't shared?

DLWoodruff:
Looks good.

Issue #134, opened 2017-04-12
Multiplying Virtual Sets causes the original Set to be modified


jsiirola:
Multiplying a Virtual Set by itself produces the desired resulting set, but also causes the original Set to become crossed with itself.

In the following example, note that the Virtual Set `m.IJ` goes from 9 to 81 elements, and that the members are 4-tuples instead of 2-tuples (even though the Set `Dimen` flag is not updated).

```python
>>> from pyomo.environ import *
>>> m = ConcreteModel()
>>> m.I = Set(initialize=[1,2,3])
>>> m.J = Set(initialize=[4,5,6])
>>> m.IJ = m.I * m.J
>>> m.IJ.display()
IJ : Dim=0, Dimen=2, Size=9, Domain=None, Ordered=False, Bounds=None
    Virtual
>>> m.IJ2 = m.IJ * m.IJ
>>> m.IJ.display()
IJ : Dim=0, Dimen=2, Size=81, Domain=None, Ordered=False, Bounds=None
    Virtual
>>> m.IJ2.display()
IJ2 : Dim=0, Dimen=4, Size=81, Domain=None, Ordered=False, Bounds=None
    Virtual
>>> [x for x in m.IJ]
[...]

PR #1101, opened 2019-08-15
bug in bigm transformation with set objects

michaelbynum:
## Summary/Motivation:
When creating a constraint using the index set of another constraint combined with another set, the index set of the original constraint gets modified unintentionally. See the following example:

```
>>> import pyomo.environ as pe
>>> m = pe.ConcreteModel()
>>> m.a = pe.Set(initialize=[1,2,3])
>>> m.b = pe.Set(initialize=['a', 'b'])
>>> m.c = pe.Constraint(m.a, m.b)
>>> m.c.index_set().pprint()
c_index : Dim=0, Dimen=2, Size=6, Domain=None, Ordered=False, Bounds=None
    Virtual
>>> for i in m.c.index_set():
...     print(i)
...
(1, 'a')
(1, 'b')
(2, 'a')
(2, 'b')
(3, 'a')
(3, 'b')
>>> m.c2 = pe.Constraint(m.c.index_set(), m.b)
>>> m.c.index_set().pprint()
c_index : Dim=0, Dimen=2, Size=12, Domain=None, Ordered=False, Bounds=None
    Virtual
>>> for i in m.c.index_set():
...     print(i)
...
(1, 'a', 'a')
(1, 'a', 'b')
(1, 'b', 'a')
(1, 'b', 'b')
(2, 'a', 'a')
(2, 'a', 'b')
(2, 'b', 'a')
(2, 'b', 'b')
(3, 'a', 'a')
(3, 'a', 'b')
(3, 'b', 'a')
(3, 'b', 'b')
```

This creates a bug in the BigM transformation in ```pyomo.gdp```. This PR fixes this problem by wrapping the original index set in a ```list()``` when creating the new constraint. 

This is may not be the correct fix. We may need a fix in the Constraint constructor or somewhere in Sets. However, the new Set rewrite may fix this problem. @jsiirola What is your suggestion on how to proceed? Is this the correct fix until the set rewrite is complete or should I go ahead and try to fix the underlying problem?

### Legal Acknowledgement

### DIFF
--- a/pyomo/gdp/plugins/bigm.py
+++ b/pyomo/gdp/plugins/bigm.py
@@ -513,7 +513,7 @@ class BigM_Transformation(Transformation):
             try:
                 # The reason list(obj.index_set()) is used instead of just obj.index_set()
                 # is that the underlying index set for obj gets modified in this step
-                # if list() is not used. See PR #TBD
+                # if list() is not used. See PR #1101
                 newConstraint = Constraint(list(obj.index_set()), transBlock.lbub)
             except TypeError:
                 # The original constraint may have been indexed by a
--- a/pyomo/gdp/plugins/bigm.py
+++ b/pyomo/gdp/plugins/bigm.py
@@ -511,7 +511,10 @@ class BigM_Transformation(Transformation):
 
         if obj.is_indexed():
             try:
-                newConstraint = Constraint(obj.index_set(), transBlock.lbub)
+                # The reason list(obj.index_set()) is used instead of just obj.index_set()
+                # is that the underlying index set for obj gets modified in this step
+                # if list() is not used. See PR #TBD
+                newConstraint = Constraint(list(obj.index_set()), transBlock.lbub)
             except TypeError:
                 # The original constraint may have been indexed by a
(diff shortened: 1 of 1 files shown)


jsiirola:
This is a duplicate of #134.  The set-rewrite (#326) does indeed fix this bug.  While this technically works around #134, the better solution is to fix _SetProduct - which now that I look into it, I don't think is a bad patch:
```diff
diff --git a/pyomo/core/base/sets.py b/pyomo/core/base/sets.py
index 4e27e98..1fe17af 100644
--- a/pyomo/core/base/sets.py
+++ b/pyomo/core/base/sets.py
@@ -1545,7 +1545,7 @@ class _SetProduct(_SetOperator):
         _SetOperator.__init__(self, *args, **kwd)
         # the individual index sets definining the product set.
         if isinstance(self._setA,_SetProduct):
-            self.set_tuple = self._setA.set_tuple
+            self.set_tuple = list(self._setA.set_tuple)
         else:
             self.set_tuple = [self._setA]
         if isinstance(self._setB,_SetProduct):
```

PR #1102, opened 2019-08-15
avoiding math range errors in fbbt

michaelbynum:
## Summary/Motivation:
This PR updates the interval arithmetic for exp to avoid math range errors when performing FBBT. 

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/fbbt/interval.py
+++ b/pyomo/contrib/fbbt/interval.py
@@ -282,14 +282,14 @@ def _inverse_power2(zl, zu, xl, xu, feasiblity_tol):
 
 
 def exp(xl, xu):
-    if xl >= 100:
-        lb = math.inf
-    else:
+    try:
         lb = math.exp(xl)
-    if xu >= 100:
-        ub = math.inf
-    else:
+    except OverflowError:
+        lb = math.inf
+    try:
         ub = math.exp(xu)
+    except OverflowError:
+        ub = math.inf
     return lb, ub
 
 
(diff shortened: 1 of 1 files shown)


pyomo/contrib/fbbt/interval.py
```diff
@@ -282,7 +282,15 @@ def _inverse_power2(zl, zu, xl, xu, feasiblity_tol):
 
 
 def exp(xl, xu):
-    return math.exp(xl), math.exp(xu)
+    if xl >= 100:
+        lb = math.inf
+    else:
+        lb = math.exp(xl)
+    if xu >= 100:
```
jsiirola: Where did 100 come from?  I would prefer to see the constant pulled from Python or with a try-except block.  Is there a similar issue for the lower limits and underflow?

michaelbynum: @jsiirola Good suggestion. I went with the try-except approach. There is not a similar issue for the lower limits and underflow. If the bounds are too small (large negative numbers), then the ```exp``` function from ```math``` will just return 0.

michaelbynum:
@jsiirola Good suggestion. I went with the try-except approach. There is not a similar issue for the lower limits and underflow. If the bounds are too small (large negative numbers), then the ```exp``` function from ```math``` will just return 0.

blnicho:
I don't think the Jenkins test failures are related to the changes in this PR but rather because this branch is a few commits behind Pyomo master. Going to go ahead and merge.

PR #1111, opened 2019-08-27
Rewrite of the Pyomo Set component hierarchy

### LINKED ISSUE #326: [PEP] Redesign Pyomo Set component

jsiirola:
## Summary/Motivation:
This implements the rewrite of the Pyomo Set component as specified in PEP #326

**NOTE:**  This PR does **NOT** switch Pyomo over to using the new Set system.  Everything is implemented in parallel to the existing system.  A second PR will switch Pyomo over to the new system later.

- Rewrite of Set to follow the current Pyomo component object model
- Adding "Range" objects for performing operations on continuous and discrete numeric ranges
- Adding a new approach for processing component initialization options
- Adding a new approach for providing errors for unconstructed (Abstract) component methods.

### Legal Acknowledgement


jsiirola:
This should be ready to discuss/review.  Tests all pass, and coverage is good: overall project coverage is up 0.87% and the new Set code is covered at 100%.

PR #1114, opened 2019-08-30
Baron exponential bugfix

### LINKED ISSUE #1113: Baron .bar writer error
The Baron file writer (**baron_writer.py**) in the expression tree parsing (class ToBaronVisitor, visit() method) seems to be checking "is" type as opposed to "is instance" of a type. Specifically, I have a certain constraint that has an "NPV_PowExpression" object that is being written as a string to the.bar file (meaning, the ** operator for power hasn't been replaced with the "^" Baron requires). I think this is happening at line 102 of the baron_writer.py where type is checked against "is EXPR.PowExpression," since NPV_PowExpression is an instance of PowExpression and not a PowExpression type itself.

qtothec:
## Summary/Motivation:
Making changes proposed by @natalieisenberg in the issue, and adding test.
Note that changes were not made to similar adjacent code.

### Legal Acknowledgement


jsiirola:
I don't think there is anything wrong with this fix, but I have two concerns:
- are we going to have the same problems with the other expression node checks?
- is a series of `isinstance` checks (most of which will return False) going to be a performance issue (`isinstance()` is rather slow when it returns False).  Should we just explicitly check for `PowExpression` and `NPV_PowExpression`?

qtothec:
I understand the performance reasons that we have the current type-checking system, but we are going to run into the problem of having to remember to update all the type-checks everywhere whenever we introduce a new class type, which is what happened with NPV_Expressions, or we may have logic errors. We could move to a "registered handler" type of scheme behind an `isinstance()` to see if it is an `Expression`, which would at least flag unhandled types?

blnicho:
I think NPV stands for "not potentially variable" so I think you should be able to just evaluate any NPV_* expression before writing it out. Couldn't you just call `is_potentially_variable` on the expression to see if it could be evaluated to a numeric constant before writing it? This would handle all of the NPV expression types.

jsiirola:
@blnicho: I was thinking along the same lines.  One of the things that `standard_repn` does is evaluate fixed subexpressions.  That includes not only removing NPV subexpressions, but also any subexpressions involving fixed variables.  The Baron/GAMS writers do not currently do that and instead output the raw expression tree.  I think some form of evaluation would be good, but it should probably check `is_fixed` and not `is_potentially_variable`.  Regardless, because of the walker that those writers use, it might be expensive: both checks themselves require walking the complete subexpression.

jsiirola:
@qtothec:  How about this: can we start the "if tree" in the writer by first checking if `node.__class in pyomo.core.expr.numeric_expr.NPV_expression_types`?  If it returns `True`, then I think the right thing to do is to evaluate the node and return the numeric value.  That way we don't need the `isinstance` check, plus this will fix the fundamental problem for all NPV objects.

Also, we should consider porting this fix to the GAMS writer as well.

Issue #1106, opened 2019-08-23
Block set_value can't assign to index of an IndexedBlock


emma58:
Calling set_value on a Block fails when assigning either a Block or a BlockData to an index in an IndexedBlock.

If you assign a Block:
```
m = ConcreteModel()
m.block_list = Block([1, 2])

movingBlock = m.some_other_block = Block()
m.del_component(movingBlock)

m.block_list[1] = movingBlock
```
```
Traceback (most recent call last):
  File "blockSetValue.py", line 11, in <module>
    m.block_list[1] = movingBlock
  File "/home/esjohn/src/pyomo/pyomo/core/base/indexed_component.py", line 412, in __setitem__
    return self._setitem_when_not_present(index, val)
  File "/home/esjohn/src/pyomo/pyomo/core/base/indexed_component.py", line 675, in _setitem_when_not_present
    obj.set_value(value)
  File "/home/esjohn/src/pyomo/pyomo/core/base/block.py", line 711, in set_value
    self.add_component(k,val[k])
  File "/home/esjohn/src/pyomo/pyomo/core/base/block.py", line 923, in add_component
    super(_BlockData, self).__setattr__(name, val)
TypeError: attribute name must be string, not 'NoneType'
```
If you assign a BlockData:
```
m = ConcreteModel()
m.block_list = Block(Any)

m.some_other_block_list = Block([3,4])
movingBlock = m.some_other_block_list[4]
m.del_component(movingBlock)

m.block_list[len(m.block_list)] = movingBlock
```
```
Traceback (most recent call last):
  File "blockSetValue.py", line 21, in <module>
    m.block_list[len(m.block_list)] = movingBlock
  File "/home/esjohn/src/pyomo/pyomo/core/base/indexed_component.py", line 412, in __setitem__
    return self._setitem_when_not_present(index, val)
  File "/home/esjohn/src/pyomo/pyomo/core/base/indexed_component.py", line 675, in _setitem_when_not_present
    obj.set_value(value)
  File "/home/esjohn/src/pyomo/pyomo/core/base/block.py", line 710, in set_value
    for k in sorted(iterkeys(val)):
  File "/projects/sems/install/rhel6-x86_64/pyomo/python/six/1.11.0/lib/python2.7/site-packages/six.py", line 599, in iterkeys
    return d.iterkeys(**kw)
  File "/home/esjohn/src/pyomo/pyomo/core/base/block.py", line 549, in __getattr__
[...]

PR #1129, opened 2019-10-04
BigM transformation rewrite

emma58:
## Summary/Motivation:

This fixes bugs in the bigM transformation which mainly occur in situations where disjunctions are iteratively added to a model and the model is transformed and solved at every iteration. It also changes how we track whether components have been transformed and simplifies our methods of mapping between components and their transformed counterparts.

~~NOTE: This is waiting for issue #1106 to be resolved.~~

- Instead of putting the Block created by the transformation on the parent model, put it on the parent block of the disjunction. (This means transforming and solving a Block works as expected.) This means dropping support for allowing the same Disjunct to appear in multiple Disjunctions.
- Add weakrefs to Disjunct, DisjunctData, Disjunction, and DisjunctionData which, after the component has been transformed, point to the transformed object (blocks of constraints in the case of disjuncts, and xor/or constraints in the case of disjunctions). 
- Change active flag of IndexedDisjunct and IndexedDisjunction to be a function of the active status of the ComponentData objects.
- Deprecate CUIDs in both targets and in the argument to specify values for M

### Legal Acknowledgement


qtothec:
Just to verify, do we have a guard against using the same Disjunct in multiple Disjunctions?

emma58:
@qtothec , yes. It gets caught in a bit of a weird place, but while transforming a Disjunction, if one of the Disjuncts has been transformed, it must have been in another Disjunction. So we yell there.

pyomo/gdp/disjunct.py
```diff
@@ -74,28 +76,45 @@ def process(arg):
 
 class _DisjunctData(_BlockData):
 
+    # [ESJ 08/16/2019] TODO: Is this a good idea? (If not, need to change back
+    # in DisjunctionData also)
+    @property
+    def transformation_block(self):
+        return self._transformation_block
+
```
qtothec: Seems like a pretty good idea to me.

pyomo/gdp/disjunct.py
```diff
@@ -74,28 +76,45 @@ def process(arg):
...
+        # if _transformation_block is not None we should yell. else call base
+        # class, check that you have indicator_var and transformation block.
+        _transformation_block = self._transformation_block
+        if not _transformation_block is None:
+            # the disjunct has been transformed, so the truth no longer lives
+            # here and this is a bad idea!
+            raise GDP_Error("Attempting to call set_value on an already-"
+                            "transformed disjunct! Since disjunct %s "
+                            "has been transformed, replacing it here will "
+                            "not effect the model." % self.name)
```
qtothec: I think you mean `affect` vs. `effect`.

pyomo/gdp/disjunct.py
```diff
@@ -74,28 +76,45 @@ def process(arg):
...
+
+        # Call the base class set_value. (Note we are changing decl_order by
+        # doing things this way: the indicator var will always be last rather
+        # than first)
+        super(_DisjunctData, self).set_value(val)
```
qtothec: Hopefully nothing cute relies on `indicator_var` existing before it's created later. I would argue that it's bad practice anyways, so *shrug*.

pyomo/gdp/disjunct.py
```diff
@@ -207,8 +244,16 @@ def __getstate__(self):
...
+            # The user gave us a proper Disjunct block 
+            # [ESJ 06/21/2019] This is really an issue with the reclassifier,
+            # but in the case where you are iteratively adding to an
+            # IndexedDisjunct indexed by Any which has already been transformed,
+            # the new Disjuncts are parading as Blocks already. This catches
+            # them for who they are anyway.
+            # [ESJ 09/13/2019] We still need this. I didn't dig again, but I am
+            # tempted to trust my past self.
+            if isinstance(e, _DisjunctData) or isinstance(e, SimpleDisjunct):
+            #if hasattr(e, 'type') and e.type() == Disjunct:
```
qtothec: Why the change to using `isinstance()`? Also, you could just do `isinstance(e, (_DisjunctData, SimpleDisjunct))`.

pyomo/gdp/plugins/bigm.py
```diff
@@ -19,29 +19,39 @@
...
 
 NAME_BUFFER = {}
 
 def _to_dict(val):
-    if val is None:
+    # [ESJ 09/14/2019] It doesn't seem like this can happen. Even if you
+    # explicitly specify it, this doesn't get called because None is the default
+    # value maybe?
+    # if val is None:
+    #     return val
```
qtothec: I think the result of this is that `None` becomes `{None: None}` instead of remaining `None`, no? Does that still give you the same result? I think so.

pyomo/gdp/plugins/bigm.py
```diff
@@ -173,76 +175,45 @@ def _apply_to_impl(self, instance, **kwds):
...
             targets = (instance, )
             _HACK_transform_whole_instance = True
         else:
             _HACK_transform_whole_instance = False
-        for _t in targets:
-            t = _t.find_component(instance)
-            if t is None:
-                raise GDP_Error(
-                    "Target %s is not a component on the instance!" % _t)
+        knownBlocks = set()
```
qtothec: Comment what this is used for?

pyomo/gdp/plugins/bigm.py
```diff
@@ -253,117 +224,125 @@ def _apply_to_impl(self, instance, **kwds):
...
+            xorConstraint = self._get_xor_constraint(obj.parent_component())
 
         xor = obj.xor
         or_expr = 0
+        # Just because it's unlikely this is what someone meant to do...    
+        if len(obj.disjuncts) == 0:
+            raise GDP_Error("Disjunction %s is empty. This is " 
+                            "likely indicative of a modeling error."  %
+                            obj.getname(fully_qualified=True,
+                                        name_buffer=NAME_BUFFER))
```
qtothec: Is this possible if someone created an `IndexedDisjunction` and wanted to `Skip` an index?

pyomo/gdp/plugins/bigm.py
```diff
@@ -253,117 +224,125 @@ def _apply_to_impl(self, instance, **kwds):
...
+            # pass it down.)
             suffix_list = self._get_bigm_suffix_list(disjunct)
             # relax the disjunct
-            self._bigM_relax_disjunct(disjunct, transBlock, bigM, suffix_list)
+            self._transform_disjunct(disjunct, transBlock, bigM, suffix_list)
+
         # add or (or xor) constraint
         if xor:
-            orConstraint.add(index, (or_expr, 1))
+            xorConstraint.add(index, (or_expr, 1))
```
qtothec: Can we do `expr=or_expr == 1`, or is the tuple for efficiency purposes?

pyomo/gdp/plugins/bigm.py
```diff
@@ -253,117 +224,125 @@ def _apply_to_impl(self, instance, **kwds):
...
         else:
-            orConstraint.add(index, (1, or_expr, None))
+            xorConstraint.add(index, (1, or_expr, None))
```
qtothec: Similarly, can we do `expr=or_expr >= 1`?

pyomo/gdp/plugins/bigm.py
```diff
@@ -421,34 +387,109 @@ def _bigM_relax_disjunct(self, obj, transBlock, bigM, suffix_list):
...
+                sort=SortComponents.deterministic, 
+                descend_into=(Block)):
+            if not obj.algebraic_constraint:
+                # This could be bad if it's active since that means its
+                # untransformed, but we'll wait to yell until the next loop
+                continue
+            # get this disjunction's relaxation block.
+            transBlock = None
+            for d in obj.disjuncts:
+                if d._transformation_block:
```
qtothec: What is this line checking? if `d` has the attribute `_transformation_block`?

pyomo/gdp/plugins/bigm.py
```diff
@@ -421,34 +387,109 @@ def _bigM_relax_disjunct(self, obj, transBlock, bigM, suffix_list):
...
+                    transBlock = d._transformation_block().parent_block()
+                    # We found it, no need to keep looking
+                    break
+            if transBlock is None:
+                raise GDP_Error(
+                    "Found transformed disjunction %s on disjunt %s, "
```
qtothec: typo: `disjunct`

pyomo/gdp/plugins/bigm.py
```diff
@@ -421,34 +387,109 @@ def _bigM_relax_disjunct(self, obj, transBlock, bigM, suffix_list):
...
         obj._deactivate_without_fixing_indicator()
-        infodict['relaxed'] = True
 
-    def _transform_block_components(self, block, disjunct, infodict,
-                                    bigM, suffix_list):
-        # Look through the component map of block and transform
+    def _transform_block_components(self, block, disjunct, bigM, suffix_list):
+        # We first need to find any transformed disjunctions that might be here
+        # because we need to move their transformation blocks up onto the parent
+        # block before we transform anything else on this block
```
qtothec: I think I must be missing something in the rationale for why we are moving these blocks up?

pyomo/gdp/plugins/bigm.py
```diff
@@ -421,34 +387,109 @@ def _bigM_relax_disjunct(self, obj, transBlock, bigM, suffix_list):
...
         for name, obj in list(iteritems(block.component_map())):
-            if hasattr(obj, 'active') and not obj.active:
+            # This means non-ActiveComponent types cannot have handlers
+            if not hasattr(obj, 'active') or not obj.active:
                 continue
             handler = self.handlers.get(obj.type(), None)
             if not handler:
                 if handler is None:
+                    # TODO: It is here that we need to make sure we are only
+                    # yelling of the offender is an ActiveComponent, right?
```
qtothec: typo: "if" the offender

Also, yes, I believe so.

pyomo/gdp/plugins/bigm.py
```diff
@@ -421,34 +387,109 @@ def _bigM_relax_disjunct(self, obj, transBlock, bigM, suffix_list):
...
-            handler(obj, disjunct, infodict, bigM, suffix_list)
-
-    def _warn_for_active_disjunction(self, disjunction, disjunct, infodict,
-                                     bigMargs, suffix_list):
+            handler(obj, disjunct, bigM, suffix_list)
+
+    def _transfer_transBlock_data(self, fromBlock, toBlock):
+        # We know that we have a list of transformed disjuncts on both. We need
+        # to move those over. Then there might be constraints on the block also
+        # (at this point only the diaggregation constraints from chull,
```
qtothec: typo: `disaggregation`

pyomo/gdp/plugins/bigm.py
```diff
@@ -624,18 +659,33 @@ def _xform_constraint(self, obj, disjunct, infodict,
...
+        if constraint in bigMargs:
+            return bigMargs[constraint]
+        elif parent in bigMargs:
+            return bigMargs[parent]
+
+        # [ESJ 08/22/2019] We apparently never actually check what is in
+        # bigMargs... So I'll just yell about CUIDs if we find them here.
+        deprecation_msg = ("In the future the bigM argument will no longer "
+                           "allow ComponentUIDs as keys. Keys should be "
+                           "constraints (in either a dict or ComponentMap)")
```
qtothec: It looks like keys can also be a `Block`?

pyomo/gdp/util.py
```diff
@@ -79,24 +84,62 @@ def clone_without_expression_components(expr, substitute=None):
...
     else:
         raise ValueError(
-            "Expected ComponentUID, Component, Component name, "
-            "or list of these.\n\tReceived %s" % (type(x),))
+            "Expected Component or list of Components."
+            "\n\tRecieved %s" % (type(x),))
+
+# [ESJ 07/09/2019 Should this be a more general utility function elsewhere?  I'm
+#  putting it here for now so that all the gdp transformations can use it
+def is_child_of(parent, child, knownBlocks=None):
```
qtothec: Might want to document what this function does.

pyomo/gdp/disjunct.py
emma58: This is a hack around the fact that we reclassify disjuncts as blocks at the end of bigm. So you could have an algorithm where you generate a new disjunction in each iteration and retransform, like this:
```
model.firstTerm = Disjunct(Any)
model.secondTerm = Disjunct(Any)
model.disjunctionList = Disjunction(Any)
        
for i in range(2):
       model.firstTerm[i].cons = Constraint(expr=model.x == 2*i)
       model.secondTerm[i].cons = Constraint(expr=model.x >= i + 2)
       model.disjunctionList[i] = [model.firstTerm[i], model.secondTerm[i]]

       TransformationFactory('gdp.bigm').apply_to(model)
```
After the first iteration, `firstTerm` and `secondTerm` are reclassified as blocks. So in all later iterations, when you add to them in the first two lines of the loop, you add blocks. The third line will throw a ValueError with the hasattr version because `e.type()` is `Block`. However, they are still `_DisjunctData` instances, hence the change.

pyomo/gdp/plugins/bigm.py
emma58: Yes, you're right, it would be the same result anyway.

pyomo/gdp/plugins/bigm.py
emma58: Because for nested disjunctions, it is sufficient to relax only the XOR constraint of the inner disjunction, so after transforming the inner disjunction, we move all the transformed constraints from its disjuncts up to avoid transforming them again.

pyomo/gdp/disjunct.py
qtothec: Ok, that makes sense. You could condense the two `isinstance()` calls into one, though: https://stackoverflow.com/a/33311330/1557256

pyomo/gdp/plugins/bigm.py
qtothec: Ah, gotcha. So that you don't end up with `g(x) <= M(1-y1) + M(1-y2)`.

pyomo/gdp/plugins/bigm.py
```diff
@@ -251,8 +251,8 @@ def _transform_blockData(self, obj, bigM):
                 descent_order=TraversalStrategy.PostfixDFS):
             self._transform_disjunction(disjunction, bigM)
 
-    def _get_xor_constraint(self, disjunction):
-        # Put the disjunction constraint on its parent block and
+    def _get_xor_constraint(self, disjunction, transBlock):
+        # Put the disjunction constraint on the transformation block and
         # determine whether it is an OR or XOR constraint.
```
qtothec: The name of the function and its description seem to diverge slightly. `get` operations usually don't modify state.

pyomo/gdp/tests/test_bigm.py
```diff
@@ -2563,21 +2573,30 @@ def test_transformed_disjunction_all_disjuncts_deactivated(self):
         m.disjunction = Disjunction(expr=[m.x == 0, m.x >= 4])
         m.disjunction_disjuncts[0].nestedDisjunction = Disjunction(
             expr=[m.y == 6, m.y <= 1])
```
qtothec: Should this be `m.disjunction.disjuncts[0]`?

pyomo/gdp/plugins/bigm.py
emma58: No, because Skip just doesn't create the index, so the DisjunctionData would not be there at all. So I think this is okay.

pyomo/gdp/plugins/bigm.py
emma58: I don't know?? @jsiirola?

jsiirola: The tuple notation is slightly faster, but not enough to warrant using it at the cost of clarity.

pyomo/gdp/disjunct.py
```diff
@@ -74,28 +76,43 @@ def process(arg):
...
+        # doing things this way: the indicator var will always be last rather
+        # than first)
+        super(_DisjunctData, self).set_value(val)
+
         if val:
             if 'indicator_var' not in val:
                 self.add_component('indicator_var', _indicator_var)
-            for k in sorted(iterkeys(val)):
-                self.add_component(k,val[k])
+            if 'transformation_block' not in val:
```
qtothec: Should there be an underscore here?

pyomo/gdp/disjunct.py
```diff
@@ -74,28 +76,43 @@ def process(arg):
...
+                self._transformation_block = _transformation_block
         else:
             self.add_component('indicator_var', _indicator_var)
+            self._transformation_block = _transformation_block
```
qtothec: Also, I am confused: wouldn't the value of `_transformation_block` always be `None` here, given that an error would have been raised above otherwise?

pyomo/gdp/plugins/bigm.py
```diff
@@ -706,3 +818,57 @@ def _estimate_M(self, expr, name):
...
+
+    def get_src_disjunction(self, xor_constraint):
+        m = xor_constraint.model()
+        for disjunction in m.component_data_objects(Disjunction):
+            if disjunction._algebraic_constraint:
+                if disjunction._algebraic_constraint() is xor_constraint:
+                    return disjunction
+        raise GDP_Error("It appears that %s is not an XOR or OR constraint "
+                        "resulting from transforming a Disjunction."
+                        % xor_constraint.name)
```
qtothec: We have to do a linear scan of the model to get the source disjunction?

pyomo/gdp/util.py
qtothec: I think it would be fair to have a future PR that adds a docstring to this and lifts it into pyomo.common.

pyomo/gdp/tests/test_disjunct.py
```diff
@@ -10,11 +10,16 @@
...
+from pyomo.core import ConcreteModel, Var, Constraint, Block, \
+    TransformationFactory
+from pyomo.gdp import Disjunction, Disjunct, GDP_Error
+import pyomo.gdp.plugins.bigm
 
 from six import iterkeys
 
+# TODO DEBUG
+from nose.tools import set_trace
+
```
qtothec: I should remember to use this. But you probably don't need to leave this here.

pyomo/gdp/plugins/bigm.py
```diff
@@ -378,41 +384,28 @@ def _bigM_relax_disjunct(self, obj, transBlock, bigM, suffix_list):
...
-        # add reference to original disjunct to info dict on transformation
-        # block
+                
+        if obj._transformation_block is not None:
+            # we've transformed it, which means this is the second time it's
+            # appearing in a Disjunction
+            raise GDP_Error(
+                    "The disjunct %s has been transformed, but a disjunction "
+                    "it appears in has not. Putting the same disjunct in "
+                    "multiple disjunctions is not supported." % obj.name)
```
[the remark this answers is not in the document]
emma58: Oh, interesting. There's not, but it wouldn't be hard to do... I'm not sure how we want to tell the transformation that this is what we want because I don't think we'd want the default to be for it to transform orphan Disjuncts. And even if you specified a Disjunct in the list of targets, how do you distinguish between treating a Disjunct target as a Block (transform everything on it) vs. transforming it without it being part of a Disjunction? But in any case, the actual transformation would be easy to do.

pyomo/gdp/plugins/bigm.py
emma58: Currently yes, because I didn't store a reverse map during the transformation. If we need this to be efficient, then we can just keep a map from xor constraints to their disjunctions on the transformation block. I didn't bother because I wasn't convinced a lot of people were going to be asking xor constraints where they came from. But if we ever are... Then yes, this should be changed.

pyomo/gdp/disjunct.py
```diff
@@ -158,15 +167,31 @@ def __init__(self, *args, **kwds):
...
-    pass
+    #
+    # HACK: this should be implemented on ActiveIndexedComponent, but
+    # that will take time and a PEP
+    #
+    @property
+    def active(self):
+        return any(d.active for d in itervalues(self._data))
+
+    @active.setter
```
emma58: @jsiirola, I'm confused: You can't set `active`, right? We added this in both IndexedDisjunct and IndexedDisjunction but it's not covered in the tests because I don't know how to call it...

pyomo/gdp/plugins/bigm.py
qtothec: You're right. It's unlikely to be used. Maybe a quick note to that effect, so that people templating from this code understand why it was done that way?

pyomo/gdp/plugins/bigm.py
```diff
@@ -208,23 +206,12 @@ def _apply_to_impl(self, instance, **kwds):
...
-            if isinstance(t, ComponentUID):
-                tmp = t
-                t = t.find_component(instance)
-                if t is None:
-                    raise GDP_Error(
-                        "Target %s is not a component on the instance!" % tmp)
-
             # check that t is in fact a child of instance
-            if not is_child_of(parent=instance, child=t,
+            elif not is_child_of(parent=instance, child=t,
```
qtothec:
```suggestion
            if not is_child_of(parent=instance, child=t,
```

Otherwise you're missing an initial `if`.

pyomo/gdp/plugins/bigm.py
```diff
@@ -150,15 +156,29 @@ def _get_bigm_suffix_list(self, block):
...
+            return arg_list
+        while block is not None:
+            if block in bigm_args:
+                arg_list.append({block: bigm_args[block]})
+            block = block.parent_block()
+        return arg_list
 
     def _apply_to(self, instance, **kwds):
         assert not NAME_BUFFER
+        assert not used_args
```
jsiirola: Why is this a global?  How hard would it be to make this a local variable (or at least a variable attached to this transformation instance)?

pyomo/gdp/plugins/bigm.py
```diff
@@ -175,76 +195,51 @@ def _apply_to_impl(self, instance, **kwds):
...
-                # contained _DisjunctData's indicator_vars!) due to our
-                # class hierarchy.  Instead, we will directly call the
-                # relevant base class (safe-ish since we are verifying
-                # that all the contained _DisjunctionData are
-                # deactivated directly above).
-                ActiveComponent.deactivate(obj)
+
+        # issue warnings about anything that was in the bigM args dict that we
+        # didn't use
+        if not bigM is None and len(bigM) > len(used_args):
```
jsiirola: PEP8 style: `if bigM is not None`

jsiirola: Also, why not actually do a set comparison (e.g., `set(bigM) - set(used_args)`)?  Because of line 713 (`used_args[block] = val`), are we guaranteed that just comparing the number of entries is sufficient (block comes from arg_list and not bigMargs)?

pyomo/gdp/plugins/bigm.py
```diff
@@ -255,117 +250,143 @@ def _apply_to_impl(self, instance, **kwds):
...
 
-    def _transformDisjunction(self, obj, transBlock, bigM):
+    def _transform_disjunction(self, obj, bigM):
+        if not obj.active:
+            return
+
+        # if this is an IndexedDisjunction we have seen in a prior call to the
+        # transformation, we already have a transformation block for it. We'll
+        # use that.
+        if not obj._algebraic_constraint is None:
```
jsiirola: PEP8 style: `obj._algebraic_constraint is not None`

pyomo/gdp/plugins/bigm.py
```diff
@@ -255,117 +250,143 @@ def _apply_to_impl(self, instance, **kwds):
...
+            arg_list = self._get_bigm_arg_list(bigM, disjunct)
             # relax the disjunct
-            self._bigM_relax_disjunct(disjunct, transBlock, bigM, suffix_list)
+            self._transform_disjunct(disjunct, transBlock, bigM, arg_list,
+                                     suffix_list)
+
         # add or (or xor) constraint
         if xor:
-            orConstraint.add(index, (or_expr, 1))
+            xorConstraint.add(index, expr=or_expr == 1)
```
jsiirola: FYI: `.add(index, expr)` is legacy syntax.  The preferred syntax should be:
```python
xorConstraint[index] = or_expr == 1
```
or
```python
xorConstraint[index].set_value(or_expr == 1)
```

pyomo/gdp/plugins/bigm.py
```diff
@@ -423,77 +438,111 @@ def _bigM_relax_disjunct(self, obj, transBlock, bigM, suffix_list):
...
+            self._transfer_transBlock_data(transBlock, destinationBlock)
+            # we leave the transformation block because it still has the XOR
+            # constraints, which we want to be on the parent disjunct.
+
+        # Now look through the component map of block and transform
         # everything we have a handler for. Yell if we don't know how
         # to handle it.
         for name, obj in list(iteritems(block.component_map())):
-            if hasattr(obj, 'active') and not obj.active:
+            # This means non-ActiveComponent types cannot have handlers
```
jsiirola: Should we just do:
```python
for obj in block.component_objects(active=True, descend_into=False): 
```
instead?  Since we will not complain about non-ActiveComponent objects, we might as well just not visit them.

Also, since we are skipping non-active components, we can remove the handler for Var, Expression, and Suffix, right?

pyomo/gdp/plugins/bigm.py
```diff
@@ -706,3 +788,131 @@ def _estimate_M(self, expr, name):
...
+                raise GDP_Error(
+                    "Constraint %s is not on a disjunct and so was not "
+                    "transformed" % constraint.name)
+            parent_disjunct = parent_disjunct.parent_block()
+
+        return parent_disjunct
+
+    def _get_constraint_transBlock(self, constraint):
+        parent_disjunct = self._find_parent_disjunct(constraint)
+        transBlock = parent_disjunct._transformation_block
```
jsiirola: Similar to @qtothec's comment above:  is `_transformation_block` guaranteed to be present?  Instead, should this be
```python
transBlock = getattr(parent_disjunct, '_transformation_block', None)
```
(similar questions for the other information methods below)

pyomo/gdp/plugins/bigm.py
```diff
@@ -706,3 +788,131 @@ def _estimate_M(self, expr, name):
...
+        # the world is really against us.
+        if not hasattr(transBlock, "_constraintMap"):
+            raise GDP_Error("Constraint %s is not a transformed constraint" 
+                            % transformedConstraint.name)
+        return transBlock._constraintMap['srcConstraints'][transformedConstraint]
+
+    def _find_parent_disjunct(self, constraint):
+        # traverse up until we find the disjunct this constraint lives on
+        parent_disjunct = constraint.parent_block()
+        while type(parent_disjunct) not in (_DisjunctData, SimpleDisjunct):
```
jsiirola: Should this use `isinstance` instead of exact type comparison?

pyomo/gdp/plugins/bigm.py
```diff
@@ -706,3 +788,131 @@ def _estimate_M(self, expr, name):
...
+        """Return the Disjunct object whose transformed components are on
+        transBlock.
+
+        Parameters
+        ----------
+        transBlock: _BlockData which is in the relaxedDisjuncts IndexedBlock
+                    on a transformation block.
+        """
+        if not hasattr(transBlock, '_srcDisjunct') or \
+           not type(transBlock._srcDisjunct) is weakref_ref:
```
jsiirola: There are other potential issues here, right?  Like could the weakref have fallen out of scope?  Would a better trap be:
```python
try:
    return transBlock._srcDisjunct()
except:
    raise GDP_Error("Block %s doesn't appear to be a transformation "
                    "block for a disjunct. No source disjunct found." 
                    "\n\t(original error: %s)" 
                    % (transBlock.name, sys.exc_info()[1]))
```

pyomo/gdp/plugins/bigm.py
emma58: Laziness... It's easy, you'd just have to pass it all the way down all the function calls because it doesn't get used until you are getting an M value while transforming a constraint. So we could either do that or attach it to the transformation instance. I'll attach it to the instance unless you have objections?

pyomo/gdp/plugins/bigm.py
emma58: Oh, yeah, the set comparison is better. We have to use ComponentSet though in case someone messes up and gives us something unhashable (Var). (I think number of entries would be sufficient because we never add anything to arg_list that is not in bigMargs, and we add all the Blocks that we could possible need. But since we do the set comparison to get the warning message, might as well do it initially.)

pyomo/gdp/plugins/bigm.py
emma58: I agree about rewriting the loop, but we still hit Var, Expression and Suffix in what you suggest. So we would still need the handlers. The only way I see to get rid of the handlers is if we only iterate through stuff we need to transform. But then we would fail silently if there was something odd, so I'm tempted to leave it.

pyomo/gdp/plugins/bigm.py
emma58: I think it is guaranteed here because `_find_parent_disjunct` can only return a _DisjunctData or SimpleDisjunct, either of which will have it.

pyomo/core/base/block.py
```diff
@@ -1887,11 +1887,12 @@ def __init__(self, *args, **kwargs):
     def _getitem_when_not_present(self, idx):
         return self._setitem_when_not_present(idx)
 
-    def find_component(self, label_or_component):
+    def find_component(self, label_or_component, context=None):
```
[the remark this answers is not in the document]
emma58: Oops, yeah, I think that is a spandrel from some other adventure. I'll revert.

emma58:
@jsiirola, okay, I think everything is addressed (except where I argued with you).

pyomo/gdp/disjunct.py
jsiirola: Testing both `_DisjunctData` and `SimpleDisjunct` is redundant.

pyomo/gdp/plugins/bigm.py
```diff
@@ -423,77 +443,109 @@ def _bigM_relax_disjunct(self, obj, transBlock, bigM, suffix_list):
...
+            # we leave the transformation block because it still has the XOR
+            # constraints, which we want to be on the parent disjunct.
+
+        # Now look through the component map of block and transform everything
+        # we have a handler for. Yell if we don't know how to handle it. (Note
+        # that because we only iterate through active components, this means
+        # non-ActiveComponent types cannot have handlers.)
+        for obj in block.component_objects(active=True, descend_into=False):
             handler = self.handlers.get(obj.type(), None)
             if not handler:
```
jsiirola: Philosophical question: should this only raise an exception if the component is an active ActiveComponent derivative?  

I think it is unfortunate that non-ActiveComponents are returned when `active=True`.  I wonder if that should be reported as a bug?

qtothec:
:-o it's passing all tests and approved.

PR #1134, opened 2019-10-13
FBBT Bugs

michaelbynum:
## Summary/Motivation:
This PR fixes 2 bugs in `pyomo.contrib.fbbt`.
- There was a bug when handling integer variables without bounds. An error was raised when trying to round the bound. This PR adds a check for infinity bounds before attempting to do any rounding for integer variables.
- There was a bug caused by floating point arithmetic. Occasionally, the lower bound of a variable can end up larger than the upper bound due to floating point arithmetic. This changes ensures that does not happen.

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/fbbt/fbbt.py
+++ b/pyomo/contrib/fbbt/fbbt.py
@@ -1023,6 +1023,23 @@ _prop_bnds_root_to_leaf_map[_GeneralExpressionData] = _prop_bnds_root_to_leaf_Ge
 _prop_bnds_root_to_leaf_map[SimpleExpression] = _prop_bnds_root_to_leaf_GeneralExpression
 
 
+def _check_and_reset_bounds(var, lb, ub):
+    """
+    This function ensures that lb is not less than var.lb and that ub is not greater than var.ub.
+    """
+    orig_lb = value(var.lb)
+    orig_ub = value(var.ub)
+    if orig_lb is None:
+        orig_lb = -interval.inf
+    if orig_ub is None:
+        orig_ub = interval.inf
+    if lb < orig_lb:
+        lb = orig_lb
+    if ub > orig_ub:
+        ub = orig_ub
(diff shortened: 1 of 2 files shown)


pyomo/contrib/fbbt/fbbt.py
```diff
@@ -1120,6 +1120,29 @@ def visiting_potential_leaf(self, node):
...
+                    orig_ub = value(node.ub)
+                    if orig_lb is None:
+                        orig_lb = -interval.inf
+                    if orig_ub is None:
+                        orig_ub = interval.inf
+                    if lb < orig_lb:
+                        lb = orig_lb
+                    if ub > orig_ub:
+                        ub = orig_ub
+                    self.bnds_dict[node] = (lb, ub)
```
jsiirola: This procedure looks good, however a comment here that indicates the logic behind this code would be helpful.  When I initially read this, I was wondering why you were relaxing both `lb` and `ub` ... but the checks against the original bounds make sense (that is, I read this as if the bounds ever cross, then relax the computed values by `feasibility_tol`).

pyomo/contrib/fbbt/fbbt.py
```diff
@@ -1130,21 +1153,22 @@ def visiting_potential_leaf(self, node):
...
+                orig_lb = value(node.lb)
+                orig_ub = value(node.ub)
+                if orig_lb is None:
+                    orig_lb = -interval.inf
+                if orig_ub is None:
+                    orig_ub = interval.inf
+                if lb < orig_lb:
+                    lb = orig_lb  # don't make the bounds worse than the original bounds
+                if ub > orig_ub:
+                    ub = orig_ub  # don't make the bounds worse than the original bounds
```
jsiirola: A comment similar to the above would be helpful.  Also, given that this is repeated code, would a helper function be warranted?

PR #1149, opened 2019-10-28
Leverage a name_buffer when generating names in pyomo.gdp

jsiirola:
## Summary/Motivation:
The pyomo.gdp Big-M and CHull transformations exhibit quadratic scaling due to inefficiencies in generating component names.  This switches to using `getname()` with a name buffer to resolve this behavior.

- use `getname(name_buffer=NAME_BUFFER)` when generating all component names

### Legal Acknowledgement

### DIFF
--- a/pyomo/gdp/plugins/bigm.py
+++ b/pyomo/gdp/plugins/bigm.py
@@ -31,6 +31,7 @@ from six import iterkeys, iteritems
 
 logger = logging.getLogger('pyomo.gdp.bigm')
 
+NAME_BUFFER = {}
 
 def _to_dict(val):
     if val is None:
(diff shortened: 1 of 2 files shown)


pyomo/gdp/plugins/bigm.py
```diff
@@ -507,7 +521,8 @@ def _xform_constraint(self, obj, disjunct, infodict,
         # Though rare, it is possible to get naming conflicts here
         # since constraints from all blocks are getting moved onto the
         # same block. So we get a unique name
```
qtothec: With @emma58 changes to Big-M, is this stipulation still true? 

> Though rare, it is possible to get naming conflicts here
> since constraints from all blocks are getting moved onto the
> same block. So we get a unique name

pyomo/gdp/plugins/bigm.py
```diff
@@ -299,8 +310,9 @@ def _getXorConstraint(self, disjunction):
         # can no longer make that distinction in the name.
         #    nm = '_xor' if xor else '_or'
         nm = '_xor'
-        orCname = unique_component_name(parent, '_gdp_bigm_relaxation_' +
-                                        disjunction.name + nm)
+        orCname = unique_component_name(
+            parent, '_gdp_bigm_relaxation_' + disjunction.getname(
+                fully_qualified=True, name_buffer=NAME_BUFFER) + nm)
```
qtothec: Similarly, do we need `unique_component_name` here?

jsiirola:
@qtothec: yes - I think you still need the calls to `unique_component_name`.  Note that this PR doesn't include @emma58's new changes - that will come as a separate PR.  We just realized that this was an easy enough patch that we felt it was important to get in quickly.

qtothec:
In that case, it seems fine to me. I guess Emma's going to have some refactoring to do to her PR.

PR #1172, opened 2019-11-14
Add MC++ interface versioning

### LINKED ISSUE #1123: Add versioning to MC++ library shared object
Versioning needs to be added to MC++ library so that it is possible to know when the `*.so` file is out of date with the Pyomo version.

qtothec:
## Summary/Motivation:
Adds a version check to the MC++ library shared object file to make sure it does not fall out of date with the python code.

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/mcpp/mcppInterface.cpp
+++ b/pyomo/contrib/mcpp/mcppInterface.cpp
@@ -20,6 +20,9 @@ std::string lastDisplay;
 
 extern "C"
 {
+    // Version number
+    const char* get_version() {return "19.11.12"}
+
     // Functions to build up an MC object
     void* newVar(double lb, double pt, double ub, int count, int index)
     {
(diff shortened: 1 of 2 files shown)


pyomo/contrib/mcpp/mcppInterface.cpp
```diff
@@ -11,15 +11,26 @@
...
 
 extern "C"
 {
+
+    // Version number
+    const char* get_version() {
+        std::ostringstream vStream;
+        vStream << "19.11.12" << std::flush;
+        version = vStream.str();
+        return version.c_str();
```
jsiirola: This is technically a seg fault: you are returning a pointer to an internal data member of an object allocated on the stack.  When the function returns, the `vStream` is deallocated (including the puffer that the `char*` is pointing to).  You are just "lucky" that the OS hasn't reclaimed / reallocated the memory yet.

A safer solution is:
```c++
const char* version_string = "19.11.12";
const char* get_version() {
    return version_string;
}
```
This way, the pointer that is returned is statically allocated as part of the program's global memory space.

qtothec:
I am more inclined to have that flexibility on the python side. The reason that I noticed this problem was that an update to the python code made it incompatible with an old pre-existing compiled library. Is there a reason to have it built in on the C++ side?

PR #1174, opened 2019-11-14
Don't call formatargspec with Python 3

### LINKED ISSUE #1173: Deprecation Warning: formatargspec
In the new version of Pyomo (3.6.7) an error/warning occurs: `DeprecationWarning: 'formatargspec' is deprecated since Python 3.5. Use 'signature' and the 'Signature' object directly`

This refers to this method: https://github.com/Pyomo/pyomo/blob/master/pyomo/core/base/util.py#L57

lafrech:
## Summary/Motivation:

Remove deprecated function call.

- Use the formatargspec trick only on Python 2. It is not needed in PY3.4+ anyway as per the comment.

### Legal Acknowledgement


jsiirola:
I partially reverted this to still create a one-off function when disabling methods in Python 3.x.  This makes sure that the function signature is preserved.  It also preserves the function name (so that the wrapped function's name is not `<lambda>`.

EliasIEE:
Thanks a lot both of you for the super fast fix!

Issue #893, opened 2019-03-09
Bug in expand arc : bounds are copied from an inlet to outlets 


ReinboldV:
I am using pyomo.network to build sort of graph with blocks and extensive flow connections. In this application flow variable are extensive (or conservative, see #882), positive or negative (general case). Block may have inlets and outlets,  counted positive when the flow come respectively in and out. But In some cases I need a flow to be reversible, which means positive or negative some times.  

During the arc expansion, some properties of the ports, like rule, up bound or low bound,  are copied to the expanded arc variables (named $arc_name$_expanded). For Extensive connection, this actually make sense if all the connected variable are positive. Let's suppose their are two sources feeding one load. If the load has boundaries, it makes sense to impose these limits to the expended arc variables. 
But it is actually quiet restrictive when considering positive and negative variables. Let's see an example to illustrate my purpose. Let's have to sources feeding one destination. An the destination has a dummy  bounds rule (lambda m, t : (-10, 10) ).

```
from pyomo.environ import *                                        
from pyomo.network import *                                        
from pyomo.dae import *                                            
                                                                   
m = ConcreteModel()                                                
                                                                   
m.b1 = Block()                                                     
m.b2 = Block()                                                     
m.b4 = Block()                                                     
                                                                                                                               
def source_block(b):                                               
    b.t = ContinuousSet(bounds=(0, 1))                             
    b.p_out = Var(b.t)                                             
    b.outlet = Port(initialize={'p': (b.p_out, Port.Extensive)})   
                                                                   
def load_block(b):                                                 
    b.t = ContinuousSet(bounds=(0, 1))                             
    b.p_in = Var(b.t,  bounds=lambda m, t: (-10, 10))              
    b.inlet = Port(initialize={'p': (b.p_in, Port.Extensive)})     
                                                                   
source_block(m.b1)                                                 
source_block(m.b2)                                                 
load_block(m.b4)                                                   
                                                                   
m.b14 = Arc(source=m.b1.outlet, destination=m.b4.inlet)            
m.b24 = Arc(source=m.b2.outlet, destination=m.b4.inlet)            
                                                                   
TransformationFactory('dae.finite_difference').apply_to(m, nfe=1)  
TransformationFactory("network.expand_arcs").apply_to(m)                                                      
``` 
A pprint() would give us this : 
```
5 Block Declarations
    b1 : Size=1, Index=None, Active=True
        1 Var Declarations
            p_out : Size=2, Index=b1.t
[...]

PR #1186, opened 2019-11-18
Network: allow disabling split_frac; correct bounds on duplicated variables

### LINKED ISSUE #882: Implement a simple Kirchoff law for Extensive port in pyomo.network
Hello,
I am experimenting the pyomo network for hierarchical modelling. First, thanks a lot for your contributions. This is a very useful tool !
I have some hard time understanding the philosophy behind the Port.Extensive rule when more than one arc is expanded.

Let's take a simple example with 3 blocks, one "source", b1, connected to 2 "destinations", b2 and b3. Let's assume that each destination is connected to the source through a port that expose an extensive variable, i.e. a flow.

Here a simple code to do this : 

```
from pyomo.environ import *
from pyomo.network import *
from pyomo.dae import *

m = ConcreteModel()

m.b1 = Block()
m.b2 = Block()
m.b3 = Block()

def source_block(b):
    b.t = ContinuousSet(bounds=(0, 1))
    b.p_out = Var(b.t)
    b.outlet = Port(initialize={'p': (b.p_out, Port.Extensive)})

def load_block(b):
    b.t = ContinuousSet(bounds=(0, 1))
    b.p_in = Var(b.t)
    b.inlet = Port(initialize={'p': (b.p_in, Port.Extensive)})

source_block(m.b1)
load_block(m.b2)
load_block(m.b3)

m.b12 = Arc(source=m.b1.outlet, destination=m.b2.inlet)
m.b13 = Arc(source=m.b1.outlet, destination=m.b3.inlet)

TransformationFactory('dae.finite_difference').apply_to(m, nfe=1)
TransformationFactory("network.expand_arcs").apply_to(m)

m.pprint()
```
I don't really understand why one would introduce the following variables : `b13_expanded.splitfact` and `b13_expanded.splitfact` the following constraints `m.b12_expanded.p_split` and `m.b13_expanded.p_split` within the new connection blocks. In a general flow  connection, one would only need a Kirchoff law (i.e. sum of flows are equal to 0) without imposing split (the physic will actually take care of this.), i.e. : `b1.p_out[i] == b2.p_in[t] + b3.pin[t]` which is equivalent to the set of generated constraints `b1.outlet_p_outsum`,  `b2.inlet_p_insum` and `b3.inlet_p_insum`. So here is my questions/remarks : 

- Would this Kirchoff law be a new kind of expansion for arcs ?
- Is there an easy way to automatically disable the constraints `b12_expanded.p_split` and `b13_expanded.p_split` ? Or make this behaviour an option ? In the call for Arc for instance.

Thanks again for your contribution and for your help, 
Greetings

ReinboldV:
## Summary/Motivation:
Motivations are discussed in issue #882 . 
There are the main points : 

pyomo.network module allows 2 types of arc expansion rule. 
- First, a rule called Equality. During the expansion of arcs, this rule allows to create an equality constraint between variables' ports. Physically, it is used for effort connexions (temperature, voltage, pressure, etc.). 

- Secondly, a rule called Extensive, that introduce constraints of the form $$\sum_i f_{in}^i = \sum_j f_{out}^j $$, joined with other constraints that impose a split fraction for non-trivial cases (more than one source or targets). Physically, this is used for flow connections where we need a constant split factor between sources and targets (my guess is that it used in hydraulics, where pressure equations are not modelled and we assume a constant split between pipes).

~~The proposal is to create a new rule, called Conservative, based on the Kirchhoff current law, i.e. without the assumption of fix split factor parameter. This will allow to represent flow connection, (hydraulics, current, power etc.) in a more general case.~~

- ~~Adding rule called Conservative~~ Updating the arguments to Extensive to suppress the generation of split fraction variables
- Adding unitest to check the behaviour
- Updating the logic for computing the bounds on created variables.

### Legal Acknowledgement


jsiirola:
@gseastream: if you have a moment, can you take a look at this?

pyomo/network/util.py
```diff
@@ -25,24 +25,25 @@ def replicate_var(comp, name, block, index_set=None):
...
-    except AttributeError:
-        pass
+    # try:
+    #     var_args['domain'] = comp.domain
+    # except AttributeError:
+    #     pass
+    # try:
+    #     var_args['bounds'] = comp.bounds
+    # except AttributeError:
+    #     pass
```
gseastream: So this part is related to #893, and @jsiirola I think it'd be good for you to weigh in.

I can definitely see the problem with copying bounds to expanded variables and perhaps even implicit/auto variables, especially with the introduction of potentially negative variables. However, I think there is also a problem if we chose to skip copying the domain to the new variables, since I can imagine we don't want to be equating reals with integers and then passing that to our solvers, if we can help it. Copying domains should present less of a problem than bounds since they can be as inclusive as Reals or Integers. Although, you could also have domains with inherent bounds like NonNegativeReals or Binary. I would think if the bounds inherent in those domains starting causing problems for a user, it would be the user's responsibility to change the domain of the parent variable to be more inclusive. Does that make sense?

If we agree to keep copying domains but remove copying bounds, then I would say uncomment the domain parts, delete the now commented-out bounds parts, and update any comments/docstring in this function that refers to copying bounds.

I also took a look at the failing tests, they all have to do with checking to make sure domains and bounds were copied, so if we go with the above plan then you should go into the failing tests in `test_arc.py` and remove the assertions for bounds.

gseastream:
Aside from my above comments, out of curiosity, if you use variables with the Conservative rule, what in your model do you use to dictate how much of the variable split goes to each specific outlet arc?

pyomo/network/util.py
ReinboldV: Copying domain is a good compromise indeed, and it make sense. 
Thanks for looking to the failing test. This PR should be fine after changing it.

ReinboldV:
My application is electric engineering. My port is (voltage:effort, current:conservative) and in the block, I have a characteristic I = f(delta V). For a simple resistance, it is I = R(v1 - v2) (the Ohm's law).
For a simple resistance, if we impose v1 and v2, the flow is dictated by the Ohm's law.

In hydraulic (I guess it is your field of application), the flow is dictated by the hydraulic resistance and the pressure difference. In real life, like heating and cooling network, this is hard to know those characteristics (pipes, valves etc.). So we make the assumption to fix the flow (by choosing a split factor), making the hypothesis that we will always be able to ensure this by controlling pressure at the source or controlling valves. The pressure is no longer present in the equations (and not present in the port) because of this hypothesis.  

In another applications, I am working with powers, connecting various sources and loads modelled as blocks. A node is thus only a power balance between blocks. The power split is the actual unknown, fixing it makes no sense. In optimization, we usually call this a unit commitment problem. And there is no reason that the split_fact is constant over the time.

gseastream:
So are you saying for current you would add a constraint to the port's block that uses the `evar` that was created on the arc's expanded block during the expansion transformation in order to relate it to resistance and voltage? (just trying to learn about this use case)

pyomo/network/util.py
jsiirola: Setting the domain can also be problematic.  Consider the case of an integer Port with two Arcs leaving it.  There is no reason to believe that the arc flow rates should be restricted to discrete values - only that the sum be integer.

This does beg the question, though if we need to augment the API to allow the user to specify domains / bounds for arc variables?

jsiirola:
Sorry for not getting to this for so long.  I think there are a couple issues with the proposed code:
 - significant repeated code: the `_Split_Conservative` helper method is a simplification of `_Split`
 - I do not find the name `Conservative` intuitive, especially as it implements a rule for quantities that I would have normally termed "extensive".
 - The solution of not propagating bounds / domains is weaker than necessary.  I think you *can* propagate bounds, but only in the case where an arc is connected as either the sole inlet or outlet arc for a port.

Thinking about this more, I do not think that we need another rule at all, provided that we update Extensive so that we can specify `include_splitfrac=False` and actually have the arc expansion omit the split fractions.  That turned out to be easier to implement than I thought, and I have submitted it as a PR to this PR; see ReinboldV/pyomo#1.

@gseastream, @ReinboldV: what do you think?

ReinboldV:
Hi John, thanks for the feed back !
This solution seems way better than the hack I was using. I've just merged your PR. 
Thanks a lot

pyomo/network/tests/test_arc.py
```diff
@@ -1061,6 +1136,129 @@ def test_extensive_single_var(self):
 13 Declarations: x y z p1 p2 p3 a1 a2 a1_expanded a2_expanded p1_v_outsum p2_v_insum p3_v_insum
 """)
 
+    def test_conservative_expansion(self):
+        m = ConcreteModel()
+        m.time = Set(initialize=[1, 2, 3])
```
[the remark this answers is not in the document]
gseastream: This was marked as resolved but I'm still not sure if this `time` declaration was intentional?

pyomo/network/util.py
gseastream: It's technically possible already (and not altogether too difficult) to modify the domains and such of arc variables. All the user needs to do is `m.myarc.flow.domain = NonNegativeReals` (or whatever the right syntax is) because Arc's getattr method:

> Returns `self.expanded_block.name` if it exists

jsiirola:
> Also see below for my comment on something that was marked resolved but I'm not sure why.

I went in and updated that test to remove the problem you brought up.

jsiirola:
OK - This should be good to as soon as @qtothec has a chance to look it over.

pyomo/network/port.py
```diff
@@ -448,7 +458,7 @@ def Equality(port, name, index_set):
             Port._add_equality_constraint(arc, name, index_set)
 
     @staticmethod
-    def Extensive(port, name, index_set, include_splitfrac=False,
+    def Extensive(port, name, index_set, include_splitfrac=None,
```
qtothec: You might want to update the documentation below to correspond with the change in function API.

jsiirola: Done.

jsiirola:
The last commit only impacted documentation, so waiting for Appveyor to catch up is unnecessary.  All other tests passed, so I will merge.

Issue #1201, opened 2019-11-25
Issues with NEOS solvers and solution output


gnowzil:
I'm with the NEOS team and I'm trying to figure out some issues with Pyomo+NEOS.

It seems that 
`pyomo solve --solver=<solver> --solver-manager=neos model.py`
works only if `<solver>` has an explicit plugin (e.g., any of the solvers in pyomo/solvers/plugins/solvers).  This is mentioned in #910.  

The following seems to work for any NEOS-valid solver:
```
manager = SolverManagerFactory('neos')
results = manager.solve(model,opt='<solver>')
```

Is there a bug with the first method?  And if not, I assume that the second method is the way to go?  I dug around a little bit and it looks like the first method tries to create `SolverFactory(<solver>)` but this results in an unknown solver for some of the NEOS solvers.

Furthermore, in either case, no solution seems to be returned.  I don't see any error messages, but printing the results always shows 
```
Solution: 
- number of solutions: 0
  number of solutions displayed: 0
```
I'm not sure what Pyomo is expecting from NEOS, but if there's anything we can do on our end to fix this, let me know.

blnicho:
@gnowzil thanks for the note, it is certainly possible that this is a bug. We're hoping to rework Pyomo's solver interfaces this December which might fix this and leave us with a more robust interface to NEOS.

gnowzil:
Hi,  I just wanted to follow up on this.  We've been getting some queries about pyomo on NEOS.  I ran small examples through all of the solvers available and was met with varying degrees of success.  I'm not an expert in using Pyomo so I may be doing this incorrectly, but I've summarized what I have seen below.

- cbc, conopt, cplex, mosek, and snopt appeared to work and returned a solution.
- filter sqp, knitro, l-bfgs-b, lancelot, lgo, loqo, minlp, minos, minto, ooqp appeared to work if I use the command `python myexample.py`, with `SolverManagerFactory('neos')` and the `solve` call in the script, but failed if I used `pyomo solve --solver-manager=neos --solver=<solver> mymodel.py`, where `mymodel.py` only has the model.  
- bonmin, couenne, and filmint didn't work at all.
- raposa is listed as a solver, but not recognized when solving with a `pyomo solve` call.  

Let me know if there's anything we can do to help.  One thing that could be causing problems is the return of solutions from NEOS -- we made some minor changes here and we may not be returning something that Pyomo expects.

carldlaird:
[...]

PR #1206, opened 2019-11-27
Fixing logic in NEOS

whart222:
New code for Python 3.x had clear errors.  The name of a function argument changed between the 2.x and 3.x
versions.

## Summary/Motivation:
This PR is a simple fix to an obvious error in new Python 3.x logic in NEOS.

- Simple changes.

### Legal Acknowledgement

### DIFF
--- a/pyomo/neos/kestrel.py
+++ b/pyomo/neos/kestrel.py
@@ -86,7 +86,7 @@ else: # Python 3.x
             self.proxy = urlparse(host)
             if not self.proxy.hostname:
                 # User omitted scheme from the proxy; assume http
-                self.proxy = urlparse('http://'+proxy)
+                self.proxy = urlparse('http://'+host)
 
         def make_connection(self, host):
             scheme = urlparse(host).scheme
--- a/pyomo/neos/kestrel.py
+++ b/pyomo/neos/kestrel.py
@@ -82,8 +82,8 @@ else: # Python 3.x
     # ProxiedTransport from Python 3.x documentation
     # (https://docs.python.org/3/library/xmlrpc.client.html)
     class ProxiedTransport(xmlrpclib.Transport):
-        def set_proxy(self, host):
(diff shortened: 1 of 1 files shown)


whart222:
@jsiirola This PR suggest that there is a problem with Pyomo's testing.  I thought that NEOS tests were being executed by default (e.g. when running test.pyomo).  And I thought we were testing Pyomo on Python 3.x.  But if so, then it isn't clear why those tests are passing, given this obvious bug.

jsiirola:
I don't understand what you are changing here - as indicated in the code comments, both `ProxiedTransport` classes were derived nearly verbatim from the respective Python 2 / 3 documentation (and the documentation changes the name of the argument between 2 and 3).  Regardless, that change has no impact (that I can find) on how our code runs.

The NEOS proxy interface is tested via Jenkins on all versions of Python.  I just looked, and it is running correctly and passing tests.

whart222:
On a new machine, this code failed on line 89, because the variable 'proxy' was not defined.

blnicho:
Is this related at all to #1201?

blnicho:
@jsiirola I think this is fixing a copy/paste error introduced in https://github.com/Pyomo/pyomo/commit/b4b1368a46f536ceb8bf5c0a890faf8a9f91b3ad. Maybe the better fix would be changing `proxy` to `host` on line 89?

```
        def set_proxy(self, host):
            self.proxy = urlparse(host)
            if not self.proxy.hostname:
                # User omitted scheme from the proxy; assume http
                self.proxy = urlparse('http://'+proxy)  # <-- Line 89
```
This bug wasn't revealed in our tests because line 89 is not covered.

jsiirola:
@blnicho: that makes sense.  I agree that the better fix would be just to change line 89.

blnicho:
@whart222 @jsiirola I made the small change to line 89 discussed above. I have no idea how to test this part of the code though.

PR #1208, opened 2019-11-27
NFC: GDP documentation update

qtothec:
## Summary/Motivation:
Documentation update for Pyomo.GDP. Includes formatting changes and describes TransformationFactory and GDPopt.

### Legal Acknowledgement

### DIFF
--- a/doc/OnlineDocs/modeling_extensions/gdp.rst
+++ b/doc/OnlineDocs/modeling_extensions/gdp.rst
@@ -82,8 +82,8 @@ If you are using a Python script, ``TransformationFactory`` accomplishes the sam
       have a variable “indicator_var” defined and attached to it.
       After that, it is just a Var like any other Var.
 
-    - The hull reformulation is an exact reformulation even for nonconvex models,
-      but the resulting MINLP will also be nonconvex.
+    - The hull reformulation is an exact reformulation at the solution points
+      even for nonconvex models, but the resulting MINLP will also be nonconvex.
 
 Direct GDP solvers
 ------------------
--- a/doc/OnlineDocs/modeling_extensions/gdp.rst
+++ b/doc/OnlineDocs/modeling_extensions/gdp.rst
@@ -70,10 +70,10 @@ If you are using a Python script, ``TransformationFactory`` accomplishes the sam
 
 .. note::
 
-    - all variables that appear in disjuncts need upper and lower bounds
+    - all variables that appear in disjuncts need upper and lower bounds for chull
 
     - for linear models, the BigM transform can estimate reasonably tight M
-      values for you
+      values for you if variables are bounded
 
     - for all other models, you will need to provide the M values through a
       “BigM” Suffix.
@@ -81,6 +81,9 @@ If you are using a Python script, ``TransformationFactory`` accomplishes the sam
     - When you declare a Disjunct, it (at declaration time) will automatically
       have a variable “indicator_var” defined and attached to it.
       After that, it is just a Var like any other Var.
+
+    - The hull reformulation is an exact reformulation even for nonconvex models,
+      but the resulting MINLP will also be nonconvex.
 
 Direct GDP solvers
 ------------------
--- a/doc/OnlineDocs/modeling_extensions/gdp.rst
+++ b/doc/OnlineDocs/modeling_extensions/gdp.rst
@@ -16,9 +16,8 @@ then they are Generalized Disjunctive Programming (GDP) models
 Disjunctions
 ============
 
-A disjunction is a set of collections of variables, parameters, and
-constraints that are linked by an OR (really exclusive or)
-constraint. The simplest case is a 2-term disjunction:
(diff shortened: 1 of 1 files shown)


doc/OnlineDocs/modeling_extensions/gdp.rst
```diff
@@ -61,32 +60,36 @@ the example returns.
 Transformation
 --------------
 
-In order to use the solvers currently available, one must convert the
-disjunctive model to a standard MIP/MINLP model.  The easiest way to do
-that is using the (included) BigM or Convex Hull transformations.  From
-the Pyomo command line, include the option ``--transform
-pyomo.gdp.bigm`` or ``--transform pyomo.gdp.chull``
+To use standard commercial solvers, you must convert the disjunctive model to a standard MIP/MINLP model.
+The two classical strategies for doing so are the (included) Big-M and Hull reformulations.
```
blnicho: Is there a reason you're calling it the Hull reformulation instead of Convex Hull? The transformation is invoked using `chull` which might be a little confusing if you just refer to it as the Hull reformulation.

qtothec: Yes, because in literature it appears as the Hull Reformulation. Practically speaking, it is possible to apply it to nonconvex functions, although the theory around this is a bit sparse.

bernalde: This addresses an issue that might be a little imprecision from the notation in Pyomo. The reformulation is named the hull reformulation because it only takes the convex hull from the elements in the disjunction (if well written) but does not yield the actual object popularly known as convex hull from the object (i.e. if we forget about the integrality condition we wouldn't necessarily get an integer solution).

emma58: I agree that the name may be misleading, but just to note: It is the true convex hull, but represented in a higher-dimensional space.

doc/OnlineDocs/modeling_extensions/gdp.rst
```diff
@@ -61,32 +60,36 @@ the example returns.
...
 
-Notes
------
+- ``TransformationFactory('gdp.bigm').apply_to(model)``
+- ``TransformationFactory('gdp.chull').apply_to(model)``
+
+.. note::
 
-Some notes:
+    - all variables that appear in disjuncts need upper and lower bounds
```
jsiirola: This is only true for thc HUll reformulation.  BigM only needs bounds if you are going to let us automatically determine M

bernalde: Even for nonlinear constraints?

doc/OnlineDocs/modeling_extensions/gdp.rst
```diff
@@ -61,32 +60,36 @@ the example returns.
...
 
-- for all other models, you will need to provide the M values through a
-  “BigM” Suffix.
+    - When you declare a Disjunct, it (at declaration time) will automatically
+      have a variable “indicator_var” defined and attached to it.
+      After that, it is just a Var like any other Var.
 
-- the convex hull reformulation is only valid for linear and convex
-  nonlinear problems.  Nonconvex problems are not supported (and are not
-  checked for).
```
jsiirola: Why are you removing this warning?

doc/OnlineDocs/modeling_extensions/gdp.rst
jsiirola: Correct.  BigM will happily relax the constraints without regard to the validity of the relaxation (if you provide an M value, we trust it is correct).

doc/OnlineDocs/modeling_extensions/gdp.rst
bernalde: I might have some results here that I would be happy to talk to you about. Even in that case, I would just leave the warning (this might go back to the whole convex hull vs. hull reformulation debate)

doc/OnlineDocs/modeling_extensions/gdp.rst
qtothec: Ok folks, I just checked with Ignacio yesterday. The Hull Reformulation is an exact reformulation from GDP into MINLP at the optimal solution, irregardless of convexity for the original GDP model constraints. The main complicating element of HR is the perspective function, but the function value is exact at indicator variable values of `y=1` and `y=0`. Therefore, its function value is exact for any feasible solutions of the GDP.

tldr: the perspective function of a nonconvex function will still be nonconvex, so the hull reformulation is still valid, but is not convex in the case of a GDP with nonconvex constraints.

blnicho: @qtothec so is this ready to merge?

qtothec: @blnicho Let me make a couple extra edits. I'll rerequest your review when I'm done.

Issue #859, opened 2019-02-08
Issue with slicers/References of unindxed components


andrewlee94:
If you create a slicer/Reference of an unindexed component (e.g. you were unsure if it was indexed or not), the resulting Reference is indexed by [None]. This means that when working with the new Reference the [None] index needs to be included, whilst working with the original component does not (i.e. the Reference has a different index to the thing it points to).

See example below:

```
m = ConcreteModel()
m.v = Var()

m.r = Reference(m.v[...])

m.r.display()

>>>> r : Size=1, Index=r_index
    Key  : Lower : Value : Upper : Fixed : Stale : Domain
    None :  None :  None :  None : False :  True :  Reals

m.r = 1

>>>> ValueError: Cannot set the value for the indexed component 'r' without specifying an index value.
	For example, model.r[i] = value
```

PR #1210, opened 2019-12-01
Compdata references

Robbybp:
## Summary/Motivation:
Would like to create references to data objects for an IDAES model DAE integrator. In particular, creating references at a particular time point to variables and constraints that are only indexed by time requires this. 
See: https://groups.google.com/d/topic/pyomo-forum/o_-RKlSlmvU/discussion

Note that this change would be unnecessary (for my purposes) if I could use start and stop values in a slice to create a reference to variables and constraints at time points within a finite element. 
See: https://groups.google.com/d/topic/pyomo-forum/xfS5fDCgm4U/discussion

-Allow Reference function to be called with VarData, ConstraintData, and BlockData objects as arguments
-For these objects, return a SimpleComponent object whose _data dict maps None to the data object to be referenced. If this object is created as:

```
m.add_component(ref_name, Reference(data_object))
```

It must be accessed as ```m.ref_name[None]``` instead of just ```m.ref_name```, which may lead to some confusion as, for variables, ```m.ref_name.value``` is distinct from ```m.ref_name[None].value```. I am not sure what other problems this may cause, for instance when sending variables to a solver/.nl file, but it seems to work for the integrator as I have it written. 

Opening the PR for discussion/feedback.

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/reference.py
+++ b/pyomo/core/base/reference.py
@@ -12,6 +12,9 @@ from pyutilib.misc import flatten_tuple
 from pyomo.common import DeveloperError
 from pyomo.core.base.sets import SetOf, _SetProduct, _SetDataBase
 from pyomo.core.base.component import Component, ComponentData
+from pyomo.core.base.var import SimpleVar, _GeneralVarData
+from pyomo.core.base.constraint import SimpleConstraint, _GeneralConstraintData
+from pyomo.core.base.block import SimpleBlock, _BlockData
 from pyomo.core.base.indexed_component import (
     IndexedComponent, UnindexedComponent_set
 )
@@ -518,10 +521,26 @@ def Reference(reference, ctype=_NotSpecified):
         pass
     elif isinstance(reference, Component):
         reference = reference[...]
+    elif isinstance(reference, _GeneralVarData):
+        obj = SimpleVar()
+        obj._costructed = True
+        obj.domain = reference.domain
+        obj._data = {None: reference}
+        return obj
+    elif isinstance(reference, _GeneralConstraintData):
+        obj = SimpleConstraint()
+        obj._costructed = True
+        obj._data = {None: reference}
+        return obj
+    elif isinstance(reference, _BlockData):
+        obj = SimpleBlock()
+        obj._costructed = True
+        obj._data = {None: reference}
+        return obj
     else:
         raise TypeError(
             "First argument to Reference constructors must be a "
-            "component or component slice (received %s)"
+            "component, component slice, or component data (received %s)"
             % (type(reference).__name__,))
 
     _data = _ReferenceDict(reference)
--- a/pyomo/core/base/reference.py
+++ b/pyomo/core/base/reference.py
@@ -523,18 +523,18 @@ def Reference(reference, ctype=_NotSpecified):
         reference = reference[...]
     elif isinstance(reference, _GeneralVarData):
         obj = SimpleVar()
-        obj._costructed = True
+        obj._constructed = True
         obj.domain = reference.domain
         obj._data = {None: reference}
         return obj
     elif isinstance(reference, _GeneralConstraintData):
         obj = SimpleConstraint()
-        obj._costructed = True
+        obj._constructed = True
         obj._data = {None: reference}
         return obj
     elif isinstance(reference, _BlockData):
         obj = SimpleBlock()
-        obj._costructed = True
+        obj._constructed = True
         obj._data = {None: reference}
         return obj
     else:
--- a/pyomo/core/base/reference.py
+++ b/pyomo/core/base/reference.py
@@ -517,33 +517,22 @@ def Reference(reference, ctype=_NotSpecified):
               4 :     1 :    10 :  None : False : False :  Reals
 
     """
-    is_compdata = False
     if isinstance(reference, _IndexedComponent_slice):
         pass
     elif isinstance(reference, Component):
         reference = reference[...]
     elif isinstance(reference, ComponentData):
-        index = reference.index()
-        parent = reference.parent_component()
-        if not isinstance(index, tuple):
-            index = (index,)
-        fixed = {i: val for i, val in enumerate(index)}
-        sliced = {}
-        ellipsis = None
-        reference = _IndexedComponent_slice(parent, fixed, sliced, ellipsis)
-        is_compdata = True
+        obj = reference.type()()
+        obj.construct()
+        obj._data = {None: reference}
+        return obj
     else:
         raise TypeError(
             "First argument to Reference constructors must be a "
             "component, component slice, or component data (received %s)"
             % (type(reference).__name__,))
 
-    if is_compdata:
-        # reference is a singleton slice referencing a ComponentData.
-        # This should be well-defined:
-        _data = {v.index(): v for v in reference}
-    else:
-        _data = _ReferenceDict(reference)
+    _data = _ReferenceDict(reference)
 
     _iter = iter(reference)
     if ctype is _NotSpecified:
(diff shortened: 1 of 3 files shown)


pyomo/core/base/reference.py
```diff
@@ -518,10 +521,26 @@ def Reference(reference, ctype=_NotSpecified):
...
+    elif isinstance(reference, _GeneralConstraintData):
+        obj = SimpleConstraint()
+        obj._constructed = True
+        obj._data = {None: reference}
+        return obj
+    elif isinstance(reference, _BlockData):
+        obj = SimpleBlock()
+        obj._constructed = True
+        obj._data = {None: reference}
+        return obj
```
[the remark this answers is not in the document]
Robbybp: I would like to know that as well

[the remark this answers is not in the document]
jsiirola: You could make this more general with something like
```
elif isinstance(reference, _ComponentData):
   obj = reference.type()()
   obj.construct()
   obj._data = {None: reference}
```

However, this PR is probably not the best solution: you are creating a new Component to hold the `_data` dict, but that component is never added to a model.  I think a better solution to this will be a way to explicitly create an `_IndexedComponent_slice` to the specific `_ComponentData`.  That is closer to what we do for IndexedComponents (the `[...]` notation in the immediately previous line creates a slice).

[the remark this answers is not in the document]
Robbybp: You're suggesting to create an ```_IndexedComponent_slice``` for the Component with all indices fixed? That seems like a better solution. In this case ```_data``` should map a singleton to the ```_ComponentData``` and ```index``` on line 575 should be a singleton set to index the new Component. When I try this, neither ```_ReferenceDict(reference)``` nor ```_ReferenceSet(reference)``` return the objects I want. I'll look into these functions and see if I can modify them appropriately.

The current implementation is creating a new Component to hold the ```_data``` dict as well, right? And the Component is only added to the model if the user does something like

```
m.r = Reference(comp)
```

or uses ```add_component```. Correct me if I'm wrong here.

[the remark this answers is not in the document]
jsiirola: @Robbybp: yes, I think creating an `_IndexedComponent_slice` is the right approach.  However, I don't think that it will work the way things are currently implemented.  The `_IndexedComponent_slice` object was never designed to work with just a single `_ComponentData`.  Supporting that would need logic that is akin to what is needed to support bounded slices - something that I originally explicitly excluded when building that class, but we should probably consider adding (for ordered Sets).

You are right about the current implementation as well - I missed that you are directly returning the component and not just assigning the `reference` variable.  The problem with your approach is that it ignores the ctype argument to the Reference function.

pyomo/core/base/reference.py
```diff
@@ -514,17 +517,34 @@ def Reference(reference, ctype=_NotSpecified):
...
+    elif isinstance(reference, ComponentData):
+        index = reference.index()
+        parent = reference.parent_component()
+        if not isinstance(index, tuple):
+            index = (index,)
+        fixed = {i: val for i, val in enumerate(index)}
+        sliced = {}
+        ellipsis = None
+        reference = _IndexedComponent_slice(parent, fixed, sliced, ellipsis)
+        is_compdata = True
```
Robbybp: @jsiirola I can create a slice at a singleton, but it doesn't behave as (I) expected with ```_ReferenceDict``` and ```_ReferenceSet```, so here I bypass these with the ```is_compdata``` flag and create my own ```_dict``` and ```index```. The result is accessed something like
```
m.r = Reference(m.v['a'])
m.r['a'].set_value(10)
```
This still creates some non-intuitive behavior though as the reference must be accessed through the index of the lowest-level container. E.g. for an ```IndexedVar``` in an ```IndexedBlock```
```
m.r2 = Reference(m.b[1].v1['b'])
m.r2['b'].set_value(10)
```
instead of
```
m.r2[1, 'b'].set_value(10)
```
as one would expect for a standard reference-to-slice:
```
m.r3 = Reference(m.b[:].v1[:])
m.r3[1, 'b'].set_value(10)
```
Not entirely sure how to fix this or if it's worth fixing. This still seems like a pretty superficial solution.

Robbybp: @jsiirola I can create a slice at a singleton, but it doesn't behave as (I) expected with ```_ReferenceDict``` and ```_ReferenceSet```, so here I bypass these with the ```is_compdata``` flag and create my own ```_dict``` and ```index```. The result is accessed something like
```
m.r = Reference(m.v['a'])
m.r['a'].set_value(10)
```
This still creates some non-intuitive behavior though as the reference must be accessed through the index of the lowest-level container. E.g. for an ```IndexedVar``` in an ```IndexedBlock```
```
m.r2 = Reference(m.b[1].v1['b'])
m.r2['b'].set_value(10)
```
instead of
```
m.r2[1, 'b'].set_value(10)
```
as one would expect for a standard reference-to-slice:
```
m.r3 = Reference(m.b[:].v1[:])
m.r3[1, 'b'].set_value(10)
```
Not entirely sure how to fix this or if it's worth fixing. This still seems like a pretty superficial solution.

pyomo/core/base/reference.py
```diff
@@ -518,13 +521,19 @@ def Reference(reference, ctype=_NotSpecified):
         pass
     elif isinstance(reference, Component):
         reference = reference[...]
+    elif isinstance(reference, ComponentData):
+        obj = reference.type()()
+        obj.construct()
+        obj._data = {None: reference}
+        return obj
```
Robbybp: I reverted to returning a SimpleComponent containing the ```ComponentData``` because the alternative broke my integrator somewhere in the nl file write. This is closer to how References are currently handled for SimpleComponents and for some reason doesn't break my integrator.

blnicho:
This PR is related to #859

pyomo/core/base/reference.py
```diff
@@ -12,6 +12,9 @@
 from pyomo.common import DeveloperError
 from pyomo.core.base.sets import SetOf, _SetProduct, _SetDataBase
 from pyomo.core.base.component import Component, ComponentData
+from pyomo.core.base.var import SimpleVar, _GeneralVarData
+from pyomo.core.base.constraint import SimpleConstraint, _GeneralConstraintData
+from pyomo.core.base.block import SimpleBlock, _BlockData
```
qtothec: With the reversion, you now have unused imports.

pyomo/core/base/reference.py
```diff
@@ -518,13 +521,19 @@ def Reference(reference, ctype=_NotSpecified):
...
+        return obj
     else:
         raise TypeError(
             "First argument to Reference constructors must be a "
-            "component or component slice (received %s)"
+            "component, component slice, or component data (received %s)"
             % (type(reference).__name__,))
 
     _data = _ReferenceDict(reference)
+
```
qtothec: Spurious newline

pyomo/core/tests/unit/test_reference.py
```diff
@@ -29,6 +29,7 @@
 )
 
 
+
```
qtothec: Spurious newline

pyomo/core/base/reference.py
Robbybp: Thanks

Robbybp:
> Code-wise, this looks fine. I don't know if it addresses your original concern about needing to index by `None`.

I have no problem indexing by `None`, but am not convinced this won't break things when being sent to a solver (although it seems to work for me, for now).

blnicho:
@Robbybp I'm going to close this. I'm pretty sure it's superseded by some of your more recent PRs in Pyomo and IDAES

PR #1221, opened 2019-12-12
Using NoArgumentGiven instead of *val for var fixing.

### LINKED ISSUE #1219: Var fix implementation
We have an interesting implementation of variable fixing to support both `var.fix()` and `var.fix(value)`:
https://github.com/Pyomo/pyomo/blob/ccde78cab8cabaf8822b66f741b63bd2797777d0/pyomo/core/base/var.py#L894

However, does it ever make sense to `var.fix(None)`? If not, it would be far more clear to implement as an optional argument.

qtothec:
## Summary/Motivation:
Replaces a functional but potentially confusing implementation with one that should be more clear.

- use a new NoArgumentGiven "static" object in `pyomo.common`
- add a simple README.md to `pyomo.common`

### Legal Acknowledgement


qtothec:
Yeah, I just changed one of the existing unit tests that covers the fixing code.

ghackebeil:
I approve as well. I’m just not logged into GitHub. No need to wait on a click from me.

PR #1222, opened 2019-12-12
BigM estimation for nonlinear expressions

qtothec:
# Summary/Motivation:
This is scratching an itch that I have had for a while now. We have had nonlinear interval arithmetic through both MC++ and FBBT for a while now. This PR puts those capabilities to use for BigM estimation in Pyomo.GDP.

There is surprisingly little that needed to be updated in the code.

### Legal Acknowledgement

### DIFF
--- a/doc/OnlineDocs/modeling_extensions/gdp.rst
+++ b/doc/OnlineDocs/modeling_extensions/gdp.rst
@@ -73,10 +73,14 @@ If you are using a Python script, ``TransformationFactory`` accomplishes the sam
     - all variables that appear in disjuncts need upper and lower bounds for chull
 
     - for linear models, the BigM transform can estimate reasonably tight M
-      values for you if variables are bounded
+      values for you if variables are bounded.
+
+    - for nonlinear models where finite expression bounds may be inferred from
+      variable bounds, the BigM transformation may also be able to automatically
+      compute M values for you.
 
     - for all other models, you will need to provide the M values through a
-      “BigM” Suffix.
+      “BigM” Suffix. A ``GDP_Error`` will be raised for missing M values.
 
     - When you declare a Disjunct, it (at declaration time) will automatically
       have a variable “indicator_var” defined and attached to it.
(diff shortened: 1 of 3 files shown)


bernalde:
An interesting edge case would be to see how it behaves on actual disjoint feasible regions (if nonconvex even better). Something like
$[x^2 + y^2 <= 2, x^3 + y^2 + xy >= 1/2] v [(x - 3)^2 + (y - 3)^2 <= 1]$
Computing a single M for the whole disjunction or without the nonconvex term is simple (it is 4 sqrt(2) - 1), but computing an individual M for each disjunct or even for each constraint (as Francisco proposed) might be tough.
The example is from [Fei's paper](https://aiche.onlinelibrary.wiley.com/doi/abs/10.1002/aic.16207) or a circle centered in (3,3), of radius 1.

qtothec:
@bernalde The assumption is that we have bounds on `x` and `y`. I do not see the difficulty of computing per-constraint bounds if we have bounded variables. Using bounds of `-4<=x<=4` and `-10<=y<=10`, I get as a result:

```
x**2 + y**2 - 114.0*(1 - Y_1) <= 2
1/2 <= x**3 + y**2 + x*y + 104.5*(1 - Y_1)
(x - 3)**2 + (y - 3)**2 - 217.0*(1 - Y_2) <= 1
```

Note that these are also "default" `M` values that are used if the user does not provide any. If you can calculate tighter `M`s, you can just give those to the transformation, right?

We are using interval arithmetic here, which doesn't rely on convexity.

jsiirola:
(because of the timing, merging will wait until after the 5.6.8 release process this morning)

qtothec:
@jsiirola Yeah, FBBT has turned out to be a quite useful functionality. I am actually a bit surprised that we didn't end up with circular reference issues, since FBBT imports from `Disjunct`

jsiirola:
The circular imports are avoided because the transformations are not imported automatically.  That is, importing `pyomo.gdp.disjunct` does not cause anything in `pyomo.gdp.plugins` to be imported, which is how FBBT can be cleanly imported.  This is part of the argument for the plugins infrastructure: various parts of the codebase can all contribute modeling components along with transformations, and Pyomo can first import all the components before attempting to import any of the transformations.

PR #1228, opened 2019-12-16
[WIP] Add indices to ComponentData

### LINKED ISSUE #1130: [PEP] Add _index slot to _ComponentData
## Proposal
This PEP proposes adding a slot to `_ComponentData` to store the index of the object within the parent container.

## Motivation
Pyomo originally had all `_ComponentData` objects contain a slot that held the index of the object (that is, the key within the parent Component's `_data` dictionary).  We removed that pointer when we were trying to slim down the Pyomo memory footprint, with the logic that the base Pyomo workflow didn't *need* to ever get the index from a data object.  We added a `index()` method that would look through the parent's `_data` dict and find the index, but that method involved a linear-time scan.

The most common use for getting the index appears to be getting the name of a data object.  To speed that up for the case where you were going to end up needing the names for most objects, we added a `name_buffer` argument to `getname()` that would generate and save *all* the names for a component when generating the first name.  In that way, while the first name was a linear time lookup, all subsequent names were O(1).  The challenge is that few developers know about the name_buffer, and few parts of the code make use of it.  The result is that portions of our code that should run in linear time actually scale quadratically (we most recently hit this in the BigM transformation).

## Discussion
I propose that we re-add the `_index` slot to `_ComponentData`.  By itself, this change will have a negative impact on overall Pyomo memory usage.  However, I believe that we can mitigate that impact through a subsequent redesign of the internal Pyomo data model that will reduce the amount of redundant data that must be stored on indexed component data objects (e.g., only storing a lower bound once if all the data objects share the same lower bound).  I have prototyped an approach that shows promise (see #202), but I would recommend deferring that discussion to a subsequent PEP.

qtothec:
## Summary/Motivation:
Add indices to ComponentData

This PR is waiting on:

- [ ] Other devs to check that the changes proposed here make sense (we touched a lot of files)
- [ ] Proper test cases to check performance implications

-
-

### Legal Acknowledgement


ghackebeil:
oops, didn't mean to click close

mrmundt:
#2351 replaces this PR.

PR #1238, opened 2019-12-18
Housekeeping

qtothec:
## Summary/Motivation:
Small housekeeping changes

### Legal Acknowledgement


qtothec:
Coverage is low because I moved `blockutils.py` from `core` to `util`, which added some more lines. I don't think `blockutils` had any coverage before. Would be fine with deprecating both and removing the single function if nobody is using them.

PR #1242, opened 2020-01-06
Add MatrixConstraint to AML interface.

ghackebeil:
Add a basic implementation of `pyomo.kernel.matrix_constraint` inrito the AML interface as `pyomo.core.base.matrix_constraint.MatrixConstraint`. Includes some minimal unit tests.

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/matrix_constraint.py
+++ b/pyomo/core/base/matrix_constraint.py
@@ -60,7 +60,7 @@ class _MatrixConstraintData(_ConstraintData):
         _index          The row index into the main coefficient matrix
     """
 
-    __slots__ = ('_index')
+    __slots__ = ('_index',)
 
     # the super secret flag that makes the writers
     # handle _MatrixConstraintData objects more efficiently
--- a/pyomo/core/base/matrix_constraint.py
+++ b/pyomo/core/base/matrix_constraint.py
@@ -1,0 +1,360 @@
+#  ___________________________________________________________________________
+#
+#  Pyomo: Python Optimization Modeling Objects
+#  Copyright 2017 National Technology and Engineering Solutions of Sandia, LLC
+#  Under the terms of Contract DE-NA0003525 with National Technology and 
+#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain 
+#  rights in this software.
+#  This software is distributed under the 3-clause BSD License.
+#  ___________________________________________________________________________
+
+
+import logging
+import weakref
+
+import pyutilib.misc
+from pyomo.core.base.set_types import Any
+from pyomo.core.expr.numvalue import value
+from pyomo.core.expr.numeric_expr import LinearExpression
+from pyomo.core.base.plugin import ModelComponentFactory
+from pyomo.repn.standard_repn import StandardRepn
+from pyomo.core.base.constraint import (IndexedConstraint,
+                                        _ConstraintData)
+
+import six
+from six.moves import xrange
+
+if six.PY3:
+    from collections.abc import Mapping as collections_Mapping
+else:
+    from collections import Mapping as collections_Mapping
+
+
+logger = logging.getLogger('pyomo.core')
+
+class _MatrixConstraintData(_ConstraintData):
+    """
+    This class defines the data for a single linear constraint
+        derived from a canonical form Ax=b constraint.
+
+    Constructor arguments:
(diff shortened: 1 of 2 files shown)


ghackebeil:
@carldlaird

ghackebeil:
@jsiirola raised a good point about this not working with abstract models. I'm not sure what to do about that.

carldlaird:
@blnicho 
> Overall I think this looks good but I would recommend expanding the docstring slightly to show how the MatrixConstraint would be used by a general user. The tests you added give a hint of how to use this component but I suspect this isn't how most users would/should use it.
While I agree, most of the intent of this component is that we would build custom constraints on top of it and document those for general users. Nevertheless, nothing wrong with documenting it.

carldlaird:
@ghackebeil Regarding comment from @jsiirola 

> @jsiirola raised a good point about this not working with abstract models. I'm not sure what to do about that.

I think this would primarily entail adding the possibility of a rule that returns the items that are currently passed into __init__ (from an interface standpoint).

For now, I am happy if it is simply documented as concrete only. I think we should find out how much we end up using it and the potential benefits before the next step of implementation.

jsiirola:
I agree with @carldlaird: I am not too concerned about supporting this for abstract models (the whole argument is efficiency, and bringing the data in through DAT files / DataPortals is more about convenience than efficiency.

pyomo/core/base/matrix_constraint.py
```diff
@@ -0,0 +1,360 @@
...
+    def set_value(self, expr):
+        """Set the expression on this constraint."""
+        raise NotImplementedError(
+            "MatrixConstraint row elements can not be updated"
+        )
+
+@ModelComponentFactory.register(
+                   "A set of constraint expressions in Ax=b form.")
+class MatrixConstraint(collections_Mapping,
+                       IndexedConstraint):
```
jsiirola: Can you comment on why you inherit from both collections_Mapping and IndexedConstraint?  In particular, you are having the collections API overwrite the IndexedConstraint API.  What "weirdness" are you attempting to circumvent?  If there are efficiency or implementational benefits of doing this, we should consider having IndexedComponent inherit from collections_Mapping...

pyomo/core/base/matrix_constraint.py
```diff
@@ -0,0 +1,360 @@
...
+        The row start-stop pointers of the CSR format sparse matrix
+    lb : list
+        The list of constraint lower bounds
+    ub : list
+        The list of constraint upper bounds
+    x : list
+        The list of pyomo variables mapped to their appropriate column
+    """
+
+    def __init__(self, A_data, A_indices, A_indptr, lb, ub, x):
```
jsiirola: I am a little concerned about the constructor API: For (most) Pyomo components, positional arguments are reserved for indexing sets and all other arguments must be passed by keyword argument.  This creates another exception to that rule (along with Piecewise, and to a certain extent, DerivativeVar)

pyomo/core/base/matrix_constraint.py
```diff
@@ -0,0 +1,360 @@
...
+            logger.debug("Constructing constraint %s"
+                         % (self.name))
+        if self._constructed:
+            return
+        self._constructed = True
+
+        ref = weakref.ref(self)
+        with pyutilib.misc.PauseGC():
+            self._data = tuple(_MatrixConstraintData(i, ref)
+                               for i in xrange(len(self._lower)))
```
jsiirola: Why prepopulate _data here?  You could instead overwrite `__getitem__` (since you are using collections_Mapping's dict API), or `_getitem_when_not_present()` (if you were using IndexedConstraint) in order to just create the data object on the fly when it was requested.  That way "intelligent" writers that looked at the Constraint Type could avoid the creation of the data objects altogether.

ghackebeil: The result of doing a big set of allocations like this with the GC disabled cuts the time from ~14s to ~4s for nrows=10,000,000. Given this significant speedup, and the fact that we don't yet have special handling of this component anywhere (so all _MatrixConstraintData objects _will_ be created), I see no need to make the implementation more complex.

pyomo/core/base/matrix_constraint.py
ghackebeil: This may just be a remnant of the implementation in kernel. I'll have to double check.

Issue #45, opened 2016-09-20
del_component issue when deleting constraints


tony121psu:
I'm working on a project that frequently requires solving a model, modifying a constraint, and then resolving. I've been using the del_component & add_component commands for removing and recreating constraints (see simple attached example). It appears that the del_component(constraint of interest) command is leaving behind some garbage if the constraint to be removed is indexed. This necessitates an additional command (i.e., "model.del_component(model.constraint1_index)") to avoid an error.

Is this a bug? Can it be fixed? Should I be doing something differently?

Thanks!
Tony Burgard (NETL)

[Simple Example.py.txt](https://github.com/Pyomo/pyomo/files/483458/Simple.Example.py.txt)

jsiirola:
Yes. I think that this behavior is a bug and can be corrected. I'll take a look and see how easy it is to fix. The only thing that I worry about is that it is possible that another component could rely on the implicit indexing set - but that would require some strange coding on the part of the modeler, so I think it should be safe to automatically remove the implicit sets that are created by indexed components.

There is also a workaround: if you create the indexed component using a single set that is already on the model, then no implicit sets will be created.

On Sep 20, 2016, at 1:30 PM, tony121psu <notifications@github.com<mailto:notifications@github.com>> wrote:


I'm working on a project that frequently requires solving a model, modifying a constraint, and then resolving. I've been using the del_component & add_component commands for removing and recreating constraints (see simple attached example). It appears that the del_component(constraint of interest) command is leaving behind some garbage if the constraint to be removed is indexed. This necessitates an additional command (i.e., "model.del_component(model.constraint1_index)") to avoid an error.

Is this a bug? Can it be fixed? Should I be doing something differently?

Thanks!
Tony Burgard (NETL)

Simple Example.py.txt<https://github.com/Pyomo/pyomo/files/483458/Simple.Example.py.txt>

-

tony121psu:
Thanks much for the fast response. It’d be definitely be very helpful if Pyomo could automatically remove the implicit sets created by the indexed components. Thanks also for the workaround. It seems to be working though it’ll be pretty tedious (especially compared to GAMS) to create a new set each time I want to use a modified constraint.

All the best,
Tony

jsiirola:
[...]

PR #1243, opened 2020-01-07
[WIP] With References check for equivalent sets instead of identical Set objects

### LINKED ISSUE #905: DAE transformation and Arc expansion are order dependent
I have been working with a a model using both a ContinuousSet and Arcs, and have discovered that the order in which the transformations are applied is critical in whether the model is constructed properly.

If I apply the DAE transformation first, and then expand the Arcs, I get the behaviour I would expect, where the Arc transformation writes equality constraints at all point in the ContinuousSet. However, if I apply the Arc expansion first, the equality constraints are only written at the points in the ContunuousSet that already exist, and no additional Constraints are written during the DAE expansion. This then results in unexpected degrees of freedom.

It seems clear that the DAE transformation is not interacting fully with the expanded Arcs for some reason, and I think this either needs to be fixed, or something done to make it clear that this will not work (an Exception or warning message?). Ultimately, I don't see a reason why these two should be order dependent, so I would express a preference for getting the trasnformations to work together properly.

blnicho:
## Summary/Motivation:
I think this will resolve an issue where for some types of models using both Pyomo.DAE and Pyomo.Network, the discretization and arc_expansion transformations had to be applied in a particular order. This makes a very small change in `reference.py` in how indexing sets are compared when iterating over the components in a slice. Instead of checking if the same Set objects are used this checks that **equivalent** sets are used. 

@andrewlee94 could you check if this change resolves the ordering problem for IDAES models?

- check for equivalent sets when creating Reference objects

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/reference.py
+++ b/pyomo/core/base/reference.py
@@ -410,7 +410,7 @@ def _identify_wildcard_sets(iter_stack, index):
         if len(index[i]) != len(level):
             return None
         # if any subset differs
-        if any(index[i].get(j,None) is not _set for j,_set in iteritems(level)):
+        if any(index[i].get(j,None) != _set for j,_set in iteritems(level)):
             return None
     return index
 
--- a/pyomo/core/tests/unit/test_reference.py
+++ b/pyomo/core/tests/unit/test_reference.py
@@ -540,7 +540,7 @@ class TestReference(unittest.TestCase):
         m.r = Reference(m.b[:].x[:])
 
         self.assertIs(m.r.type(), Var)
-        self.assertIs(type(m.r.index_set()), SetOf)
+        self.assertIs(type(m.r.index_set()), _SetProduct)
         self.assertEqual(len(m.r), 2*2)
         self.assertEqual(m.r[1,3].lb, 1)
         self.assertEqual(m.r[2,4].lb, 2)


pyomo/core/base/reference.py
```diff
@@ -410,7 +410,7 @@ def _identify_wildcard_sets(iter_stack, index):
         if len(index[i]) != len(level):
             return None
         # if any subset differs
-        if any(index[i].get(j,None) is not _set for j,_set in iteritems(level)):
+        if any(index[i].get(j,None) != _set for j,_set in iteritems(level)):
```
jsiirola: Can you comment on the need for this change?  Th new code will be slower (`is` is an O(1) operation, where `!=` in general is O(n)).  Also, by not using `is`, something like the following:
```
m.b = Block([1,2])
m.I = Set(initialize=[1,2,3])
m.J = Set(initialize=[1,2,3])
m.b[1].x = Var(m.I)
m.b[2].x = Var(m.J)
s = m.b[:].x[:]
```
will identify wildcard sets for `s` as `[m.b_index, m.I]`, even though `m.b[2].x` is indexed by `m.J` and not `m.I`.  When we originally wrote this logic, the idea was to only return wildcard sets if the indexing sets were demonstrably "known".

andrewlee94:
@blnicho This appears to fix the problem. I just tried running the IDAES Tutorial 3 under both your branch and the IDAES tag (or something close to it).

With your branch, the order of the transformations did not matter, whilst on master it did.

pyomo/core/base/reference.py
blnicho: This fixes the toy model in issue #905 where a model containing a Block constructed using a rule like the one below had to have the discretization transformation come before the arc_expansion transformation.

```
m = ConcreteModel()

m.s = ContinuousSet(initialize=[0,10])

def block_rule(b, i):
    b.v = Var(['a','b'], initialize=1)
    return

m.b1 = Block(m.s, rule=block_rule)
m.b2 = Block(m.s, rule=block_rule)

m.p1 = Port()
m.p2 = Port()

m.r1 = Reference(m.b1[:].v[:])
```

The problem is that `b1[0].v` and `b1[10].v` are not identified as having the same indexing set because a new Set is being implicitly created from the values of the Python list. This causes the ContinuousSet index in the Reference component `r1` to be obscured so the indexed equality constraint that is eventually added by the arc_expansion transformation is not explicitly indexed by a ContinuousSet and therefore not expanded by the discretization transformation. 

I realize this change violates the original logic which requires the user to be much more explicit when declaring their indexing sets but I also think that the above implementation is pretty common and most users would expect `b1[0].v` and `b1[10].v` to be identified as having the same indexing set (they shouldn't have to know that a Set object is being created for them implicitly)

It's possible that there are edge cases here where if we were creating References for Params or Constraints (which aren't necessarily dense) then checking `is not` vs. `!=` leads to more or less intuitive behavior but for Vars I think it makes sense to check if all the elements of the sets are the same. 

A couple other ideas for solving this problem of transformation ordering are:

1. In the discretization transformation check component indexing sets for the `SetOf` type and automatically try to expand the component if it is found (since it might contain an obscured ContinuousSet) - _I'm worried this could lead to weird side-effects if the user does anything weird with adding elements to non-ContinuousSets_

2. Make References or SetOf aware of ContinuousSets - _Seems weird to make core code aware of something in Pyomo.DAE_

3. Modify the wildcard identification functionality to preserve any indexing sets that are identified as the same even if it encounters ones that aren't the same at a different level - _I'm not sure how hard this would be to implement_

jsiirola: Let me add another option:
  4. Modify the wildcard identification so that it handles `SetOf differently (that is, use `is` for `Set` and `==` for `SetOf`).

That said, I think I like option 3.  Looking at the wildcard identification, I suspect that we *could* support a partial identification without too much hassle.  The challenge is that we would be potentially creating several pseudoset objects, and those objects would not be attached to the model.  Given that I am in favor of moving toward *not* requiring all set objects to be attached to the model (see #45), I don't see too much of a problem going down that path.  However, I think we can only handle "floating sets" correctly after we switch over to using the set rewrite (#326) by default.

blnicho:
We decided not to go with the solution proposed in this PR and this is no longer a high priority so I'm going to close it

PR #1250, opened 2020-01-12
Deprecating pyomo.contrib.simplemodel

whart222:
## Summary/Motivation:
Removing support for pyomo.contrib.simplemodel.  This package will be supported separately in pyomo_community

- Adding a deprecation warning that pyomo.contrib.simplemodel is no longer supported.

### Legal Acknowledgement


qtothec:
Did we already remove the auto-import of this contrib module from environ, etc? Otherwise, it looks like this code would generate a warning at import time.

PR #1256, opened 2020-01-15
[WIP] Units convert_value PR-1

carldlaird:
Implemented the convert_value method to convert numerical values using Pyomo units objects.

## Summary/Motivation:
- need a convert value method for working with constants containing Pyomo units.

- implemented the convert_value method on the PyomoUnitsContainer to allow
for conversion of numerical values based on Pyomo Units.
- Changed the singleton behavior to access the singleton units container through a module function.

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/importer.py
+++ b/pyomo/common/importer.py
@@ -1,0 +1,81 @@
+#  ___________________________________________________________________________
+#
+#  Pyomo: Python Optimization Modeling Objects
+#  Copyright 2017 National Technology and Engineering Solutions of Sandia, LLC
+#  Under the terms of Contract DE-NA0003525 with National Technology and
+#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain
+#  rights in this software.
+#  This software is distributed under the 3-clause BSD License.
+#  ___________________________________________________________________________
+
+import importlib
+
+class DeferredImportError(ImportError):
+    pass
+
+class ModuleUnavailable(object):
+    """Dummy object that raises a DeferredImportError upon attribute access
+
+    This object is returned by attempt_import() in liu of the module in
+    the case that the module import fails.  Any attempts to access
+    attributes on this object will raise a DeferredImportError
+    exception.
+
+    Parameters
+    ----------
+    message: str
+        The string message to return in the raised exception
+    """
+    def __init__(self, message):
+        self._error_message_ = message
+
+    def __getattr__(self, attr):
+        raise DeferredImportError(self._error_message_)
+
+def attempt_import(name, error_message=None, only_catch_importerror=True):
+    """Attempt to import the specified module.
+
+    This will attempt to import the specified module, returning a
+    (module, available) tuple.  If the import was successful, `module`
+    will be the imported module and `available` will be True.  If the
+    import results in an exception, then `module` will be an instance of
+    :py:class:`ModuleUnavailable` and `available` will be False
+
+    The following is equivalent to ``import numpy as np``:
+
+    .. doctest::
+
+       >>> from pyomo.common import attempt_import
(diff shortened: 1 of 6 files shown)


jsiirola:
Note also that the documentation builds are failing due to the API change from `pyo.units` to `pyo.units()`.

carldlaird:
@jsiirola Addressing your comment above where you provide sample code that would allow us to keep the module.units instead of module.units(). I don't think the complexity of that shell class is worth it just to prevent someone from accessing the singleton through a module function instead of a module property. It seems like the solution implemented (model.units()) is really straightforward to see what is happening.

@jsiirola Addressing your comment above about going to the full convert - it is not quite as simple as you illustrate, but it is close. I have the code and will work on making the change tomorrow (there needs to be some checking for offset units). I propose then, that we have two methods:
- convert takes in a Pyomo expression (hopefully with units) as well as "to_units" and returns the expression that does the conversion (close to your code above).
- convert_value takes in a numerical value or Pyomo expression, and must have a from_units and a to_units and just does the numerical conversion.

Maybe most people would just use "pyo.value(u.convert(expr, to_units=u.kg/u.s))" instead of convert_value?

carldlaird:
Thinking about it more, maybe there is no need for the convert_value method since wrapping the result with pyo.value basically does this anyway.

jsiirola:
I think the thing that bothers me about `pyo.units()` stems from the old guidance to make methods verbs (or verb phrases). It is not clear to me what action `units()` is doing without consulting the documentation.  

The simplicity of a module property was nice and by design, the wrapper class requires no maintenance (it will “automatically track” any future changes to the PyomoUnitsContainer API.

michaelbynum:
@jsiirola I am not convinced that your suggestion does not require maintenance. The code you suggested may cause future developers confusion. I also don't see a problem with having a function as long as the examples and documentation are clear.

michaelbynum:
Another simple options would be to not import units with pyomo.environ. If someone wants to import units, they can import it explicitly, and an ImportError would be raised if pint is not available.

carldlaird:
@michaelbynum That's not a bad idea. I have to look and see if the imports will be required in variables or parameters to support the next step. I will get back to you as soon as I can.

jsiirola:
Not pulling it into `pyomo.environ` would force users to start off with `from pyomo.core.base.units_container import units`, correct?

FWIW, the mock (duck-typed) interface it is a paradigm that we already use in one form or another in several places in Pyomo (Abstract classes in the Set rewrite, pyqt in `contrib.viewer`, and resolving Py2/Py3 API changes that are not handled by `six` are three that come to mind).

I went back and looked at the use of the `units` singleton, and I am 99% certain we can get away with a much simpler class:
```python
class PyomoUnitsContainer_noPint(object):
    def __getattr__(self, attr):
         raise RuntimeError(
            "The default `units` singleton is not available because the "
            "required Python package 'pint' could not be imported."
            " Please make sure you have 'pint' installed.")

if pint_module is not None:
    units = PyomoUnitsContainer()
else:
    units = PyomoUnitsContainer_noPint()
```

michaelbynum:
Is there something wrong with requiring `from Pyomo.core.base.units_container import units`?

michaelbynum:
@jsiirola I do like this simpler class much better.

jsiirola:
@michaelbynum: Only that it would be the first "significant" piece of functionality that a normal user might run across that would require a direct import from inside of `base`.

carldlaird:
What if we put the import at pyomo.units? Then units would look like other packages and this would also have the benefit that these imports were not executed if units is not used.

If you think this is a good idea, what is the best way to implement? Put a units folder at the main level with a __init__? If so, we could also move the code here.

Thoughts?

PR #1265, opened 2020-01-17
Cplex direct fix

### LINKED ISSUE #1260: `CPLEXDirect` (and `CPLEXPersistent`) do not support v12.10
1. Install `cplex==12.10.0`
2. Run a model using the persistent solver
3. The following error occurs

```
  File "...venv/lib/python3.7/site-packages/cplex/_internal/_ostream.py", line 66, in __init__
    raise TypeError("outputfile must have a write method")
TypeError: outputfile must have a write method
```

This is due to `pyomo/solvers/plugins/solvers/cplex_direct.py` line 112, (in `_apply_solver`) using a call to `set_results_stream()` that is now deprecated. A deprecation warning was raised in version 12.9.

The fix looks pretty straightforward; `CPLEXDirect._log_file` is a string path whereas `Cplex.set_results_stream()` expects a file-like object.

michaelbynum:
## Summary/Motivation:
This PR updates the CPLEX direct interface to support CPLEX 12.10. It also adds 

`pip install cplex`

to the GitHub workflow so the cplex direct interface can be tested.

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/ubuntu_python_matrix_test.yml
+++ b/.github/workflows/ubuntu_python_matrix_test.yml
@@ -29,7 +29,7 @@ jobs:
         pip install numpy scipy ipython openpyxl sympy pyodbc pyyaml networkx xlrd matplotlib dill
         pip install pandas # Pandas needs to be installed after its dependencies to work correctly for Python 2.7 
         pip install seaborn pymysql pyro4 pint pathos
-        pip install cplex
+        pip install cplex || echo "CPLEX Community Edition is not available for ${{ matrix.python-version }}"
         echo "Installing GAMS..."
         wget -q https://d37drm4t2jghv5.cloudfront.net/distributions/24.8.5/linux/linux_x64_64_sfx.exe
         chmod +x linux_x64_64_sfx.exe
--- a/.github/workflows/ubuntu_python_matrix_test.yml
+++ b/.github/workflows/ubuntu_python_matrix_test.yml
@@ -28,7 +28,8 @@ jobs:
         echo "Install extras..."
         pip install numpy scipy ipython openpyxl sympy pyodbc pyyaml networkx xlrd matplotlib dill
         pip install pandas # Pandas needs to be installed after its dependencies to work correctly for Python 2.7 
-        pip install seaborn pymysql pyro4 pint pathos 
+        pip install seaborn pymysql pyro4 pint pathos
+	pip install cplex
         echo "Installing GAMS..."
         wget -q https://d37drm4t2jghv5.cloudfront.net/distributions/24.8.5/linux/linux_x64_64_sfx.exe
         chmod +x linux_x64_64_sfx.exe
--- a/.github/workflows/ubuntu_python_matrix_test.yml
+++ b/.github/workflows/ubuntu_python_matrix_test.yml
@@ -29,7 +29,7 @@ jobs:
         pip install numpy scipy ipython openpyxl sympy pyodbc pyyaml networkx xlrd matplotlib dill
         pip install pandas # Pandas needs to be installed after its dependencies to work correctly for Python 2.7 
         pip install seaborn pymysql pyro4 pint pathos
-	pip install cplex
+        pip install cplex
         echo "Installing GAMS..."
         wget -q https://d37drm4t2jghv5.cloudfront.net/distributions/24.8.5/linux/linux_x64_64_sfx.exe
         chmod +x linux_x64_64_sfx.exe
--- a/pyomo/solvers/plugins/solvers/cplex_direct.py
+++ b/pyomo/solvers/plugins/solvers/cplex_direct.py
@@ -105,13 +105,16 @@ class CPLEXDirect(DirectSolver):
             for block in self._pyomo_model.block_data_objects(descend_into=True, active=True):
                 for var in block.component_data_objects(ctype=pyomo.core.base.var.Var, descend_into=False, active=True, sort=False):
                     var.stale = True
+        _log_file = self._log_file
+        if (self.version()[0] > 12) or (self.version()[0] == 12 and self.version()[1] >= 10):
+            _log_file = open(self._log_file, 'w')
         if self._tee:
             def _process_stream(arg):
                 sys.stdout.write(arg)
                 return arg
-            self._solver_model.set_results_stream(self._log_file, _process_stream)
+            self._solver_model.set_results_stream(_log_file, _process_stream)
         else:
-            self._solver_model.set_results_stream(self._log_file)
+            self._solver_model.set_results_stream(_log_file)
 
         if self._keepfiles:
             print("Solver log file: "+self._log_file)
@@ -181,6 +184,8 @@ class CPLEXDirect(DirectSolver):
         self._solver_model.solve()
         t1 = time.time()
         self._wallclock_time = t1 - t0
+        if (self.version()[0] > 12) or (self.version()[0] == 12 and self.version()[1] >= 10):
+            _log_file.close()
 
         # FIXME: can we get a return code indicating if CPLEX had a significant failure?
         return Bunch(rc=None, log=None)
--- a/pyomo/solvers/plugins/solvers/cplex_direct.py
+++ b/pyomo/solvers/plugins/solvers/cplex_direct.py
@@ -106,86 +106,88 @@ class CPLEXDirect(DirectSolver):
                 for var in block.component_data_objects(ctype=pyomo.core.base.var.Var, descend_into=False, active=True, sort=False):
                     var.stale = True
         _log_file = self._log_file
-        if (self.version()[0] > 12) or (self.version()[0] == 12 and self.version()[1] >= 10):
+        if self.version() >= (12, 10):
             _log_file = open(self._log_file, 'w')
-        if self._tee:
-            def _process_stream(arg):
-                sys.stdout.write(arg)
-                return arg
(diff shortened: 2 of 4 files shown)


ghackebeil:
Does pip installed Cplex automatically use a trial license these days?

ghackebeil:
Maybe it always has, and we didn’t do it on the Travis containers because it would trigger PySP test failures. But if we’re only running smoke tests on these then it’s not a problem.

Is main solver test suite running on these containers? Because that would be nice.

ruaridhw:
> Does pip installed Cplex automatically use a trial license these days?

Yes, I believe so. To use a fully licensed package you should be installing it from the package that [comes with your cplex installation](https://www.ibm.com/support/knowledgecenter/SSSA5P_12.7.1/ilog.odms.cplex.help/CPLEX/GettingStarted/topics/set_up/Python_setup.html).

pyomo/solvers/plugins/solvers/cplex_direct.py
```diff
@@ -105,13 +105,16 @@ def _apply_solver(self):
             for block in self._pyomo_model.block_data_objects(descend_into=True, active=True):
                 for var in block.component_data_objects(ctype=pyomo.core.base.var.Var, descend_into=False, active=True, sort=False):
                     var.stale = True
+        _log_file = self._log_file
+        if (self.version()[0] > 12) or (self.version()[0] == 12 and self.version()[1] >= 10):
```
jsiirola: Can we just do `self.version()[:2] >= (12,10)` for clarity?

.github/workflows/ubuntu_python_matrix_test.yml
```diff
@@ -28,7 +28,8 @@ jobs:
         echo "Install extras..."
         pip install numpy scipy ipython openpyxl sympy pyodbc pyyaml networkx xlrd matplotlib dill
         pip install pandas # Pandas needs to be installed after its dependencies to work correctly for Python 2.7 
-        pip install seaborn pymysql pyro4 pint pathos 
+        pip install seaborn pymysql pyro4 pint pathos
+        pip install cplex
```
jsiirola: Is there a reason to make this a separate `pip install` command?

pyomo/solvers/plugins/solvers/cplex_direct.py
```diff
@@ -181,6 +184,8 @@ def _process_stream(arg):
         self._solver_model.solve()
         t1 = time.time()
         self._wallclock_time = t1 - t0
+        if (self.version()[0] > 12) or (self.version()[0] == 12 and self.version()[1] >= 10):
```
jsiirola: Same as above: can we just do `self.version()[:2] >= (12,10)`?

pyomo/solvers/plugins/solvers/cplex_direct.py
ruaridhw: This should do the same thing:

```suggestion
        if self.version() >= (12, 10):
```
using tuple comparison.

pyomo/solvers/plugins/solvers/cplex_direct.py
```diff
@@ -105,13 +105,16 @@ def _apply_solver(self):
             for block in self._pyomo_model.block_data_objects(descend_into=True, active=True):
                 for var in block.component_data_objects(ctype=pyomo.core.base.var.Var, descend_into=False, active=True, sort=False):
                     var.stale = True
+        _log_file = self._log_file
+        if (self.version()[0] > 12) or (self.version()[0] == 12 and self.version()[1] >= 10):
+            _log_file = open(self._log_file, 'w')
```
jsiirola: As a general question, can we do this in a `try ... finally` block so that we can guarantee that the _log_file is closed (even in the case of exception)?

pyomo/solvers/plugins/solvers/cplex_direct.py
ruaridhw: You shouldn't even need the `[:2]` if that's cleaner 😉 

```python
assert (12, 7, 1) < (12, 10)
```

jsiirola:
Following up detecting CPLEX license status, it appears that there isn't an easy way to check if the installed version of cplex is community or full.  See https://stackoverflow.com/questions/48446468/cplex-how-can-i-programmatically-check-whether-a-user-is-running-the-community

.github/workflows/ubuntu_python_matrix_test.yml
jsiirola: (answering my own question): yes.  This needs to be it's own command, as cplex is not available for all versions of python (notably not 3.8), so we need to update this so that failure to install cplex does not necessarily kill the job.  Something like this might work:
```
pip install cplex || python -c "import sys; assert sys.version_info[:2] in ((3,4), (3,8))"
```

jsiirola: That said, it may not be that useful to perform the version check.  The following is probably sufficient:
```
pip install cplex || echo "CPLEX Community Edition is not available for ${{ matrix.python-version }}"
```

We should also propagate this to the Windows and OSX github actions as well.

michaelbynum:
> Maybe it always has, and we didn’t do it on the Travis containers because it would trigger PySP test failures. But if we’re only running smoke tests on these then it’s not a problem.
> 
> Is main solver test suite running on these containers? Because that would be nice.

@ghackebeil I can't remember what smoke tests are. I believe the main solver test suite is running. At the very least, I am now seeing some cplex direct and cplex persistent tests running from `pyomo.solvers.tests.checks`. Is that what you mean by the main solver test suite?

pyomo/solvers/plugins/solvers/cplex_direct.py
michaelbynum: @ruaridhw Thanks. Good suggestion.

.github/workflows/ubuntu_python_matrix_test.yml
michaelbynum: @jsiirola Thanks for the help here.

pyomo/solvers/plugins/solvers/cplex_direct.py
michaelbynum: @jsiirola Good suggestion

michaelbynum:
> Following up detecting CPLEX license status, it appears that there isn't an easy way to check if the installed version of cplex is community or full. See https://stackoverflow.com/questions/48446468/cplex-how-can-i-programmatically-check-whether-a-user-is-running-the-community

@jsiirola Thanks for looking into this.

michaelbynum:
@mrmundt Can you help me update the GitHub workflow for Windows, please? I need to add something along the lines of

`conda install -q -y --no-update-deps -c ibmdecisionoptimization cplex`

jsiirola:
> > Following up detecting CPLEX license status, it appears that there isn't an easy way to check if the installed version of cplex is community or full. See https://stackoverflow.com/questions/48446468/cplex-how-can-i-programmatically-check-whether-a-user-is-running-the-community
> 
> @jsiirola Thanks for looking into this.

I think the following is the fastest/easiest implementation to check the license:
```python
import cplex
try:
    cpx = cplex.Cplex()
    cpx.variables.add(lb=[0]*1001)
    cpx.solve()
    license_valid = True
except cplex.exceptions.CplexSolverError as err:
    # assert err.args[2] == cplex.exceptions.error_codes.CPXERR_RESTRICTED_VERSION
    license_valid = False
```

mrmundt:
> @mrmundt Can you help me update the GitHub workflow for Windows, please? I need to add something along the lines of
> 
> `conda install -q -y --no-update-deps -c ibmdecisionoptimization cplex`

@michaelbynum - Sure thing. I added the flag and pushed the commit for it.

michaelbynum:
@mrmundt Thanks!

mrmundt:
> @mrmundt Thanks!

Anytime! By the way, it looks like CPlex isn't compatible yet with Python 3.8. I tried a 3.8 environment on my machine and got this error:
```
UnsatisfiableError: The following specifications were found                                                                                           
to be incompatible with the existing python installation in your environment:

Specifications:

  - cplex -> python[version='2.7.*|3.5.*|>=2.7,<2.8.0a0|>=3.5,<3.6.0a0|>=3.6,<3.7.0a0|>=3.7,<3.8.0a0']

Your python: python=3.8
```

michaelbynum:
@mrmundt Sorry. I ran into this for the linux tests too. For linux, @jsiirola suggested 
```
pip install cplex || echo "CPLEX Community Edition is not available for ${{ matrix.python-version }}"
```
which worked well. Can we do something similar for Windows?

mrmundt:
> @mrmundt Sorry. I ran into this for the linux tests too. For linux, @jsiirola suggested
> 
> ```
> pip install cplex || echo "CPLEX Community Edition is not available for ${{ matrix.python-version }}"
> ```
> 
> which worked well. Can we do something similar for Windows?

@michaelbynum Absolutely! I added a `try-catch` block for CPLEX.

ruaridhw:
> By the way, it looks like CPlex isn't compatible yet with Python 3.8

@mrmundt, yes, the `setup.py` in `cplex=12.10.0.0` specifically checks you're using 3.6 or 3.7. Annoyingly we've found that despite this enforcement, newer versions of Python can still work perfectly; I think this was the case with 12.7.1 and 3.7 from memory...

michaelbynum:
I will re-open this PR shortly.

PR #1266, opened 2020-01-21
Deprecate 'Any' being the default Param domain.

### LINKED ISSUE #1185: Default domain for mutable Params
Suppose I have:
```
from pyomo.environ import *

m = ConcreteModel()

m.x = Var(bounds=(0, 10))
m.param = Param(initialize=5, mutable=True)
m.cons = Constraint(expr=m.x >= m.param)
m.obj = Objective(expr=m.x)

m.param = '10' #oops

# Now the error we get doesn't tell us where or what the problem is.
SolverFactory('cplex').solve(m)
# TypeError: unsupported operand type(s) for -: 'str' and 'int'
```
This is dumb on my part, but I doubt I'll be the last. I think the question is what the default domain for Params should be. Right now it looks like it is Any. So I could have saved myself by doing:
```
m.param = Param(initialize=5, mutable=True, within=Reals)
```
and I would get a good error message when I tried to set `m.param` to a string. But my concern is that the `TypeError` certainly isn't what led me to that answer. And in addition, this seems inconsistent because the writer is clearly expecting a number, but, by default, Param doesn't validate that its value is a number.

jsiirola:
## Summary/Motivation:
This deprecates, but does not change, Any being the default domain for Param objects.  The deprecation warning is only issued when non-real values are encountered.

- Issue a deprecation warning when encountering non-real values on Params using the default `domain` value.

### Legal Acknowledgement


jsiirola:
All tests are passing and this is ready for review.

PR #1267, opened 2020-01-23
GDPopt-LBB set default BigM

qtothec:
## Summary/Motivation:
Set a token Big M value for re-activated nonlinear constraints so that Big M transformation does not have to calculate an M (and also therefore does not complain when it is unable to do so).
It is fine to use M = 1 arbitrarily, because Y is fixed to True for these disjuncts. Hence, `g(x) <= M(1-y)` becomes `g(x) <= M(1-1)` = 0.

### Legal Acknowledgement


qtothec:
Actually, going to combine a few PRs together. Closing this.

PR #1271, opened 2020-01-27
Correct mistake in pynumero nlp sensitivity example

### LINKED ISSUE #1269: bug in pynumero example: sensitivity.py
It seems like the [sensitivity.py](https://github.com/Pyomo/pyomo/blob/master/pyomo/contrib/pynumero/examples/sensitivity.py) code is not calculating the sensitivity correctly. The sensitivity based x_new is a bit different from the result obtained by solving the perturbed NLP in the code. The cause of this seems to be the creation of `Np`

```
Np = BlockMatrix(2, 1)
Np[0, 0] = nlp.extract_submatrix_hessian_lag(pyomo_variables_rows=nlp.get_pyomo_variables(), pyomo_variables_cols=[m.eta1, m.eta2])
Np[1, 0] = nlp.extract_submatrix_jacobian(pyomo_variables=[m.eta1, m.eta2], pyomo_constraints=nlp.get_pyomo_constraints())
```
It seems that in this example Np should be 9x2 matrix with all zeros except elements (8, 1) and (9, 2) which should be one (based on eq 13 in this [reference paper](http://www.optimization-online.org/DB_FILE/2011/04/3008.pdf)). 

I have a [jupyter notebook](https://github.com/notesofdabbler/learn_pyomo/blob/master/wk_pynumero/wk_sensitivity.ipynb) that recreates this example and the proposed modification which seems to fix the issue.

santiagoropb:
## Summary/Motivation:
This PR fixes a mistake in an example in pyomo.contrib.pynumero.examples.sensitivity.

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/pynumero/examples/sensitivity.py
+++ b/pyomo/contrib/pynumero/examples/sensitivity.py
@@ -81,9 +81,6 @@ y = compute_init_lam(nlp, x=x)
 nlp.set_primals(x)
 nlp.set_duals(y)
 
-#J = nlp.evaluate_jacobian()
-#H = nlp.evaluate_hessian_lag()
-
 J = nlp.extract_submatrix_jacobian(pyomo_variables=[m.x1, m.x2, m.x3], pyomo_constraints=[m.const1, m.const2])
 H = nlp.extract_submatrix_hessian_lag(pyomo_variables_rows=[m.x1, m.x2, m.x3], pyomo_variables_cols=[m.x1, m.x2, m.x3])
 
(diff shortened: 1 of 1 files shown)


michaelbynum:
I am waiting for the Jenkins tests to run before merging this.

blnicho:
It looks like the Jenkins autotester has been failing for a few days now and it's not clear to me from the console output why that is happening. I think we can go ahead and merge this.

carldlaird:
@santiagoropb Nice. I like what you did with the print statement at the end - I assume that this was to ensure the correct ordering for the print?

PR #1272, opened 2020-01-27
Raise error on failed Param validation

### LINKED ISSUE #930: validate does not raise Exception
When `AbstractModel.create_instance` is called, if the `validate` function of a Parameter returns `False`, the value is not assigned but no Exception is raised.

Example:
```python
from pyomo import environ as pe

m = pe.AbstractModel()
    
m.p = pe.Param(validate = lambda m,v: False)
    
data = {None:{
    'p': {None: 0.2}}}

cm = m.create_instance(data=data)
```

does not raise any error. When one tries to retrieve the value though, `ValueError` is raised because the value has not been set.

```python
>>> cm.p.value
ValueError: Error retrieving immutable Param value (p):
	The Param value is undefined and no default value is specified.
```

There might be reasons not to raise an exception, but failing to set a value silently can create a lot of confusion.

Besides, the [docs](https://pyomo.readthedocs.io/en/latest/pyomo_modeling_components/Parameters.html?#parameters) clearly say:
> If a value is provided that is less than that, the model instantation would be terminated and an error message issued

Imho, validation is a great feature but I understand it introduces a lot of challenges.
If the functionality does not behave as expected, it should at least be (temporarily, hopefully) deprecated, or issue a warning.

goroderickgo:
## Summary/Motivation:
As noted in #930, the `validate` option for Params does not raise an error when the validation function returns False (i.e., the functionality described in [the documentation](https://pyomo.readthedocs.io/en/stable/pyomo_modeling_components/Parameters.html) does not work as described). 

- Raise the error that was caught in the `try/except` block. This proposed fix is modeled after the pattern used for the methods with the same name (`_setitem_when_not_present`) in for [IndexedComponent](https://github.com/goroderickgo/pyomo/blob/69b2e7bf9b4422750fba108a04d0d3216be02805/pyomo/core/base/indexed_component.py#L715-L736) and [Var](https://github.com/goroderickgo/pyomo/blob/69b2e7bf9b4422750fba108a04d0d3216be02805/pyomo/core/base/var.py#L629-L641).

I'm not very familiar with adding tests, so I would gladly take guidance if I need to add any.

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/param.py
+++ b/pyomo/core/base/param.py
@@ -601,6 +601,7 @@ class Param(IndexedComponent):
                 return value
         except:
             del self._data[index]
+            raise
 
 
     def _validate_value(self, index, value, validate_domain=True):
--- a/pyomo/core/tests/unit/test_param.py
+++ b/pyomo/core/tests/unit/test_param.py
@@ -1169,6 +1169,58 @@ class MiscParamTests(unittest.TestCase):
             return 0.0
         model.p = Param(model.A, initialize=rule)
 
+    def test_param_validate(self):
+        """Test Param `validate` and `within` throw ValueError when not valid.
+
+        The `within` argument will catch the ValueError, log extra information
+        with of an "ERROR" message, and reraise the ValueError.
+
+        1. Immutable Param (unindexed)
+        2. Immutable Param (indexed)
+        3. Immutable Param (arbitrary validation rule)
+        4. Mutable Param (unindexed)
+        5. Mutable Param (indexed)
+        6. Mutable Param (arbitrary validation rule)
+        """
+        def validation_rule(model, value):
+            """Arbitrary validation rule that always returns False."""
+            return False
+
+        # 1. Immutable Param (unindexed)
+        with self.assertRaisesRegex(ValueError, "Value not in parameter domain"):
+            m = ConcreteModel()
+            m.p1 = Param(initialize=-3, within=NonNegativeReals)
+
+        # 2. Immutable Param (indexed)
+        with self.assertRaisesRegex(ValueError, "Value not in parameter domain"):
+            m = ConcreteModel()
+            m.A = RangeSet(1, 2)
+            m.p2 = Param(m.A, initialize=-3, within=NonNegativeReals)
+
+        # 3. Immutable Param (arbitrary validation rule)
+        with self.assertRaisesRegex(ValueError, "Invalid parameter value"):
+            m = ConcreteModel()
+            m.p5 = Param(initialize=1, validate=validation_rule)
+
+        # 4. Mutable Param (unindexed)
+        with self.assertRaisesRegex(ValueError, "Value not in parameter domain"):
+            m = ConcreteModel()
+            m.p3 = Param(within=NonNegativeReals, mutable=True)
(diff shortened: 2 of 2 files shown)


blnicho:
I tested this locally and it seems to fix things. I don't think this will have any unintended side effects but we should wait for @jsiirola to take a look before merging. 

For tests, take a look at https://github.com/Pyomo/pyomo/blob/master/pyomo/core/tests/unit/test_param.py. I think we just need to add a method under `MiscParamTests` that declares a few different types of Params (indexed/nonindexed, mutable/immutable) and asserts that a ValueError is raised when the Param is set to an invalid value. The syntax for that is something like: 

    with self.assertRaises(ValueError):
        m.p = -3

goroderickgo:
Thanks @blnicho! Agreed on waiting for @jsiirola, just figured I’d take a stab at fixing the issue. I’ll take a look at the tests soon.

jsiirola:
I think this solution looks fine, and will approve once we have some tests to make sure that this doesn't sneak back in.  To add to what @blnicho said, I generally prefer that we verify the exception message as well as the type (especially when we are looking for a common type like `ValueError`) - I have gotten burned before by just looking at the type: the test was passing because the right type was being raised, but it wasn't the exception I was expecting.

In this case, something like
```python
with self.assertRaisesRegexp(ValueError, "Value failed parameter validation rule"):
    m.p = -3
```
would be good.

goroderickgo:
Hi everyone, I could use a bit more guidance on the tests:

1. If we initialize a Param like so: `m.x = Param(initialize=-3, within=NonNegativeReals)`, it also raises a ValueError, but it's wrapped in a more generic "ERROR" that doesn't pass the validation test. Seems like I should be testing for this, but I'm not sure how, since "ERROR" is not a defined type of exception.

    ```
    ERROR: Constructing component 'p1' from data=None failed:
            ValueError: Invalid parameter value: p1[None] = '-3', value
            type=<class 'int'>.
            Value not in parameter domain NonNegativeReals
    ```

2. I'm on Python 3.8 and am getting a deprecation warning for `self.assertRaisesRegexp` (see below). Should I use `self.assertRaisesRegex` as it suggests instead?

    ```
    DeprecationWarning: Please use assertRaisesRegex instead.
    ```

Thanks!

jsiirola:
@goroderickgo: for Q1, while Block (and Model) will catch the `ValueError` exception and log additional context, the original exception is re-raised and you can catch it as normal.  So, something like the following should work:
```python
with self.assertRaisesRegexp(ValueError, "Value not in parameter domain NonNegativeReals"):
    m.p = Param(initialize=-3, domain=NonNegativeReals)
```

For Q2: until PyUtilib/pyutilib#79 is accepted and merged in, using `assertRaisesRegex` will cause Python 2.7 tests to fail (which would prevent merging this PR).  My recommendation would be patience - I expect that PR to be reviewed and merged in about a day.  After that, using `assertRaisesRegex` will be the preferred syntax for all of Pyomo.

blnicho:
@goroderickgo I merged the PyUtilib PR so you should be fine to move ahead with the `assertRaisesRegex` syntax.

PR #1273, opened 2020-01-28
PyNumero support on Windows

### LINKED ISSUE #1261: PyNumero support on Windows
We need to make sure PyNumero installs/runs on Windows and make sure any extra installation steps are well documented. Some discussion on this was started in #1253. When this is resolved we also need to make sure to enable the tests currently being skipped on Windows.

blnicho:
## Summary/Motivation:
@santiagoropb updated the `pynumero_libraries` conda package to support Windows. This PR makes a small change in PyNumero to get things working and enables all of the tests that were disabled by default on Windows. Technically this only adds Windows support for Anaconda users but I think that is enough to close #1261. 

@mrmundt could you help me add the `pynumero_libraries` package from conda-forge to the Windows Github Actions build? 

I also need to double check which PyNumero tests are being skipped on our other testing platforms and figure out how to update/install the PyNumero libraries when we're not using conda.

- Change to the library finder in PyNumero to support Windows library names
- Enable all PyNumero tests on Windows

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/pynumero/extensions/utils.py
+++ b/pyomo/contrib/pynumero/extensions/utils.py
@@ -17,6 +17,11 @@ def find_pynumero_library(library_name):
     asl_path = find_library(library_name)
     if asl_path is not None:
         return asl_path
+
+    # On windows the library is prefixed with 'lib'
+    asl_path = find_library('lib'+library_name)
+    if asl_path is not None:
+        return asl_path
     else:
         # try looking into extensions directory now
         file_path = os.path.abspath(__file__)
--- a/.github/workflows/win_python_matrix_test.yml
+++ b/.github/workflows/win_python_matrix_test.yml
@@ -22,7 +22,7 @@ jobs:
       with:
         auto-update-conda: true
         python-version: ${{ matrix.python-version }}
-    - name: Install dependencies
+    - name: Install Pyomo dependencies
       shell: pwsh
       run: |
         $env:PYTHONWARNINGS="ignore::UserWarning"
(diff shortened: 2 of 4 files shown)


pyomo/contrib/pynumero/extensions/utils.py
```diff
@@ -15,6 +15,11 @@
 def find_pynumero_library(library_name):
 
     asl_path = find_library(library_name)
+    if asl_path is not None:
+        return asl_path
+
+    # On windows the library is prefixed with 'lib'
+    asl_path = find_library('lib'+library_name)
```
carldlaird: This is minor and should not hold-up the merge, however, it is my understanding that the pynumero library can or  does provide more than ASL interfaces.

blnicho: Yeah, the local variable name in that function is unnecessarily specific. I can make that quick change. Currently the `pynumero_libraries` package only includes ASL.

mrmundt:
@blnicho - I added `pynumero_libraries` to the workflow

mrmundt:
@blnicho - It failed, but I think I know why. Let me test and I'll make the appropriate changes.

santiagoropb:
It does not seem like the builds failing have something to do with pynumero_libraries. The conda recipe does not have any restrictions on the version of python. Is there any other package that has some sort of conflict with python27?

mrmundt:
> @mrmundt It does not seem like the builds failing have something to do with pynumero_libraries. The conda recipe does not have any restrictions on the version of python. Is there any other package that has some sort of conflict with python27?

If you look at the [files list](https://anaconda.org/conda-forge/pynumero_libraries/files) for pynumero_libraries, the issue is specifically with Windows and any Python older than 3.6. There isn't a distribution for older Python versions.

santiagoropb:
Yeah it seems like by default conda builds python >= 3.6 and python 2.7. Can you skip the tests on 3.5? I thought of adding py3.5 to the conda recipe but it turns out they do no longer support 3.5. Perhaps pyomo shouldn’t support 3.5 either given all the issues it has with dictionaries?

mrmundt:
> @mrmundt Yeah it seems like by default conda builds python >= 3.6 and python 2.7. Can you skip the tests on 3.5? I thought of adding py3.5 to the conda recipe but it turns out they do no longer support 3.5. Perhaps pyomo shouldn’t support 3.5 either given all the issues it has with dictionaries?

@santiagoropb - I've just added a try-catch block for the installation. So if it doesn't work because of Python version, it simply will print a message and move on.

blnicho:
@mrmundt awesome! Lets get that one merged in first and then we can merge those changes into this PR.

blnicho:
This was just waiting on some updates to the GitHub Actions workflows. All of those jobs passed so this should be safe to merge.

PR #1277, opened 2020-01-30
Remove Python 2.7 from Linux/Mac Github Actions

mrmundt:
## Summary/Motivation:
We want to remove Python 2.7 tests from Mac and Linux Github Actions workflows. Because the Github runners are frequently changing, these tests are too unstable, and in the case of Linux, the tests are covered by regular Jenkins jobs as well.

- Remove Python 2.7 Tests for Mac/Linux
- Reorganize the Workflows to all have the same format/printout messages for consistency

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/mac_python_matrix_test.yml
+++ b/.github/workflows/mac_python_matrix_test.yml
@@ -1,4 +1,4 @@
-name: Mac no Py27
+name: Mac no Python 2.7
 
 on:
   push:
--- a/.github/workflows/mac_python_matrix_test.yml
+++ b/.github/workflows/mac_python_matrix_test.yml
@@ -32,24 +32,32 @@ jobs:
         brew install unixodbc
         brew install freetds # Now install Python modules
         python -m pip install --upgrade pip
+        echo ""
         echo "Install Pyomo dependencies..."
+        echo ""
         pip install cython numpy scipy ipython openpyxl sympy pyyaml pyodbc networkx xlrd pandas matplotlib dill seaborn pymysql pyro4 pint pathos
     - name: Install Pyomo and extensions
       run: |
         echo "Clone Pyomo-model-libraries..."
         python -m pip install --upgrade pip
         git clone --quiet https://github.com/Pyomo/pyomo-model-libraries.git
+        echo ""
         echo "Install PyUtilib..."
+        echo ""
         pip install --quiet git+https://github.com/PyUtilib/pyutilib
+        echo ""
         echo "Install Pyomo..."
+        echo ""
         python setup.py develop # Install Pyomo
+        echo ""
         echo "Install extensions..."
+        echo ""
         pyomo download-extensions # Get Pyomo extensions
         pyomo build-extensions
     - name: Run nightly, not fragile tests with test.pyomo
       run: |
+        echo "Run test.pyomo..."
         python -m pip install --upgrade pip
         pip install nose
-        echo "Run test.pyomo..."
         KEY_JOB=1
         test.pyomo -v --cat="nightly" pyomo `pwd`/pyomo-model-libraries # Run nightly, stable tests
--- a/.github/workflows/mac_python_matrix_test.yml
+++ b/.github/workflows/mac_python_matrix_test.yml
@@ -1,9 +1,9 @@
-name: Mac no Python 2.7
+name: continuous-integration/github/pr/osx
 
 on:
-  push:
+  pull_request:
     branches:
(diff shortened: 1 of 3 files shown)


.github/workflows/mac_python_matrix_test.yml
```diff
@@ -12,9 +12,9 @@ jobs:
     runs-on: macos-latest
     strategy:
       fail-fast: false
-      max-parallel: 5
+      max-parallel: 4
```
allevin: What is the implication of leaving max-parallel at 5 or even slightly higher.   
From what I see, currently we have to change the max-parallel every time we add or remove a python version.

mrmundt: I don't think there is any implication in this case. I was simply maintaining the "Can't run more or fewer jobs than Python versions" precedent that we (nonverbally) have created. I'm happy to put it back to 5.

allevin: I can see the argument for both cases... @blnicho do you have an opinion?

jsiirola: Out of curiosity, what happens if you leave `max-parallel` out entirely?  Does it default to 1 or to unlimited?

mrmundt: I just tested it - it looks like it is "unlimited" (in that I think you still can't run more than 20 workflows simultaneously, so if there are some prior ones pending, it'll wait for a job to finish).

allevin: I would then just leave out out... @jsiirola do you agree?

jsiirola: Yes.

.github/workflows/ubuntu_python_matrix_test.yml
```diff
@@ -11,45 +11,51 @@ jobs:
...
+        echo ""
         echo "Download and install extensions..."
+        echo ""
         pyomo download-extensions
         pyomo build-extensions
     - name: Run nightly tests with test.pyomo
       run: |
-        export PYTHONWARNINGS="ignore::UserWarning"
         echo "Run test.pyomo..."
+        python -m pip install --upgrade pip
```
jsiirola: Why upgrade pip again?  I think this could/should be removed.

ghackebeil:
Why not upgrade pip?

.github/workflows/mac_python_matrix_test.yml
blnicho: I would also vote to leave it out based on @mrmundt's comment that the default is "unlimited"

mrmundt:
> Why not upgrade pip?
> […](#)
> On Jan 30, 2020, at 12:08 PM, John Siirola ***@***.***> wrote: ﻿ @jsiirola commented on this pull request. In .github/workflows/mac_python_matrix_test.yml: > pyomo download-extensions # Get Pyomo extensions pyomo build-extensions - name: Run nightly, not fragile tests with test.pyomo run: | + echo "Run test.pyomo..." + python -m pip install --upgrade pip Why are you upgrading pip again? I think this could/should be removed. In .github/workflows/mac_python_matrix_test.yml: > @@ -12,9 +12,9 @@ jobs: runs-on: macos-latest strategy: fail-fast: false - max-parallel: 5 + max-parallel: 4 Yes. In .github/workflows/ubuntu_python_matrix_test.yml: > echo "Run test.pyomo..." + python -m pip install --upgrade pip Why upgrade pip again? I think this could/should be removed. — You are receiving this because you are subscribed to this thread. Reply to this email directly, view it on GitHub, or unsubscribe.

@jsiirola @ghackebeil I think the question is more: "Why upgrade it AGAIN"? It's not strictly necessary, but I've found that some changes don't stick once you move to a new step in the workflow. I haven't seen any issues that make me think it doesn't stick - it was mostly a precaution. I'm happy to call the update only once.

.github/workflows/mac_python_matrix_test.yml
mrmundt: Has been corrected in the newest commit

.github/workflows/ubuntu_python_matrix_test.yml
mrmundt: Removed.

mrmundt:
> I think this is ready once all the Github Actions jobs finish (we don't need to wait for the other testing harnesses)

@blnicho @allevin - Github Action jobs have passed!

Issue #1280, opened 2020-01-30
Update GAMS version for Jenkins/Github


jsiirola:
We currently test against an older GAMS version.  Should we consider moving to the current GAMS version (as of the time of writing, 30.1.0)?

We should probably resolve #1279 first.

blnicho:
Yes, I think moving to the current version is a good idea.

mrmundt:
#1279 introduces GAMS 30.1.0 to MacOS (since older distributions have issues installing on Mac 10.15), so that's one thing already done.

mrmundt:
This has been done.

PR #1281, opened 2020-01-30
Add GAMS to MacOS

### LINKED ISSUE #1279: Add GAMS to OSX Github Actions
Can we add a GAMS installation to the OSX workflow?  The URL for a self-extracting zip file is: https://d37drm4t2jghv5.cloudfront.net/distributions/24.8.5/macosx/osx_x64_64_sfx.exe

mrmundt:
## Summary/Motivation:
GAMS is represented in Windows and Linux testing. We need to add it to MacOS.

- Add GAMS installation to MacOS
- GAMS version 29.1.0 (older distributions do not install on Mac 10.15)
- Fix Windows/Linux path variables for GAMS solver

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/mac_python_matrix_test.yml
+++ b/.github/workflows/mac_python_matrix_test.yml
@@ -42,20 +42,18 @@ jobs:
         chmod +x osx_x64_64_sfx.exe
         ./osx_x64_64_sfx.exe -q -d gams
         cd gams/gams29.1_osx_x64_64_sfx/apifiles/Python/
-        py_ver="$(python --version)"
-        if [[ "$py_ver" == *"2.7"* ]]; then
-          cd api
-          python setup.py -q install
-        elif [[ "$py_ver" == *'3.6'* ]]; then
-          cd api_36
-          python setup.py -q install
-        elif [[ "$py_ver" == *'3.7'* || "$py_ver" == *'3.8'* ]]; then
-          cd api_37
-          python setup.py -q install -noCheck
-        else 
-          cd api_34
-          python setup.py -q install -noCheck
-        fi
+        py_ver=$(python -c 'import sys;print("%s%s" % sys.version_info[:2])')
+        echo $py_ver
+        gams_ver=api
+        for ver in api_*; do
+            echo $ver
+            if test ${ver:4} -le $py_ver; then
+                gams_ver=$ver
+            fi  
+        done
+        echo $gams_ver
+        cd $gams_ver
+        python setup.py -q install -noCheck
     - name: Install Pyomo and extensions
       run: |
         echo "Clone Pyomo-model-libraries..."
--- a/.github/workflows/mac_python_matrix_test.yml
+++ b/.github/workflows/mac_python_matrix_test.yml
@@ -1,9 +1,9 @@
 name: continuous-integration/github/pr/osx
 
 on:
-  pull_request:
+  push:
     branches:
-      - master
+      - mac_gams
       # Can add additional branches if desired
 
 jobs:
@@ -35,6 +35,13 @@ jobs:
         echo "Install Pyomo dependencies..."
         echo ""
         pip install cython numpy scipy ipython openpyxl sympy pyyaml pyodbc networkx xlrd pandas matplotlib dill seaborn pymysql pyro4 pint pathos
+        echo ""
+        echo "Install GAMS..."
+        echo ""
+        wget -q https://d37drm4t2jghv5.cloudfront.net/distributions/30.1.0/macosx/osx_x64_64_sfx.exe
+        chmod +x osx_x64_64_sfx.exe
+        ./osx_x64_64_sfx.exe -q -d gams
+        PATH=$PATH:/gams/gams30.1_osx_x64_64
     - name: Install Pyomo and extensions
       run: |
         echo "Clone Pyomo-model-libraries..."
--- a/.github/workflows/mac_python_matrix_test.yml
+++ b/.github/workflows/mac_python_matrix_test.yml
@@ -41,7 +41,7 @@ jobs:
         wget -q https://d37drm4t2jghv5.cloudfront.net/distributions/29.1.0/macosx/osx_x64_64_sfx.exe
         chmod +x osx_x64_64_sfx.exe
         ./osx_x64_64_sfx.exe -q -d gams
-        python gams/gams29.1_osx_x64_64_sfx/apifiles/Python/api/setup.py install -noCheck
+        python gams/gams29.1_osx_x64_64_sfx/apifiles/Python/api/setup.py install 
     - name: Install Pyomo and extensions
       run: |
         echo "Clone Pyomo-model-libraries..."
--- a/.github/workflows/mac_python_matrix_test.yml
+++ b/.github/workflows/mac_python_matrix_test.yml
@@ -41,6 +41,7 @@ jobs:
         wget -q https://d37drm4t2jghv5.cloudfront.net/distributions/29.1.0/macosx/osx_x64_64_sfx.exe
         chmod +x osx_x64_64_sfx.exe
         ./osx_x64_64_sfx.exe -q -d gams
+        python /gams/gams29.1_osx_x64_64_sfx/apifiles/Python/api/setup.py install -noCheck
     - name: Install Pyomo and extensions
       run: |
         echo "Clone Pyomo-model-libraries..."
--- a/.github/workflows/mac_python_matrix_test.yml
+++ b/.github/workflows/mac_python_matrix_test.yml
@@ -41,7 +41,8 @@ jobs:
         wget -q https://d37drm4t2jghv5.cloudfront.net/distributions/30.1.0/macosx/osx_x64_64_sfx.exe
         chmod +x osx_x64_64_sfx.exe
         ./osx_x64_64_sfx.exe -q -d gams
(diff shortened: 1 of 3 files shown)


.github/workflows/mac_python_matrix_test.yml
```diff
@@ -35,6 +35,13 @@ jobs:
         echo "Install Pyomo dependencies..."
         echo ""
         pip install cython numpy scipy ipython openpyxl sympy pyyaml pyodbc networkx xlrd pandas matplotlib dill seaborn pymysql pyro4 pint pathos
+        echo ""
+        echo "Install GAMS..."
+        echo ""
+        wget -q https://d37drm4t2jghv5.cloudfront.net/distributions/30.1.0/macosx/osx_x64_64_sfx.exe
+        chmod +x osx_x64_64_sfx.exe
+        ./osx_x64_64_sfx.exe -q -d gams
+        PATH=$PATH:/gams/gams30.1_osx_x64_64
```
jsiirola: This is not quite right: we are still getting that the GAMS python bindings and the gams executable are not available; e.g.:
```
Thu, 30 Jan 2020 21:30:15 GMT test_keepfiles_gms (pyomo.solvers.tests.checks.test_GAMS.GAMSTests) ... SKIP: The 'gams' executable is not available
Thu, 30 Jan 2020 21:30:15 GMT test_keepfiles_py (pyomo.solvers.tests.checks.test_GAMS.GAMSTests) ... SKIP: The 'gams' python bindings are not available
```

mrmundt: Lovely. I'll look into it.

mrmundt: @jsiirola - It looks like licensing is part of the issue (as per [step 5](https://www.gams.com/latest/docs/UG_MAC_INSTALL.html#UG_MAC_SFX_INSTALL)). Do we have an existing demo license, or shall I request one from GAMS?

jsiirola: We do not have a demo license.  That is a relatively new development: for the 24.8.5 release, the download defaults to demo mode and does not require a separate license.  It would be great to reach out to GAMS and see what we can work out.  I worry about how we would secure access to the license: while we can use it on Jenkins where we can restrict who has access to the workspace and build environment, on github actions, the world would have access to the license file, which is almost certainly problematic.

mrmundt: We'll have to work something out if we want to use a newer distribution of GAMS. Do you mind if I reach out to GAMS about a community license for use in Pyomo testing?

Also, I wonder if there might be a Github Action that can grab input files (like licenses) from a private repo, provided that the workflow executor has the correct permissions. Let me look into that as a solution for the access control concern.

jsiirola: @mrmundt: that would be great.

bernalde: Guys, starting from version 30.1, GAMS is providing a free community license for linear problems with at most 5000 variables and constraints and 25000 variables and constraints for the remaining problems (this should cover all the tests within Pyomo). Unfortunately, the default demo mode is no longer default after this new version. Here is the info https://www.gams.com/latest/docs/UG_License.html#GAMS_Community_Licenses

mrmundt:
@jsiirola - Fun fact: I checked and the same issue was happening with Ubuntu and Windows. I have fixed the issues.
In addition, I tried a one-older GAMS distribution for Mac (29.1.0). It installs and runs without a license file, so I've changed the workflow to incorporate this while we wait on #1280.

.github/workflows/mac_python_matrix_test.yml
mrmundt: I rolled back one distribution to 29.1.0 for Mac. It has the old default (demo license).

jsiirola:
We are getting closer.  We are still seeing:
```
test_file_removal_py (pyomo.solvers.tests.checks.test_GAMS.GAMSTests) ... SKIP: The 'gams' python bindings are not available
```
It looks like we can install them with:
```
cd gams/gams29.1_linux_x64_64_sfx/apifiles/Python/api && python setup.py install
```
(substituting `api_34`, `pai_36`, and `api_37` for those versions of Python... 3.5 and 3.8 *might* be able to get away with (automatically) rebuilding the bindings drom the `api` directory?)

mrmundt:
> We are getting closer. We are still seeing:
> 
> ```
> test_file_removal_py (pyomo.solvers.tests.checks.test_GAMS.GAMSTests) ... SKIP: The 'gams' python bindings are not available
> ```
> 
> It looks like we can install them with:
> 
> ```
> cd gams/gams29.1_linux_x64_64_sfx/apifiles/Python/api && python setup.py install
> ```
> 
> (substituting `api_34`, `pai_36`, and `api_37` for those versions of Python... 3.5 and 3.8 _might_ be able to get away with (automatically) rebuilding the bindings drom the `api` directory?)

Another one that is happening for all OS'! Interesting point there - you can just give it the flag `-noCheck` to power through without worrying about the Python version. The original `api` folder looks like it's intended for Py27. I tried `python setup.py install -noCheck` on my local machine (which runs Py37) and it built, so it might work out for Py35/38.

I think, in terms of "closeness", we might want to try the `api_34` bindings for Py35 and `api_37` for Py38. Thoughts?

jsiirola:
> I think, in terms of "closeness", we might want to try the `api_34` bindings for Py35 and `api_37` for Py38. Thoughts?

No - the code for the various versions should be the same.  The difference is how the compiled module libraries are linked.  So, api_34 should *only* work for Python 3.4.x.  When we use `api`'s setup.py for other python versions, it should be recognizing that the Python version is different and regenerating / recompiling the libraries from scratch (the point of `api_3?` is that on those platforms, no compilation is necessary).

mrmundt:
> > I think, in terms of "closeness", we might want to try the `api_34` bindings for Py35 and `api_37` for Py38. Thoughts?
> 
> No - the code for the various versions should be the same. The difference is how the compiled module libraries are linked. So, api_34 should _only_ work for Python 3.4.x. When we use `api`'s setup.py for other python versions, it should be recognizing that the Python version is different and regenerating / recompiling the libraries from scratch (the point of `api_3?` is that on those platforms, no compilation is necessary).

Gotcha! Makes sense to me. I'll work on that.

mrmundt:
> We are getting closer. We are still seeing:
> 
> ```
> test_file_removal_py (pyomo.solvers.tests.checks.test_GAMS.GAMSTests) ... SKIP: The 'gams' python bindings are not available
> ```
> 
> It looks like we can install them with:
> 
> ```
> cd gams/gams29.1_linux_x64_64_sfx/apifiles/Python/api && python setup.py install
> ```
> 
> (substituting `api_34`, `pai_36`, and `api_37` for those versions of Python... 3.5 and 3.8 _might_ be able to get away with (automatically) rebuilding the bindings drom the `api` directory?)

@jsiirola - It turns out, this issue exists in ALL of our testing suites. If you look at Appveyor, Jenkins, and Travis, the GAMS python binding tests are skipped in every single one. I am experimenting with getting it working on Ubuntu first (I was able to get them working on my local RHEL).

mrmundt:
@jsiirola @allevin @blnicho - This is finally complete. The Python bindings don't work for Python 3.5 and 3.8 on Windows, but that's it. Everything else is fine. 

Related to this: It turns out there is (at least on Windows) a bug with conda that occurs when a package fails to install. I submitted [an issue to conda](https://github.com/conda/conda/issues/9675) for it.

.github/workflows/mac_python_matrix_test.yml
```diff
@@ -35,6 +35,27 @@ jobs:
...
         pip install cython numpy scipy ipython openpyxl sympy pyyaml pyodbc networkx xlrd pandas matplotlib dill seaborn pymysql pyro4 pint pathos
+        echo ""
+        echo "Install GAMS..."
+        echo ""
+        wget -q https://d37drm4t2jghv5.cloudfront.net/distributions/29.1.0/macosx/osx_x64_64_sfx.exe
+        chmod +x osx_x64_64_sfx.exe
+        ./osx_x64_64_sfx.exe -q -d gams
+        cd gams/gams29.1_osx_x64_64_sfx/apifiles/Python/
+        py_ver="$(python --version)"
+        if [[ "$py_ver" == *"2.7"* ]]; then
```
jsiirola: This is an edge case, but one that may show up... 3.2.7 will match the above.  A better test might be:
```
if `python -c 'import sys;sys.exit(0 if sys.version_info[:2] == (2,7) else 1)'`; then
```
For the others you can do `sys.version_info[:2] == (3,6)` and `sys.version_info[:2] >= (3,7)`.

mrmundt: I could change it to match how I did the Windows ones - those checks are based off the python-version matrix value.

jsiirola: I think either would be OK.  The advantage of the relational operator is we won't have to change it for future releases.  You could do something like this as well:
```
py_ver=`python -c 'import sys;print(''.join(sys.version_info[:2]))'`
gams_ver=api
for ver in api_*; do
    if test ${ver:4} -lt $py_ver; then
        gams_ver=$ver
    fi
done
cd $gams_ver
python setup.py -q install -noCheck
```

mrmundt: Changes made and passed.

PR #1289, opened 2020-02-11
GDP examples refactor: use `build_model()`

qtothec:
## Summary/Motivation:
Wrap GDP examples in `build_model()` function rather than directly executing at import-time.

Lots of lines changed because it adds a tab, but I claim no other functional changes.

Doing this so that I can more easily import and execute these within a horse-race framework.

### Legal Acknowledgement


qtothec:
Oh, you've got quite a bit of code in test_gdp.py. I don't necessarily want to break the `pyomo` command tests. I wonder if adding `if __name__ == '__main__'` to these would help...

qtothec:
I have to admit, I think it would be easier for me to create a "concrete" variant of the models in question than to figure out how to rework the testing code. It bugs me a bit that it's so difficult to go from the command line version of a model to the scripting version, and vice versa.

emma58:
Yes, I would be fine with that. Concrete ones would be much more useful anyway. (I haven't written an abstract model since I coded those examples.) I don't think reworking the testing code is hard though... Because I don't think there's any reason to test with the pyomo command. So we can just import your build_model function, instantiate with the .dat file, and run the transformation. Actually, all we need to do is write the LP file. We don't test anything about the solve anyway (though we should... But that's a later problem.)

qtothec:
I think what's giving me trouble is that it's unclear exactly which asserts or conditions are being tested. It looks like there's some YAML output being generated from somewhere.

emma58:
@jsiirola, will you make sure you can forgive me for the hacking in test_gdp? If so, I think this is good to go...

examples/gdp/jobshop.py
```diff
@@ -29,51 +29,64 @@
...
 
-# minimize makespan
-model.makespan = Objective(expr=model.ms)
+    # minimize makespan
+    model.makespan = Objective(expr=model.ms)
+    return model
+
+
+def build_small_concrete():
+    return build_model().create_instance('jobshop-small.dat')
```
jsiirola: Why make this a hard-coded function?  It seems to only be called in the `if __name__ == '__main__'` below.  I think it can be removed.

examples/gdp/medTermPurchasing_Literal.py
```diff
@@ -1,602 +1,616 @@
...
+       return [ model.FD_1mo_contract[j,t], model.FD_2mo_contract[j,t],
+                model.FD_3mo_contract[j,t], model.FD_no_contract[j,t], ]
+    model.FD_contract = Disjunction(model.RawMaterials, model.TimePeriods,
+       rule=FD_contract)
+
+    return model
+
+
+def build_concrete():
+    return build_model().create_instance('medTermPurchasing_Literal_Chull.dat')
```
jsiirola: Why have this function?  It only seems to be used in the `if __name__ == '__main__'` below.

examples/gdp/jobshop.py
qtothec: I added these in here to associate the abstract model files with the `.dat` files that accompany them. Otherwise you're left with an abstract model and you'd be hard-coding the `dat` association somewhere else.

PR #1291, opened 2020-02-12
Minor FBBT Fixes and Tests

michaelbynum:
- Fixes a bug in the iteration limit for FBBT.
- Ensures `compute_bounds_on_expr` returns None instead of `inf`
- Adds tests for the above changes
- Fixed the BigM _estimate_M method to account for changes to FBBT
- Added quadratic=False to the call to generate_standard_repn in _estimate_M to ensure quadratic constraints are handled properly in the BigM transformation

### Legal Acknowledgement


qtothec:
Wait, `compute_bounds_on_expr` is suddenly going to start returning `(None, None)` instead of `(-inf, inf)`?

qtothec:
Because I just told XY that he can use `compute_bounds_on_expr` for `MatOpt` a couple days ago, and at that point the expected output was `(-inf, inf)`.

michaelbynum:
@qtothec I made this change to be consistent with the return values of `fbbt`. Can you explain how this is problematic for MatOpt?

michaelbynum:
And to be consistent with variable bounds, which use `None` to indicate infinite.

qtothec:
@michaelbynum No strong opinions on whether it should be one way or another. Just wanted more of a warning that an API change was happening.

PR #1296, opened 2020-02-16
Fix component decorator return value

jsiirola:
## Summary/Motivation:
Currently, the "`@model.Constraint()`" notation masks the decorated function:
```python
@model.Constraint()
def foo(m):
    return m.x >= 0
```
results in `foo` being declared in the local namespace, but `type(foo) is None`.  This PR fixes it so that `foo` is the original rule function.

- Fix return value for component decorator

### Legal Acknowledgement


qtothec:
This certainly makes sense, although I'm surprised that anyone noticed.

Issue #1244, opened 2020-01-07
Fixed variables are not always updated correctly in persistent solvers


michaelbynum:
Fixed variables are not updated correctly in the persistent solvers if they are fixed before the constraints/objectives they are used in get added to the solver. If that happens, the variables get replaced by their values when `generate_standard_repn` is called. This should be easy to fix by setting the `compute_values` argument to `False` when `generate_standard_repn` is called.

mrmundt:
@michaelbynum - did you fix this within APPSI?

michaelbynum:
This is fixed.

Issue #1169, opened 2019-11-13
with cplex, solver_io='python' is much slower than executable=(solver_path) 


songbotao:
With cplex, when the model is large, the method of solver_io='python' is much slower than executable=(solver_path) . When the model is small, the method of solver_io='python' is the same as  than executable=(solver_path) . 
I would know how about this problem, and why the difference happen

songbotao:
The version of my pyomo is 5.6.6, the version of my pyomo is 12.9

jsiirola:
Archived on the master Performance Proposals Issue (#1430).  Closing this performance proposal until active development has begun.

mrmundt:
This is a duplicate to #51 ; we are working on refactoring the solver interfaces and will aim to address this as part of that effort.

PR #1300, opened 2020-02-19
Add branching priorities to CPLEXSHELL

### LINKED ISSUE #1299: Issue Branching Priorities
Several commercial solvers allow for issuing branching directions and priorities for variables of MIP problems however Pyomo doesn't support this.

See [Gurobi](https://www.gurobi.com/documentation/9.0/refman/branchpriority.html), [CPLEX](https://www.ibm.com/support/knowledgecenter/SSSA5P_12.10.0/ilog.odms.cplex.help/CPLEX/FileFormats/topics/ORD.html), [GAMS](https://www.gams.com/latest/docs/UG_LanguageFeatures.html#UG_LanguageFeatures_SettingPrioritiesForBranching)

One should be able to apply a "direction" and "priority" for variables which are read and applied by the solver upon solve time.

Note, this is highly solver-specific and will require separate implementations for the Shell vs Persistent solvers. For example, the CPLEXSHELL solver can write a `.ord` file to disk in exactly the same manner as the warm start file which is then read by the CPLEX script. Meanwhile, the CPLEXDirect solver would utilise `self._solver_model.order.set()`

I would be happy to raise a PR given some design direction:
1. Should this be implemented using the `Suffix` interface? If not, would a simple `List[Tuple[PyomoVar, PriorityType, BranchDirectionType]]` passed in as a `solver.solve()` kwarg be sufficient? 
2. Is it acceptable to only implement this for CPLEXSHELL and CPLEXDirect for now?

ruaridhw:
## Summary/Motivation:
To be able to use the branching priority capability of commercial solvers when solving MIP problems

- Add branching priority / direction data to variables
- Add branching priority capability to the CPLEXSHELL interface

### Legal Acknowledgement

### DIFF
--- a/pyomo/opt/base/problem.py
+++ b/pyomo/opt/base/problem.py
@@ -8,7 +8,13 @@
 #  This software is distributed under the 3-clause BSD License.
 #  ___________________________________________________________________________
 
-__all__ = [ 'AbstractProblemWriter', 'WriterFactory', 'ProblemConfigFactory', 'BaseProblemConfig' ]
+__all__ = [
+    "AbstractProblemWriter",
+    "WriterFactory",
+    "ProblemConfigFactory",
+    "BaseProblemConfig",
+    "BranchDirection",
+]
 
 from pyomo.common import Factory
 
@@ -44,3 +50,12 @@ class AbstractProblemWriter(object):
     def __exit__(self, t, v, traceback):
         pass
 
+
+class BranchDirection(object):
+    """ Allowed values for MIP variable branching directions in the `direction` Suffix of a model. """
+
+    default = 0
+    down = -1
+    up = 1
+
+    ALL = {default, down, up}
--- a/pyomo/solvers/plugins/solvers/CPLEX.py
+++ b/pyomo/solvers/plugins/solvers/CPLEX.py
@@ -25,6 +25,7 @@ from pyomo.opt.solver import *
 from pyomo.solvers.mockmip import MockMIP
 from pyomo.core.base import Var, ComponentMap, Suffix, active_export_suffix_generator
 from pyomo.core.kernel.block import IBlock
+from pyomo.util.components import iter_component
 
 logger = logging.getLogger('pyomo.solvers')
 
@@ -257,12 +258,8 @@ class CPLEXSHELL(ILMLicensedSystemCallSolver):
     def _convert_priorities_to_rows(self, instance, priorities, directions):
         if isinstance(instance, IBlock):
             smap = getattr(instance, "._symbol_maps")[self._smap_id]
-            var_is_indexed = lambda var: hasattr(var, 'components')
-            var_iter = lambda var: var.components()
         else:
             smap = instance.solutions.symbol_map[self._smap_id]
-            var_is_indexed = lambda var: var.is_indexed()
-            var_iter = lambda var: var.values()
         byObject = smap.byObject
 
         rows = []
@@ -275,14 +272,7 @@ class CPLEXSHELL(ILMLicensedSystemCallSolver):
 
             var_direction = directions.get(var, BranchDirection.default)
 
-            if not var_is_indexed(var):
-                if id(var) not in byObject:
-                    continue
-
-                rows.append((byObject[id(var)], priority, var_direction))
-                continue
-
-            for child_var in var_iter(var):
+            for child_var in iter_component(var):
                 if id(child_var) not in byObject:
                     continue
 
--- a/pyomo/solvers/plugins/solvers/CPLEX.py
+++ b/pyomo/solvers/plugins/solvers/CPLEX.py
@@ -18,7 +18,6 @@ import pyomo.common
 import pyutilib.common
 import pyutilib.misc
 
-from pyomo.core import ComponentMap, Suffix, active_export_suffix_generator
 from pyomo.opt.base import *
 from pyomo.opt.base.solvers import _extract_version
 from pyomo.opt.results import *
@@ -238,7 +237,7 @@ class CPLEXSHELL(ILMLicensedSystemCallSolver):
 
     def _write_priorities_file(self, instance) -> None:
         """ Write a variable priorities file in the CPLEX ORD format. """
-        from pyomo.core.base import Var
+        from pyomo.core.base import Var, ComponentMap, Suffix, active_export_suffix_generator
 
         if isinstance(instance, IBlock):
             smap = getattr(instance, "._symbol_maps")[self._smap_id]
--- a/pyomo/solvers/plugins/solvers/CPLEX.py
+++ b/pyomo/solvers/plugins/solvers/CPLEX.py
@@ -18,6 +18,7 @@ import pyomo.common
 import pyutilib.common
 import pyutilib.misc
 
+from pyomo.core import ComponentMap, Suffix, active_export_suffix_generator
 from pyomo.opt.base import *
 from pyomo.opt.base.solvers import _extract_version
 from pyomo.opt.results import *
@@ -122,10 +123,9 @@ class CPLEXBranchDirection:
     @staticmethod
     def to_str(branch_direction):
         try:
-            return {
-                CPLEXBranchDirection.down: "DN",
-                CPLEXBranchDirection.up: "UP",
-            }[branch_direction]
+            return {CPLEXBranchDirection.down: "DN", CPLEXBranchDirection.up: "UP"}[
+                branch_direction
+            ]
         except KeyError:
             return ""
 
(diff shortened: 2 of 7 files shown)


ruaridhw:
This should be ready for review as all the checks are passing now. GitHub is giving me a few suggestions (@ghackebeil, @jsiirola, @whart222) though I'm not sure who would be best placed so whenever one of you is free!

pyomo/solvers/tests/checks/test_cplex.py
```diff
@@ -54,5 +59,198 @@ def test_validate_file_name(self):
         with self.assertRaisesRegexp(ValueError, msg):
             _validate_file_name(_128, fname, 'xxx')
 
+
+class CPLEXShellWritePrioritiesFile(unittest.TestCase):
+    def setUp(self):
+        from pyomo.solvers.plugins.converter.model import PyomoMIPConverter  # register the `ProblemConverterFactory`
+        from pyomo.repn.plugins.cpxlp import ProblemWriter_cpxlp  # register the `WriterFactory`
```
jsiirola: Instead if explicitly importing the plugins here, I think it would be preferable to just
```python
import pyomo.environ
```
at the top of the module (which will ensure that standard plugins are registered.

pyomo/solvers/tests/checks/test_cplex.py
```diff
@@ -54,5 +59,198 @@ def test_validate_file_name(self):
...
+        direction_val = BranchDirection.down
+        self.mock_model.direction.set_value(self.mock_model.x, direction_val)
+
+        CPLEXSHELL._write_priorities_file(self.mock_cplex_shell, self.mock_model)
+        priorities_file = self.get_priorities_file_as_string(self.mock_cplex_shell)
+
+        self.assertEqual(
+            priorities_file,
+            "* ENCODING=ISO-8859-1\nNAME             Priority Order\n %s x1 %d\nENDATA\n"
+            % (ORDFileSchema._direction_to_str(direction_val), priority_val),
```
jsiirola: I think I would prefer this string to be hard-coded (not generated with `%s` placeholders).  Doing so will ensure that this also tests that the `ORDFileSchema._direction_to_str` is correct.

pyomo/solvers/tests/checks/test_cplex.py
```diff
@@ -54,5 +59,198 @@ def test_validate_file_name(self):
...
+            self.mock_model.x, "invalid_branching_direction"
+        )
+
+        CPLEXSHELL._write_priorities_file(self.mock_cplex_shell, self.mock_model)
+        priorities_file = self.get_priorities_file_as_string(self.mock_cplex_shell)
+
+        self.assertEqual(
+            priorities_file,
+            "* ENCODING=ISO-8859-1\nNAME             Priority Order\n  x1 %d\nENDATA\n"
+            % (priority_val,),
```
jsiirola: Same as above: I would recommend hard-coding the reference string.

pyomo/opt/base/problem.py
```diff
@@ -44,3 +50,12 @@ def __enter__(self):
     def __exit__(self, t, v, traceback):
         pass
 
+
+class BranchDirection(object):
```
jsiirola: Open question: is this the best place for this enum-like object?

ruaridhw:
Thanks @michaelbynum. I am certainly up for improving the direct/persistent interface as well and would be happy to raise a separate PR to implement this functionality for that.

As it happens, we were previously using `CPLEXPersistent` for this advanced functionality and found it to be a little unreliable (see known issues such as #1244, #1169) hence the (hopefully temporary) move back to `CPLEXSHELL` and this PR.

With regards to supporting advanced functionality in `CPLEXSHELL`, from what I can see, there are only a [handful of file-based interfaces](https://www.ibm.com/support/knowledgecenter/SSSA5P_12.10.0/ilog.odms.cplex.help/CPLEX/homepages/reffileformatscplex.html) that `pyomo` *doesn't already* support. Personally, the only other input format that would be much use would be "ANN" for model annotations. I understand that all other settings and parameters are already exposed by `pyomo` via `solver.options`.

michaelbynum:
@ruaridhw Good points. I am working on improving the direct/persistent solvers separately. This is a good solution for now. If you can address @jsiirola 's comments, we will get this merged.

pyomo/opt/base/problem.py
michaelbynum: I wondered the same thing, but I'm not sure of a better place for it. Gurobi has the same options (up, down, and default).

ruaridhw: Yeah, this was the only place general enough that I could think of as there's nothing else as suitable. We could always introduce a new module for this though?

pyomo/solvers/tests/checks/test_cplex.py
ruaridhw: Sure. I've updated this and the below

ruaridhw:
@michaelbynum, @jsiirola, back you you.

pyomo/solvers/tests/checks/test_cplex.py
jsiirola: Below got updated, but I was thinking something closer to this here:
```python
self.assertEqual(
    priorities_file,
    "* ENCODING=ISO-8859-1\n"
    "NAME             Priority Order\n"
    " DN x1 10\n"
    "ENDATA\n")
```

pyomo/solvers/tests/checks/test_cplex.py
```diff
@@ -54,5 +59,198 @@ def test_validate_file_name(self):
...
+        )
+
+
+class CPLEXShellSolvePrioritiesFile(unittest.TestCase):
```
jsiirola: Why is this a separate class from the previous priorities tester?

ruaridhw: The former class is a unit test specifically on the writing of priorities via `_write_priorities_file()`. The latter is an integration test on the end-to-end application of priorities via the `Suffix` through a `solve()`. I've added docs to that effect

pyomo/solvers/tests/checks/test_cplex.py
```diff
@@ -54,5 +59,198 @@ def test_validate_file_name(self):
...
+        self.mock_cplex_shell._priorities_file_name = pyutilib.services.TempfileManager.create_tempfile(
+            suffix=".cplex.ord"
+        )
+
+    def tearDown(self):
+        pyutilib.services.TempfileManager.clear_tempfiles()
+
+    def get_mock_model(self):
+        model = ConcreteModel()
+        model.x = Var(within=Binary)
```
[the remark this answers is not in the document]
ruaridhw: > an indexed variable m.y = Var([1,2]) to this test

Thanks, added.

Yeah, it sure didn't work for `kernel` 🤦‍♂ 

I'm not too familiar with that API so I've had to add special checks as to whether the variable is indexed (see `var_is_indexed()` `var_iter()`). Any alternatives to make that neater would be great!

pyomo/solvers/plugins/solvers/CPLEX.py
```diff
@@ -268,14 +275,14 @@ def _convert_priorities_to_rows(self, instance, priorities, directions):
 
             var_direction = directions.get(var, BranchDirection.default)
 
-            if not var.is_indexed():
```
jsiirola: I *think* the easiest general solution is:
```python
if hasattr(var, 'values'):
    # catches Var, IndexedVar, and kernel's variable_dict
    var_iter = var.values()
elif hasattr(var, '__iter__'):
    # catches list and kernel's variable_list and variable_tuple
    var_iter = iter(var)
else:
    # kernel's variable
    var_iter = iter((var,))
for v in var_iter:
    if id(v) not in byObject:
        continue
    v_direction = directions.get(v, var_direction)
    rows.append((byObject[id(v)], priority, v_direction))
```

ruaridhw: Thanks @jsiirola! I've added this logic as a general-purpose helper in `pyomo.util.components` however let me know if you'd rather this logic is kept private to this method or this isn't the best place for it. (To be honest, I'm surprised this hasn't come up before!)

michaelbynum:
OSX builds are failing, but it does not have anything to do with this PR. I just created another PR which will hopefully fix the problem. Then, we can get this merged.

pyomo/util/components.py
```diff
@@ -52,3 +53,30 @@ def rename_components(model, component_list, prefix):
...
+        # catches `IndexedComponent`, and kernel's `_dict`
+        return iter(obj.values())
+    except AttributeError:
+        pass
+
+    try:
+        # catches list and kernel's `_list` and `_tuple`
+        return iter(obj)
+    except TypeError:
+        return iter((obj,))
```
michaelbynum: @ruaridhw Thanks for adding this. This will be very useful in other places too.

pyomo/util/tests/test_components.py
```diff
@@ -7,10 +7,13 @@
 #  rights in this software.
 #  This software is distributed under the 3-clause BSD License.
 #  ___________________________________________________________________________
+from itertools import zip_longest
```
jsiirola: The import error can be fixed by:
```python
from six.moves import zip_longest
```

ruaridhw: Thanks, updated.

Issue #1295, opened 2020-02-13
Newer CBC no longer works with Pyomo


mrmundt:
In testing CBC 2.10.3 (from https://ampl.com/dl/open/cbc/cbc-linux64.zip) on RHEL7 with Python 3.7, these errors/failures are now occurring: 
```
======================================================================
ERROR: test_cbc_nl_symbolic_labels (pyomo.solvers.tests.checks.test_writers.transplant_class.<locals>.C)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/mmundt/devel-pyomo/pyomo/pyomo/solvers/tests/checks/test_writers.py", line 72, in writer_test
    load_solutions)
  File "/home/mmundt/devel-pyomo/pyomo/pyomo/solvers/tests/models/base.py", line 126, in solve
    **io_options)
  File "/home/mmundt/devel-pyomo/pyomo/pyomo/opt/base/solvers.py", line 600, in solve
    "Solver (%s) did not exit normally" % self.name)
pyutilib.common._exceptions.ApplicationError: Solver (cbc) did not exit normally
-------------------- >> begin captured logging << --------------------
pyomo.opt: ERROR: Solver (cbc) returned non-zero return code (-11)
pyomo.opt: ERROR: Solver log:
CBC 2.10.3: printingOptions=all
 
--------------------- >> end captured logging << ---------------------
 
======================================================================
FAIL: test_optimal_mip (pyomo.solvers.tests.checks.test_CBCplugin.TestCBC)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/mmundt/devel-pyomo/pyomo/pyomo/solvers/tests/checks/test_CBCplugin.py", line 121, in test_optimal_mip
    self.assertEqual(1.0, results.problem.lower_bound)
AssertionError: 1.0 != -1.0
 
======================================================================
FAIL: test_unbounded_mip (pyomo.solvers.tests.checks.test_CBCplugin.TestCBC)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/mmundt/devel-pyomo/pyomo/pyomo/solvers/tests/checks/test_CBCplugin.py", line 106, in test_unbounded_mip
    self.assertEqual(TerminationCondition.unbounded, results.solver.termination_condition)
[...]

PR #1303, opened 2020-02-19
NFC: Remove CBC from our installation documentation

### LINKED ISSUE #1302: Platform dependence: Special case needed for coincbc in installation.rst
In `/doc/OnlineDocs/installation.rst`, under the "Using CONDA" paragraph, the following line:
```
conda install -c conda-forge ipopt coincbc glpk
```
should be changed to:
```
conda install -c conda-forge ipopt glpk
```
and these lines added:

On Linux-64 and OSx-64 platforms, this other solver can be installed as well:
```
conda install -c conda-forge coincbc
```

blnicho:
## Summary/Motivation:
This removes CBC from the list of solvers that we mention can be installed using conda in our installation documentation. @CatChenal pointed out that the conda package only works on Linux and OSX and @mrmundt pointed out that the newer version of CBC causes test failures (#1295). Rather than update the documentation with a bunch of caveats for a particular solver, I think the better approach is to remove it completely for now. We should probably add a section to the documentation that gives more guidance on installing different solvers but that is outside the scope of this PR.

- Remove CBC from our conda installation documentation

### Legal Acknowledgement

### DIFF
--- a/doc/OnlineDocs/installation.rst
+++ b/doc/OnlineDocs/installation.rst
@@ -23,7 +23,7 @@ optimization solvers can be installed with conda as well:
 
 ::
 
-   conda install -c conda-forge ipopt coincbc glpk
+   conda install -c conda-forge ipopt glpk
 
 
 Using PIP


mrmundt:
Is this something we plan to add back eventually, or is the idea to remove it permanently?

blnicho:
My intention was to remove it permanently. It only appeared in the installation docs to provide a very brief note on a few solvers that were easy to install with Anaconda. 

I think our users would certainly benefit from a more detailed/comprehensive guide on installing various solvers however this type of documentation would be time consuming to develop and difficult to keep up-to-date.

Issue #1030, opened 2019-05-23
[PEP] Redesign of Pyomo Solvers


michaelbynum:
I would like to start redesigning the solver interfaces in Pyomo.

Motivation
-----------

1. The current solver interfaces are more complex than necessary.
2. The current solver interfaces have many inconsistencies.
3. The current solver interfaces do not allow the solution of a model/block with constraints that use variables which live elsewhere.

Preliminary Design Decisions (to be updated with time)
-------------------------------------------------------
- When solving a model/block, the model is defined by the set of active components and any components used by those active components. This means that variables are handed to the solver if and only if they are used in an active constraint or objective.

Implementation
----------------
- I intend to develop new solver interfaces alongside the existing ones. I do not intend to remove existing solver interfaces. Once the new solver interfaces are sufficiently mature, we can rename/archive the old interfaces.
- I intend to start with a couple simple solver interfaces and get feedback before proceeding further.

Thoughts? Concerns?

Related Issues
---------------

- #1169 
- #1490 
- #1776 🇺🇸 
- #2845 
- #3064 
- #3110 
- #3243 
- #3448 
- #3450

ghackebeil:
I think you should reduce the scope of this PEP. The third bullet point is an orthogonal and controversial issue. There is no need to have it possibly get in the way the other (much needed) changes.

jwatsonnm:
Regarding the third bullet Gabe references, can you clarify? Is said variable at least below the current block in the model component hierarchy? If it's not in the hierarchy, then I agree with Gabe that this is highly controversial.

carldlaird:
I updated the motivation in the pep to have numbers.

It would be good to document the cases where "Bullet 3" is required. I also admit that the first time I heard that we needed to support solves where variables live outside the hierarchy, I thought this was very strange because it breaks encapsulation of sub-models. Having said that, I would like to see a use case where this is required to be convinced one way or another.

michaelbynum:
@ghackebeil @jwatsonnm I will open a separate PEP for (3).

michaelbynum:
@ghackebeil @jwatsonnm @carldlaird Let's move the discussion of (3) to #1032.

DLWoodruff:
Now that the controversy has moved to #1032, I predict that everyone will think the new PEP with only items one and two is non-controversial great idea and that you should proceed with your plan to create  a couple of examples.

jsiirola:
[...]

PR #1313, opened 2020-02-24
Parse `noSolution` status from CPLEXShell and CPLEXDirect solvers correctly

ruaridhw:
## Summary/Motivation:
Currently when CPLEX returns "No Solution", this isn't captured in any of the Pyomo solution or solver data objects. We can identify "No Solution" because when both of (1) max time limit is reached and (2) the model is infeasible occur.

### `CPLEXDirect`
- Clean up the status codes in `_postsolve()` to use the actual CPLEX library constants
- Add new logic to handle `MIP_time_limit_infeasible` and `MIP_dettime_limit_infeasible` as indicating a solver `termination_condition` of `noSolution`
- Get the solver return code if possible from any `CplexSolverError`s (but don't add this to the returned `Bunch` due to downstream code raising `ApplicationError`s)
- Set `results.solver.return_code` and `results.solver.termination_message` using the available information

### `CPLEXSHELL`
- Fix `process_logfile()` for CPLEX 12.10 as it looks like IBM have updated the log file schema and is no longer being parsed correctly
- Add new logic to handle `'no integer solution.'` in the log file as indicating a solver `termination_condition` of `noSolution`
- Get the solver return code if possible from any `CPLEX Error` lines in the log file

### Legal Acknowledgement

### DIFF
--- a/pyomo/opt/base/problem.py
+++ b/pyomo/opt/base/problem.py
@@ -8,7 +8,13 @@
 #  This software is distributed under the 3-clause BSD License.
 #  ___________________________________________________________________________
 
-__all__ = [ 'AbstractProblemWriter', 'WriterFactory', 'ProblemConfigFactory', 'BaseProblemConfig' ]
+__all__ = [
+    "AbstractProblemWriter",
+    "WriterFactory",
+    "ProblemConfigFactory",
+    "BaseProblemConfig",
+    "BranchDirection",
+]
 
 from pyomo.common import Factory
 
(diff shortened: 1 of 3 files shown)


michaelbynum:
@ruaridhw It looks like the changes in #1300 are also included in this PR. However, the changes to the direct interface do not depend on the changes in #1300. If you do not want this PR to be held up by #1300, then I would recommend creating a PR with just the changes to the direct interface.

ruaridhw:
@michaelbynum, sorry about that. Branch is fixed now.

ruaridhw:
@michaelbynum, this should be ready to review whenever you get a chance

michaelbynum:
@ruaridhw Thanks. I will review these PRs early next week.

mrmundt:
This PR is going to be superseded by work to complete #1030 - which builds in base support for "no solution" to all solver interfaces. As such, I am going to close this PR.

PR #1318, opened 2020-02-28
Added a test for the csvwriter to rapper tests

DLWoodruff:
## Summary/Motivation:
Adds a test for the csvwriter to the rapper tests because users have been claiming
that the csvwriter is failing when they use rapper.

- adds one test to the rapper tests.
-

### Legal Acknowledgement

### DIFF
--- a/pyomo/pysp/tests/rapper/rapper_tester.py
+++ b/pyomo/pysp/tests/rapper/rapper_tester.py
@@ -1,5 +1,5 @@
 # Provide some test for rapper; most are smoke because PySP is tested elsewhere
-# Author: David L. Woodruff (circa March 2017 and Sept 2018)
+# Author: David L. Woodruff (circa March 2017; Sept 2018; Feb 2020)
 
 import pyutilib.th as unittest
 import tempfile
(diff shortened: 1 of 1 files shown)


DLWoodruff:
All I added was a test, so code coverage going down is an artifact of how coverage is computed and
compared.

pyomo/pysp/tests/rapper/rapper_tester.py
```diff
@@ -110,6 +111,19 @@ def test_ef_solve_with_gap(self):
...
+    def test_ef_solve_with_csvwriter(self):
+        """ solve the ef and report gap"""
+        stsolver = rapper.StochSolver("ReferenceModel.py",
+                                      fsfct = "pysp_instance_creation_callback",
+                                tree_model = self.farmer_concrete_tree)
+        res, gap = stsolver.solve_ef(solvername, tee=True, need_gap=True)
+        csvw.write_csv_soln(stsolver.scenario_tree, "testcref")
+        with open("testcref.csv", 'r') as f:
+            line = f.readline()
+        assert(line.split(",")[0] == "FirstStage")
```
jsiirola: This test needs to delete the file it created:
```python
fname = "testcref.csv"
try:
    csvw.write_csv_soln(stsolver.scenario_tree, fname)
    with open(fname, 'r') as f:
        data = f.readline(s)
finally:
    if os.path.exists(fname):
        os.remove(fname)
```
Other notes:
- This should probably use the tempfile creation logic from either python or pyutilib - we cannot always guarantee that the CWD is writable.
- can we test more than the first entry of the first line?  For example, a sanity check that the remainder of the file is not empty?

DLWoodruff:
@jsiirola 
This test creates a temp directory and cd's to it for all tests, then cd's back at the end. So all files
go in the temp dir.

jsiirola:
@DLWoodruff: Ahh. I didn't catch that from the diff.  That said, we need to add code to the TestSuite `tearDown` to delete the temporary directory.  This (partially) explains why the test server has >32k `/tmp/tmp*` directories on it...

PR #1319, opened 2020-03-02
Integrate new Set rewrite into Pyomo

jsiirola:
## Summary/Motivation:
This integrates the set rewrite (#1111 / #326) into all of Pyomo.

- updating process_setarg to better handle function/generator arguments
- improve simple_set_rule decorator to preserve wrapped rule API
- adding `indices`, `contains_indices` methods to initializer objects
- Adding TUplizeValuseInitializer to standardize tuplization of lists for multidimensional Sets
- Improved deprecation support for old Set methods and attributes
- adding a Set.get() method for fetchng the value stored in the set
- SetOperators not connected to Blocks are always copied when cloning models/blocks
- Improved implementation of DeclareGlobalSet
- RealSet, IntegerSet, BinarySet moed from `core.kernal.set_tyes` to `core.base.set`.  Stubs (sufficient for kernel's needs) were reimplemented in Kernel.
- improved testing and handling of various edge cases encountered when converting Pyomo

### Legal Acknowledgement


examples/pysp/scripting/apps/compile_scenario_tree.py
```diff
@@ -102,7 +102,7 @@ def _pickle_compiled_scenario(worker,
             param._validate = None
         for set_ in block.component_objects(Set):
             set_.initialize = None
-            set_.filter = None
+            #set_.filter = None
```
blnicho: Can this commented line be removed?

pyomo/core/base/block.py
```diff
@@ -821,7 +821,7 @@ def transfer_attributes_from(self, src):
                or k in self._decl:
                 setattr(self, k, src_raw_dict[k])
 
-    def _add_temporary_set(self, val):
+    def _add_implicit_sets(self, val):
         """TODO: This method has known issues (see tickets) and needs to be
         reviewed. [JDS 9/2014]"""
```
blnicho: Is this TODO still relevant? Seems to pre-date the move to GitHub

doc/OnlineDocs/pyomo_modeling_components/Sets.rst
```diff
@@ -106,14 +106,17 @@ Note that the element number starts with 1 and not 0:
 .. doctest::
 
     >>> model.X.pprint()
-    X : Dim=0, Dimen=1, Size=10, Domain=None, Ordered=False, Bounds=(1, 19)
-        [1, 3, 5, 7, 9, 11, 13, 15, 17, 19]
+    X : Size=1, Index=None, Ordered=Insertion
+        Key  : Dimen : Domain : Size : Members
+        None :     1 :    Any :   10 : {1, 3, 5, 7, 9, 11, 13, 15, 17, 19}
```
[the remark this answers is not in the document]
jsiirola: I agree that it is confusing.  I am open to options for renaming the table entry (the header should probably remain the same for consistency with other IndexedComponents)

examples/doc/pyomobook/pyomo-components-ch/var_declaration.txt
```diff
@@ -14,9 +14,9 @@ A : Dim=0, Dimen=1, Size=3, Domain=None, Ordered=False, Bounds=(1, 3)
     [1, 2, 3]
 s : Size=3, Index=A
     Key : Lower : Value : Upper : Fixed : Stale : Domain
-      1 :     1 :  None :     2 : False :  True : IntegerInterval(1, 2)
-      2 :     2 :  None :     3 : False :  True : IntegerInterval(2, 3)
-      3 :     3 :  None :     4 : False :  True : IntegerInterval(3, 4)
+      1 :     1 :  None :     2 : False :  True : [1:2]
```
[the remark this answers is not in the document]
jsiirola: No - the information is still available.  The notation is: 
- `[0]` scalar
- `[1:10]` integer interval (1,2,3,4,5,6,7,8,9,10)
- `[1:9:2]` integer interval, with step of 2 (1,3,5,7,9)
- `[1..10]` closed real interval between 1 and 10
- `(1..10)` open real interval between 1 and 10

examples/pysp/scripting/apps/compile_scenario_tree.py
jsiirola: Yes, but I wanted to leave it in for the review so that @ghackebeil, @jwatsonnm, and @DLWoodruff would see it and could comment.

pyomo/core/base/block.py
jsiirola: Yes.  This is still an issue.  I am separately working on changing how to resolve it.  I didn't want to further complicate this PR with that change.  The problem is documented in, e.g., #45.

pyomo/core/base/global_set.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def __str__(self):
+        # Override str() to always print out the global set name
+        return self.name
+
+# FIXME: This mocks up part of the Set API until we can break up the set
+# module to resolve circular dependencies and can make this a proper
+# GlobalSet (Scalar IndexedComponent objects are indexed by
+# UnindexedComponent_set, but we would like UnindexedComponent_set to be
+# a proper scalar IndexedComponent).
```
michaelbynum: Is there an existing issue to make sure this gets addressed?

pyomo/core/base/set.py
```diff
@@ -69,19 +126,23 @@ def process_setarg(arg):
         raise TypeError("Cannot apply a Set operator to a non-Set "
                         "component data (%s)" % (arg.name,))
 
-    # TODO: DEPRECATE this functionality? It has never been documented,
+    # DEPRECATED: This functionality has never been documented,
     # and I don't know of a use of it in the wild.
```
michaelbynum: Should there be a deprecation warning in here somewhere?

pyomo/core/base/set.py
```diff
@@ -147,12 +226,26 @@ def A_rule(model, i, j):
...
             return Set.End
         return value
-    return wrapper_function
+""" % (formatargspec(fn),)
+    # Create the wrapper in a temporary environment that mimics this
+    # function's environment.
+    _env = dict(globals())
+    _env.update(locals())
+    exec(_funcdef, _env)
+    return _env['wrapper_function']
```
michaelbynum: I understand the need to preserve the original function signature. However, is a string followed by exec the best way to do this? Maybe I do not quite understand the goal here. Can you expand on this further?

pyomo/core/base/set.py
```diff
@@ -205,20 +307,44 @@ def __call__(self, parent, idx):
...
+                    raise ValueError(
+                        "SetIntersectInitializer contains two "
+                        "sub-initializers with inconsistent external indices")
+            return self._A.indices()
+        else:
+            # It is OK (and desirable) for this to raise the exception
+            # if B does not contain external indices
+            return self._B.indices()
+
+class BoundsInitializer(InitializerBase):
```
michaelbynum: Can you add some very brief class documentation to describe what these initializer classes are used for? I think it will be useful for developers in the future.

pyomo/core/base/set.py
```diff
@@ -306,12 +461,36 @@ def isordered(self):
         """Returns True if this is an ordered finite discrete (iterable) Set"""
         return False
 
+    def subsets(self, expand_all_set_operators=None):
```
michaelbynum: I like the addition of the subsets method a lot.

pyomo/core/base/set.py
```diff
@@ -1386,6 +1695,9 @@ def _sort(self):
     'set_value', 'add', 'remove', 'discard', 'clear', 'update', 'pop',
 )
 
+
+@ModelComponentFactory.register(
+    "Set data that is used to define a model instance.")
```
michaelbynum: This is the first I have seen or heard of the ModelComponentFactory. Is this new? Can you explain the purpose of this?

pyomo/core/base/set.py
jsiirola: I spent a long time on this, and have not been able to find a way to preserve the original function signature without using `exec`.  The challenge is that some of the obscure (and now generally unused) rules for functions added an additional positional argument to the rule that was the number of times the rule had been called (this was legacy support from before we understood the power of generators).  To preserve backwards comparability with that way of calling rules, we need to know how many positional arguments the rule was expecting.  If we create the wrapper in the normal way (with `*args`), that information is lost.  So, when we wrap a rule, we generate a new function with *exactly* the same calling signature.  I consider this reasonably safe, as the string is "just" a function definition (no side effects), and the variable data does not come from the user, but instead from a Python internal function.

pyomo/core/base/set.py
jsiirola: This is ancient (pre-Coopr3).  It is only used by a legacy subcommand that will print out the list of known components and the `@model.Constraint()` decorator syntax.

pyomo/core/base/global_set.py
michaelbynum: Okay, that's fine with me.

pyomo/core/base/set.py
michaelbynum: Ahh... Okay.

pyomo/core/base/set.py
michaelbynum: Okay, I figured you had a good reason, but I had to ask.

Robbybp:
Just merged this with my dae-tools branch and discovered a bug. When creating a `SetOperator` from a `ContinuousSet`, the `_fe` attribute will change from containing only the finite element points to containing the entire set. The following code illustrates:

```python
from pyomo.environ import *
from pyomo.dae import *

m = ConcreteModel()

m.time = ContinuousSet(bounds=(0,1))
m.set1 = Set(initialize=['a','b','c'])

disc = TransformationFactory('dae.collocation')
disc.apply_to(m, wrt=m.time, nfe=2, ncp=2, scheme='LAGRANGE-RADAU')

op = m.time * m.set1

print(m.time._fe)
```

displaying `[0, 0.166667, 0.5, 0.666667, 1]`, instead of `[0, 0.5, 1]` as in previous versions of pyomo. This is significant as it renders `get_finite_elements` useless. This could be caused by an erroneous call to `construct`, which sets `self._fe = sorted(self)`, but I didn't see an obvious culprit in a quick scan through the code of the set operators.

Robbybp:
This also seems to happen when expanding arcs.

jsiirola:
@Robbybp ContinuousSet should be fixed now (including a test to catch this case in the future).  Thanks!

Robbybp:
> @Robbybp ContinuousSet should be fixed now (including a test to catch this case in the future). Thanks!

Yep, the code that led to me finding this works now, thanks for fixing

ghackebeil:
Hoping to get around to looking over this PR this week.

pyomo/core/base/set.py
```diff
@@ -2314,9 +2999,42 @@ def isdiscrete(self):
         """Returns True if this set admits only discrete members"""
         return all(r.isdiscrete() for r in self.ranges())
 
+    def subsets(self, expand_all_set_operators=None):
+        if not isinstance(self, SetProduct):
+            if expand_all_set_operators is None:
+                logger.warning("""
+                Extracting subsets for Set %s, which is a SetOperator
+                other than a SetProduct.  Returning this set and not
+                decending into the set operands.  To descend into this
```
blnicho: Typo: descending

pyomo/core/base/set.py
```diff
@@ -2818,12 +3546,30 @@ def dimen(self):
                 ans += s_dim
         return UnknownSetDimen if _unknown else ans
 
+    def _flatten_product(self, val):
+        """Flatten any nested set product terms (due to nested products)
+
+        Note that because this is called in a recursive context, this
+        method is assued that there is no more than a single level of
```
blnicho: Typo: assured?

pyomo/core/base/set.py
```diff
@@ -3067,103 +3890,212 @@ def DeclareGlobalSet(obj):
...
+        global_name = None
+
+        def __new__(cls, **kwds):
+            """Hijack __new__ to mock up old RealSet el al. interface
+
+            In the original Set implementation (Pyomo<=5.6.7), the
+            global sets were instances of their own virtual set classes
+            (RealSet, IntegerSet, BooleanSet), and one could create new
+            instances of those sets with modified bounds.  Since the
+            GlobalSet mechansism also declares new classes for every
```
blnicho: Typo: mechanism

pyomo/core/tests/unit/test_set.py
```diff
@@ -2730,6 +3265,165 @@ def test_name(self):
...
+        with self.assertRaisesRegex(
+                NameError, "name 'TrinarySet' is not defined"):
+            TrinarySet
+
+    def test_exceptions(self):
+        with self.assertRaisesRegex(
+                RuntimeError, "Duplicate Global Set declaration, Reals"):
+            DeclareGlobalSet(RangeSet( name='Reals', ranges=(NR(0,2,1),) ))
+
+        # But repeat delcarations are OK
```
blnicho: Typo: declarations

jsiirola:
@blnicho, I think I addressed all your comments except for updating the `ContinuousSet` docstring.

examples/kernel/variables.py
```diff
@@ -1,14 +1,15 @@
 import pyomo.kernel as pmo
+import pyomo.environ as pe
```
ghackebeil: I'm hoping to keep pyomo.kernel as self contained as possible until the two modeling layers are compatible. Is this import necessary?

examples/pysp/scripting/apps/compile_scenario_tree.py
ghackebeil: Feel free to delete. This code was just about removing initialization stuff that no longer seemed to be needed once a model becomes "static".

pyomo/core/kernel/set_types.py
```diff
@@ -20,232 +20,62 @@
...
+    @staticmethod
+    def is_integer():
+        return True
+
+    @staticmethod
+    def is_binary():
+        return False
+
+
+class BinarySet(object):
```
ghackebeil: The BinarySet, BooleanSet objects don't seem to serve a purpose in kernel.

examples/kernel/variables.py
jsiirola: Unfortunately, for this example., yes: `Reals` is a proper Pyomo Set.  The alternative was to change all the `Reals` in the example to `RealSet` from kernel.

pyomo/core/kernel/set_types.py
jsiirola: I agree, but I think I ran into examples that used BinarySet and BooleanSet from kernel.  Regardless, I left them in for backwards compatibility.  If we can later confirm that they are not necessary (and no one is using them), then we can delete.

ghackebeil: I'm fairly certain this these are not needed in kernel. Binary, yes. BinarySet, no.

jsiirola: BooleanSet is used in `pyomo/core/tests/unit/kernel/test_variable.py` (and historically `BooleanSet` and `BinarySet` have been the same class), but it was not imported into `pyomo.kernel`.

Part of my refactoring was so that `kernel` proper only needs `RealSet` and `IntegerSet`, and can infer Binary from a standard `get_interval()` method.  Longer term, I would like to standardize on `BinarySet` where needed so that we can re-purpose `BooleanSet` to be a proper Boolean (i.e., True/False, and not 0/1) set.

PR #1321, opened 2020-03-03
Repair damage from a bug in tree_structure that manifests itself only…

DLWoodruff:
… in csvwriter

## Summary/Motivation:
csvwriter was not working correctly when NetworkX was used to specify the tree.

- Add comments about the underlying problem in tree_structure.py, but this would take longer
to fix and is probably not worth the effort,
- so I fixed the only place it bites, which is in csvsolutionwriter and added a test.
(TMI: The problem is that when NetworkX is used to specify the tree, the stage cost names (which are private attributes) are set to None on the stage objects because they are not known at the time of stage creation; however, they are OK on the node and scenario objects. The only code that uses the private cost names on the stage objects is in csvsolutionwriter, so I edited that code to pick up names from the tree nodes if they are None on the stages.)

### Legal Acknowledgement


blnicho:
@DLWoodruff looks like the new test is failing on the jobs that don't include optional dependencies. I'd recommend checking if `networkx` is available and skipping the new test you added if it isn't

PR #1323, opened 2020-03-03
Standardizing github actions naming

jsiirola:
## Summary/Motivation:
This standardizes the naming for the various GitHub Actions drivers.

- standardize test names as `continuous-integration/github/{push,pr} / [mpi/]{linux,osx,win}/py{N}.{M}`

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/parallel_tests.yml
+++ b/.github/workflows/parallel_tests.yml
@@ -1,4 +1,4 @@
-name: parallel_tests
+name: continuous-integration/github/pr
 
 on:
   pull_request:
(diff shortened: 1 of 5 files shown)


.github/workflows/parallel_tests.yml
```diff
@@ -1,4 +1,4 @@
-name: parallel_tests
+name: continuous-integration/github/pr
```
mrmundt: (Left my comment in the wrong place!) 
This is a new workflow. What is the added value?

jsiirola:
@mrmundt: I assume you are referring to the parallel workflow?  It came in as part of #1285, where @michaelbynum wanted to be able to test some things within mpi4py.  We tried to (efficiently, correctly) launch MPI from within nose, but in the end it ended up being easier to to launch nose within MPI and just run the subset of tests that required MPI.  A future PR will introduce an "mpi" category and then categorize the tests that are currently called out explicitly.  Another open todo is to see if we can't get the MPI driver closer to the regular driver.

...having just written all of this, it occurs to me that we should change the "parallel" name to just "mpi".

blnicho:
Can I tack on one more change to this PR? Could we update our actions to use: 

`- uses: actions/checkout@v2` 

This should fix the retrying Github Actions issue.

blnicho:
Since this only impacts the Github Actions builds (which have all passed) I'm going to go ahead and merge this.

PR #1326, opened 2020-03-05
Docs for working on forks

michaelbynum:
## Summary/Motivation:
This PR adds some documentation to help developers work on forks.

### Legal Acknowledgement

### DIFF
--- a/doc/OnlineDocs/contribution_guide.rst
+++ b/doc/OnlineDocs/contribution_guide.rst
@@ -77,6 +77,11 @@ workflow, labeled
 a process that defines separate remotes for your fork and the main
 Pyomo repository.
 
+More information on git can be found at
+https://git-scm.com/book/en/v2. Section 2.5 has information on working
+with remotes.
+
+
 .. _forksgithubui:
 
 Working with my fork and the GitHub Online UI
@@ -230,10 +235,6 @@ these changes to the master branch on your fork,
 
    git push my-fork master
    
-More information on git can be found at
-https://git-scm.com/book/en/v2. Section 2.5 has information on working
-with remotes.
-
 
 Review Process
 --------------
--- a/doc/OnlineDocs/contribution_guide.rst
+++ b/doc/OnlineDocs/contribution_guide.rst
@@ -68,15 +68,19 @@ https://github.com/Pyomo/pyomo, click the "Fork" button in the
 upper right corner, and follow the instructions.
 
 This section discusses two recommended workflows for contributing
-pull-requests to Pyomo. The first workflow, labeled 'Working with my fork
-and the GitHub Online UI', does not require the use of 'remotes', and
+pull-requests to Pyomo. The first workflow, labeled
+:ref:`Working with my fork and the GitHub Online UI <forksgithubui>`,
+does not require the use of 'remotes', and
 suggests updating your fork using the GitHub online UI. The second
-workflow, labeled 'Working with remotes and the git command-line', outlines
+workflow, labeled
+:ref:`Working with remotes and the git command-line <forksremotes>`, outlines
 a process that defines separate remotes for your fork and the main
 Pyomo repository.
 
+.. _forksgithubui:
+
 Working with my fork and the GitHub Online UI
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
++++++++++++++++++++++++++++++++++++++++++++++
 
 After creating your fork (per the instructions above), you can
 then clone your fork of the repository with
@@ -130,7 +134,7 @@ the Pyomo master development branch into the feature branch on your
 fork and in your local clone of the repository.
 
 Using GitHub UI to merge Pyomo master into a branch on your fork
-""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""""
+****************************************************************
 
 To update your fork, you will actually be merging a pull-request from
 the main Pyomo repository into your fork.
@@ -161,10 +165,10 @@ the main Pyomo repository into your fork.
     * You can then merge the pull request by clicking the green "Merge
       pull request" button from your fork on GitHub.
 
-
+.. _forksremotes:
 
 Working with remotes and the git command-line
-^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
++++++++++++++++++++++++++++++++++++++++++++++
 
 After you have created your fork, you can clone the fork and setup
 git 'remotes' that allow you to merge changes from (and to) different
--- a/doc/OnlineDocs/contribution_guide.rst
+++ b/doc/OnlineDocs/contribution_guide.rst
@@ -60,36 +60,61 @@ reviewed or merged by the core development team. In addition, any
 active development may be marked 'stale' and closed.
 
 Working on Forks and Branches
-+++++++++++++++++++++++++++++
+-----------------------------
 
 All Pyomo development should be done on forks of the Pyomo
 repository. In order to fork the Pyomo repository, visit
-https://github.com/Pyomo/pyomo, click the "Fork" button in the upper
-right corner, and follow the instructions. You can then clone your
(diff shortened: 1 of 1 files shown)


doc/OnlineDocs/contribution_guide.rst
```diff
@@ -59,6 +59,128 @@ reviewed or merged by the core development team. In addition, any
...
+
+::
+
+   git remote -v
+
+In this case, you would push to your fork with
+
+::
+
+   git push my_fork <branch_name>
```
jsiirola: You technically don't have to list the branch name for remotes other than origin, unless you want the remote branch to have a different name than your local branch name.

mrmundt:
I like this PR. A lot.

jsiirola:
One other comment.  I would recommend making the documentation self-consistent: either keep one of the remotes as `origin` in all the examples, or else always rename `origin` to something else.  The PR as it stands now has both modes, which is more confusing.

doc/OnlineDocs/contribution_guide.rst
michaelbynum: I personally like the clarity. This explicitly states which branch is being pushed. However, I am still open to making this change. Let's see what other say.

michaelbynum:
@jsiirola I was trying to setup the documentation so that someone could use a fork without ever knowing anything about remotes. I feel like remotes confuse many people, and they are not strictly required to develop on a fork. I like using remotes, but not everyone does. This is the reason for the inconsistency in remote naming.

doc/OnlineDocs/contribution_guide.rst
jsiirola: Technically, it doesn't say which branch is being pushed - only what branch you are pushing to.  The danger is that you are sitting on the local branch `foo` and then `git push my_fork bar` pushes your local branch `foo` into a different remote branch (I only mention this because I have made this mistake in the past).

doc/OnlineDocs/contribution_guide.rst
```diff
@@ -59,6 +59,128 @@ reviewed or merged by the core development team. In addition, any
...
+**************************************
+
+First, suppose you have set up your remotes with the following:
+
+::
+   
+   git clone https://github.com/<username>/pyomo.git
+   git remote rename origin my_fork
+   git remote add main_fork https://github.com/pyomo/pyomo.git
+
```
carldlaird: Is the main Pyomo repository really a "fork"? Is there a better name for this? I could see people thinking that the "main_fork" is their fork.

carldlaird: Also, while I agree with this renaming, it does complicate the use of other instructions from the web which would likely use the "origin" and "upstream" names. (Maybe this is better - I don't like the origin and upstream names, because "origin" to me feels like it refers to the main Pyomo repository - when it is really the "upstream")... Other thoughts on naming? I use my_fork and main_pyomo. But I could be convinced otherwise.

jsiirola: FWIW, I have settled on `origin` for `Pyomo/pyomo`, `me` for `jsiirola/pyomo`, and first names for everyone else's forks.

doc/OnlineDocs/contribution_guide.rst
```diff
@@ -59,6 +59,128 @@ reviewed or merged by the core development team. In addition, any
...
+The "--ff-only" only allows a merge if the merge can be done by a
+fast-forward. If you do not require a fast-forward, you can drop this
+option. The most common concrete example of this would be
+
+::
+
+   git checkout master
+   git fetch main_fork
+   git merge main_fork/master --ff-only
```
carldlaird: This doesn't push the changes to your fork - only merges it locally. How does this differ from the recommendations in some online sources that suggest:
> git checkout master
> git pull upstream master
> git push origin master

doc/OnlineDocs/contribution_guide.rst
carldlaird: But then, isn't your "origin" exactly the reverse of the typical instructions online? Where "origin" is your fork (what you call me), and "upstream" is the main repos (what you call "origin")?

michaelbynum: I have no real preference on the naming.

blnicho: I personally use John's approach but I agree with Carl that using the `origin` name in a slightly different way from other online documentation would likely cause confusion. I like the idea of calling it `main_pyomo`.

doc/OnlineDocs/contribution_guide.rst
blnicho: I agree with John on this one. I never specify the branch name when I push and instead I repeatedly run `git status` while I'm developing to verify the branch I'm on before making any commits or pushing anything. Maybe we could just expand the comment on the `git status` line to double check the active branch (in addition to the staged files).

doc/OnlineDocs/contribution_guide.rst
qtothec: We're going headlong into a messy developer preference case, but the preponderance of online documentation has `origin` be the user's fork and `upstream` refer to the... well, upstream.

https://www.atlassian.com/git/tutorials/comparing-workflows/forking-workflow

I don't have an issue with individual developers choosing something different from the mainstream, but since we don't want to have to field all the questions, we probably don't want to publicize a practice deviating from the "I Googled it" answer.

doc/OnlineDocs/contribution_guide.rst
jsiirola: FWIW, I was wrong.  @michaelbynum showed me that when you `git push remote branch`, your local branch `branch` will be pushed up to the remote as the branch `branch` *regardless* of which branch you happen to currently have checked out.  We are pretty sure that what I was doing when I made the mistake was slightly different.

doc/OnlineDocs/contribution_guide.rst
jsiirola: I agree with @qtothec and @carldlaird that for the purposes of this documentation, our recommendations should at least be not inconsistent with "I Googled it" (either rename both remotes like @michaelbynum is doing now, or keep them with the standard `origin` and `upstream`).

doc/OnlineDocs/contribution_guide.rst
```diff
@@ -59,6 +59,128 @@ reviewed or merged by the core development team. In addition, any
...
+Using git commands to update your fork
+**************************************
+
+First, suppose you have set up your remotes with the following:
+
+::
+   
+   git clone https://github.com/<username>/pyomo.git
+   git remote rename origin my_fork
+   git remote add main_fork https://github.com/pyomo/pyomo.git
```
jsiirola: I am not sure if we want to document this, but I find my life to be significantly easier if my local master branch tracks `Pyomo/pyomo` and not `jsiirola/pyomo`.  I think in the context of this setup, I think that after setting up the remotes, I would switch the upstream for master with:
```bash
git branch master --set-upstream-to main_fork/master
```

doc/OnlineDocs/contribution_guide.rst
jsiirola: @michaelbynum's text is equivalent to the first two commands.  I generally prefer using `fetch` and `merge` explicitly instead of `pull`, as when things failed I felt it was more clear what state my local clone was in.  Also, if you move upstream to `Pyomo/pyomo`, you can use an even "simpler" workflow, either:
```bash
git checkout master
git fetch --all
gir merge --ff-only
```
or
```bash
git checkout master
git pull
```

carldlaird:
@jsiirola @michaelbynum @blnicho @qtothec I made several changes, please look this over again.

michaelbynum:
I like the changes. I think we should move the link to the git documentation to the top.

michaelbynum:
Looks good to me.

qtothec:
A little apprehensive about renaming `origin` to `my-fork` (due to external conventions), but I don't feel strongly enough about it to hold this up.

PR #1335, opened 2020-03-10
Added helper function to retrieve solver success (or not)

carldlaird:
## Summary/Motivation:
This PR adds a small helper function to encapsulate the common check for solver status.
This will simplify and improve several examples in the book.

-
-

### Legal Acknowledgement

### DIFF
--- a/pyomo/opt/results/solver.py
+++ b/pyomo/opt/results/solver.py
@@ -8,7 +8,7 @@
 #  This software is distributed under the 3-clause BSD License.
 #  ___________________________________________________________________________
 
-__all__ = ['SolverInformation', 'SolverStatus', 'TerminationCondition', 'solve_optimal']
+__all__ = ['SolverInformation', 'SolverStatus', 'TerminationCondition', 'optimal_termination']
 
 from pyutilib.enum import Enum
 from pyomo.opt.results.container import MapContainer, ScalarType
@@ -68,10 +68,10 @@ TerminationCondition = Enum(
     )
 
 
-def solve_optimal(results):
+def optimal_termination(results):
     """
     This function returns True if the termination condition for the solver
-    is 'optimal', 'lcoallyOptimal', or 'globallyOptimal', and the status is 'ok'
+    is 'optimal', 'locallyOptimal', or 'globallyOptimal', and the status is 'ok'
 
     Parameters
     ----------
--- a/pyomo/opt/results/solver.py
+++ b/pyomo/opt/results/solver.py
@@ -8,7 +8,7 @@
 #  This software is distributed under the 3-clause BSD License.
 #  ___________________________________________________________________________
 
-__all__ = ['SolverInformation', 'SolverStatus', 'TerminationCondition', 'check_optimal_termination']
+__all__ = ['SolverInformation', 'SolverStatus', 'TerminationCondition', 'check_optimal_termination', 'assert_optimal_termination']
 
 from pyutilib.enum import Enum
 from pyomo.opt.results.container import MapContainer, ScalarType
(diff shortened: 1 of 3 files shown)


qtothec:
Didn't we have an issue some time ago with CONOPT reporting `locaillyOptimal` instead of `Optimal`? We'll need to make sure this doesn't trip on those kind of cases.

carldlaird:
OK, I added checks for the other two as well, and updated the documentation.

pyomo/opt/results/solver.py
```diff
@@ -68,6 +68,26 @@
     )
 
 
+def solve_optimal(results):
+    """
+    This function returns True if the termination condition for the solver
+    is 'optimal', 'lcoallyOptimal', or 'globallyOptimal', and the status is 'ok'
```
blnicho: Typo: locallyOptimal

carldlaird:
@blnicho OK, originally, I really liked your name - then I recalled my initial aversion to the name - the "results" object usually does not actually contain the results. I agree, however, that I did not like solve_optimal. I changed the name to optimal_termination. Does everyone agree that this is a reasonable name?

carldlaird:
@blnicho @michaelbynum Sorry folks, I made another change. We realized we would like this to throw an exception if the solution is not optimal. By default, this is now a one-liner for what  I believe is the most common use case.

pyomo/opt/results/solver.py
```diff
@@ -68,6 +68,33 @@
     )
 
 
+def check_optimal_termination(results, suppress_exception=False):
```
qtothec: With `raise` by default, this doesn't strike me as a `check` function. It's more of an `assert` function, in which case it is odd for it to be returning `True`/`False` values. I see what you're trying to accomplish, but wouldn't it make more sense to retain the original behavior (and not have to add a flag) by just using `assert optimal_termination(results)` when you need a one-liner?

michaelbynum: @qtothec makes a good point...

jsiirola:
I like the idea of this function.

However, I would avoid methods like this raising an exception by default.  I am less opposed to providing an option to raise an exception if the user wants to be lazy and not code in their own, but in general, users should check the return values of the functions they call and act accordingly.  We have problems elsewhere in the code where we raise exceptions for effectively unexceptional events, and then we start passing "`exception=False`" flags through large portions of the API.  Maybe it is less of an issue here because the intent is that this is a user utility and that it won't be used in the core of Pyomo?

DLWoodruff:
@jsiirola John, your speculation at the end of your comment was correct: this is intended to be used only by end-users. Actually, we intend for it to be used by almost all new users. We need a short, simple way of letting new users add an optimality check.  For
the purposes of writing the book, we don't want that checking code to detract from the main messages. But I think also most new users need to be given a simple way to have their code exit if their assumption that they got an optimal solution is invalid. Since we are moving people away from the pyomo command to their own scripts this is probably overdue.

qtothec:
@jsiirola and @DLWoodruff, I made the mistake of making this remark as a code comment, and it seems you two might have missed it among the other emails: https://github.com/Pyomo/pyomo/pull/1335#discussion_r391325796

> Wouldn't it make more sense to retain the original behavior (and not have to add a flag) by just using `assert optimal_termination(results)` when you need a one-liner?

carldlaird:
@qtothec Sorry, Qi. I saw your comment. It took me a bit to get the changes done. I like this better - there are two methods - one that checks (and returns a bool), and one that "asserts" and raises an exception. Let me know what you think.

@DLWoodruff I think this one-liner is still fine - it is clear what it does - we have both methods if we ever need the other one. Agree?

DLWoodruff:
@carldlaird I agree that this is an improvement and that having both function is a good idea.

michaelbynum:
I like the changes. This looks good.

PR #1341, opened 2020-03-13
Units support for Pyomo components and models

carldlaird:
## Summary/Motivation:
This PR provides support for Units handing within Pyomo. It builds off of the units support in a previous PR and adds the ability to assign units to Var, Param, or ExternalFunction components. It also contains methods to convert values between units, and convert expressions to different units. Furthermore, it includes methods to check the units on the models.

-
-

### Legal Acknowledgement


pyomo/core/base/units_container.py
```diff
@@ -1013,13 +1049,27 @@ def exitNode(self, node, data):
         # first check if the node is a leaf
         if type(node) in nonpyomo_leaf_types \
                 or not node.is_expression_type():
-            if isinstance(node, _PyomoUnit):
+            if type(node) in native_numeric_types:
+                # this is a number - return dimensionless                                                                      
+                return (None, None)
```
carldlaird: This if-block is not necessary - it will get caught by the fall-through below - but maybe this is better for performance...

pyomo/core/base/param.py
```diff
@@ -218,6 +223,9 @@ def __init__(self, *args, **kwd):
         self._mutable       = kwd.pop('mutable', Param.DefaultMutable )
         self._default_val   = kwd.pop('default', _NotValid )
         self._dense_initialize = kwd.pop('initialize_as_dense', False)
+        self._units         = kwd.pop('units', None)                                                                           
+        if self._units is not None:                                                                                            
+            self._mutable = True
```
jsiirola: Should this generate a warning?  That is, should users be told that if they are using units, they *must* use mutable Params?

pyomo/core/base/units_container.py
jsiirola: Yes, because most trees have a lot of native numeric leaf nodes, and `isinstance` is quite slow when it returns False.

pyomo/core/base/units_container.py
```diff
@@ -437,7 +449,7 @@ def _pint_units_equivalent(self, lhs, rhs):
         : bool
            True if they are equivalent, and False otherwise
         """
-        if lhs == rhs:
+        if id(lhs) == id(rhs):
```
jsiirola: I think the preferred syntax here is 
```python
if lhs is rhs:
```

pyomo/core/base/units_container.py
```diff
@@ -1285,33 +1343,222 @@ def check_units_equivalent(self, expr1, expr2):
...
+                                                                   pint_expected_unit):
+            raise UnitsError("drop_units called, however the units on the expression: {}"
+                             " do not match the expected units: {}."
+                             "".format(str(pyomo_expr_unit), str(pyomo_expected_unit)))
+
+        return expr/pyomo_expr_unit
+
+    def check_constraint_data_units(self, condata):
+        if condata.equality:
+            if condata.lower == 0.0:
```
jsiirola: We should add a comment here documenting the edge case on why we allow an apparent units inconsistency for equality expressions with a 0 RHS.

pyomo/core/base/units_container.py
```diff
@@ -1285,33 +1343,222 @@ def check_units_equivalent(self, expr1, expr2):
...
+           Pyomo expression
+
+        """
+        pyomo_expr_unit, pint_expr_unit = self._get_units_tuple(expr)
+        pyomo_expected_unit, pint_expected_unit = self._get_units_tuple(expected_units)
+        if not _UnitExtractionVisitor(self)._pint_units_equivalent(pint_expr_unit,
+                                                                   pint_expected_unit):
+            raise UnitsError("drop_units called, however the units on the expression: {}"
+                             " do not match the expected units: {}."
+                             "".format(str(pyomo_expr_unit), str(pyomo_expected_unit)))
```
jsiirola: You perform this units check all the time (even if the user did not request units checking.  This is potentially a performance hit that is unconditionally incurred by any model that uses drop_units.

pyomo/core/base/units_container.py
```diff
@@ -1285,33 +1343,222 @@ def check_units_equivalent(self, expr1, expr2):
...
+           to retrieve the units from the src as a Pyomo expression
+        to_units : Pyomo units expression
+           The desired target units for the new value
+
+        Returns
+        -------
+           float : The converted value
+
+        """
+        src_pyomo_unit, src_pint_unit = self._get_units_tuple(src)
```
jsiirola: You get the src_units regardless of if you actually need them.  Should this only be done if either units checking is enabled or the user didn't provide from_units?

pyomo/core/base/units_container.py
```diff
@@ -1285,33 +1343,222 @@ def check_units_equivalent(self, expr1, expr2):
...
+        if from_units is None:
+            from_pyomo_unit, from_pint_unit = src_pyomo_unit, src_pint_unit
+        else:
+            from_pyomo_unit, from_pint_unit = self._get_units_tuple(from_units)
+            if src_pint_unit is not None and \
+               not _UnitExtractionVisitor(self)._pint_units_equivalent(src_pint_unit, from_pint_unit):
```
jsiirola: You perform this check unconditionally, regardless of if the user requested units checking.  That is an "unconditional performance hit".

pyomo/core/base/param.py
carldlaird: Not sure. Is the main concern that we would be making everything immutable without them knowing?
Note: We are going to have to do some performance testing with this as well to see if the model build times change significantly.

pyomo/core/base/units_container.py
carldlaird: You are right. Easy fix.

pyomo/core/base/units_container.py
carldlaird: Agreed. Done.

pyomo/core/base/units_container.py
carldlaird: @jsiirola After our discussion, I think I am going to remove drop units. Users can do this simply by dividing by the correct units themselves if they need to.

pyomo/core/base/units_container.py
carldlaird: I am going to rework these methods to simplify the api - it will remove this issue as well.

pyomo/core/base/units_container.py
carldlaird: While I can remove some of the checks, we need the factors, so I cannot remove the _get_units_tuple call. Also remember that the check is done later on the final expression (not when the expression is created), so anything that would need to be checked must be a node (or associated with a node) on the expression tree. I will be reworking this method a little as well.

Note: We will have to do some performance testing - performance has not really been my focus during the development, and I am sure there are improvements we can make.

pyomo/core/base/param.py
jsiirola: Right.  Mutable Params are measurably slower than immutable: they have to be constructed, and thy occupy a position in the expression tree (i.e., we cannot simplify them out).  I wonder if it is "evil" that we silently switch a user's Param if they happen to provide units.  A warning would force them to make it explicit, and could help educate them as to why "I added units to my model and suddenly everything is slower!"

I think what I was thinking was something along the lines of:
```python
if self._units is not None and not self._mutable:
    self._mutable = True
    logger.warning("Converting Param %s from immutable to mutable: Params with units must be mutable." % (self.name,))
```

PR #1370, opened 2020-03-31
Add Fourier-Motzkin elimination transformation

emma58:
## Summary/Motivation:

Adds an implementation of Fourier-Motzkin elimination as a transformation in contrib.

- Adds transformation which takes a linear model and a list of continuous variables to project out. It deactivates all the constraints involving those variables, executes FME, and puts the projected constraints on their own block on the model.
- Adds tests for the transformation.

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/fme/fourier_motzkin_elimination.py
+++ b/pyomo/contrib/fme/fourier_motzkin_elimination.py
@@ -14,7 +14,7 @@ from pyomo.core.base import (TransformationFactory, _VarData)
 from pyomo.core.base.block import _BlockData
 from pyomo.core.base.param import _ParamData
 from pyomo.core.base.constraint import _ConstraintData
-from pyomo.core.plugins.transform.hierarchy import LinearTransformation
+from pyomo.core.plugins.transform.hierarchy import Transformation
 from pyomo.common.config import ConfigBlock, ConfigValue
 from pyomo.common.modeling import unique_component_name
 from pyomo.repn.standard_repn import generate_standard_repn
@@ -32,17 +32,7 @@ def vars_to_eliminate_list(x):
     elif hasattr(x, '__iter__'):
         ans = ComponentSet()
         for i in x:
-            if isinstance(i, (Var, _VarData)):
-                # flatten indexed things
-                if i.is_indexed():
-                    for j in i.index_set():
-                        ans.add(i[j])
-                else:
-                    ans.add(i)
-            else:
-                raise ValueError(
-                    "Expected Var or list of Vars."
-                    "\n\tRecieved %s" % type(x))
+            ans.update(vars_to_eliminate_list(i))
         return ans
     else:
         raise ValueError(
@@ -52,7 +42,7 @@ def vars_to_eliminate_list(x):
 @TransformationFactory.register('contrib.fourier_motzkin_elimination',
                                 doc="Project out specified (continuous) "
                                 "variables from a linear model.")
-class Fourier_Motzkin_Elimination_Transformation(LinearTransformation):
+class Fourier_Motzkin_Elimination_Transformation(Transformation):
     """Project out specified variables from a linear model.
 
     This transformation requires the following keyword argument:
@@ -82,6 +72,8 @@ class Fourier_Motzkin_Elimination_Transformation(LinearTransformation):
     def __init__(self):
         """Initialize transformation object"""
         super(Fourier_Motzkin_Elimination_Transformation, self).__init__()
+        self.ctypes_not_to_transform = set((Block, Param, Objective, Set,
+                                        Expression, Suffix))
     
     def _apply_to(self, instance, **kwds):
         config = self.CONFIG(kwds.pop('options', {}))
@@ -108,26 +100,23 @@ class Fourier_Motzkin_Elimination_Transformation(LinearTransformation):
                 descend_into=Block,
                 sort=SortComponents.deterministic,
                 active=True):
-            if obj.type() in (Block, _BlockData, Param, _ParamData, Objective,
-                              Set, Expression, Suffix):
+            if obj.type() in self.ctypes_not_to_transform:
                 continue
-            elif obj.type() in (Constraint, _ConstraintData):
+            elif obj.type() is Constraint:
                 cons_list = self._process_constraint(obj)
                 constraints.extend(cons_list)
                 obj.deactivate() # the truth will be on our transformation block
-            elif obj.type() in (Var, _VarData):
+            elif obj.type() is Var:
                 # variable bounds are constraints, but we only need them if this
                 # is a variable we are projecting out
                 if obj not in vars_to_eliminate:
                     continue
                 if obj.lb is not None:
                     constraints.append({'body': generate_standard_repn(obj),
-                                        'upper': None,
                                         'lower': value(obj.lb),
                                         'map': ComponentMap([(obj, 1)])})
                 if obj.ub is not None:
                     constraints.append({'body': generate_standard_repn(-obj),
-                                        'upper': None,
                                         'lower': -value(obj.ub),
                                         'map': ComponentMap([(obj, -1)])})
             else:
@@ -154,7 +143,7 @@ class Fourier_Motzkin_Elimination_Transformation(LinearTransformation):
                 else:
                     # This would actually make a lot of sense in this case...
                     #projected_constraints.add(Constraint.Infeasible)
-                    raise RuntimeError("Fourier-Motzkin found that model is "
+                    raise RuntimeError("Fourier-Motzkin found the model is "
                                        "infeasible!")
             else:
                 projected_constraints.add(lhs >= lower)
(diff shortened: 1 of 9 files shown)


pyomo/contrib/fme/plugins/fourier_motzkin_elimination.py
```diff
@@ -0,0 +1,329 @@
...
+            waiting_list = []
+
+            while(constraints):
+                cons = constraints.pop()
+                leaving_var_coef = cons['map'].get(the_var)
+                if leaving_var_coef is None or leaving_var_coef == 0:
+                    waiting_list.append(cons)
+                    continue
+
+                # we know the constraints is a >= constraint, using that
```
jsiirola, line 253: typo: constraint

pyomo/contrib/fme/plugins/fourier_motzkin_elimination.py
```diff
@@ -0,0 +1,329 @@
...
+                    vars_that_appear.append(var)
+
+        # we actually begin the recursion here
+        while vars_that_appear:
+            # first var we will project out
+            the_var = vars_that_appear.pop()
+
+            # we are 'reorganizing' the constraints, we will map the coefficient
+            # of the_var from that constraint and the rest of the expression and
+            # sorting based on whether we have the_var <= other stuff or vice
```
jsiirola, line 240: typo: 'sort'?

pyomo/contrib/fme/plugins/fourier_motzkin_elimination.py
```diff
@@ -0,0 +1,329 @@
...
+            # versa.
+            leq_list = []
+            geq_list = []
+            waiting_list = []
+
+            while(constraints):
+                cons = constraints.pop()
```
jsiirola, line 247: Why `pop`?  Couldn't you just iterate over the constraints and at the end of the loop replace the constraints with 
```python
constraints = waiting_list
for leq in leq_list:
    for geq in geq_list:
        constraints.append(self._add_linear_constraints(leq, geq))
```

pyomo/contrib/fme/plugins/fourier_motzkin_elimination.py
```diff
@@ -0,0 +1,329 @@
...
+        """Adds two >= constraints"""
+        ans = {'lower': None, 'upper': None, 'body': None, 'map': ComponentMap()}
+
+        # This is not beautiful, but it needs to be both deterministic and
+        # account for the fact that Vars aren't hashable.
+        seen = ComponentSet()
+        all_vars = []
+        for v in cons1['body'].linear_vars:
+            all_vars.append(v)
+            seen.add(v)
```
jsiirola, line 305: I think you can simplify this:
```python
all_vars = list(cons1['body'].linear_vars)
seen = ComponentSet(all_vars)
```

pyomo/contrib/fme/plugins/fourier_motzkin_elimination.py
```diff
@@ -0,0 +1,329 @@
...
+            cons1_coef = cons1['map'].get(var)
+            cons2_coef = cons2['map'].get(var)
+            if cons2_coef is not None and cons1_coef is not None:
+                ans['map'][var] = new_coef = cons1_coef + cons2_coef
+            elif cons1_coef is not None:
+                ans['map'][var] = new_coef = cons1_coef
+            elif cons2_coef is not None:
+                ans['map'][var] = new_coef = cons2_coef
+            expr += new_coef*var
+        ans['body'] = generate_standard_repn(expr)
```
jsiirola, line 321: Can this be simplified to
```python
ans['body'] = generate_standard_repn(sum(
    (cons1['map'].get(var,0) + cons2['map'].get(var,0))*var for var in all_vars))
```

pyomo/contrib/fme/plugins/fourier_motzkin_elimination.py
```diff
@@ -0,0 +1,329 @@
...
+
+        # upper is None, so we just deal with the constants here.
+        cons1_lower = cons1['lower']
+        cons2_lower = cons2['lower']
+        if cons1_lower is not None and cons2_lower is not None:
+            ans['lower'] = cons1_lower + cons2_lower
```
jsiirola, line 327: When can `cons?_lower` be None?  Given that you are dropping the bound if either one is, It appears that it never can be...  is that correct?

pyomo/contrib/fme/plugins/fourier_motzkin_elimination.py
```diff
@@ -0,0 +1,329 @@
...
+        return ans
+    else:
+        raise ValueError(
+            "Expected Var or list of Vars."
+            "\n\tRecieved %s" % type(x))
+
+@TransformationFactory.register('contrib.fourier_motzkin_elimination',
+                                doc="Project out specified (continuous) "
+                                "variables from a linear model.")
+class Fourier_Motzkin_Elimination_Transformation(LinearTransformation):
```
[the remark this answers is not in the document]
emma58: I don't think it actually provides anything... And also you are right that as long as the variables we are eliminating don't play inside of nonlinear expressions, we are actually okay. So sure, let's just have it build off of Transformation.

pyomo/contrib/fme/plugins/fourier_motzkin_elimination.py
```diff
@@ -0,0 +1,329 @@
...
+        (The constant will be moved to the RHS).
+        """
+        body = constraint.body
+        std_repn = generate_standard_repn(body)
+        # linear only!!
+        if not std_repn.is_linear():
+            raise RuntimeError("Found nonlinear constraint %s. The "
+                               "Fourier-Motzkin Elimination transformation "
+                               "can only be applied to linear models!"
+                               % constraint.name)
```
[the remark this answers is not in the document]
emma58: Ah, oops. You are absolutely right. I think we even don't mind if some of the constraints have nonlinear parts, as long as the variables we are eliminating don't appear in those expressions.

pyomo/contrib/fme/plugins/fourier_motzkin_elimination.py
```diff
@@ -0,0 +1,329 @@
...
+        ub and a lb, it is transformed into two constraints. Otherwise it is
+        flipped if it is <=. Each dictionary contains the keys 'lower', 'upper'
+        and 'body' where, after the process, 'upper' will be None, 'lower' will
+        be a constant, and 'body' will be the standard repn of the body.
+        (The constant will be moved to the RHS).
+        """
+        body = constraint.body
+        std_repn = generate_standard_repn(body)
+        # linear only!!
+        if not std_repn.is_linear():
```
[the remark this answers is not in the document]
emma58: It looks like it does the right thing. Assuming you mean something like:
```
m = ConcreteModel()
m.p = Param(initialize=5, mutable=True)
m.p2 = Param(initialize=5)
m.x = Var()
m.x.fix(3)
# All these are True:
generate_standard_repn(m.p).is_linear() 
generate_standard_repn(m.p2).is_linear() 
generate_standard_repn(5).is_linear() 
generate_standard_repn(m.x).is_linear() 
```
But these constraints can't possibly be needed since they have no variables, so they'll get filtered out after they make it through that check.

pyomo/contrib/fme/plugins/fourier_motzkin_elimination.py
emma58: Ha! Oops, yes, that makes much more sense. :)

pyomo/contrib/fme/plugins/fourier_motzkin_elimination.py
emma58: Good point, it can't.

pyomo/contrib/fme/plugins/fourier_motzkin_elimination.py
jsiirola: No - `+=` shouldn't be a problem.  I just missed that you were updating the `map` as well in my rush to simplify things.

pyomo/contrib/fme/fourier_motzkin_elimination.py
```diff
@@ -82,6 +72,8 @@ class Fourier_Motzkin_Elimination_Transformation(LinearTransformation):
     def __init__(self):
         """Initialize transformation object"""
         super(Fourier_Motzkin_Elimination_Transformation, self).__init__()
+        self.ctypes_not_to_transform = set((Block, Param, Objective, Set,
+                                        Expression, Suffix))
```
jsiirola, line 76: This could be moved to the class scope, since this is effectively static.

pyomo/contrib/fme/tests/test_fourier_motzkin_elimination.py
```diff
@@ -61,7 +61,7 @@ def test_no_vars_specified(self):
     def check_projected_constraints(self, m):
         constraints = m._pyomo_contrib_fme_transformation.projected_constraints
         # x - 0.01y <= 1
-        cons = constraints[4]
+        cons = constraints[1]
```
jsiirola, line 64: Why did the resulting constraints change order?

pyomo/contrib/fme/fourier_motzkin_elimination.py
```diff
@@ -0,0 +1,312 @@
...
+
+        return constraints
+
+    def _nonneg_scalar_multiply_linear_constraint(self, cons, scalar):
+        """Multiplies all coefficients and the RHS of a >= constraint by scalar.
+        There is no logic for flipping the equality, so this is just the 
+        special case with a nonnegative scalar, which is all we need.
+        """
+        cons['body'].linear_coefs = [scalar*coef for coef in
+                                     cons['body'].linear_coefs]
```
jsiirola, line 278: Don't you need to update the quadratic and nonlinear terms as well (if you are going to allow nonlinear constraints)?

pyomo/contrib/fme/fourier_motzkin_elimination.py
```diff
@@ -0,0 +1,312 @@
...
+        for var, coef in cons['map'].items():
+            cons['map'][var] = coef*scalar
+
+        # assume scalar >= 0 and constraint only has lower bound
+        if cons['lower'] is not None:
+            cons['lower'] *= scalar
+        
+        return cons
+
+    def _add_linear_constraints(self, cons1, cons2):
```
jsiirola, line 290: If you are going to allow constraints with (unrelated) nonlinear components, then these need to be addressed in here as well, correct?

pyomo/contrib/fme/tests/test_fourier_motzkin_elimination.py
emma58: They changed because I changed the order the remaining constraints appear in after one iteration of FME, when I stopped using `pop` on the list of constraints. Specifically, `waiting_list` comes first now, and then the new constraints. It was the opposite before. It should still be deterministic, so I think it's okay.

jsiirola: Great.  That's fine as long as we know why the order changed (and we didn't let some nondeterminism sneak in).

emma58:
@jsiirola, oops, you were very much right. I think that it should actually work with unrelated nonlinear expressions now.

PR #1374, opened 2020-04-07
Return meaningful errors when an Expression is treated as a Var

### LINKED ISSUE #1091: More helpful error message when trying to fix an Expression
The IDAES tools make extensive use of `Expressions` in places, which is leading some inexperienced (and the occasional experienced) user to try to `fix` these (which obviously results in an exception). Whilst the resulting error message (_GeneralExpressionData object has no method fix) is meaningful to an experienced user, it is not clear to inexperienced users what this means.

In order to provide more guidance to inexperienced users, I am suggesting adding a `fix` method to `Expressions` which raises an exception with a more useful error message. The error should inform the user that the component they are trying to fix is in fact an `Expression` and cannot be fixed, and some guidance on how to proceed.

Similar messages would be useful for other common actions on `Vars`, e.g. `unfix`, `setlb` and `setub`.

andrewlee94:
## Summary/Motivation:
Within IDAES we have had cases where a user of a model has mistaken an `Expression` for a `Var` and tried to use `fix` to set its value. The returns an error saying the `_GeneralExpressionData` has not implemented a `fix` method, which is not meaningful to many users.

- Implement methods for `setlb`, `setub`, `fix` and `unfix` in `_GeneralExpressionData` and `IndexedEpression` which raise `TypeErrors` telling the user that this action cannot be applied to an `Expression`.

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/expression.py
+++ b/pyomo/core/base/expression.py
@@ -222,6 +222,29 @@ class _GeneralExpressionDataImpl(_ExpressionData):
         """A boolean indicating whether this expression is fixed."""
         return self._expr.is_fixed()
 
+    # Define methods for common activities on Vars in case user mistakes
+    # an Expression for a Var
+    def setlb(self, val):
+        raise TypeError(
+            "Expressions cannot have bounds: %s"
+            % (self.name))
+
+    def setub(self, val):
+        raise TypeError(
+            "Expressions cannot have bounds: %s"
+            % (self.name))
+
+    def fix(self, val):
+        raise TypeError(
+            "Expressions cannot be fixed: %s"
+            % (self.name))
+
+    def unfix(self):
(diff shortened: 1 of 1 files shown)


qtothec:
Might as well tell them to add a constraint instead, then?

andrewlee94:
@qtothec Can do that. Should we suggest adding inequalities for `setlb` and `setub` as well?

qtothec:
I am a bit indifferent here. Inequalities are also constraints, which is why I just suggested the "constraint" wording.

blnicho:
Rather than modify the `Expression` API with "var-like" methods we decided to add a new derived component to IDAES to resolve #1091. See https://github.com/IDAES/idaes-dev/pull/830.

PR #1376, opened 2020-04-08
Rename Component.type() to Component.ctype

### LINKED ISSUE #613: Should the type() method actually be called ctype?
Continuing discussion brought up in #593.

For pyomo components, the way to get the ctype is a method called `type()`, but for kernel components it's a property called `ctype`. Should we move to make these consistent with each other? This could mean both having a property called `ctype`, a method called `ctype()`, or either of the same for `type`. I think I agree with @jsiirola that calling it `ctype` would be a lot clearer since I for one was confused at first when `.type()` wasn't the same as its actual type. As for whether it should be a property or a method, the only information I have on that would be that I noticed kernel components much more often use properties instead of methods. I don't know if this is an actual design choice or if it's faster or something, so I'll let @ghackebeil weigh in on that.

jsiirola:
## Summary/Motivation:
This renames the `Component.type()` method to be a `Component.ctype` property.  This change matches the development team's vernacular when describing that field, as well as its use in methods (like `component_data_objects()`).  The switch from method to property makes it consistent with the kernel implementation.

- switch `Component.type()` (method) to `Component.ctype` (property)
- rename the internal `_type` attribute to `_ctype`
- remove the `type()` hack & test from kernel
- bugfix: rename Complementarity's use of `_type` to `_complementarity_type` for recording the standard form format
- (unrelated: switch a Component deprecation log message to use `pyomo.common.deprecation.deprecated`)

### Legal Acknowledgement

### DIFF
--- a/pyomo/repn/plugins/gams_writer.py
+++ b/pyomo/repn/plugins/gams_writer.py
@@ -160,7 +160,7 @@ class ToGamsVisitor(EXPR.ExpressionValueVisitor):
         if isinstance(comp, ICategorizedObject):
             return comp.ctype
         else:
-            return comp.ctype()
+            return comp.ctype
 
 
 def expression_to_string(expr, treechecker, labeler=None, smap=None):
--- a/pyomo/repn/plugins/gams_writer.py
+++ b/pyomo/repn/plugins/gams_writer.py
@@ -134,15 +134,15 @@ class ToGamsVisitor(EXPR.ExpressionValueVisitor):
             return False, None
 
         if node.is_component_type():
-            if self.ctype(node) not in valid_expr_ctypes_minlp:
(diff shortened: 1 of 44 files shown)


pyomo/repn/plugins/gams_writer.py
```diff
@@ -160,7 +160,7 @@ def ctype(self, comp):
         if isinstance(comp, ICategorizedObject):
             return comp.ctype
         else:
-            return comp.type()
+            return comp.ctype
```
gseastream, line 163: If the name of the `ctype` property is the same as kernel components now, code like this (found at least in the GAMS and Baron writers) can probably be removed. It looks like this line is part of a helper function meant specifically for this workaround.

Issue #1383, opened 2020-04-11
Error with TransformationFactory('core.relax_integers')


ZedongPeng:
When I use `TransformationFactory('core.relax_integrality').apply_to(m)`, a warning came up.

`WARNING: DEPRECATED: core.relax_integrality is deprecated.  Use core.relax_integers  (deprecated in TBD)`

When I changed the code to `TransformationFactory('core.relax_integers').apply_to(m)`, an error came up:

```
TransformationFactory('core.relax_integers').apply_to(m)
AttributeError: 'NoneType' object has no attribute 'apply_to'
```

qtothec:
That's a typo. Try `core.relax_integer_vars`.

qtothec:
@jsiirola Would it be difficult for `TransformationFactory` to return an object that raises an error listing the valid transformations if you happen to give it an invalid argument, instead of returning `None`, as it currently does?

jsiirola:
@qtothec: I don't think that something that provides a more informative error would be too difficult to implement; we already do something similar for the `SolverFactory`.  I will submit two PR's: the first will fix the typo in the deprecation warning (and I expect it will be reviewed / merged quickly).  The second will draft the change to the TransformationFactory, which will probably have a longer discussion/review process.

PR #1386, opened 2020-04-13
[WIP] Switch TransformationFactory to return an "UnknownTransformation" object

jsiirola:
## Summary/Motivation:
In #1383 @qtothec asked if we could return a more helpful return value from `TransformationFactory()` when the requested transformation was not found.  Currently, the Factory returns `None` for unknown transformations, and it is the responsibility of the user to check that before attempting to apply the transformation.  As users typically do not bother with the check, they end up getting less informative errors like:
```
AttributeError: 'NoneType' object has no attribute 'apply_to'
```

This changes the behavior of the TransformationFactory singleton to follow the pattern established by the `SolverFactory` and return an `UnknownTransformation` object instead of `None` when the factory doesn't have a registered matching transformation.  The `UnknownTransformation` instance implements the Transformation API, with public methods that raise errors like:
```
Attempting to use an unavailable transformation.

The TransformationFactory was unable to create "core._bad_transformation"
and returned an UnknownTransformation object.  This error is raised at
the point where the UnknownTransformation object was used as if it were
valid (by calling method "apply_to").

The original transformation was created with the following arguments:
        {}
```

Open design questions
- Should the Transformation API implement an `available()` method so that users can query if the transformation object returned by the factory is available / valid to use (right know, users need to know to import the `UnknownTransformation` clsas and compare types)?
- Should we extend *both* the Transformation and Solver APIs to incluse a `valid()` method to help distinguish between valid, but unavailable solvers/transformations (e.g., the solver is not installed), and invalid solvers/transformations (i.e., nothing is registered under that name)?
- Should the Factories (solver and transformation) raise exceptions for unregistered names?  Users who want to avoid the exception should either use try-except blocks or else check the name first, e.g.:
   ```python
   if name in TransformationFactory:
      TransformationFactory(name).apply_to(model)
   ```

- return `UnknownTransformation` instance from `TransformationFactory` when the requested name is not registered.

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/PyomoModel.py
+++ b/pyomo/core/base/PyomoModel.py
@@ -47,7 +47,9 @@ from pyomo.core.base.numvalue import *
 from pyomo.core.base.block import SimpleBlock
 from pyomo.core.base.set import Set, UnknownSetDimen
 from pyomo.core.base.component import Component, ComponentUID
-from pyomo.core.base.plugin import ModelComponentFactory, TransformationFactory
+from pyomo.core.base.plugin import (
+    ModelComponentFactory, TransformationFactory, UnknownTransformation,
+)
 from pyomo.core.base.label import CNameLabeler, CuidLabeler
 
 import pyomo.opt
@@ -933,7 +935,7 @@ TransformationFactory('%s').apply_to(model) to directly apply the
 transformation to the model instance.""" % (name,name,) )
 
         xfrm = TransformationFactory(name)
-        if xfrm is None:
+        if type(xfrm) is UnknownTransformation:
             raise ValueError("Unknown model transformation '%s'" % name)
         return xfrm.apply_to(self, **kwds)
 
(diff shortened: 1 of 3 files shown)


qtothec:
It occurred to me that another strategy is just to raise an error instead of returning None, but since you made this already...

jsiirola:
Discussion around this PEP is leaning toward updating this proposal to have the Solver and Transformation Factories raise exceptions when the specified object is not found.  The "implicit" solver behavior (assuming that unregistered names in the SolverFactory are ASL solvers) will only apply when the user specifically creates it as `"asl:solver_name"`.  We will register "known" ASL solvers (e.g., from COIN-OR), and provide deprecation warnings for implicitly created solvers to ease the transition.

ghackebeil:
I think I like this idea

jsiirola:
This has been replaced by PEP #1487.

PR #1388, opened 2020-04-14
Fix Constraint Tightener `TransformationFactory`

ruaridhw:
## Summary/Motivation:
- The Constraint Tightener is untested for negative coefficients.
- In light of this, I believe I've found an issue with the bounds being applied (see 
f85a5bf)

- The logic is being reversed twice; `has_lb()` and `has_ub()` have swapped as well as the setting of `LB` and `UB`. This is corrected to the best of my knowledge of this `TransformationFactory`
- Add `rounding_ndigits` functionality for rounding off the bounds being applied
- Some other general cleanup / refactoring

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/preprocessing/plugins/constraint_tightener.py
+++ b/pyomo/contrib/preprocessing/plugins/constraint_tightener.py
@@ -1,5 +1,6 @@
 import logging
-import textwrap
+
+from six.moves import zip
 
 from pyomo.core import Constraint, value, TransformationFactory
 from pyomo.core.plugins.transform.hierarchy import IsomorphicTransformation
@@ -31,27 +32,28 @@ class TightenContraintFromVars(IsomorphicTransformation):
             LB = UB = 0
             if repn.constant:
                 LB = UB = repn.constant
+
             # loop through each coefficent and variable pair
-            for i, coef in enumerate(repn.linear_coefs):
+            for var, coef in zip(repn.linear_vars, repn.linear_coefs):
                 # TODO: Rounding issues
                 # Calculate bounds using interval arithmetic
                 if coef >= 0:
-                    if repn.linear_vars[i].has_ub():
-                        UB = UB + coef * value(repn.linear_vars[i].ub)
+                    if var.has_ub():
+                        UB = UB + coef * value(var.ub)
                     else:
                         UB = float('Inf')
-                    if repn.linear_vars[i].has_lb():
-                        LB = LB + coef * value(repn.linear_vars[i].lb)
+                    if var.has_lb():
+                        LB = LB + coef * value(var.lb)
                     else:
                         LB = float('-Inf')
                 else:
                     # coef is negative, so signs switch
-                    if repn.linear_vars[i].has_lb():
-                        UB = UB + coef * value(repn.linear_vars[i].lb)
+                    if var.has_lb():
+                        UB = UB + coef * value(var.lb)
                     else:
                         UB = float('Inf')
-                    if repn.linear_vars[i].has_ub():
-                        LB = LB + coef * value(repn.linear_vars[i].ub)
+                    if var.has_ub():
+                        LB = LB + coef * value(var.ub)
                     else:
                         LB = float('-Inf')
 
--- a/pyomo/contrib/preprocessing/plugins/constraint_tightener.py
+++ b/pyomo/contrib/preprocessing/plugins/constraint_tightener.py
@@ -20,9 +20,16 @@ class TightenContraintFromVars(IsomorphicTransformation):
     For now, this only operates on linear constraints.
 
     """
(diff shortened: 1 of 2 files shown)


qtothec:
Thanks for the fix and the test. Small request: could you use `NoArgumentGiven` from `pyomo.common.modeling`? https://github.com/Pyomo/pyomo/blob/fdcddf473ced8977add932bcf3fba78d08f9f3d1/pyomo/common/modeling.py#L38

pyomo/contrib/preprocessing/plugins/constraint_tightener.py
```diff
@@ -19,9 +20,16 @@ class TightenContraintFromVars(IsomorphicTransformation):
     For now, this only operates on linear constraints.
 
     """
+    class _MissingArg(object):
+        pass
+
+    def _apply_to(self, model, rounding_ndigits=_MissingArg, **kwds):
```
jsiirola, line 26: Why are you adding `*kwds` to the signature?  It is not used, nor is it verified to be empty.

pyomo/contrib/preprocessing/plugins/constraint_tightener.py
```diff
@@ -31,33 +39,38 @@ def _apply_to(self, instance):
...
+                    else:
+                        LB = float('-Inf')
 
             # if inferred bound is tighter, replace bound
             new_ub = min(value(constr.upper), UB) if constr.has_ub() else UB
             new_lb = max(value(constr.lower), LB) if constr.has_lb() else LB
+
+            if rounding_ndigits is not self._MissingArg:
+                new_ub = round(new_ub, rounding_ndigits)
+                new_lb = round(new_lb, rounding_ndigits)
```
jsiirola, line 72: Why round?  I would think that the safer alternative would be ceil(UB) and floor(LB).

qtothec:
@jsiirola Short answer is yes, the functionality is subsumed by `fbbt`, unless you have a need to apply only a specific kind of tightening operation. Might be worth a provisional deprecation, with a message to complain loudly here if `fbbt` doesn't work for their use case?

pyomo/contrib/preprocessing/plugins/constraint_tightener.py
ruaridhw: It is not used, not. No reason other than conforming to the same pattern as many other `TransformationFactory`s (eg. `BigM_Transformation`). Happy to revert if you'd prefer that?

pyomo/contrib/preprocessing/plugins/constraint_tightener.py
ruaridhw: If `new_ub` is say `10.0001` or `new_lb` is `9.9999` then I wouldn't expect them to become 11 and 9.

michaelbynum:
I don't think fbbt replaces the functionality in this transformation. FBBT does not changes the bounds on constraints. It uses the bounds on constraints to improve the bounds on variables. However, this transformation could make use of the `compute_bounds_on_expr` function from the `fbbt` module.

jsiirola:
@michaelbynum: doh.  Of course you are right.  Is there a particular reason why FBBT doesn't update the constraint bounds when it can?

pyomo/contrib/preprocessing/plugins/constraint_tightener.py
jsiirola: The problem is that when you move the bound in (in your example, from a computed upper bound of 10.0001 to 10), you are potentially cutting off part of the feasible region (therefore it is an invalid cut).

Rounding is a nonstandard way of handling what is effectively a tolerance issue (depending on the computed bound, it either makes the bound tighter or looser).  If you look in, for example, fbbt, you will see the use of a `feas_tol`.  I believe that the general model is to always use the computed value, but to only update the bound if the computed bound is tighter than the original bound by at least `feas_tol`. @michaelbynum?

ghackebeil:
You probably want the option either deactivate the constraint when the body can no longer reach the original bounds or tighten the bounds. So maybe it would be good to report feasible body bounds separately and let the user decide what to do.

michaelbynum:
I agree with Gabe. In most cases, I would think deactivating the constraint makes more sense than tightening the bounds of the constraint (you have proven that the constraint will always be satisfied). The `fbbt` module has an option for that functionality though.

jsiirola:
Note: merging #1458 marked this PR (#1388) as merged and closed.  That is not quite accurate, as the rounding feature from this PR was removed from #1458 as it could potentially generate invalid cuts.  @ruaridhw, if you want to rework that portion of this PR, feel free to open a new PR.

ruaridhw:
Thanks @qtothec and @jsiirola! That other branch is perfectly fine for now.

PR #1391, opened 2020-04-15
mindtpy: implementation of lp/nlp algorithm, bug fix and documentation update

### LINKED ISSUE #1139: Error whil using mindtpy
I was trying to start the MINLP example presented here:
https://pyomo.readthedocs.io/en/stable/contributed_packages/mindtpy.html

When i run the code:

```
from pyomo.environ import *

# Create a simple model
model = ConcreteModel()

model.x = Var(bounds=(1.0, 10.0), initialize=5.0)
model.y = Var(within=Binary)

model.c1 = Constraint(expr=(model.x-3.0)**2 <= 50.0*(1-model.y))
model.c2 = Constraint(expr=model.x*log(model.x)+5.0 <= 50.0*(model.y))

model.objective = Objective(expr=abs(model.x*model.x), sense=minimize)

# Solve the model using MindtPy
SolverFactory('mindtpy').solve(model, mip_solver='glpk', nlp_solver='ipopt')
model.objective.display()
model.display()
model.pprint()
```

I get the following error:

pyomo.common.errors.DeveloperError: Internal Pyomo implementation error:
        "sympy expression type 're' not found in the operator map"
        Please report this to the Pyomo Developers.

I have installed pyomo 5.6.6, pyomo.extras 3.3, and pyomo.solvers 1.0 using Anaconda on Windows 10.

### LINKED ISSUE #1187: Mindtpy error with stale variables when no linear constraints exist
This issue is being opened based on a post by @julian-belina: https://github.com/Pyomo/pyomo/issues/1139#issuecomment-555036213

The example file is:
```
from pyomo.environ import *

#Create a simple model
model = ConcreteModel()

model.x = Var(bounds=(1.0,10.0),initialize=5.0)
model.y = Var(within=Binary)

model.c1 = Constraint(expr=(model.x-3.0)**2 <= 50.0*(1-model.y))
model.c2 = Constraint(expr=model.x*log(model.x)+5.0 <= 50.0*(model.y))

model.objective = Objective(expr=model.x, sense=minimize)

#Solve the model using MindtPy
SolverFactory('mindtpy').solve(model, mip_solver='glpk', nlp_solver='ipopt',tee=True) 

model.objective.display()
model.display()
model.pprint()
```

The issue lies in that the first master problem does not involve the variable `y` at all, so its value is not updated by the optimizer. Note that both constraints are nonlinear, and OA cuts appear not to have been added for some reason.

ZedongPeng:
## Summary/Motivation:
Fixing bugs in MindtPy and adding single-tree implementation of OA using CPLEX callback functionality

- The Single-tree implementation.
- Add a description of the single tree implementation in **mindtpy.rst**.
- Add the options and some configuration confirmation related to the single tree implementation in **MindtPy.py**.
- Fix the bug in OA cut expression in **cut_generation.py**.
- Add the support for the persistent solver.
- Add bound for variables in **MINLP2_simple.py** and **MINLP3_simple.py**.

### Legal Acknowledgement


pyomo/contrib/mindtpy/cut_generation.py
```diff
@@ -50,48 +50,190 @@ def add_oa_cuts(target_model, dual_values, solve_data, config,
...
+
+                    target_model.MindtPy_utils.MindtPy_linear_cuts.oa_cuts.add(
+                        expr=(sum(value(jacs[constr][var])*(var - var.value)
+                                  for var in constr_vars)
+                              + value(constr.body)
+                              >= constr.lower)
+                    )
+
+
+'''
```
bernalde, line 141: This commented part from GDPopt should be removed.

pyomo/contrib/mindtpy/initialization.py
```diff
@@ -20,8 +24,9 @@ def MindtPy_initialize_master(solve_data, config):
     """
     m = solve_data.mip = solve_data.working_model.clone()
     MindtPy = m.MindtPy_utils
+    m.dual.deactivate()
 
-    m.dual.activate()
+    # m.dual.activate()
```
bernalde, line 29: What is the motivation for deactivating the duals here if you deactivate them in max_binary again and you sort of need them in init_rNLP?

pyomo/contrib/mindtpy/tests/online_doc_example.py
```diff
@@ -0,0 +1,61 @@
...
+# model.objective.pprint()
+# model.pprint()
+# model = EightProcessFlowsheet()
+# print('\n Solving problem with Outer Approximation')
+# SolverFactory('mindtpy').solve(model, strategy='OA',
+#                                init_strategy='rNLP',
+#                                mip_solver='cplex',
+#                                nlp_solver='ipopt',
+#                                bound_tolerance=1E-5)
+# print(value(model.cost.expr))
```
jsiirola, line 61: Can these comments be removed?

pyomo/contrib/mindtpy/tests/eight_process_problem.py
```diff
@@ -144,6 +144,6 @@ def __init__(self, *args, **kwargs):
 
         """Bound definitions"""
         # x (flow) upper bounds
-        x_ubs = {3: 2, 5: 2, 9: 2, 10: 1, 14: 1, 17: 2, 19: 2, 21: 2, 25: 3}
+        x_ubs = {3: 2, 5: 2, 9: 2, 10: 1, 14: 1, 17: 2, 18: 1.4, 19: 2, 21: 2, 25: 3}
```
jsiirola, line 147: Out of curiosity, was this bound in the original problem, or are you adding it to support testing with the lp/nlp solver?

pyomo/contrib/mindtpy/mip_solve.py
```diff
@@ -2,17 +2,280 @@
...
+from math import fabs
+
+from pyomo.repn import generate_standard_repn
+
+try:
+    import cplex
+    from cplex.callbacks import LazyConstraintCallback
+except ImportError:
+    print("Cplex python API is not found. Therefore, lp-nlp is not supported")
+    '''Other solvers (e.g. Gurobi) are not supported yet'''
```
jsiirola, line 31: Please do not print your message here.  The way this is currently implemented, this message is printed out *every* time `pyomo.environ` is imported on platforms that do not have cplex.  Also, messages should be emitted using the logger and not with `print()`.  

Look at `pyomo.common.dependencies.attempt_import()` for an alternative approach for working with optional dependencies.  You should be able to just:
```python
cplex, cplex_available = attempt_import('cplex')
```

More generally, you need to re-think either how you define `LazyOACallback`, or else when/where this module is imported.  Because you automatically import this module, your current implementation causes a fatal exception that prevents importing `pyomo.environ` if the cplex python library is not present (see the test failures).

pyomo/contrib/mindtpy/nlp_solve.py
```diff
@@ -51,7 +52,8 @@ def solve_NLP_subproblem(solve_data, config):
                + (0 if c.lower is None else c.lower))
         sign_adjust = 1 if value(c.upper) is None else -1
         fix_nlp.tmp_duals[c] = sign_adjust * max(0,
-                sign_adjust * (rhs - value(c.body)))
+                                                 sign_adjust*(rhs - value(c.body)))
```
jsiirola, line 55: If you are going to reformat (PEP8) this line, I would recommend:
```python
fix_nlp.tmp_duals[c] = sign_adjust * max(
    0, sign_adjust * (rhs - value(c.body)))
```

pyomo/contrib/mindtpy/mip_solve.py
```diff
@@ -2,17 +2,280 @@
...
+
+
+class LazyOACallback(LazyConstraintCallback):
```
jsiirola, line 34: As this class is hard-coded to the CPLEX API, I would recommend including 'cplex' in the class name.  Maybe `LazyOACallback_cplex`?  That hints at what you would need to change/reimplement to add, e.g., Gurobi support.

pyomo/contrib/mindtpy/initialization.py
ZedongPeng: Here, m is the master problem which is a MIP. Therefore, we deactivate the dual for MIP. Otherwise, some warnings will be reported.

pyomo/contrib/mindtpy/tests/online_doc_example.py
ZedongPeng: Thanks for the advice. I have removed these comments in the latest commits.

pyomo/contrib/mindtpy/tests/eight_process_problem.py
ZedongPeng: I tested the case again without `18: 1.4`, it made no difference to the results.
For the other two bound change in MINLP2_simple.py and MINLP3_simple.py, MindtPy is able to deal with these problems without the changes. However, some warnings will be reported if the bounds are not added.

pyomo/contrib/mindtpy/mip_solve.py
ZedongPeng: I created a new file name **single_tree.py** and moved the `LazyOACallback` class to it.
I also used `attempt_import` to import both `cplex` and `pyomo.contrib.mindtpy.single_tree`.
Please let me know if `attempt_import` is inappropriately used or related code can be further improved.

pyomo/contrib/mindtpy/mip_solve.py
ZedongPeng: Thanks for the advice. `LazyOACallback` has been changed into `LazyOACallback_cplex` in the latest commit.

pyomo/contrib/mindtpy/nlp_solve.py
ZedongPeng: Thanks. I have changed it in the latest commit.

pyomo/contrib/mindtpy/mip_solve.py
ZedongPeng: Hi John. I pushed several commits after the pull request. Now there is an **Install dependencies** bug on several tests. Do you know how to fix it? Thank you. @jsiirola

ZedongPeng: Hi John. Do you know why the test `continuous-integration/jenkins/pr` failed? It seems a Jenkins failure.
Thank you.
@jsiirola

doc/OnlineDocs/contributed_packages/mindtpy.rst
```diff
@@ -58,6 +58,46 @@ The solution may then be displayed by using the commands
...
+  >>> model.y = Var(within=Binary)
+
+  >>> model.c1 = Constraint(expr=(model.x-3.0)**2 <= 50.0*(1-model.y))
+  >>> model.c2 = Constraint(expr=model.x*log(model.x)+5.0 <= 50.0*(model.y))
+  
+  >>> model.objective = Objective(expr=model.x, sense=minimize)
+
+  Solve the model using single tree implementation in MindtPy
+  >>> SolverFactory('mindtpy').solve(model, strategy='OA', 
+                                     mip_solver='cplex_persistent', nlp_solver='ipopt', single_tree=True)
```
qtothec, line 95: You might need to have the following:

```
>>> SolverFactory('mindtpy').solve(model, strategy='OA', 
...                                mip_solver='cplex_persistent', nlp_solver='ipopt', single_tree=True)
```

ZedongPeng: OK

pyomo/contrib/mindtpy/tests/test_mindtpy.py
```diff
@@ -161,11 +164,129 @@ def test_OA_Proposal_with_int_cuts(self):
...
+            model = OnlineDocExample()
+            print('\n Solving problem with Outer Approximation')
+            opt.solve(model, strategy='OA',
+                      mip_solver=required_solvers[1],
+                      nlp_solver=required_solvers[0]
+                      )
+            self.assertAlmostEqual(value(model.objective.expr), 3, places=2)
+
+    # the following tests are used to improve code coverage
+    def test_OA_OnlineDocExample2(self):
```
qtothec, line 185: Would prefer if these titles indicated what was being tested. E.g. `iteration limit`

pyomo/contrib/mindtpy/single_tree.py
```diff
@@ -0,0 +1,257 @@
...
+                # TODO What about PositiveIntegers etc?
+                elif 'is not in domain Integers' in err_msg and (
+                        fabs(var_val - rounded_val) <= config.integer_tolerance):
+                    v_to.set_value(rounded_val)
+                # Value is zero, but shows up as slightly less than zero.
+                elif 'is not in domain NonNegativeReals' in err_msg and (
+                        fabs(var_val) <= config.zero_tolerance):
+                    v_to.set_value(0)
+                else:
+                    raise
```
jsiirola, line 74: I think that you could make this more robust by doing something like:
```python
try:
    v_val = self.get_values(opt._pyomo_var_to_solver_var_map[v_from])
    v_to.set_value(v_val)
    if skip_stale:
        v_to.stale = False
except ValueError:
    # Snap the value to the bounds
    if v_to.lb is not None and v_val < v_to.lb and v_to.lb - v_val <= config.zero_tolerance:
        v_to.set_value(v_to.lb)
    elif v_to.ub is not None and v_val > v_to.ub and v_val - v_to.ub <= config.zero_tolerance:
        v_to.set_value(v_to.ub)
    #... or the nearest integer
    elif v_to.is_integer():
        rounded_val = int(round(v_val))
        if (ignore_integrality or fabs(v_val - rounded_val) <= config.integer_tolerance) \
                and rounded_val in v_to.domain:
            v_to.set_value(rounded_val)
    else:
        raise
```

pyomo/contrib/mindtpy/tests/test_mindtpy.py
jsiirola: I agree with @qtothec: it s best if the test names / documentation give an indication of what is being tested.

pyomo/contrib/mindtpy/tests/test_mindtpy_lp_nlp.py
```diff
@@ -0,0 +1,155 @@
...
+from pyomo.contrib.mindtpy.tests.from_proposal import ProposalModel
+from pyomo.contrib.mindtpy.tests.online_doc_example import OnlineDocExample
+from pyomo.environ import SolverFactory, value
+
+required_solvers = ('ipopt', 'cplex_persistent')
+required_solvers_temp = ('ipopt', 'cplex')
+if all(SolverFactory(s).available() for s in required_solvers_temp):
+    subsolvers_available = True
+else:
+    subsolvers_available = False
```
jsiirola, line 19: I believe that `cplex_persistent` is required for these tests.  This is why the tests are failing on Jenkins (on that platform, for some Python versions, we have the lp/nl cplex interface, but not the python bindings, so `cplex` is available, but `cplex_persistent` is not).

This needs to be changed to:
```python
required_solvers = ('ipopt', 'cplex_persistent')
if all(SolverFactory(s).available(False) for s in required_solvers):
    subsolvers_available = True
else:
    subsolvers_available = False
```

pyomo/contrib/mindtpy/single_tree.py
```diff
@@ -0,0 +1,257 @@
...
+import logging
+from pyomo.common.dependencies import attempt_import
+
+cplex, cplex_available = attempt_import('cplex')
+if cplex_available:
+    from cplex.callbacks import LazyConstraintCallback
+else:
+    logging.warning(
+        "Cplex python API is not found. Therefore, lp-nlp is not supported")
+    # Other solvers (e.g. Gurobi) are not supported yet
```
jsiirola, line 24: Since `single_tree` is only imported conditionally if it is going to be used, I think you can just `import cplex` here.  In fact there is no reason not to: if `cplex` is not available, line 27 will fail with "`NameError: name 'LazyConstraintCallback' is not defined`".  I think it is better to get the `ImportError` from '`import cplex`...

pyomo/contrib/mindtpy/MindtPy.py
```diff
@@ -137,7 +131,8 @@ class MindtPySolver(object):
     ))
     CONFIG.declare("mip_solver", ConfigValue(
         default="gurobi",
-        domain=In(["gurobi", "cplex", "cbc", "glpk", "gams"]),
+        domain=In(["gurobi", "cplex", "cbc", "glpk", "gams",
+                   "gurobi_persistent", "cplex_persistent"]),
```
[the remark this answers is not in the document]
qtothec: I should give this some thought as well. GDPopt has the same issue.

pyomo/contrib/mindtpy/nlp_solve.py
```diff
@@ -50,8 +51,9 @@ def solve_NLP_subproblem(solve_data, config):
         rhs = ((0 if c.upper is None else c.upper)
                + (0 if c.lower is None else c.lower))
         sign_adjust = 1 if value(c.upper) is None else -1
-        fix_nlp.tmp_duals[c] = sign_adjust * max(0,
-                sign_adjust * (rhs - value(c.body)))
+        fix_nlp.tmp_duals[c] = sign_adjust * max(
+            0, sign_adjust*(rhs - value(c.body)))
+        pass
         # TODO check sign_adjust
```
RomeoV, line 57: Can we resolve this `#TODO`? I think I put it there originally because I didn't fully understand which way the sign needed to be, but if we can verify it's correct we can remove it.

RomeoV: Since we use `sign_adjust` the same way everywhere I guess the sign is probably correct, but we might think about referring to a little derivation somewhere, since it's not super obvious which way the sign needs to be.

pyomo/contrib/mindtpy/single_tree.py
```diff
@@ -0,0 +1,257 @@
...
+                         use_slack_var=False):
+        """Add oa_cuts through Cplex inherent function self.add()"""
+
+        for (constr, dual_value) in zip(target_model.MindtPy_utils.constraint_list,
+                                        dual_values):
+            if constr.body.polynomial_degree() in (0, 1):
+                continue
+
+            constr_vars = list(identify_variables(constr.body))
+            jacs = solve_data.jacobians
```
RomeoV, line 89: Can we guarantee this is always available? If I remember correctly we need to calculate them first.

pyomo/contrib/mindtpy/single_tree.py
```diff
@@ -0,0 +1,257 @@
...
+        for c in fix_nlp.tmp_duals:
+            if fix_nlp.dual.get(c, None) is None:
+                fix_nlp.dual[c] = fix_nlp.tmp_duals[c]
+        dual_values = list(fix_nlp.dual[c]
+                           for c in fix_nlp.MindtPy_utils.constraint_list)
+
+        main_objective = next(
+            fix_nlp.component_data_objects(Objective, active=True))
+        if main_objective.sense == minimize:
+            solve_data.UB = min(value(main_objective.expr), solve_data.UB)
```
RomeoV, line 176: In line 149 you use `self.get_objective_value()` - can we do the same here?

pyomo/contrib/mindtpy/single_tree.py
```diff
@@ -0,0 +1,257 @@
...
+
+        # add oa cuts
+        if fix_nlp_result.solver.termination_condition is tc.optimal:
+            self.handle_lazy_NLP_subproblem_optimal(
+                fix_nlp, solve_data, config, opt)
+        elif fix_nlp_result.solver.termination_condition is tc.infeasible:
+            self.handle_lazy_NLP_subproblem_infeasible(
+                fix_nlp, solve_data, config, opt)
+        else:
+            # TODO
```
RomeoV, line 256: I guess there should be something here?

pyomo/contrib/mindtpy/tests/test_mindtpy.py
```diff
@@ -34,11 +38,12 @@ def test_OA_8PP(self):
...
                       init_strategy='rNLP',
                       mip_solver=required_solvers[1],
-                      nlp_solver=required_solvers[0])
+                      nlp_solver=required_solvers[0],
+                      bound_tolerance=1E-5)
 
             # self.assertIs(results.solver.termination_condition,
             #               TerminationCondition.optimal)
-            self.assertTrue(fabs(value(model.cost.expr) - 68) <= 1E-2)
+            self.assertAlmostEqual(value(model.cost.expr), 68, places=1)
```
RomeoV, line 46: Could this be replaced with `model.get_objective_value()`? It should be the same of course, I'm just curious.

pyomo/contrib/mindtpy/tests/test_mindtpy_lp_nlp.py
```diff
@@ -0,0 +1,155 @@
...
+            print('\n Solving problem with Outer Approximation')
+            opt.solve(model, strategy='OA',
+                      init_strategy='rNLP',
+                      mip_solver=required_solvers[1],
+                      nlp_solver=required_solvers[0],
+                      bound_tolerance=1E-5,
+                      single_tree=True)
+
+            # self.assertIs(results.solver.termination_condition,
+            #               TerminationCondition.optimal)
```
RomeoV, line 45: Why do we not need to check for this anymore?

ZedongPeng: Sure, we need to check the termination condition. I have added them back.

pyomo/contrib/mindtpy/single_tree.py
ZedongPeng: `self.get_objective_value()` is an function from cplex package. It is used to get the objective value of the current integer solution in the branch and bound tree. When solving master problem, we need to use `self.get_objective_value()`. However, when we are solving subproblems, we still need to use the functions in pyomo.

pyomo/contrib/mindtpy/single_tree.py
ZedongPeng: Thanks for reminding me. I have added `self.handle_lazy_NLP_subproblem_other_termination` here.

pyomo/contrib/mindtpy/tests/test_mindtpy.py
ZedongPeng: Pyomo has no function of `model.get_objective_value()`. The same answer as above.

pyomo/contrib/mindtpy/single_tree.py
ZedongPeng: Thanks for the advice.

pyomo/contrib/mindtpy/tests/test_mindtpy_lp_nlp.py
ZedongPeng: Thanks for the advice.

pyomo/contrib/mindtpy/single_tree.py
ZedongPeng: Thanks for the advice and the code.

pyomo/contrib/mindtpy/initialization.py
```diff
@@ -82,10 +93,14 @@ def init_rNLP(solve_data, config):
...
                solve_data.LB, solve_data.UB))
         if config.strategy == 'OA':
-            copy_var_list_values(m.MindtPy_utils.variable_list, 
+            copy_var_list_values(m.MindtPy_utils.variable_list,
                                  solve_data.mip.MindtPy_utils.variable_list,
                                  config, ignore_integrality=True)
             add_oa_cuts(solve_data.mip, dual_values, solve_data, config)
+            # TODO check if value of the binary or integer varibles is 0/1 or integer value.
+            for var in solve_data.mip.component_data_objects(ctype=Var):
+                if var.domain.name == 'Integer' or var.domain.name == 'Binary':
```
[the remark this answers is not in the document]
ZedongPeng: Thanks. I have corrected this. I have also checked this is the only case in MindtPy.

pyomo/contrib/mindtpy/MindtPy.py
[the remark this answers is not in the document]
ZedongPeng: These are the solvers we have tried. Once more solvers have been tried. We will update the list.

jsiirola:
Why are you deleting the following tests?
-  pyomo/contrib/mindtpy/tests/flay03m.py 
- pyomo/contrib/mindtpy/tests/example_PSE.py
- pyomo/contrib/mindtpy/tests/batchdes.py
- pyomo/contrib/mindtpy/tests/alan.py

bernalde:
> Why are you deleting the following tests?
> 
> * pyomo/contrib/mindtpy/tests/flay03m.py
> * pyomo/contrib/mindtpy/tests/example_PSE.py
> * pyomo/contrib/mindtpy/tests/batchdes.py
> * pyomo/contrib/mindtpy/tests/alan.py

Those files were never actually part of the MindtPy tests. Those are instances from MINLPlib that I left behind from the early developments of MindtPy and were just laying there untouched.

bernalde:
> Several comments that you can decide if you want to address; but one fix that must be made before merging (skip lp_nlp tests if cplex_persistent is not available) to fix test failures.

The commits have already addressed all the comments above. Thanks for the advice!

pyomo/contrib/mindtpy/single_tree.py
```diff
@@ -0,0 +1,253 @@
...
+                continue  # Skip fixed variables.
+            try:
+                v_val = self.get_values(
+                    opt._pyomo_var_to_solver_var_map[v_from])
+                v_to.set_value(v_val)
+                if skip_stale:
+                    v_to.stale = False
+            except ValueError:
+                # Snap the value to the bounds
+                if v_to.lb is not None and v_val < v_to.lb and v_to.lb - v_val <= config.zero_tolerance:
```
qtothec, line 45: I would recommend `v_to.has_lb()` in case the default value of `lb` ever changes.

pyomo/contrib/mindtpy/tests/eight_process_problem.py
qtothec: Since here it makes no difference, I will remove these.

Users will often submit models that do not have all bounds specified. @bernalde is it your intent to warn them and have them provide some bounding information, in that case? If so, it may be good to have some tests to that effect.

pyomo/contrib/mindtpy/MindtPy.py
```diff
@@ -196,7 +191,7 @@ class MindtPySolver(object):
         description="Tolerance on variable bounds."
     ))
     CONFIG.declare("zero_tolerance", ConfigValue(
-        default=1E-15,
+        default=1E-10,
```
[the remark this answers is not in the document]
bernalde: I think there should be one at least for the Metasolvers (GDPOpt, MindtPy, Pynumero).

pyomo/contrib/mindtpy/iterate.py
```diff
@@ -2,90 +2,122 @@
...
-                            <= (log[-i - 1] + progress_required)
-                            * sign_adjust):
-                        making_progress = False
-                    else:
+                handle_NLP_subproblem_other_termination(fix_nlp, fix_nlp_result.solver.termination_condition,
+                                                        solve_data, config)
+            # Call the NLP post-solve callback
+            config.call_after_subproblem_solve(fix_nlp, solve_data)
+
+            if config.strategy == 'PSC':
```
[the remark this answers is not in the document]
bernalde: The PSC implementation is going to be revisited in the next PR. We could either leave it there or remove it for this one. The same applies to GDB and ECP.

pyomo/contrib/mindtpy/tests/eight_process_problem.py
bernalde: Currently, the warning is raised by the MILP solvers when a variable is unbounded in the master problem (it may be bounded for the MINLP given the nonlinear constraints), and depending on each they add redundant bounds to avoid the problem being unbounded.
MindtPy by itself does not add those bounds since we do not know them a-priori. Maybe if we used some bound propagation (@michaelbynum) although I do not know what would be the efficiency effect of that.

pyomo/contrib/mindtpy/iterate.py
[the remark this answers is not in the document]
qtothec: Perhaps comment it out, so that it doesn't show up as uncovered lines of code?

pyomo/contrib/mindtpy/MindtPy.py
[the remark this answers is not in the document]
qtothec: Ok, this ought not to block merge of this PR, but wanted to pose the question.

pyomo/contrib/mindtpy/nlp_solve.py
```diff
@@ -45,14 +44,25 @@ def solve_NLP_subproblem(solve_data, config):
...
+    # tmp_duals are the value of the dual variables stored before using deactivate trivial contraints
+    # The values of the duals are computed as follows: (Complementary Slackness)
+    #
+    # | constraint | c_leq | status at x1 | tmp_dual  |
+    # |------------|-------|--------------|-----------|
+    # | g(x) <= b  | -1    | g(x1) <= b   | 0         |
+    # | g(x) <= b  | -1    | g(x1) > b    | b - g(x1) |
+    # | g(x) >= b  | +1    | g(x1) >= b   | 0         |
+    # | g(x) >= b  | +1    | g(x1) < b    | b - g(x1) |
+
```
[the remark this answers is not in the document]
ZedongPeng: The condition for the duals allows us to identify which side of the inequality is active for us to add the cuts.
We prefer to include the upper bound as the right hand side since we are considering c by default a (hopefully) convex function, which would make c >= lb a nonconvex inequality which we wouldn't like to add linearizations if we don't have to.

pyomo/contrib/mindtpy/single_tree.py
ZedongPeng: Since we initialize the iteration loop by evaluating the Jacobian, this case should never happen.

pyomo/contrib/mindtpy/MindtPy.py
```diff
@@ -365,8 +368,9 @@ def solve(self, model, **kwds):
             #     MindtPy.feas_inverse_map[n] = c
 
             # Create slack variables for OA cuts
-            lin.slack_vars = VarList(
-                bounds=(0, config.max_slack), initialize=0, domain=NonNegativeReals)
+            if config.add_slack:
+                lin.slack_vars = VarList(
+                    bounds=(0, config.max_slack), initialize=0, domain=NonNegativeReals)
```
qtothec: Change not required for approval:

This isn't necessary, since you just end up with a dimension 0 singleton VarList.

bernalde: We decided to you remove the variables since it pollutes the master problem with an empty variable definition

pyomo/contrib/mindtpy/mip_solve.py
```diff
@@ -45,13 +45,11 @@ def solve_OA_master(solve_data, config):
...
-    sign_adjust = 1 if main_objective.sense == minimize else -1
-    if MindtPy.find_component('MindtPy_oa_obj') is not None:
-        del MindtPy.MindtPy_oa_obj
+    sign_adjust = 1 if main_objective.sense == minimize else - 1
+    MindtPy.del_component('MindtPy_oa_obj')
 
-    if config.add_slack == True:
-        if MindtPy.find_component('MindtPy_penalty_expr') is not None:
-            del MindtPy.MindtPy_penalty_expr
+    if config.add_slack is True:
```
jsiirola, line 51: @qtothec: Unless the config can take something other than True/False (e.g., is 'None' an acceptable value?), I think you should use Python's build in bool context and not check that the value is exactly True/False.  That is:
```python
if config.add_slack:
```
The same comment applies to lines 73 and 90 below.

pyomo/contrib/mindtpy/mip_solve.py
```diff
@@ -60,20 +58,19 @@ def solve_OA_master(solve_data, config):
         MindtPy.MindtPy_oa_obj = Objective(
             expr=main_objective.expr + MindtPy.MindtPy_penalty_expr,
             sense=main_objective.sense)
-    elif config.add_slack == False:
+    elif config.add_slack is False:
```
jsiirola, line 61: This is redundant given the test above (and that `add_slack` is in domain bool with a default value).  It would be more clear to implement this as:
```python
else: # not config.add_slack
```

PR #1402, opened 2020-04-22
Performance: only compute `is_fixed()` when necessary

ruaridhw:
## Summary/Motivation:
- `is_fixed()` is a very expensive operation for large constraint expressions
- The default of `skip_trivial_constraint` is False so it isn't even being used
- Therefore don't compute it if we don't need it

- Use Python's shortcircuting to ignore `is_fixed()` computation
-

### Legal Acknowledgement

### DIFF
--- a/pyomo/solvers/plugins/solvers/cplex_direct.py
+++ b/pyomo/solvers/plugins/solvers/cplex_direct.py
@@ -291,9 +291,8 @@ class CPLEXDirect(DirectSolver):
         if not con.active:
             return None
 
-        if is_fixed(con.body):
-            if self._skip_trivial_constraints:
-                return None
+        if self._skip_trivial_constraints and is_fixed(con.body):
+            return None
 
         conname = self._symbol_map.getSymbol(con, self._labeler)
 


ruaridhw:
> This is good. It would be nice if there was a test that covers this code, but as this was uncovered code to begin with, I will not hold up this PR waiting for it.

@jsiirola, I'd be happy to write one however the way I would typically test this is to use `unittest.mock.patch()` to mock `is_fixed()` and assert that it is now never called. I noticed though that Pyomo uses `pyutilib.th` rather than importing `unittest` directly. Is there a reason for not exposing/using `unittest.mock`? If so, how would you go about testing this?

jsiirola:
@ruaridhw, the test I was really thinking about was actually different (testing that trivial constraints were skipped in the interface when `_skip_trivial_constraints` was True), and is rather unrelated to this PR.  Although, the test you are proposing would be good, too.

This has been rendered somewhat moot (the PR has been merged), but to answer your questions: `pyutilib.th` derives from `unittest` in order to implement some categorization functionality that is not (easily) handled by `unittest` or `nose` and to provide additional `assert` methods.  There should be no problems using other functionality from `unittest`, and if `pyutilib.th` should import additional things from `unittest` to make the api more compatible we can handle that on a PR to PyUtilib.

As for compatibility, Pyomo currently supports Python 2.7 and 3.4+, and as such, we generally expect the tests to also support all versions of Python, unless they are testing a Python version-specific part of the Pyomo codebase.

ruaridhw:
> There should be no problems using other functionality from unittest, and if pyutilib.th should import additional things from unittest to make the api more compatible we can handle that on a PR to PyUtilib.

@jsiirola, see PyUtilib/pyutilib#98

PR #1403, opened 2020-04-22
`_collect_linear()` does not account for fixed variables' coefficients

### LINKED ISSUE #1399: `_collect_linear()` does not account for fixed variables' coefficients
`_collect_linear()` is not computing the `constant` term correctly when fixed variables have coefficients. This is resulting in the wrong LP file being written when using `quicksum()` (or `LinearExpression`s in general) in constraint rules.

### Current

```python
for c,v in zip(exp.linear_coefs, exp.linear_vars):
    if v.fixed:
        if compute_values:
            ans.constant += multiplier*v.value
        else:
            ans.constant += multiplier*v
    ...
```

### Suggested

```python
for c,v in zip(exp.linear_coefs, exp.linear_vars):
    if v.fixed:
        if compute_values:
            ans.constant += multiplier * c * value(v)
        else:
            ans.constant += multiplier * c * v
    ...
```

ruaridhw:
-
-

### Legal Acknowledgement

### DIFF
--- a/pyomo/repn/standard_repn.py
+++ b/pyomo/repn/standard_repn.py
@@ -867,9 +867,9 @@ def _collect_linear(exp, multiplier, idMap, compute_values, verbose, quadratic):
     for c,v in zip(exp.linear_coefs, exp.linear_vars):
         if v.fixed:
             if compute_values:
-                ans.constant += multiplier*v.value
+                ans.constant += multiplier * value(c) * value(v)
             else:
-                ans.constant += multiplier*v
+                ans.constant += multiplier * c * v
         else:
             id_ = id(v)
             if id_ in idMap[None]:
--- a/pyomo/repn/tests/test_standard.py
+++ b/pyomo/repn/tests/test_standard.py
@@ -4000,7 +4000,7 @@ class Test(unittest.TestCase):
         m = ConcreteModel()
         m.A = RangeSet(5)
         m.v = Var(m.A, initialize=1)
-        m.p = Param(m.A, initialize={1:-2, 2:-1, 3:0, 4:1, 5:2})
+        m.p = Param(m.A, initialize={1: -2, 2: -1, 3: 0, 4: 1, 5: 2})
 
         e = summation(m.v) + sum_product(m.p, m.v)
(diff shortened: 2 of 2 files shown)


pyomo/repn/tests/test_standard.py
```diff
@@ -4010,9 +4010,9 @@ def test_linear1(self):
 
         m.v[1].fixed=True
         rep = generate_standard_repn(e, compute_values=True)
-        self.assertEqual(str(rep.to_expression()), "2 + v[3] + 2*v[4] + 3*v[5]")
+        self.assertEqual(str(rep.to_expression()), "-1 + v[3] + 2*v[4] + 3*v[5]")
         rep = generate_standard_repn(e, compute_values=False)
-        self.assertEqual(str(rep.to_expression()), "v[1] + v[1] + v[3] + 2*v[4] + 3*v[5]")
+        self.assertEqual(str(rep.to_expression()), "v[1] - 2*v[1] + v[3] + 2*v[4] + 3*v[5]")
```
ruaridhw, line 4015: Unless I'm mistaken, this test was asserting that the behaviour was wrong.

If the expression is `- v[1] + v[3] + 2*v[4] + 3*v[5]` and we fix `v[1]` to `1` then I can't see why the constant term is `2`.

michaelbynum: @ruaridhw I agree.

pyomo/repn/standard_repn.py
```diff
@@ -867,9 +867,9 @@ def _collect_linear(exp, multiplier, idMap, compute_values, verbose, quadratic):
     for c,v in zip(exp.linear_coefs, exp.linear_vars):
         if v.fixed:
             if compute_values:
-                ans.constant += multiplier*v.value
+                ans.constant += multiplier * c * value(v)
```
michaelbynum, line 870: @ruaridhw If `compute_values` is True, then we probably want `value(c)` in case c is a parameter.

ruaridhw: Good point, thanks!

PR #1405, opened 2020-04-22
Fix polynomial degree and `_is_fixed()` of linear expressions

### LINKED ISSUE #1362: `quicksum()` causes trivial constraints to remain in the model
I have a constraint with a `body` of the following form:

```
var_1[x] - (var_2[x] + (1/6*var_3[x,y] + 1/6*var_3[x,y]))
```

When all of the variables are fixed, I would like the constraint to be deactivated.

This functionality already appears to be implemented with `TrivialConstraintDeactivator` and/or `skip_trivial_constraints=True` which both check `body.polynomial_degree() == 0` as the test for whether the constraint should be skipped.

As it happens, I'm finding that when all the variables are fixed I get the following:

```python
from pyomo.core.expr.visitor import polynomial_degree, _expression_is_fixed
assert polynomial_degree(body) == 1
assert _expression_is_fixed(body) is True
```

That is, the degree is non-zero yet the constraint is also wholly fixed. I've narrowed this down to the final two terms forming a `LinearExpression` for which the above assertions still hold.

Most importantly, if I rewrite the building of the constraint so that the final two terms fall under a `SumExpression` instead, `polynomial_degree(body) == 0` as expected. I've done this by swapping out `quicksum()` for `builtins.sum()`.

Is this a bug or is this expected behaviour? If it's the latter, performance is an issue for this constraint so what would be the suggested approach?

ruaridhw:
### Polynomial Degree
- `LinearExpression` causes trivial constraints to remain in the model
- This is because their polynomial degree evaluates to 1 even when all vars are fixed
- It should instead evaluate to 0

### Is Fixed
- Conversely, `LinearExpression` causes non-trivial constraints to be removed from the model
- This is because when it is part of another expression, `LinearExpression._is_fixed()` evaluates to `True` because `values` (ie its children) is empty and `all([]) is True`
- It should instead evaluate to the same as `LinearExpression.is_fixed()`

-
-

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/tests/unit/test_numeric_expr.py
+++ b/pyomo/core/tests/unit/test_numeric_expr.py
@@ -16,6 +16,8 @@ import pickle
 import math
 import os
 import re
+from collections import defaultdict
+
 import six
 import sys
 from os.path import abspath, dirname
@@ -5211,6 +5213,26 @@ class TestDirect_LinearExpression(unittest.TestCase):
         self.assertAlmostEqual(repn.constant, 1.0)
         self.assertTrue(len(repn.linear_coefs) == N)
         self.assertTrue(len(repn.linear_vars) == N)
+
+    def test_LinearExpression_polynomial_degree(self):
+        m = ConcreteModel()
+        m.S = RangeSet(2)
+        m.var_1 = Var(initialize=0)
+        m.var_2 = Var(initialize=0)
+        m.var_3 = Var(m.S, initialize=0)
+
+        def con_rule(model):
+            return model.var_1 - (model.var_2 + sum_product(defaultdict(lambda: 1/6), model.var_3)) <= 0
+
+        m.c1 = Constraint(rule=con_rule)
+
+        m.var_1.fix(1)
+        m.var_2.fix(1)
+        m.var_3.fix(1)
+
+        self.assertTrue(is_fixed(m.c1.body))
+        self.assertEqual(polynomial_degree(m.c1.body), 0)
+
 
 if __name__ == "__main__":
     unittest.main()
--- a/pyomo/core/tests/unit/test_numeric_expr.py
+++ b/pyomo/core/tests/unit/test_numeric_expr.py
@@ -5233,6 +5233,25 @@ class TestDirect_LinearExpression(unittest.TestCase):
         self.assertTrue(is_fixed(m.c1.body))
         self.assertEqual(polynomial_degree(m.c1.body), 0)
 
+    def test_LinearExpression_is_fixed(self):
+        m = ConcreteModel()
+        m.S = RangeSet(2)
+        m.var_1 = Var(initialize=0)
+        m.var_2 = Var(initialize=0)
+        m.var_3 = Var(m.S, initialize=0)
+
+        def con_rule(model):
+            from collections import defaultdict
+            return model.var_1 - (model.var_2 + sum_product(defaultdict(lambda: 1 / 6), model.var_3)) <= 0
+
+        m.c1 = Constraint(rule=con_rule)
+
+        m.var_1.fix(1)
+        m.var_2.fix(1)
+
+        self.assertFalse(is_fixed(m.c1.body))
+        self.assertEqual(polynomial_degree(m.c1.body), 1)
+
 
 if __name__ == "__main__":
     unittest.main()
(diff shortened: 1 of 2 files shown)


michaelbynum:
@ruaridhw Thanks again for this PR.

michaelbynum:
The more I look at this, the more I'm think the `is_fixed` and `_is_fixed` methods can use the base class methods. Thoughts?

ruaridhw:
> The more I look at this, the more I'm think the `is_fixed` and `_is_fixed` methods can use the base class methods. Thoughts?

Yeah I thought about this a bit too. The conclusion I came to was:

1. To use the existing logic, a sensible `values` to `_is_fixed()` would have to be provided that actually pertains to the expression's children. This would require modifying `_args_` and `nargs()` so that `ExpressionValueVisitor()` actually walks the `LinearExpression` (which could likely break other places too and be much less performant).
1. `_is_fixed()` specifically says that it can (and should) be overridden by subclasses.
1. My interpretation is that `LinearExpression` is quite a bit more performant than the other expression classes. Especially since it is "walked" much more easily, we may as well make use of the easy access to `linear_vars` rather than triggering yet another traversal with a call to `is_fixed()`

jsiirola:
I am fighting with LinearExpression at the moment, too.  The problem with it is that it is a weird hybrid between an ExpressionBase node and a leaf node.  In a normal s-expression system, expression nodes are internal tree nodes and the atoms (leaf nodes) are constants/variables/etc.  The problem is that a LinearExpression is both.  Sort of.  Logically, it is an expression, as you and I know it is an operator.  BUT, you cannot descend into it in any meaningful way: the logical children would be monomial terms (`c_i * x_i`), and the whole point of the LinearExpression object is to never form those nodes.  For this issue, I *think* the solution is to start treating LinearExpression as a (compicated) atom and *not* as an Expression, but that is a significant change to a lot of things that touch the expression system.

As a side note, LinearExpression is performant exactly because you can reduce a summation from N+1 nodes to a single node.  That significantly reduces the tree size and simplifies walking it and processing it.  But, before we start pushing people to use it more often, we need to resolve how it should be handled by the expression system (and in particular all the walkers).  I for one and mot thrilled that -- at least currently -- it has to be handled as a special case in virtually every walker.

I am working on some documentation of components / expressions / etc, so hopefully we will be able to codify some of this soon.

pyomo/core/tests/unit/test_numeric_expr.py
```diff
@@ -5212,5 +5214,44 @@ def test_LinearExpression_expression(self):
...
+
+    def test_LinearExpression_is_fixed(self):
+        m = ConcreteModel()
+        m.S = RangeSet(2)
+        m.var_1 = Var(initialize=0)
+        m.var_2 = Var(initialize=0)
+        m.var_3 = Var(m.S, initialize=0)
+
+        def con_rule(model):
+            from collections import defaultdict
```
jsiirola, line 5244: This import is redundant to the import above.

pyomo/core/tests/unit/test_numeric_expr.py
```diff
@@ -5212,5 +5214,44 @@ def test_LinearExpression_expression(self):
...
+            return model.var_1 - (model.var_2 + sum_product(defaultdict(lambda: 1 / 6), model.var_3)) <= 0
+
+        m.c1 = Constraint(rule=con_rule)
+
+        m.var_1.fix(1)
+        m.var_2.fix(1)
+
+        self.assertFalse(is_fixed(m.c1.body))
```
jsiirola, line 5252: This test is failing on Python 2.7 because `sum_product` is returning an empty `LinearExpression`.  This is probably indicative of a bug in `sum_product`?

michaelbynum:
I see. That all makes sense.

pyomo/core/tests/unit/test_numeric_expr.py
ruaridhw: Thanks, fixed

pyomo/core/tests/unit/test_numeric_expr.py
ruaridhw: Nah it's because in Py2 `1 / 6 = 0` 😫 Maybe I'm biased but now that 2.7 is four months past its End of Life... 😉

ruaridhw:
> it has to be handled as a special case in virtually every walker

Personally I didn't see this as a particularly bad design. Say I implement my own `ExpressionBase` as a special "cached" expression that "remembers" the properties of its children and expects them to be immutable. I would certainly want to override the typical walking behaviour so as not to traverse the children and just use the cached information instead. In that scenario, I too would expect to override `_is_fixed()`, `_compute_polynomial_degree()` and any other methods required by the walkers to exploit the performance of the cache.

Issue #1160, opened 2019-11-07
Fast(er) direct solver interfaces


whart222:
I think users would generally expect direct solver interfaces to be faster, since we're avoiding file I/O.  However, I think this is not always the case.  For example, see #515 

We need to critique our direct solver interfaces and assess whether we can setup/analyze problems more effectively.  This activity could involve:

1. Identifying bottlenecks in existing solver interfaces
2. Exploring strategies for avoiding canonicalization in Pyomo (e.g. directly generating solver expressions from Pyomo expression trees)
3. Interfacing with POEK/COEK to see if their interfaces can resolve known bottlenecks.

NOTE: Aside from #51 and #1169, I only have anecdotal evidence that direct solver interfaces aren't faster.  I would recommend confirming that they *are* faster before closing this issue.

emma58:
I'm not sure whether @michaelbynum's recent enhancements change this, but the other thing that can make the persistent interface slower is hacking around functionality that's not yet in the persistent interface. The obvious one is changing the right-hand side of a constraint: The only way I know how to do this now is to remove and add the constraint, but I've never found this strategy to be worth the time. I just point this out because I wonder if there aren't other cases where the potential performance enhancements might be in what is *not* there rather than what is.

qtothec:
I have been talking to Steve about making the GAMS direct "solver" faster by constructing in-memory GMO objects rather than writing a file, as well.

qtothec:
As a follow-up to this, Steve Dirkse sent me an example for building up and solving an NLP using in-memory GMO and GEV.

carldlaird:
I am not sure that #515 is the correct issue to reference. Does anyone know the issue number that shows a case where the direct solvers are slower than the indirect solvers?

qtothec:
Well, for the GAMS case, I have this: https://github.com/Pyomo/pyomo/pull/215#issuecomment-328152064

I don't know about the MIP solver cases.

whart222:
@carldlaird I updated the references to related issues.  Both of these concern the CPLEX python solver interface.

jsiirola:
Archived on the master Performance Proposals Issue (#1430).  Closing this performance proposal until active development has begun.

staadecker:
Fyi I know that for the large models I'm working with the direct gurobi solver interfaces is slower than the IO gurobi solver interface

PR #331, opened 2018-01-31
Improvements for Gurobi/Cplex Interfaces (mostly fixing tests and switching to bulk extraction of solution values)

michaelbynum:
## Summary/Motivation:
1) The slacks for range constraints was causing many tests to fail.
2) Bulk extraction of variable values, duals, etc. from Cplex/Gurobi is more efficient - especially for those using Gurobi cloud licenses.

## Changes proposed in this PR:
1) The direct/persistent solver interfaces now report slacks for range constraints. 
2) I converted much of the code that loads solutions from Gurobi/Cplex to extract variable values, duals, etc. in bulk rather than one at a time. This was done for the Cplex direct/persistent interfaces, the Gurobi direct/persistent interfaces, and for the Gurobi lp file interface.

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

Issue #51, opened 2016-10-11
With cplex, solver_io='python' is much slower than solver_io='nl' or default


mfripp:
A [comment on stackoverflow](http://stackoverflow.com/a/39925803/3830997) mentioned that CPLEX's python bindings should be faster than the 'nl' (AMPL-style) interface or the default (which I think interacts with the cplex executable directly), especially with Pyomo versions >= 4.3.11323. However, on my system I have not found this to be the case. In the example below, the solver takes  about 8x longer when using the solver_io='python' option than it does with the other options. This is on Mac OS X 10.11.6, with Python 2.7, CPLEX 12.6.0.0 and Pyomo 4.4.1.

    import time
    from pyomo.environ import *
    from pyomo.opt import SolverFactory

    # def arg_str(args, kwargs):
    #     args_str = ', '.join(str([a])[1:-1]
    #     kwargs_str = ', '.join(str(k)+"="+str([a])[1:-1] for k, a in kwargs.iteritems())
    #     sep = ', ' if args_str and kwargs_str else ''
    #     return args_str + sep + kwargs_str
    #     return ', '.join(
    #         [repr(a) for a in args] + [str(k) + "=" + repr(a) for k, a in kwargs.items()]
    #     )

    def solve(*args, **kwargs):
        # define the model
        n_steps = 100000
        m = ConcreteModel()
        m.STEPS = Set(initialize=range(n_steps))
        m.target = Param(m.STEPS, initialize=lambda m, s: 1.0)
        m.Build = Var()
        m.Operate = Var(m.STEPS)
        m.Meet_Target = Constraint(m.STEPS, rule=lambda m, s: m.Operate[s] >= m.target[s])
        m.Operate_Limit = Constraint(m.STEPS, rule=lambda m, s: m.Operate[s] <= m.Build)
        m.MinCost = Objective(rule=lambda m: n_steps * m.Build + sum(m.Operate[s] for s in m.STEPS), sense=minimize)

        # solve the model
        opt = SolverFactory(*args, **kwargs)
        start_time = time.time()
        results = opt.solve(m)
        m.solutions.load_from(results)

        print "solve({}): MinCost={}, time={}".format(
            ", ".join([repr(a) for a in args] + [str(k)+"="+repr(a) for k, a in kwargs.items()]),
            value(m.MinCost), time.time() - start_time)

    solve("cplex")                      # 9.2s
    solve("cplex", solver_io="nl")      # 9.2s
[...]

PR #1416, opened 2020-04-28
CPLEXDirect performance improvements

ruaridhw:
This PR contains a number of performance improvements to the `CPLEXDirect` class, and by extension, the `CPLEXPersistent` class.

# Notes

There are just a couple of functions (if not lines) that are resulting in very large bottlenecks. On the whole, things are quite performant as-is.

## Batching "transactions" with CPLEX

The `cplex` package's Linear Constraints and Variable interfaces allow for batched transactions. I think an appropriate design is to generate all the necessary data and add these objects as one call to the `solver_model`. I've also removed unnecessary transactions such as resetting variable bounds immediately after adding that variable with an obsolete bound.

## Querying results from CPLEX

Asking the `cplex` solution for *specific* variables' values is extremely slow due to the conversion on `cplex`'s end to lookup the variables you've asked for. It is orders of magnitude faster to get the full solution vector. Even worse, using the "specific variable interface" to get the full solution vector. If we must get specific variables (ie. `_load_vars()`) then their index should be used instead of their name.

## Expected performance vs LP solvers

It would be interesting to benchmark specifically writing the LP file vs the total time spent interfacing with the CPLEX library in isolation. I think it's a fallacy to suggest that the Direct interfaces *should* be faster because "there's no file IO" as suggested in GitHub issues and on SO. Once the repn has been generated, writing a string to disk takes no time at all. If anything I would imagine interfacing with a third-party library should be expected to take *more* time. To me, the gains are realised when using the Persistent interfaces to apply incremental resolves of a model. The LP solver has to regen the entire model whereas the Persistent interface only has to generate the changes.
Not to mention that a slight decline in performance is a small price to pay for access to the full advanced functionality of the `cplex` package for those who need it.

# Benchmark

- Ran the "benchmark-ish" case in #51 ten times for each solver:

| **Solver** | `master`                      | This branch |
| ------------- | ------------- | ------------- |
| **LP**     | 10.74s to 11.06s. Mean 10.84s | 10.94s to 11.27s. Mean 11.08s |
| **Direct** | 32.48s to 34.00s. _Mean 33.14s_ | 11.67s to 12.23s. _Mean 12.00s_ |

- There was still a speedup when I set `n_steps = 10` so it's not as though we're optimising for very large models only
- The total time is now almost wholly taken up by (a) model build and (b) `generate_standard_repn()` which both of these interfaces have to do (and is therefore out-of-scope here).
- There are still a few inefficiencies but I didn't bother digging / fixing to shave part-seconds on a large test case. For example, generators are slower than list comprehension.
	There are many places where generators are used where the iterable is sufficiently small to consume into memory (eg. `ComponentSet.update()`)
- Clearly the LP solver appears to for whatever reason gotten worse as well. I reran it as a control and I don't really have an explanation for that... Might need the reviewer's help in case I've somehow inadvertently impacted that solver.

### TODO

- [x] Unit tests

## Related

- #51
- #331
- #1160
- #1169

### Legal Acknowledgement

### DIFF
--- a/pyomo/solvers/plugins/solvers/cplex_direct.py
+++ b/pyomo/solvers/plugins/solvers/cplex_direct.py
@@ -48,7 +48,7 @@ class _CplexExpr(object):
     ):
         self.variables = variables
         self.coefficients = coefficients
-        self.offset = offset or 0.
+        self.offset = offset or 0.0
         self.q_variables1 = q_variables1 or []
         self.q_variables2 = q_variables2 or []
         self.q_coefficients = q_coefficients or []
--- a/pyomo/solvers/plugins/solvers/cplex_direct.py
+++ b/pyomo/solvers/plugins/solvers/cplex_direct.py
@@ -442,7 +442,6 @@ class CPLEXDirect(DirectSolver):
             self._objective = None
 
         self._solver_model.objective.set_linear([(i, 0.0) for i in range(len(self._pyomo_var_to_solver_var_map.values()))])
-        self._solver_model.objective.set_quadratic([[[0], [0]] for i in self._pyomo_var_to_solver_var_map.keys()])
 
         if obj.active is False:
             raise ValueError('Cannot add inactive objective to solver.')
--- a/pyomo/solvers/plugins/solvers/cplex_direct.py
+++ b/pyomo/solvers/plugins/solvers/cplex_direct.py
@@ -76,10 +76,7 @@ class _VariableData(object):
         self.types.append(type_)
         self.names.append(name)
 
-    def __enter__(self):
-        return self
-
-    def __exit__(self, *excinfo):
+    def store_in_cplex(self):
         self._solver_model.variables.add(
             lb=self.lb, ub=self.ub, types=self.types, names=self.names
         )
@@ -101,10 +98,7 @@ class _LinearConstraintData(object):
         self.range_values.append(range_values)
         self.names.append(name)
 
-    def __enter__(self):
-        return self
-
-    def __exit__(self, *excinfo):
+    def store_in_cplex(self):
         self._solver_model.linear_constraints.add(
             lin_expr=self.lin_expr,
             senses=self.senses,
@@ -112,27 +106,6 @@ class _LinearConstraintData(object):
             range_values=self.range_values,
             names=self.names,
         )
-
-
-# `nullcontext()` is part of the standard library as of Py3.7
-# This is verbatim from `cpython/Lib/contextlib.py`
-class nullcontext(object):
-    """Context manager that does no additional processing.
-    Used as a stand-in for a normal context manager, when a particular
-    block of code is only sometimes used with a normal context manager:
-    cm = optional_cm if condition else nullcontext()
-    with cm:
-        # Perform operation, using optional_cm if condition is True
-    """
-
-    def __init__(self, enter_result=None):
-        self.enter_result = enter_result
-
-    def __enter__(self):
-        return self.enter_result
-
-    def __exit__(self, *excinfo):
-        pass
 
 
 @SolverFactory.register('cplex_direct', doc='Direct python interface to CPLEX')
@@ -321,7 +294,7 @@ class CPLEXDirect(DirectSolver):
 
         return cplex_expr, referenced_vars
 
-    def _add_var(self, var, cplex_var_data=None):
+    def _add_var(self, var, var_data=None):
         varname = self._symbol_map.getSymbol(var, self._labeler)
         vtype = self._cplex_vtype_from_var(var)
         if var.has_lb():
@@ -337,13 +310,12 @@ class CPLEXDirect(DirectSolver):
             lb = value(var)
             ub = value(var)
 
-        ctx = (
-            _VariableData(self._solver_model)
-            if cplex_var_data is None
-            else nullcontext(cplex_var_data)
+        cplex_var_data = (
+            _VariableData(self._solver_model) if var_data is None else var_data
         )
-        with ctx as cplex_var_data:
-            cplex_var_data.add(lb=lb, ub=ub, type_=vtype, name=varname)
+        cplex_var_data.add(lb=lb, ub=ub, type_=vtype, name=varname)
+        if var_data is None:
+            cplex_var_data.store_in_cplex()
 
         self._pyomo_var_to_solver_var_map[var] = varname
         self._solver_var_to_pyomo_var_map[varname] = var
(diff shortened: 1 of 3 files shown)


blnicho:
We're going to mark this as a WIP until tests are added.

ruaridhw:
This is ready for review though I'm not sure whom would be best placed. Perhaps @michaelbynum?

pyomo/solvers/plugins/solvers/cplex_direct.py
```diff
@@ -287,7 +382,49 @@ def _set_instance(self, model, kwds={}):
                             "by overwriting its bounds in the CPLEX instance."
                             % (var.name, self._pyomo_model.name,))
 
-    def _add_constraint(self, con):
+    def _add_block(self, block):
+        with _VariableData(self._solver_model) as cplex_var_data:
+            for var in block.component_data_objects(
+                ctype=pyomo.core.base.var.Var, descend_into=True, active=True, sort=True
```
jsiirola, line 388: Here (and elsewhere), I believe that you can make this more performant (avoid unnecessary sorting) by using `sort=SortComponents.deterministic` instead of `True`.

ruaridhw:
> Overall I think this is really good. My biggest concern is around the licensing of `nullcontext`, but I think that is easy to get around.

Interesting! I never thought about there being an issue with copying CPython code given all of its standard libraries are used anyway. Happy to drop the contextmanager design, your alternative is much more explicit.

> One final question: your new tests all rely on mock. Are there at least a few tests of this functionality that actually engage the cplex solver (just so that our tests will catch if the underlying cplex argument API drifts)?

All of the tests engage the CPLEX solver. `mock` is only used to "spy" on the API calls to the CPLEX solver rather than mocking the solver object. The tests can therefore assert on the underlying arguments that Pyomo provided CPLEX in addition to ensuring the API call actually worked. This is illustrated by assertions such as:

```python
self.assertEqual(opt._solver_model.linear_constraints.get_num(), 1)
```
which would fail if `_solver_model` was not an actual `cplex.Cplex()` object.

pyomo/solvers/plugins/solvers/cplex_direct.py
ruaridhw: Happy to change this, just note it would be a behaviour change whereas the rest of this MR maintains identical behaviour to the existing implementation.

jsiirola: Good point.  Preserving behavior is a good thing.  We can revisit sorting later (probably as part of #677 / #1030).

ruaridhw:
> Happy to drop the contextmanager design, your alternative is much more explicit.

I've updated accordingly.

> the rest of this MR maintains identical behaviour to the existing implementation

Just to add a caveat to this... The _user-facing_ behaviour is identical however we are now adding linear constraints to the CPLEX object _after_ quadratic and SOS constraints. I guess it's fine for this to change though since this is an internal implementation detail and I don't think CPLEX is actually impacted if this order is changed. The order of the constraints _within_ each interface (Linear, Quadratic, SOS) is still the same.

pyomo/solvers/plugins/solvers/cplex_direct.py
```diff
@@ -426,7 +539,6 @@ def _set_objective(self, obj):
             self._objective = None
 
         self._solver_model.objective.set_linear([(i, 0.0) for i in range(len(self._pyomo_var_to_solver_var_map.values()))])
-        self._solver_model.objective.set_quadratic([[[0], [0]] for i in self._pyomo_var_to_solver_var_map.keys()])
```
michaelbynum, line 429: @ruaridhw I don't think this line should be removed completely. However, it can be simplified to:

```
self._solver_model.objective.set_quadratic([0]*len(self._pyomo_var_to_solver_var_map))
```

If this line is not included, then the quadratic part of the objective will not be updated correctly between solves (for the persistent interface).

ruaridhw: @michaelbynum, could you provide a small example of how this would break? I'd like to add a test case to that effect.

michaelbynum: @ruaridhw  This should demonstrate the issue:

```
import pyomo.environ as pe

m = pe.ConcreteModel()
m.x = pe.Var(bounds=(-2, 2))
m.y = pe.Var(bounds=(-2, 2))
m.obj = pe.Objective(expr=m.x**2 + m.y**2)
m.c1 = pe.Constraint(expr=m.y >= 2*m.x - 1)
m.c2 = pe.Constraint(expr=m.y >= -m.x + 2)
opt = pe.SolverFactory('cplex_persistent')
opt.set_instance(m)
opt.solve()
print(m.x.value, m.y.value)  # should be 1, 1
del m.obj
m.obj = pe.Objective(expr=m.x**2)
opt.set_objective(m.obj)
opt.solve()
print(m.x.value, m.y.value)  # should be 0, 2 but result is 1, 1
opt.set_instance(m)
opt.solve()
print(m.x.value, m.y.value)  # to demonstrate that the result should be 0, 2
```

ruaridhw: @michaelbynum, I've added a fix and this test case. The main issue I had was that the quadratic objective interface shouldn't be triggered if the model is a MILP. Let me know if 1b555dc is acceptable.

michaelbynum:
I think the test failures are due to older cplex versions. It looks like `set_objective([0]*num_cols)` only works in newer versions of cplex....

jsiirola:
> I think the test failures are due to older cplex versions. It looks like `set_objective([0]*num_cols)` only works in newer versions of cplex....

Per the CPLEX [documentation](https://www.ibm.com/support/knowledgecenter/SSSA5P_12.7.1/ilog.odms.cplex.help/refpythoncplex/html/cplex._internal._subinterfaces.ObjectiveInterface-class.html#set_quadratic):
> If the quadratic objective function is separable, the entries of the list must all be of type float

I looked in `lib/python3.5/site-packages/cplex/_internal/_subinterfaces.py`, and there is an explicit test against the `float` type.

Changing `[0] * num_vars` to `[0.] * num_vars` should fix things.

michaelbynum:
@jsiirola Oh, good catch!

ruaridhw:
Great spot, thanks both. I've added explicit conversion of the coeffs to `float` inside `_CplexExpr` as well.

jsiirola:
Fantastic!  Once this round of tests pass we will merge this.

jsiirola:
The one failing test was due to a codecov upload problem.  Merging.

PR #1418, opened 2020-04-29
Add ConfigEnum class

Robbybp:
## Summary/Motivation:
Sometimes would like to accept user input in form of an Enum, int, or string, each of which is translated to the corresponding Enum value.

- Add ConfigEnum class, which implements the from_enum_or_string method, intended for use as a domain in a ConfigValue
- Test

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/config.py
+++ b/pyomo/common/config.py
@@ -11,6 +11,7 @@
 import os
 import platform
 
+import enum
 import six
 from pyutilib.misc.config import ConfigBlock, ConfigList, ConfigValue
 
@@ -157,3 +158,14 @@ def add_docstring_list(docstring, configblock, indent_by=4):
             indent_spacing=0,
             width=256
         ).splitlines(True))
+
+
+class ConfigEnum(enum.Enum):
+    @classmethod
+    def from_enum_or_string(cls, arg):
+        if type(arg) is str:
+            return cls[arg]
+        else:
+            # Handles enum or integer inputs
+            return cls(arg)
+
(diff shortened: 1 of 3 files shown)


pyomo/common/config.py
```diff
@@ -11,6 +11,7 @@
 import os
 import platform
 
+import enum
```
jsiirola, line 14: FYI: This places a hard dependency on `enum34` for Python 2.7.  I am not entirely sure how to resolve this.

Robbybp: This is an unfamiliar problem for me. The issue is that `enum` was added in Python 3.4 and backported to Python 2.7, but there is no guarantee that a user on 2.7 will have the backported enum module? 

My first thought would be to put the import and class definition in a try-except, then mark the test with skipIf.

jsiirola: The problem is that this isn't in a test: you need `enum` in the *definition* of the `ConfigEnum` class, so the entire `ConfigEnum` class wouldn't be available if enum wasn't present (only an issue for Python 2.7 at this point).  I think the right answer is to add enum34 to the dependencies for python2.7 in setup.py, but adding hard dependencies is something that we should at least make sure more developer weigh in on.

I will update this PR to include the new dependency and then wait for others to weigh in.

qtothec: @jsiirola It helps that you are a main proponent of keeping 2.7 support in the first place. Fwiw, I believe it makes sense to keep support as long as it doesn't make life too difficult. Most people are content if you don't mess with 3.x.

michaelbynum:
I think a dependency on enum34 is perfectly reasonable.

PR #1419, opened 2020-05-04
PyNumero: HSL interface

Robbybp:
## Summary/Motivation:
Would like to use HSL MA27 and MA57 linear solvers with PyNumero. 

- C interfaces to MA27 and MA57
- Very basic makefile to compile `libpynumero_MA?7.so`
- Python interfaces to these libraries
- Tests (very simple - have only tested on the 5x5 matrix that HSL provides as an example for these codes)

## Yet to be implemented:
- MA97 interface
- Python interfaces to MA57's iterative refinement and memory reallocation functions, and tests for these functions
- Tests on more complicated matrices
- Incorporation into the PyNumero build harness, or any attempt at all to make this cross-platform
- More sophisticated Makefile/build script. Right now, MA27 and Metis build processes need to be modified to include -fPIC if they are to be used with the code in this PR. The user may need to set `LD_LIBRARY_PATH` to find some of the required external shared libraries (MA57/coinhsl). I had to do this to get it working with `libcoinhsl.so`.

### Legal Acknowledgement


pyomo/contrib/pynumero/src/hsl_interface/ma27Interface.c
```diff
@@ -0,0 +1,242 @@
...
+			&(ma27->LIW_b),
+			ma27->W,
+			&(ma27->MAXFRT),
+			RHS,
+			ma27->IW1,
+			&(ma27->NSTEPS), 
+			ma27->ICNTL, 
+			ma27->INFO
+			);
+
```
santiagoropb, line 187: Is there a particular reason not to have an iterative refinement subroutine here like the one you have for ma57?

pyomo/contrib/pynumero/src/hsl_interface/ma27Interface.c
```diff
@@ -0,0 +1,242 @@
...
+	if (!ma27->A_allocated) {
+		int info5 = ma27->INFO[5-1];
+		int size = (int)(ma27->A_factor*info5);
+		alloc_a(ma27, size);
+		// A is now allocated
+	}
+	// Regardless of ma27->A's previous allocation status, copy values from A.
+	memcpy(ma27->A, A, NZ*sizeof(double));
+
+	if (!ma27->IW_b_allocated) {
```
santiagoropb, line 138: will this allow for allocating more memory from the python side? If I recall correctly  for large ill conditioned problems A and IW_a  typically need to be reallocated

pyomo/contrib/pynumero/src/hsl_interface/ma57Interface.c
```diff
@@ -0,0 +1,343 @@
...
+			ma57->IWORK,
+			ma57->ICNTL,
+			ma57->CNTL,
+			ma57->INFO,
+			ma57->RINFO
+			);
+
+	free(ma57->IWORK);
+	free(ma57->WORK);
+	ma57->WORK_allocated = false;
```
santiagoropb, line 256: My guess is ma27 does not have a subroutine like this one?

pyomo/contrib/pynumero/src/hsl_interface/ma57Interface.c
```diff
@@ -0,0 +1,343 @@
...
+}
+
+void do_reallocation(struct MA57_struct* ma57, int N, double realloc_factor, int IC) {
```
santiagoropb, line 259: cool

pyomo/contrib/pynumero/src/hsl_interface/ma27Interface.c
```diff
@@ -0,0 +1,242 @@
...
+void set_cntl(struct MA27_struct* ma27, int i, double val) {
+	ma27->CNTL[i] = val;
+}
+double get_cntl(struct MA27_struct* ma27, int i) {
+	return ma27->CNTL[i];
+}
+int get_info(struct MA27_struct* ma27, int i) {
+	return ma27->INFO[i];
+}
+
```
santiagoropb, line 67: Can the number of negative eigenvalues be gotten from get_info?

pyomo/contrib/pynumero/src/hsl_interface/Makefile
```diff
@@ -0,0 +1,54 @@
+rootdir:=$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
```
santiagoropb, line 1: I am curious why using makefiles and not include this in the cmake project for pynumero to build a separate library for hsl

michaelbynum:
This is phenomenal, @Robbybp. Nice work.

pyomo/contrib/pynumero/src/hsl_interface/ma27Interface.c
Robbybp: Yes, you can allocate more memory from the Python side by calling `alloc_a` and `alloc_iw_a` via ctypes. For MA57 this can be done by either calling the `alloc_*` functions or calling `do_reallocation`. I haven't gotten around to writing Python-side reallocation functions for either solver, however.

pyomo/contrib/pynumero/src/hsl_interface/ma57Interface.c
Robbybp: Correct

pyomo/contrib/pynumero/src/hsl_interface/ma27Interface.c
Robbybp: Yes, it is INFO(15) and (24) in MA27 and 57 respectively

pyomo/contrib/pynumero/src/hsl_interface/Makefile
Robbybp: This should be incorporated into the pynumero cmake project. The makefiles are basically just placeholders until I can learn enough cmake to incorporate these libraries into the existing project. 

Also, at some point I was planning to include targets for the dependency libraries/objects, e.g. `libmetis.a`, `ma27d.o`, etc, as some of these need special rules (adding -fPIC) if compiled individually and not as part of `libcoinhsl.so`. I thought this would be easier to do with makefiles than with cmake. Not sure if I'm still planning to do this, and regardless of how the dependencies are compiled, I'm still planning to have the pynumero libraries be part of the cmake project.

pyomo/contrib/pynumero/src/hsl_interface/ma27Interface.c
Robbybp: MA27 doesn't have a built-in iterative refinement subroutine, and I wanted the C code to just provide wrappers around the built-in routines. Iterative refinement can be implemented for MA27 on the Python side, but I haven't gotten around to this yet. (Haven't gotten around to writing a Python-side iterative refinement function for MA57 either, although this will hopefully be much simpler due to the built-in routine.)

santiagoropb: so m57 and mumps have that in the C side. It would be convenient to have the iterative refinement for ma27 in C too if possible. I implemented some subroutines for iterative refinement in python at some point and it was not very clean. It also add overhead given that the matrix vector products are fairly cheap

santiagoropb:
I agree with Michael. Very clean implementation. Nicely done

pyomo/contrib/pynumero/src/hsl_interface/ma27Interface.c
michaelbynum: I think doing the iterative refinement on the python side is a good approach. I'd like to keep the C code as simple as possible.

michaelbynum: Plus, we can always add it in later if needed. I think this is good for now.

Robbybp:
@jsiirola This is ready for you to take a look at when you get the chance. Note the following:
- Both .c and .cpp interfaces are included in the branch for now. `CMakeLists.txt` is setup to use the .cpp interfaces. One set will need to be deleted.
- I altered `CMakeLists.txt` to use a `MA27_OBJECT` if it is provided, and to look for `libma27` and `libma57` if `libcoinhsl` cannot be found. MA27 and MA57 don't compile with their dependencies by default, so the user will need to manually compile them if they want to use these individually. I figure that for now, any user who wants to do this is probably capable enough to figure it out.

pyomo/contrib/pynumero/extensions/ma57_interface.py
```diff
@@ -0,0 +1,242 @@
...
+            % (i, array_name, array_len))
+
+def validate_value(val, dtype, array_name=''):
+    if not isinstance(val, dtype):
+        raise ValueError(
+            'Members of %s array must have type %s. Got %s'
+            % (array_name, dtype, type(val)))
+
+class _NotSet:
+    pass
```
jsiirola, line 37: This is duplicated in the MA27 interface.  Can you extract it into a common location (e.g., `extensions/utils.py`)>

pyomo/contrib/pynumero/extensions/tests/test_ma27_interface.py
```diff
@@ -0,0 +1,140 @@
...
+        ent = np.array([2.,3.,4.,6.,1.,5.,1.], dtype=np.double)
+        ent_copy = ent.copy()
+        ma27.do_symbolic_factorization(n, irn, icn)
+
+        status = ma27.do_numeric_factorization(irn, icn, n, ent)
+        self.assertEqual(status, 0)
+
+        expected_ent = [2.,3.,4.,6.,1.,5.,1.,]
+        for i in range(ne):
+            self.assertAlmostEqual(ent_copy[i], expected_ent[i])
```
jsiirola, line 90: This is a strange test: you are asserting that the values in `ent` haven't changed?  You are comparing the copy of an array you explicitly create with another array you explicitly create...

pyomo/contrib/pynumero/extensions/tests/test_ma27_interface.py
```diff
@@ -0,0 +1,140 @@
...
+        # numeric factorization with the same ma27 struct
+        #
+        # n is still 5, ne has changed to 8.
+        irn = np.array([1,1,2,2,3,3,5,1], dtype=np.intc)
+        icn = np.array([1,2,3,5,3,4,5,5], dtype=np.intc)
+        ent = np.array([2.,3.,4.,6.,1.,5.,1.,3.], dtype=np.double)
+        status = ma27.do_symbolic_factorization(n, irn, icn)
+        self.assertEqual(status, 0)
+        status = ma27.do_numeric_factorization(irn, icn, n, ent)
+        self.assertEqual(status, 0)
```
jsiirola, line 117: Is there any way to test the result of numeric factorization?  Just asserting that the routine returned a `status == 0` seems weak.

pyomo/contrib/pynumero/extensions/tests/test_ma27_interface.py
Robbybp: Yes, I am asserting that the double array I passed to MA27 has not changed. This is the array whose memory was getting corrupted before I started memcpy-ing A. The rest of the process, including back solve, proceeded correctly, so this is my only test for such corruption.

jsiirola: Right, but shouldn't the test be:
```python
for i in range(ne):
    self.assertAlmostEqual(ent[i], expected_ent[i])
```
(that is, test the numpy array that you passed to ma27?)

pyomo/contrib/pynumero/extensions/tests/test_ma27_interface.py
Robbybp: Agreed that it's a pretty weak test, but I'm not sure what else I could do. Accessing the stored factors (or workspace) arrays would be difficult - it would require me to create a ctypes struct that mimics the memory layout of my `MA27_struct` and to access e.g. the `FACT` array. For the first numeric factorization, I check that two of the `INFO` parameters are what I expect.

The next two factorizations are to make sure I don't have any memory issues when performing numeric factorization then numeric factorization, or numeric factorization then symbolic factorization. I've added checks for the expected number of negative eigenvalues in these matrices, but I'm not sure how worthwhile it would be to try and test beyond the `INFO` arrays.

pyomo/contrib/pynumero/extensions/ma57_interface.py
Robbybp: Done

pyomo/contrib/pynumero/extensions/tests/test_ma27_interface.py
Robbybp: Oh right, no need to be copying ent here. I added that after my first test failed, and FWIW, this test failed too. Changed

pyomo/contrib/pynumero/src/ma27Interface.cpp
```diff
@@ -0,0 +1,285 @@
...
+#  else
+#    define PYNUMERO_HSL_EXPORT __declspec(dllimport)
+#  endif
+#else
+#  define PYNUMERO_HSL_EXPORT
+#endif
+
+// Forward declaration of MA27 fortran routines
+extern "C" {
+   void ma27id_(int* ICNTL, double* CNTL);
```
Robbybp, line 20: If I'm reading the commit history correctly, you removed this `extern "C"` then re-added it. I added this after googling "how to declare external fortran functions in c++." Do you know the real reason this should or shouldn't be necessary?

pyomo/contrib/pynumero/src/ma27Interface.cpp
```diff
@@ -0,0 +1,285 @@
...
+      if (ma27->IW_a == NULL) { abort_bad_memory(1); }
+   }
+
+   PYNUMERO_HSL_EXPORT
+   void alloc_iw_b(MA27_struct* ma27, int l) {
+      if ( ma27->IW_b ) {
+         delete[] ma27->IW_b;
+      }
+      ma27->LIW_b = l;
+      ma27->IW_b = new int[l];
```
Robbybp, line 152: Is there a reason to use new/delete rather than malloc/free other than that it is the norm for C++?

pyomo/contrib/pynumero/src/ma27Interface.cpp
```diff
@@ -17,18 +17,18 @@
 
 // Forward declaration of MA27 fortran routines
 void ma27id_(int* ICNTL, double* CNTL);
-void ma27ad_(int *N, int *NZ, int *IRN, int* ICN,
```
Robbybp: What is NFC?

Robbybp:
@jsiirola Changes look great, thanks for making them. Left some questions above.

pyomo/contrib/pynumero/src/ma27Interface.cpp
jsiirola: I didn't think originally that you had to wrap the Fortran functions in "extern C", but of course you do: the names used in those declarations have to match the name mangling used by the fortran compiler (so we need the simpler, C-style mangling so that things work out).

pyomo/contrib/pynumero/src/ma27Interface.cpp
jsiirola: Mostly because it is the C++ standard, but also because it automatically handles the "sizeof()" adjustment for you, so the code is simpler.

pyomo/contrib/pynumero/src/CMakeLists.txt
```diff
@@ -179,6 +191,8 @@ IF( BUILD_MA27 )
   ADD_LIBRARY( pynumero_MA27 SHARED ${PYNUMERO_MA27_SOURCES} )
   IF( MA27_OBJECT )
     TARGET_LINK_LIBRARIES( pynumero_MA27 ${MA27_OBJECT} )
+  ELSEIF( HSL_LIBRARY AND NOT USE_MA27_LIBRARY )
+    TARGET_LINK_LIBRARIES( pynumero_MA27 ${HSL_LIBRARY} )
   ELSE()
     TARGET_LINK_LIBRARIES( pynumero_MA27 ${MA27_LIBRARY} )
   ENDIF()
```
jsiirola: I think I would recommend getting rid of the `USE_MA??_LIBRARY` option completely and just:
```cmake
  IF( MA27_OBJECT )
    TARGET_LINK_LIBRARIES( pynumero_MA27 ${MA27_OBJECT} )
  ELSEIF( MA27_LIBRARY )
    TARGET_LINK_LIBRARIES( pynumero_MA27 ${MA27_LIBRARY} )
  ELSE()
    TARGET_LINK_LIBRARIES( pynumero_MA27 ${HSL_LIBRARY} )
  ENDIF()
```

Robbybp: I'd prefer to have a way to specify which to use in the case that both libraries are present.

jsiirola: Then we should go back to the original implementation: in the original implementation, the user can overwrite MA27_LIBRARY to point to whatever they want: the way cmake's FIND_LIBRARY works, the automated search only happens if the incoming cache variable is _NotFound.

Robbybp: Okay, thanks for the reminder, that functionality slipped my mind. Reverted.

pyomo/contrib/pynumero/extensions/tests/test_ma27_interface.py
michaelbynum: I agree with @Robbybp. The real test of the factorization comes from testing the back solve. If the factorization is not correct, the backsolve will not be correct.

pyomo/contrib/pynumero/src/CMakeLists.txt
```diff
@@ -139,27 +168,32 @@ IF( BUILD_ASL )
...
 
 IF( BUILD_MA27 )
   ADD_LIBRARY( pynumero_MA27 SHARED ${PYNUMERO_MA27_SOURCES} )
-  TARGET_LINK_LIBRARIES( pynumero_MA27 ${HSL_LIBRARY} )
+  IF( MA27_OBJECT )
+    TARGET_LINK_LIBRARIES( pynumero_MA27 ${MA27_OBJECT} )
+  ELSE()
+    TARGET_LINK_LIBRARIES( pynumero_MA27 ${MA27_LIBRARY} )
+  ENDIF()
   SET_TARGET_PROPERTIES( pynumero_MA27 PROPERTIES ENABLE_EXPORTS 1 )
```
michaelbynum, line 185: @jsiirola Do we need to add the following here (similar to what is done for ASL above)? If I am correct, let me know, and I'll add it in.

```
  if ( DL_LIBRARY )
    TARGET_LINK_LIBRARIES( pynumero_MA27 ${DL_LIBRARY} )
  ENDIF()
  TARGET_COMPILE_DEFINITIONS( pynumero_MA27 PRIVATE BUILDING_PYNUMERO_MA27 )
```

pyomo/contrib/pynumero/src/ma27Interface.cpp
```diff
@@ -0,0 +1,285 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdbool.h>
+#include <string.h>
+
+// This would normally be in a header file, but as we do not need one,
+// we will explicitly include it here.
+#if defined(_WIN32) || defined(_WIN64)
+#  if defined(BUILDING_PYNUMERO_ASL)
```
michaelbynum, line 9: I think this should be `BUILDING_PYNUMERO_MA27`, right?

pyomo/contrib/pynumero/src/ma57Interface.cpp
```diff
@@ -0,0 +1,411 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <stdbool.h>
+
+// This would normally be in a header file, but as we do not need one,
+// we will explicitly include it here.
+#if defined(_WIN32) || defined(_WIN64)
+#  if defined(BUILDING_PYNUMERO_ASL)
```
michaelbynum, line 8: And `BUILDING_PYNUMERO_MA57`?

pyomo/contrib/pynumero/src/ma27Interface.cpp
jsiirola: Yes.  Done.

pyomo/contrib/pynumero/src/ma57Interface.cpp
jsiirola: Yes.  Done.

pyomo/contrib/pynumero/src/ma27Interface.cpp
michaelbynum: Thanks!

pyomo/contrib/pynumero/src/ma57Interface.cpp
michaelbynum: Thanks!

pyomo/contrib/pynumero/src/CMakeLists.txt
jsiirola: Good question.  I don't think it will hurt, but I am a bit surprised that tests were passing.  Maybe it was because they were being skipped everywhere except Jenkins.  I will add it and see if tests clear up.

jsiirola:
Assuming tests pass, I think this is ready to merge.

@Robbybp: in the future, would it make sense to add a new (python) method to the MA27/57 interfaces that returns a `namedtuple` for the INFO data structure?  That would help new users get information back from the libraries (and make the client code more "self-documenting").  So, a user could do:

```python
info = lin_solver.get_info()
print(info.num_neg_eigenvalues)
```
instead of needing to know:
```python
if isinstance(lin_solver, MA27Interface):
    print(lin_solver.get_info(15))
else:
    print(lin_solver.get_info(24))
```

michaelbynum:
I agree. It looks like it is in pretty good shape.

Robbybp:
> @Robbybp: in the future, would it make sense to add a new (python) method to the MA27/57 interfaces that returns a `namedtuple` for the INFO data structure?

It could be done, but I'm not sure it should be. I like the idea of having tags or docstrings for each info/cntl option as a reminder of what they do, but I don't think we should invest too much effort into standardizing an interface for the linear solvers at this level. A couple complications:
- `INFO`, etc. could be called something else in different linear solvers. In MUMPS for instance, the corresponding method would be `get_infog` (which should not be conflated with `get_info`).
- We would need to assign a name to every piece of solver info. This could be dangerous as similar entries could have different behavior in different solvers, and different solvers will vary greatly in the info they report. 

I think whether to do this comes down to where we want to standardize the linear solver interfaces. My current philosophy is that the interfaces should be standardized around a given application, as in the interior point method, where we have additional interfaces that implement the `get_num_neg_eigenvalues` method.

That said, I think supporting a `namedtuple` interface for some (relatively) unambiguous subset of the cntl/info arrays would okay, supporting, for instance:
```python
info = linear_solver.get_info()
if hasattr(info, 'num_neg_eigenvalues'):
    # probably implement with dict + enum
    num_neg = info.num_neg_eigenvalues
else:
    raise NotImplementedError()

PR #1421, opened 2020-05-06
Rewrite of the GDP Chull transformation

emma58:
## Summary/Motivation:

This rewrites chull to fix many of the bugs addressed in the recent rewrite of bigm in addition to many of its own (mainly around local variables). It also sorts out tests which are common to bigm and chull so that everywhere we expect common behavior, we run the same test on both transformations.

- Puts the transformation block on the parent block of the Disjunction rather than on the model (so that solving a Block after transformation is supported)
- Adds an option to include a LocalVars Suffix on the model which maps Disjuncts to lists of variables which are local to that Disjunct (used *nowhere* else on the model). This fixes bugs with wrongly assuming variables are local based on where they are declared, and enables us to not double-disaggregate variables when transforming nested disjunctions.
- Redoes internal mapping of transformed and original components, and adds methods which allow to querying these maps. Notably, in both chull and bigm, changes `get_transformed_constraint` to `get_transformed_constraints` and has it always return a list of _DisjunctDatas (since one original constraint often corresponds to multiple transformed constraints).
- Reorganizes tests for bigm and chull
- Fixes bug in bigm where we did not pick up BigM Suffix on subblock of a Disjunct

### Legal Acknowledgement


qtothec:
Impressive. Did you also want me to have a look over this, @emma58?

jsiirola:
@qtothec: that would be awesome.  Thanks!

pyomo/gdp/plugins/chull.py
```diff
@@ -210,231 +233,205 @@ def _apply_to_impl(self, instance, **kwds):
...
                 else:
-                    self._transformDisjunctionData(t, transBlock, t.index())
+                    self._transform_disjunctionData(t, t.index())
             elif t.ctype in (Block, Disjunct):
                 if t.parent_component() is t:
-                    self._transformBlock(t, transBlock)
+                    self._transform_block(t)
                 else:
-                    self._transformBlockData(t, transBlock)
+                    self._transform_blockData(t)
```
jsiirola, line 243: An alternative test to this is:
```python
elif t.ctype is Disjunction:
    if t.is_indexed():
        self._transform_disjunction(t)
    else:
        self._transform_disjunctionData(t, t.index())
elif t.ctype in (Block, Disjunct):
    if t.is_indexed():
        self._transform_block(t)
    else:
        self._transform_blockData(t)
```
...except that SimpleDisjunctions will now be processed by `_transform_disjunctionData` instead of `_transform_disjunction` (and the same for SimpleBlock and SimpleDisjunct)

pyomo/gdp/plugins/chull.py
```diff
@@ -470,22 +468,21 @@ def _transformDisjunctionData(self, obj, transBlock, index):
             disaggregationConstraint.add(
                 consIdx,
                 var == disaggregatedExpr)
```
jsiirola, line 470: I think this should be changed.  Now that the new Sets are in, you can declare the disaggregationConstraint as:
```python
Constraint(NonNegativeIntegers, Any)
```
Then, this while mess (including the computation of the `consIdx` can be simplified to:
```python
disaggregationConstraint.add((i, index), var==disaggregatedExpr)
```

pyomo/gdp/plugins/chull.py
```diff
@@ -210,231 +233,205 @@ def _apply_to_impl(self, instance, **kwds):
...
+        transBlockName = unique_component_name(
+            instance,
+            '_pyomo_gdp_chull_relaxation')
+        transBlock = Block()
+        instance.add_component(transBlockName, transBlock)
+        transBlock.relaxedDisjuncts = Block(Any)
+        transBlock.lbub = Set(initialize = ['lb','ub','eq'])
+        # We will store all of the disaggregation constraints for any
+        # Disjunctions we transform onto this block here.
+        transBlock.disaggregationConstraints = Constraint(Any)
```
jsiirola, line 273: Now that we can, I think these should be declared as `Constraint(NonNegativeIntegers, Any)`

pyomo/gdp/plugins/chull.py
```diff
@@ -210,231 +233,205 @@ def _apply_to_impl(self, instance, **kwds):
...
 
+    def _add_transformation_block(self, instance):
+        # make a transformation block on instance where we will store
+        # transformed components
+        transBlockName = unique_component_name(
+            instance,
+            '_pyomo_gdp_chull_relaxation')
+        transBlock = Block()
+        instance.add_component(transBlockName, transBlock)
+        transBlock.relaxedDisjuncts = Block(Any)
```
jsiirola, line 269: Now that we can, I think these should be declared as `Block(NonNegativeIntegers)`

pyomo/gdp/plugins/chull.py
```diff
@@ -497,42 +494,59 @@ def _transform_disjunct(self, obj, transBlock, varSet, localVars):
...
-                    "appear to have been relaxed. This makes no sense."
+                    "appear to have been relaxed. This makes no sense. "
+                    "(If the intent is to deactivate the disjunct, fix its "
+                    "indicator_var to 0.)"
                     % ( obj.name, ))
 
-        if 'chull' in infodict:
-            # we've transformed it (with CHull), so don't do it again.
-            return
+        if not obj._transformation_block is None:
```
jsiirola, line 506: PEP8: "`is not None`"

pyomo/gdp/plugins/chull.py
```diff
@@ -723,6 +706,8 @@ def _xform_constraint(self, obj, disjunct, infodict, var_substitute_map,
         if obj.is_indexed():
             try:
                 newConstraint = Constraint(obj.index_set(), transBlock.lbub)
+            # ESJ: TODO: John, is this except block still reachable in the
+            # post-set-rewrite universe? I can't figure out how to test it...
```
jsiirola, line 710: No.  In fact, I bet with a little work, we could get rid of the `if obj.is_indexed()`, too...

pyomo/gdp/plugins/chull.py
```diff
@@ -840,5 +853,131 @@ def _xform_constraint(self, obj, disjunct, infodict, var_substitute_map,
...
+        if disjunct._transformation_block is None:
+            raise GDP_Error("Disjunct %s has not been transformed" 
+                            % disjunct.name)
+        transBlock = disjunct._transformation_block()
+        try:
+            return transBlock._disaggregatedVarMap['disaggregatedVar'][v]
+        except:
+            raise GDP_Error("It does not appear %s is a "
+                            "variable which appears in disjunct %s"
+                            % (v.name, disjunct.name))
```
jsiirola, line 911: This is dangerous as it can hide underlying bugs.  Either catch a specific exception, or (probably better) log this message as an error and re-raise the original exeception.

pyomo/gdp/plugins/chull.py
```diff
@@ -840,5 +853,131 @@ def _xform_constraint(self, obj, disjunct, infodict, var_substitute_map,
...
+                           transformation as a disaggregated variable 
+                           (and so appears on a transformation block 
+                           of some Disjunct)
+        """
+        transBlock = disaggregated_var.parent_block()
+        try:
+            src_disjunct = transBlock._srcDisjunct()
+        except:
+            raise GDP_Error("%s does not appear to be a disaggregated variable"
+                            % disaggregated_var.name)
```
jsiirola, line 930: This is dangerous as it can hide underlying bugs.  Either catch a specific exception, or (probably better) log this message as an error and re-raise the original exeception.

pyomo/gdp/plugins/chull.py
```diff
@@ -840,5 +853,131 @@ def _xform_constraint(self, obj, disjunct, infodict, var_substitute_map,
...
+            raise GDP_Error("Disjunction %s has not been properly transformed: "
+                            "None of its disjuncts are transformed." 
+                            % disjunction.name)
+        try:
+            return transBlock().parent_block()._disaggregationConstraintMap[
+                original_var][disjunction]
+        except:
+            raise GDP_Error("It doesn't appear that %s is a variable that was "
+                            "disaggregated by Disjunction %s" % 
+                            (original_var.name, disjunction.name))
```
jsiirola, line 961: This is dangerous as it can hide underlying bugs.  Either catch a specific exception, or (probably better) log this message as an error and re-raise the original exeception.

pyomo/gdp/plugins/chull.py
```diff
@@ -840,5 +853,131 @@ def _xform_constraint(self, obj, disjunct, infodict, var_substitute_map,
...
+           block of some Disjunct)
+        """
+        # This can only go well if v is a disaggregated var
+        transBlock = v.parent_block()
+        try:
+            return transBlock._bigMConstraintMap[v]
+        except:
+            raise GDP_Error("Either %s is not a disaggregated variable, or "
+                            "the disjunction that disaggregates it has not "
+                            "been properly transformed." % v.name)
```
jsiirola, line 983: This is dangerous as it can hide underlying bugs.  Either catch a specific exception, or (probably better) log this message as an error and re-raise the original exeception.

pyomo/gdp/plugins/bigm.py
```diff
@@ -147,11 +149,21 @@ def _get_bigm_suffix_list(self, block):
         # SimpleBlocks. Though it is possible at this point to stick them
         # on whatever components you want, we won't pick them up.
         suffix_list = []
+        # first descend into the subblocks here to see if anything is there
+        for b in block.component_data_objects(Block, descend_into=(Block),
+                                              active=True,
+                                              sort=SortComponents.deterministic):
+            bigm = b.component('BigM')
+            if type(bigm) is Suffix:
+                suffix_list.append(bigm)
```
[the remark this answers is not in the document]
emma58: Oh, oops, you're definitely right. But we still have a problem, which is that
```
m.b = Block()
m.b.d = Disjunct()
m.b.d.foo = Block()

m.b.BigM = Suffix()
m.b.BigM[None] = 10
m.b.d.foo.c = Constraint(...)
m.b.d.foo.BigM = Suffix()
m.b.d.foo.BigM[None] = 1
```
doesn't work with the original implementation, since `_get_bigm_suffix_list` only gets called on Disjuncts (we never looked on `foo`). But that's the real bug: We should be calling it on the parent block of every constraint...

pyomo/gdp/plugins/chull.py
emma58: Oh, nice, thank you! There's no problem with that change, it makes the code make a lot more sense, actually.

pyomo/gdp/plugins/chull.py
emma58: Ummm, for reasons I no longer remember, we index these with the original constraint index followed by an integer... So I guess this could be `Constraint(Any, NonNegativeIntegers)`, but does that provide any validation actually? (Or we could change the order and make what you said true.)

jsiirola: With the set rewrite, both will work and provide the same validation (one that the index starts with a nonnegative integer, and the other that it ends with one).  I tend to like to have the non-dimensioned Set be at the end, though (things print nicer).

pyomo/gdp/plugins/chull.py
```diff
@@ -210,231 +233,205 @@ def _apply_to_impl(self, instance, **kwds):
...
+            for cons in disjunct.component_data_objects(
+                    Constraint,
+                    active = True,
+                    sort=SortComponents.deterministic,
+                    descend_into=Block):
+                # [ESJ 02/14/2020] We *will* disaggregate fixed variables on the
+                # philosophy that fixing is not a promise for the future and we
+                # are mathematically wrong if we don't transform these correctly
+                # and someone later unfixes them and keeps playing with their
+                # transformed model
```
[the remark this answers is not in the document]
emma58: Oh... Yikes. The answer *is* no (even in the linear case). And I'm of two minds about it: Because if you are striving for the tightest-possible M values, then it would be very annoying if fixed variables were treated as unfixed. But it's definitely counter to the above philosophy. Perhaps the user should have a choice about how to treat fixed variables in the transformation?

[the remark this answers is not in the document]
jsiirola: Having an option for the transformation makes a lot of sense.

pyomo/contrib/fme/tests/test_fourier_motzkin_elimination.py
```diff
@@ -20,8 +20,14 @@
 from pyomo.gdp import Disjunction, Disjunct
 from pyomo.repn.standard_repn import generate_standard_repn
 from pyomo.core.kernel.component_set import ComponentSet
+import random
 
 class TestFourierMotzkinElimination(unittest.TestCase):
```
qtothec, line 25: Is this here intentionally, Emma? I think FME is in your other PR?

pyomo/gdp/plugins/bigm.py
```diff
@@ -144,16 +146,23 @@ def __init__(self):
...
         # Note that you can only specify suffixes on BlockData objects or
         # SimpleBlocks. Though it is possible at this point to stick them
         # on whatever components you want, we won't pick them up.
         suffix_list = []
-        while block is not None:
+        orig_block = block
+
+        # go searching above block in the tree, stop when we hit stopping_block
+        # (This is so that we can search on each Disjunct once, but get any
+        # information between a cosntraint and its Disjunct while transforming
```
qtothec, line 158: typo: constraint

pyomo/gdp/plugins/bigm.py
```diff
@@ -265,7 +274,7 @@ def _add_transformation_block(self, instance):
             '_pyomo_gdp_bigm_relaxation')
         transBlock = Block()
         instance.add_component(transBlockName, transBlock)
-        transBlock.relaxedDisjuncts = Block(Any)
+        transBlock.relaxedDisjuncts = Block(NonNegativeIntegers)
```
qtothec, line 277: Is this the new way to accomplished indexed blocks? Looks like an improvement over `Any`, but in isolation I had to blink a couple times to understand what it's accomplishing.

pyomo/gdp/plugins/bigm.py
```diff
@@ -515,47 +517,11 @@ def _transfer_transBlock_data(self, fromBlock, toBlock):
...
-        # the disjunction should only have been active if it wasn't transformed
-        assert problemdisj.algebraic_constraint is None
-        _probDisjName = problemdisj.getname(
-            fully_qualified=True, name_buffer=NAME_BUFFER)
-        raise GDP_Error("Found untransformed disjunction %s in disjunct %s! "
-                        "The disjunction must be transformed before the "
-                        "disjunct. If you are using targets, put the "
-                        "disjunction before the disjunct in the list."
-                        % (_probDisjName, disjunct.name))
+        _warn_for_active_disjunction(disjunction, disjunct, NAME_BUFFER)
```
qtothec, line 520: It looks like we are changing the function signature, but keeping the old version? Why?

pyomo/gdp/plugins/bigm.py
```diff
@@ -515,47 +517,11 @@ def _transfer_transBlock_data(self, fromBlock, toBlock):
...
-
-        raise GDP_Error("Found active disjunct {0} in disjunct {1}! "
-                        "Either {0} "
-                        "is not in a disjunction or the disjunction it is in "
-                        "has not been transformed. "
-                        "{0} needs to be deactivated "
-                        "or its disjunction transformed before {1} can be "
-                        "transformed.".format(problemdisj.name,
-                                              outerdisjunct.name))
+        _warn_for_active_disjunct(innerdisjunct, outerdisjunct, NAME_BUFFER)
```
qtothec, line 524: It looks like we are changing the function signature, but keeping the old version? Why?

pyomo/gdp/util.py
```diff
@@ -134,3 +136,154 @@ def is_child_of(parent, child, knownBlocks=None):
...
+    transBlock: _BlockData which is in the relaxedDisjuncts IndexedBlock
+                on a transformation block.
+    """
+    try:
+        return transBlock._srcDisjunct()
+    except:
+        raise GDP_Error("Block %s doesn't appear to be a transformation "
+                        "block for a disjunct. No source disjunct found." 
+                        "\n\t(original error: %s)" 
+                        % (transBlock.name, sys.exc_info()[1]))
```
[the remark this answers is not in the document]
qtothec: Yeah, Pycharm will make angry noises at me for bare `except`.

pyomo/gdp/util.py
```diff
@@ -134,3 +136,157 @@ def is_child_of(parent, child, knownBlocks=None):
...
+    # traverse up until we find the disjunct this constraint lives on
+    parent_disjunct = constraint.parent_block()
+    while not isinstance(parent_disjunct, _DisjunctData):
+        if parent_disjunct is None:
+            raise GDP_Error(
+                "Constraint %s is not on a disjunct and so was not "
+                "transformed" % constraint.name)
+        parent_disjunct = parent_disjunct.parent_block()
+
+    return parent_disjunct
```
qtothec, line 212: This is helpful other places, too. I probably have duplicated code in `contrib.gdp_bounds` to accomplish similar functionality.

pyomo/gdp/tests/test_chull.py
```diff
@@ -263,29 +221,14 @@ def test_disaggregatedVar_bounds(self):
...
-        check_linear_coef(self, repn, m.d[0].indicator_var, 1)
-        check_linear_coef(self, repn, m.d[1].indicator_var, 1)
-
     def test_error_for_or(self):
         m = models.makeTwoTermDisj_Nonlinear()
         m.disjunction.xor = False
 
         self.assertRaisesRegexp(
             GDP_Error,
             "Cannot do convex hull transformation for disjunction disjunction "
```
qtothec, line 230: Heh. We must've missed a typo at some point.

pyomo/gdp/plugins/chull.py
```diff
@@ -171,6 +178,33 @@ def __init__(self):
...
+        localVars = block.component('LocalVars')
+        if type(localVars) is Suffix:
+            for disj, var_list in iteritems(localVars):
+                if local_var_dict.get(disj) is None:
+                    local_var_dict[disj] = ComponentSet(var_list)
+                else:
+                    local_var_dict[disj].update(var_list)
+
+    def _get_local_var_suffixes(self, block, local_var_dict):
+        # You can specify suffixes on any block (dijuncts included). This method
```
qtothec, line 191: typo: disjunct

pyomo/gdp/plugins/chull.py
```diff
@@ -171,6 +178,33 @@ def __init__(self):
...
+        # starts from a Disjunct (presumably) and checks for a LocalVar suffixes
+        # going up the tree, adding them into the dictionary that is the second
+        # argument.
+
+        # first look beneath where we are (there could be Blocks on this
+        # disjunct)
+        for b in block.component_data_objects(Block, descend_into=(Block),
+                                              active=True,
+                                              sort=SortComponents.deterministic):
+            self._add_local_vars(b, local_var_dict)
```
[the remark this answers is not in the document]
qtothec: Is there a possibility for suffix settings to conflict here? If so, maybe search upwards and then down, so that the "closest"  value wins?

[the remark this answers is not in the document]
emma58: @qtothec, no, there can't be a conflict because you can't declare a variable as *not* local using this Suffix. So the order doesn't matter: If we find the variable anywhere, it's local. Otherwise, no.

pyomo/contrib/fme/tests/test_fourier_motzkin_elimination.py
emma58: Yes, but only because I broke the FME tests with changes to the transformation block.

pyomo/gdp/plugins/bigm.py
jsiirola: It's not a new way to accomplish indexed blocks: Blocks are still indexable by any Set.  However, since the set-rewrite, all the global Sets (like NonNegativeIntegers) are proper Sets and can be used as indexing sets for components (like Block).  By making the index more restrictive than "Any", @emma58 is making a stronger statement about what a user should expect as a valid index.

pyomo/gdp/plugins/bigm.py
emma58: I'm not quite sure I understand your question. But basically, because this function is a handler, its arguments are determined by that (even though we don't actually need any of them). It does the same thing as the corresponding function in chull, so I moved the logic to util, but I can't call that directly because the chull handlers have different arguments than the bigm ones.

emma58:
OK, @jsiirola, @qtothec, I think I've addressed everything, if you'd be willing to have another look when you get a chance. (Though the FME tests are going to keep failing on this branch until #1436 is in.)

pyomo/gdp/plugins/chull.py
```diff
@@ -790,6 +775,9 @@ def _xform_constraint(self, obj, disjunct, infodict, var_substitute_map,
 
             if c.equality:
                 if NL:
+                    # ESJ TODO: This can't happen right? This is the only
+                    # obvious case where someone has messed up, but this has to
+                    # be nonconvex, right? Shouldn't we tell them?
                     newConsExpr = expr == c.lower*y
```
emma58, line 781: @jsiirola, one question: Shouldn't we throw an error here since it is definitely non-convex? Or am I crazy?

pyomo/gdp/plugins/bigm.py
```diff
@@ -792,113 +799,26 @@ def _estimate_M(self, expr, name):
...
-        # that map.
-        m = xor_constraint.model()
-        for disjunction in m.component_data_objects(Disjunction):
-            if disjunction._algebraic_constraint:
-                if disjunction._algebraic_constraint() is xor_constraint:
-                    return disjunction
-        raise GDP_Error("It appears that %s is not an XOR or OR constraint "
-                        "resulting from transforming a Disjunction."
-                        % xor_constraint.name)
+        return get_src_disjunction(xor_constraint)
```
qtothec, line 821: I agree with the code reduction that comes with aggregating everything into `gdp.util`, but I am a bit sad that we are losing the nice docstrings. I think there ought to be a way to do something similar to `functool.wraps` in this case.

Maybe something to consider after merge. I wouldn't hold this up on that account.

pyomo/gdp/plugins/chull.py
qtothec: You are correct that it would be non-convex, but I think the Hull Reformulation described by doing the following is still a valid reformulation of the original problem. As such, `gdp.chull` has always been somewhat false advertising.

@bernalde You want to double-check this?

pyomo/gdp/tests/test_chull.py
qtothec: Should 'disjunction' be repeated here?

pyomo/gdp/plugins/bigm.py
jsiirola: Fixed.  I made use of `@functools.wraps` so that the transformation methods inherif the docstring from the implementations in utils.

jsiirola:
@qtothec: The duplicated 'disjunction disjunction' was actually correct.  The Disjunction component in question was actnally named 'disjunction'.  I went through chull and bigm and put component names in quotes to hopefully make this more clear in the future,

I think we are good to go and will merge as soon as this most recent round of tests pass.

qtothec:
@jsiirola Does the `functool.wraps` actually work as intended here? I was saying that more as something to look into later on, but I see that it's been incorporated.

jsiirola:
@qtothec, if by "work as intended" you mean that it picks up the docstring, then yes (at least for my quick test):
```python
>>> t = TransformationFactory('gdp.chull')
>>> help(t)
...
 |  get_src_constraint(self, transformedConstraint)
 |      Return the original Constraint whose transformed counterpart is
 |      transformedConstraint
 |      
 |      Parameters
 |      ----------
 |      transformedConstraint: Constraint, which must be a component on one of
 |      the BlockDatas in the relaxedDisjuncts Block of
 |      a transformation block
 |  
 |  get_src_disjunct(self, transBlock)
 |      Return the Disjunct object whose transformed components are on
 |      transBlock.
 |      
 |      Parameters
 |      ----------
 |      transBlock: _BlockData which is in the relaxedDisjuncts IndexedBlock
 |                  on a transformation block.
```

PR #1422, opened 2020-05-06
GDPOpt fix use of numeric derivatives

jsiirola:
## Summary/Motivation:
GDPOpt currently caches symbolic derivatives for small constraints between iterations, making iterative cut generation more efficient.  However, if the derivatives are obtained numerically, the cached derivative values are incorrect.  This PR clears out numeric derivative values after the cut is generated.  It also adds a fallback option: if generation of symbolic derivatives fails, then it will automatically fall back on the generation of numeric derivatives (this is in preparation for the support of external functions in the numeric differentiation code).

Tests of the fallback routine will be added once efternal functions are supported by the numeric differentiation code.

- clear (do not cache) numeric derivative values after the cut is generated
- add fallback to numeric derivatives if symbolic derivative generation fails.

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/gdpopt/cut_generation.py
+++ b/pyomo/contrib/gdpopt/cut_generation.py
@@ -3,6 +3,7 @@ from __future__ import division
 
 from collections import namedtuple
 from math import copysign, fabs
+from six import iteritems
 from pyomo.contrib.gdp_bounds.info import disjunctive_bounds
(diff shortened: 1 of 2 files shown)


bernalde:
Should we use something along the same lines for MindtPy?

qtothec:
@bernalde I don't see why not. Mindtpy differentiation was built on the same principle.

PR #1426, opened 2020-05-08
Unified GitHub Actions workflow

jsiirola:
## Summary/Motivation:
Keeping 4 different GitHub Actions workflow drivers in sync was becomming problemmatic.  This PR introduces a single "unified" workflow that manages Linux, OSX, and Windows tests with a minimum amount of duplicated code.  There are now only two workflows, and they ONLY differ by the "on" clause (one for PRs / master, one for all other branches), and a restricted set of Python versions for the branch tests.

The only downside of a single workflow is that GitHub only supports re-running the *entire* workflow, so in the event that a one job exhibits a transient failure, the only recourse is to rerun *all* of the jobs.  As long as the transient failures are infrequent, that shouldn't be a huge problem.  If it becomes a problem in the future, we can duplicate the workflow and break apart the matrix definition to enable rerunning smaller subsets of the jobs.

- Introduce a single "unified" workflow that defines *all* jobs
- (unrelated) Catch an additional exception in the Kestrel interface to make the NEOS tests more reliable.
- (unrelated) Add a timeout to `pyomo help --solvers` so NEOS doesn't cause builds to hang

### Legal Acknowledgement


mrmundt:
👏 👏 👏 
That's all I have to say.

Just kidding - I have a few more things. 
1. Did you notice that the number of skipped tests is pretty small? In comparison, I've been messing around on a branch on my fork and the push test (Linux, Py3.7) has ~1500 skipped tests. Equivalent here: Only 1088. WOW.
2. The coverage in solvers is amazing. It's better than Appveyor even! 
3. Speaking of Appveyor. 💤  Let's put it to rest.

.github/workflows/pr_master_test.yml
```diff
@@ -0,0 +1,397 @@
...
+    branches:
+      - master
+
+defaults:
+  run:
+    shell: bash -l {0}
+
+env:
+  PYTHONWARNINGS: ignore::UserWarning
+  PYTHON_BASE_PKGS: >
```
mrmundt, line 17: A thought - we should add `wheel` to the base packages. I noticed yesterday that we get a lot of these messages: 
```
Could not build wheels for pyodbc, since package 'wheel' is not installed.
Could not build wheels for pytz, since package 'wheel' is not installed.
```
It doesn't stop the package from building. That being said, I added `wheel` to my local branch, and it makes the setup somewhat faster. It's not a huge speed-up in test time, but I did notice a difference in when it's able to build and install from `wheel` as opposed to from source, and we might as well be efficient, right?

jsiirola:
@mrmundt: Thanks!  Re:
1. Yes, but...  Some of those tests are entire Suites being skipped.  When you skip an entire suite (either by calling `unittest.SkipTest()` at the module scope or by decorating the class), unittest only records that as a single skipped test, even though it could have skipped 10's or 100's of tests.
2. Yup - the original Appveyor tests were actually pretty weak.  This is an improvement there.
3. Yes.  Once this is merged.
4. So this PR does update the branch push test.  It turns out that there is no real problem pushing branch coverage to codecov, so I thought I would leave it in (it can be nice for developers).  As fot only exercising a sumset of tests, I'd rather not: a developer can always do that locally.  By running the "full" test suite, a developer can get a pretty good feel if their branch will pass PR testing.  If anything, I might *add* tests.  Right now, we run py3.7 on all three platforms.  It might be more useful to run 2.7, 3.5, and 3.8; one on each platform (python differences tend to cause more problems than platform differences).

.github/workflows/pr_master_test.yml
jsiirola: Seems reasonable.  Done.

mrmundt:
I do have one more observation. In a previous PR (#1277 ), we had removed Unix Python 2.7 testing due to instability, but it looks like you've re-introduced the Python 2.7 testing. 

No change is needed - just something to keep in mind.

jsiirola:
@mrmundt: correct.  We had pulled it from linux/osx but left it in for Windows because of weirdness in pip.  I checked, and things looked stable, so I put it back.  If we have problems again, we can add it to the `exclude` list in the future.

.github/workflows/pr_master_test.yml
```diff
@@ -0,0 +1,401 @@
...
+    # invoke bash as a login shell ('shell: bash -l {0}') so that the
+    # conda environment is properly activated.  However, running within
+    # a login shell appears to foul up the link to python from
+    # setup-python.  Further, we have anecdotal evidence that
+    # subprocesses invoked through $(python -c ...) and `python -c ...`
+    # will not pick up the python activated by setup-python on OSX.
+    #
+    # Our solution is to define a PYTHON_EXE environment variable that
+    # can be explicitly called within subprocess calls to reach the
+    # correct interpreter.  Note that we must explicitly run nin a *non*
```
blnicho, line 149: Typo: in

.github/workflows/push_branch_test.yml
```diff
@@ -1,114 +1,400 @@
...
+    # invoke bash as a login shell ('shell: bash -l {0}') so that the
+    # conda environment is properly activated.  However, running within
+    # a login shell appears to foul up the link to python from
+    # setup-python.  Further, we have anecdotal evidence that
+    # subprocesses invoked through $(python -c ...) and `python -c ...`
+    # will not pick up the python activated by setup-python on OSX.
+    #
+    # Our solution is to define a PYTHON_EXE environment variable that
+    # can be explicitly called within subprocess calls to reach the
+    # correct interpreter.  Note that we must explicitly run nin a *non*
```
blnicho, line 148: Typo: in

blnicho:
@jsiirola can we go ahead and merge this? Travis finished successfully (even though the results aren't being published for some reason) and we'll be disabling appveyor soon after merging this

jsiirola:
I was waiting for Jenkins to get around to it to test the changes to the NEOS interface (there is a backlog due to a configuration change).  Jenkins should catch up in 2-3 hours.

That said, those changes have been successful in other environments and are relatively minor, so I am OK merging.

PR #1428, opened 2020-05-08
Fix handling of deactivated blocks in relax_integer_vars transformation

emma58:
## Summary/Motivation:

The set rewrite merge changed the behavior of `core.relax_integer_vars` to not descend into deactivated components when searching for Vars to relax. This reverts the default behavior to back to descending into deactivated components and adds an argument to the transformation to tell it which behavior you want.

- Adds `descend_into_deactivated_components` argument to relax_integer_vars transformation, with default value of True (which means actually that we use `active=None` in the call to `component_data_objects`)
- Adds a couple tests to check the default behavior and the behavior with the argument.

### Legal Acknowledgement


emma58:
A couple notes:
1) I hate the name of the argument, if anyone has suggestions...
2) This doesn't allow for someone to specify `active=False` to the `component_data_objects` call that controls which variables are transformed. If we think there is a need for that, then we should have an `active` argument to the transformation and pass that value through, I think.

qtothec:
I recall having a long conversation about whether it ever makes sense to have `active=False` on one of the component iterators. I can't think of a sane use case for it.

I think the argument is fine. You're saying that we unconsciously changed default behavior for the transformation? :fearful:

pyomo/core/plugins/transform/discrete_vars.py
```diff
@@ -44,11 +44,16 @@ def _apply_to(self, model, **kwds):
                 v.setub(bounds[1])
             model.del_component("_relaxed_integer_vars")
             return
+        # True by default, you can specify False if you want
+        descend = kwds.get('descend_into_deactivated_components', 
+                           options.get('descend_into_deactivated_components', 
+                                       True))
```
jsiirola, line 50: For consistency with the other Pyomo iterators, why not just:
```python
active = kwds.get('active', options.get('active', True))
```

emma58: I thought about that, but I'm worried that in that case giving the argument `active=False` to the transformation will be really misleading (since from the outside someone maybe doesn't know they're just setting the `active` arg in `component_data_objects`). It is consistent with the other iterators, but I think that the intuitive behavior someone would expect from it is actually what they would get from `active=None`. But... That's kind of true for component_data_objects also. So I'm fine with doing it here too for consistency if you think that's better.

jsiirola: Good point.  How about `transform_deactivated_blocks` (after all, only blocks will be descended into)?

emma58: OK, sounds good. :)

Issue #1153, opened 2019-11-05
Performance suggestion: templates


michaelbynum:
Implementing solver interfaces that utilize templates could significantly improve performance (including memory consumption).

whart222:
Can you flush this idea out further?  That will help us critique which ideas we focus on at the hackathon.

blnicho:
I think the idea here is to try "templatizing" indexed constraints by default so that only a single expression tree needs to be constructed and then it could be reused when writing out each individual index.

michaelbynum:
@blnicho Described the basic idea. Let me flush this out a little further. 

First, I should note that there is a range of possibilities here. I will discuss a rather extreme version.

Consider the constraint:

y[i] = a[i] * x[i] + b[i] * v[i]

defined over a large index set. The idea of templates is to generate a single expression tree rather than an expression tree for each constraint. When the model goes to a solver, we can walk the expression tree once and then hand all of the constraints to the solver through whatever mechanism is effective for that solver interface. 

As an example, let's consider the above constraint with the Gurobi direct interface. One way to create constraints in the Gurobi python interface is with a list of coefficients and a list of variables. We can avoid creating pyomo expression trees for all of the constraints and work primarily with native python containers. In this case, we could directly create the following lists:

[1, -a[i], -b[i]]

and 

[y[i], x[i], v[i]]

for each i in the index set. At first, it may seem that this approach has very limited application. However, @jsiirola and @carldlaird have ideas on how to make this very flexible.

jsiirola:
To follow what @michaelbynum said, we can (today) generate template expressions like the example he gave.  We can also get templates for more complex expressions like
```
m.x[i] <= sum(m.y[i,j] for j in model.J[i])
```
where the `sum` is a node in the expression tree and has _not_ been expanded.

I believe that it should be possible to "cannonicalize" template expressions using a slight variant of the visitor-based `generate_standard_repn` function I prototyped a couple months ago.

Doing this should have both significant speed and memory improvements: we don't have to generate _or_ walk all the individual expressions.  There is a little extra work, as we will still have to perform variable aggregation when we expand the templatized standard repn (`m.x[i]` and `m.x[j]` could resolve to the same `_VarData`; `m.x[i]` could be fixed, etc...), but that should be much faster than walking an expression.

[...]

Issue #752, opened 2018-11-18
Automatic generation of model documentation


GiorgioBalestrieri:
(not sure how to flag this as a feature request)

It would be a great, great feature to be able to automatically generate model documentation. I'm thinking of something like [GAMS Model2Tex command](https://www.gams.com/latest/docs/T_MODEL2TEX.html), possibly through a [Sphinx](http://www.sphinx-doc.org/en/master/)-like tool.

Clearly, the fact that Pyomo models can include arbitrary Python code will make quite difficult to make a tool that will always work, but I think using a combination of docstrings, doc fields and user-inputs should work.

I tend to think it would be somewhat easier to generate documentation from a `ConcreteModel` than an `AbstractModel`, but I would be happy to be corrected.

If anyone with a better knowledge of the inner workings of Pyomo would mind elaborating on some possible strategies and challenges in doing this, it would be helpful.

A non fully automated way - but very effective - is to include LaTeX formulations in docstring, and then rely on Sphix to generate the documentation, as done for the (awesome) [Calliope](https://www.callio.pe/) project. Not sure how this would work for Constraints defined through lambda functions or expressions though.
The mathematical formulation is [here](https://calliope.readthedocs.io/en/stable/user/ref_formulation.html), and an example of docstring is [here](https://github.com/calliope-project/calliope/blob/master/calliope/backend/pyomo/constraints/energy_balance.py#L93).

blnicho:
We do have a prototype for something like this in another project built on top of Pyomo. We may be able to generalize that functionality for pure Pyomo models to support what you're asking for. I don't have an estimate for when that might happen though.

GiorgioBalestrieri:
Ok, I gave it a try. It's still quite hacky, it basically reads the `doc` field of all components in a model and builds reStructuredText based on that.

See an example [here](file:///C:/Users/gbalestrieri/mine/pyomo-sphinx-docs/docs/_build/html/index.html)

fleimgruber:
When I gave the thumbs up, I obviously misread the OP. I thought that since Pyomo Params and Vars objects contain their indexing specifics they could be derived automatically and the LaTeX could be built from that in a generic way. Same goes for the constraints using the Pyomo internal expression parser results. I had a chat with @jsiirola during a conference where he mentioned that this could be pull off.

[...]

PR #1433, opened 2020-05-11
Expanded expression template support

jsiirola:
## Fixes #N/A
(but provides the first steps for #752 and #1153)

## Summary/Motivation:
This PR expands the template system to support hierarchical models and embedded sum expressions.  Templates can now be formed for hierarchical models where the IndexTemplate(s) appear in the blocks and not just in Var/Param components.  This also adds support for generating unexpanded templates of expressions containing summations.  For example, considering the following "complex" rule:
```python
m = ConcreteModel()
m.I = RangeSet(3)
m.J = RangeSet(3)
m.K = Set(m.I, initialize={1:[10], 2:[10,20], 3:[10,20,30]})
m.x = Var(m.I,m.J,[10,20,30])
@m.Constraint()
def c(m):
    return sum( sum(m.x[i,j,k] for k in m.K[i])
                for j in m.J for i in m.I ) <= 0
```
The following now works:
```python
>>> template, indices = templatize_constraint(m.c)
>>> print(str(template))
SUM(SUM(x[_2,_1,_3] for _3 in K[_2]) for _1 in J for _2 in I)  <=  0.0
```

To implement this, the PR proposes several core changes:
1. Add a new `PyomoModelingObject` base class.  This slotized interface is a new common base class for `NumericValue` and `_ComponentBase` class hierarchies and defines the fundamental `is_*_type` methods (all of which return False).   By implementing this as a common base class, it makes it easier to support classes that "blur" the distinction between numeric types, component types, and expression types.  This declares the minimal set of methods that Pyomo classes (that is, classes appearing in models / expression trees that are *not* in `native_types`) are expected to implement.  These are:
   - `is_expression_type()`: True if this object is a non-leaf expression node.  If True, the object is expected to implement the `ExpressionBase` API.
   - `is_numeric_type()`: True if the object can represent a numeric value.  If True, then this object is expected to implement the `NumericValue` API.
   - `is_component_type()`: True if this object is a Pyomo component or ComponentData.  If True, the object is expected to implement the `_ComponentBase` API.
   - `is variable_type()` :  Preserved for backwards compatibility.  A future PR will look at the possibility of deprecating / removing this method.
   - `is_parameter_type()`: Preserved for backwards compatibility.  It is currently not used anywhere in the codebase, and will likely be deprecated in a future PR.
   - `is_named_expression_type()`: Preserved for backwards compatibility.  A future PR will look at the possibility of deprecating / removing this method.

2. Expanding the `StreamBasedExpressionVisitor`.  This adds a new `initializeWalker` callback, and expands the API for the `beforeChild`, `acceptChildResult`, and `afterChild` methods to also include the index of the child node in the parent node's args list.  The walker also allows nodes to return args that implement a context manager api (without forcing the walker to define `enterNode` & `exitNode`).

3. Expanded `IndexedComponent_slice` class.  This class is now "public", more robust, and hashable (can be a key in a `dict`)

- Updates to the `StreamBasedExpressionVisitor`
   - pass the index if the child in the node's args list to the `beforeChild`, `acceptChildResult`, and `afterChild` methods
   - Add `initializeWalker` callback 
   - Add support for the context manager API for expression node `args` data
- Updates to `_IndexedComponent_slice`
   - make `_IndexedComponent_slice` a "public" class (rename, dropping the leading `_`)
   - allow `IndexedComponent_slice` objects as `dict` keys
   - update `IndexedComponent_slice` internal data structures to make them less susceptible to unintended side effects by borrowing concepts from SumExpression views.
- Add a new `PyomoModelingObject` base class from which both `NumericValue` and `_ComponentBase` derive.
   - standardize the definitionof the fundamental `is_*_type` methods.
- Flush out the `_UnindexedComponent_set` class with additional methods from the Set API
- Switch the Set API to have derived sets implement `_iter_impl` instead of directly overloading `__iter__`.
- Move all template logic (including definition of expression tree nodes and the resolution walkers) into `pyomo.core.expr.template_expr` (deprecate `pyomo.core.base.template_expr`)
- Add `GetAttrExpression` and `TempateSumExpression` expression nodes
- Add `resolve_template()`, and `templatize_constraint()` methods

### Legal Acknowledgement


qtothec:
@jsiirola I can't seem to track down documentation on how the template expressions work. Do you happen to know where that lives?

jsiirola:
@qtothec, alas there is very little formal documentation written up (in part because things are still evolving).

The short version is a "template" expression is just an extended expression tree; that is normal expression tree that also contains `GetItemExpression` nodes (and with this PR, potentially also `GetAttrExpression` and `TempateSumExpression` nodes).  You trigger template generation by creating `IndexTemplate` objects and using them when retrieving indices (e.g., by passing them to rules).  `IndexTemplate` objects change the behavior of IndexedComponent's `__getitem__` to return `GetItemExpression` nodes instead of returning a "Data" object.  To avoid expanding internal `sum` expressions, you need to leverage the `templatize_rule` method from this PR (it temporarily overrides the `sum()` builtin and Pyomo `Set` iteration so that we can detect and get back a `TemplateSumExpression` node.

pyomo/core/base/param.py
```diff
@@ -188,22 +188,6 @@ def is_parameter_type(self):
         """
         return True
 
-    def is_variable_type(self):
-        """
-        Returns False because this is not a variable object.
-        """
-        return False
-
-    def is_expression_type(self):
```
carldlaird, line 197: I find the name of this method confusing - this is a separate conversation, but now that we have a base class PyomoModelObject, we can have a discussion about the names and intent of these methods.

PR #1436, opened 2020-05-12
Fix component_data_objects for scalar components with no len()

### LINKED ISSUE #1435: component_data_objects fails for cross products with infinite sets
In this example:
```
m = ConcreteModel()
m.s = NonNegativeIntegers*['a', 'b']
for comp in m.component_data_objects():
    print(comp.name)
```
we get this error:
```
  File "infinite_set_bug.py", line 8, in <module>
    for comp in m.component_data_objects():
  File "/home/esjohn/src/pyomo/pyomo/core/base/block.py", line 1415, in component_data_objects
    sort=sort):
  File "/home/esjohn/src/pyomo/pyomo/core/base/block.py", line 1339, in _component_data_iter
    elif len(comp) or not hasattr(comp, '_data'):
  File "/home/esjohn/src/pyomo/pyomo/core/base/set.py", line 2955, in __len__
    "The length of a non-finite Set is Inf; however, Python "
OverflowError: The length of a non-finite Set is Inf; however, Python requires len() to return a non-negative integer value. Check isfinite() before calling len() for possibly infinite Sets
```
The call to `len()` in _component_data_iter is marked as a hack, so I'm not sure if adding `not isfinite(comp)` to that `or` statement is the desired solution... I'm also not sure why this happens for cross products, but not when a component is indexed by `NonNegativeIntegers`, for example?

jsiirola:
## Summary/Motivation:
This fixes an edge case where `component_data_objects()` failed when the model contained scalar components with no `len()` (i.e., non-finite Sets).

- Correct logic when iterating over component datas
- Update tests to include a non-finite Set.

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/tests/unit/test_block.py
+++ b/pyomo/core/tests/unit/test_block.py
@@ -61,6 +61,7 @@ class TestGenerators(unittest.TestCase):
         model = ConcreteModel()
         model.q = Set(initialize=[1,2])
         model.Q = Set(model.q,initialize=[1,2])
+        model.qq = NonNegativeIntegers*model.q
         model.x = Var(initialize=-1)
         model.X = Var(model.q,initialize=-1)
         model.e = Expression(initialize=-1)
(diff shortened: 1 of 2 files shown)


pyomo/core/tests/unit/test_block.py
```diff
@@ -186,7 +187,8 @@ def generator_test(self, ctype):
                 generator = list(block.component_objects(ctype, active=True, descend_into=False))
             except:
                 if issubclass(ctype, Component):
-                    self.fail("component_objects(active=True) failed with ctype %s" % ctype)
+                    print("component_objects(active=True) failed with ctype %s" % ctype)
+                    raise
```
blnicho, line 191: Why did you change these from `self.fail` statements to `print` and `raise`?

jsiirola: The test started failing, but because it was suppressing the exception, I could not diagnose the error.  This way, if the test fails you still get the exception stack trace and can resolve the problem.

PR #1437, opened 2020-05-12
[NFC] RangeSet docstring type hint

qtothec:
## Summary/Motivation:
Pycharm was telling me that I was using RangeSet wrong because of the type hint.

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/set.py
+++ b/pyomo/core/base/set.py
@@ -2527,7 +2527,7 @@ class RangeSet(Component):
 
     Parameters
     ----------
-    *args: tuple | int | None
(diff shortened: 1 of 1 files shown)


qtothec:
Yes, that makes sense. I was about to protest about `float`, but then I remembered that DAE is a thing.

jsiirola:
> Yes, that makes sense. I was about to protest about `float`, but then I remembered that DAE is a thing.

Actually, DAE doesn't use RangeSet.  That turns out to be behavior that dates back to the dawn of Pyomo.

PR #1438, opened 2020-05-12
Ensure block rules are always called

jsiirola:
## Summary/Motivation:
This resolves an issue where Block rules were not being called for blocks created outside of `construct()`.  As the `rule` is the user's "initializer" for a Block, it was surprising (and problematic for DAE) that the rule was not being triggered automatically for Blocks that were created after `construct()` (e.g., when the underlying set was extended, or for Blocks indexed by non-finite Sets).

- Ensure that the Block rule is always called, regardless of when the BlockData was created.
- Update tests to cover this behavior

### Legal Acknowledgement

### DIFF
--- a/pyomo/mpec/complementarity.py
+++ b/pyomo/mpec/complementarity.py
@@ -21,6 +21,9 @@ from pyomo.core.base.plugin import ModelComponentFactory
 from pyomo.core.base.numvalue import ZeroConstant, _sub
 from pyomo.core.base.misc import apply_indexed_rule, tabular_writer
 from pyomo.core.base.block import _BlockData
+from pyomo.core.base.util import (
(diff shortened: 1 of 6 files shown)


pyomo/mpec/complementarity.py
```diff
@@ -319,6 +309,10 @@ def __init__(self, **kwargs):
         args = (Set(),)
         self._nconditions = 0
         Complementarity.__init__(self, *args, **kwargs)
+        # disable the implicit rule; construct will exhause the
```
blnicho, line 312: Typo: exhaust

michaelbynum:
Again, it only looks like these tests failed due to connectivity issues with codecov, so I am going to merge this.

PR #1440, opened 2020-05-14
FME transformation constraint pruning improvements

emma58:
## Summary/Motivation:

Adds a callback to the fourier motzkin elimination transformation which allows for pruning constraints during the projection. Also adds a function for post-processing away redundant constraints.

- Adds additional argument to transformation which is a callback which takes a constraint and returns True or False on whether or not to keep it.
- Adds a default function for the above which eliminates constraints which are already implied by variable bounds
- Adds a post-processing function (to be called separately) which, for each constraint, checks if it is redundant by checking if it can be violated subject to the other constraints. (That's a lot of LP solves.)

### Legal Acknowledgement


pyomo/contrib/fme/fourier_motzkin_elimination.py
```diff
@@ -68,6 +98,20 @@ class Fourier_Motzkin_Elimination_Transformation(Transformation):
...
+    CONFIG.declare('constraint_filtering_callback', ConfigValue(
+        default=_check_var_bounds_filter,
+        domain=constraint_filtering_function,
+        description="Specifies whether or not and how the transformation should"
+        " filter out trivial constraints during the transformation.",
+        doc="""
+        Specify None in order for no constraint filtering to occur during the
+        transformation.
+
+        Specify a function with accepts a constraint (represtned in the >=
```
jsiirola, line 110: Typo: *that* accepts
Typo: represented

pyomo/contrib/fme/fourier_motzkin_elimination.py
```diff
@@ -68,6 +98,20 @@ class Fourier_Motzkin_Elimination_Transformation(Transformation):
         Note that these variables must all be continuous and the model must be 
         linear."""
     ))
+    CONFIG.declare('constraint_filtering_callback', ConfigValue(
+        default=_check_var_bounds_filter,
+        domain=constraint_filtering_function,
+        description="Specifies whether or not and how the transformation should"
+        " filter out trivial constraints during the transformation.",
```
jsiirola, line 105: "a callback that determines whether or not new FME constraints are added to the model."

pyomo/contrib/fme/fourier_motzkin_elimination.py
```diff
@@ -322,3 +370,83 @@ def _add_linear_constraints(self, cons1, cons2):
         ans['lower'] = cons1['lower'] + cons2['lower']
 
         return ans
+
+    def post_process_fme_constraints(self, m, solver_factory):
+        """Function which solves a sequence of LPs problems to check if
```
jsiirola, line 375: function *that*  😄

pyomo/contrib/fme/fourier_motzkin_elimination.py
```diff
@@ -322,3 +370,83 @@ def _add_linear_constraints(self, cons1, cons2):
...
+               TerminationCondition.optimal:
+                raise RuntimeError("Unsuccessful subproblem solve when checking"
+                                   "constraint %s.\n\t"
+                                   "Termination Condition: %s" %
+                                   (constraints[i].name, 
+                                    results.solver.termination_condition))
+            else:
+                obj_val = value(obj)
+            # if we couldn't make it infeasible, it's useless
+            if obj_val >= 0:
```
jsiirola, line 441: Should this have a tolerance?  If there is a feasible solution where it is active, then it could end up with a non-zero dual and be relevant to a downstream algorithm?

pyomo/contrib/fme/fourier_motzkin_elimination.py
```diff
@@ -39,6 +62,13 @@ def vars_to_eliminate_list(x):
             "Expected Var or list of Vars."
             "\n\tRecieved %s" % type(x))
 
+def constraint_filtering_function(f):
+    if f is None:
+        return
+    if not inspect.isfunction(f):
+        raise ValueError("Expected function. \n\tRecieved %s" % type(f))
+    return f
+
```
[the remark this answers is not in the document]
emma58: Ah, probably not... I was very confused about how to do this. I can leave off the domain, and find out later, yes.

pyomo/contrib/fme/fourier_motzkin_elimination.py
emma58: I'm not sure I understand... If this is 0, then it could never be violated. So yes, it could be active, but by keeping this around, we're making that solution degenerate. Would we want that?

jsiirola:
@emma58: I see the problem with the `doc` for the transformation registration: you are calling register twice and the last call was the one that was missing the `doc=` field.  Long term, we should probably put a warning in the PyUtilib Factory infrastructure for duplicate registrations).  In the near term, I would recommend removing the registrations.  Also, as a matter of managing import times, I would recommend:
1. remove the import from `__init__.py`
2. either:
  a. only call `register()` in `plugins.join()`
  b. only call `register()` as a decorator on the class, and put the `import pyomo.contrib.fme.fourier_motzkin_elimination` in the `plugins.load()` function.

...the idea is that to speed things up, Pyomo should not import everything in the codebase just because someone imports `pyomo`.  The exception is that importing `pyomo.environ` should ensure that all plugins have been registered.  Importing `pyomo.envron` eventually calls `load()` in all the `plugins` namespaces (`pyomo.*.plugins` and `pyomo.contrib.*.plugins`), which is what should ensure that the plugins are registeres.

emma58:
OK, @jsiirola, I think this is ready for you to have another look. :)

PR #1446, opened 2020-05-18
Add GDX interface for returning solution values from GAMS

renkekuhlmann:
## Summary/Motivation:
Using put files to read GAMS solutions and solution status is of course feasible, but not the preferred way. This PR suggests to read gdx files instead. A benchmark comparing the different Pyomo/GAMS interfaces (gams_direct, gams_shell and gams_shell_new (this PR)) on the minlplib shows that this PR improves efficiency of the Pyomo/GAMS interface:
stat_Efficiencyprofilerel006
stat_Efficiencygeommean001
(Note that the filter 'no fail by all solver' in the images is irrelevant here, because the solution status is the same for all three runs.)

- use GAMS GDX to read solution
- use further GAMS options to improve efficiency
- include `ws.working_directory` in problem name in gams_direct interface to make it consistent with gams_shell interface.
- use `gams audit` to query GAMS version

### Legal Acknowledgement


pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -331,7 +331,7 @@ def solve(self, *args, **kwds):
         extract_rc = ('rc' in model_suffixes)
 
         results = SolverResults()
-        results.problem.name = t1.name
+        results.problem.name = os.path.join(ws.working_directory, t1.name + '.gms')
```
qtothec, line 334: Should be fine, though hopefully nobody else was doing checks based on this value.

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -576,15 +576,34 @@ def available(self, exception_flag=True):
...
+                raise ImportError("Import of gams failed - GAMS direct "
+                                  "solver functionality is not available.\n"
+                                  "GAMS message: %s" % (e,))
+        except:
+            logger.warning(
+                "Attempting to import gams generated unexpected exception:\n"
+                "\t%s: %s" % (sys.exc_info()[0].__name__, sys.exc_info()[1]))
+            if not exception_flag:
+                return False
+            raise
```
qtothec, line 605: We might want to use the deferred import system for this. @jsiirola do you have documentation somewhere for that?

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -606,12 +625,22 @@ def _get_version(self):
...
+    def _parse_special_values(value):
+        if value == 1.0e300 or value == 2.0e300:
+            return float('nan')
+        if value == 3.0e300:
+            return float('inf')
+        if value == 4.0e300:
+            return -float('inf')
+        if value == 5.0e300:
+            return sys.float_info.epsilon
+        return value
```
qtothec, line 642: out of curiosity, are these documented somewhere? I wasn't able to find them doing a cursory scan of GAMS documentation.

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -722,7 +755,8 @@ def solve(self, *args, **kwds):
         ####################################################################
 
         exe = self.executable()
-        command = [exe, output, "o=" + lst, "curdir=" + tmpdir]
+        command = [exe, output, "o=" + lst, "curdir=" + tmpdir, "solvelink=5", 
+                   "limrow=0", "limcol=0", "solprint=off"]
```
qtothec, line 759: `solvelink`, `limrow`, `limcol` and `solprint` can be set via options. Do we really want to enforce them by default here for all users?

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -644,6 +673,10 @@ def solve(self, *args, **kwds):
         # Make sure available() doesn't crash
         self.available()
 
+        from gdxcc import new_gdxHandle_tp, gdxCreateD, gdxClose, gdxFree
+        from gdxcc import gdxOpenRead, gdxDataReadRawStart, gdxDataReadRaw
+        from gdxcc import gdxDataReadDone, gdxSymbolInfo
```
qtothec, line 678: same comment about deferred import system.

pyomo/solvers/plugins/solvers/GAMS.py
renkekuhlmann: Very good point. I haven't thought about that earlier. I would suggest to use those values on default, but if a user wishes to use something else, that should be fine. This may also apply to the `$offlisting` in https://github.com/Pyomo/pyomo/pull/1446/files#diff-44a99086316025ea73571ead9ddbfb19R581.

qtothec: `$offlisting` I'm not too worried about, since if `keepfiles=True`, they have the original `*.gms` file right next to it.

I should clarify: I am not opposed to having these be the default options, but I worry that feeding them in via command line here might override options set within `add_options`.

pyomo/solvers/plugins/solvers/GAMS.py
renkekuhlmann: Not sure. Usually, those values are transformed in most interfaces except the 'expert-level' interfaces (I guess). So a user wouldn't normally see them. We should improve the documentation of the 'expert-level' APIs though...

pyomo/solvers/plugins/solvers/GAMS.py
jsiirola: I think it is "6 of one, half dozen of the other".  If you wanted to use the deferred import, it would look something like:
```python
gdxcc, gdxcc_available = attempt_import('gdxcc', defer_check=True)
```
Then `available()` just needs to use:
```python
if gdxcc_available:
    return True
elif exception_flag:
    raise ImportError("Import of gams failed - GAMS direct "
                      "solver functionality is not available.\n"
                      "GAMS message: %s" % (e,))
else:
    return False
```

pyomo/solvers/plugins/solvers/GAMS.py
jsiirola: I agree, I think that overwriting any options that the user thinks they are setting could be potentially problematic.

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -287,20 +287,20 @@ def solve(self, *args, **kwds):
             finally:
                 # Always name working directory or delete files,
                 # regardless of any errors.
-                if keepfiles:
+                if keepfiles and tee:
```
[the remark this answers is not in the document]
renkekuhlmann: It seemed appropriate to me. I don't mind using any other option instead of `tee`. Any suggestions? The only thing that mattered to me was to have the possibility to fully disable output.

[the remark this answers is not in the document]
jsiirola: I would advocate removing the dependence on `tee`, and instead have all solver output be sent out through a logger with either `pyomo.solvers` or `__name__` scoping.  

(One of my longstanding TODOs is to replace all `print()` and `sys.stdout.write()` calls with appropriate logger calls... the solver interfaces are some of the worst offenders)

pyomo/solvers/plugins/solvers/GAMS.py
renkekuhlmann: These options are now set through gams option statements in the gams_writer if they haven't been set otherwise.

pyomo/solvers/plugins/solvers/GAMS.py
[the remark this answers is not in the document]
renkekuhlmann: I undid my changes here and leave the printing issue for you. I will simply disable those lines locally for me then.

jsiirola:
@renkekuhlmann, I think this is great and was just about to approve and get @qtothec to do a final review.  However, I just realized something I should have caught earlier (after all, you explicitly put it in the summary): merging this will mean that users will *only* be able to use the GAMS solvers if hey have set up the GAMS python bindings, which (depending on the GAMS version the user has installed) is a subset of the Python versions that Pyomo supports.  How difficult would it be to have the interface fall back on the `put` interface when the GAMS executable was available, but the python bindings were not?

ghackebeil:
I think in the long term, the behavior would be more transparent if we don't make it look like the solver_io="python" version of gams is available when it is not. It would also make debugging more difficult. If we ask a user what interface to GAMS they are using and they say solver_io="python", we really won't know what they mean. Is this what you are suggesting @jsiirola?

jsiirola:
@ghackebeil, in a sense, yes.  Before this PR, we had two interfaces: 
- `gms` worked by generating GMS files and parsing the files generated by GAMS
- `python` worked directly through the python bindings.

So, `gms` worked as long as GAMS was installed, but `python` required both that GAMS and the GAMS python bindings were installed.  This PR changes the `gms` IO to use direct python bindings to GDX for retrieving the results.  It is significantly faster, and a great development, but now both solver interfaces explicitly require the python bindings.  My concern is around the "loss" of functionality for the case where a user would have GAMS, but not the python bindings (which happens on some of our build environments).

renkekuhlmann:
@jsiirola What is the reason for a user having GAMS but not the Python bindings?

jsiirola:
Pyomo supports Python 2.7, and 3.4-3.8.  GAMS generally supports a subset of that: for example 29.1.0 (which is what we test against) only supports Python 2.7, 3.4, 3.6, and 3.7.  So, a user running Python 3.8 (or 3.5) could have GAMS, but not the GAMS Python bindings.

qtothec:
@renkekuhlmann That, and to avoid explaining to someone who might have multiple GAMS versions installed on their computer that their python bindings are hooked up to the wrong one, when they just want to see if this idea works.

renkekuhlmann:
I think `gdxcc` located in `<gams_system_dir>/apifiles/Python/src/gdxcc.py` is more or less Python version independent (it needs at least Python 2.7). You can install it using `python gdxsetup.py install` located in the same directory. I tested it with Python 3.8. `test_GAMS.py` worked.

renkekuhlmann:
My colleague replied that the procedure described above should indeed work for Python 2.7+, but as @jsiirola pointed out, not all Python versions are officially supported by GAMS. Important to note is that we plan to drop official support for Python 2.7 for some upcoming GAMS release. So there is no guarantee that this will work for Python 2.7 in the future for upcoming GAMS versions. However, if you stay with GAMS 29.1.0 on your build machines this shouldn't be a problem.

jsiirola:
I just submitted [PR#1](https://github.com/renkekuhlmann/pyomo/pull/1) to this PR to re-introduce the "put" mechanism for retrieving results from GAMS as a fallback when the GDX python bindings are not available.

PR #1282, opened 2020-02-01
[WIP] add reduced hessian example

santiagoropb:
## Summary/Motivation:
Implement a basic example that demonstrates how to compute the reduced hessian of a nonlinear equality constrained problem with pynumero. Extensions to general nonlinear optimization problems with bounds can be implemented in pynumero following a very similar approach where the KKT matrix is the augmented system of the barrier subproblem at the optimal point. @carldlaird @blnicho @michaelbynum please refer to A.35 in [Here](https://github.com/santiagoropb/paper_pynumero/blob/master/draft0.pdf) for the computation of the KKT matrix.

The calculation of the reduced hessian can be beneficial for different applications of pyomo models. These include advanced step NMPC (e.g. pyomo.dae  @blnicho ) and estimation of covariance matrices in parameter estimation problems (e.g. Parmest @kaklise, KIPET @salvadorgarciamunoz). Note that now that pynumero is supported on windows the implementation of the reduced hessian can be done in pynumero without the need of compiling any additional packages 

@carldlaird for PyNumero itself, efficient calculation of the reduced hessian can be very useful for designing new algorithms. An interesting example is presented in [here](https://9d378b24-a-bf4ca052-s-sites.googlegroups.com/a/wisc.edu/zavalab/publications/journalpubs/optima_reg_parest_v8.pdf?attachauth=ANoY7coxNMWlU6r9Det_YQy7Wh4GZjfCusqHtsAe4mHzbf5h2O2pxfOY_aOavr_BH26QCoFIU6UCTK7fERhTf73VFdhueosQSx0LTVX4USrjGl_EmIlBeSF1SkCsveHtVII9QcswhuwpcEElAEMU5bkc6w48Xn-WwJx7gfROrdW2_Y98It_XhDdI1LEfgp6fKzL1_jQTgtQ6CfyhjcH0Nirn72ApLLBHzFGZTOvw6I7gtwp6DwKQW3mefHKvxibtQLKzugxqYYiC&attredirects=0). There information from the reduced hessian is used within the inertia correction.

## Changes proposed in this PR:
-
-

### Legal Acknowledgement

By contributing to this software project, I agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #1450, opened 2020-05-18
Interior Point

michaelbynum:
## Summary/Motivation:
This PR implements a basic interior point algorithm using Pynumero. There is still plenty of work to do, but I think this is a good start.

### Legal Acknowledgement


pyomo/contrib/interior_point/interface.py
```diff
@@ -0,0 +1,624 @@
...
+
+    def regularize_hessian(self, kkt, coef, copy_kkt=True):
+        raise RuntimeError(
+            'Hessian of Lagrangian regularization is necessary but no '
+            'function has been implemented for doing so.')
+
+
+class InteriorPointInterface(BaseInteriorPointInterface):
+    def __init__(self, pyomo_model):
+        if type(pyomo_model) is str:
```
carldlaird, line 240: Should this take in a PyNumero NLP instead of a pyomo model or ampl model? I don't see the code making use of anything out of the higher level models.

pyomo/contrib/interior_point/interface.py
```diff
@@ -0,0 +1,624 @@
...
+        return self._nlp.evaluate_grad_objective()
+
+    def evaluate_jacobian_eq(self):
+        return self._nlp.evaluate_jacobian_eq()
+
+    def evaluate_jacobian_ineq(self):
+        return self._nlp.evaluate_jacobian_ineq()
+
+    def regularize_equality_gradient(self, kkt, coef, copy_kkt=True):
+        # Not technically regularizing the equality gradient ...
```
carldlaird, line 540: Hmmm. I agree with your comment. Maybe we should change the name?

pyomo/contrib/interior_point/interior_point.py
```diff
@@ -0,0 +1,669 @@
...
+                                  coef=total_hess_reg_coef, neg_eig=neg_eig, status=status)
+                reg_iter += 1
+                if reg_iter > self.max_reg_iter:
+                    raise RuntimeError('Exceeded maximum number of regularization iterations.')
+                last_hess_reg_coef = total_hess_reg_coef
+                total_hess_reg_coef *= self.reg_factor_increase
+
+        return last_hess_reg_coef
+
+    def process_init(self, x, lb, ub):
```
carldlaird, line 376: What do these methods do?

pyomo/contrib/interior_point/interior_point.py
```diff
@@ -0,0 +1,669 @@
...
+    if np.any((ub - lb) < 0):
+        raise ValueError(
+            'Lower bounds for variables/inequalities should not be larger than upper bounds.')
+    if np.any((ub - lb) == 0):
+        raise ValueError(
+            'Variables and inequalities should not have equal lower and upper bounds.')
+
+    lb_mask = build_bounds_mask(lb)
+    ub_mask = build_bounds_mask(ub)
+
```
carldlaird, line 643: Do these masks not exist already somewhere?

pyomo/contrib/interior_point/inverse_reduced_hessian.py
```diff
@@ -0,0 +1,124 @@
+import pyomo.environ as pyo
+from pyomo.opt import check_optimal_termination
+from pyomo.common.dependencies import attempt_import
+from .interface import InteriorPointInterface
+from .linalg.scipy_interface import ScipyInterface
+
```
carldlaird, line 6: I know that I am responsible for this, but should we move this file somewhere else?
Also, does it work with this interior-point method, or does it truly need IPOPT?

pyomo/contrib/interior_point/linalg/results.py
```diff
@@ -0,0 +1,14 @@
+import enum
+
+
+class LinearSolverStatus(enum.Enum):
+    successful = 0
+    not_enough_memory = 1
+    singular = 2
+    error = 3
+    warning = 4
+
```
carldlaird, line 10: Do we call out not_enough_memory and singular because these are the only errors we do anything special with?

pyomo/contrib/interior_point/linalg/scipy_interface.py
```diff
@@ -0,0 +1,67 @@
...
+from scipy.linalg import eigvals
+from scipy.sparse import isspmatrix_csc
+from pyomo.contrib.pynumero.sparse.block_vector import BlockVector
+import logging
+import numpy as np
+
+
+class ScipyInterface(LinearSolverInterface):
+    def __init__(self, compute_inertia=False):
+        self._lu = None
```
carldlaird, line 13: I think we should add a comment here to indicate that compute_inertia could make things pretty slow for large problems (computing the eigenvalues can be very expensive)

pyomo/contrib/interior_point/interface.py
michaelbynum: Good idea.

pyomo/contrib/interior_point/interface.py
michaelbynum: We could have a single method named `regularize_kkt(self, kkt, hess_coef, grad_coef, copy_kkt=True)`?

michaelbynum: I can't think of better names if we keep the methods split apart.

pyomo/contrib/interior_point/interior_point.py
michaelbynum: They move variables to the interior of their bounds.

pyomo/contrib/interior_point/interior_point.py
michaelbynum: Good question. I originally had a bunch of compression matrices floating around with lots of matrix vector dot products. However, when I started profiling, those matrix vector dot products were the first thing that showed up. It turns out that, throughout most of the algorithm, numpy handles the bounds of `-inf` and `inf` quite nicely. In most parts of the algorithm, we are dividing by `x - x_L` or `x_U - x`. When the denominator is `inf`, numpy returns `0`, so things work out nicely. Because of this, I just used the full arrays of primals, duals, etc. throughout the algorithm instead of compressed arrays. The duals for variables with bounds are just always `0`. This just leaves a couple places where we have to worry about infinite bounds: when processing initial values and in the convergence check.

pyomo/contrib/interior_point/inverse_reduced_hessian.py
michaelbynum: I'm not sure where we would move it to? Do we need a directory for this one file? I vote we move it later. I'm sure it will be more clear where it should live after this module matures a bit.

I think it should work with this interior-point method if it converges. The algorithm has an inertia correction now, so if it converges, it should be correct.

pyomo/contrib/interior_point/linalg/results.py
michaelbynum: Exactly. We can add more to this enum as needed.

pyomo/contrib/interior_point/linalg/scipy_interface.py
michaelbynum: Good point.

pyomo/contrib/interior_point/interface.py
carldlaird: I like regularize_kkt. hess_coef and eqcon_coef?

pyomo/contrib/interior_point/inverse_reduced_hessian.py
carldlaird: Ya, I think it is the reliance on Ipopt suffixes that is more problematic.

pyomo/contrib/interior_point/interface.py
michaelbynum: There are two methods that make use of features of a Pynumero `PyomoNLP`: `_get_full_duals_primals_bounds` and `load_primals_into_pyomo_model`. Should these methods be moved elsewhere? Should we require a `PyomoNLP`? I'm not sure what the best thing to do here is.

michaelbynum: I was wrong. There are several methods at the bottom of the class that implement methods from the `PyomoNLP`. Maybe we should have a class that inherits from this class for Pyomo models? `InteriorPointInterface` and `PyomoInteriorPointInterface`? Then we can move these methods to the derived class?

michaelbynum:
Okay, I am going to merge this and work on some of these updates in a separate PR.

pyomo/contrib/interior_point/interface.py
santiagoropb: What about hess_coefficients and eqcon_coefficients in a vector format instead of a scalar. Even a block vector format in case the regularization wants to target a particular subproblem.

pyomo/contrib/interior_point/interface.py
```diff
@@ -0,0 +1,624 @@
...
+    def ineq_ub(self):
+        return self._nlp.ineq_ub()
+
+    def set_barrier_parameter(self, barrier):
+        self._barrier = barrier
+
+    def pyomo_nlp(self):
+        return self._nlp
+
+    def evaluate_primal_dual_kkt_matrix(self, timer=None):
```
santiagoropb, line 391: Can this be used to extract the KKT system of the barrier problem at x*, lambda*? If so this matrix and the rhs is all that is needed to compute the inverse of the reduced hessian in a similar fashion to what it is in #1282

pyomo/contrib/interior_point/linalg/ma27_interface.py
```diff
@@ -0,0 +1,125 @@
+from .base_linear_solver_interface import LinearSolverInterface
+from .results import LinearSolverStatus, LinearSolverResults
+from pyomo.contrib.pynumero.linalg.ma27 import MA27Interface
+from scipy.sparse import isspmatrix_coo, tril
+from pyomo.contrib.pynumero.sparse import BlockVector
+
+
+class InteriorPointMA27Interface(LinearSolverInterface):
```
santiagoropb, line 8: This is pretty cool

pyomo/contrib/interior_point/interior_point.py
```diff
@@ -0,0 +1,669 @@
...
+            timer.start('back solve')
+            with self.linear_solve_context:
+                self.logger.info('Iter: %s' % self._iter)
+                delta = linear_solver.do_back_solve(rhs)
+            timer.stop('back solve')
+
+            interface.set_primal_dual_kkt_solution(delta)
+            timer.start('frac boundary')
+            alpha_primal_max, alpha_dual_max = \
+                    self.fraction_to_the_boundary()
```
santiagoropb, line 300: No line search?

michaelbynum: Not yet. Starting simple. We'll get there.

pyomo/contrib/interior_point/interface.py
michaelbynum: Yes, it can. Good point.

PR #1456, opened 2020-05-18
WIP - Pynumero poek

whart222:
- A POEK interface using the COEK NL writer
- A POEK interface using the direct interface to the nlp_model class

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/pynumero/interfaces/poeknl_nlp.py
+++ b/pyomo/contrib/pynumero/interfaces/poeknl_nlp.py
@@ -12,7 +12,6 @@
 This module defines the classes that provide an NLP interface based on
 the Ampl Solver Library (ASL) implementation
 """
-import poek as pk
 from pyomo.contrib.pynumero.interfaces.ampl_nlp import AslNLP
 import pyutilib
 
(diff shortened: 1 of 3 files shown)


whart222:
I don't see an obvious way to enable the POEK tests.  That would require installing COEK, for which we don't have a CONDA build yet.

whart222:
Note:  This code isn't ready for review.  The PR is intended to get testing started.

jsiirola:
@whart222: FWIW, you don't need to create a PR to get testing.  We also have individual branch testing set up for pyomo.  You are already getting a single (old) build (see: https://github.com/whart222/pyomo/actions).  If you update your branch from master, you will get 4 builds covering 4 python versions and 3 OS's (plus MPI)

pyomo/contrib/pynumero/interfaces/poeknl_nlp.py
```diff
@@ -0,0 +1,242 @@
...
+try:
+    import poek as pk
+    poek_available=True
+    __all__ = ['PoekNLP_NLP']
+except:
+    poek_available=False
+    __all__ = []
+
+
+class PoekNL_NLP(AslNLP):
```
santiagoropb, line 31: My guess is that PoekNL_NLP is mostly for testing performance. Is that correct? It would be very interesting to see how does performance of the AD from Poek compare against ASL. If the AD of Poek is as fast or close as fast to ASL then PoekNLP would be always a better choice than PoekNL_NLP? It would also be interesting to see how does the nl writer of Poek compares to the one in Pyomo (you probably have already some metrics on that?)

Its cool seeing that pynumero is facilitating this comparisons :)

blnicho:
@whart222 this has been open as a WIP for a while. I'm going to close this as the main motivation for opening this was to start looking at test failures and we now have testing set up for branches on forks.

PR #1460, opened 2020-05-21
Bug fixes and improvements to Pyomo units

carldlaird:
## Fixes # .
- Fixes bug when converting from dimensionless to dimensionless (in convert())
- adds functionality to support specifying units for arguments of external functions

-
-

### Legal Acknowledgement


pyomo/core/base/units_container.py
```diff
@@ -1409,6 +1409,17 @@ def convert(self, src, to_units=None):
...
+        # check if they are both dimensionless
+        src_dimensionless = \
+            _UnitExtractionVisitor(self)._pint_unit_equivalent_to_dimensionless(src_pint_unit)
+        to_dimensionless = \
+            _UnitExtractionVisitor(self)._pint_unit_equivalent_to_dimensionless(to_pint_unit)
+        if src_dimensionless and to_dimensionless:
+            return src
+        elif src_dimensionless or to_dimensionless:
+            raise InconsistentUnitsError(src_pint_unit, to_pint_unit,
+                                         'Error in convert: units not compatible.')
```
jsiirola, line 1421: Consistency question: below the exception is:
```python
raise UnitsError('Cannot convert {0:s} to {1:s}. Units are not compatible.'.format(str(src_pyomo_unit), str(to_pyomo_unit)))
```
Is there a reason that this raises a different message?

andrewlee94:
@jsiirola the get units methods do return `None`, which is expected based on the way the method are structured. If only one of the two sets of units is `None`, then an exception stating that units are not consistent is raised, again as expected.

However, if both sets of units are `None`, then the units are correctly identified as consistent but the final return statement ends up being:

`fac_b_src/fac_b_dest*to_pyomo_unit/src_pyomo_unit*src =  1/1*None/None*src`

This raises the following exception:

 `TypeError: unsupported operand type(s) for *: 'float' and 'NoneType'`

jsiirola:
@andrewlee94: If that's the case, would a simpler (and more efficient?) solution just be to:
```python
if src_pyomo_unit is None:
    return src
else:
    return fac_b_src/fac_b_dest*to_pyomo_unit/src_pyomo_unit*src
```

andrewlee94:
@jsiirola That was my first suggestion actually - I can only assume @carldlaird found something I didn't.

pyomo/core/base/units_container.py
```diff
@@ -33,22 +33,23 @@
 be used directly in expressions (e.g., defining constraints). You can also
 verify that the units are consistent on a model, or on individual components
 like the objective function, constraint, or expression using
-`assert_units_consistent`. There are other methods that may be helpful
-for verifying correct units on a model.
+`assert_units_consistent` (from pyomo.util.units_checking).
```
jsiirola, line 36: Silly naming question, but would `pyomo.util.units_checker` be a better (more intuitive) name for the new module?

pyomo/core/base/units_container.py
```diff
@@ -1408,28 +1419,17 @@ def convert(self, src, to_units=None):
...
+            return src
 
         # no offsets, we only need a factor to convert between the two
         fac_b_src, base_units_src = self._pint_registry.get_base_units(src_pint_unit, check_nonmult=True)
         fac_b_dest, base_units_dest = self._pint_registry.get_base_units(to_pint_unit, check_nonmult=True)
 
         if base_units_src != base_units_dest:
-            raise UnitsError('Cannot convert {0:s} to {1:s}. Units are not compatible.'.format(str(src_pyomo_unit), str(to_pyomo_unit)))
+            raise InconsistentUnitsError(src_pint_unit, to_pint_unit,
+                                             'Error in convert: units not compatible.')
```
jsiirola, line 1432: Would it make sense to add the units to the exception message?

Also, silly thing, but it looks like there is a weird indentation for the second line.

pyomo/core/expr/numeric_expr.py
```diff
@@ -638,8 +638,12 @@ def _apply_operation(self, result):
     def _to_string(self, values, verbose, smap, compute_values):
         return "{0}({1})".format(self.getname(), ", ".join(values))
 
+    def get_arg_units(self):
+        """ Return the units for this external functions arguments """
+        return self._fcn.get_arg_units()
+
     def get_units(self):
-        """ Return the units for this external function expression """
+        """ Get the returned units for this external function """
```
jsiirola, line 646: Possible clarification:
```
Get the units of the return value of this external function
```

pyomo/util/units_checking.py
```diff
@@ -0,0 +1,175 @@
...
+from pyomo.core.base.constraint import Constraint
+from pyomo.core.base.var import Var
+from pyomo.core.base.param import Param
+from pyomo.core.base.suffix import Suffix
+from pyomo.core.base.set import Set, RangeSet
+from pyomo.gdp import Disjunct
+from pyomo.gdp import Disjunction
+from pyomo.core.base.block import Block
+from pyomo.core.base.external import ExternalFunction
+from pyomo.core.base.expression import Expression
```
jsiirola, line 27: Would the following be simpler / more "future proof":
```python
from pyomo.core.base import (
    Objective, Constraint, Var, Param, Suffix, Set, RangeSet, Block, 
    ExternalFunction, Expression,
)
```

pyomo/core/base/units_container.py
carldlaird: to me, units_checker implies that there is an class (just because it is a noun, I guess). This module is currently only a list of functions, which is why I named it the way I did. I don't feel strongly about it.

pyomo/core/base/units_container.py
carldlaird: The units are included in the exception message (they are passed in as the first two arguments, and added to the message in InconsistentUnitsError). I can look at the indentation. Have we ever standardized on what the indentation for continuation stuff should be?

pyomo/core/expr/numeric_expr.py
carldlaird: OK, I can make this change.

pyomo/util/units_checking.py
carldlaird: Sure. I didn't know that this would work.

carldlaird: I will make this change too.

pyomo/core/base/units_container.py
jsiirola: Perfect.

As for standards, nope.  My editor (emacs) aligns things within the parens (so the "`'`" character would be directly under the '`s`' from `src_pint_unit`).

PR #1463, opened 2020-05-22
GAMS: handle fixed variables in LinearExpressions

### LINKED ISSUE #924: GAMS writer fails to identify fixed variables in LinearExpression

jsiirola:
## Summary/Motivation:
Alternative implementation to #917: fixes the bug where fixed variables appearing in a Linear Expression are sent to GAMS as free variables.

This implementation adds extra processing to make sure that any fixed variables that are emitted by the expression walkers (e.g., in the case of LinearExpression) are marked as 'fixed' in the GAMS input.  An alternative implementation (#917) adds special processing for the LinearExpression node in the `ToGamsVisitor`.

The advantage of this implementation is that it should also correctly handle other (future) cases where a fixed variable could sneak through the visitor logic.  The advantage of the #917 implementation is that fixed variables are completely hidden from GAMS.

- catch fixed variables in the variable categorization logic and output them to GAMS as fixed.
- add a test to the writer for this situation (with LinearExpression nodes)

### Legal Acknowledgement

### DIFF
--- a/pyomo/repn/plugins/gams_writer.py
+++ b/pyomo/repn/plugins/gams_writer.py
@@ -180,11 +180,14 @@ class Categorizer(object):
         self.ints = []
         self.positive = []
         self.reals = []
+        self.fixed = []
 
         # categorize variables
         for var in var_list:
             v = symbol_map.getObject(var)
-            if v.is_binary():
+            if v.is_fixed():
+                self.fixed.append(var)
+            elif v.is_binary():
                 self.binary.append(var)
             elif v.is_integer():
                 if (v.has_lb() and (value(v.lb) >= 0)) and \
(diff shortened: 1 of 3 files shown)


jsiirola:
@qtothec: we should pick one of these fixes (#917 or #1463) for fixed variables in GAMS.  What are your thoughts?

qtothec:
I forgot to mention: the added commit is just porting the tests over from the other PR.

jsiirola:
@qtothec, do you think we need the same sort of fix for the Baron writer?

qtothec:
@jsiirola Yes, it at least looks like the result you get back can be incorrect:
```
Model unknown

  Variables:
    y : Size=1, Index=None
        None :     0 :     0 :     1 :  True :  True : Binary
    x : Size=1, Index=None
        None :     0 :   0.0 :  None : False : False :  Reals

  Objectives:
    obj : Size=1, Index=None, Active=True
        Key  : Active : Value
        None :   True :   0.0

  Constraints:
    c1 : Size=1
        Key  : Lower : Body : Upper
        None :   0.0 :    0 :  None
    c2 : Size=1
        Key  : Lower : Body : Upper
        None :   1.0 :  0.0 :   1.0
```

Note `c2` on the test problem you added.

PR #1470, opened 2020-05-29
Ensure sympy configuration happens

michaelbynum:
## Fixes
This PR fixes the error below.

```
from pyomo.core.expr.sympy_tools import PyomoSympyBimap, sympy2pyomo_expression, sympy_available
import sympy
import pyomo.environ as pe
x = sympy.Symbol('x')
y = sympy.Symbol('y')
e = x + y
m = pe.ConcreteModel()
m.x = pe.Var()
m.y = pe.Var()
obj_map = PyomoSympyBimap()
obj_map.sympy2pyomo[x] = m.x
obj_map.sympy2pyomo[y] = m.y
try:
    sympy2pyomo_expression(e, obj_map)
except Exception as err:
    print(err)
bool(sympy_available)
print(sympy2pyomo_expression(e, obj_map))
```

Output:
```
Internal Pyomo implementation error:
	"sympy expression type '<class 'sympy.core.add.Add'>' not found in the operator map"
	Please report this to the Pyomo Developers.
x + y
```

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/expr/sympy_tools.py
+++ b/pyomo/core/expr/sympy_tools.py
@@ -139,6 +139,7 @@ class PyomoSympyBimap(object):
 class Pyomo2SympyVisitor(EXPR.StreamBasedExpressionVisitor):
 
     def __init__(self, object_map):
+        sympy.Add  # this ensures _configure_sympy gets run
         super(Pyomo2SympyVisitor, self).__init__()
         self.object_map = object_map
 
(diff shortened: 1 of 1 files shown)


jsiirola:
Good catch!  Some thoughts: I believe the same bug applies to both `Pyomo2SympyVisitor` and `Sympy2PyomoVisitor` (assuming in the latter case that the user hand-created the object map).  Would it be better to put the check in the two visitors' `__init__()` to catch the case where users directly worked with the visitors and not the conversion functions?  Also, what do you think of just starting off the `__init__()` with:
```python
sympy.Add
```
...this would trigger the normal `DeferredImportError` instead of another (different) exception message?

jsiirola:
Test failures are unrelated to this PR (either NEOS connectivity, or have already been resolved by #1473).  This is safe to merge.

PR #1476, opened 2020-06-01
Fix warmstart and add cycling check in MindtPy

ZedongPeng:
## Summary/Motivation:

- Fix the warmstart for both Outer Approximation algorithm and LP/NLP algrithm in MindtPy. 
- Add cycling check in Outer Approximation algorithm in MindtPy.
- Fix several bugs in MindtPy.
- Rename the online doc example to constraint qualification example, add a new online doc example.
- Add the support for gams solvers both for MILP and NLP.

### Legal Acknowledgement


pyomo/contrib/mindtpy/MindtPy.py
```diff
@@ -210,7 +210,7 @@ class MindtPySolver(object):
         domain=bool
     ))
     CONFIG.declare("add_integer_cuts", ConfigValue(
-        default=True,
+        default=False,
```
bernalde, line 213: I'm still debating what is best to do here. In general, the integer cuts are not required for convex programs, and we do not have an implementation for general Integer variables yet (only binaries so far). Including them might hurt our capability of providing valid lower bounds to the optimal solution.
DICOPT has them on by default, Pajarito doesn't.

pyomo/contrib/mindtpy/single_tree.py
```diff
@@ -125,18 +125,18 @@ def handle_lazy_master_mip_feasible_sol(self, master_mip, solve_data, config, op
...
-                # self.get_best_objective_value(),
-                solve_data.LB)
-            solve_data.LB_progress.append(solve_data.LB)
-        else:
-            solve_data.UB = min(
-                self.get_objective_value(),
-                # self.get_best_objective_value(),
-                solve_data.UB)
-            solve_data.UB_progress.append(solve_data.UB)
+        # if main_objective.sense == minimize:
```
bernalde, line 128: We should remove these lines instead of commenting them @ZedongPeng

pyomo/contrib/mindtpy/initialization.py
```diff
@@ -144,7 +149,11 @@ def init_max_binaries(solve_data, config):
     opt = SolverFactory(config.mip_solver)
     if isinstance(opt, PersistentSolver):
         opt.set_instance(m)
-    results = opt.solve(m, options=config.mip_solver_args)
+    mip_args = dict(config.mip_solver_args)
+    if config.mip_solver == 'gams':
+        mip_args['add_options'] = mip_args.get('add_options', [])
+        mip_args['add_options'].append('option optcr=0.01;')
```
bernalde, line 155: For convergence to be guaranteed we actually require the MILP to be solved to optimality (`optcr=0.0`)

pyomo/contrib/mindtpy/mip_solve.py
```diff
@@ -83,16 +83,20 @@ def solve_OA_master(solve_data, config):
         masteropt._solver_model.set_log_stream(None)
         masteropt._solver_model.set_error_stream(None)
         masteropt.options['timelimit'] = config.time_limit
+    mip_args = dict(config.mip_solver_args)
+    if config.mip_solver == 'gams':
+        mip_args['add_options'] = mip_args.get('add_options', [])
+        mip_args['add_options'].append('option optcr=0.01;')
```
bernalde, line 89: Same comment as above.

pyomo/contrib/mindtpy/initialization.py
```diff
@@ -15,13 +15,18 @@
 from pyomo.contrib.mindtpy.nlp_solve import (solve_NLP_subproblem,
                                              handle_NLP_subproblem_optimal, handle_NLP_subproblem_infeasible,
                                              handle_NLP_subproblem_other_termination)
+from pyomo.contrib.mindtpy.single_tree import var_bound_add
```
jsiirola, line 18: This unconditionally imports `single_tree`, which in turn unconditionally imports cplex.  This leads to failures on platforms where the cplex python bindings are not present.

pyomo/contrib/mindtpy/tests/test_mindtpy_lp_nlp.py
```diff
@@ -121,19 +121,33 @@ def test_lazy_OA_Proposal(self):
...
+    def test_lazy_OA_ConstraintQualificationExample(self):
+        with SolverFactory('mindtpy') as opt:
+            model = ConstraintQualificationExample()
+            print('\n Solving ConstraintQualificationExample with LP/NLP')
+            results = opt.solve(model, strategy='OA',
+                                mip_solver=required_solvers[1],
+                                nlp_solver=required_solvers[0],
+                                single_tree=True
+                                )
+            # TODO: constraint qualification hold true in this case
```
bernalde, line 133: I guess we can remove this TODO. We can also add an assertion for a feasible solution instead of optimal (uncomment and correct next line).

pyomo/contrib/mindtpy/MindtPy.py
```diff
@@ -191,7 +191,7 @@ class MindtPySolver(object):
         description="Tolerance on variable bounds."
     ))
     CONFIG.declare("zero_tolerance", ConfigValue(
-        default=1E-10,
+        default=1E-8,
```
bernalde, line 194: @jsiirola does Pyomo have constants for tolerances and maximum values such that we can use them here? In this case, we just used values that made sense but maybe there are some global constants we can set by default.

jsiirola: @bernalde, Unfortunately, no.  Someone else (@qtothec? @michaelbynum?) was asking a similar question.  I think it would be reasonable to set up a "global" repository of default tolerances / constants (and make them settable through the Pyomo user configuration), but nothing exists at the moment.

pyomo/contrib/mindtpy/initialization.py
qtothec: @bernalde w.r.t. your question about how to handle this. You want to make sure that your symbols are properly scoped. Otherwise you're going to run yourself into an import loop someday. I would argue that `var_bound_add` should be moved to a more general utility module.

pyomo/contrib/mindtpy/MindtPy.py
qtothec: @jsiirola That was me. Hardest part is likely consistent naming and scoping. "zero_tolerance" to one person might be "feasibility_tolerance" to another. Frankly, the registry would help with that issue.

pyomo/contrib/mindtpy/MindtPy.py
```diff
@@ -231,6 +231,21 @@ class MindtPySolver(object):
                     "slack variables here are used to deal with nonconvex MINLP",
         domain=bool
     ))
+    CONFIG.declare("continuous_var_bound", ConfigValue(
+        default=1e24,
```
jsiirola, line 235: I think I would make this `sys.float_info.max` (usually 1.7976e+308), unless you have a reason to do something different.  1e24 is probably excessively restrictive.

pyomo/contrib/mindtpy/util.py
```diff
@@ -94,3 +94,26 @@ def add_feas_slacks(m):
...
+    m = solve_data.working_model
+    MindtPy = m.MindtPy_utils
+    for c in MindtPy.constraint_list:
+        if c.body.polynomial_degree() not in (1, 0):
+            for var in list(EXPR.identify_variables(c.body)):
+                if var.has_lb() and var.has_ub():
+                    continue
+                elif not var.has_lb():
+                    if var.is_integer():
+                        var.setlb(-config.integer_var_bound)
```
jsiirola, line 112: Note that you should make this `var.setlb(-(config.integer_var_bound-1))` due to ints being stored as 2's-compliment.

jsiirola:
@bernalde, The (new) standard we are using for optional dependencies is to import the library using `pyomo.common.dependencies.attempt_import`, e.g.:
```python
cplex, cplex_available = attempt_import('cplex')
```
There is a bit of overhead with `attempt_import`, so it is generally preferable to call it only once and then import the result into other modules where you might need it.

You can then skip any test or test suite that requires code that requires cplex with something like:
```python
@unittest.skipIf(not cplex_available, "These tests require CPLEX")
class Test_NNN(unittest.TestCase):
    # ....
```

pyomo/contrib/mindtpy/nlp_solve.py
```diff
@@ -53,18 +46,28 @@ def solve_NLP_subproblem(solve_data, config):
     # | g(x) <= b  | -1    | g(x1) > b    | g(x1) - b            |
     # | g(x) >= b  | +1    | g(x1) >= b   | 0                    |
     # | g(x) >= b  | +1    | g(x1) < b    | b - g(x1)            |
-
+    flag = False
```
bernalde, line 49: The name of this variable does not say what it means (if `flag=True`, then there was an evaluation error in the constraints). May I suggest changing its name to something else? `evaluation_error` ?

pyomo/contrib/mindtpy/MindtPy.py
```diff
@@ -231,6 +231,21 @@ class MindtPySolver(object):
...
+        domain=PositiveFloat
+    ))
+    CONFIG.declare("integer_var_bound", ConfigValue(
+        default=1e9,
+        description="default bound added to unbounded integral variables in nonlinear constraint if single tree is activated.",
+        domain=PositiveFloat
+    ))
+    CONFIG.declare("cycling_check", ConfigValue(
+        default=True,
+        description="whether check the cycling in OA algorithm.",
```
qtothec, line 246: Suggested: "Check if OA algorithm is stalled in a cycle and terminate."

I assume terminate, but this doesn't actually say.

pyomo/contrib/mindtpy/util.py
```diff
@@ -109,9 +109,9 @@ def var_bound_add(solve_data, config):
                     continue
                 elif not var.has_lb():
                     if var.is_integer():
-                        var.setlb(-config.integer_var_bound)
+                        var.setlb(-config.integer_var_bound - 1)
                     else:
-                        var.setlb(-config.continuous_var_bound)
+                        var.setlb(-config.continuous_var_bound - 1)
```
jsiirola, line 114: Technically, you should not subtract 1 for the continuous bound.  The "`- 1`' only accounts for 2's complement in integers and does not apply (in the same way) to floating point numbers.

PR #1479, opened 2020-06-02
Fix bug in bigm transformation for nested indexed disjunctions

emma58:
## Summary/Motivation:

With the recent rewrite, the following transformation would throw an AttributeError because the `relaxedDisjuncts` container was prematurely deleted when moving up information to transform the disjuncts in the first nested DisjunctionData, meaning we lost the transformation blocks for the disjuncts of the second DisjunctionData.
```
m = ConcreteModel()
m.d1 = Disjunct()
m.d1.indexedDisjunct1 = Disjunct([0,1])
m.d1.indexedDisjunct2 = Disjunct([0,1])
@m.d1.Disjunction([0,1])
def innerIndexed(d, i):
    return [d.indexedDisjunct1[i], d.indexedDisjunct2[i]]
m.d2 = Disjunct()
m.outer = Disjunction(expr=[m.d1, m.d2])

TransformationFactory('gdp.bigm').apply_to(m)
```

- Fixes the above bug by only deleting BlockDatas from relaxedDisjuncts, never the whole container.
- Adds a test

### Legal Acknowledgement

### DIFF
--- a/pyomo/gdp/plugins/bigm.py
+++ b/pyomo/gdp/plugins/bigm.py
@@ -535,8 +535,6 @@ class BigM_Transformation(Transformation):
 
         # delete everything we moved.
         for idx in to_delete:
-            # Note we have to do it by index because python scoping is
-            # obnoxious...
             del fromBlock.relaxedDisjuncts[idx]
 
         # Note that we could handle other components here if we ever needed
--- a/pyomo/gdp/plugins/bigm.py
+++ b/pyomo/gdp/plugins/bigm.py
@@ -42,7 +42,6 @@ from functools import wraps
 from six import iterkeys, iteritems
 from weakref import ref as weakref_ref
 
-
 logger = logging.getLogger('pyomo.gdp.bigm')
 
 NAME_BUFFER = {}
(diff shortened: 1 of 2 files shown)


pyomo/gdp/plugins/bigm.py
```diff
@@ -530,8 +530,14 @@ def _transfer_transBlock_data(self, fromBlock, toBlock):
...
 
-        # we delete this container because we just moved everything out
-        del fromBlock.relaxedDisjuncts
+            # save index of what we just moved so that we can delete it
+            to_delete.append(idx)
+
+        # delete everything we moved.
+        for idx in to_delete:
+            # Note we have to do it by index because python scoping is
+            # obnoxious...
```
qtothec, line 539: Huh? What do you mean?

emma58: Eh, nothing, I just forgot to remove my annoyance at how hard it is to delete something in python sometimes...

jsiirola:
Failing tests are unrelated to this PR (coverage reporting, baron test failures).

PR #1489, opened 2020-06-04
Fix DAE flattener for non-time-indexed blocks

Robbybp:
## Fixes # .
Flattener would throw an error if it encountered a non-time-indexed IndexedBlock before a time-indexed IndexedBlock in its top-down search of the model hierarchy. This was due to calling `component_objects()` on an IndexedBlock.

- Fix bug and add test to catch

### Legal Acknowledgement


pyomo/dae/flatten.py
```diff
@@ -123,15 +123,18 @@ def flatten_dae_variables(model, time):
...
-        for v in b.component_objects(SubclassOf(Var), descend_into=False):
-            v_sets = v.index_set().subsets()
-            if time in v_sets:
-                for _slice in generate_time_only_slices(v, time):
-                    time_indexed_vars.append(Reference(_slice))
-            else:
-                regular_vars.extend(list(v.values()))
+        for blkdata in b.values():
+            block_queue.extend(
+                list(blkdata.component_objects(Block, descend_into=False))
```
qtothec, line 128: random comment, but fwiw, you don't need the cast to `list` here. `extend()` will happily take an iterable.

jsiirola:
This is good to merge: the Jenkins test failures are due to network connectivity issues.

Issue #379, opened 2018-03-03
Do not silently fail to perform slicing


qtothec:
```
from pyomo.environ import *

m = ConcreteModel()
m.s = RangeSet(2)
m.s2 = RangeSet(2)
m.ss = Set(initialize=m.s * m.s2)
m.s3 = RangeSet(3)
m.x = Var(m.ss, m.s3)

print("Case 1")
for var in m.x[:, 3]:
    print(var.name)
    # nothing happens. Expect reasonable error message.

print("Case 2")
for var in m.x[:, :, 3]:
    print(var.name)
    # expected behavior
```

Output:
```
Case 1
Case 2
x[1,2,3]
x[1,1,3]
x[2,1,3]
x[2,2,3]
```

Robbybp:
Another example:

```python
from pyomo.environ import *

m = ConcreteModel()

m.time = Set(initialize=[0,1,2,3])
m.space = Set(initialize=[2,4,6,8])
m.comp = Set(initialize=['a','b','c'])

m.v = Var(m.time, m.space, m.comp, initialize=1)

# Would like this to throw an error:
_slice = m.v[:, 1, :, :]
```

Robbybp:
An example from `test_indexed_slice.py` seems to indicate that some behavior like this is intended. In the following test, 1 is not valid for the second index of `b[1,:].c`, but the expected behavior seems to be to return an empty slice rather than to raise an error.
```python
    def test_empty_slices(self):
        _slicer = self.m.b[1,:].c[:,1].x
        self.assertIsInstance(_slicer, IndexedComponent_slice)
        ans = [ str(x) for x in _slicer ]
        self.assertEqual( ans, [] )
```
I think all the cases where an index is invalid (extra slice, missing fixed index/slice, or invalid fixed index) should raise exceptions. Thoughts?

mrmundt:
@Robbybp - I'm going through old bug reports - do you know if this bug still exists in the current version of Pyomo?

Robbybp:
@mmrundt This issue can be closed. To recap, slices with inconsistent dimension now raise exceptions. Slices with an invalid index, but the correct dimension, simply don't return anything. This behavior is intentional as of the discussion in #1492.

PR #1492, opened 2020-06-09
Check for consistent dimensions when slicing

Robbybp:
## Fixes # .
Addresses part of #379 

## Summary/Motivation:
Slices silently "fail" in several scenarios, some of which are when too many or not enough `slice` objects are provided in the index. While the tests in test_indexed_slice.py indicated that this behavior was expected, I don't think it should be. 

- Raise an error if the dimension of the index provided does not match the dimension of the `IndexedComponent`
- Change and add tests to expect this error

I'm not sure how these consistency checks might need to be altered if `normalize_index.flatten is False`. It seems like in this case I should relax the requirements I impose for dimension consistency, i.e. allow
```
len(fixed) + len(sliced) < self.dim()
```
as each item of `fixed` or `sliced` could correspond to a set of dimension greater than one. I don't know the context in which this flag is False, so there many be more implications I haven't realized. Suggestions on how to address are appreciated.

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/indexed_component.py
+++ b/pyomo/core/base/indexed_component.py
@@ -666,7 +666,15 @@ value() function.""" % ( self.name, i ))
                 fixed[i - len(idx)] = val
 
         if sliced or ellipsis is not None:
-            if (not normalize_index.flatten and 
+            if self.dim() == 0 and idx == (slice(None),):
+                # If dim == 0 and idx is slice(None), the component was
+                # a scalar passed a single slice. Since scalar components
+                # can be accessed with a "1-dimensional" index of None,
+                # this behavior is allowed.
+                #
+                # NOTE: dim will still be 0 if nomalize_index.flatten is False
+                pass
+            elif (not normalize_index.flatten and 
                 ellipsis is None and 
                 len(idx) != len(self._implicit_subsets)):
                 # dim will be None. Absent an ellipsis, the number of indices
@@ -686,12 +694,6 @@ value() function.""" % ( self.name, i ))
             elif self.dim() is None:
                 # Assume that the right thing to do here is return
                 # an IndexedComponent_slice
-                pass
-            elif self.dim() == 0 and idx == (slice(None),):
-                # If dim == 0 and idx is slice(None), the component was
-                # a scalar passed a single slice. Since scalar components
-                # can be accessed with a "1-dimensional" index of None,
-                # this behavior is allowed.
                 pass
             elif ellipsis is None and len(idx) != self.dim():
                 # If there is no ellipse and the index doesn't match the 
--- a/pyomo/core/base/indexed_component.py
+++ b/pyomo/core/base/indexed_component.py
@@ -666,7 +666,24 @@ value() function.""" % ( self.name, i ))
                 fixed[i - len(idx)] = val
 
         if sliced or ellipsis is not None:
-            if self.dim() is None:
+            if (not normalize_index.flatten and 
+                ellipsis is None and 
+                len(idx) != len(self._implicit_subsets)):
+                # dim will be None. Absent an ellipsis, the number of indices
+                # passed in should be the length of _implicit_subsets.
+                raise IndexError(
+                    "Index %s contains an invalid number of entries for "
+                    "component %s. Expected %s, got %s." 
+                    % (idx, self.name, len(self._implicit_subsets), len(idx)))
+            elif (not normalize_index.flatten and 
+                len(fixed) + len(sliced) > len(self._implicit_subsets)):
+                # dim is still None. Raise an error if the number of fixed
+                # and sliced indices exceeds the length of _implicit_subsets.
+                raise IndexError(
+                    "Index %s contains an invalid number of entries for "
+                    "component %s. Expected no more than %s, got %s." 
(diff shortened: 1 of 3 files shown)


pyomo/core/base/indexed_component.py
```diff
@@ -666,6 +666,27 @@ def _processUnhashableIndex(self, idx):
...
+                # If there is no ellipse and the index doesn't match the 
+                # component's dimension, raise an error.
+                # NOTE: Should this do something different for unflattened 
+                # indices?
+                (len(fixed) + len(sliced) > self.dim())):
+                # If an ellipsis is present and the index exceeds the dimension
+                # of the component, raise an error.
+                raise IndexError(
+                    "Index %s contains an invalid number of entries for "
+                    "component %s." % (idx, self.name))
```
qtothec, line 689: You probably want to tell them `got: 3, expected 2` or something.

pyomo/core/tests/unit/test_indexed_slice.py
```diff
@@ -548,5 +548,48 @@ def test_duplicate(self):
...
+            _slicer = m.b[:].c[:,:,:].x
+
+        with self.assertRaisesRegexp(
+            IndexError, 'Index .* contains an invalid number of '
+            'entries for component .*'):
+            _slicer = m.b[2, :].c[:,:,:].x
+            # Error not raised immediately because accessing c is deferred
+            # until iteration.
+            for var in _slicer:
+                pass
```
qtothec, line 584: This is fine, but if you want a single-line alternative, `list(_slicer)` ought to work.

pyomo/core/base/indexed_component.py
```diff
@@ -666,6 +666,27 @@ def _processUnhashableIndex(self, idx):
                 fixed[i - len(idx)] = val
 
         if sliced or ellipsis is not None:
+            if self.dim() is None:
+                # Assume that the right thing to do here is return
+                # an IndexedComponent_slice
```
qtothec, line 671: Huh? What case would cause this?

pyomo/core/base/indexed_component.py
```diff
@@ -666,6 +666,27 @@ def _processUnhashableIndex(self, idx):
...
+                # this behavior is allowed.
+                pass
+            elif ((ellipsis is None and len(idx) != self.dim()) or 
+                # If there is no ellipse and the index doesn't match the 
+                # component's dimension, raise an error.
+                # NOTE: Should this do something different for unflattened 
+                # indices?
+                (len(fixed) + len(sliced) > self.dim())):
+                # If an ellipsis is present and the index exceeds the dimension
+                # of the component, raise an error.
```
qtothec, line 686: Bit of devil's advocate: is there a use case of slicing where you'd slice a slice?

I personally cannot think of a good one right now.

pyomo/core/base/indexed_component.py
Robbybp: Yep good call, added.

pyomo/core/tests/unit/test_indexed_slice.py
Robbybp: Thanks, update.

pyomo/core/base/indexed_component.py
Robbybp: I think some uninstantiated sets in abstract models can have dimension None? It shows up in one of the tests, which is why I added it.

pyomo/core/base/indexed_component.py
Robbybp: I'm not sure. To achieve the functionality you'd presumably want by slicing a slice, I would slice a reference-to-slice. What is the implication of being able to slice a slice? And to be clear, you're talking about trying to run code like this?
```
m.v[:,2,:][:,'a'].fix(0)
```

pyomo/core/base/indexed_component.py
Robbybp: Also sets with inconsistent dimension can be created with dimension None. From `test_reference.py`:
```
        m.js = Set(initialize=[1, (2,3)], dimen=None)
```

blnicho:
I'm in favor of these changes but I'd like to hear from @jsiirola if there was a motivation behind the current behavior that we aren't aware of.

Robbybp:
I think whether this should be exceptional depends on your use case for slices. If you're using them as a convenient way to, say, fix initial conditions, then you probably want to throw an error if the slice index is invalid in any way. If you're using them to access data from components indexed by different sets, and do different things depending on whether a set has a particular index (say, copy values versus interpolate for the case of ContinuousSets with different discretizations), you might want to be able to pass a component a slice with an invalid index.
Maybe cases in which to fail should be controlled by flags somewhere. Regardless, I think that at least structurally invalid slice indices should fail by default.

To your point 3, I think that sets with `dimen == None` are rare and weird enough that having different behavior here is acceptable. 

For the case where `normalize_index.flatten == False`, would it be enough to just check that the number of indices, `len(fixed) + len(sliced)`, is consistent with `len(_implicit_subsets)` before checking if `dim is None`? I've implemented this and it seems to work in some simple test cases.

What is wrong with `_implicit_subsets` other than the name? Should there be a separate PR to replace it?

PR #1495, opened 2020-06-12
Bug fix in interior point/MUMPS memory reallocation logic

Robbybp:
## Fixes 
Fixes a bug in the Interior Point/MUMPS memory reallocation logic, where `_prev_allocation` was not being updated correctly after a symbolic factorization. This was only a problem because symbolic factorization is being performed for every numeric factorization attempt, and would only become apparent if multiple reallocations were necessary.

- Fix bug by updating `_prev_allocation` to be the max of ICNTL(23) and INFOG(16) after symbolic factorization
- Cast `_prev_allocation` as an int to accurately represent the info sent to MUMPS in ICNTL(23).
- Test to catch the erroneous behavior.

The test to catch this bug is very slow, about 3s. This is because the bug only became a problem if 2 or more reallocation iterations were necessary, so I needed to find an NLP for which this was the case. This is complicated by the fact that memory allocations via ICNTL(23) are rounded to the nearest MB, so I needed a KKT matrix whose actual memory requirement was at least 2 MB greater than the prediction. The NLP in the test file is the smallest/fastest I could find with this property.

I could switch to a test that only requires one iteration of reallocation, which could be made much faster, but would not cover the specific bug I encountered. Also up for debate whether this is more of a "MUMPS test" or an "Interior Point test." It's located in the latter's directory because that's how I encountered it. Suggestions appreciated on both of these counts, or on a smaller KKT matrix that requires 2 MB of reallocation.  

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/interior_point/interior_point.py
+++ b/pyomo/contrib/interior_point/interior_point.py
@@ -140,11 +140,13 @@ class InteriorPointSolver(object):
 
         self.linear_solver_logger = self.linear_solver.getLogger()
         self.linear_solve_context = LinearSolveContext(self.logger,
-                                                       self.linear_solver_logger,
-                                                       self.linear_solver_log_filename)
+                                        self.linear_solver_logger,
+                                        self.linear_solver_log_filename)
 
     def update_barrier_parameter(self):
-        self._barrier_parameter = max(self._minimum_barrier_parameter, min(0.5 * self._barrier_parameter, self._barrier_parameter ** 1.5))
+        self._barrier_parameter = max(self._minimum_barrier_parameter, 
+                                    min(0.5 * self._barrier_parameter, 
+                                        self._barrier_parameter ** 1.5))
 
     def set_linear_solver(self, linear_solver):
         """This method exists to hopefully make it easy to try the same IP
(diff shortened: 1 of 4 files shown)


jsiirola:
Note that this has a failing test under pypy on Travis:
```
FAIL: test_mumps (pyomo.contrib.interior_point.tests.test_realloc.TestReallocation)
  File "/home/travis/build/Pyomo/pyomo/pyomo/contrib/interior_point/tests/test_realloc.py", line 66, in test_mumps
    self.assertEqual(predicted, 12)
AssertionError: 11 != 12
```

Robbybp:
Noticed. I suspect this is due to a different version of Mumps/PyMumps being used, and will have to either modify my test or add a check for the expected version of Mumps.

blnicho:
Note: This PR will have conflicts with #1505 so we'll have to be careful with the merge

Robbybp:
> Note: This PR will have conflicts with #1505 so we'll have to be careful with the merge

Let me know if I should merge that branch

blnicho:
@Robbybp yes, #1505 has been merged into master so could you merge master into this PR and resolve the conflicts? You'll probably need to update some of your imports.

PR #1502, opened 2020-06-16
Deprecate GDP Reclassify Transformation in Favor of Using References

emma58:
## Summary/Motivation:

Previously, `TransformationFactory('gdp.bigm').apply_to(m.block)` and `TransformationFactory('gdp.bigm').apply_to(m, targets=m.block)` behaved differently because of the hack to expose the `indicator_var`s and other local variables to the writers. This is because the first automatically called the `gdp.reclassify` transformation whereas the second did not, since we have no way of knowing if someone expects their model to be completely transformed after using targets.

This PR deprecates `gdp.reclassify` and instead has the bigm and hull transformations create References to locally declared variables, which are stored on their own sub-block of the Disjunct's transformation Block. This exposes the local variables to the writers without ever changing the ctype of the `Disjunct`.

- Modifies `gdp.bigm` and `gdp.hull` to add References to all locally declared Vars on the Disjunct transformation block.
- Adds `assert_model_algebraic` function to gdp.util to take the place of the validation we were doing in `gdp.reclassify` to check that the model was indeed completely transformed.
- Moves `hull` disaggregated variables to their own sub-block of the Disjunct transformation block. (This is peripheral to this PR, but means we have complete control over names on the Disjunct transformation block again, which is nice.)

### Legal Acknowledgement

### DIFF
--- a/pyomo/gdp/plugins/gdp_var_mover.py
+++ b/pyomo/gdp/plugins/gdp_var_mover.py
@@ -23,7 +23,6 @@ from pyomo.core import TraversalStrategy, TransformationFactory
 from pyomo.core.base.indexed_component import ActiveIndexedComponent
 from pyomo.common.deprecation import deprecated
 from pyomo.common.modeling import unique_component_name
-from pyomo.common.config import ConfigBlock, ConfigValue
 
 from six import itervalues
 
--- a/pyomo/gdp/plugins/gdp_var_mover.py
+++ b/pyomo/gdp/plugins/gdp_var_mover.py
@@ -42,8 +42,8 @@ class HACK_GDP_Disjunct_Reclassifier(Transformation):
                 "variables local to each Disjunct during the transformation. "
                 "Validation that the model has been completely transformed "
                 "to an algebraic model has been moved to the "
-                "assert_model_algebraic function in gdp.util.",
-                version='5.6.10')
+                "check_model_algebraic function in gdp.util.",
+                version='TBD')
     def _apply_to(self, instance, **kwds):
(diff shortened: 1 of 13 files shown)


pyomo/gdp/plugins/gdp_var_mover.py
```diff
@@ -22,39 +22,13 @@
 from pyomo.core import TraversalStrategy, TransformationFactory
 from pyomo.core.base.indexed_component import ActiveIndexedComponent
 from pyomo.common.deprecation import deprecated
+from pyomo.common.modeling import unique_component_name
+from pyomo.common.config import ConfigBlock, ConfigValue
```
jsiirola, line 26: This import is not used

pyomo/gdp/plugins/gdp_var_mover.py
```diff
@@ -64,7 +38,13 @@ class HACK_GDP_Disjunct_Reclassifier(Transformation):
...
 
     """
-
+    @deprecated(msg="The gdp.reclasify transformation has been deprecated in "
+                "favor of the gdp transformations creating References to "
+                "variables local to each Disjunct during the transformation. "
+                "Validation that the model has been completely transformed "
+                "to an algebraic model has been moved to the "
+                "assert_model_algebraic function in gdp.util.",
+                version='5.6.10')
```
jsiirola, line 47: The next release will be 5.7 (not 5.6.10)

blnicho:
@jsiirola is this ready to merge?

jsiirola:
Yes.

PR #1526, opened 2020-06-20
Add community detection

rahuljoglekar47:
## Fixes # .

This PR adds a convenient way to perform Louvain community detection on a Pyomo model by automating the process of creating a NetworkX graph based on the model components.

## Summary/Motivation:

The community detection package allows users to obtain a map of the communities in a model based on Louvain community detection. The user is also given a wide variety of options in how this community detection is to be performed. Lastly, using a community map like the one the main function (detect_communities) generates could be very useful as an input into some existing algorithms.

- No changes are proposed for the already-existing repository

### Legal Acknowledgement


qtothec:
@rahuljoglekar47 Could I ask you to change the captions on the 8PP image to say more like "nodes are constraints, edges are shared/common variables", so that it's more intuitive what someone is looking at? I'll give a more detailed review later.

doc/OnlineDocs/contributed_packages/community.rst
```diff
@@ -0,0 +1,274 @@
...
+detection package. Both of these packages can be installed at the following URLs (respectively):
+
+https://pypi.org/project/networkx/
+
+https://pypi.org/project/python-louvain/
+
+The pip install and conda install commands are included below as well:
+
+
+>>> pip install networkx
```
ZedongPeng, line 85: I think `.. code::` or `commands::` should be added here to format the commands. `>>>` could be removed here.

doc/OnlineDocs/contributed_packages/community.rst
```diff
@@ -0,0 +1,274 @@
...
+>>> conda install -c conda-forge python-louvain
+
+Usage Examples
+--------------
+
+Let's take the following model as an example (taken from `Allman et al, 2019`_) (the necessary
+imports to run all of the code examples are included below):
+
+.. _Allman et al, 2019: https://doi.org/10.1007/s11081-019-09450-5
+
```
ZedongPeng, line 97: `.. code::` is also needed here. Otherwise, the following code will be commented out.

ZedongPeng: Line 174 too.

doc/OnlineDocs/contributed_packages/community.rst
```diff
@@ -0,0 +1,274 @@
...
+    >>> c3 c5 {'weight': 2}
+    >>> c4 c5 {'weight': 2}
+    >>> Adjacency List:
+    >>> c1 c2 c3
+    >>> c2 c3
+    >>> c3 c4 c5
+    >>> c4 c5
+    >>> c5
+
+Main Function List
```
ZedongPeng, line 254: I am not sure if it is better to use `.. autoclass:: ` here. An example can be found in [gdpopt.rst](https://github.com/Pyomo/pyomo/blob/master/doc/OnlineDocs/contributed_packages/gdpopt.rst)

doc/OnlineDocs/contributed_packages/community.rst
qtothec: `>>>` is relevant in the context of a `.. doctest::` block to distinguish input from expected output.

doc/OnlineDocs/contributed_packages/community.rst
qtothec: @ZedongPeng This is a reference definition in RST. This is valid. I slightly prefer using footnotes instead, but that's stylistic.

doc/OnlineDocs/contributed_packages/community.rst
qtothec: Might be `.. automodule::` instead here.

doc/OnlineDocs/contributed_packages/community.rst
```diff
@@ -0,0 +1,274 @@
...
+The community detection package allows users to obtain a dictionary of the communities in a Pyomo model. The package
+takes in a model, organizes the model components into a graph of nodes and edges, then uses Louvain
+community detection (`Blondel et al, 2008`_) to determine the communities that exist within the model.
+
+.. _Blondel et al, 2008: https://dx.doi.org/10.1088/1742-5468/2008/10/P10008
+
+In graph theory, a community is defined as a subset of nodes that have a greater degree of connectivity within
+themselves than they do with the rest of the nodes in the graph. In the context of Pyomo models, a community
+represents a subproblem within the overall optimization problem. Identifying these subproblems and then solving them
+independently can save computational work as compared with trying to solve the entire model at once. Thus, it
```
qtothec, line 19:
```suggestion
independently can save computational work compared with trying to solve the entire model at once. Thus, it
```

doc/OnlineDocs/contributed_packages/community.rst
```diff
@@ -0,0 +1,274 @@
...
+    `detect_communities(model[, type_of_community_map, with_objective, weighted_graph, random_seed])`
+
+The main graph features the user can specify are the type of community map, whether the graph is weighted or
+unweighted, and whether the objective function(s) is included in the graph generation. Below, the significance
+and reasoning behind including each of these options are explained in greater depth.
+
+Type of Community Map
+    In this package's main function (detect_communities), the user can select 'b', 'c', or 'v' as an input for
+    the 'type_of_community_map' argument, and these result in a community map based on a bipartite graph, a constraint
+    node graph, or a variable node graph (respectively).
```
qtothec, line 36: Can I suggest instead `nodes_are=['variables'|'constraints'|'variables_and_constraints']`?

I had originally thought `both` instead of `variables_and_constraints`, but it is conceivable to add more classes later on.

doc/OnlineDocs/contributed_packages/community.rst
```diff
@@ -0,0 +1,274 @@
...
+represents a subproblem within the overall optimization problem. Identifying these subproblems and then solving them
+independently can save computational work as compared with trying to solve the entire model at once. Thus, it
+can be very useful to know the communities that exist in a model.
+
+The manner in which the graph of nodes and edges is constructed from the model directly affects the community
+detection. Thus, this package provides the user with a lot of control over the construction of the graph. The
+parameters of this function that are used for this graph construction and subsequent community detection
+are shown below.
+
+    `detect_communities(model[, type_of_community_map, with_objective, weighted_graph, random_seed])`
```
qtothec, line 27:
```suggestion
.. autofunction:: pyomo.contrib.community_detection.detection.detect_communities
```

doc/OnlineDocs/contributed_packages/community.rst
```diff
@@ -0,0 +1,274 @@
...
+    the user intends to use the community detection information. For example, if a user plans on feeding that
+    information into an algorithm, the algorithm may be better suited to the communities detected in a weighted
+    graph (or vice versa).
+
+With/Without Objective in the Graph
+    This argument determines whether the objective function(s) will be included when creating the graphical
+    representation of the model and thus whether the objective function(s) will be included in the community map.
+    Some models have an objective function that contains so many of the model variables that it obscures potential
+    communities within a model. Thus, it can be useful to call `detect_communities(model, with_objective=False)`
+    on such a model to see whether isolating the other components of the model provides any new insights.
```
qtothec, line 71: Would it make sense to move this description into the function docstring itself, given that we suggest `autofunction` above?

Or, have a condensed version in docstring, and leave this here.

doc/OnlineDocs/contributed_packages/community.rst
```diff
@@ -0,0 +1,274 @@
...
+
+https://pypi.org/project/python-louvain/
+
+The pip install and conda install commands are included below as well:
+
+
+>>> pip install networkx
+>>> pip install python-louvain
+>>> conda install -c anaconda networkx
+>>> conda install -c conda-forge python-louvain
```
qtothec, line 88: Fwiw, I agree with Zedong--these `>>>` don't need to be here. They're used for Python doctest inputs. You don't have a doctest here, just a code block.

doc/OnlineDocs/contributed_packages/community.rst
```diff
@@ -0,0 +1,274 @@
...
+    >>> # Output:
+    >>> {0: ([<pyomo.core.base.constraint.SimpleConstraint object at 0x0000028DA74BB588>,
+    >>>       <pyomo.core.base.constraint.SimpleConstraint object at 0x0000028DA74BB5F8>],
+    >>>      [<pyomo.core.base.var.SimpleVar object at 0x0000028DA74BB3C8>,
+    >>>       <pyomo.core.base.var.SimpleVar object at 0x0000028DA74BB438>]),
+    >>>  1: ([<pyomo.core.base.constraint.SimpleConstraint object at 0x0000028DA74BB668>,
+    >>>       <pyomo.core.base.constraint.SimpleConstraint object at 0x0000028DA74BB6D8>,
+    >>>       <pyomo.core.base.constraint.SimpleConstraint object at 0x0000028DA74BB748>],
+    >>>      [<pyomo.core.base.var.SimpleVar object at 0x0000028DA74BB4A8>,
+    >>>       <pyomo.core.base.var.SimpleVar object at 0x0000028DA74BB518>])}
```
qtothec, line 132: Don't use `>>>` for output. See https://www.sphinx-doc.org/en/master/usage/extensions/doctest.html for how to setup and run a doctest.

doc/OnlineDocs/contributed_packages/community.rst
```diff
@@ -0,0 +1,274 @@
...
+    >>>
+    >>> # As we did before, we will use the pos argument to create a consistent graph layout
+    >>>
+    >>> middle_fig, _ = visualize_model_graph(model, type_of_graph='v', type_of_community_map='b',
+    >>> random_seed=seed, pos=pos)
+    >>> plt.show()
+    >>>
+    >>> right_fig, _ = visualize_model_graph(model, type_of_graph='v', type_of_community_map='v',
+    >>> random_seed=seed, pos=pos)
+    >>> plt.show()
```
qtothec, line 189: Would recommend making these doctests.

doc/OnlineDocs/contributed_packages/index.rst
```diff
@@ -22,8 +22,9 @@ Contributed packages distributed with Pyomo:
    mcpp.rst
    mindtpy.rst
    satsolver.rst
+   community.rst
 
 Contributed packages distributed independently of Pyomo, but accessible
 through ``pyomo.contrib``:
 
-* `pyomo.contrib.simplemodel <http://pyomocontrib-simplemodel.readthedocs.io/en/latest/source.html>`_
+* `pyomo.contrib.simplemodel <http://pyomocontrib-simplemodel.readthedocs.io/en/latest/source.html>`_
```
qtothec, line 30: I don't see the difference. Did you add a windows EOL here?

pyomo/contrib/community_detection/community_graph.py
```diff
@@ -0,0 +1,343 @@
...
+    with_objective: bool, optional
+        a Boolean argument that specifies whether or not the objective function is included in the graph; the
+        default is True
+    weighted_graph: bool, optional
+        a Boolean argument that specifies whether a weighted or unweighted graph is to be created from the Pyomo
+        model; the default is True (type_of_graph='b' creates an unweighted graph regardless of this parameter)
+
+    Returns
+    -------
+    bipartite_model_graph/collapsed_model_graph: NetworkX graph
```
qtothec, line 40:
```suggestion
    nx_graph: nx.Graph
```

pyomo/contrib/community_detection/community_graph.py
```diff
@@ -0,0 +1,343 @@
...
+                if weighted_graph:
+                    new_edge_weights = {edge: edge_weight_dict.get(edge, 0) + 1 for edge in edges_between_variables}
+                    edge_weight_dict.update(new_edge_weights)
+                else:
+                    edge_set.update(edges_between_variables)
+
+            # If the node is not a constraint/objective, then it is a variable and we want it as
+            # a node in our variable node graph
+            else:
+                collapsed_model_graph.add_node(node)
```
qtothec, line 193: First thought: these should be a separate private functions.

Second thought: does networkx have a functionality for this type of graph operation? I would presume so. If so, I would use that.

pyomo/contrib/community_detection/community_graph.py
```diff
@@ -0,0 +1,343 @@
...
+
+    # Log important information with the following logger function
+    _event_log(model, collapsed_model_graph, set(constraint_variable_map), type_of_graph, with_objective)
+
+    # Return the collapsed NetworkX graph, the dictionary of node numbers mapped to their respective Pyomo
+    # components, and the map of constraints to the variables they contain
+    return collapsed_model_graph, number_component_map, constraint_variable_map
+
+
+def _event_log(model, model_graph, constraint_set, type_of_graph, with_objective):
```
qtothec, line 221: I would move this into a separate utility module, since it's not business logic.

pyomo/contrib/community_detection/detection.py
```diff
@@ -0,0 +1,466 @@
...
+
+    This function takes in a Pyomo optimization model and organizes the variables and constraints into a graph of nodes
+    and edges. Then, by using Louvain community detection on the graph, a dictionary (community_map) is created, which
+    maps (arbitrary) community keys to the detected communities within the model.
+
+    Parameters
+    ----------
+    model: Block
+         a Pyomo model or block to be used for community detection
+    type_of_community_map: str, optional
```
qtothec, line 41: see my earlier suggestion on naming. You can use Pycharm's refactoring tool if needed.

pyomo/contrib/community_detection/detection.py
```diff
@@ -0,0 +1,466 @@
...
+        logger.warning("Community detection found that with the given parameters, the model could not be decomposed - "
+                       "only one community was found")
+
+    # Return community_map, which has integer keys that now map to a tuple of two lists
+    # containing Pyomo modeling components
+    return community_map
+
+
+def visualize_model_graph(model, community_map=None, type_of_graph='c', with_objective=True, weighted_graph=True,
+                          random_seed=None, type_of_community_map=None, pos=None):
```
qtothec, line 169: Why not just have usage:
```
community_map = detect_communities(pyomo_model, ...)
visualize_communities(community_map)
```

pyomo/contrib/community_detection/detection.py
```diff
@@ -0,0 +1,466 @@
...
+        community_map_type = node_name_map[type_of_community_map]
+        plot_title = "%s graph - colored using %s community map" % (graph_type, community_map_type)
+    plt.title(plot_title)
+
+    # Return the figure and the position dictionary used for the graph layout
+    return fig, pos
+
+
+def stringify_community_map(model=None, community_map=None, type_of_community_map='c', with_objective=True,
+                            weighted_graph=True, random_seed=None):
```
qtothec, line 384: Similar question to `visualize_communities`

qtothec (dismissed):
Commendable work. Few comments below for discussion.

pyomo/contrib/community_detection/detection.py
rahuljoglekar47: I'm not sure what you mean - are you saying this function should only take in community_map as an argument?

doc/OnlineDocs/contributed_packages/index.rst
rahuljoglekar47: Hmm, I don't think so (or at the very least it shouldn't be there) - I was playing around with this file when git kept telling me I had an unresolved conflict (which I later realized was normal)

pyomo/contrib/community_detection/detection.py
bernalde: Hi Qi, this actually has to do with the fact that the visualization depends not only on the communities that you derived (which are based on a kind of graph where you applied the decomposition method `type_of_community_map`) but also on the graph you want to plot them into `type_of_graph`. You need the model to derive the original graph shape for plotting.

pyomo/contrib/community_detection/detection.py
bernalde: Same as above, without the original model, we have no way of knowing what the object's actually mean in strings.

pyomo/contrib/community_detection/community_graph.py
bernalde: You are absolutely right. This was an oversight from me. There is a function `weighted_projected_graph` in NetworkX that does this exactly (see https://networkx.github.io/documentation/networkx-1.10/reference/generated/networkx.algorithms.bipartite.projection.weighted_projected_graph.html#networkx.algorithms.bipartite.projection.weighted_projected_graph). I will discuss this with Rahul and try to change this implementation

pyomo/contrib/community_detection/detection.py
qtothec: So, what you're telling me is that `detect_communities(pyomo_model,...)` should return a type `CommunityMap` inheriting from `dict` that includes methods for visualization and a `__str__` override?

rahuljoglekar47: `detect_communities` should return a standard Python `dict`, which is named `community_map`. Visualizing and "stringifying" this `community_map` is done with the functions `visualize_model_graph` and `stringify_community_map` (respectively).

doc/OnlineDocs/contributed_packages/community.rst
rahuljoglekar47: Yeah, the intention was for the docstring to be a condensed version of the longer explanation of the community detection in the rst file. But let me know if you're saying I should modify the docstring based on the descriptions in the rst file or if you're saying the docstring is fine as it is.

pyomo/contrib/community_detection/detection.py
qtothec: @rahuljoglekar47 Apologies, I should be more clear. You're handing the user back a community map from `detect_communities()`. Now, if all the relevant information associated with a community map fits neatly as a python base type (e.g. `dict`), then that's all well and good. However, David is saying that there's additional metadata associated with the communities (i.e. its type). So, under OOP principles, you're no longer just returning a `dict`. You're returning some new object type that has the original `dict` information, but also additional metadata attributes, hence, a new class type.

bernalde: The reason why we thought of `detect_communities(...)` to return a `dict` is that it is the simplest output, independent of the `type_of_graph` selected. We have not considered the general case where we want to return that more complicated object. Let us look into it.

qtothec:
@rahuljoglekar47 You might want to take a look at those image captions again. Pretty sure nodes in a constraint community map are supposed to be constraints, with shared variables as the edges?

rahuljoglekar47:
> @rahuljoglekar47 You might want to take a look at those image captions again. Pretty sure nodes in a constraint community map are supposed to be constraints, with shared variables as the edges?

In the case of  "Variable graph - colored with Constraint community map", this means we take a variable graph (nodes as variables) and then use a community map that was generated by performing community detection on a constraint-node graph. In short, the "coloring" is based on the community map and the actual graph that we use is supposed to be independent of that community map - so we could color a given bipartite/variable/constraint graph with any type of community map (including a bipartite, variable, or constraint community map).

qtothec:
> > @rahuljoglekar47 You might want to take a look at those image captions again. Pretty sure nodes in a constraint community map are supposed to be constraints, with shared variables as the edges?
> 
> In the case of "Variable graph - colored with Constraint community map", this means we take a variable graph (nodes as variables) and then use a community map that was generated by performing community detection on a constraint-node graph. In short, the "coloring" is based on the community map and the actual graph that we use is supposed to be independent of that community map - so we could color a given bipartite/variable/constraint graph with any type of community map (including a bipartite, variable, or constraint community map).

Ah, I see. My mistake. If I recall correctly, Andrew Allman had also included the constraint graph in his presentations, hence my confusion.

pyomo/contrib/community_detection/detection.py
```diff
@@ -168,7 +170,225 @@ def detect_communities(model, type_of_community_map='constraint', with_objective
...
+    def keys(self):
+        return self.community_map.keys()
+
+    def values(self):
+        return self.community_map.values()
+
+    def items(self):
+        return self.community_map.items()
+
+    def draw_graph(self, type_of_graph='constraint', with_objective=True, weighted_graph=True, random_seed=None, pos=None):
```
qtothec, line 205: Why have all of these parameters when `self` should have the information?

pyomo/contrib/community_detection/detection.py
```diff
@@ -168,7 +170,225 @@ def detect_communities(model, type_of_community_map='constraint', with_objective
...
+
+        assert random_seed is None or (type(random_seed) == int and random_seed >= 0), \
+            "Invalid value for random_seed: 'random_seed=%s' - random_seed must be a non-negative integer" % random_seed
+
+        # No assert statement for pos; the NetworkX function can handle issues with the pos argument
+
+        # Use the generate_model_graph function to create a NetworkX graph of the given model (along with
+        # number_component_map and constraint_variable_map, which will be used to help with drawing the graph)
+        model_graph, number_component_map, constraint_variable_map = generate_model_graph(
+            self.model, type_of_graph=type_of_graph, with_objective=with_objective, weighted_graph=False)
```
qtothec, line 267: Should the networkx representation be an attribute of the CommunityMap object?

pyomo/contrib/community_detection/detection.py
```diff
@@ -168,7 +170,225 @@ def detect_communities(model, type_of_community_map='constraint', with_objective
...
+        # Return the figure and pos, the position dictionary used for the graph layout
+        return fig, pos
+
+    def stringify(self):
+        """
+        This function takes in a community map of Pyomo components and returns the same community map but with the strings
+        of the Pyomo components. Alternatively, if the user does not provide a community map, this function can take in
+        a model and then create a community map (using the function detect_communities). Then, the same process of
+        "stringifying" the community map is performed, again resulting in a community map but with the strings
+        of the Pyomo components.
```
qtothec, line 374: Documentation seems a bit out of date, but you might choose to update it once the API is solidified.

On this function, does it make sense to override `__str__`?

pyomo/contrib/community_detection/detection.py
```diff
@@ -168,7 +170,225 @@ def detect_communities(model, type_of_community_map='constraint', with_objective
...
+        for key in self.community_map:
+            str_community_map[key] = ([str(component) for component in self.community_map[key][0]],
+                                      [str(component) for component in self.community_map[key][1]])
+
+        # Return str_community_map, which is identical to community_map except it has the strings of all of the Pyomo
+        # components instead of the actual components
+        return str_community_map
 
 
 def visualize_model_graph(model, community_map=None, type_of_graph='constraint', with_objective=True,
```
qtothec, line 394: Again, if the community map includes some of the parameter information already, why input it again here?

pyomo/contrib/community_detection/detection.py
rahuljoglekar47: In the case of the draw_graph function, the CommunityMap object only tells us how to color the nodes according to their communities. But we can control whether we draw a constraint-node graph, a variable-node graph, or a bipartite graph (this is what the 'type_of_graph' parameter is for). The other parameters are similarly involved in the specifics of the graph that we draw (whether we treat the objective as a node and whether we want to provide a layout with 'pos').

I forgot to remove 'random_seed' as a parameter (this is used for creating community maps), so maybe this led to some confusion as well - sorry about that!

pyomo/contrib/community_detection/detection.py
rahuljoglekar47: The networkX graph generated in those lines may be a little different from the one used to create the CommunityMap object (based on my response above) - however, I can add the networkX graph as an attribute and then also add a check before these lines to avoid creating the same networkX graph as we already made to create the CommunityMap object

pyomo/contrib/community_detection/detection.py
rahuljoglekar47: Yeah, sorry about the delay with the documentation. As for your second point, I was considering the following:
- remove the above function and put that code into the __str__ function so that when the user prints the CommunityMap, they will see a user-friendly community map (a bunch of string variable names in the dict)
- for __repr__, return the community_map attribute of the CommunityMap object (which would be the community map of actual Pyomo components)

Not sure if this answers your question, but I haven't made any changes yet as I'm unsure what would be best - let me know what you think!

pyomo/contrib/community_detection/community_graph.py
```diff
@@ -56,11 +60,12 @@ def generate_model_graph(model, type_of_graph, with_objective=True, weighted_gra
...
     if with_objective:
         component_number_map = ComponentMap((component, number) for number, component in enumerate(
-            model.component_data_objects(ctype=(Constraint, Var, Objective), active=True, descend_into=True,
+            model.component_data_objects(ctype=(Constraint, Var, Objective), active=use_only_active_components,
+                                         descend_into=True,
                                          sort=SortComponents.deterministic)))
     else:
         component_number_map = ComponentMap((component, number) for number, component in enumerate(
-            model.component_data_objects(ctype=(Constraint, Var), active=True, descend_into=True,
+            model.component_data_objects(ctype=(Constraint, Var), active=use_only_active_components, descend_into=True,
```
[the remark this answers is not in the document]
rahuljoglekar47: Oh, thanks for pointing that out! I changed it so the user can only specify True or None

rahuljoglekar47:
@bernalde @ZedongPeng @qtothec @jsiirola I believe all of the code is complete - if possible, it would be great if anyone could take a look at the `generate_structured_model` method of the CommunityMap class to ensure that the logic is sound - thanks!

bernalde:
The tests that are failing here are coming from the plot tests in `pyomo.contrib.parmest.tests.test_parmest.parmest_object_Tester_RB`. Even after merging master, these are still failing.
I went into the commit and it merges the changes to `pyomo/contrib/parmest/graphics.py`, therefore I would like to raise the warning about those tests.

bernalde:
I cannot make it to today's developers' meeting (gotta meet Ignacio), but I would like to raise some questions regarding the test failures in this PR:
1. The older versions of Python fail since the tests are designed as asserts from an expected output given the deterministic behavior of an underlying Python library (Python-louvain). Even after fixing the random seed, the underlying library does not yield the same results (we added an extra test to verify this). What can we do? Have our test outputs depend on what the library answers (testing that the package does what it is expected to do given a library output) or having Python version dependent tests?
2. The newer versions of Python fail the tests since matplotlib is now taking more time to load. How can we solve this?
Thanks and talk to you next week!

jsiirola:
@bernalde:
- Re (1):  I think I would prefer that the Pyomo code is verified to be correct, given the behavior of the TPL (python-louvain).  Python version-specific baselines tend to be *very* fragile.
- Re (2): This test is catching that matplotlib is being imported when a user runs `import pyomo.environ`.  This is indicative of an unconditional import of matplotlib in a community_detection module that is automatically imported by `pyomo.environ`.  Looking through the code, that is what you are doing (you copied code over from `parmest.graphics`, but the difference is that `parmest.graphics` is not automatically imported by `pyomo.envioron`.  That said, I have gone in and extended the deferred import system to better handle deferred import of submodules (like `matplotlib.pyplot`).  See #1683.  Once that is merged, you can replace the matplotlib imports in `detection.py` with:
    ```python
    from pyomo.common.dependencies import matplotlib, matplotlib_available
    plt = matplotlib.pyplot
    ```

pyomo/contrib/community_detection/detection.py
```diff
@@ -0,0 +1,643 @@
...
+                    if not_found and node in numbered_community_map[community_key][position]:
+                        color_list.append(community_key)
+                        not_found = False
+
+            # Note - there is no strong reason to choose spring layout; it just creates relatively clean graphs
+            if pos is None:  # The case where the user has not provided their own layout
+                pos = nx.spring_layout(model_graph)
+
+        # Define color_map
+        color_map = cm.get_cmap('viridis', len(numbered_community_map))
```
bernalde, line 434: See comment above

pyomo/contrib/community_detection/detection.py
```diff
@@ -0,0 +1,643 @@
...
+
+logger = getLogger('pyomo.contrib.community_detection')
+
+# Attempt import of louvain community detection package
+community_louvain, community_louvain_available = attempt_import(
+    'community', error_message="Could not import the 'community' library, available via 'python-louvain' on PyPI.")
+
+# Import matplotlib
+plt = matplotlib.pyplot
+cm = matplotlib.cm
```
bernalde, line 32: Remove this import with 
`color_map = plt.cm.get_cmap('viridis', len(numbered_community_map))`
which is done here 
https://github.com/Pyomo/pyomo/blob/2a09bf8b6b6f364a6d7a37a8f45b8079989356f2/pyomo/contrib/parmest/graphics.py#L162

pyomo/environ/__init__.py
```diff
@@ -9,9 +9,11 @@
 #  ___________________________________________________________________________
 
 import sys as _sys
+
```
bernalde, line 12: I don't know why this and all the changes to this file (except line 63) were introduced (I want to blame Pycharm), but let's stick to what master says in terms of this file. The only change that should appear is the addition of `'pyomo.contrib.community_detection',` to `_optional_packages`

pyomo/contrib/community_detection/detection.py
```diff
@@ -0,0 +1,643 @@
...
+        nx.draw_networkx_nodes(model_graph, pos, nodelist=node_list, node_size=40, cmap=color_map,
+                               node_color=color_list)
+        nx.draw_networkx_edges(model_graph, pos, alpha=0.5)
+
+        # Make the main title
+        graph_type = type_of_graph.capitalize()
+        community_map_type = self.type_of_community_map.capitalize()
+        main_graph_title = "%s graph - colored using %s community map" % (graph_type, community_map_type)
+
+        main_font_size = 14
```
bernalde, line 447: I don't think it is a good practice to fix font sizes here. Consider removing this and the one on line 457.

doc/OnlineDocs/contributed_packages/index.rst
```diff
@@ -23,8 +23,9 @@ Contributed packages distributed with Pyomo:
    mindtpy.rst
    satsolver.rst
    sensitivity_toolbox.rst
+   community.rst
 
 Contributed packages distributed independently of Pyomo, but accessible
 through ``pyomo.contrib``:
 
-* `pyomo.contrib.simplemodel <http://pyomocontrib-simplemodel.readthedocs.io/en/latest/source.html>`_
+* `pyomo.contrib.simplemodel <http://pyomocontrib-simplemodel.readthedocs.io/en/latest/source.html>`_
```
bernalde, line 31: I can't see the difference here, but it should reflect what is happening in master.

bernalde (changes requested):
We are encountering issues with matplotlib and its import time. In order to delay the import of that package and avoid problems during doctesting and overall testing, let's follow what is done in the parameter estimation contributed package.
A great example is this function https://github.com/Pyomo/pyomo/blob/2a09bf8b6b6f364a6d7a37a8f45b8079989356f2/pyomo/contrib/parmest/graphics.py#L190
where they import the whole matplotlib and show the plots if no filename is specified to save the figures. We should transfer our visualize function to be like this, avoiding importing matplotlib.pyton in the .rst file and calling plt.show() (which I believe is what is making the import time of matplotlib explode).

bernalde:
If possible I would like to ask for some advice from the other developers. The tests that we have are failing because of matplotlib related issues. We tried to use exactly the same import statements as in `parmest` but still could get it running correctly. Any thoughts? @jsiirola @michaelbynum @blnicho

michaelbynum:
@bernalde It is not immediately clear to me why these tests are not passing. The failing tests appear to be tests from parmest, which is clearly not modified by this PR. Can you make sure your `add_community_detection` branch is up to date (i.e., merge Pyomo `master` into rahuljoglekar47 `add_community_detection`)?

jsiirola:
#1729 has a fix that should resolve the matplotlib/parmest test failures in this PR.

pyomo/contrib/community_detection/detection.py
```diff
@@ -0,0 +1,655 @@
...
+
+logger = getLogger('pyomo.contrib.community_detection')
+
+# Attempt import of louvain community detection package
+community_louvain, community_louvain_available = attempt_import(
+    'community', error_message="Could not import the 'community' library, available via 'python-louvain' on PyPI.")
+
+# Import matplotlib
+if matplotlib_available:
+    plt = matplotlib.pyplot
```
jsiirola, line 32: With the recent change to `pyomo.common.dependencies`, you should revert f342de2.  That will prevent matplotlib from being unconditionally imported.

The way the deferred imports work, if you cast the `_available` flag to a `bool` (i.e., in an `if`), then the import is attempted.  So, you want to avoid checking if an optional dependency is available at the module scope (i.e., when the module is imported).  The deferred import system is designed so that in most cases you just grab the deferred module and use it like you normally would - it will be imported the first time it is needed.

pyomo/contrib/community_detection/tests/test_detection.py
```diff
@@ -30,8 +30,8 @@
 from pyomo.solvers.tests.models.LP_inactive_index import LP_inactive_index
 from pyomo.solvers.tests.models.SOS1_simple import SOS1_simple
 
-if matplotlib_available:
-    matplotlib.use('Agg')  # added to avoid $DISPLAY errors on Travis (from parmest)
+matplotlib.use('Agg')  # added to avoid $DISPLAY errors on Travis (from parmest)
```
jsiirola, line 33: Ironically, this still needs to be guarded by:
```python
if matplotlib_available:
    matplotlib.use('Agg')  # added to avoid $DISPLAY errors on Travis (from parmest)
```

Because we are in the test harness, it is OK to trigger the import check for optional modules.  Further, we need to guard the call to `.use`, as when that is parsed, the `DeferredImportModule` (i.e., `matplotlib`) will trigger the import and then if `matplotlib` is not available, the attempt to return the `.use` attribute will generate an exception.

rahuljoglekar47: Okay I see - making the changes now!

jsiirola:
Tests are now passing (the remaining test failure is due to an intermittent GitHub infrastructure failure unrelated to this PR).

@rahuljoglekar47: note that PR #1739 has also been merged into master.

bernalde:
I tried contacting @qtothec to get his approval of this PR but was unsuccessful. Since his comments were addressed by Rahul some time ago and are included in the latest version of this PR, I would be in favor of bypassing his review and merging this as we had agreed during our call @jsiirola

jsiirola (approved):
I think this is reasonable.  There are a number of things that can be improved in future iterations, including:
- @bernalde comment about embedded constants (font sizes)
- testing of the plotting capability
- conversion of the structured model generation to leverage `Reference()` and avoid duplication of variables / constraints.

blnicho (approved):

PR #1530, opened 2020-06-24
Resolve table formatting with unicode character data

jsiirola:
## Summary/Motivation:
The `tabular_writer` does not produce aligned columns in Python 2.7 when provided unicode characters in the table data.  This updates the `tabular_writer` so that strings are properly converted to unicode if necessary so that the resulting tables are well-aligned.

This bug was identified by @qtothec as part of PR #1507.

- Encode python 2.7 strings as unicode (UTF-8) before computing the string length so the resulting table is correctly formatted.
- Add a test to verify this behavior

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/misc.py
+++ b/pyomo/core/base/misc.py
@@ -178,15 +178,15 @@ def sorted_robust(arg):
         return sorted(arg, key=_robust_sort_keyfcn())
 
 
-def _safe_to_str(obj):
+def _to_ustr(obj):
     if not isinstance(obj, string_types):
         try:
             obj = str(obj)
         except:
-            return "None"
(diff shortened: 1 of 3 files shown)


jsiirola:
@qtothec: good point.  I was working with a "minimal change" approach, but you are correct: `str(obj)` could easily return a Python 2.7 `str` with embedded Unicode characters.  The safe thing is to convert all strings.  I have updated the PR, renaming a few things to make it more obvious what is going on, and caught another edge case where we needed to call `_to_ustr`.

Assuming this is merged, then we can include it in #1507 and go back to being whitespace sensitive (which in general is a good thing to verify in our testing).

PR #1541, opened 2020-07-01
Construct ComponentUID with wildcard

Robbybp:
## Summary/Motivation:
I'd like a way to construct ComponentUID objects from ComponentData objects with certain indices, or sets, specified as wildcards. I'd use these to identify components on different models which are identically named, other than the indices in a certain location, specified here by the (1-dimensional) set corresponding to that location. These CUIDs also give a nice way of getting a name for a slice without attaching it to a model as a reference.

- Add `wildcard_set` argument to `ComponentUID` constructor. Everywhere an index of this set appears in the component, that index should be replaced with a wildcard `*` in the ComponentUID constructed
- Simple tests
- `get_location_of_coordinate_set` function, which I copied from `idaes.core.util.dyn_utils`. Not sure where this function should go. I have tests for it I can add if we decide to move forward with this PR.

Wasn't sure what's happening when `cuid_buffer is not None` in `_generate_cuid`, so I didn't add any code for identifying the wildcard set here. I assume the same logic would be necessary, i.e. identify the wildcard location and send to `_partial_cuid_from_index`.

If there is a better way to do what I describe above (identifying nearly identically-named components on different models) without messing with `ComponentUID`, I'm all ears. Feedback appreciated.

### Legal Acknowledgement


qtothec:
Is CUID generation still expensive, or did that change?

Robbybp:
@qtothec No idea. Was/is it more expensive than you would expect traveling up the block hierarchy to be?

jsiirola:
@Robbybp I am a little unclear as to the intent of this PR.  Can you provide some examples?  In particular, how is this different from iterating over a slice and generating CUIDs?

As to @qtothec's question: yes: until PR #1228 is merged, generating a CUID from an object is "expensive" (can involve linear scans over indexing sets).

Robbybp:
@jsiirola Say I have two models that are nearly identical. The only difference is that one set (could be extended to multiple sets) has different values between the two models. Call this set time. Given a ComponentData in the first model at some time point, I want to access a ComponentData in the second model at some time point. The best way I could think to do this was with CUIDs. The problem is that I can't create a CUID from a component with a time point from the other model. I may not even know what these time points are. So I'd like to create a CUID with a wildcard in place of the time index. This can then be used to access the component(s) I want on both models. 

If I could create a CUID from a slice and have it generate a CUID with corresponding wildcards, that would be ideal, but I don't think that's possible. I think creating a CUID from a reference-to-slice will just have a name like `IndexedVar:**`, which is not what I want.

Robbybp:
As an example of use, I have a data structure to hold plant simulation data in CAPRESE. The data could come from different models, and thus have different time indices. What I need to do is, for a reference to time-only slice `_slice`, get a CUID that can be used to identify "the same" components on different models. The core functionality I need, that is not currently available, is the following function:
```python
def cuid_from_timeslice(_slice, time):
    t0 = time.first()
    compdata = _slice[t0]
    return ComponentUID(compdata, wildcard_set=time)
```
The alternative to this PR is to parse `compdata.name` to construct a string that I can use to construct the exact same CUID that I am constructing here. This string might look like: `fs.MB.gas_phase.properties[*,0.5].mole_frac['CH4']`.
To get the values I need for each model, I run the following code:
```
cuid = cuid_from_timeslice(_slice, time)
new_data = [comp.value for comp in cuid.list_components(plant_model)]
data_dict[str(cuid)].extend(new_data)
```
Here, `list_components` will work regardless of whether `_slice` belongs to `plant_model`. What I really do is generate these CUIDs when I initialize my plant simulation data structure, then use them to access their components on whatever model I want to update it from.

Let me know what questions there are.

Robbybp:
@jsiirola @carldlaird Added a paraphrased example of how I'm using this.

Robbybp:
Todo on this PR:
- [ ] Choose a reasonable name for `get_location_of_coordinate_set`
- [ ] Add tests for this function
- [ ] What needs to be done when `cuid_buffer is not None`? I am inclined to fail if `cuid_buffer` and `wildcard_set` are both used, but only because I have not idea what `cuid_buffer` is doing.

blnicho:
@Robbybp is going to rework this functionality and open a new PR

PR #1546, opened 2020-07-06
Pynumero: set AMPLFUNC before loading NL files

### LINKED ISSUE #1215: Pynumero External Function Support
Pynumero doesn't automatically include the AMPLExternalFunctions by setting the AMPLFUNC environment variable.  I can set AMPLFUNC before running Pyomo to make it work, but then stuff in AMPLFUNC is duplicated when solving the problem.  This can cause trouble, at least it causes solvers to print  warnings about duplicated functions.

eslickj:
### Motivation
This adds setting the AMPLFUNC environment variable to Pynumero.  Since Pynumero is implemented as a shared library, setting environment variables on Windows is a bit tricky, so there is a call to ```_wputenv_s()``` in msvcrt on Windows.  I think for Python 3.0 and on string encoding is unicode by default so you need ```_wputenv_s()``` instead of ```_putenv_s()```.  On Python 2.7 I'm not really sure what's right, or what needs to be done about encoding.

This is based on a patch suggested by @jsiirola, I just added the fix for Windows. 

### Changes in this PR
- Add a `CtypesEnviron` context manager for cleanly setting and restoring environment variables, with special handling for propagating environment variables into MSVCRT libraries.
- Use the `CtypesEnviron` context manager to temporarily update the `AMPLFUNC` environment variable to include the `PYOMO_AMPLFUNC` environment variable from the NL writer before loading the NL file in Pynumero

### Legal Acknowledgement


pyomo/contrib/pynumero/interfaces/pyomo_nlp.py
```diff
@@ -66,9 +69,31 @@ def __init__(self, pyomo_model):
...
+                    os.environ['AMPLFUNC'] += "\n" + os.environ['PYOMO_AMPLFUNC']
+                else:
+                    os.environ['AMPLFUNC'] = os.environ['PYOMO_AMPLFUNC']
+                if os.name in ['nt', 'dos']:
+                    ctypes.cdll.msvcrt._wputenv_s("AMPLFUNC", os.environ["AMPLFUNC"])
+
             # now call the AslNLP with the newly created nl_file
             super(PyomoNLP, self).__init__(nl_file)
 
+            # Restore the AMPLFUNC environment variable
```
jsiirola, line 86: I think you want to put the `__init__` call in a `try` with the restoration of the environment in a `finally` block

eslickj: Ok, I'm pretty sure I understand, but just to be sure, this is to ensure the environment gets restored if ```__init__``` fails?  Because if someone catches the exception and it doesn't terminate the process, it will mess with my subsequent calls to the nl-writer?

jsiirola: Correct.  One of the things that I am trying to get us to be more consistent on is that if we modify a global setting (including the environment) with the intent that the change is temporary, then we do it in a try-finally so that we can ensure that the change is still temporary in the face of exceptions / errors / etc.

jsiirola:
@eslickj: can you add a (simple) test that exercises this?  I frequently use a small model that calls a function from the GSL for testing AMPL external functions.  See, e.g. `pyomo/core/tests/unit/test_external.py`

eslickj:
@jsiirola I was just thinking about asking if you had some functions for testing.  Yes, I can write a test.

eslickj:
@jsiirola, the test is in.  I pretty much just copied it from your example, and it to works for me.

pyomo/contrib/pynumero/interfaces/tests/test_external.py
```diff
@@ -0,0 +1,42 @@
...
+#  Copyright 2017 National Technology and Engineering Solutions of Sandia, LLC
+#  Under the terms of Contract DE-NA0003525 with National Technology and
+#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain
+#  rights in this software.
+#  This software is distributed under the 3-clause BSD License.
+#  ___________________________________________________________________________
+#
+import os
+import pyutilib.th as unittest
+from pyomo.contrib.pynumero.interfaces.pyomo_nlp import PyomoNLP
```
jsiirola, line 13: I think you need to change this to:
```python
from pyomo.contrib.pynumero.asl import AmplInterface
if not AmplInterface.available():
    raise unittest.SkipTest(
        "Pynumero needs the ASL extension to run NLP tests")
 from pyomo.contrib.pynumero.interfaces.pyomo_nlp import PyomoNLP 
```
To guard against `pypy`, where the pynumero interface doesn't appear to work.

pyomo/contrib/pynumero/interfaces/tests/test_external.py
```diff
@@ -0,0 +1,42 @@
...
+from pyomo.common.getGSL import find_GSL
+from pyomo.environ import *
+from pyomo.core.base.external import (PythonCallbackFunction,
+                                      AMPLExternalFunction)
+from pyomo.opt import check_available_solvers
```
jsiirola, line 18: This import is not used

eslickj:
I think the Python 2.7 test failed because the strings are ASCII instead of unicode, so I used six to make sure they are consistent with Python 3.  I can't seem to get Pynumero working on Python 2.7 locally, so hopefully it's fixed.  We'll see.

eslickj:
Okay, I'm trying something else.  For Python 2.7 I'm using ```_putenv_s``` (not wide character) and for Python 3.x I'm using ```_wputenv_s``` (wide version).  I'm not sure why this would work any better than converting strings to unicode in Python 2.7, but I'll hope for the best.  Windows with Python 2.7 works fine on my machine either way.

jsiirola:
Okay, progress:  I have been able to reproduce the test failure on my development machine.  I needed the current version of Miniconda3, with a python2.7 environment with the same conda-forge packages GHA uses, and the MSVS 2019 Community Edition.  (The IDAES build of `libpynumero_ASL.dll` is too old for master, and `pyomo build-extensions` with MSVS 2017 fails to build ampl/mp).

eslickj:
@jsiirola I made a new IDAES binary release it if helps.  I built pynumero from master (as of last night).  https://github.com/IDAES/idaes-ext/releases/tag/2.1.1.

jsiirola:
I actually have a working fix. It got complicated enough that I pulled the logic out into a `CtypesEnviron` context manager. There is one side thing I am working on cleaning up and then I’ll push it to this PR. Hopefully today/tonight.

jsiirola:
OK - this should be ready for review.  There were two challenges that had to be resolved:
- It turns out to be insufficient to just transfer the environment into `msvcrt`.  There are a number of different `msvcr*` libraries that a DLL could have been linked against, and we need to update all of them.  In particular, the Python 2.7 test was failing because it was linked against `api-ms-win-crt-environment-l1-1-0.dll`.  The current solution is to look for every known DLL that could contain copies of the environment and update *all* of them.
- When solving the first problem, I ran into an issue where loading `msvcr90.dll` on some platforms (Python 3.5, 3.6, and 3.7) resulted in the `ctypes.CDLL` call to load the library hanging indefinitely.  This PR now implements an approach where we attempt the load in a disposable subprocess (where we can implement a timeout) and then only "do it for real" in the main process if that succeeded.

eslickj:
@jsiirola, thanks.  That was unexpectedly complicated to make work.

pyomo/common/env.py
```diff
@@ -0,0 +1,498 @@
...
+
+    def get_env_dict(self):
+        ans = {}
+        _str_buf = self._envstr()
+        # I am sure there is an easier way to parse this.
+        # GetEnvironmentStringsW returns a single const char* that
+        # points to a block of memory that contains the environment
+        # strings.  Each environment string is NULL terminated, and an
+        # empty string (effectively two consecutive NULLs) indicates the
+        # end of the block.  The following jult parses that data
```
blnicho, line 347: Is "jult" a typo or a programming term I'm unfamiliar with?

pyomo/common/env.py
```diff
@@ -0,0 +1,498 @@
...
+        # environ back to its original state.  Note that there are cases
+        # where some of the libs refer to the same DLL space, so we will
+        # restore them in the opposite order that we set them, so that
+        # cases where multiple libs point to the same environment things
+        # are all restored correctly.
+        for lib in reversed(self.interfaces):
+            lib.restore()
+
+    def __getitem__(self, key):
+        """Retun the current environment variable value from os.environ"""
```
blnicho, line 485: Typo: return

pyomo/common/env.py
```diff
@@ -0,0 +1,498 @@
...
+        return os.environ[key]
+
+    def __contains__(self, key):
+        """Retun true if the key is in os.environ"""
```
blnicho, line 489: Typo: return

pyomo/common/env.py
```diff
@@ -0,0 +1,498 @@
...
+def _as_unicode(val):
+    """Helper function to coerce a string to a unicode() object"""
+    if isinstance(val, six.text_type):
+        return val
+    elif val is not None:
+        return val.decode()
+    return None
+
+
+def _attempt_ctypes_cdll(name):
```
mrmundt, line 36: Seemingly one of the few without a docstring.

pyomo/common/env.py
```diff
@@ -0,0 +1,498 @@
...
+                        "exceeded max environment block size (32767)")
+            key, val = _str.split('=', 1)
+            ans[key] = val
+            i += len(_str_buf[i]) # Skip the NULL
+        self._free_envstr(_str_buf)
+        return ans
+
+
+class CtypesEnviron(object):
+    # Windows has a number of C Runtime Libraries, each of which can
```
mrmundt, line 373: Again, docstring? Is this starter comment meant to be one?

jsiirola:
I have fixed @blnicho and @mrmundt's comments, and @blnicho has reviewed.

PR #1553, opened 2020-07-10
Fix Model.compute_statistics()

### LINKED ISSUE #1551: Incorrect statistics on Model
In Model.compute_statistics() the computed statics object often contains incorrect results when using Blocks.
```
        model = ConcreteModel()
        model.x = Var([1])

        model.B = Block()
        model.B.c = ConstraintList()
        model.B.c.add(model.x[1] == 0)
        print(model.nvariables())
        print(model.nconstraints())
```
The variable count is evaluated to 2 even though there is only one variable and the constraint count is evaluated to 0.

Also no recalculation is performed.
```
        model = ConcreteModel()
        model.x = Var([1])

        print(model.nvariables())
        model.clear()
        print(model.nvariables())
```
The statistic is only evaluated on the first use. Both times nvariables evaluates to 2.

sebuer:
## Summary/Motivation:
compute_statistics should recompute the amount of variables, constraints and objectives when a Model changes. Models containing blocks should also calculate the amount of these statistics correctly. Currently compute_statistics never re-evaluates the statistics. When summing variables, constraints and objectives of blocks, only the amount on the Model is currently summed multiple times.

- Add new test for a Model containing a Block.
- Fix component summation in compute_statistics.
- Add recompute argument to compute_statistics.

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/PyomoModel.py
+++ b/pyomo/core/base/PyomoModel.py
@@ -602,11 +602,11 @@ use the AbstractModel or ConcreteModel class instead.""")
         self.statistics.number_of_constraints = 0
         self.statistics.number_of_objectives = 0
         for block in self.block_data_objects(active=active):
-            for data in self.component_map(Var, active=active).itervalues():
+            for data in block.component_map(Var, active=active).itervalues():
                 self.statistics.number_of_variables += len(data)
-            for data in self.component_map(Objective, active=active).itervalues():
+            for data in block.component_map(Objective, active=active).itervalues():
                 self.statistics.number_of_objectives += len(data)
-            for data in self.component_map(Constraint, active=active).itervalues():
+            for data in block.component_map(Constraint, active=active).itervalues():
                 self.statistics.number_of_constraints += len(data)
 
     def nvariables(self):
--- a/pyomo/core/base/PyomoModel.py
+++ b/pyomo/core/base/PyomoModel.py
@@ -592,12 +592,10 @@ use the AbstractModel or ConcreteModel class instead.""")
         self.solutions = ModelSolutions(self)
         self.config.preprocessor = 'pyomo.model.simple_preprocessor'
 
-    def compute_statistics(self, recompute=True, active=True):
+    def compute_statistics(self, active=True):
         """
         Compute model statistics
         """
-        if not recompute and len(self.statistics) > 0:
-            return
         self.statistics.number_of_variables = 0
         self.statistics.number_of_constraints = 0
         self.statistics.number_of_objectives = 0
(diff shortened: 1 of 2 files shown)


jsiirola:
@sebuer: Thank you for the issue and PR!  I haven't had a chance to discuss this with others yet, but I think I would like to propose a more substantial change to fix the issue:
- deprecate the `compute_statistics()` method
- deprecate the `model.statistics` object completely
- take the logic out of `compute_statistics()` and put it in `nvariables()`, `nconstraints()`, and `nobjectives()`.  Rework those methods so that they *do not* cache the result and instead recompute it every time.  We should probably also add the `active=` argument to those methods. 
  - While we are at it, I would advocate changing the logic to leverage `component_data_objects`; for example: 
     ```python
     def nvariables(self):
         return sum(1 for _ in self.component_data_objects(Var))
     ```
  - ...and this is probably the "right" time to resolve a longstanding ambiguity around the relationship between `active` and `descend_into` in `component_data_objects()`

This also leaves several open questions:
- should `nvariables()`, `nconstraints()`, and `nobjectives()` be moved from `Model` and promoted to `Block`?
- given that summing over `component_data_objects` is pretty straightforward, should we deprecate  `nvariables()`, `nconstraints()`, and `nobjectives()`?  One argument for removing them is it is not always obvious what they are counting.  For example, right now, those methods only "work" for "standard" LP/MIP/NLP/MINLP models.  Models that use derived Blocks that advertise a non-Block `ctype` (e.g., GDP, or the old bilevel models) will not have any variables/constraints/objectives on those subblocks (`Disjunct` or `Submodel`) counted.  

@Pyomo/pyomo-developers: Thoughts?

@sebuer: I realize that we are taking your simple fix and turning it into a bigger design change.  I think a reasonable path might be to update this PR to remove the `recompute=True` argument and just always recompute the values.  That would fix the immediate bugs and we could move the more involved changes to a subsequent PR.

qtothec:
I was embarrassingly not aware of this (or thought it was abandoned code) when I wrote https://github.com/Pyomo/pyomo/blob/master/pyomo/util/model_size.py. The functionality ought to be combined, and made aware of other extension ctypes.

sebuer:
I changed the PR to not include the recompute argument as suggested. For me https://github.com/Pyomo/pyomo/blob/master/pyomo/util/model_size.py is a good replacement for `compute_statistics()` in another project.

jsiirola:
Only test failures are due to Python 3.7 and 3.8 configuration on OSX (which were resolved by #1555).  This is safe to merge.

PR #1554, opened 2020-07-10
Cyipopt scaling

carldlaird:
## Fixes # .
- Added scaling functionality to the CyIpopt interface
- Scaling was added to the pynumero nlp class hierarchy
- Scaling was also added to the PyomoExternalCyIpoptProblem class

## Summary/Motivation:
Scaling is often required on NLP problems, and this is also true when they have external models.

-
-

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/pynumero/interfaces/ampl_nlp.py
+++ b/pyomo/contrib/pynumero/interfaces/ampl_nlp.py
@@ -424,6 +424,26 @@ class AslNLP(ExtendedNLP):
     def get_duals_ineq(self):
         return self._duals_ineq.copy()
 
+    # overloaded from NLP
+    def get_obj_scaling(self):
+        return None
+
+    # overloaded from NLP
+    def get_primals_scaling(self):
+        return None
+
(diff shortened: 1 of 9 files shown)


pyomo/contrib/pynumero/interfaces/ampl_nlp.py
```diff
@@ -424,6 +424,26 @@ def set_duals_ineq(self, duals_ineq):
...
+    def get_constraints_scaling(self):
+        return None
+
+    # overloaded from ExtendedNLP
+    def get_eq_constraints_scaling(self):
+        return None
+
+    # overloaded from ExtendedNLP
+    def get_ineq_constraints_scaling(self):
+        None
```
jsiirola, line 445: I don't see these methods being used.  Is this put in for future-proofing?  If that is the case, does it make sense to omit these methods and only add them when we need to support them?

carldlaird: I implemented them. I would rather the API be complete.

PR #1559, opened 2020-07-14
Add find_nearest_index method to ContinuousSet

Robbybp:
## Summary/Motivation:
A method for finding a point in a ContinuousSet would be useful. Especially when you need to get a point in a ContinuousSet from a nearby point obtained by some floating point arithmetic. "Nearby" can be defined as "within some tolerance." The point is found with binary search via the `bisect` module.

Unresolved:
- Should this method return `None` if the input point is not nearby any point in the set? The behavior right now is to return the closest point in the set. (Arbitrarily, a tie goes to the larger point.)
- If the current behavior is kept, having a tolerance is pretty useless.
- If the current behavior is changed, where should the tolerance be set and how should it be updated?
- Should this functionality be put into a `binary_search_with_tolerance` utility function somewhere? 

- `find` method
- tests

### Legal Acknowledgement


Robbybp:
The behavior proposed is:
- Find the index of the nearest point in the ContinuousSet
- Return this index only if it is within `tolerance` of the target. Otherwise return `None`
- The default tolerance is `None`, which is effectively an infinite tolerance
- Tolerance violation must be strict (`delta > tolerance`). This is so `tolerance=0` may be specified to search with strict equality.
- If there is a tie, the index on the right is returned
- It is the user's responsibility to use a small enough tolerance if they want "equal-within-tolerance" to describe a unique point in the ContinuousSet

Robbybp:
@blnicho @jsiirola This should be ready to review. The Travis test seems to time out somewhere, and the post-build test seems to fail due to codecov not being uploaded.

pyomo/dae/contset.py
```diff
@@ -247,3 +248,54 @@ def construct(self, values=None):
...
 
+    def find_nearest_index(self, target, tolerance=None):
+        """ Returns the index of the nearest point in the 
+        :py:class:`ContinuousSet <pyomo.dae.ContinuousSet>`.
+
+        If a tolerance is specified, the index will only be returned
+        if the distance between the target and the closest point is
+        less than or equal to that tolerance. If the tolerance is less
+        than half the minimum spacing between points in the set, the
+        "closest-point-within-tolerance" will be unique.
```
jsiirola, line 259: This last sentence seems like it no longer applies: the function should always return a unique (and deterministic) value.

pyomo/dae/contset.py
```diff
@@ -247,3 +248,54 @@ def construct(self, values=None):
...
+            # delta_right = p_g - target
+            # neighbors = {
+            #     delta_left: i-1,
+            #     delta_right: i,
+            #     }
+            # If delta_left == delta_right, delta_right will be kept as it is 
+            # added last. In this way, a tie goes to the index on the right.
+            neighbors = dict((abs(target - self[j]), j) for j in [i-1, i])
+            delta = min(neighbors)
+            nearest_index = neighbors[delta]
```
jsiirola, line 296: Why make neighbors a dict?  If the key is the same, I cannot guarantee that `i` will (or will not) always replace `i-1`.  What about:
```python
delta, nearest_index = min((abs(target - self[j]), j) for j in [i-1, i])
```
You should note that this changes the handling for points exactly in the middle (ties now return the left point, not the right).

Robbybp: Constructing a dict from a generator does not guarantee that the insertion order of the dict is the same as the generator? Regardless, your implementation is cleaner. I just didn't consider calling `min` on tuples.

pyomo/dae/tests/test_contset.py
```diff
@@ -201,6 +201,81 @@ def test_duplicate_construct(self):
...
+        self.assertEqual(i, 1)
+
+        # This test fails. I get:
+        # delta_left == 2.075-2.0 == 0.07500000000000018
+        # delta_right == 2.15-2.075 == 0.07499999999999973
+        # Index 8 is returned (time[8] == 2.15), even though
+        # "tie"-break logic suggests index 7 should be.
+        #i = m.time.find_nearest_index(2.075)
+        #self.assertEqual(i, 7)
+
```
Robbybp, line 278: Tiebreak behavior is harder to reproduce than I thought it would be

PR #1562, opened 2020-07-16
Add `get_units` to Param API

alexdetsch:
## Fixes # .
* Move the get_units-function from SimpleParam-class to Param-class so it also can be used in IndexedParam-objects (Or as alternative, implement the function also in IndexedParam-class.

## Summary/Motivation:
* conversion of units is not possible in IndexedParam-objects due to the missing getter-function in that class (currently only implemented in SimpleParam-class)

-
-

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/param.py
+++ b/pyomo/core/base/param.py
@@ -965,6 +965,10 @@ This has resulted in the conversion of the source to dense form.
                  ("Value",),
                  dataGen,
                  )
+    
+    def get_units(self):
+    """Return the units expression for this parameter"""
+    return self._units
 
 
 class SimpleParam(_ParamData, Param):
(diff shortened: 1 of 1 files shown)


jsiirola:
I am not opposed to this change, but I am curious as to the use case: when do you need to know the units of the container (`IndexedParam`)?  We store them there out of convenience and to save memory, but I am not clear when you would actually need id.  Can you provide an example?

blnicho:
@alexdetsch any response to John's questions?

carldlaird:
@alexdetsch I have submitted a PR that actually removes the get_units call from SimpleParam since the implementation in the baseclass _ParamData serves the same purpose. As written above, I don't think any routines will make use of get_units on the IndexedParam, since the units are accessed from the _ParamData object (e.g., m.p[i] instead of m.p). If you have a use case, please let me know. And thanks for the interest - we are working on other issues associated with units as well (e.g., performance).

blnicho:
@alexdetsch I think @carldlaird addressed your comments in the PR description so I'm going to close this. If you have additional comments or questions please open a new issue.

PR #1563, opened 2020-07-17
Deprecate CUID targets in add_slack_variables transformation

emma58:
## Summary/Motivation:

Deprecates the requirement that targets for `core.add_slack_variables` are a list of ComponentUIDs. Instead, they are expected to be a Constraint or list of Constraints.

- Changes to ConfigBlock for handling transformation arguments
- Deprecates cuid targets in favor of Constraints or lists of Constraints
- Updates tests accordingly

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/plugins/transform/add_slack_vars.py
+++ b/pyomo/core/plugins/transform/add_slack_vars.py
@@ -10,9 +10,6 @@ from pyomo.core.base.component import ComponentUID
 from pyomo.core.base import Constraint, _ConstraintData
 from pyomo.common.deprecation import deprecation_warning
 
-import logging
-logger = logging.getLogger('pyomo.core')
-
 def target_list(x):
     deprecation_msg = ("In future releases ComponentUID targets will no longer "
                        "be supported. Specify targets as a Constraint or list "
--- a/pyomo/core/plugins/transform/add_slack_vars.py
+++ b/pyomo/core/plugins/transform/add_slack_vars.py
@@ -5,11 +5,43 @@ from pyomo.opt import SolverFactory
 
 from pyomo.common.modeling import unique_component_name
 from pyomo.core.plugins.transform.hierarchy import NonIsomorphicTransformation
-
-from random import randint
+from pyomo.common.config import ConfigBlock, ConfigValue
+from pyomo.core.base.component import ComponentUID
+from pyomo.core.base import Constraint, _ConstraintData
+from pyomo.common.deprecation import deprecation_warning
 
 import logging
 logger = logging.getLogger('pyomo.core')
+
+def target_list(x):
+    deprecation_msg = ("In future releases ComponentUID targets will no longer "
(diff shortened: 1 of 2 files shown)


pyomo/core/plugins/transform/add_slack_vars.py
```diff
@@ -5,11 +5,40 @@
...
+from pyomo.common.deprecation import deprecation_warning
 
-from random import randint
-
-import logging
-logger = logging.getLogger('pyomo.core')
+def target_list(x):
+    deprecation_msg = ("In future releases ComponentUID targets will no longer "
+                       "be supported. Specify targets as a Constraint or list "
+                       "of Constraints.")
```
jsiirola, line 16: This should include the name of the transformation in the message.

pyomo/core/plugins/transform/add_slack_vars.py
```diff
@@ -5,11 +5,40 @@
...
+    if isinstance(x, ComponentUID):
+        deprecation_warning(deprecation_msg)
```
jsiirola, line 18: We should probably only emit the deprecation message once for every call to the transformation.  Maybe something like:
```python
if deprecation_msg:
    deprecation_warning(deprecation_msg)
    deprecation_msg = None
```

pyomo/core/plugins/transform/add_slack_vars.py
```diff
@@ -5,11 +5,40 @@
...
+        # [ESJ 07/15/2020] We have to just pass it through because we need the
+        # instance in order to be able to do anything about it...
+        return [ x ]
+    elif isinstance(x, (Constraint, _ConstraintData)):
+        return [ x ]
+    elif hasattr(x, '__iter__'):
+        ans = []
+        for i in x:
+            if isinstance(i, ComponentUID):
+                deprecation_warning(deprecation_msg)
```
jsiirola, line 28: ...same as above: only emit the deprecation message once for every call?

pyomo/core/tests/transform/test_add_slacks.py
```diff
@@ -260,28 +260,50 @@ def test_target_objective_create_using(self):
...
+    def test_error_for_non_constraint_target_in_list(self):
+        m = self.makeModel()
+        self.assertRaisesRegexp(
+            ValueError,
+            "Expected Constraint or list of Constraints.\n\tRecieved "
+            "<class 'pyomo.core.base.var.SimpleVar'>",
+            TransformationFactory('core.add_slack_variables').apply_to,
+            m,
+            targets=[m.rule1, m.x]
+            )
```
jsiirola, line 299: You should add a test that confirms that ComponentUID still works, and that it emits a deprecation warning (probably using LoggingIntercept)

emma58:
@jsiirola, OK, I think I addressed everything. Good point about only yelling at people once...

Issue #1571, opened 2020-07-23
Scaling transformation does not work with References


Robbybp:
References break the scaling transformation because it renames the underlying components, then tries to scale the now-broken `Reference` components. Some options:
1. Rename all components (breaking references), then don't try to scale references
2. Don't rename components
3. Create new, renamed components but don't delete the old components

I think I'm in favor of option 1 for now as long as we document that the scaling transformation breaks references. Maybe we can have a `rename_scaled_components` option?

carldlaird:
I wonder if this exposes another problem - are the referenced variables scaled twice? I guess we don't know if it fails.

Robbybp:
This supersedes https://github.com/Pyomo/pyomo/issues/1571

Robbybp:
I think we try to scale the referenced variables twice, but fail when we get to the Reference.

carldlaird:
@blnicho Any thoughts?

PR #1589, opened 2020-08-11
Don't attempt to scale references in scaling transformation

Robbybp:
## Fixes # .
Addresses part of #1571 

## Summary/Motivation:
The scaling transformation currently breaks references (by renaming the underlying components), then tries to scale these references, causing errors to be raised from the depths of indexed_component_slice.py. The underlying issue is that references are being broken by the scaling transformation. This PR is a band-aid that allows the scaling transformation to be called when references are present. As long as these references are not "de-referenced" after the scaling transformation, the rest of the model should work just fine.

- Scaling transformation no longer attempts to scale references
- Adds a small test where a model with references is scaled

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/plugins/transform/scaling.py
+++ b/pyomo/core/plugins/transform/scaling.py
@@ -9,6 +9,7 @@
 #  ___________________________________________________________________________
 
 from pyomo.core.base import Var, Constraint, Objective, _ConstraintData, _ObjectiveData, Suffix, value
+from pyomo.core.base.reference import _ReferenceDict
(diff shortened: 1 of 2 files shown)


jsiirola:
One issue with this is approach is what happens if the scaling transformation is called on a "sub block" that has a reference to a Var (or Constraint) that points to an object outside of the subtree that is being scaled.

Robbybp:
Closing this in favor of #1596 as identifying references through ```python type(comp._data)``` is not supported.

PR #1594, opened 2020-08-13
Add integer arithmetic option to FME transformation and add tolerance for floating point version

emma58:
## Summary/Motivation:

Adds an option to the Fourier-Motzkin elimination transformation to perform only integer arithmetic (if all the coefficients of the variables being projected out are integer), and adds tolerance at which floating points are considered and used as 0 to mitigate numerical issues when the data isn't integer.

- Adds a flag (default to False for backwards compatibility) to tell FME to only to do integer operations. When set, as long as all the data is integer, it will multiply the relevant constraints so that the coefficients are all equal to the least common multiplier (in absolute value), before adding the constraints. (As opposed to dividing so that all the coefficients are 1). It throws an error if this is set and not all the data is integer
- Adds a zero_tolerance argument where all floats less than this tolerance (in absolute) value will be treated as 0 during the projection.
- Adds tests for both the above

### Legal Acknowledgement


pyomo/contrib/fme/fourier_motzkin_elimination.py
```diff
@@ -68,6 +68,22 @@ def vars_to_eliminate_list(x):
             "Expected Var or list of Vars."
             "\n\tRecieved %s" % type(x))
 
+def _is_integer(x):
+    if int(x) == x:
```
jsiirola, line 72: Is there any benefit of doing this within a tolerance?  Since you will raise an exception if this is False, I am thinking something along the lines of:
```python
def _to_integer(x):
    int_x = int(x)
    if abs(x-int_x) < int_tolerance:
        return int_x
    raise ValueError("%x is not an integer" % (x,))
```

pyomo/contrib/fme/fourier_motzkin_elimination.py
```diff
@@ -313,14 +347,39 @@ def _fourier_motzkin_elimination(self, constraints, vars_to_eliminate):
...
+                                cons, -1.0/leaving_var_coef))
+                    else:
+                        geq_list.append(
+                            self._nonneg_scalar_multiply_linear_constraint(
+                                cons, 1.0/leaving_var_coef))
                 else:
-                    geq_list.append(
-                        self._nonneg_scalar_multiply_linear_constraint(
-                            cons, 1.0/leaving_var_coef))
+                    if not _is_integer(leaving_var_coef):
```
jsiirola, line 360: I think you actually need to convert x to an integer to get exact integer arithmetic.  Just checking that the value is integer without ensuring that the `type` is integer will still result in floating point operations.

pyomo/contrib/fme/fourier_motzkin_elimination.py
emma58: I don't know? But it's easy enough to do, and if you want to be really neurotic, you can set the tolerance to 0 (and not have a strict inequality), so yes, seems worth implementing.

pyomo/contrib/fme/fourier_motzkin_elimination.py
emma58: This one is I think okay because it is cast to an int in line 410. But I think you are still quite right, because I'm not making sure to cast the actual coefficients to integers before I start taking linear combinations of them...

pyomo/contrib/fme/fourier_motzkin_elimination.py
emma58: @jsiirola I took your suggestion, kind of, but I kept it returning a boolean so that I could have more specific error messages when I hit non-integer data. Though it seems a bit silly because I cast each of them twice that way...

emma58:
@jsiirola, OK, this is ready for review when you have a chance. :) (Assuming the Jenkins failures aren't related?)

pyomo/contrib/fme/fourier_motzkin_elimination.py
```diff
@@ -301,56 +394,144 @@ def _fourier_motzkin_elimination(self, constraints, vars_to_eliminate):
...
+                            self._nonneg_scalar_multiply_linear_constraint(
+                                cons, 1.0/leaving_var_coef))
                 else:
-                    geq_list.append(
-                        self._nonneg_scalar_multiply_linear_constraint(
-                            cons, 1.0/leaving_var_coef))
+                    if not _is_integer(leaving_var_coef, self.integer_tolerance):
+                        raise ValueError(self._get_noninteger_coef_error_message(
+                            the_var.name, leaving_var_coef))
+                    coefs.append(int(leaving_var_coef))
```
jsiirola, line 426: This is dangerous.  Consider the case:
```python
leaving_var_coef = 0.999
self.integer_tolerance = 0.01
_is_integer(leaving_var_coef, self.integer_tolerance) # == True!
int(leaving_var_coef)  # 0!
```

The preferred way to round is `int(leaving_var_coef + 0.5)`

pyomo/contrib/fme/fourier_motzkin_elimination.py
```diff
@@ -301,56 +394,144 @@ def _fourier_motzkin_elimination(self, constraints, vars_to_eliminate):
...
         """
         body = cons['body']
-        body.linear_coefs = [scalar*coef for coef in body.linear_coefs]
+        new_coefs = []
+        for i, coef in enumerate(body.linear_coefs):
+            v = body.linear_vars[i]
+            new_coefs.append( 
+                self._multiply(scalar, coef,
+                               self._get_noninteger_coef_error_message(v.name,
+                                                                       coef)))
```
jsiirola, line 504: This has performance implications.  The error message will always be created / formatted, even when it is not needed.  Maybe we can redesign this to pass callbacks around and only generate the messages when they are actually going to be raised?

If `_multiply` was:
```python
def _multiply(self, scalar, coef, error_message, error_args):
    #...
    raise ValueError(error_message if error_args is None
                     else error_message(*error_args))
```
then this line could be
```python
self._multiply(scalar, coef,
               self._get_noninteger_coef_error_message,
               (v.name, coef))
```
...of course, this applies to `_add` as well.

pyomo/contrib/fme/fourier_motzkin_elimination.py
```diff
@@ -301,56 +394,144 @@ def _fourier_motzkin_elimination(self, constraints, vars_to_eliminate):
...
+                                            cons['lower'], coef))
         return cons
 
     def _add_linear_constraints(self, cons1, cons2):
-        """Adds two >= constraints"""
+        """Adds two >= constraints
+        
+        Because this is always called after 
+        _nonneg_scalar_multiply_linear_constraint, though it is implemented
+        more generally.
```
jsiirola, line 533: Sentence fragment.  Can you clarify this comment?

jsiirola (changes requested):
One bug that needs to be fixed, and one performance idea (that you may or may not want to implement)...

jsiirola (approved):

jsiirola:
I went ahead and implemented the fixes needed from my review.  @blnicho do you mind looking this over?

blnicho (approved):

PR #1590, opened 2020-08-12
Add Extended Cutting Plane(ECP) and Global OA strategies in MindtPy

ZedongPeng:
## Summary/Motivation:
This PR adds Extended Cutting Plane(ECP) strategy and Global Outer Approximation(GOA) strategy in MindtPy. The GOA strategy is used for solving nonconvex problems using the affine cuts from MC++. The GOA strategy supports single-tree implementation using LazyConstraint Callback. The documentation for each function in MindtPy is also completed in this PR. 

## Changes proposed in this PR:
- Add 'ECP' strategy and 'GOA' strategy.
- Add related configurations
- Complete the documentation for each function
- Move all the configurations to a new file.
- Update the online doc of MindtPy

### Legal Acknowledgement

By contributing to this software project, I have read the [contribution guide](https://pyomo.readthedocs.io/en/stable/contribution_guide.html) and agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #1596, opened 2020-08-14
Option to not rename in scaling transformation

### LINKED ISSUE #1571: Scaling transformation does not work with References

Robbybp:
## Fixes #1571 
Hopefully.

## Summary/Motivation:
Scaling transformation breaks references due to renaming. By not renaming, scaling transformation can be used on models with references. Components with references may be scaled twice, however, so we add a filter to make sure they are only scaled once.

- `rename=True` option in `_apply_to` method of scaling transformation
- Filter for components that have already been scaled (due to references)
- A simple test.

I'm not sure how to reconcile this with https://github.com/Pyomo/pyomo/pull/1589. On one hand, if I have references and want to rename, I should be able to do that, knowingly breaking my references in the process. On the other, if I have references and don't want to rename, the underlying components should be scaled at least once. Seems like this could be handled by a simple 'if' statement, but there may be some logic I'm missing, or an option we should provide.

### Legal Acknowledgement


Robbybp:
Test failures appear to be: Ply not found (linux/3.5), unrecognized mpi argument oversubscribe (linux/3.7/mpi), error installing GAMS (osx/3.5)

jsiirola:
I have been working these test failures over on #1590, and should be pretty close.

blnicho:
@carldlaird could you take a look at this PR? We'd like to get it into the 5.7.2 release this week

carldlaird (approved):
This looks reasonable to me and doesn't change the default behavior.

PR #1600, opened 2020-08-16
Add an option to specify the path to ipopt

### LINKED ISSUE #1566: parmest: allow user to specify path to ipopt executable
This would be a nice enhancement for folks with multiple versions of Ipopt installed. Alternately, we could let the user provide an Ipopt solver object.

DLWoodruff:
## Summary/Motivation:

Adds an optional keyword argument to the constructor to allow for the user to specify the path to ipopt. There are pros and cons:
Cons:
- One could change one's PATH rather than using this option
- I don't know how to test it
- I didn't document it other than in the api docstring because I don't know how to test an example either

Pros:
- It works
- It is documented in the API section of the documentation
- It only adds one line, so if we are worried about code coverage, surely we can find a place to use a list
comprehension instead of a loop to get that line back.

- add an optional constructor arg and an attribute to the estimator object that specifies a path to the solver 
- add the executable keyword arg to all SolverFactory calls

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/parmest/parmest.py
+++ b/pyomo/contrib/parmest/parmest.py
@@ -339,9 +339,12 @@ class Estimator(object):
         If True, print diagnostics from the solver
     solver_options: dict, optional
         Provides options to the solver (also the name of an attribute)
+    solver_path: string, optional
+        Absolute or relative path to the solver executable (e.g., ~/Ipopt/ipopt)
     """
     def __init__(self, model_function, data, theta_names, obj_function=None, 
-                 tee=False, diagnostic_mode=False, solver_options=None):
+                 tee=False, diagnostic_mode=False, solver_options=None,
+                 solver_path=None):
         
         self.model_function = model_function
         self.callback_data = data
@@ -355,6 +358,7 @@ class Estimator(object):
         self.tee = tee
         self.diagnostic_mode = diagnostic_mode
         self.solver_options = solver_options
+        self.solver_path = solver_path
         
         self._second_stage_cost_exp = "SecondStageCost"
         self._numbers_list = list(range(len(data)))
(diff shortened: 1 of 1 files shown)


pyomo/contrib/parmest/parmest.py
```diff
@@ -581,9 +585,9 @@ def _Q_opt(self, ThetaVals=None, solver="ef_ipopt",
 
             model = stsolver.make_ef()
             stream_solver = True
-            ipopt = SolverFactory('ipopt')
-            sipopt = SolverFactory('ipopt_sens')
-            kaug = SolverFactory('k_aug')
+            ipopt = SolverFactory('ipopt', executable=self.solver_path)
+            sipopt = SolverFactory('ipopt_sens', executable=self.solver_path)
+            kaug = SolverFactory('k_aug', executable=self.solver_path)
```
jsiirola, line 590: I am not sure that this is valid, as the executable specifies the absolute path to the executable (file).  Wouldn't we need to be able to specify a separate executable flag for ipopt, sipopt and kaug?

Would a better / easier approach be to pass an actual solver object into the method (and leave it up to the user to create it)?  If the solver was `None`, you could still create one using default options.

adowling2: I agree with John. It would be better to ask the user to pass solver objects.

DLWoodruff: Or maybe sound the retreat. We are adding a lot complication for something that can be accomplished by changing the PATH.

blnicho:
I'm going to close this for now. The issue is still active but I think the discussion here has revealed that there are some issues with the design of the solution in this PR and questions about if the issue should be resolved on the Pyomo side or not. I think we should resolve this discussion in the issue and open a redesigned PR if needed.

PR #1606, opened 2020-08-27
Error and warning options in solve_consistent_initial_conditions

Robbybp:
## Summary/Motivation:
I want to set error/warning options of `deactivate_model_at` from a call to `solve_consistent_initial_conditions`.

- Add `allow_skip` and `suppress_warnings` options in `solve_consistent_initial_conditions`, which are just sent to `deactivate_model_at`.
- Add a test that raises an error due to `allow_skip=False`

### Legal Acknowledgement

### DIFF
--- a/pyomo/dae/initialization.py
+++ b/pyomo/dae/initialization.py
@@ -99,6 +99,11 @@ def solve_consistent_initial_conditions(model, time, solver, tee=False,
         time: Set whose initial conditions will remain active for solve
         solver: Something that implements a solve method that accepts
                 a model and tee keyword as arguments
+        tee: tee argument that will be sent to solver's solve method
+        allow_skip: If True, KeyErrors due to Constraint.Skip being
+                    used will be ignored
+        suppress_warnings: If True, warnings due to ignored
+                           KeyErrors will be suppressed
 
     Returns:
         The object returned by the solver's solve method
--- a/pyomo/dae/initialization.py
+++ b/pyomo/dae/initialization.py
@@ -87,7 +87,8 @@ def get_inconsistent_initial_conditions(model, time, tol=1e-8, t0=None,
     return list(inconsistent)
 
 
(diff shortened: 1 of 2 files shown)


pyomo/dae/initialization.py
```diff
@@ -87,7 +87,8 @@ def get_inconsistent_initial_conditions(model, time, tol=1e-8, t0=None,
     return list(inconsistent)
 
 
-def solve_consistent_initial_conditions(model, time, solver, tee=False):
+def solve_consistent_initial_conditions(model, time, solver, tee=False,
+        allow_skip=True, suppress_warnings=False):
```
blnicho, line 91: @Robbybp could you expand the docstring with descriptions of these new options? It isn't really clear to me what the `allow_skip` option is doing by the name alone.

jsiirola (approved):

jsiirola:
I am not going to block this, but I am not entirely sure I am a fan of these changes.  As @blnicho noted, we should probably explore a better name than `allow_skip` (it is not clear what it is really allowing).  Also, why do we need a `suppress warnings` flag?  If a user wants to do that, shouldn't they just:
```python
with TemporaryLogLevel(logging.getLogger('pyomo.dae'), logging.WARNING+1):
    solve_consistent_initial_conditions(model, t, ipopt)
```

Robbybp:
The reason for the `suppress_warnings` argument is that I didn't know that Pyomo has a logger. I'll probably remove this option from `deactivate_model_at` altogether though, as it is clumsy and these warnings have never been useful to me. I guess the user can set `allow_skip=False` and catch the `KeyError`s if they really want a warning here. 

I agree that `allow_skip` is not the best name. Alternatives could be `allow_constraint_skip`, `fail_on_index_miss`, or `allow_skipped_index`. 

I'll make these changes in a separate PR.

PR #1618, opened 2020-09-09
GDP: consistent key for calculated M values in bigM dictionary

emma58:
## Summary/Motivation:

When M values are retrieved from Suffixes or args, `get_m_value_src` returns (src, key) where src is the Suffix or dictionary passed as the argument and key is the key in the Suffix or dictionary that was used to get the value of M. Prior to this change, if the M was calculated by the transformation, `(m_lower, m_upper)` was returned. This assumes that the src and key are the same for both the lower and upper values, which is a false assumption.

This PR deprecates the `get_m_value_src` method, and adds a new method `get_M_value_src` which returns `((lower_m_value, lower_m_source, lower_m_key), (upper_m_value, upper_m_source, upper_m_key))`, and additionally it adds a simpler method `get_m_value` which returns `(lower_m_value, upper_m_value`.

- Changes src_bigm dict in bigM transformation to accomodate the lower and upper values coming form different places 
- Adds `get_M_value` method
- Deprecates `get_m_value_src` and replaces with `get_M_value_src`
- Adjusts tests to match

### Legal Acknowledgement


emma58:
@jsiirola, 38 minutes to spare. :P By the way, this is a breaking change, but I think of something that's new enough it's worth it, right? Also, the current behavior was undocumented...

pyomo/gdp/plugins/bigm.py
```diff
@@ -654,10 +654,10 @@ def _transform_constraint(self, obj, disjunct, bigMargs, arg_list,
 
             if c.lower is not None and M[0] is None:
                 M = (self._estimate_M(c.body, name)[0] - c.lower, M[1])
-                bigm_src[c] = M
+                bigm_src[c] = (None, M)
             if c.upper is not None and M[1] is None:
                 M = (M[0], self._estimate_M(c.body, name)[1] - c.upper)
-                bigm_src[c] = M
+                bigm_src[c] = (None, M)
```
jsiirola, line 660: Will this end up "lying" about the source if a user sets a suffix with something like `(None, 10)` for an equality constraint?

jsiirola (approved):
I think this is ok, although I somewhat wonder about the validity of the "src" if a source like the kwds or the Suffix only partially specify the M (i.e., `(0, None)` for an equality constraint).

pyomo/gdp/plugins/bigm.py
emma58: Yikes, you're right. There can be more than one source for a constraint... Should this return a list? Ick.

jsiirola: @carldlaird: thoughts?  

What do we want to know here?  In general, is can be useful to know *where* the M value came from (and there could be more than one M value), and *what( the M value is.  The original idea was if you know where (e.g., from which suffix/dict, and what index/item), then the value is just an item lookup.  Carl advocated returning all three.

What about the following:
```python
(lower_M, lower_src, lower_idx), (upper_M, upper_src, upper_idx)
```
...or would it make more sense to keep the src, idx together:
```python
(lower_M, (lower_src, lower_idx)), (upper_M, (upper_src, upper_idx))
```
...but this makes a "triply-nested" structure - perhaps too complex?

emma58:
@jsiirola, assuming tests pass, I think this is ready, if you wouldn't mind having another look. :)

pyomo/gdp/plugins/bigm.py
```diff
@@ -704,66 +722,144 @@ def _transform_constraint(self, obj, disjunct, bigMargs, arg_list,
...
+                                                             need_upper,
+                                                             bigMargs, None,
+                                                             constraint_name,
+                                                             from_args=True)
+            if not need_lower and not need_upper:
+                return lower, upper
+
+        return lower, upper
+
+    def _get_M_from_suffixes(self, constraint, suffix_list, lower, upper):
```
jsiirola, line 810: Is this a good name?  When I read above where this is called, it looks like this method completely overrides the M values from the args.  Maybe a better name is `_update_M_from_suffixes`?

jsiirola (approved):
This looks good, with one (non-blocking) question about naming.

pyomo/gdp/plugins/bigm.py
emma58: It doesn't override, but I agree that 'update' is more apt, because it will skip anything already found in args.

jsiirola (approved):

jsiirola:
The test failures are unrelated to this PR (and will be addressed by #1625).  This should be good to merge.

PR #1620, opened 2020-09-10
Remove truthiness assumption for sequence in TuplizeValuesInitializer

### LINKED ISSUE #1564: ValueError: The truth value of a MultiIndex is ambiguous is caused by set.py
The PyomoGallery network interdiction examples fails with the error:
`ValueError: The truth value of a MultiIndex is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().`

The error is fixed if line 414 in pyomo/core/base/set.py is changed from:
`elif not _val:`
to
`elif _val is None:`

I am running Anaconda Python 3.6.5 under Windows 10 (64 bit)

samryan18:
## Summary/Motivation:
`TuplizeValuesInitializer` should not assume that the sequence can be truthy. pandas indices [do not support this](https://pandas.pydata.org/pandas-docs/version/0.15/gotchas.html) and will raise a ValueError.

- remove assumption that the param in this initializer can be considered truthy/falsey and instead, just explicitly check whether it is None
- behavior should remain unchanged for empty sequences (or empty non-sequences) which will still be caught by the `if len(_val) == 0` check

Thanks @stigrs !

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/tests/unit/test_set.py
+++ b/pyomo/core/tests/unit/test_set.py
@@ -1781,7 +1781,10 @@ class Test_SetOperator(unittest.TestCase):
         # ValueError: The truth value of a MultiIndex is ambiguous. 
         # Use a.empty, a.bool(), a.item(), a.any() or a.all().
         iterables = [['bar', 'baz', 'foo', 'qux'], ['one', 'two']]
-        pandas_index = pd.MultiIndex.from_product(iterables, names=['first', 'second'])
+        pandas_index = pd.MultiIndex.from_product(
+            iterables, 
+            names=['first', 'second']
+        )
 
         model = ConcreteModel()
         model.a = Set(initialize=pandas_index,
--- a/pyomo/core/tests/unit/test_set.py
+++ b/pyomo/core/tests/unit/test_set.py
@@ -1785,13 +1785,16 @@ class Test_SetOperator(unittest.TestCase):
         model.a = Set(initialize=pandas_index,
                       dimen=pandas_index.nlevels)
 
-        # test that dimension is inferred correctly
+        # we will confirm that dimension is inferred correctly
         model.b = Set(initialize=pandas_index)
 
-        for s in (model.a, model.b):
-            self.assertIsInstance(s, Set)
-            self.assertEquals(list(s), list(pandas_index))
-            self.assertEquals(s.dimen, pandas_index.nlevels)
+        self.assertIsInstance(model.a, Set)
+        self.assertEquals(list(model.a), list(pandas_index))
+        self.assertEquals(model.a.dimen, pandas_index.nlevels)
+
+        self.assertIsInstance(model.b, Set)
+        self.assertEquals(list(model.b), list(pandas_index))
+        self.assertEquals(model.b.dimen, pandas_index.nlevels)
 
 
 class TestSetUnion(unittest.TestCase):
--- a/pyomo/core/tests/unit/test_set.py
+++ b/pyomo/core/tests/unit/test_set.py
@@ -1785,6 +1785,7 @@ class Test_SetOperator(unittest.TestCase):
         model.a = Set(initialize=pandas_index,
                       dimen=pandas_index.nlevels)
         self.assertIsInstance(model.a, Set)
+        self.assertEquals(list(model.a), list(pandas_index))
 
 
 class TestSetUnion(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


jsiirola:
This seems reasonable.  Can you please add a test that verifies the expected behavior from #1564 (appropriately guarded by a `unittest.skipIf(not pandas_available)`)?

pyomo/core/tests/unit/test_set.py
```diff
@@ -64,6 +64,12 @@
     Objective,
 )
 
+try:
+    import pandas as pd
+    pandas_available = True
+except:
+    pandas_available = False
```
jsiirola, line 71: FYI: This is fine and does not need to be changed (as it is in a test file).  But for future reference, you can also:
```python
from pyomo.common.dependencies import pandas as pd, pandas_available
```
This leverages a deferred import mechanism so that the import is not attempted until either you cast `pandas_available` to a `bool`, or you use an attribute from `pandas`/`pd`.  As I mentioned, the deferred import is not as useful here, as you are guaranteed to need to check the availability, but in core code, this avoids the pandas import (which is very slow) unless you *actually* are going to need pandas.

samryan18: Neat! Good to know, thanks.

pyomo/core/tests/unit/test_set.py
```diff
@@ -1768,6 +1769,24 @@ def x_init(m,i):
...
+        iterables = [['bar', 'baz', 'foo', 'qux'], ['one', 'two']]
+        pandas_index = pd.MultiIndex.from_product(
+            iterables, 
+            names=['first', 'second']
+        )
+
+        model = ConcreteModel()
+        model.a = Set(initialize=pandas_index,
+                      dimen=pandas_index.nlevels)
+        self.assertIsInstance(model.a, Set)
```
jsiirola, line 1787: It would be good to also verify that the set ends up containing the appropriate entries, e.g., with an `assertEqual(model.a, [...])`

pyomo/core/tests/unit/test_set.py
```diff
@@ -1768,6 +1769,24 @@ def x_init(m,i):
...
+        # Use a.empty, a.bool(), a.item(), a.any() or a.all().
+        iterables = [['bar', 'baz', 'foo', 'qux'], ['one', 'two']]
+        pandas_index = pd.MultiIndex.from_product(
+            iterables, 
+            names=['first', 'second']
+        )
+
+        model = ConcreteModel()
+        model.a = Set(initialize=pandas_index,
+                      dimen=pandas_index.nlevels)
```
jsiirola, line 1786: Do you have to specify the dimen?  Is there a second test where you leave the dimen unspecified and Pyomo (hopefully correctly) infers that the set is 2D (and of course after creating the set, we check that dimen ends up 2)?

jsiirola (commented):
This is very close.  Some questions about the tests, and then I think we can merge.

pyomo/core/tests/unit/test_set.py
samryan18: Setting `dimen` explicitly is actually what triggers the call that leads to the ValueError but that's a reasonable test in and of itself. 

When not setting dimen explicitly in the example, the set is created appropriately but the following warning is raised which seems unexpected:
```
WARNING: Implicitly replacing the Component attribute a (type=<class
    'pyomo.core.base.set.OrderedSimpleSet'>) on block unknown with a new
    Component (type=<class 'pyomo.core.base.set.AbstractOrderedSimpleSet'>).
    This is usually indicative of a modelling error. To avoid this warning,
    use block.del_component() and block.add_component().
```

Any idea what the "modelling error" might be?

Here is the example:
```python
        iterables = [['bar', 'baz', 'foo', 'qux'], ['one', 'two']]
        pandas_index = pd.MultiIndex.from_product(
            iterables, 
            names=['first', 'second']
        )

        model = ConcreteModel()
        model.a = Set(initialize=pandas_index) # not specifying dimen
        # this will raise the warning but list(model.a) returns what we expect
```

jsiirola: Yeah - you are using the same model for the two "halves" of the test.  The first part is creating `model.a`, then in this part you are creating another Set and attempting to also add it to the model using the `a` attribute.  The quick fix is to make the second `model.a = ` to just be `model.b = `.

samryan18: [Here](https://github.com/Pyomo/pyomo/blob/f607ebf301165caa8c78a3d5806fd5fcd6f946ec/pyomo/core/base/block.py#L565) is the line that gives the warning. Bit tricky to trace back without a stack trace (this doesn't seem like a big issue but flagging it anyways).

samryan18: Ahh 🤦‍♂️. Makes sense

jsiirola (approved):

blnicho (approved):

PR #1633, opened 2020-09-17
Update warning message for loading SolverResults with warning status

### LINKED ISSUE #1634: Loading a SolverResults object warning message
This warning message (generated from [`PyomoModel.py`](https://github.com/Pyomo/pyomo/blob/0031882b6e0f4a51e7a501b5672c42b0e9af65d4/pyomo/core/base/PyomoModel.py#L226)) can read very cryptically:
```
WARNING: Loading a SolverResults object with a warning status into
    model=unknown;
        message from solver=<undefined>
```

So much so that there is a [contrib class](https://github.com/Pyomo/pyomo/blob/f607ebf301165caa8c78a3d5806fd5fcd6f946ec/pyomo/contrib/gdpopt/util.py#L45) specifically written to suppress this. There are also lots of questions on this various places online.

Part of the problem is the [initialization](https://github.com/Pyomo/pyomo/blob/1e3bb15da5107ac769bcfc029861b7b22e7c5949/pyomo/core/base/PyomoModel.py#L579) of `name` for models to "unknown".

See #1633. Will propose an improved warning soon.

samryan18:
## Summary/Motivation:
This error message can read very cryptically:
```
WARNING: Loading a SolverResults object with a warning status into
    model=unknown;
        message from solver=<undefined>
```

There is lots of confusion about this online. So much so that there is a [contrib class](https://github.com/Pyomo/pyomo/blob/f607ebf301165caa8c78a3d5806fd5fcd6f946ec/pyomo/contrib/gdpopt/util.py#L45) specifically written to suppress this. Generally the issue is infeasibility. This simple addition will capture that case and include it in the warning and should alleviate a lot of developer confusion.

Also clarifying the "unknown" to be referencing `model.name` so that it is clear this is a successfully initialized pyomo model (in case users have not specified a name in the constructor).

Note: Part of the problem is the [initialization](https://github.com/Pyomo/pyomo/blob/1e3bb15da5107ac769bcfc029861b7b22e7c5949/pyomo/core/base/PyomoModel.py#L579) of `name` for models to "unknown" but that is probably a separate issue.

This small change will update the message in cases of infeasibility to:
```
WARNING: Loading a SolverResults object with a warning status into model.
    model.name="unknown"; message from solver=<undefined>; termination
    condition=infeasible
```

Not sure we really need tests here as this is just logging and capturing logs for testing purposes is a bit tricky across multiple python versions.

I am using this script to validate:
```python
import pyomo.environ as pyo


def test(model, infeasible: bool):
    model.s = pyo.Set(initialize=range(3))
    model.x = pyo.Var(model.s, within=pyo.Reals)

    def obj_expression(model):
        return sum(model.x[val] for val in model.s)

    def con1(model):
        return sum(model.x[val] for val in model.s) <= -1

    def con2(model):
        return sum(model.x[val] for val in model.s) >= 1

    model.OBJ = pyo.Objective(rule=obj_expression)
    model.con1 = pyo.Constraint(model.s, rule=con1)
    if infeasible:
        model.con2 = pyo.Constraint(model.s, rule=con2)

    optimizer = SolverFactory("cbc")
    results = optimizer.solve(model)


test(pyo.ConcreteModel(name="my_model"), infeasible=True)
test(pyo.ConcreteModel(), infeasible=True)

test(pyo.ConcreteModel(name="my_model"), infeasible=False)
test(pyo.ConcreteModel(), infeasible=False)
```

### Legal Acknowledgement


pyomo/core/base/PyomoModel.py
```diff
@@ -217,11 +218,27 @@ def load_from(self,
         # If there is a warning, then print a warning message.
         #
         if (results.solver.status == SolverStatus.warning):
+            if hasattr(results.solver, "termination_condition") and (
+                results.solver.termination_condition
+                == TerminationCondition.infeasible
+            ):
```
jsiirola, line 224: Is there any reason you are only including the termination condition if it is infeasible?  Also, why remove the message formatting?

Would you consider the following solution:
```python
            tc = getattr(results.solver, 'termination_condition', None)
            msg = getattr(results.solver, 'msg', None)
            logger.warning(
                'Loading a SolverResults object with a '
                'warning status into model=%s;\n'
                '    termination condition=%s\n'
                '    message from solver=%s'
                % (instance.name, tc, msg))
```

blnicho:
This PR was reworked and merged in #1752 so closing.

pyomo/core/base/PyomoModel.py
samryan18: Whoops missed this. Looks like you addressed this in #1752 so thanks!

Good point—no good reason to only include termination condition if infeasible. 

And I'm not quite sure I understood the point of the way it was formatted before and marginally preferred the update I made but don't feel strongly at all about that one (I assume just convention pyomo has adopted?)

PR #1639, opened 2020-09-21
Utility function to get slice of a component along certain sets

Robbybp:
## Summary/Motivation:
This functionality is very useful if you have a small number of components you want to generate slices of, and don't necessarily know where the sets you want to slice are located. I envision that these slices could be used to generate a CUID.

- Function to generate slices in indices corresponding to certain sets
- Helper functions to accomplish this goal
- Tests

Includes the changes in #1638.

### Legal Acknowledgement


pyomo/util/slices.py
```diff
@@ -0,0 +1,289 @@
...
+
+    location = len_index - 1
+    # Step through subsets in reverse order, assigning locations,
+    # until a set of dimen==None is encountered.
+    for sub_coord, sub in enumerate(reversed(subsets)):
+        sub_coord = n_subsets - sub_coord - 1
+        dimen = sub.dimen
+        if dimen is None:
+            if (dimen_none_set is not sub or
+                    dimen_none_set_coord != sub_coord):
```
jsiirola, line 267: Isn't it sufficient to just check `dimen_none_set_coord != sub_coord`, as `dimen_none_set is not sub` always implies the coordinates are different?

pyomo/util/slices.py
```diff
@@ -0,0 +1,289 @@
...
+from pyomo.core.base.indexed_component import normalize_index
+from pyomo.core.base.indexed_component_slice import IndexedComponent_slice
+from pyomo.core.base.global_set import UnindexedComponent_set
+from pyomo.common.collections import ComponentSet, ComponentMap
+
+def _get_index_if_present(comp):
+    if hasattr(comp, 'index'):
+        return comp.index()
+    else:
+        return None
```
jsiirola, line 22: I don't think this method is necessary.  See the comment in `get_component_call_stack`.

pyomo/util/slices.py
Robbybp: But the same set could appear in multiple locations, and I want to count that as invalid. The test on lines test_slices.py:517-521 would not raise the desired error if I omitted the second check.

pyomo/core/base/indexed_component_slice.py
```diff
@@ -276,6 +276,14 @@ def expanded_items(self):
         return ((_iter.get_last_index(), _) for _ in _iter)
 
 
+def _tuple_from_possible_scalar(source):
+    if type(source) is not tuple:
+        # This will behave poorly for non-tuple,
+        # non-string iterables, but we do not 
+        # expect non-tuple, non-string iterables.
+        return (source,)
+    return source
```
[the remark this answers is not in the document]
Robbybp: Probably thought I would need to use it more often. Inlined.

pyomo/util/slices.py
```diff
@@ -0,0 +1,289 @@
...
+
+def _generate_subsets(setprod):
+    if hasattr(setprod, 'subsets'):
+        for sub in setprod.subsets():
+            yield sub
+    else:
+        # setprod is not a SetProduct. 
+        # Its only "subset" is itself
+        yield setprod
```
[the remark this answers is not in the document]
Robbybp: Noted, removed.

pyomo/util/slices.py
```diff
@@ -0,0 +1,289 @@
...
+        # to the call stack.
+        if comp is context:
+            # We are done
+            break
+
+        # Add (get_item, index) to the call stack
+        index = _get_index_if_present(comp)
+        if index is not None:
+            call_stack.append((IndexedComponent_slice.get_item, index))
+        parent_component = comp.parent_component()
```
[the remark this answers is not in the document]
Robbybp: Good call. `comp` can be an indexed component though so I need
```python
if parent_component.is_indexed() and comp is not parent_component:
    ...
```

pyomo/util/slices.py
Robbybp: You're right. Removed.

pyomo/util/slices.py
```diff
@@ -0,0 +1,289 @@
...
+    -----------
+    comp : `pyomo.core.base.component.Component`
+        The component to locate
+    context : `pyomo.core.base.block.Block`
+        The block within which to locate the component. If `None`, the
+        top-level model will be used.
+
+    Returns:
+    --------
+    `list` : Contains the necessary method calls and their arguments
```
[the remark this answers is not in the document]
Robbybp: Included a note to this effect in the documentation of the return value.

pyomo/util/slices.py
jsiirola: Agreed.  My point was that I think you *only* need the second check (`dimen_none_set_coord != sub_coord`) and you don't need the first (`dimen_none_set is not sub`)

Robbybp: D'oh yep you're right. Changed.

pyomo/util/slices.py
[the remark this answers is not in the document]
jsiirola: Good call.  In fact, I think you *only* need
```python
if comp is not parent_component:
    ...
```

pyomo/core/base/indexed_component_slice.py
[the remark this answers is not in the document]
jsiirola: FYI: It looks like you haven't pushed the inlining commit to the PR...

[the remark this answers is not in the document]
Robbybp: Oh I forgot that this change (in indexed_component_slice.py) was included in this PR, so I think I inlined something else when I read that comment. I've now inlined the function above.

pyomo/util/slices.py
```diff
@@ -0,0 +1,275 @@
...
+from pyomo.core.base.global_set import UnindexedComponent_set
+from pyomo.common.collections import ComponentSet, ComponentMap
+
+def _to_iterable(source):
+    iterable_scalars = six.string_types + (six.binary_type, six.text_type)
+    if isinstance(source, iterable_scalars):
+        yield source
+    elif hasattr(source, '__iter__'):
+        for obj in source:
+            yield obj
```
jsiirola, line 24: For efficiency, it might make sense to switch this:
```python
if hasattr(source, '__iter__'):
    if isinstance(source, iterable_scalars):
        yield source
    for obj in source:
        yield obj
```
This reduces the umber of `isinstance` calls that are expected to return False (which can be slow).

jsiirola (approved):
I think this is good to go.  One comment that would improve efficiency (slightly), but nothing that should hold up the PR.

pyomo/util/slices.py
Robbybp: Easy enough, and makes sense given the high volume of non-iterable scalars we expect.

blnicho (approved):

PR #1653, opened 2020-10-06
External cyipopt phase II

carldlaird:
## Fixes #
- Implements a new interface for dealing with external codes (e.g., black-box or grey-box) that can provide multiple inputs, outputs, and equality constraints.
- Also includes a cyipopt solver interface that is now callable with a standard Pyomo model (and a model containing external code)

## Summary/Motivation:
- The motivation is to provide support for multi-input / multi-output external models within Pyomo.

-
-

### Legal Acknowledgement


jsiirola (approved):

pyomo/contrib/pynumero/interfaces/ampl_nlp.py
```diff
@@ -440,14 +440,24 @@ def get_constraints_scaling(self):
     def get_eq_constraints_scaling(self):
         constraints_scaling = self.get_constraints_scaling()
         if constraints_scaling is not None:
-            return np.compress(self._con_full_eq_mask, constraints_scaling)
+            # It is possible that derived classes may implement
+            # additional constraints.  We will assume that the first
+            # _n_con_full entries from get_constraints_scaling
+            # correspond to the constraints from AslNLP.
+            return np.compress(self._con_full_eq_mask,
+                               constraints_scaling[:self._n_con_full])
```
michaelbynum, line 448: This seems strange to me. If this method is problematic for derived classes, shouldn't they just override this method?

pyomo/contrib/pynumero/interfaces/ampl_nlp.py
```diff
@@ -440,14 +440,24 @@ def get_constraints_scaling(self):
...
     def get_ineq_constraints_scaling(self):
         constraints_scaling = self.get_constraints_scaling()
         if constraints_scaling is not None:
-            return np.compress(self._con_full_ineq_mask, constraints_scaling)
+            # It is possible that derived classes may implement
+            # additional constraints.  We will assume that the first
+            # _n_con_full entries from get_constraints_scaling
+            # correspond to the constraints from AslNLP.
+            return np.compress(self._con_full_ineq_mask,
+                               constraints_scaling[:self._n_con_full])
```
michaelbynum, line 460: Same comment as above.

michaelbynum (changes requested):

pyomo/contrib/pynumero/interfaces/ampl_nlp.py
jsiirola: Yes.  The original implementation of `PyomoGreyBoxNLP` inherited from `PyomoNLP` and just "tagged on" the grey box constraints / variables to the end of the base pyomo components.  The derived class did indeed overrode this method.  This implementation (and the associated comment) was done this way so that derived classes did not have to directly access private attributes from the base class.  That is, the `PyomoGreyBoxNLP.get_eq_constraints_scaling()` looked something like:
```python
def get_eq_constraints_scaling(self):
    return np.concatenate(
        super(PyomoGreyBoxNLP, self).get_eq_constriaints_scaling(),
        self._greybox_eq_con_scaling
    )
```

However, during development, we ran across other issues where some methods did not behave cleanly when subclassed, to the implementation of `PyomoGreyBoxNLP` in this PR uses containment instead of inheritance.  I would like to revisit the implementation / data structures in `PyomoNLP` so that it can more cleanly be subclassed, but that will wait for a future PR.

michaelbynum (approved):

michaelbynum:
The failing test is unrelated to this PR and is fixed on master, so I am going to merge.

PR #1664, opened 2020-10-16
Treat bounds of of variables being projected as constraints in FME transformation

emma58:
## Summary/Motivation:

The FME transformation only collected variable bounds on the Block it was called on. This is silly because for variables not being projected, you don't need these constraints, and for variables being projected, their bounds should be included regardless of where the Var was declared.

- No longer adds all variable bounds on `instance`: Instead adds all variable bounds for Vars in `vars_to_eliminate`
- Adds a test for the above
- This changes baselines in existing tests because it changes constraint ordering
- Slips in an unrelated test to mitigate some gdp.cuttingplanes paranoia

### Legal Acknowledgement


jsiirola (approved):
This looks good; we can merge pending successful tests.

PR #1679, opened 2020-10-22
Add skip_stale_vars option for store_to function

### LINKED ISSUE #1658: Skip stale=True variables in store_to function
I am now solving a model iteratively. In different iterations, I'll fix the value of some variables. Since the I wanna record the optimal solution of each iteration, I use the [`store_to`](https://github.com/Pyomo/pyomo/blob/1e3bb15da5107ac769bcfc029861b7b22e7c5949/pyomo/core/base/PyomoModel.py#L290) function (`model.solutions.store_to(result)`). Since I'll also fix some variables in each iteration, the `stale` flag of these variable will become `True` after solving. The fixed variables in each iteration may not be the same.
However, the store_to function will skip the variables whose `stale` flag is `True`.

```
if obj.stale:
    continue
```
Ref: https://github.com/Pyomo/pyomo/blob/1e3bb15da5107ac769bcfc029861b7b22e7c5949/pyomo/core/base/PyomoModel.py#L321

I quite understand that this will improve the efficiency of storing and loading solutions. However, if I am not loading the solution right after storing, it will cause some problems. For example, after 10 iterations, I wanna load the solution of the 3rd iteration. Since the variables fixed in iteration 3 and iteration 10 are not the same. The `stale` flag of all variables is not the same. The loading will be incomplete or even wrong.(There are many options in [`load_from`](https://github.com/Pyomo/pyomo/blob/1e3bb15da5107ac769bcfc029861b7b22e7c5949/pyomo/core/base/PyomoModel.py#L206) function to control these stuff. But I am not quite familiar ). 

I am now manually setting the `stale` flag to `True` before storing the solutions. Is it possible to add an option in the `store_to` function, which stops skipping the `stale=True` variables?

ZedongPeng:
## Summary/Motivation:
This Pull Request inserts add a `skip_stale_vars` option to control whether to skip the `stale=True` variables when storing variables.

- `store_to(self, results, cuid=False)` to `store_to(self, results, cuid=False, skip_stale_vars=True)`

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/tests/unit/test_model.py
+++ b/pyomo/core/tests/unit/test_model.py
@@ -266,14 +266,35 @@ class Test(unittest.TestCase):
         self.assertMatchesJsonBaseline(
             join(currdir,"solve1b.out"), join(currdir,"solve1b.txt"),
             tolerance=1e-4)
-
+            
+    def test_store_to_skip_stale_vars(self):
         # test store_to() function with skip_stale_vars=True
-        results = opt.solve(model)
-        model.solutions.store_to(results, skip_stale_vars=True)
-        results.write(filename=join(currdir,"solve1b.out"), format='json')
-        self.assertMatchesJsonBaseline(
-            join(currdir,"solve1b.out"), join(currdir,"solve1b.txt"),
-            tolerance=1e-4)
+        model = ConcreteModel()
+        model.A = RangeSet(1,4)
+        model.x = Var(model.A, bounds=(-1,1))
+        def obj_rule(model):
+            return sum_product(model.x)
+        model.obj = Objective(rule=obj_rule)
+        def c_rule(model):
+            expr = 0
+            for i in model.A:
+                expr += i*model.x[i]
+            return expr == 0
+        model.c = Constraint(rule=c_rule)
+        opt = SolverFactory('glpk')
+        results = opt.solve(model, symbolic_solver_labels=True)
+        model.solutions.store_to(results,skip_stale_vars=False)
+        for index in model.A:
+            self.assertIn(model.x[index].getname(), results.solution.variable.keys())
+        model.x[1].fix()
+        results = opt.solve(model, symbolic_solver_labels=True)
+        model.solutions.store_to(results,skip_stale_vars=True)
+        for index in model.A:
+            if index == 1:
+                self.assertNotIn(model.x[index].getname(), results.solution.variable.keys())
+            else:
+                self.assertIn(model.x[index].getname(), results.solution.variable.keys())
+
 
     def test_display(self):
         model = ConcreteModel()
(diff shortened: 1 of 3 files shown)


jsiirola:
I don't see a problem with this, but this PR should include a test to verify behavior with both `skip_stale_vars=True` and `skip_stale_vars=False`.

An alternative name could be `store_stale_vars`.  @blnicho, @carldlaird: thoughts?

blnicho:
> An alternative name could be `store_stale_vars`. @blnicho, @carldlaird: thoughts?

I think both names are clear and I don't have a strong preference. If I had to pick one I think I like `skip_stale_vars` slightly more than `store_stale_vars` just because I think the if-condition reads a little more clearly:

`if obj.stale and skip_stale_vars:` vs. `if obj.stale and not store_stale_vars:`

carldlaird:
I agree with @blnicho, but I have a strong preference for the default being skip_stale_vars=False. This way, the default is correct for those that have no idea what "stale" means.

ZedongPeng:
I'll agree with @blnicho and @carldlaird about the name. Should I also follow @carldlaird's suggestion about keeping `skip_stale_vars=False` by default? 

Btw, any suggestion about where to put the new tests? Thanks. @jsiirola

blnicho:
@ZedongPeng I think yes, we should try setting the default as `skip_stale_vars=False`. However, I'm a little worried that this might break some baseline tests that use the `store_to` method. Go ahead and try making this change and we'll help you resolve test failures if there are any.

I'd recommend putting the tests in `pyomo/core/tests/unit/test_model.py`. There are some tests near the bottom labeled `test_solve_with_store` and you can add an extra test there for this new option.

pyomo/core/tests/unit/test_model.py
```diff
@@ -267,6 +267,14 @@ def e_rule(model, i):
             join(currdir,"solve1b.out"), join(currdir,"solve1b.txt"),
             tolerance=1e-4)
 
+        # test store_to() function with skip_stale_vars=True
```
blnicho, line 270: You're on the right track with this test but I think it could be improved by moving it to its own test function and using `store_to` on a model that has both `stale=True` and `stale=False` variables (tested with both `skip_stale_vars=True` and `skip_stale_vars=False`). If there is a way to query the results object to verify whether or not it contains variables you know to be stale that would be a more robust test than the file baselines.

ZedongPeng: Hey @blnicho , I added a test function for `store_to` with `skip_stale_vars=True` and `skip_stale_vars=False`. Let me know if it needs to further modified. Thanks.

blnicho (approved):

jsiirola (approved):

PR #1682, opened 2020-10-23
Create Pyomo-specific unittest.TestCase and update fragile test

jsiirola:
## Summary/Motivation:
The `pyomo.core.tests.diet.test_diet:Test.test_sqlite_equality` test case was failing periodically on various platforms.  The root cause was that the test was comparing JSON result files and the `results['Solver'][0]['Time']` field was differing more than the allowable variance.  While we could have increased the variance, it would have become large enough that significant differences in the results would have still passed. 

This PR adds a new assertion, `assertRelativeEqual` that recursively compares two structured objects using a relative tolerance.  As it is aware of both sequences and mappings, it allows us to resolve the test failure in a clean way by loading the two JSON results files, removing the `Time` key and then performing a relative comparison of the remaining structure.

While we could have added this assertion to `pyutilib.th`, as part of the upcoming separation of Pyomo and PyUtilib, this PR sets the framework for future PRs that will migrate the necessary parts of `pyutilib.th` into `pyomo.common.testing`.  Currently, it declares a `TestCase` that inherits from `pyutilib.th.TestCase` and only implements the new `assertRelativeEqual` assertion.

- Create a `unittest.TestCase` wrapper in `pyomo.common.testing`
- Implement the `assertRelativeEqual` test assertion
- Resolve fragile tests in `pyomo.core.tests.diet.test_diet`

### Legal Acknowledgement


pyomo/common/unittest.py
```diff
@@ -0,0 +1,152 @@
...
+#  Under the terms of Contract DE-NA0003525 with National Technology and
+#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain
+#  rights in this software.
+#  This software is distributed under the 3-clause BSD License.
+#  ___________________________________________________________________________
+
+import math
+import six
+
+# Import base classes privately (so that we hae handles on them)
```
blnicho, line 14: Typo: have

pyomo/common/unittest.py
```diff
@@ -0,0 +1,152 @@
...
+        """Test that first and second are equal up to a relative tolerance
+
+        This compares first and second using a relative tolerance
+        (`delta`).  It will recursively decend into Sequence and Mapping
+        containers (allowing for the relative comparison of structured
+        data including lists and dicts).
+
+        If `places` is supplied, `delta` is computed as `10**-places`.
+
+        If neither `places` not `delta` is provided, delta defaults to 1e-7.
```
blnicho, line 51: Typo: nor

blnicho (approved):
I found a couple typos but otherwise I think this looks fine.

pyomo/common/unittest.py
```diff
@@ -0,0 +1,152 @@
...
+import six
+
+# Import base classes privately (so that we hae handles on them)
+import pyutilib.th.pyunit as _pyunit
+from pyutilib.th.pyunit import unittest as _unittest
+
+# Now, import the base unittest environment.  We will override things
+# specifically later
+from unittest import *
+from pyutilib.th import *
```
mrmundt, line 21: I will remind you - I did just get rid of the rest of the `import *` statement.

pyomo/common/unittest.py
```diff
@@ -0,0 +1,152 @@
...
+from pyomo.common.collections import Mapping, Sequence
+
+class TestCase(_pyunit.TestCase):
+    """A Pyomo-specific class whose instances are single test cases.
+
+    This class derives from unittest.TestCase and provides the following
+    additional functionality:
+      - extended suport for test categories
+      - additional assertions:
+        * :py:meth:`assertRelativeEqual`
```
mrmundt, line 32: I don't know if `meth` is a good word choice.

pyomo/common/unittest.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+    __doc__ += _unittest.TestCase.__doc__
+
+    def assertRelativeEqual(self, first, second,
+                            places=None, msg=None, delta=None,
+                            allow_second_superset=False):
+        """Test that first and second are equal up to a relative tolerance
+
+        This compares first and second using a relative tolerance
+        (`delta`).  It will recursively decend into Sequence and Mapping
```
mrmundt, line 45: Typo: descend

pyomo/common/unittest.py
jsiirola: Not our choice: this is a Sphinx RST flag: see the Sphinx [docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#role-py-meth)

pyomo/common/unittest.py
jsiirola: I debated this for a long time.  The challenge is that this module should be "unittest with extra bells and whistles".  We can explicitly list the objects from `unittest.__all__`, but I worry about "drift" in the underlying unittest library, where different versions have different exported symbols.  Thoughts?

(The `from pyutilib.th import *` will disappear as we complete the separation of Pyomo/PyUtilib and the required parts of pyutilib.th are moved here)

mrmundt: I'm fine with leaving this as-is. No need to do extra work re: pyutilib nonsense for now.

mrmundt (approved):

PR #1683, opened 2020-10-24
Add support for deferred import of submodules

jsiirola:
## Summary/Motivation:
The `pyomo.common.dependencies.attempt_import()` system greatly simplifies management of slow-loading optional dependencies.  However, recent work with #1526 highlighted a limitation when working with deferred imports of submodules.  This PR extends the deferred import system to allow for declaring specific submodules to be deferred as well (that is, accessing that attribute on a `DeferredImportModule` will not immediately cause that module to be imported.

The primary motivation for this change is `matplotlib`, where the standard usage is not `import matplotlib`, but rather `import matplotlib.pyplot as plt`.  Which this PR, users can now do the following:

```python
from pyomo.common.dependencies import matplotlib, matplotlib_available
plt = matpltolib.pyplot
```

`matplotlib` will not actually be imported until `plt` is used (e.g., with `plt.gca()` or `plt.plot([0,1],[2,3])`).  At that point, both the `matplotlib` and `plt` objects in the current `globals()` will be replaced by the appropriate modules (as is the normal case with `attempt_import`; thereby ensuring that using `attempt_import` has minimal performance impact).

- add a `defer_submodules` argument to `attempt_import` to support the deferred import of specific submodules
- add an `attempt_import` of `matplotlib` to `pyomo.common.dependencies`
- convert `parmest.graphics` to use the `scipy` / `numpy` / `matplotlib` from `pyomo.common.dependencies`
- add testing

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/dependencies.py
+++ b/pyomo/common/dependencies.py
@@ -11,6 +11,7 @@
 import inspect
 import importlib
 import logging
+from six import iteritems
 
 class DeferredImportError(ImportError):
     pass
@@ -48,12 +49,31 @@ class DeferredImportModule(object):
     attributes on this object will raise a DeferredImportError
     exception.
     """
-    def __init__(self, indicator):
+    def __init__(self, indicator, deferred_submodules, submodule_name):
         self._indicator_flag = indicator
+        self._submodule_name = submodule_name
+
+        if not deferred_submodules:
+            return
+        if submodule_name is None:
+            submodule_name = ''
+        for name in deferred_submodules:
+            if not name.startswith(submodule_name + '.'):
+                continue
+            _local_name = name[(1+len(submodule_name)):]
+            if '.' in _local_name:
+                continue
+            setattr(self, _local_name, DeferredImportModule(
+                indicator, deferred_submodules,
+                submodule_name + '.' + _local_name))
 
     def __getattr__(self, attr):
         self._indicator_flag.resolve()
-        return getattr(self._indicator_flag._module, attr)
+        _mod = self._indicator_flag._module
+        if self._submodule_name:
+            for _sub in self._submodule_name[1:].split('.'):
+                _mod = getattr(_mod, _sub)
+        return getattr(_mod, attr)
 
 
 class _DeferredImportIndicatorBase(object):
@@ -84,7 +104,8 @@ class DeferredImportIndicator(_DeferredImportIndicatorBase):
     """
 
     def __init__(self, name, alt_names, error_message, only_catch_importerror,
-                 minimum_version, original_globals, callback, importer):
+                 minimum_version, original_globals, callback, importer,
+                 deferred_submodules):
         self._names = [name]
         if alt_names:
             self._names += list(alt_names)
@@ -99,6 +120,7 @@ class DeferredImportIndicator(_DeferredImportIndicatorBase):
         self._importer = importer
         self._module = None
         self._available = None
+        self._deferred_submodules = deferred_submodules
 
     def resolve(self):
         if self._module is None:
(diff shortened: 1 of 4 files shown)


pyomo/common/dependencies.py
```diff
@@ -325,20 +392,31 @@ def _finalize_yaml(module, available):
...
+    'scipy', callback=_finalize_scipy, deferred_submodules={'stats':None})
 networkx, networkx_available = attempt_import('networkx', alt_names=['nx'])
 pandas, pandas_available = attempt_import('pandas', alt_names=['pd'])
 dill, dill_available = attempt_import('dill')
+
+# Note that matplotlib.pyplot can generate a runtime error on OSX when
+# not installed as a Framework (as is the case in the CI systems)
+matplotlib, matplotlib_available = attempt_import(
+    'matplotlib', only_catch_importerror=False, callback=_finalize_matplotlib,
+    deferred_submodules={'pyplot': ['plt']})
```
blnicho, line 422: If a developer doesn't follow the standard naming convention for this submodule (i.e. `import matplotlib.pyplot as pt`) what will happen? Will it just bypass the deferred import stuff altogether and eventually fail the test checking the import time?

pyomo/common/dependencies.py
```diff
@@ -325,20 +392,31 @@ def _finalize_yaml(module, available):
...
 
 def _finalize_pympler(module, available):
     if available:
         # Import key subpackages that we will want to assume are present
         import pympler.muppy
 
+def _finalize_matplotlib(module, available):
+    if available:
+        import matplotlib.pyplot
+
```
kaklise, line 407: Since pyplot is imported when finalizing matplotlib, the submodule import is not deferred.  Is that intended?

jsiirola: The submodule import is deferred until the main module import is triggered.  So, given:
```python
from pyomo.common.dependencies import matplotlib
plt = matplotlib.pyplot # neither matplotlib nor pyplot are imported
plt.plot([0,1],[2,3]) # both matplotlib and matplotlib.pyplot are imported; plot() works as expected
```
and 
```python
from pyomo.common.dependencies import matplotlib`
plt = matplotlib.pyplot # neither matplotlib nor pyplot are imported
matplotlib.locale # both matplotlib and matplotlib.pyplot are imported
```
The only downside for the `_finalize_matplotlib` approach is that if *any* part of matplotlib is needed, then pyplot will be imported.  When I looked at things, that didn't seem to be a big hit, as pyplot was *always* used if someone was using matplotlib (and this was way easier than implementing a more complex layered deferred import system).

kaklise (approved):
This looks good to me.  Thanks for updating parmest.

pyomo/common/dependencies.py
jsiirola: If the developer runs
```python
import matplotlib.pyplot as pt  # or even as plt
```
then this is a direct Python import: matplotlib and pyplot will be immediately imported.  If a user tries to do this from dependencies with something like:
```python
import pyomo.common.dependencies.matplotlib.pyplot as plt
```
they will get an import error.  Similarly, 
```python
from pyomo.common.dependencies import matplotlib.pyplot as plt
```
will get a parse error.  Finally (what I think you were originally asking), if they:
```python
from pyomo.common.dependencies import matplotlib
pt = matplotlib.pyplot
```
Then `pyplot` will not be imported until the user accesses an attribute on `pt`, at which point the module is imported and the attribute is returned.  The only difference between `plt` and `pt` is that because `pt` is not one of the known alternate names, then the deferred import occurs, the deferred importer cannot replace `pt` in the module `globals()` dictionary.  Subsequent calls to `pt.<any attribute>` will work just find, but will incur a slight overhead as they have to go through the `__getattr__` on `pt` (because `pt` is still an instance of `DeferredImportModule`)

blnicho (approved):

PR #1684, opened 2020-10-26
Fix integer conversion of string arguments in ProgressiveHedging

jetuk:
## Summary/Motivation:

The comment `ProgressiveHedging` suggests that an attempt is made to convert string arguments to numbers. It says that integer is tried first, and then float. However, the implementation is opposite and consequently all integer arguments are converted to floats. This has caused a problem in my CBC model when I try to pass an integer value.

- The changes here switch the order of the conversion attempts to reflect the written comment and allow integer arguments to be passed correctly.

### Legal Acknowledgement

### DIFF
--- a/pyomo/pysp/ph.py
+++ b/pyomo/pysp/ph.py
@@ -1945,10 +1945,10 @@ class ProgressiveHedging(_PHBase):
         # and in that order.
         def convert_value_string_to_number(s):
             try:
-                return float(s)
+                return int(s)
             except ValueError:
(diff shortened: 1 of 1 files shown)


DLWoodruff (approved):
This looks good.

blnicho (approved):

jsiirola:
I see no problem with this fix (and we can merge it), but am mildly surprised that the code is never executed by any PySP test.

PR #1703, opened 2020-11-12
Block initialization improvement

jsiirola:
## Summary/Motivation:
Block was inconsistent in how it applied certain initialization routines.  This made it particularly onerous when declaring derived Block classes that wanted to perform additional initialization (outside of a rule).  This PR implements a general cleanup of the Block initialization, including:
- switch to using the `Initializer()` function
- deprecation of the (best that I can tell completely unused) `options=` argument
- centralizing all block data initialization in `_getitem_when_not_present`

- switch Block to leverage `Initializer()`
- standardize initialization in `_getitem_when_not_present`
- resolve edge cases for constructing nested abstract blocks (and add tests)
- deprecate the `Block(options=...)` argument
- add a `Block(dense=...)` argument to control if indexed blocks are created densely when no rule is provided (default = `True`)
- general cleanup of imports
- switch to using the `@deprecated()` decorator

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/block.py
+++ b/pyomo/core/base/block.py
@@ -1844,6 +1844,9 @@ class Block(ActiveIndexedComponent):
         # As concrete applies to the Block at declaration time, we will
         # not use an initializer.
         _concrete = kwargs.pop('concrete', False)
+        # As dense applies to the whole container, we will not use an
+        # initializer
+        self._dense = kwargs.pop('dense', True)
         kwargs.setdefault('ctype', Block)
         ActiveIndexedComponent.__init__(self, *args, **kwargs)
         if _options is not None:
@@ -1938,7 +1941,8 @@ class Block(ActiveIndexedComponent):
         try:
             if self.is_indexed():
                 # We can only populate Blocks with finite indexing sets
-                if self.index_set().isfinite():
+                if self.index_set().isfinite() and (
+                        self._dense or self._rule is not None):
                     for _idx in self.index_set():
                         # Trigger population & call the rule
                         self._getitem_when_not_present(_idx)
--- a/pyomo/core/base/block.py
+++ b/pyomo/core/base/block.py
@@ -24,14 +24,9 @@ from operator import itemgetter
 from six import iteritems, iterkeys, itervalues, StringIO, string_types, \
     advance_iterator, PY3
 
-if PY3:
-    from collections.abc import Mapping as collections_Mapping
-else:
-    from collections import Mapping as collections_Mapping
-
 from pyutilib.misc.indent_io import StreamIndenter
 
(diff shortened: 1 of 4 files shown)


pyomo/core/base/block.py
```diff
@@ -1857,22 +1880,27 @@ def _getitem_when_not_present(self, idx):
...
-                self, self._rule, _block, idx, self._options)
+            obj = self._rule(_block, idx)
+            # If the user returns a block, transfer over everything
+            # they defined into the empty one we created.  We do
+            # this inside the try block so that any abstract
+            # components declared by the rule have the opportunity
+            # to be initialized with data from
+            # _BlockConstruction.data as they are transferred over.
+            if obj is not _block and isinstance(obj, _BlockData):
+                _block.transfer_attributes_from(obj)
```
michaelbynum, line 1899: Just for my understanding, does this work for classes that inherit from blocks? I think so...

pyomo/core/base/block.py
```diff
@@ -1836,11 +1839,31 @@ def __new__(cls, *args, **kwds):
...
-        self._rule = kwargs.pop('rule', None)
-        self._options = kwargs.pop('options', None)
+        _rule = kwargs.pop('rule', None)
+        _options = kwargs.pop('options', None)
+        # As concrete applies to the Block at declaration time, we will
+        # not use an initializer.
         _concrete = kwargs.pop('concrete', False)
+        # As dense applies to the whole container, we will not use an
+        # initializer
+        self._dense = kwargs.pop('dense', True)
```
michaelbynum, line 1849: I don't think I understand the implications of this. Were blocks not dense before?

michaelbynum (approved):

blnicho (approved):

pyomo/core/base/block.py
jsiirola: Blocks were dense if you specified a rule (and indexed them with a finite set).  This was unfortunate, as `Complementary` relied on the face that if the rule was omitted the Block was sparse.  This does change the default behavior so that Blocks (indexed by finite sets) are not dense by default.  I don't think that it really changes anything for people, especially as I don't know of anyone who was relying on that particular edge case.

pyomo/core/base/block.py
jsiirola: Yes.  This should work for derived blocks.  That said, if the derived block class has non-component attributes that hold state information that shouldn't be moved over, then they need to define a `_Block_reserved_words` class attribute that extends the set defined on `Block._Block_reserved_words`.

michaelbynum:
@jsiirola Thanks for the explanations. They were very helpful.

PR #1713, opened 2020-11-17
Update ComponentUID to improve serialization and leverage slicing

### LINKED ISSUE #1069: CUIDs for components indexed by Any
It seems that Component UIDs don't behave the same way when you index by Any. I ran into this with Disjunctions, but here's a simple example with Constraints:
```
from pyomo.environ import *

m = ConcreteModel()
m.x = Var()
m.a = Constraint(Any)
@m.Constraint([0,1])
def b(m, i):
    return m.x >= i

m.a[0] = m.x >= 8

m.pprint()

if ComponentUID(m.b[0]).find_component(m) is m.b[0]:
    print("Life is good")

if ComponentUID(m.a[0]).find_component(m) is None:
    print("Should this happen?")
```
Both of the prints statements run... I should expect m.a to behave like m.b, right?

### LINKED ISSUE #1312: Issues with find_component
The first issue is that you can't call `find_component` from a `_BlockData` object. The following code raises an attribute error:

```python
from pyomo.environ import *

m = ConcreteModel()

@m.Block([1,2,3])
def b1(b, i):
    b.v1 = Var()

vname = m.b1[1].v1.name
comp = m.b1[1].find_component(vname)
```

The second issue is that `find_component` doesn't work with decimal indices. 

```python
from pyomo.environ import *

m = ConcreteModel()

m.t = Set(initialize=[1.0, 1.5, 2.0])
m.x = Set(initialize=[1,2,3])

@m.Block(m.t, m.x)
def b1(b, t, x):
    b.v1 = Var()
```

After the above code, the following `find_component` fails:

```python
>>> m.b1[1,1].v1.name
'b1[1.0,1].v1'
>>> vname1 = m.b1[1,1].v1.name
>>> m.find_component(vname1) is None
True
```

while the following works as expected:

```python
>>> m.find_component('b1[1,1].v1') is m.b1[1.0,1.0].v1
True
```

This is especially problematic for `ContinuousSet`s, where decimal indices are more common.

jsiirola:
## Summary/Motivation:
The original implementation of `ComponentUID` was fragile and contained a number of known edge case errors.  In particular, it did not cleanly support all numeric values, strings with `.`, and had no support for index types other than `str`, `int`, and `float`.  

This PR reworks the `ComponentUID` implementation to make use of component slicing (instead of reimplementing a nearly identical functionality) and to resolve all known edge cases from the previous implementation.

- Reverse the internal `_cids` list to match the tree order / order in Component slices
- Resolve edge cases
  - `.` in string indices
  - ambiguity between `1`, `'1'`, and `(1,)`
  - support for `float('inf')`
  - support for index types other than `str`, `int`, and `float`
- introduce a new string format and a more robust string parser
- improve testing

### Legal Acknowledgement


Robbybp (commented):
A typo and a couple questions but I'm very happy about these changes.

Robbybp:
This will address part of #1312

pyomo/core/base/componentuid.py
```diff
@@ -0,0 +1,576 @@
...
+    The CUID has a string representation that can specify a specific
+    component or a group of related components through the use of index
+    wildcards (* for a single element in the index, and ** for all
+    indexes)
+    """
+
+    __slots__ = ( '_cids', )
+
+    def _safe_str_tuple(x):
+        return '(' + ','.join(ComponentUID._safe_str(_) for _ in x) + ',)'
```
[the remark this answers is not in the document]
jsiirola: Yes.  I had originally tried, but the naive implementation generated errors.  After a little digging I figured out how to make it a proper `@staticmethod`

pyomo/core/base/componentuid.py
```diff
@@ -0,0 +1,576 @@
...
+
+from six import PY2, string_types, iteritems
+
+from pyomo.common.collections import ComponentMap
+from pyomo.common.dependencies import pickle
+from pyomo.common.deprecation import deprecated
+from pyomo.core.base.indexed_component_slice import IndexedComponent_slice
+
+class _PickleEllipsis(object):
+    "A work around for the non-pigklability of Ellipsis in Python 2"
```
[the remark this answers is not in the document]
jsiirola: Yes.  Fixed.

pyomo/core/base/componentuid.py
```diff
@@ -0,0 +1,576 @@
...
+    """
+
+    __slots__ = ( '_cids', )
+
+    def _safe_str_tuple(x):
+        return '(' + ','.join(ComponentUID._safe_str(_) for _ in x) + ',)'
+
+    @staticmethod
+    def _pickle(x):
+        return '|'+repr(pickle.dumps(x, protocol=2))
```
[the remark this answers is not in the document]
jsiirola: Yes.  As both string indices and pickled "other" indices are embedded as strings, the "`|`" allows the parser to distinguish between them.

pyomo/core/base/componentuid.py
```diff
@@ -0,0 +1,576 @@
...
+            for self_i, other_i in zip(self_idx, other_idx):
+                if self_i != other_i:
+                    if other_i is Ellipsis:
+                        return True
+                    if self_i is Ellipsis:
+                        return False
+                    if other_i.__class__ is slice:
+                        return True
+                    if self_i.__class__ is slice:
+                        return False
```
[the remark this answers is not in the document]
jsiirola: Good catch.  Resolved.

pyomo/core/base/componentuid.py
```diff
@@ -0,0 +1,576 @@
...
+    def find_component_on(self, block):
+        """
+        Return the (unique) component in the block.  If the CUID contains
+        a wildcard in the last component, then returns that component.  If
+        there are wildcards elsewhere (or the last component was a partial
+        slice), then returns None.  See list_components below.
+        """
+        obj = self._resolve_cuid(block)
+        if isinstance(obj, IndexedComponent_slice):
+            return None
```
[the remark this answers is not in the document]
jsiirola: I had thought about the same thing and originally elected to leave the behavior unchanged.  However,I think that is sufficiently useful that the API change is a good idea.

This turned out to be a little challenging to implement because it introduced a circular import for Python < 3.7.  I figured out how to make a more robust `relocated_module_attribute()` method that breaks the circular import for all versions of Python.

pyomo/core/base/componentuid.py
```diff
@@ -74,7 +75,8 @@ def _safe_str(x):
         int: repr,
         float: repr,
         str: repr,
-        tuple: _safe_str_tuple,
+        # Note: the function is unbound at this point; extract with __func__
+        tuple: _safe_str_tuple.__func__,
```
Robbybp: Oh, didn't know this would be necessary

pyomo/common/deprecation.py
```diff
@@ -97,3 +108,126 @@ def wrapper(*args, **kwargs):
...
+            target_obj = _import_object(name, *info)
+            setattr(self, name, target_obj)
+            return target_obj
+        raise AttributeError("module '%s' has no attribute '%s'"
+                             % (self.__name__, name))
+
+def relocated_module_attribute(local, target, version, remove_in=None):
+    """Provide a deprecation path for moved / renamed module attributes
+
+    This runction declares that a local module attribute has been moved
```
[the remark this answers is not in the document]
jsiirola: Fixed.

jsiirola:
@Robbybp: I believe this PR fixes both the parts of #1312 (find_component now correctly handles floating point indices)..

Robbybp:
@jsiirola Oh didn't realize find_component was moved to _BlockData here. You're right.

pyomo/core/base/componentuid.py
```diff
@@ -0,0 +1,589 @@
...
+    def __gt__(self, other):
+        "Return True if this CUID > the 'other' CUID"
+        return not (self <= other)
+
+    def __ge__(self, other):
+        "Return True if this CUID >= the 'other' CUID"
+        return not (self < other)
+
+    def __eq__(self, other):
+        """Return True is this CUID is exactly equal to `other`
```
blnicho, line 224: Typo: "**is** this" should be "**if** this"

pyomo/core/base/componentuid.py
```diff
@@ -0,0 +1,589 @@
...
+        is not a ComponentUID.
+        """
+        try:
+            other_cids = other._cids
+        except AttributeError:
+            return False
+        return self._cids == other_cids
+
+    def __ne__(self, other):
+        """Return True is this CUID is not exactly equal to `other`
```
blnicho, line 236: Typo: "**is** this" should be "**if** this"

pyomo/core/base/componentuid.py
```diff
@@ -0,0 +1,589 @@
...
+
+    def _generate_cuid(self, component, cuid_buffer=None, context=None):
+        "Return the list of (name, idx) pairs for the specified component"
+        model = component.model()
+        if context is None:
+            context = model
+        orig_component = component
+        # if component.is_indexed():
+        #     yield( component.local_name, (Ellipsis,))
+        #     component = component.parent_block()
```
blnicho, line 306: Can these lines be removed?

pyomo/core/tests/unit/test_componentuid.py
```diff
@@ -0,0 +1,891 @@
...
+        self.assertIs(cList[0], self.m.b[1,'2'].c.a)
+
+    def test_find_wildcard_not_exists(self):
+        cuid = ComponentUID('b[*,*].c.x')
+        self.assertIsNone(cuid.find_component_on(self.m))
+
+    # def test_find_implicit_exists(self):
+    #     cuid = ComponentUID('b:1,2.c.a:3')
+    #     self.assertTrue(cuid.find_component_on(self.m) is
+    #                     self.m.b[1,'2'].c.a[3])
```
blnicho, line 312: Why is this test commented out?

pyomo/core/tests/unit/test_componentuid.py
```diff
@@ -0,0 +1,891 @@
...
+
+    def test_matches_explicit(self):
+        cuid = ComponentUID(self.m.b[1,'2'].c.a[3])
+        self.assertTrue(cuid.matches(self.m.b[1,'2'].c.a[3]))
+        self.assertFalse(cuid.matches(self.m.b[1,'2'].c.a['2']))
+
+    # def test_matches_implicit(self):
+    #     cuid = ComponentUID('b:1,2.c.a:3')
+    #     self.assertTrue(cuid.matches(self.m.b[1,'2'].c.a[3]))
+    #     self.assertFalse(cuid.matches(self.m.b[1,'2'].c.a['2']))
```
blnicho, line 381: Why is this test commented out?

pyomo/core/tests/unit/test_componentuid.py
jsiirola: ComponentUID used to support "fuzzy matching" of component names.  That is, the "name" of a component is ambiguous:
```
m.x = Var(Any, dense=False)
m.x[1].name == "x[1]"  # True!
m.x['1'].name == "x[1]"  # True!
m.x['1'] is not m.x[1]  # True!
```
The old ComponentUID would take strings like `"x[1]"` and attempt to perform "implicit" matching (looking first for `1` and then for `'1'`).  This PR removed that functionality: it was too prone to generating error, plus ComponentUID can now correctly parse quoted indices, so it is not as big of an issue.  A future PR will update `Component.name` so that ambiguous strings (containing spaces or commas, or consisting of parsable floating point numbers) will always be quoted.

pyomo/core/tests/unit/test_componentuid.py
jsiirola: See above comment: ComponentUID no longer supports "fuzzy" matching of ambiguous strings.

blnicho (approved):

PR #1724, opened 2020-11-24
PyNumero Updates

michaelbynum:
## Summary/Motivation:
This PR primarily makes performance enhancements to `pynumero.sparse`. There are a couple other minor changes listed below.

- Add functionality to `HierarchicalTimer` to iterate over all timers.
- HSL's MA27 is actually very robust to changes in the ordering of the row indices and col indices between the symbolic and numeric factorizations. It even allows the length of these arrays to be different, as long as the symbolic factorization had all of the possible nonzero entries. Given this, I removed some of the assert statements in the ma27 wrapper.
- The rest of the changes are `pynumero.sparse` performance enhancements

### Legal Acknowledgement


blnicho (approved):

blnicho:
Here is the output from the failing Jenkins test:

```
 FAIL: test_do_numeric_factorization (test_ma27.TestMA27Interface)
   File ".../pyomo/pyomo/contrib/pynumero/linalg/tests/test_ma27.py", line 107, in test_do_numeric_factorization
    ma27.do_numeric_factorization(irn, icn, n, bad_ent)
 AssertionError: AssertionError not raised
```

jsiirola (approved):
I think this looks good, pending the resolution of the ma27 test.

I do have a question around the `assert_correct_owners` API, though: what is the intent of the flag?  I noticed that this PR changes the default from `True` to `False`.  Is there any situation where you would want to set this `True`?  Is this really more of a debugging flag?  It seems to be a strange thing to add to the interface.  I wonder if a cleaner solution would be a global (debugging) flag, instead of something that needs to be passed through all function calls?

michaelbynum:
@jsiirola It is a debugging flag. I think there is a lot of debugging infrastructure that we need to put in, but I think it will come in a separate PR.

I'll get the test fixed shortly.

PR #1745, opened 2020-12-14
Fix cloning AMPLExternalFunction objects after the external library is loaded

jsiirola:
## Summary/Motivation:
`AMPLExternalFunction` objects contain "uncopyable" attributes after teh SO/DLL has been loaded.  This PR updates `AMPLExternalFunction.__getstate__()` to *not* copy the `_so` and `_known_functions` attributes.  This makes the resulting state copyable/picklable, but forces the clone to re-load the external library.

This bug was originally reported by @esrawli when using GDPopt with models containing compiled external functions.

- Map the `AMPLExternalFunction` `_so` and `_known_functions` attributes to `None` when deepcopying/cloning/pickling the component.
- Add a test to verify the previous error is resolved.

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/external.py
+++ b/pyomo/core/base/external.py
@@ -111,6 +111,13 @@ class AMPLExternalFunction(ExternalFunction):
         self._so = None
         ExternalFunction.__init__(self, *args, **kwds)
 
+    def __getstate__(self):
+        state = super(AMPLExternalFunction, self).__getstate__()
+        # Remove reference to loaded library (they are not copyable or
+        # picklable)
+        state['_so'] = state['_known_functions'] = None
+        return state
+
     def _evaluate(self, args, fgh, fixed):
         if self._so is None:
             self.load_library()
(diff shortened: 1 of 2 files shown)


pyomo/core/base/external.py
```diff
@@ -111,6 +111,13 @@ def __init__(self, *args, **kwds):
         self._so = None
         ExternalFunction.__init__(self, *args, **kwds)
 
+    def __getstate__(self):
+        state = super(AMPLExternalFunction, self).__getstate__()
+        # Remove reference to loaded library (they are not copyable or
+        # picklable)
+        state['_so'] = state['_known_functions'] = None
+        return state
+
```
carldlaird, line 120: I assume that removing the reference to the loaded library ensures it will be reloaded when required?

carldlaird (approved):
Approved pending successful tests and addressing of comment.

pyomo/core/base/external.py
jsiirola: That is correct.  Both `_so` and `_known_functions` are initialized to `None` in `__init__()`, and `_evaluate()` begins with:
```python
if self._so is None:
    self.load_library()
```

PR #1768, opened 2021-01-03
Extend flattener to allow multiple sets

Robbybp:
## Summary/Motivation:
The current `pyomo.dae.flatten` module has a function `flatten_dae_components` that partitions indexed components into new (reference) components, indexed only (if at all) by a specified set. This PR extends this functionality by allowing multiple sets to be specified. The returned components may be indexed by any or all of the specified sets. This is useful especially for PDE modeling/analysis, where a unique "variable" is defined as something indexed by e.g. time and space.

- Overhaul flattener helper functions to allow the more general functionality
- Add a user-facing function `flatten_components_along_sets` that implements this partition
- Rewrite `flatten_dae_components` as a special case of `flatten_components_along_sets`

### Legal Acknowledgement


pyomo/dae/flatten.py
```diff
@@ -7,147 +7,371 @@
...
+    if len(filled_index) == 1:
+        return filled_index[0]
+    else:
+        return filled_index
+
+def _fill_indices_from_product(partial_index_list, product):
+    """ 
+    `partial_index_list` is a list of indices, each corresponding to a
+    set. If an entry in `partial_index_list` is `_NotAnIndex`, that
+    slot will get filled in by an entry from `setprod`.
```
blnicho, line 73: Should `setprod` be `product`?

pyomo/dae/flatten.py
```diff
@@ -7,147 +7,371 @@
...
+    hierarchy. Note that `b` is a data object, so any index
+    of its parent component should be included in the stack.
+
+    `_slice` is the slice generated so far. Our goal here is to
+    yield extensions to `_slice` at this level of the hierarchy.
+
+    `ctype` is the type we are looking for.
+
+    `index_map` is potentially a map from each set in `sets` to a "representative
+              index" to use if we ever have
```
blnicho, line 137: Is this description incomplete?

pyomo/dae/flatten.py
```diff
@@ -7,147 +7,371 @@
...
+
+            normalize_index.flatten = False
+            # Want to get the unflattened factors when we advance the
+            # iterator of `product`
+    finally:
+        # Reset `normalize_index.flatten`
+        normalize_index.flatten = _normalize_index_flatten
+
+def generate_sliced_components(b, index_stack, _slice, sets, ctype, index_map):
+    """
```
blnicho, line 121: This docstring is missing a description of the `sets` argument

pyomo/dae/tests/test_flatten.py
```diff
@@ -12,7 +12,92 @@
 from pyomo.environ import ConcreteModel, Block, Var, Reference, Set, Constraint
 from pyomo.dae import ContinuousSet
 # This inport will have to change when we decide where this should go...
```
blnicho, line 14: Typo: import

pyomo/dae/tests/test_flatten.py
```diff
@@ -208,7 +293,997 @@ def c_rule(m, t):
...
+
+        #for sets, comps in zip(sets_list, comps_list):
+        #    if len(sets) == 1 and sets[0] is m.d2:
+        #        ref_data = {
+        #                self._hashRef(Reference(m.v_2n[:,('a',1)])),
+        #                self._hashRef(Reference(m.v_2n[:,('b',2)])),
+        #                }
+        #        # Cannot access the data of ^these references
+        #        for comp in comps:
+        #            self.assertIn(self._hashRef(comp), ref_data)
```
blnicho, line 598: Can this commented code be removed? At a higher level, is this test complete and doing what you expect? The comments seem to indicate that there may have been some issues debugging it.

blnicho (commented):
A few minor comments but otherwise I think this looks fine.

pyomo/dae/flatten.py
```diff
@@ -7,147 +7,371 @@
...
-                    for _slice in generate_time_only_slices(v, time):
-                        time_indexed_comps.append(Reference(_slice))
+    j = 0
+    for i, val in enumerate(filled_index):
+        if val is _NotAnIndex:
+            filled_index[i] = index[j]
+            # `index` is always a tuple, so this is valid
+            j += 1
+    # Make sure `filled_index` had the same number of vacancies
+    # as `product` has factors. Not _strictly_ necessary.
```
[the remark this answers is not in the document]
Robbybp: `product` has the same number of factors as the length of `index`. Changed the wording to not reference a variable from another function, though.

pyomo/dae/flatten.py
Robbybp: Yep, changed.

pyomo/dae/flatten.py
Robbybp: Yes, updated.

pyomo/dae/flatten.py
Robbybp: Added.

pyomo/dae/tests/test_flatten.py
Robbybp: Don't think this comment is relevant anymore, deleted.

pyomo/dae/tests/test_flatten.py
Robbybp: This test is basically not doing what is expected. The line
```python
self.assertNotIn(('a', 1), ref1._index._ref)
```
confirms this. `('a', 1)` should be a valid index. This is a problem with compatibility between slices and `normalize_index.flatten == False`. Basically a one-dimensional slice does not properly slice a higher-dimension set in this case. I'm not sure there is an easy fix. The commented code is what I would have expected to work.

Probably should just open an issue about this then remove this test?

blnicho: Thanks for the clarification. I think the best thing to do is to open an issue and then add a "TODO" comment in this test that references the issue. I think leaving the commented test for future reference is fine as long as we can follow the context of why it's there.

pyomo/dae/flatten.py
```diff
@@ -7,147 +7,373 @@
...
+    with the sets that were sliced for each component.
+
+    `b` is a _BlockData object.
+
+    `index_stack` is a list of indices "above" `b` in the
+    hierarchy. Note that `b` is a data object, so any index
+    of its parent component should be included in the stack.
+
+    `_slice` is the slice generated so far. Our goal here is to
+    yield extensions to `_slice` at this level of the hierarchy.
```
jsiirola, line 132: If the intent of this argument is to be "public", I would recommend making it `slice_` and not `_slice`.

pyomo/dae/tests/test_flatten.py
```diff
@@ -208,7 +292,997 @@ def c_rule(m, t):
...
-    # TODO: Add tests for Sets with dimen==None
+class TestFlatten(TestCategorize):
+
+    def _hashRef(self, ref):
+        if not ref.is_indexed():
+            return (id(ref),)
+        else:
+            return tuple(sorted(id(_) for _ in ref.values()))
+
+    def _model_1(self):
```
jsiirola, line 303: Can you provide more descriptive model names (maybe incorporate some of the subsequent comment into the function name; possibly something like `_model1_1D_sets`)

pyomo/dae/tests/test_flatten.py
```diff
@@ -208,7 +292,997 @@ def c_rule(m, t):
...
+                }
+                ref_data.update(self._hashRef(Reference(m.v_tst[:,x,:])) for x in S)
+                ref_data.update(self._hashRef(Reference(m.b.b2[:,x].v2[:,j])) for x, j in SC)
+                assert len(comps) == len(ref_data)
+                for comp in comps:
+                    self.assertIn(self._hashRef(comp), ref_data)
+            else:
+                raise RuntimeError()
+
+    def _model_2(self):
```
jsiirola, line 534: Can you provide more descriptive model names (maybe incorporate some of the subsequent comment into the function name; possibly something like `_model2_nD_sets`)

pyomo/dae/tests/test_flatten.py
```diff
@@ -208,7 +292,997 @@ def c_rule(m, t):
...
+                ref_data.update(self._hashRef(Reference(m.b[:,i2,i_n].v1[:])) for i2, i_n in D2N)
+                assert len(ref_data) == len(comps)
+                for comp in comps:
+                    self.assertIn(self._hashRef(comp), ref_data)
+            else:
+                raise RuntimeError()
+
+        normalize_index.flatten = True
+
+    def _model_3(self):
```
jsiirola, line 656: Can you provide more descriptive model names (maybe incorporate some of the subsequent comment into the function name; possibly something like `_model2_nD_flattened_sets`)

pyomo/dae/tests/test_flatten.py
```diff
@@ -208,7 +292,997 @@ def c_rule(m, t):
...
+                        self._hashRef(Reference(m.b[:,:].v1)),
+                        self._hashRef(incomplete_ref),
+                        }
+                self.assertEqual(len(ref_data), len(comps))
+                for comp in comps:
+                    self.assertIn(self._hashRef(comp), ref_data)
+            else:
+                raise RuntimeError()
+
+    def _model_4(self):
```
jsiirola, line 1101: Can you provide more descriptive model names (maybe incorporate some of the subsequent comment into the function name; possibly something like `_model4_slice_3_sets`)

pyomo/dae/tests/test_flatten.py
Robbybp: Added comment and opened issue

blnicho (approved):

jsiirola (commented):
I think this looks good (up do some questions on naming / documentation).

I do have a side question for you: how challenging do you think it would be it be to add a wrapper object around ReferenceDict to support reordering the indices of a reference?  That way you wouldn't have to create separate results for `[I, J, I]` and `[I, I, J]`.  I suspect we could do that pretty easily (in a separate PR).

Robbybp:
This may be a larger can of worms than I should open right now, but I'm not sure I like the name `flatten_components_along_sets`. "Flatten" seems to imply that we are going from a "higher dimension" object to a "lower dimension" object by multiplying sets into a single index, like we might do for finite elements and collocation points. 

This functionality seems closer to a partition of component data objects into containers that preserve indexing by the specified sets. Any thoughts on the name `partition_components_by_sets`?

jsiirola:
> This may be a larger can of worms than I should open right now, but I'm not sure I like the name `flatten_components_along_sets`. "Flatten" seems to imply that we are going from a "higher dimension" object to a "lower dimension" object by multiplying sets into a single index, like we might do for finite elements and collocation points.

I think this *does* move from a higher dimension down to a lower one, in that all the index sets *other* than the ones you are flattening along disappear.

Robbybp:
> I think this _does_ move from a higher dimension down to a lower one, in that all the index sets _other_ than the ones you are flattening along disappear.

I usually think of flattening as going from a single 2d array to a single 1d array. Here we are going from high-dimensional objects to many low-dimensional objects. I suppose it still is flattening but it seems to be closer to some sort of partition. It feels like cheating to say we "flattened" an object when really we broke it up enough that we could call each piece flat.

pyomo/dae/flatten.py
Robbybp: Done

pyomo/dae/tests/test_flatten.py
Robbybp: Done

pyomo/dae/tests/test_flatten.py
Robbybp: Done

pyomo/dae/tests/test_flatten.py
Robbybp: Done

pyomo/dae/tests/test_flatten.py
Robbybp: Done

jsiirola (approved):

Robbybp:
After a little more thought, I think we are actually doing three things here. First we split components by eliminating any sets we don't care about, second we partition these components according to what sets they're still indexed by, and third we create references that allow the user to access the new components directly.

The first and second steps can be thought of as partitions, the third as flattening. I think the entire procedure can be thought of as viewing components "from the perspective of" the specified sets. Not sure how much this helps naming. Best I can come up with is something like
`view_components_from_sets` or
`view_components_through_sets`.

blnicho:
I suggest we keep the current name for now. This is a fairly advanced and specific tool and I think expanding the documentation to include the purpose and intended use of this functionality will be more helpful to "future-us" than trying to find a more precise name at this point.

PR #1774, opened 2021-01-05
Parmest pairwise plot using the covariance matrix

kaklise:
## Summary/Motivation:
Updates to use the covariance matrix in parmest pairwise graphics.

- Updated ``graphics.pairwise_plot`` to make use of a mean and covariance in the pairwise plotting function.  Added an example to the end of parmest_example_cov.py
-  The covariance matrix, returned by ``pest.theta_est(calc_cov=True)`` data type was changed from a numpy array to a DataFrame using the variable names as index/columns.  
- Updated the way graphics are import.  Graphics functions are now accessed using ``parmest.graphics``.
- Updated API docs and a small update to an online docs rst file
- Merged recent updates from #1770 

### Legal Acknowledgement


blnicho (approved):

pyomo/contrib/parmest/examples/reactor_design/datarec_example.py
```diff
@@ -11,7 +11,7 @@
 import numpy as np
 import pandas as pd
 import pyomo.contrib.parmest.parmest as parmest
-from .reactor_design import reactor_design_model
+from reactor_design import reactor_design_model
```
jsiirola, line 14: Why are you changing this from a relative import to an absolute import?  I would think this could cause problems (i.e., if someone were to define a `reactor_design` package).  I think if you are not going to use relative imports, then this needs to be a fully-qualified absolute import.

jsiirola (commented):
I think this looks good, but have one question on an import change before we merge.

pyomo/contrib/parmest/examples/reactor_design/datarec_example.py
kaklise: That makes sense.  After looking over the parmest examples again, I see that most examples import the model using an absolute import.  I'll update the examples so that these imports are all relative.

kaklise: I tested that out and I get an Import Error using relative import (ImportError: attempted relative import with no known parent package).  The datarec example was the only file that had a relative import.

jsiirola: @kaklise: yes: if you attempt to directly run a python file (e.g., `python datarec_example.py`), then you cannot use relative imports in the file.  What I was suggesting is that instead of
```python
 from reactor_design import reactor_design_model 
```
(an implicit relative import), you use an absolute import:
```python
 from pyomo.contrib.parmest.examples.reactor_design import reactor_design_model 
```

jsiirola (approved):

PR #1778, opened 2021-01-07
parmest with, or without, mpi-sppy 

DLWoodruff:
- drop use of PySP
- add optional use of mpi-sppy
- the bad news: use two files of code copied from mpi-sppy (this is legal, but less than ideal) so that we can avoid use of  mpi-sppy but still have all the hooks
- TBD: if we like this, we will want to add tests that do depend on the presence of mpi-sppy, which is easy to assure in the test environment (it is pip installable); however, for the time being, those tests should probably run in mpi-sppy.

### Legal Acknowledgement


pyomo/contrib/parmest/tests/edgecase.py
```diff
@@ -0,0 +1,111 @@
...
+        model.Boiler[1].P_in[t] * model.Boiler[1].cost_per_kWh * model.cost_error for t in model.T))
+    return model
+
+
+def parameter_estimation():
+    """Estimate Parameter Boiler[1].eta"""
+    data = {'global': {'T': [0, 1, 2, 3]},
+            'Boiler': {1: {'cost_per_kWh': {None: 1},
+                           'eta': {None: 10}}},  # during construction eta will be fixed to wrong value.
+            # paramest schould unfix the variable and estimate the correct value.
```
blnicho, line 51: Typo: parmest, should

pyomo/contrib/parmest/tests/edgecase.py
```diff
@@ -0,0 +1,111 @@
...
+    # The third argument is a list of variable names that the user wants to estimate.
+    # The list contains strings with Var names from the Pyomo model.
+    theta_names = ['Boiler[1].eta']
+
+    # objective_function
+    def objective_function(model, data):
+        T = data['global']['T']
+        expr = 0
+        for t in T:
+            # (observed_value - value_calculated_form_model)^2*1/nr_time_steps
```
blnicho, line 66: Typo: from

pyomo/contrib/parmest/tests/test_scenariocreator.py
```diff
@@ -29,7 +29,8 @@
 
 testdir = os.path.dirname(os.path.abspath(__file__))
 
-
+print("parmest.parmest_available", parmest.parmest_available)
+print("ipopt_available",ipopt_available)
```
blnicho, line 33: These look like debugging print statements, can they be removed?

blnicho (dismissed):

doc/OnlineDocs/bibliography.rst
```diff
@@ -13,6 +13,10 @@ Bibliography
 
 .. [GAMS] http://www.gams.com
 
+.. [ParmestPaper] Katherine A. Klise, Bethany L. Nicholson, Andrea Staid, David L.Woodruff.
+	     Parmest: Parameter Estimation Via Pyomo.
+	     Computer Aided Chemical Engineering, 47: 41-46.
```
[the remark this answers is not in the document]
DLWoodruff: Thanks for all of the comments in this thorough review. I am way behind due to a family emergency, but I should be able to make all of these changes next week.

pyomo/contrib/parmest/parmest.py
```diff
@@ -585,19 +496,22 @@ def _Q_opt(self, ThetaVals=None, solver="ef_ipopt",
...
                     vals = {}
                     for var in return_values:
                         exp_i_var = exp_i.find_component(str(var))
-                        temp = [pyo.value(_) for _ in exp_i_var.itervalues()]
+                        if exp_i_var is None:  # we might have a block such as _mpisppy_data
+                            continue
+                        temp = [pyo.value(_) for _ in exp_i_var.values()]
                         if len(temp) == 1:
                             vals[var] = temp[0]
                         else:
```
kaklise, line 517: The current code adds additional rows of NaNs when exp_i_var is None. I suggest adding the following lines at 519.
```
if len(vals) > 0:
    var_values.append(vals) 
```

kaklise (changes requested):

pyomo/contrib/parmest/parmest.py
DLWoodruff: Thanks! This solves the mystery.  We could do the fix you suggest, or maybe do something to avoid looking at the
extra blocks; either way, this should be the last piece of the puzzle.

pyomo/contrib/parmest/parmest.py
```diff
@@ -449,65 +393,46 @@ def _Q_opt(self, ThetaVals=None, solver="ef_ipopt",
...
-        callback.  Side note (feb 2018, dlw): if you later decide to
-        construct the tree just once and reuse it, then remember to
-        remove thetavals from it when none is desired.
         """
-        
-        assert(solver != "k_aug" or ThetaVals == None)
-        # Create a tree with dummy scenarios (callback will supply when needed).
-        # Which names to use (i.e., numbers) depends on if it is for bootstrap.
+        if (solver == "k_aug"):
+            raise RuntimeError("k_aug no longer supported.")
```
blnicho, line 398: We might want to add an issue to support k_aug in the future now that we have easier access to it through `idaes-ext` in our testing environment.

pyomo/contrib/parmest/parmest.py
```diff
@@ -23,133 +42,46 @@
...
-    # pull off the index
-    l = vstr.find('[')
-    if l == -1:
-        indexstr = None
-        basestr = vstr
+def ef_nonants(ef):
+    # Wrapper to call someone's ef_nonants
+    # (the function being called is very short, but it might be changed)
+    if use_mpisppy:
+        return sputils.ef_nonants(ef)
```
blnicho, line 64: Since the `ef_nonants` function has the same name and arguments regardless of if we're using mpi-sppy or not can we just explicitly import it above and avoid this wrapper? Is the comment implying that this is for future-proofing in case the functions diverge?

pyomo/contrib/parmest/parmest.py
```diff
@@ -449,65 +393,46 @@ def _Q_opt(self, ThetaVals=None, solver="ef_ipopt",
...
-            scenariotree.tree_structure.CUID_repr_version = _cuidver
-                
+            outer_cb_data["BootList"] = bootlist
+        outer_cb_data["cb_data"] = self.callback_data  # None is OK
+        outer_cb_data["theta_names"] = self.theta_names
+
+        options = {"solver": "ipopt"}
+        scenario_creator_options = {"cb_data": outer_cb_data}
+        if use_mpisppy:
+            ef = sputils.create_EF(scen_names,
```
blnicho, line 420: Similar question here as above, can we just explicitly import `create_EF` and avoid this if-tree?

blnicho (commented):
I have a few minor questions

pyomo/contrib/parmest/parmest.py
```diff
@@ -204,12 +135,15 @@ def _pysp_instance_creation_callback(scenario_tree_model,
...
-    cb_data = scenario_tree_model.cb_data # cb_data might be None.
+    cb_data = outer_cb_data["cb_data"] # cb_data might be None.
 
     # at least three signatures are supported. The first is preferred
     try:
         instance = callback(experiment_number = exp_num, cb_data = cb_data)
     except TypeError:
+        raise RuntimeError("Only one callback signature is supported: "
+                           "callback(experiment_number, cb_data) ")
+        """
```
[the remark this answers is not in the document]
DLWoodruff: I deleted it and updated the comment above.

pyomo/contrib/parmest/parmest.py
DLWoodruff: I was not clever enough to think of using "future-proofing," but yes, I think we should keep it this way because I think it might
help future maintainers who might not be aware of the relationship and history between the stochastic programming terminology and least squares parameter estimation.

pyomo/contrib/parmest/parmest.py
DLWoodruff: I think we should leave it this way to help future maintainers.

blnicho (approved):

PR #1780, opened 2021-01-11
Fix subproblem initializations in gdp.cuttingplane

emma58:
## Summary/Motivation:

Especially in numerically less-ideal situations, ipopt was failing to find a feasible solution to the relaxed bigM problem in the second iteration. This is not shocking given that it was being initialized with the solution that the cut from the first iteration had just cut off. The primary change in this PR is to initialize with the projection of that solution on the the relaxed hull region, which is sure to be feasible for relaxed bigM with the cut.

- Don't load solutions into any subproblems until we know there is a solution (so that cut generation can stop gracefully)
- Initializing the relaxed bigM problem with the separation problem solution (feasible for relaxed hull, hence relaxed bigM)
- Fixing some typos, adding more descriptive logging for *why* cuttingplanes terminated
- Sneaking some more informative logging into `contrib.fourier_motzkin_elimination`
- (One test needed tolerance after initialization changes... It's within the solvers' constraint tolerance, so I think this is OK)

### Legal Acknowledgement


pyomo/gdp/plugins/cuttingplane.py
```diff
@@ -65,13 +65,17 @@ def _constraint_tight(model, constraint):
     """
     val = value(constraint.body)
     ans = [0, 0]
+    # TODO: Should we let the user set this?
+    TOL = 1e-6
     if constraint.lower is not None:
-        if value(constraint.lower) >= val:
+        if value(constraint.lower) >= val or \
+        val - value(constraint.lower) <= TOL:
```
jsiirola, line 72: The first clause is redundant (implied by the second).

pyomo/gdp/plugins/cuttingplane.py
```diff
@@ -65,13 +65,17 @@ def _constraint_tight(model, constraint):
...
             # tight or in violation of LB
             ans[0] -= 1
 
     if constraint.upper is not None:
-        if value(constraint.upper) <= val:
+        if value(constraint.upper) <= val or \
+           value(constraint.upper) - val <= TOL:
```
jsiirola, line 78: The first clause is redundant (implied by the second).

jsiirola (commented):
This looks good, although we should probably remove the redundant test clauses in `_constraint_tight` before merging.

jsiirola (approved):

PR #1797, opened 2021-01-26
Overhaul Pyomo logging configuration

### LINKED ISSUE #1760: "pyomo" Logger Sends Logs To Stdout Irrespective of User's Logging Config
Hello,

Thank you for all the hard work on Pyomo!

We are having some trouble with `pyomo` not respecting our custom logging configuration, and unconditionally emitting its logs to `stdout` in a hardcoded format.

Consider, for example, how `pyomo` ignores a Python logging configuration that sends all logs into a `StringIO` buffer with a custom format message:

```python
>>> import logging
>>> from io import StringIO
>>> import pyomo.environ
>>> 
>>> string_io = StringIO()
>>> handler = logging.StreamHandler(stream=string_io)
>>> logging.basicConfig(format='custom:%(message)s', handlers=[handler])
>>> 
>>> logging.getLogger("pyomo").warning("pyomo warning")
WARNING: pyomo warning
>>> logging.warning("example warning")
>>> string_io.getvalue()
'custom:pyomo warning\ncustom:example warning\n'
>>> 
```

Unfortunately, the `"pyomo"` logger emits its log to both `stdout` and the `string_io` buffer...Additionally, the log it emits to `stdout` also does not respect the custom format configured in the `logging.basicConfig` call---the message `pyomo` emits to stdout doesn't have `custom` as its prefix.

I think the root cause of this issue is that `pyomo` adds a custom handler to itself in `common/log.py`:

https://github.com/Pyomo/pyomo/blob/263d5b1a89dbd13eab1db01cdff687bbb1c4e795/pyomo/common/log.py#L165-L171

This custom handler, evidently, is responsible for the log messages that always wind up in `stdout` in the wrong format irrespective of the users's choices. The Python docs [advise against adding a handler in a library's logger because it means that developers are unable to configure logging to suit their needs](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library):

> Note It is strongly advised that you do not add any handlers other than NullHandler to your library’s loggers. This is because the configuration of handlers is the prerogative of the application developer who uses your library. The application developer knows their target audience and what handlers are most appropriate for their application: if you add handlers ‘under the hood’, you might well interfere with their ability to carry out unit tests and deliver logs which suit their requirements.

In our application, for example, we want all log messages to have a custom JSON format and be sent to `stderr`. But, we wind up with many logs from `pyomo` in `stdout` without our custom JSON formatting.

We can work around this problem by preemptively running `import pyomo.environ` and then clobbering all of the handlers on the `"pyomo"` logger, but---of course---that solution is pretty brittle.

Thanks for your help.

jsiirola:
## Summary/Motivation:
The original Pyomo logging configuration was heavily geared around the `pyomo` script and did not conform to the general behavior expected by users.  For example, calling `logging.basicConfig()` was effectively ignored by the Pyomo loggers.  This PR does an overhaul of the Pyomo logger configuration to more closely follow the accepted (and expected) Python logging behavior.  In particular, the bulk of the logic is now implemented through Formatter and Filter objects (allowing users to reuse the formatting in their own Handlers), and the Pyomo default logging will defers to a root logger if the user / calling environment created one.

This also creates a standardized `is_debug_set` function (in `pyomo.common.log`) to replace our use of `if __debug__ and logger.isEnabledFor(logging.DEBUG)`.  The old idiom would generate debugging log records when the root logger was left in "NOTSET" (which is the case not what the core Pyomo log configuration does not explicitly set a level).  The new function preserves the old behavior (of only emitting DEBUG log records when the user explicitly set a debugging level) by only returning `True` if the effective level is in `NOTSET < level <= DEBUG`.
- Note to reviewers: I am not completely thrilled with the name of this function and am soliciting alternative suggestions.

This should be fully-backwards compatible with the existing logging setup (in that existing scripts that log correctly should continue to log in the same manner).

- Overhaul the Pyomo logger to move the bulk of the formatting (and line wrapping) logic into a proper Formatter class.
- Rely on a Filter class to defer to a root level logger (if configured)
- Overhaul the logging configuration in the scripting interface to more robustly restore the logging state after the command exits
- define a standardized `is_debug_set` function (and propagate throughout code base)
- Deprecate the `LogHandler` class and `configure_loggers` function

### Legal Acknowledgement


pyomo/common/log.py
```diff
@@ -23,152 +23,249 @@
...
         # simple introductory line followed by an intented line (our
         # tests did this!), and cannot specify it without adding an
         # extra blank line to the output.  In contrast, it is possible
         # for the user to fix the scenario above that motivated this
         # code by just indenting their first line correctly.
+        msg = textwrap.dedent(msg).strip()
+
+        # Split the formatted log message (that currently has _flag in
+        # lieu of the actual message content) into lines, then
+        # recombine, subtitiuting and wrapping any lines that contain
```
blnicho, line 125: Typo: substituting

pyomo/common/log.py
```diff
@@ -23,152 +23,249 @@
...
+    base=PYOMO_ROOT_DIR,
+    verbosity=lambda: _pyomoLogger.isEnabledFor(logging.DEBUG),
+))
+_handler.addFilter(_GlobalLogFilter())
+_pyomoLogger.addHandler(_handler)
+
+
+class LogHandler(logging.StreamHandler):
+    @deprecated('The pyomo.common.log.LogHandler class has been deprecated '
+                'in favor of standard Handlers from the python logging module '
```
blnicho, line 256: Capitalize "Python"

blnicho (approved):
I found a couple typos but otherwise I think this looks fine

pyomo/common/log.py
jsiirola: Fixed.

pyomo/common/log.py
jsiirola: Fixed.

jsiirola (commented):
Resolved @blnicho comments.

pyomo/contrib/fbbt/tests/test_fbbt.py
```diff
@@ -727,33 +727,24 @@ def test_inf_bounds_on_expr(self):
         self.assertEqual(lb, None)
         self.assertEqual(ub, None)
 
-    @unittest.skip('This test passes locally, but not on travis or appveyor. I will add an issue.')
```
mrmundt, line 730: I'm guessing this is a sneaky bug fix?

mrmundt (approved):

pyomo/contrib/fbbt/tests/test_fbbt.py
jsiirola: It is actually a fix to the implementation of the test (leveraging the `LoggingIntercept` context manager resolves portability problems)

PR #1804, opened 2021-01-28
Add writer consistency checks between filenames and formats

jsiirola:
## Summary/Motivation:
There is an issue where we do not perform consistency checks between filenames specified by the user and the file format specified by the user.  The simplest example of this is:
```
pyomo solve --solver cbc model.py --save-results=results.json
```
In this case, Pyomo will create a `results.json` file and write data to it in YAML format.  This PR does a couple things: the default results format is pulled from the filename extension (if applicable), and for the cases where both a filename and a format is specified, if they don't match (in the Windows sense of checking extensions), then a warning is emitted to the user.

- infer the results format from the specified filename (if applicable)
- warn when a writer (model or results) is asked to write to a file in a different format than would normally be inferred by the file extension.

### Legal Acknowledgement

### DIFF
--- a/pyomo/opt/results/results_.py
+++ b/pyomo/opt/results/results_.py
@@ -14,8 +14,10 @@ import math
 import sys
 import copy
 import json
+import logging
+import os.path
 
-from pyomo.common.dependencies import yaml, yaml_load_args
+from pyomo.common.dependencies import yaml, yaml_load_args, yaml_available
 import pyomo.opt
 from pyomo.opt.results.container import (undefined,
                                          ignore,
@@ -29,6 +31,7 @@ import pyomo.opt.results.solver
 from six import iteritems, StringIO
 from six.moves import xrange
 
+logger = logging.getLogger(__name__)
 
 class SolverResults(MapContainer):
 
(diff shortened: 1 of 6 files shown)


pyomo/opt/results/results_.py
```diff
@@ -75,18 +78,38 @@ def _repn_(self, option):
...
+                    "writing results to file (%s) using what appears "
+                    "to be an incompatible format (%s)" % (fname, _fmt))
+            with open(fname, "w") as OUTPUT:
+                kwds['ostream'] = OUTPUT
+                kwds['format'] = _fmt
+                self.write(**kwds)
         else:
-            self.write_json(**kwds)
+            if not _fmt:
+                _fmt = 'yaml' if yaml_available else 'json'
```
blnicho, line 106: @jsiirola we can write a yaml file without yaml being available and some of our tests rely on this behavior. I'm conflicted over removing the `if yaml_available else 'json'` part of this line to match the previous behavior or skipping 3 tests when yaml is not available.

Any preference?

blnicho (approved):
@jsiirola I'm pretty sure I fixed all the test failures. If you agree with the minor changes I made then this should be ready to merge once all the tests finish.

jsiirola (commented):
Agreed.  This looks good.

PR #1812, opened 2021-02-02
Replace pyutilib.subprocess with Python subprocess

mrmundt:
## Summary/Motivation:
A Pyomo 6.0.0 initiative is to remove the PyUtilib package. This replaces pyutilib.subprocess with python's base `subprocess` package.

- Remove `pyutilib.subprocess`
- Add Python `subprocess` library

### Legal Acknowledgement


pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -194,6 +193,8 @@ def _run_simple_model(self, n):
             from gams import GamsWorkspace, DebugLevel
             GamsWorkspace(debug=DebugLevel.Off,
                           working_directory=tmpdir)
+            ## !!THIS SEEMS LIKE A BUG!! - mrmundt ##
+            ## ws is undefined
```
blnicho, line 197: I agree. I think line 194 is missing `ws = ` before `GamsWorkspace(...`

pyomo/solvers/plugins/solvers/ASL.py
```diff
@@ -93,8 +93,12 @@ def _get_version(self):
...
         if solver_exec is None:
             return _extract_version('')
-        results = run( [solver_exec,"-v"], timelimit=1 )
-        return _extract_version(results[1])
+        try:
+            results = subprocess.run( [solver_exec,"-v"], timeout=1,
+                                 stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+            return _extract_version(results.stdout.decode("utf-8"))
+        except OSError:
+            pass
```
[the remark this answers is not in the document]
mrmundt: Yup! There is some logic in pyutilib.subprocess that says, "If you get an OSError, just pass and keep going." This was the only place that error was ever triggered (I believe for Windows).

pyomo/scripting/driver_help.py
```diff
@@ -16,8 +16,9 @@
 import textwrap
 import logging
 import socket
+import subprocess
 
-import pyutilib.subprocess
+import pyutilib.misc
```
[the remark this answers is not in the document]
mrmundt: The `import pyutilib.subprocess` was sufficient to allow access to `pyutilib.misc.sort_index` later in the file. Since I removed the `pyutilib.subprocess` import, I needed to add back the `pyutilib.misc` (at least until I rip those out later).

pyomo/repn/tests/ampl/test_ampl_comparison.py
```diff
@@ -98,16 +98,18 @@ def nlwriter_asl_test(self, name):
...
 
     # obtain the nl file summary information for comparison with ampl
-    p = pyutilib.subprocess.run(
+    p = subprocess.run(
         'gjh_asl_json '+currdir+name+'.test.nl rows='
-        +currdir+name+'.test.row cols='+currdir+name+'.test.col')
-    self.assertTrue(p[0] == 0, msg=p[1])
+        +currdir+name+'.test.row cols='+currdir+name+'.test.col',
+        stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
+    self.assertTrue(p.returncode == 0, msg=p.stdout.decode("utf-8"))
```
jsiirola, line 105: I think this can be converted to a list with:
```python
p = subprocess.run(['gjh_asl_json',
    currdir+name+'.test.nl',
    'rows='+currdir+name+'.test.row',
    'cols='+currdir+name+'.test.col'],
    stdout=subprocess.PIPE, stderr=subprocess.PIPE)

File: pyomo/opt/testing/pyunit.py [hunk 5]
```diff
@@ -32,13 +32,16 @@ def _failIfPyomoResultsDiffer(self, cmd=None, baseline=None, cwd=None):
...
     finally:
         os.chdir(oldpwd)
     
-    if output[0] != 0:
+    if output.returncode != 0:
         self.fail("Command terminated with nonzero status: '%s'" % cmd)
-    results = extract_results(re.split('\n',output[1]))
+    # !!THIS SEEMS LIKE A BUG!! - mrmundt #
+    # There is nothing imported called "extract_results" nor "compare_results"
+    results = extract_results(re.split('\n', output.stdout.decode("utf-8")))
```
jsiirola, line 44: I agree.  I not sure this code is / can be run?  I didn't find an extract_results anywhere in either Pyomo or PyUtilib.  I wonder if this entire directory (`pyomo/opt/testing`) is actually abandoned code?

pyomo/pysp/tests/convert/test_ddsip.py
```diff
@@ -85,15 +74,14 @@ def _get_cmd(self,
...
-        self.assertNotEqual(rc[0], 0)
-        self._assert_contains(
-            outfile,
-            b"ValueError: One or more deterministic parts of the problem found in file")
+        result = subprocess.run(cmd, stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE)
+        self.assertNotEqual(result.returncode, 0)
+        self.assertIn(
+           b"ValueError: One or more deterministic parts of the problem found in file",
+           result.stderr)
```
jsiirola, line 82: Out of curiosity, is there a reason to look for a binary string instead of decoding stderr?

pyomo/solvers/plugins/converter/ampl.py
```diff
@@ -78,7 +78,8 @@ def apply(self, *args, **kwargs):
         #
         # Execute command and cleanup
         #
-        output = pyutilib.subprocess.run(cmd)
+        output = subprocess.run(cmd, stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE)
         if not os.path.exists(output_filename):       #pragma:nocover
-            raise ApplicationError("Problem launching 'ampl' to create '%s': %s" % (output_filename, output))
+            raise ApplicationError("Problem launching 'ampl' to create '%s': %s" % (output_filename, output.stdout.decode("utf-8")))
```
jsiirola, line 84: This only outputs stdout.  It might be better to make the run call use `stderr=subprocess.STDOUT` so that we get both stdout and stderr.

pyomo/core/tests/examples/test_kernel_examples.py
```diff
@@ -64,8 +64,9 @@ def testmethod(self):
             if (not testing_solvers['ipopt','nl']) or \
                (not testing_solvers['mosek_direct','python']):
                 self.skipTest("Ipopt or Mosek is not available")
-        rc, log = pyutilib.subprocess.run([sys.executable,example])
-        self.assertEqual(rc, 0, msg=log)
+        result = subprocess.run([sys.executable,example],
+                                stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        self.assertEqual(result.returncode, 0, msg=result.stdout.decode("utf-8"))
```
jsiirola, line 69: Is it sufficient to only output the stdout is the test fails?  should we consider `stderr=subprocess.STDOUT`?

pyomo/repn/tests/ampl/test_ampl_comparison.py
```diff
@@ -98,16 +98,20 @@ def nlwriter_asl_test(self, name):
...
     # obtain the nl file summary information for comparison with ampl
-    p = pyutilib.subprocess.run(
-        'gjh_asl_json '+currdir+name+'.test.nl rows='
-        +currdir+name+'.test.row cols='+currdir+name+'.test.col')
-    self.assertTrue(p[0] == 0, msg=p[1])
+    p = subprocess.run(['gjh_asl_json',
+                        currdir+name+'.test.nl',
+                        'rows='+currdir+name+'.test.row',
+                        'cols='+currdir+name+'.test.col'],
+                       stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
jsiirola, line 105: Should we use `stderr=subprocess.STDOUT` so that the test failure message will contain the error?

pyomo/repn/tests/ampl/test_ampl_comparison.py
```diff
@@ -98,16 +98,20 @@ def nlwriter_asl_test(self, name):
...
     # obtain the nl file summary information for comparison with pyomo
-    p = pyutilib.subprocess.run(
-        'gjh_asl_json '+currdir+name+'.ampl.nl rows='
-        +currdir+name+'.ampl.row cols='+currdir+name+'.ampl.col')
-    self.assertTrue(p[0] == 0, msg=p[1])
+    p = subprocess.run(['gjh_asl_json',
+                        currdir+name+'.ampl.nl',
+                        'rows='+currdir+name+'.ampl.row',
+                        'cols='+currdir+name+'.ampl.col'],
+                       stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
jsiirola, line 113: Should we use `stderr=subprocess.STDOUT` so that the test failure message will contain the error?

pyomo/solvers/plugins/solvers/ASL.py
```diff
@@ -93,8 +93,12 @@ def _get_version(self):
         solver_exec = self.executable()
         if solver_exec is None:
             return _extract_version('')
-        results = run( [solver_exec,"-v"], timelimit=1 )
-        return _extract_version(results[1])
+        try:
+            results = subprocess.run( [solver_exec,"-v"], timeout=1,
```
jsiirola, line 97: PEP8: whitespace (x2): `run([solver_exec, "-v"],`

pyomo/solvers/plugins/solvers/ASL.py
```diff
@@ -93,8 +93,12 @@ def _get_version(self):
...
+                                 stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
jsiirola, line 98: Is capturing and ignoring stderr a problem here?

pyomo/solvers/plugins/solvers/BARON.py
```diff
@@ -177,8 +176,10 @@ def _get_version(self):
         else:
             fnames = self._get_dummy_input_files(check_license=False)
             try:
-                results = run([solver_exec, fnames[0]])
-                ver = _extract_version(results[1])
+                results = subprocess.run([solver_exec, fnames[0]],
+                                         stdout=subprocess.PIPE,
+                                         stderr=subprocess.PIPE)
```
jsiirola, line 181: Is capturing and ignoring stderr a problem here?

pyomo/solvers/plugins/solvers/CBCplugin.py
```diff
@@ -53,11 +52,13 @@ def configure_cbc():
     if not executable:
         return
     cbc_exec = executable.path()
-    results = run( [cbc_exec,"-stop"], timelimit=1 )
-    _cbc_version = _extract_version(results[1])
-    results = run(
-        [cbc_exec,"dummy","-AMPL","-stop"], timelimit=1 )
-    _cbc_compiled_with_asl = not ('No match for AMPL' in results[1])
+    results = subprocess.run( [cbc_exec,"-stop"], timeout=1,
```
jsiirola, line 55: PEP8: whitespace (x2): `run([cbc_exec, "-stop"]`

pyomo/solvers/plugins/solvers/CBCplugin.py
```diff
@@ -53,11 +52,13 @@ def configure_cbc():
...
+                             stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
jsiirola, line 56: Is capturing and ignoring stderr a problem here?

pyomo/solvers/plugins/solvers/CBCplugin.py
```diff
@@ -53,11 +52,13 @@ def configure_cbc():
...
+    _cbc_version = _extract_version(results.stdout.decode("utf-8"))
+    results = subprocess.run(
+        [cbc_exec,"dummy","-AMPL","-stop"], timeout=1, stdout=subprocess.PIPE,
```
jsiirola, line 59: PEP8: whitespace: `[cbc_exec, "dummy", "-AMPL", "-stop"]`

pyomo/solvers/plugins/solvers/CBCplugin.py
```diff
@@ -53,11 +52,13 @@ def configure_cbc():
...
+        stderr=subprocess.PIPE)
```
jsiirola, line 60: Is capturing and ignoring stderr a problem here?  Specifically, does '`No match for AMPL'` come out stdout or stderr?

pyomo/solvers/plugins/solvers/CONOPT.py
```diff
@@ -79,8 +79,9 @@ def _get_version(self):
         solver_exec = self.executable()
         if solver_exec is None:
             return _extract_version('')
-        results = run( [solver_exec], timelimit=1 )
-        return _extract_version(results[1])
+        results = subprocess.run( [solver_exec], timeout=1,
+                                 stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
jsiirola, line 83: Is capturing and ignoring stderr a problem here?

pyomo/solvers/plugins/solvers/CPLEX.py
```diff
@@ -397,8 +397,9 @@ def _get_version(self):
         solver_exec = self.executable()
         if solver_exec is None:
             return _extract_version('')
-        results = run( [solver_exec,'-c','quit'], timelimit=1 )
-        return _extract_version(results[1])
+        results = subprocess.run( [solver_exec,'-c','quit'], timeout=1,
+                                 stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
jsiirola, line 401: Is capturing and ignoring stderr a problem here?

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -659,8 +660,9 @@ def _get_version(self):
         else:
             # specify logging to stdout for windows compatibility
             cmd = [solver_exec, "audit", "lo=3"]
-            _, txt = pyutilib.subprocess.run(cmd, tee=False)
-            return _extract_version(txt)
+            results = subprocess.run(cmd, stdout=subprocess.PIPE,
+                                     stderr=subprocess.PIPE)
```
jsiirola, line 664: Is capturing and ignoring stderr a problem here?

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -811,7 +813,18 @@ def solve(self, *args, **kwds):
...
         try:
-            rc, txt = pyutilib.subprocess.run(command, tee=tee)
+            with subprocess.Popen(command, stdout=subprocess.PIPE,
+                                  stderr=subprocess.STDOUT, bufsize=1,
+                                  universal_newlines=True) as p, \
+                open('logfile', 'w') as f:
+                for line in p.stdout:
+                    if tee:
+                        sys.stdout.write(line)
+                    f.write(line)
```
jsiirola, line 823: This is a problem, as the stdout does not appear until after the subprocess terminates (and is not the intent of `tee=True`).

pyomo/solvers/plugins/solvers/GLPK.py
```diff
@@ -34,8 +34,11 @@ def configure_glpk():
     _glpk_version = _extract_version("")
     if not Executable("glpsol"):
         return
-    errcode, results = pyutilib.subprocess.run(
-        [Executable('glpsol').path(), "--version"], timelimit=2)
+    result = subprocess.run([Executable('glpsol').path(), "--version"],
+                            stdout=subprocess.PIPE, stderr=subprocess.PIPE,
```
jsiirola, line 38: Is capturing and ignoring stderr a problem here?

pyomo/solvers/plugins/solvers/GLPK_old.py
```diff
@@ -455,8 +455,10 @@ def _get_version(self):
         solver_exec = self.executable()
         if solver_exec is None:
             return _extract_version('')
-        errcode, results = pyutilib.subprocess.run(
-        [solver_exec, "--version"], timelimit=1)
+        result = subprocess.run([solver_exec, "--version"], timeout=1,
+                                stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
jsiirola, line 459: Is capturing and ignoring stderr a problem here?

pyomo/solvers/plugins/solvers/GUROBI.py
```diff
@@ -278,14 +277,13 @@ def _get_version(self):
...
         else:
-            f = StringIO()
-            results = run([solver_exec],
-                          stdin=('from gurobipy import *; '
-                                 'print(gurobi.version()); exit()'),
-                          ostream=f)
+            results = subprocess.run([solver_exec],
+                                 input=('from gurobipy import *; print(gurobi.version()); exit()'.encode()),
+                                 stdout=subprocess.PIPE,
+                                 stderr=subprocess.PIPE)
```
jsiirola, line 283: Is capturing and ignoring stderr a problem here?

pyomo/solvers/plugins/solvers/IPOPT.py
```diff
@@ -78,8 +78,9 @@ def _get_version(self):
         solver_exec = self.executable()
         if solver_exec is None:
             return _extract_version('')
-        results = run( [solver_exec,"-v"], timelimit=1 )
-        return _extract_version(results[1])
+        results = subprocess.run( [solver_exec,"-v"], timeout=1,
+                                 stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
jsiirola, line 82: Is capturing and ignoring stderr a problem here?

pyomo/solvers/plugins/solvers/SCIPAMPL.py
```diff
@@ -77,8 +77,9 @@ def _get_version(self):
         solver_exec = self.executable()
         if solver_exec is None:
             return _extract_version('')
-        results = run( [solver_exec], timelimit=1 )
-        return _extract_version(results[1])
+        results = subprocess.run( [solver_exec], timeout=1,
+                                 stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
jsiirola, line 81: Is capturing and ignoring stderr a problem here?

jsiirola (changes requested):
Several small comments and one really big one:

- I am pretty sure we need a different implementation of `_execute_command()` in `pyomo/opt/solver/shellcmd.py`.
- There are a lot of places where you are suppressing (capturing then ignoring) both stdout and stderr.  Should we consider only capturing stdout so that the stderr propagates out (for information / debugging)?

pyomo/common/download.py
```diff
@@ -98,9 +97,11 @@ def _get_distver_from_redhat_release(cls):
 
     @classmethod
     def _get_distver_from_lsb_release(cls):
-        rc, dist = run(['lsb_release', '-si'])
-        rc, ver = run(['lsb_release', '-sr'])
-        return cls._map_dist(dist.lower().strip()), ver.strip()
+        dist = subprocess.run(['lsb_release -si'], stdout=subprocess.PIPE,
+                              stderr=subprocess.PIPE)
+        ver = subprocess.run(['lsb_release -sr'], stdout=subprocess.PIPE,
```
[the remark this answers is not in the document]
mrmundt: Already did - it was an error on my side.

examples/doc/samples/scripts/test_scripts.py
```diff
@@ -34,7 +35,8 @@ def tearDown(self):
     def run_script(self, test, filter, yaml=False):
         cwd = self.cwd+os.sep+test+os.sep
         os.chdir(cwd)
-        pyutilib.subprocess.run(sys.executable+' script.py', outfile=cwd+os.sep+'script.log', cwd=cwd)
+        with open(cwd+os.sep+'script.log', 'w') as f:
+            subprocess.run([sys.executable, 'script.py'], stdout=f, stderr=f)
```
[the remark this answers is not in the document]
mrmundt: Error - added back.

pyomo/core/tests/examples/test_convert.py
```diff
@@ -62,10 +62,10 @@ def convert(self, cmd, type, **kwds):
 
     # in the convert tests, make sure everything is generating symbolic solver labels - aids debugging.
     def run_convert2nl(self, cmd, file=None):
-        return pyutilib.subprocess.run('pyomo convert --format=nl --symbolic-solver-labels '+cmd, outfile=file)
+        return subprocess.run('pyomo convert --format=nl --symbolic-solver-labels '+cmd)
```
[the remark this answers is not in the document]
mrmundt: Question... Does this even run? When I trigger it locally, it returns:
```
$ nosetests -v pyomo/core/tests/examples/test_convert.py 

Ran 0 tests in 0.000s

OK
```
The class is defined (line 42-43):
```
class xTest(object):
#class Test(unittest.TestCase):
```

pyomo/opt/solver/shellcmd.py
```diff
@@ -299,14 +299,20 @@ def _execute_command(self,command):
...
                 env   = command.env,
-                tee   = self._tee,
-                define_signal_handlers = self._define_signal_handlers
+                stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT,
+                universal_newlines=True
              )
+            if self._tee:
+                for line in result.stdout:
+                    sys.stdout.write(line)
```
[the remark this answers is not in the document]
mrmundt: Can we have a bigger conversation on this? It looks like the [section that does this](https://github.com/PyUtilib/pyutilib/blob/master/pyutilib/subprocess/processmngr.py#L587) in pyutilib is based on threading, and it's quite long.

pyomo/pysp/tests/convert/test_ddsip.py
mrmundt: No particular reason (besides that it was less typing). I'm not married to this choice if you want me to decode stderr instead.

pyomo/core/tests/examples/test_kernel_examples.py
mrmundt: Good question. I was recreating the capability of the call previously - which I assume was previously considered to be sufficient. I'm happy to change it if you disagree with previous design choices.

pyomo/core/tests/examples/test_convert.py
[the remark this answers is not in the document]
mrmundt: Answer: This does NOT currently run.
Next steps: We are going to leave it alone for now until a time when we do a deep-dive into coverage.

pyomo/opt/solver/shellcmd.py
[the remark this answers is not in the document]
mrmundt: Reading more on the vast interwebs, it looks like threading is the only way to do this to produce the output as desired. How would you like to proceed?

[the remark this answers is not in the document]
michaelbynum: I'm think I'm still not convinced that the solution I developed here (https://github.com/michaelbynum/pyomo/blob/appsi/pyomo/contrib/appsi/utils/tee_thread.py) and here (https://github.com/michaelbynum/pyomo/blob/47beab745db7f4601aaac3e247b15f7ef434a82b/pyomo/contrib/appsi/solvers/ipopt.py#L277) is not sufficient. I think we discussed a problem where the last few lines of the log file might not be printed to stdout, but I think that can be easily fixed by giving `TeeThread` a file object instead of a filename. Then put subprocess in a try-finally block where the finally clause makes sure everything was written. Maybe there are other issues that I am forgetting now.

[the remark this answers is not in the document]
jsiirola: There are a number of issues here.  I will try to (somewhat) succinctly summarize what I have been thinking.

First, the bulk of this PR is good (we want to use Python's `subprocess` directly whenever possible).  However, there are a couple cases (notably the solver subprocesses) where we want to be able to send stdout/stderr to more than one location.  This includes 1) to the screen (in real time) for the user's benefit, 2) to a log file for archival purposes, and 3) [for some solvers] to a string buffer for subsequent parsing to get relevant results information back (like iteration count from Ipopt).  After some Googling, it appears that there is no way to do this without using either `threading` or `asyncio`.  For these situations I feel that we should have a separate utility (*not* called `subprocess`) that wraps `subprocess` handles the "tee" activity.

@michaelbynum has suggested a (beautifully simple) approach whereby we don't actually "tee", but instead send the stream to a file and have a separate thread that basically does `tail -f`.  This has the benefit of being the simplest implementation that I have run across.  However, it has some limitations.  The biggest is that it is tied to the filesystem.  In other projects (notably Dakota where inter-process communication can occur through the filesystem) we have seen numerous challenges when writing and then immediately reading from files.  This is particularly true on HPC systems where the filesystem is NFS or Lustre.  I forget the exact root cause, but I recall it being something like all the file data was not available, even after the process terminates.  It had to do with the (significant) buffering that occurs (both in the file descriptors and the OS).  It is aggravated by filesystems that are optimized for writing large amounts of data, but not for write/read operations (like Lustre).  There is also the principle that we really shouldn't be creating files if we don't actually need a file.

The solutions that are more akin to `tee` have the benefit of being entirely in-memory: so if you are in a situation where the user did not ask for the log file, but you need to parse the solver log, then you only create the string buffer and dump things to the console: no files are created, so there is nothing to clean up.

A separate (but related) question is how we should handle merging stdout/stderr.  If we let the underlying OS handle the merge (i.e., by using `stderr=subprocess.STDOUT`), then you get the (unfortunately common) situation where a stderr message gets stuck in the middle of a line of text sent to stdout.  This makes parsing the resulting combined stream very challenging (both for automated parsers trying to extract information, and for users).  A lot of the work in the PyUtilib implementation was geared around making the merge process more reliable and easier to parse (in particular, it tried to ensure that entire lines were kept together).

Finally, there are libraries out there other than PyUtilib that (mostly?) solve this problem.  As an alternative to rolling our own, we could add another dependency to Pyomo.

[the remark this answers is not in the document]
mrmundt: The idea of adding another dependency that we do not maintain is appealing to me. We are ripping out PyUtilib and Enum in 6.0.0, which means Pyomo still only requires:
```
      install_requires=[
          'ply',
          'six>=1.4',
      ],
```
Plus whatever library you think might be a good alternative, @jsiirola .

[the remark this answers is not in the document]
michaelbynum: I think the challenge of not all of the data being in the file can be overcome pretty easily, but I'm not entirely sure.

Are we sure that subprocess does not directly handle the correct merging of stdout/stderr?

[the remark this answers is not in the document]
michaelbynum: Forget my last comment about merging stdout/stderr. I understand.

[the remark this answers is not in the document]
michaelbynum: The correct merging of stdout/stderr is certainly more difficult, but I'm not convinced that it is terribly important (yet).

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -630,9 +630,9 @@ def _run_simple_model(self, n):
             test = os.path.join(tmpdir, 'test.gms')
             with open(test, 'w') as FILE:
                 FILE.write(self._simple_model(n))
-            rc, txt = pyutilib.subprocess.run(
+            result = subprocess.run(
                 [self.executable(), test, "curdir=" + tmpdir, 'lo=0'])
```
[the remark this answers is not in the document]
mrmundt: We didn't use any suppression here at all actually - which implies that it needs it!

examples/doc/samples/scripts/test_scripts.py
```diff
@@ -34,7 +45,9 @@ def tearDown(self):
     def run_script(self, test, filter, yaml=False):
         cwd = self.cwd+os.sep+test+os.sep
         os.chdir(cwd)
-        pyutilib.subprocess.run(sys.executable+' script.py', outfile=cwd+os.sep+'script.log', cwd=cwd)
+        with open(cwd+os.sep+'script.log', 'w') as f:
+            subprocess.run([sys.executable, 'script.py'], 
+                           stdout=f, stderr=f, cwd=cwd)
```
jsiirola, line 50: `f` is being opened in text mode; do we need `universal_newlines=True` here?

pyomo/core/tests/examples/test_convert.py
```diff
@@ -62,10 +62,16 @@ def convert(self, cmd, type, **kwds):
 
     # in the convert tests, make sure everything is generating symbolic solver labels - aids debugging.
     def run_convert2nl(self, cmd, file=None):
-        return pyutilib.subprocess.run('pyomo convert --format=nl --symbolic-solver-labels '+cmd, outfile=file)
+        cmd = ('pyomo convert --format=nl --symbolic-solver-labels '+cmd).split(' ')
+        with open(file, 'w') as f:
+            result = subprocess.run(cmd, stdout=f, stderr=f)
```
jsiirola, line 67: `f` is being opened in text mode; do we need `universal_newlines=True` here?

pyomo/core/tests/examples/test_convert.py
```diff
@@ -62,10 +62,16 @@ def convert(self, cmd, type, **kwds):
...
+        return result
 
     def run_convert2lp(self, cmd, file=None):
-        return pyutilib.subprocess.run('pyomo convert --format=lp --symbolic-solver-labels '+cmd, outfile=file)
+        cmd = ('pyomo convert --format=lp --symbolic-solver-labels '+cmd).split(' ')
+        with open(file, 'w') as f:
+            result = subprocess.run(cmd, stdout=f, stderr=f)
```
jsiirola, line 73: `f` is being opened in text mode; do we need `universal_newlines=True` here?

pyomo/core/tests/examples/test_pyomo.py
```diff
@@ -86,9 +86,11 @@ def tearDown(self):
             os.remove(currdir+'results.jsn')
 
     def run_pyomo(self, cmd, root=None):
-        cmd = 'pyomo solve --solver=glpk --results-format=json ' \
-              '--save-results=%s.jsn %s' % (root, cmd)
-        return pyutilib.subprocess.run(cmd, outfile=root+'.out')
+        cmd = ('pyomo solve --solver=glpk --results-format=json ' \
+              '--save-results=%s.jsn %s' % (root, cmd)).split(' ')
+        with open(root+'.out', 'w') as f:
+            result = subprocess.run(cmd, stdout=f, stderr=f)
```
jsiirola, line 92: `f` is being opened in text mode; do we need `universal_newlines=True` here?

pyomo/opt/testing/pyunit.py
```diff
@@ -32,13 +32,16 @@ def _failIfPyomoResultsDiffer(self, cmd=None, baseline=None, cwd=None):
             baseline = "\n".join(INPUT.readlines())
             INPUT.close()
     
-        output = pyutilib.subprocess.run(cmd)
+        output = subprocess.run(cmd, stdout=subprocess.PIPE,
+                                stderr=subprocess.STDOUT)
```
jsiirola, line 36: Should we switch this over to using `universal_newlines=True` here and avoid the call to `decode()` on line 44?

pyomo/pysp/tests/convert/test_ddsip.py
```diff
@@ -118,52 +106,43 @@ def test_bad_constraint_rhs(self):
...
-            b"StochasticConstraintBodyAnnotation. To correct this "
-            b"issue, ensure that multiple container components under "
-            b"which the component might be stored \(such as a Block "
-            b"and an indexed Constraint\) are not simultaneously set in "
-            b"this annotation.")
+        result = subprocess.run(cmd, stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE)
+        self.assertNotEqual(result.returncode, 0)
+        self.assertIn(
+            """RuntimeError: Component b.c was assigned multiple declarations in annotation type StochasticConstraintBodyAnnotation. To correct this issue, ensure that multiple container components under which the component might be stored (such as a Block and an indexed Constraint) are not simultaneously set in this annotation.""".encode('utf-8'),
```
jsiirola, line 113: Three questions: 
- why did you not make this a `bytes` string like you did on line 81 (instead of unicode / `decode()`), 
- can you restore the previous version where the long line was broken across multiple lines?
- As the new PySP repo has already been created/populated, should we just revert all changes to this file (because they are not likely to be propagated to PySP)?

pyomo/pysp/tests/convert/test_smps.py
```diff
@@ -150,52 +138,43 @@ def test_bad_constraint_rhs_LP(self):
...
-            b"StochasticConstraintBodyAnnotation. To correct this "
-            b"issue, ensure that multiple container components under "
-            b"which the component might be stored \(such as a Block "
-            b"and an indexed Constraint\) are not simultaneously set in "
-            b"this annotation.")
+        result = subprocess.run(cmd, stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE)
+        self.assertNotEqual(result.returncode, 0)
+        self.assertIn(
+            """RuntimeError: Component b.c was assigned multiple declarations in annotation type StochasticConstraintBodyAnnotation. To correct this issue, ensure that multiple container components under which the component might be stored (such as a Block and an indexed Constraint) are not simultaneously set in this annotation.""".encode('utf-8'),
```
jsiirola, line 145: Three questions: 
- why did you not make this a `bytes` string like you did on line 81 (instead of unicode / `decode()`), 
- can you restore the previous version where the long line was broken across multiple lines?
- As the new PySP repo has already been created/populated, should we just revert all changes to this file (because they are not likely to be propagated to PySP)?

pyomo/scripting/commands.py
```diff
@@ -123,7 +122,7 @@ def pyomo_python(args=None):
         console.interact('Pyomo Python Console\n'+sys.version)
     else:
         cmd = sys.executable+' '+ ' '.join(args)
-        pyutilib.subprocess.run(cmd, stdout=sys.stdout, stderr=sys.stderr)
+        subprocess.run(cmd, stdout=sys.stdout, stderr=sys.stderr)
```
jsiirola, line 125: Why are we specifying `stdout=sys.stdout, stderr=sys.stderr`? Shouldn't we just omit those arguments?

pyomo/solvers/plugins/converter/glpsol.py
```diff
@@ -104,7 +104,7 @@ def apply(self, *args, **kwargs):
                 OUTPUT.write("end;\n")
             OUTPUT.close()
             cmd.append(modfile)
-        pyutilib.subprocess.run(cmd)
+        subprocess.run(cmd)
```
jsiirola, line 107: Should this have `stdout=subprocess.DEVNULL, stderr=DEVNULL`, because the default behavior of the `pyutilib.subprocess.run` was to suppress the command output?

pyomo/solvers/plugins/converter/pico.py
```diff
@@ -72,7 +71,7 @@ def apply(self, *args, **kwargs):
                     raise ConverterError("File "+item+" does not exist!")
                 cmd = cmd + " "+item
         print("Running command: "+cmd)
-        pyutilib.subprocess.run(cmd)
+        subprocess.run(cmd)
```
jsiirola, line 74: Should this have `stdout=subprocess.DEVNULL, stderr=DEVNULL`, because the default behavior of the `pyutilib.subprocess.run` was to suppress the command output?

jsiirola (commented):
I think we are really close to being ready.  Just a few final questions.

examples/doc/samples/scripts/test_scripts.py
mrmundt: I tried running a `subprocess` command locally using this same format, both with and without `universal_newlines`, and it didn't seem to make any difference. So... I don't think so.

pyomo/scripting/commands.py
mrmundt: Good point - I forgot that pyutilib's default was to suppress and Python's default is to print out.

jsiirola (approved):

pyomo/common/tests/test_tee.py
```diff
@@ -0,0 +1,113 @@
...
+        with tee.TeeStream() as t:
+            out = t.STDOUT
+            self.assertIs(out, t.STDOUT)
+            err = t.STDERR
+            self.assertIs(err, t.STDERR)
+            self.assertIsNot(out, err)
+
+    @unittest.skipIf(not tee._peek_available,
+                     "This test is not deterministic if select/peek "
+                     "are available")
```
blnicho, line 41: Should this be "if select/peek are **not** available"?

pyomo/common/tests/test_tee.py
```diff
@@ -0,0 +1,113 @@
...
+        a = StringIO()
+        b = StringIO()
+        try:
+            _tmp, tee._peek_available = tee._peek_available, False
+            with tee.TeeStream(a,b) as t:
+                # Ensure both threads are running
+                t.STDOUT
+                t.STDERR
+                # ERR should come out before OUT, but this is slightly
+                # nondeterministic, so a short pause should help
```
blnicho, line 67: Given this comment should this test have the same skipif condition as the previous test?

blnicho (approved):
Two minor comments but otherwise this looks fine

pyomo/common/tests/test_tee.py
jsiirola: Yes.  Fixed.

pyomo/common/tests/test_tee.py
jsiirola: No.  The previous comment was unclear (the determinism wasn't really the issue: the previous test tested the expected behavior of the `_mergedReader` (which requires `_peek_available`),  This test should always work.  Both are slightly nondeterministic, but the delays are hopefully long enough to prevent false failures.

Issue #1040, opened 2019-05-28
[PEP] Clean up Pyomo / Pyutilib console scripts


ghackebeil:
We've had discussions about this in the past, but its time to make it official. I think we need to cut back on the number of console scripts we install, biggly.

Here is the list of scripts added to my bin directory after a fresh Pyomo install.

By Pyutilib:
 - checkCopyright
 - dispatch_srvr
 - lbin
 - lpython
 - pypi_downloads
 - pyutilib_test_driver
 - replaceCopyright
 - test.pyutilib

By Pyomo:
 - computeconf
 - evaluate_xhat
 - kill_pyro_mip_servers
 - launch_pyro_mip_servers
 - OSSolverService
 - phsolverserver
 - pyomo
 - pyomo_ns
 - pyomo_nsc
 - pyomo_old
 - pyomo_python
 - pyro_mip_server
 - readsol
 - results_schema
 - runbenders
 - runef
 - runph
 - scenariotreeserver
 - test.pyomo

This exceeds the number of console scripts in my system bin directory (associated with all python packages I've ever installed, over multiple Python versions) by almost a factor of two.

Let's discuss how to clean this up.

michaelbynum:
I agree.

jsiirola:
I agree.  As an opening proposal:

Pyutilib:
   - Keep
   - Deprecate immediately, replace with "executable modules" (e.g., `python -m pyutilib.foo`), remove the console script in 5.8
      - checkCopyright
      - lbin
      - lpython
      - pypi_downloads
      - replaceCopyright
      - test.pyutilib (assuming that we can make coverage work correctly)
   - Deprecate immediately, do not replace, remove the console script in 5.8
      - dispatch_srvr
      - pyutilib_test_driver

By Pyomo:
   - Keep
      - pyomo
   - Deprecate immediately, replace with pyomo subcommands, remove the console script in 5.7
      - computeconf
      - evaluate_xhat
      - phsolverserver
      - pyomo_python
      - runbenders
      - runef
      - runph
      - test.pyomo (assuming that we can make coverage work correctly)
   - Deprecate immediately, remove the console script in 5.7
      - kill_pyro_mip_servers
      - launch_pyro_mip_servers
      - OSSolverService
      - pyomo_ns
      - pyomo_nsc
      - pyro_mip_server
      - readsol
      - results_schema
      - scenariotreeserver
   - Remove immediately
      - pyomo_old

[...]

PR #1830, opened 2021-02-19
Cleanup console scripts

ghackebeil:
Working on https://github.com/Pyomo/pyomo/issues/1040

Removed:
 - `readsol`: broken since `pyomo.environ` import was required to register plugins (so a long time ago)
 - `OSSolverService`: broken since `pyomo.environ` import was required to register plugins (so a long time ago)
 - `results_schema`: I don't think anyone knows what this is
 - `pyomo_old`: broken, points to a function that does not exist
 - everything under the `[pyomo.commands]` section (I don't think these lines did anything)

Deprecated:
 - `pyomo_python`: it runs a subprocess and hides the return code, nobody should be using it
 -  ~~`SolverManager_Pyro` along with `pyro_mip_server`, `pyomo_ns`, `pyomo_nsc`, `kill_pyro_mip_servers`, and `launch_pyro_mip_server`.~~
 
With all of the above changes, it leaves us with just `pyomo` and `test.pyomo` getting installed into bin (when deprecated items are removed). We might just move all Pyro functionality into the standalone pysp project.

### Legal Acknowledgement


ghackebeil:
I took a shot at migrating the Pyro-related commands under a `pyomo pyro ...` subcommand. It turns out that each of these ingest command-line arguments in subtly different ways, and are pretty much incompatible with how argparse subparsers work.

Rather than just deprecate the entry-point commands (and provide an alternative subcommand for them), my vote is to deprecate them, including SolverManagerPyro, and plan to remove them sometime in the future. Those pieces of code are fragile, and they have been **barely** maintained over the last decade.

pyomo/scripting/commands.py
```diff
@@ -80,40 +81,12 @@ def launch_pyro_mip_servers():
...
-@pyomo_command('readsol', "Read and print a *.sol file")
-def readsol():
-    import pyomo.opt
-
-    reader = pyomo.opt.ReaderFactory("sol")
-    soln = reader(sys.argv[1])
-    soln.write()
-
 @pyomo_command('pyomo_python', "Launch script using Pyomo's python installation")
+@deprecated(
```
mrmundt, line 85: So one thing about deprecation: 
We normally put `TBD` as the `version` instead of explicitly stating the version (we replace all appropriate `TBD`s with the correct version during the release process). We also aren't in the habit of using the `remove_in`.

mrmundt (changes requested):

mrmundt (approved):

jsiirola (approved):
This looks good.  My only question is if we should go farther and flat out remove all console scripts other than `pyomo`

ghackebeil:
Besides `test.pyomo`, we basically are if we decide to move all Pyro functionality into pysp.

mrmundt:
> Besides `test.pyomo`, we basically are if we decide to move all Pyro functionality into pysp.

Which we decided to do today during the dev call. Hooray!

mrmundt:
@jsiirola , I vote we merge this and then get rid of the residual console scripts that are caused by the Pyro functionality.

setup.py
```diff
@@ -134,25 +134,14 @@ def run_setup():
...
-        OSSolverService = pyomo.scripting.commands:OSSolverService
-        pyomo_python = pyomo.scripting.commands:pyomo_python
-        pyomo_old=pyomo.scripting.pyomo_command:main
-
-        [pyomo.command]
-        pyomo.help = pyomo.scripting.driver_help
-        pyomo.test.pyomo = pyomo.scripting.runtests
-        pyomo.pyro_mip_server = pyomo.scripting.pyro_mip_server
-        pyomo.results_schema=pyomo.scripting.commands
-        pyomo.viewer=pyomo.contrib.viewer.pyomo_viewer
```
jsiirola, line 155: I think we need to leave these.  The pyomo.command entrypoints are all loaded by the main pyomo script (in `pyomo/scripting/pyomo_main.py`).  This is nominally to provide a registration mechanism outside of `pyomo.environ`.  I believe that it even allows other packages to register subcommands for the `pyomo` command.  We may even make use of it, e.g., for PySP to register subcommands.

jsiirola (changes requested):

setup.py
ghackebeil: Thanks for clearing that up. I could not find any documentation about subsections in entry points. I'll take a closer look at each of these.

jsiirola (approved):

PR #1833, opened 2021-02-23
Replace `six` usage in pyomo.core

mrmundt:
## Summary/Motivation:
A task in the Pyomo 6.0 release is to replace instances of `six` with the appropriate Python 3 uses. This starts that effort in `pyomo.core`

- Replace instances of `six` in `pyomo.core`

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/block.py
+++ b/pyomo/core/base/block.py
@@ -1401,11 +1401,11 @@ Components must now specify their rules explicitly using 'rule=' keywords.""" %
         generator recursively descends into sub-blocks.
         """
         if not descend_into:
-            for x in self.component_map(ctype, active, sort).itervalues():
+            for x in self.component_map(ctype, active, sort).values():
                 yield x
             return
         for _block in self.block_data_objects(active, sort, descend_into, descent_order):
-            for x in _block.component_map(ctype, active, sort).itervalues():
+            for x in _block.component_map(ctype, active, sort).values():
                 yield x
 
     def component_data_objects(self,
--- a/pyomo/core/base/block.py
+++ b/pyomo/core/base/block.py
@@ -20,8 +20,7 @@ import textwrap
 
 from inspect import isclass
 from operator import itemgetter
-from six import iteritems, iterkeys, itervalues, StringIO, string_types, \
-    advance_iterator, PY3
+from io import StringIO
 
 from pyomo.common.collections import Mapping
 from pyomo.common.deprecation import deprecated, deprecation_warning
@@ -289,7 +288,7 @@ class PseudoMap(object):
         sort_order = self._sorted
         try:
             self._sorted = False
-            for x in itervalues(self):
+            for x in self.values():
                 return True
             return False
         finally:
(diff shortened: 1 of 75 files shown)


pyomo/core/tests/unit/test_var_set_bounds.py
```diff
@@ -152,6 +151,8 @@ def Xtest_list_domain(self):
         self.model.con2 = Constraint(expr=self.model.y[2] <= 2.9)
         
         self.instance = self.model.create_instance()
+        # !! THIS SEEMS LIKE A BUG!! - mrmundt
+        # I think it's supposed to be SolverFactory
         self.opt = solver["glpk"]
```
blnicho, line 156: I think you're right but it's extra weird that this test isn't failing. Do we have `glpk` in our testing environment? Could you check if this test is being run or not?

blnicho (approved):

pyomo/core/expr/visitor.py
```diff
@@ -12,28 +12,17 @@
...
 from copy import deepcopy
 from collections import deque
 
-if six.PY2:
-    getargspec = inspect.getargspec
-else:
-    # For our needs, getfullargspec is a drop-in replacement for
-    # getargspec (which was removed in Python 3.x)
-    getargspec = inspect.getfullargspec
+getargspec = inspect.getfullargspec
```
jsiirola, line 18: We should either `from inspect import getfullargspec` or just use `inspect.getfullargspec` below.

pyomo/core/tests/unit/test_block.py
```diff
@@ -11,13 +11,11 @@
 # Unit Tests for Elements of a Block
 #
 
+import io
 import os
 import sys
-import six
 import types
 
-from six import StringIO, iterkeys
```
jsiirola, line 19: Any reason why you didn't just `from io import StringIO`?

pyomo/core/tests/unit/test_var_set_bounds.py
mrmundt: It's not running. But I also tried it locally - nosetests doesn't pick up any of the names because they all start with `X`. So someone turned off all these tests.

pyomo/core/base/block.py
```diff
@@ -435,12 +434,10 @@ def items(self):
...
-# In Python3, the items(), etc methods of dict-like things return
-# generator-like objects.
-if PY3:
-    PseudoMap.keys = PseudoMap.iterkeys
-    PseudoMap.values = PseudoMap.itervalues
-    PseudoMap.items = PseudoMap.iteritems
+
+PseudoMap.keys = PseudoMap.iterkeys
+PseudoMap.values = PseudoMap.itervalues
+PseudoMap.items = PseudoMap.iteritems
```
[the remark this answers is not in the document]
mrmundt: Any preferences on what we rename them to?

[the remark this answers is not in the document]
jsiirola: We should rename `PseudoMap.iteritems()` to `PseudoMap.items()`, then add:
```python
class Pseudomae(object):
    #...
    @deprecated()
    def iteritems(self):
         return self.items()
```

pyomo/core/tests/unit/test_var_set_bounds.py
mrmundt: DECISION: We're going to handle all `Xtest` instances in a separate PR.

jsiirola (approved):

PR #1837, opened 2021-02-24
GDPopt RIC Option

michaelbynum:
- Add an option to GDPopt to solve a relaxation with integer cuts but without outer approximation cuts or MC++
- Remove the default GDPopt strategy. The strategy must be specified now.

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/gdpopt/config_options.py
+++ b/pyomo/contrib/gdpopt/config_options.py
@@ -20,6 +20,7 @@ def _get_GDPopt_config():
         'LOA',  # Logic-based outer approximation
         'GLOA',  # Global logic-based outer approximation
         'LBB',  # Logic-based branch-and-bound
+        'RIC',  # Relaxation with Integer Cuts
     }
     CONFIG = ConfigBlock("GDPopt")
     CONFIG.declare("iterlim", ConfigValue(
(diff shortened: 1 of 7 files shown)


bernalde:
I think this makes sense from a pure enumeration way of using GDPopt.
The failures in the tests were not part of the PR (both were from timing) and it makes sense to remove the default behavior to be LOA.

carldlaird (approved):
This looks reasonable to me.

pyomo/contrib/gdpopt/config_options.py
```diff
@@ -35,7 +36,7 @@ def _get_GDPopt_config():
             "need to set subsolver time limits as well."
     ))
     CONFIG.declare("strategy", ConfigValue(
-        default="LOA", domain=In(_supported_strategies),
+        default=None, domain=In(_supported_strategies),
```
jsiirola, line 39: Why are you taking away the default strategy?

michaelbynum: Because the strategy that should be used depends heavily on the problem being solved (convex/nonconvex, number of disjunctions, etc). Users should be forced to specify a strategy based on the problem they are solving.

PR #1851, opened 2021-03-05
MANIFEST File

michaelbynum:
## Summary/Motivation:
This PR adds a MANIFEST.in file so that the cpp files for pynumero and the MC++ interface get distributed by PyPI. This allows users to pip install and then use the pyomo build-extensions command. Currently, that command only works if you install from source (because the cpp files don't exist if you install with pip).

### Legal Acknowledgement

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,7 +1,4 @@
 include README.md
 include LICENSE.txt
-include pyomo/contrib/mcpp/mcppInterface.cpp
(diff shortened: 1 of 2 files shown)


michaelbynum:
I had to modify `setup.py` slightly to get this to work correctly. I tested this locally, and the cpp files do get distributed.

blnicho (approved):

PR #1878, opened 2021-03-18
Fix bug in identify_mutable_parameters

### LINKED ISSUE #1877: identify_mutable_parameters sometimes identifies numeric constants
There seem to be some circumstances in which `identify_mutable_parameters` identifies numeric constants. As far as I can tell, this happens when the numeric constant is alone as one argument of a logical expression. The following code reproduces the issue:
```python
import pyomo.environ as pyo
from pyomo.core.expr.visitor import identify_mutable_parameters

m = pyo.ConcreteModel()
m.p = pyo.Param(mutable=True)
expr = m.p - 5 == 0

for p in identify_mutable_parameters(expr):
    print(p.to_string())
```
This prints:
```
0.0
p
```

Robbybp:
## Summary/Motivation:
Fixes an apparent bug in `identify_mutable_parameters`. 
The logic used to identify a "mutable parameter" was
```python
not node.is_variable_type() and node.is_fixed()
```
I'm not sure how this has not been identifying all numeric constants as mutable params, but the bug only seems to appear in certain types of expressions. This logic has been changed to the following, which seems to fix the bug.
```python
not node.is_variable_type() and not node.is_constant()
```

- Change logic as described above
- Add test that fails with old logic

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/expr/visitor.py
+++ b/pyomo/core/expr/visitor.py
@@ -1181,7 +1181,8 @@ class _MutableParamVisitor(SimpleExpressionVisitor):
             return
 
         # TODO: Confirm that this has the right semantics
-        if not node.is_variable_type() and not node.is_constant():
+        if (not node.is_variable_type() and node.is_fixed()
+                and not node.is_constant()):
             if id(node) in self.seen:
                 return
             self.seen.add(id(node))
--- a/pyomo/core/expr/visitor.py
+++ b/pyomo/core/expr/visitor.py
@@ -1181,7 +1181,7 @@ class _MutableParamVisitor(SimpleExpressionVisitor):
             return
 
         # TODO: Confirm that this has the right semantics
-        if not node.is_variable_type() and node.is_fixed():
+        if not node.is_variable_type() and not node.is_constant():
             if id(node) in self.seen:
                 return
             self.seen.add(id(node))
--- a/pyomo/core/tests/unit/test_visitor.py
+++ b/pyomo/core/tests/unit/test_visitor.py
@@ -33,6 +33,7 @@ from pyomo.core.expr.visitor import (
 from pyomo.core.base.param import _ParamData, SimpleParam
 from pyomo.core.expr.template_expr import IndexTemplate
 from pyomo.core.expr.expr_errors import TemplateExpressionError
+from pyomo.common.collections import ComponentSet
 from pyomo.common.log import LoggingIntercept
 from io import StringIO
 
@@ -190,6 +191,17 @@ class TestIdentifyParams(unittest.TestCase):
         self.assertEqual( list(identify_mutable_parameters(m.b+m.e)), [ m.b, m.a ] )
         self.assertEqual( list(identify_mutable_parameters(m.E[0])), [ m.a ] )
         self.assertEqual( list(identify_mutable_parameters(m.E[1])), [ m.b ] )
+
+    def test_identify_mutable_parameters_logical_expr(self):
+        #
+        # Identify mutable params in logical expressions
+        #
+        m = ConcreteModel()
+        m.a = Param(initialize=0, mutable=True)
+        expr = m.a+1 == 0
+        param_set = ComponentSet(identify_mutable_parameters(expr))
+        self.assertEqual(len(param_set), 1)
+        self.assertIn(m.a, param_set)
 
     def test_identify_mutable_parameters_params(self):
         m = ConcreteModel()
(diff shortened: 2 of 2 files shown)


michaelbynum (approved):
This looks good to me. The test failures seem unrelated. I'll restart them.

pyomo/core/expr/visitor.py
```diff
@@ -1181,7 +1181,7 @@ def visit(self, node):
             return
 
         # TODO: Confirm that this has the right semantics
-        if not node.is_variable_type() and node.is_fixed():
+        if not node.is_variable_type() and not node.is_constant():
```
jsiirola, line 1184: I think you still need the `is_fixed` test to make sure you don't catch other nodes types like Expression objects.  That is,
```python
if not node.is_variable_type() and node.is_fixed() and not node.is_constant():
```

jsiirola (changes requested):
I think you need to re-add the `is_fixed()` test so that we filter out things like `Expression` objects.  Also note that this doesn't actually *fix* `identify_mutable_parameters`, as it still misses mutable params in places like `LinearExpression` objects.

pyomo/core/expr/visitor.py
michaelbynum: I was thinking about this too. I'm certainly fine with leaving the `is_fixed` check, but doesn't the `xbfs_yield_leaves` method already filter out expressions?

michaelbynum: Also, is it possible to have expression objects for which is_fixed returns True? I'm thinking about the NPV expressions.

jsiirola: You are correct: some part of the walker does correctly handle things like `Expression` and `MonomialTermExpression`.

Yes: `is_fixed()` should return correct values for any expression object (it triggers a walk of the expression tree).  That said, NPV objects have `is_constant()` True by decree (no walk required - in fact, as I recall, NPV nodes were added explicitly to speed up the `is_constant` test).

michaelbynum: I think the `LinearExpression` issue points to a bigger problem, which is that the `args` and `nargs()` properties/methods do not return anything meaningful for `LinearExpression`.

michaelbynum: So we also need to make sure that the `identify_mutable_parameters` function does not return NPV expressions:

```

In [3]: m.p = pe.Param(mutable=True)

In [4]: e = 2*m.p

In [5]: e
Out[5]: <pyomo.core.expr.numeric_expr.NPV_ProductExpression at 0x7ff1becfcb10>

In [6]: e.is_constant()
Out[6]: False

In [7]: e.is_fixed()
Out[7]: True
```

Robbybp: I just re-added the `is_fixed()` check, but I can't think of a way to test it. That is, I can't think of a non-variable leaf node that has both `is_constant()` False and `is_fixed()` False.

Robbybp: > So we also need to make sure that the `identify_mutable_parameters` function does not return NPV expressions

Is it possible for an NPV expression to be leaf node?

jsiirola: No - I don't think so.  Apart from `Expression` and `LinearExpression`, no expression-like objects should be leaf nodes.

jsiirola (approved):

jsiirola:
Actually, can you add a test that ensures that things like NumericConstants and ScalarParams are not picked up?

pyomo/core/tests/unit/test_visitor.py
```diff
@@ -167,6 +168,23 @@ def test_identify_mutable_parameters(self):
...
+    def test_identify_mutable_parameters_constants(self):
+        #
+        # SimpleParams and NumericConstants are not recognized
+        #
+        m = ConcreteModel()
+        m.x = Var(initialize=1)
+        m.x.fix()
+        m.p = Param(initialize=2, mutable=False)
+        m.p_m = Param(initialize=3, mutable=True)
+        e1 = (m.x + m.p + 5)
```
jsiirola, line 180: Can we make `5` be `NumericConstant(5)` (from `pyomo.core.expr.numvalue`)

Robbybp: Done. Assumed this was happening under the hood

pyomo/core/expr/visitor.py
michaelbynum: Sorry - I apparently can't keep it in my head that this uses `xbfs_yield_leaves`.

PR #1888, opened 2021-03-25
`incidence_analysis` contrib package for maximum matching and block triangularization

Robbybp:
## Summary/Motivation:
Performing a maximum matching or block triangularization on (some subset of) the incidence graph of variables and constraints in a Pyomo model is useful for debugging structured models. This PR adds some functions that perform these analyses on sparse matrices (via NetworkX), as well as an interface to perform them on Pyomo models/variables/constraints.

- Add a `matching` contrib package
- `maximum_matching` function
- `block_triangularize` function
- `IncidenceGraphInterface` class to call these functions on Pyomo incidence matrices

### Legal Acknowledgement


pyomo/contrib/matching/interface.py
```diff
@@ -0,0 +1,79 @@
...
+class IncidenceGraphInterface(object):
+    """
+    The purpose of this class is to allow the user to easily
+    analyze graphs of variables and contraints in a Pyomo
+    model without constructing multiple PyomoNLPs.
+    """
+
+    def __init__(self, model):
+        self.nlp = PyomoNLP(model)
+
```
jsiirola, line 25: I think I would like to see another method here to reduce the amount of duplicated code:

```python
def get_incidence_matrix(self, obj, variables, constraints):
    if isinstance(model, PyomoNLP):
        return self.get_pyomonlp_incidence_matrix(obj, variables, constraints)
    else:
        return self.get_pyomo_incidence_matrix(obj, variables, constraints)

def get_pyomonlp_incidence_matrix(self, nlp, variables, constraints):
    if variables is None:
        variables = nlp.get_pyomo_variables()
    if constraints is None:
        constraints = nlp.get_pyomo_constraints()

    for comp in variables+constraints:
        if comp.is_indexed():
            raise ValueError(
                        "Variables and constraints must be unindexed "
                        "ComponentData objects. Got %s, which is indexed."
                        % comp.name
                        )
    matrix = nlp.extract_submatrix_jacobian(variables, constraints)
    return variables, constraints, matrix

def get_pyomo_incidence_matrix(self, model, variables, constraints):
    _check = []
    if variables is None:
        variables = list(model.component_data_objects(Var))
    else:
        _check.extend(variables)
    if constraints is None:
        constraints = list(model.component_data_objects(Constraint, active=True))
    else:
        _check.extend(constraints)

    for comp in _check:
        if comp.is_indexed():
            raise ValueError(
                        "Variables and constraints must be unindexed "
                        "ComponentData objects. Got %s, which is indexed."
                        % comp.name
                        )
    _varID = ComponentMap((v,i) for i,v in enumerate(variables))
    _rows = []
    _cols = []
    for i, c in enumerate(constraints):
        _cols.extend(_varID[v] for v in identify_variables(c.body) if v in _varID)
        _rows.extend([i]*(len(_cols)-len(_rows)))
    matrix = scipy.sparse.coo_matrix( ([1]*len(_rows), (_rows, _cols)), 
            shape=(len(constraints), len(variables)) )
    return variables, constraints, matrix

def maximum_matching(self, variables=None, constraints=None):
    var, con, matrix = self.get_incidence_matrix(self.nlp, variables, constraints)
    matching = maximum_matching(matrix.tocoo())
    # Matching maps row (constraint) indices to column (variable) indices
    return ComponentMap((con[i], varj]) for i, j in matching.items())

def block_triangularize(self, variables=None, constraints=None):
    var, con, matrix = self.get_incidence_matrix(self.nlp, variables, constraints)
    ....
```


```

File: pyomo/contrib/matching/interface.py [hunk 2]
```diff
@@ -0,0 +1,79 @@
...
+class IncidenceGraphInterface(object):
+    """
+    The purpose of this class is to allow the user to easily
+    analyze graphs of variables and contraints in a Pyomo
+    model without constructing multiple PyomoNLPs.
+    """
+
+    def __init__(self, model):
+        self.nlp = PyomoNLP(model)
```
jsiirola, line 24: I don't really see the value in caching a local PyomoNLP.  Wouldn't it make more sense to just work with whatever the user give us?

pyomo/contrib/matching/tests/test_interface.py
```diff
@@ -0,0 +1,169 @@
...
+#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain 
+#  rights in this software.
+#  This software is distributed under the 3-clause BSD License.
+#  ___________________________________________________________________________
+
+import pyomo.environ as pyo
+from pyomo.common.dependencies import networkx_available
+from pyomo.common.dependencies import scipy_available
+from pyomo.common.collections import ComponentSet
+from pyomo.contrib.matching.interface import IncidenceGraphInterface
```
jsiirola, line 15: I think you need to trap for `networkx_available` before this import: `interface` imports `block_triangularize` which unconditionally imports `networkx`.  Alternatively, you should remove the unconditional imports (see comment in `block_triangularize` and `maximal_matching`

pyomo/contrib/matching/maximum_matching.py
```diff
@@ -0,0 +1,42 @@
...
+#  Under the terms of Contract DE-NA0003525 with National Technology and 
+#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain 
+#  rights in this software.
+#  This software is distributed under the 3-clause BSD License.
+#  ___________________________________________________________________________
+
+from pyomo.common.dependencies import networkx as nx
+nxb = nx.algorithms.bipartite
+nxc = nx.algorithms.components
+from_biadjacency_matrix = nxb.matrix.from_biadjacency_matrix
```
jsiirola, line 14: These cause `networkx` to be unconditionally imported when this module is imported.  You might consider moving them into the `maximal_matching()` function.

pyomo/contrib/matching/block_triangularize.py
```diff
@@ -0,0 +1,98 @@
...
+#  rights in this software.
+#  This software is distributed under the 3-clause BSD License.
+#  ___________________________________________________________________________
+
+from pyomo.contrib.matching.maximum_matching import maximum_matching
+from pyomo.common.dependencies import networkx as nx
+nxb = nx.algorithms.bipartite
+nxc = nx.algorithms.components
+nxd = nx.algorithms.dag
+from_biadjacency_matrix = nxb.matrix.from_biadjacency_matrix
```
jsiirola, line 16: These cause `networkx` to be unconditionally imported when this module is imported.  You might consider moving them into the `block_triangularize()` function.

jsiirola (changes requested):
I think you need to add `__init__.py` files in `pyomo/contrib/matching` and `pyomo/contrib/matching/tests` (in additional to the comments below).

pyomo/contrib/matching/interface.py
Robbybp: For this interface class, I think I prefer just working with a Pyomo model. My reasoning was that if the user already has a PyomoNLP that they want to work with, they can call `extract_submatrix_jacobian` themself and use the `maximum_matching` and `block_triangularize` functions directly on the matrix. I cache the PyomoNLP so we can call the methods repeatedly with different components without repeating the work of writing/parsing the nl file. The desire to cache the PyomoNLP then block triangularize many subsets of components was the motivation for creating this class in the first place.

I can think of a couple improvements:
1. Don't cache the PyomoNLP, then at the time the submatrix is needed, construct a smaller PyomoNLP based on only the variables and constraints the user asked for (by attaching these components to a dummy block as references).
2. Don't create a PyomoNLP at all, then construct the sparse matrix with `identify_variables`, or if we need numeric derivatives, Pyomo's AD/symbolic differentiation functions.

For both of these options, we don't need to cache anything, and also don't need access to the user's model, so I don't think they need to be part of this (or any) class at all. We could add additional functions for these in the `block_triangularize` and `maximum_matching` modules. We should at least do this for an implementation that only uses a structural Jacobian.

pyomo/contrib/matching/interface.py
Robbybp: See comment below. I added a method to get rid of the repeated code, but I believe "structural incidence matrix" code should probably be added as a function independent of this class. I can do that as part of this PR.

pyomo/contrib/matching/interface.py
```diff
@@ -0,0 +1,74 @@
...
+        """
+        variables, constraints = self._validate_input(variables, constraints)
+
+        matrix = self.nlp.extract_submatrix_jacobian(variables, constraints)
+        row_block_map, col_block_map = block_triangularize(matrix.tocoo())
+        con_block_map = ComponentMap((constraints[i], idx)
+                for i, idx in row_block_map.items())
+        var_block_map = ComponentMap((variables[j], idx)
+                for j, idx in col_block_map.items())
+        # Switch the order of the maps here to matching the method call.
```
blnicho, line 72: Typo: "to matching" should be "to match"

blnicho (approved):
I found one typo but otherwise this looks fine

pyomo/contrib/matching/interface.py
```diff
@@ -0,0 +1,119 @@
...
+        if constraints is None:
+            constraints = self.nlp.get_pyomo_constraints()
+
+        for comp in variables+constraints:
+            if comp.is_indexed():
+                raise ValueError(
+                        "Variables and constraints must be unindexed "
+                        "ComponentData objects. Got %s, which is indexed."
+                        % comp.name
+                        )
```
jsiirola, line 85: Shouldn't this just call `_check_unindexed()` instead of repeating the code here?

jsiirola (approved):
I don't see any reason to block this PR; however, I still feel that the asymmetry between the `Block` and `PyomoNLP` interfaces is troublesome: the user interacts with this code in a very different manner if they want to work with a PyomoNLP (there is an interface class that handles everything for them) versus a Pyomo `Block` (they have to roll everything themselves).

I very much like that the code functions are exposed (getting the incidence graph, `maximum_matching()` and `block_triangularize()`), but I feel somewhat strongly that if you are going to have an "Interface" class, it should present the user the same interface regardless of if they are working on a Block or a PyomoNLP object, especially because the "user-facing" interface is always the same (Pyomo variables & constraints).

I also feel that `IncidenceGraphInterface` as written is a misleading name, as it is really the Pynumero Incidence Graph interface.

Robbybp:
@jsiirola I actually never intended for there to be a PyomoNLP interface. I embed a PyomoNLP as a convenient way to get a numeric incidence matrix. You're saying that if the user happens to already have a PyomoNLP, I should be able to use it instead of constructing another one. I'm inclined to agree. What I think is more troublesome is the difference in API between, say, block triangularizing a structural incidence matrix and a numeric one:
1. No model is needed to get a structural incidence matrix, whereas a model is needed to get a numeric incidence matrix (because of the embedded PyomoNLP)
2. The user must wrap the structural incidence matrix themself, whereas the numeric incidence matrix is wrapped for them by the interface class.

---

As of just now, there is a function to get the numeric incidence matrix that does not require a model, i.e. only constructs a PyomoNLP that contains the relevant subset of variables and constraints. I think the next questions are:
1. Whether to have an interface class
2. If not, how (or whether) to wrap the "lower level" functions to return a matching and triangularization in terms of the Pyomo variables and constraints.
I think I lean towards:
1. No.
2. Put `identify_maximum_matching` and `identify_block_triangularization` functions that act on Pyomo components in the `interface` module, then rename the lower level functions, e.g. `maximum_matching_matrix` or something like that.
I could see some other ways to do the "Pyomo interface," however.

Robbybp:
I'm also not entirely sure about some of the names. If we do away with the interface class, the name is moot, although I agree that the name is too general for its current functionality. I'm also not sure what I think the name for this contrib package should be. "matching" seems too specific. It's really an interface to create incidence matrices/graphs, and a couple algorithms for ordering/partitioning these structures. Something related to "incidence" or "ordering" might be better.

pyomo/contrib/structural_analysis/interface.py
```diff
@@ -0,0 +1,237 @@
...
+        # WARNING: This cache will become invalid if the user alters their
+        #          model.
+        self.cached = IncidenceMatrixType.NONE
+        if isinstance(model, PyomoNLP):
+            nlp = model
+            self.cached = IncidenceMatrixType.NUMERIC
+            self.variables = nlp.get_pyomo_variables()
+            self.constraints = nlp.get_pyomo_constraints()
+            self.var_index_map = nlp._vardata_to_idx
+            self.con_index_map = nlp._condata_to_idx
```
jsiirola, line 138: Shouldn't we rebuild these maps instead of accessing private members of the PyomoNLP class?

jsiirola (commented):
This looks really good.  Only one (minor) question.

pyomo/contrib/structural_analysis/interface.py
Robbybp: Yep, changed.

Robbybp:
Renamed this again to `incidence_analysis`, because "structure" could mean many things (hierarchical structure, structure due to indexing, etc.). Maybe "incidence" can too but I find it more telling.

jsiirola (approved):

PR #1895, opened 2021-03-29
[WIP] Model partition package

rahuljoglekar47:
## Summary/Motivation:
To take the code from PR #1526 and to make changes to make the code easier to use with other model partitioning software (since PR #1526 was designed for Louvain community detection)

- This PR builds off of PR #1526 and makes some minor changes to make the code easier to use with other model partitioning software (since PR #1526 was designed for Louvain community detection)
- This PR also makes the generate_structured_model method of the CommunityMap class much less complicated

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/community_detection/detection.py
+++ b/pyomo/contrib/community_detection/detection.py
@@ -10,7 +10,7 @@ Original implementation developed by Rahul Joglekar in the Grossmann research gr
 from logging import getLogger
 
 from pyomo.common.dependencies import attempt_import
-from pyomo.core import ConcreteModel, ComponentMap, Block, Var, Constraint, Objective, ConstraintList
+from pyomo.core import ConcreteModel, ComponentMap, Block, Var, Constraint, Objective, ConstraintList, Reference
 from pyomo.core.base.objective import _GeneralObjectiveData
 from pyomo.core.expr.current import identify_variables
 from pyomo.core.expr.visitor import replace_expressions
@@ -144,14 +144,14 @@ def detect_communities(model, type_of_community_map='constraint', with_objective
         # separate the nodes into their two groups; thus, we create a list of constraints and a list of variables
 
         for community_key in community_map:
-            constraint_node_list, variable_node_list = [], []
+            constraint_list, variable_list = [], []
             node_community_list = community_map[community_key]
             for numbered_node in node_community_list:
                 if numbered_node in constraint_variable_map:
-                    constraint_node_list.append(number_component_map[numbered_node])
+                    constraint_list.append(number_component_map[numbered_node])
                 else:
-                    variable_node_list.append(number_component_map[numbered_node])
-            community_map[community_key] = (constraint_node_list, variable_node_list)
+                    variable_list.append(number_component_map[numbered_node])
+            community_map[community_key] = (constraint_list, variable_list)
 
     elif type_of_community_map == 'constraint':
         # If the community map was created for a constraint node graph, then for a given community, we want to create a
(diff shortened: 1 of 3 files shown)


pyomo/contrib/community_detection/detection.py
```diff
@@ -494,10 +494,21 @@ def generate_structured_model(self):
 
         # Initialize a new model (structured_model) which will contain variables and constraints in blocks based on
         # their respective communities within the CommunityMap
+
+        original_model = self.model
         structured_model = ConcreteModel()
 
         structured_model.b = Block(self.keys())
 
+        for variable in original_model.component_data_objects(ctype=Var, descend_into=True):
```
rahuljoglekar47, line 503: One question I have at the moment regarding this generate_structured_model code:
1.) I created this for loop because without it, I get the following error when I try to solve the new model:

```
  File "D:/College/Git/pyomo/pyomo/contrib/community_detection/scratch_work_pt2.py", line 613, in <module>
    opt_success = opt.solve(new_model)
  File "D:\College\Git\pyomo\pyomo\solvers\plugins\solvers\GAMS.py", line 785, in solve
    io_options=io_options)
  File "D:\College\Git\pyomo\pyomo\core\base\block.py", line 1814, in write
    io_options)
  File "D:\College\Git\pyomo\pyomo\repn\plugins\gams_writer.py", line 487, in __call__
    put_results_format=put_results_format,
  File "D:\College\Git\pyomo\pyomo\repn\plugins\gams_writer.py", line 646, in _write_model
    tc(var)
  File "D:\College\Git\pyomo\pyomo\repn\plugins\gams_writer.py", line 238, in __call__
    self.raise_error(comp)
  File "D:\College\Git\pyomo\pyomo\repn\plugins\gams_writer.py", line 255, in raise_error
    "All components must have the same parent model." % comp.name)
RuntimeError: GAMS writer: found component 'x1' not on same model tree.
All components must have the same parent model.
```

Robbybp: Are you trying to solve the `structured_model` that gets returned by this method, or one of its sub-blocks? Assuming the former, I'm not entirely sure why you're getting this error. Shouldn't all the variables belong to some community, and thus should get added onto a subblock by the call to `Reference` below? If there is some reason that a variable is not added to a subblock, then I think this error makes sense. 

That said, I would not delete the variables from the user's model. It should be sufficient to add them as references, right? Or is part of the problem that this is not sufficient?
```
structured_model.add_component("variables",
        Reference(list(original_model.component_data_objects(Var))))
```
It would be easier to debug if we had a copy of the model you're using here. Is this available somewhere, or simple enough to paste?

rahuljoglekar47: That's right, I am trying to solve the `structured_model` that is returned by the method. If I do not include the for loop to delete a variable from the original model and then add the variable to the new model, I get the error above (so I believe adding them as references is not sufficient for some reason).

Also, here is a code snippet that results in the error above (only when run without the above for loop):

```
def create_model_6():  # Toy model
    model = m = ConcreteModel()
    m.x1 = Var(bounds=(0, 1))
    m.x2 = Var(bounds=(0, 1))
    m.x3 = Var(bounds=(0, 1))
    m.x4 = Var(bounds=(0, 1))
    m.obj = Objective(expr=m.x1, sense=minimize)
    m.c1 = Constraint(expr=m.x1 + m.x2 >= 1)
    m.c2 = Constraint(expr=m.x3 + m.x4 >= 1)
    return model

seed = 5  # To be used as a random seed value for the heuristic Louvain community detection
community_map_object = detect_communities(create_model_6(), type_of_community_map='bipartite', random_seed=seed)
new_model = community_map_object.generate_structured_model()
opt = SolverFactory('gams')
opt_success = opt.solve(new_model)
```

blnicho:
@rahuljoglekar47 @bernalde is this still being worked on or can we close it?

bernalde:
I think that this PR can be safely closed as Rahul is no longer involved in the project and it might be superseded by some of the work done by @michaelbynum

PR #1904, opened 2021-04-01
Allow string args to external fuctions

### LINKED ISSUE #1903: NL-writer External Function Issues
I have two issues with the NL-writer related to external functions.

1. If you pass a constant number directly to an external function like:  ``...expr=3 == m.ext_func(m.x, 4.0)...`` you get the error "AttributeError: 'float' object has no attribute 'is_fixed'".  You only see this if not all the args are const, otherwise the writer seems to realize the function evaluation is a constant and do it before the nl-file.
2. If you pass a string, at least with ipopt like ``...expr=3 == m.ext_func(m.x, "a string")...`` you get an error from ipopt complaining of a malformed line in the nl file.  The malformed line is the one where the string arg is defined.  As far as I can tell the line isn't really malformed, but something isn't right.  This also only happens if one arg is variable, otherwise the function is pre-evaluated and everything works fine.

eslickj:
## Fixes #1903 

This adds support for passing strings to external AMPL functions and fixes issues in #1903.  The string args on Windows get messed up by the \r\n newline, so I forced \n to be used everywhere. Also catch float and int before ``is_fixed()`` is checked so you can directly pass constant args to external functions.  

### Legal Acknowledgement


eslickj:
Attached is c source for a function I've been using for testing.  It could be used to write tests for this with some modification. 
[functions.zip](https://github.com/Pyomo/pyomo/files/6250167/functions.zip)

blnicho:
@eslickj could you add some tests for this PR in the following 2 files: https://github.com/Pyomo/pyomo/blob/main/pyomo/repn/tests/ampl/test_ampl_nl.py#L112 and https://github.com/Pyomo/pyomo/blob/main/pyomo/core/tests/unit/test_external.py

If you can use AmplGSL functions to test this I think that would be preferable to adding new C code to the repo but I realize you need an external function that accepts both floats and strings as arguments and I don't know if such a function exists in AmplGSL.

eslickj:
@blnicho, I'm still looking, but I don't think they have functions with string args. I'm going to look some more, but I could add a function specifically for testing to the IDAES function library, but I'm not sure how hard that would be to use in a test.

eslickj:
Note so I don't forget.  The string problem could be a windows issue if the NL-files are written with windows style line breaks on windows.  Not sure, but should check that out.  If that's the case this may fix string args for windows and break them on Linux, but it should be easy to handel.

eslickj:
Windows test fail now due to the nl files switching from the \r\n new line to just \n, so the nl file is different from expected.  If the general approach in this PR is okay, those tests will need to be updated.

eslickj:
@jsiirola, I really opened a can of worms with the new line issue.  I added a new file compare utility function for the tests.  It may not be the greatest, but I don't really see any alternative a good alternative.  There are a lot of ways end-of-line characters could end-up not matching.  An alternative to this could to update all the text file writes for use "\n" and make git not auto set the new lines on windows to "\r\n" when you checkout on Windows, but I think that's even messier.

jsiirola:
@eslickj I am just now trying to get caught up on this PR, but I am a bit confused: why did so many tests unrelated to this change have to be updated to use different file comparisons?

eslickj:
> 
> 
> @eslickj I am just now trying to get caught up on this PR, but I am a bit confused: why did so many tests unrelated to this change have to be updated to use different file comparisons?

@jsiirola Good question.  Every test that writes an NL file and checks it against a baseline would need updated.  I think that accounts for most. The cplex and mps ones, actually probably didn't need updated.  I changed my own git setting to not convert line ending to CRLR, and that caused a whole new set of tests to fail, since those file still get written with the CRLF EOL on Windows.  We can roll those test changes back.  Line endings are a mess.

jsiirola:
I am still confused: we have been writing and testing NL files against baselines on Linux/Windows/OSX for a long time and haven't seen this issue.  I know we changed our file comparison utilities a little while ago (moving to standard utilities wherever we could and away from "home grown" comparisons).  Maybe one of those tests is more fragile than it should be.  @mrmundt: any thoughts here?

mrmundt:
> I am still confused: we have been writing and testing NL files against baselines on Linux/Windows/OSX for a long time and haven't seen this issue. I know we changed our file comparison utilities a little while ago (moving to standard utilities wherever we could and away from "home grown" comparisons). Maybe one of those tests is more fragile than it should be. @mrmundt: any thoughts here?

When we recently switched over to more standard utilities, we did have to implement some finicky hacks. My approach was to generally compare line-by-line and ignore any newlines as opposed to comparing the whole contents of the files. For example:

```
    def filter_items(self, items):
        filtered = []
        for i in items:
            if not (i.startswith('/') or i.startswith(":\\", 1)):
                try:
                    filtered.append(float(i))
                except:
                    filtered.append(i)
        return filtered

    def compare_files(self, file1, file2):
        try:
            self.assertTrue(cmp(file1, file2),
                        msg="Files %s and %s differ" % (file1, file2))
        except:
            with open(file1, 'r') as f1, open(file2, 'r') as f2:
                f1_contents = f1.read().strip().split('\n')
                f2_contents = f2.read().strip().split('\n')
                f1_filtered = []
                f2_filtered = []
                for item1, item2 in zip_longest(f1_contents, f2_contents):
                    if not item1.startswith('['):
                        items1 = item1.strip().split()
                        items2 = item2.strip().split()
                        f1_filtered.append(self.filter_items(items1))
                        f2_filtered.append(self.filter_items(items2))
                self.assertStructuredAlmostEqual(f2_filtered, f1_filtered,
                                                 abstol=1e-6,
                                                 allow_second_superset=True)
```
I attempt to compare the files directly first, but if that doesn't work, I transform them into a structured data sets that are stripped and filtered (as appropriate for the scenario - in this example, I removed any instance of paths since of course those would vary from system to system)

eslickj:
@jsiirola, you're seeing it now because I changed the way NL files are written.  The end of line used to be system dependent.  CRLF for Windows and LF for Linux.  I made the NL files always use LF to get around a bug in string function arguments when CRLF end of lines are used.  When you checkout on git, it converts line endings in text files to the one appropriate for you OS, so the baselines will have CRLF on Windows.  Now the NL files have LF even on Windows, and that's why there is this end-of-line issue now that wasn't there before.

eslickj:
Added test for external functions with a constant argument mixed with a variable argument.  Now I just need one for string arguments.

eslickj:
@jsiirola and @blnicho, I added a test function to contrib under ``ampl_function_demo`` with a build plugin copied from pynumero.  Luckily the cmake file already handles the ASL dependency.  There is also a test there.  It should complete this, except for some clean up.  The test function should be a fairly complete example of writing external ampl functions with strings and variable length argument lists.  As usual for me the comments may need some editing.

I tested ``pyomo build-extensions`` and running the test and at least on my machine it works.  Should be close anyway.

pyomo/contrib/ampl_function_demo/tests/test_ampl_function_demo.py
```diff
@@ -0,0 +1,34 @@
...
+from pyomo.opt import check_available_solvers
+
+flib = find_library("function_ASL")
+
+class TestAMPLExternalFunction(unittest.TestCase):
+    def test_eval_function(self):
+        if not flib:
+            self.skipTest("Could not find the function_ASL.dll library")
+        if platform.python_implementation().lower().startswith('pypy'):
+            self.skipTest("AMPLExternalFunction interface does not work on pypy")
```
blnicho, line 15: Why don't you need this same `skipTest` check for the other test in this file (`test_solve_function`)?

pyomo/core/tests/unit/test_external.py
```diff
@@ -65,33 +65,6 @@ def test_extra_kwargs(self):
...
-        out = StringIO()
-        m.pprint(ostream=out)
-        self.assertEqual(out.getvalue().strip(), """
-2 ExternalFunction Declarations
-    h : function=_g, units=None, arg_units=None
-    i : function=_h, units=kg, arg_units=['m', 's']
-
-2 Declarations: h i
-        """.strip())
-
```
blnicho, line 94: Why did you remove this test?

pyomo/environ/__init__.py
```diff
@@ -58,6 +58,7 @@ def _do_import(pkg_name):
     'pyomo.contrib.multistart',
     'pyomo.contrib.petsc',
     'pyomo.contrib.preprocessing',
+    'pyomo.contrib.ampl_function_demo',
```
blnicho, line 61: Could you move this to the top of the set to keep the contrib packages in alphabetical order?

pyomo/contrib/ampl_function_demo/tests/test_ampl_function_demo.py
eslickj: I think you can't evaluate the function from pypy, but you can send it to the solver without issue.

pyomo/core/tests/unit/test_external.py
eslickj: I didn't mean to.  I guess something may have gotten messed up merging the latest stuff in main.  I'll take a closer look.

jsiirola:
This capability got merged as part of #2295 (with a slightly different solution to the line endings problem).

PR #1907, opened 2021-04-03
Handle CBC infinite gap output

### LINKED ISSUE #1442: CBC plugin error when gap returned by solver is "1.#J"
Looks like the CBC solver returns `1.#J` if the gap is infinity/NaN, which isn't handled correctly by the `CBCplugin.py`, as that string cannot be converted to a float:

https://github.com/Pyomo/pyomo/blob/7a018c52591a9534e32f4a5d62b714f89da21536/pyomo/solvers/plugins/solvers/CBCplugin.py#L563

When this happens, we get a `ValueError: could not convert string to float: '1.#J'`

Also mentioned in this question: https://or.stackexchange.com/questions/3652/pyomo-cbc-1-j-error

Moisan:
## Summary/Motivation:

Loose copy-paste of my answer on [StackExchange about this issue](https://or.stackexchange.com/questions/3652/pyomo-cbc-1-j-error):

`1.#J` in C++ is the output obtained when printing a formatted double equal to infinity (`1.#INF`). 
If the lower bound found by CBC is equal to `0.0`, it will obtain a gap of `1.#INF` which will then be formatted to `1.#J`. The current CBC plugin of pyomo didn't handle that case and crashed when it tries to parse `1.#J` to a float.

- Return `float('inf')` when CBC returns an infinite gap.
- Add tests based on two outputs from CBC, one with a float gap and another with infinite gap. Both are based on test5 that has been stopped after two seconds.

### Legal Acknowledgement

### DIFF
--- a/pyomo/solvers/plugins/solvers/CBCplugin.py
+++ b/pyomo/solvers/plugins/solvers/CBCplugin.py
@@ -448,12 +448,12 @@ class CBCSHELL(SystemCallSolver):
             if n_tokens > 1:
                 # https://projects.coin-or.org/Cbc/browser/trunk/Cbc/src/CbcSolver.cpp?rev=2497#L3769
                 if n_tokens > 4 and tokens[:4] == ('Continuous', 'objective', 'value', 'is'):
-                    lower_bound = float(tokens[4])
+                    lower_bound = _float(tokens[4])
                 # Search completed - best objective %g, took %d iterations and %d nodes
                 elif n_tokens > 12 and tokens[1:3] == ('Search', 'completed') \
                         and tokens[4:6] == ('best', 'objective') and tokens[9] == 'iterations' \
                         and tokens[12] == 'nodes':
-                    optim_value = float(tokens[6][:-1])
+                    optim_value = _float(tokens[6][:-1])
                     results.solver.statistics.black_box.number_of_iterations = int(tokens[8])
                     nodes = int(tokens[11])
                 elif tokens[1] == 'Exiting' and n_tokens > 4:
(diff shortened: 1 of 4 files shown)


Moisan:
I see from some CI failure that `numpy` is not a hard requirement of pyomo. What would be the best way then to return infinity?

Moisan:
I updated the PR to use `float('inf')` instead of `np.inf`.

pyomo/solvers/plugins/solvers/CBCplugin.py
```diff
@@ -563,7 +563,7 @@ def process_logfile(self):
                 # https://projects.coin-or.org/Cbc/browser/trunk/Cbc/src/CbcSolver.cpp?rev=2497#L7918
                 elif tokens[0] == 'Gap:':
                     # This is relative and only to 2 decimal places - could calculate explicitly using lower bound
-                    gap = float(tokens[1])
+                    gap = float(tokens[1]) if tokens[1] != '1.#J' else float('inf')
```
jsiirola, line 566: Is this the only time that CBC can return `1.#J`?  I wonder if a better solution would be a standard function that was something like:
```python
def _float(x):
    if x == '1.#J':
        return float('inf')
    return float(x)
```
and then use that function in lieu of `float(x)` throughout the parser?

pyomo/solvers/tests/mip/test_mip.py
```diff
@@ -13,9 +13,34 @@
 
 import os
 from os.path import abspath, dirname
+
+from pyomo.solvers.plugins.solvers.CBCplugin import CBCSHELL
+
```
jsiirola, line 18: Instead of hijacking the `test_mip` infrastructure, can you add these tests to `pyomo/solvers/tests/checks/test_CBCplugin:TestCBCUsingMock`?

jsiirola (changes requested):
One question (does this need to be generalized), and one necessary change (please integrate these tests into the existing `test_CBCplugin.py` module).

pyomo/solvers/plugins/solvers/CBCplugin.py
Moisan: Good idea, I'll include that.

Moisan:
I updated the PR based on the comments and rebased on top of main.

jsiirola (approved):

blnicho (approved):

PR #1930, opened 2021-04-15
PyNumero PyomoNLP should expose nl file options

michaelbynum:
- PyNumero PyomoNLP should write NL file with skip_trivial_constraints=True

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/pynumero/interfaces/pyomo_nlp.py
+++ b/pyomo/contrib/pynumero/interfaces/pyomo_nlp.py
@@ -32,7 +32,7 @@ __all__ = ['PyomoNLP']
 
 # TODO: There are todos in the code below
 class PyomoNLP(AslNLP):
-    def __init__(self, pyomo_model):
+    def __init__(self, pyomo_model, nl_file_options=None):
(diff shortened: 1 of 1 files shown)


blnicho (commented):
@michaelbynum could you expand the PR description to include the motivation or issue behind this PR? It seems like this will change some default behavior in PyNumero and I'm a little surprised you didn't have to update any tests.

michaelbynum:
I should not have done this in such a hurry. What I really wanted was to be able to specify options to the nl writer.

michaelbynum:
With the most recent changes, the default behavior won't change. However, I should still answer your question. I think the "trivial constraints" (when all of the variables in a constraint are fixed) case is not a terribly common edge case. Also, I'm sure we don't have any tests with trivial constraints. I'm not sure those tests would even belong in PyNumero or not. They might be more appropriate for algorithms developed on top of PyNumero. I'm not sure.

jsiirola (approved):
Seems reasonable to me.

blnicho (approved):

PR #1946, opened 2021-04-23
reduce errors logged with value when exception is False

michaelbynum:
## Summary/Motivation:
When evaluating expressions with `value(expr, exception=False)`, errors still get logged when variables do not have values. This PR propagates the `exception` flag to the `_EvaluationVisitor` to prevent logging such errors.

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/expr/visitor.py
+++ b/pyomo/core/expr/visitor.py
@@ -929,6 +929,9 @@ def sizeof_expression(expr):
 
 class _EvaluationVisitor(ExpressionValueVisitor):
 
+    def __init__(self, exception):
+        self.exception = exception
+
     def visit(self, node, values):
         """ Visit nodes that have been expanded """
         return node._apply_operation(values)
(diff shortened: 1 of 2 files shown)


blnicho (commented):
Could you add a test for this new flag?

michaelbynum:
done

jsiirola:
I don't see a reason not to merge this, because it fixes a known problem.  

However, I am not enthusiastic about this change as we are now using `exception=False` as an alias for "`log errors = False`", which is confusing.  I think that long term I would actually like to move away from the `exception=` flag entirely.  Instead, I think that evaluation errors should *never* generate an exception but *always* be logged to `pyomo.core.expr` (or maybe `pyomo.core.expr.evaluation`?).  The operators should just return `None` or `float('inf')` or `float('nan')` as appropriate.  

This would remove the need for the exception flag.  For use cases where a user wants to suppress the errors, they can either set the log level on `pyomo.core.expr` to, e.g., `logging.CRITICAL` or use the `LoggingIntercept` to capture the errors and (conditionally) route/log/output them.

blnicho (approved):

PR #1949, opened 2021-04-26
improving interval arithmetic for some special cases

michaelbynum:
## Summary/Motivation:
The ultimate goal of this PR is for 
```
import pyomo.environ as pe
from pyomo.contrib.fbbt.fbbt import compute_bounds_on_expr

m = pe.ConcreteModel()
m.x = pe.Var()
m.y = pe.Var()
e = (m.x**2 + m.y**2)**(-0.5)
compute_bounds_on_expr(e)
```

to produce

```
(0, None)
```

instead of 

```
(None, None)
```

Both of these are valid, but the former is much more useful. In the end, this took a rework of the interval `div` and `inv` functions. 

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/fbbt/interval.py
+++ b/pyomo/contrib/fbbt/interval.py
@@ -44,7 +44,7 @@ def inv(xl, xu, feasibility_tol):
     but should be very slightly positive) should not be an issue. Suppose xu is 2 and xl is 1e-15 but should be -1e-15.
     The bounds obtained from this function will be [0.5, 1e15] or [0.5, inf), depending on the value of
     feasibility_tol. The true bounds are (-inf, -1e15] U [0.5, inf), where U is union. The exclusion of (-inf, -1e15]
-    should be acceptable. Additionally, it very important to return [0, 1/xu] when xl is 0.
+    should be acceptable. Additionally, it very important to return a non-negative interval when xl is non-negative.
     """
     if xu <= 0 <= xl:
         if abs(xl - xu) <= feasibility_tol:
--- a/pyomo/contrib/fbbt/interval.py
+++ b/pyomo/contrib/fbbt/interval.py
@@ -39,19 +39,43 @@ def mul(xl, xu, yl, yu):
 
 
 def inv(xl, xu, feasibility_tol):
-    if xl <= feasibility_tol and xu >= -feasibility_tol:
-        # if the denominator (x) can include 0, then 1/x is unbounded.
-        lb = -inf
+    """
+    The case where xl is very slightly positive but should be very slightly negative (or xu is very slightly negative
+    but should be very slightly positive) should not be an issue. Suppose xu is 2 and xl is 1e-15 but should be -1e-15.
+    The bounds obtained from this function will be [0.5, 1e15] or [0.5, inf), depending on the value of
+    feasibility_tol. The true bounds are (-inf, -1e15] U [0.5, inf), where U is union. The exclusion of (-inf, -1e15]
+    should be acceptable. Additionally, it very important to return [0, 1/xu] when xl is 0.
+    """
+    if xu <= 0 <= xl:
+        if abs(xl - xu) <= feasibility_tol:
+            raise IntervalException(f'Division by zero in inv; xl: {xl}; xu: {xu}')
+        else:
+            raise InfeasibleConstraintException(f'lower bound is greater than upper bound in inv; xl: {xl}; xu: {xu}')
+    elif 0 <= xl <= feasibility_tol:
+        # xu must be strictly positive
         ub = inf
-    else:
+        lb = 1.0 / xu
+    elif xl > feasibility_tol:
+        # xl and xu must be strictly positive
         ub = 1.0 / xl
         lb = 1.0 / xu
+    elif -feasibility_tol <= xu <= 0:
+        # xl must be strictly negative
+        lb = -inf
+        ub = 1.0 / xl
+    elif xu < -feasibility_tol:
+        # xl and xu must be strictly negative
+        ub = 1.0 / xl
+        lb = 1.0 / xu
+    else:
+        # everything else
+        lb = -inf
+        ub = inf
     return lb, ub
 
 
 def div(xl, xu, yl, yu, feasibility_tol):
-    if yl <= feasibility_tol and yu >= -feasibility_tol:
-        # if the denominator (y) can include 0, then x/y is unbounded.
+    if xl <= 0 <= xu and yl <= 0 <= yu:
         lb = -inf
         ub = inf
     else:
@@ -59,7 +83,7 @@ def div(xl, xu, yl, yu, feasibility_tol):
     return lb, ub
 
 
-def power(xl, xu, yl, yu):
+def power(xl, xu, yl, yu, feasibility_tol):
     """
     Compute bounds on x**y.
     """
@@ -67,7 +91,7 @@ def power(xl, xu, yl, yu):
         """
         If x is always positive, things are simple. We only need to worry about the sign of y.
         """
-        if yl < 0 and yu > 0:
+        if yl < 0 < yu:
             lb = min(xu ** yl, xl ** yu)
             ub = max(xl ** yl, xu ** yu)
         elif yl >= 0:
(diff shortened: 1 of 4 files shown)


pyomo/contrib/fbbt/interval.py
```diff
@@ -39,35 +39,59 @@ def mul(xl, xu, yl, yu):
...
+def power(xl, xu, yl, yu, feasibility_tol):
     """
     Compute bounds on x**y.
     """
     if xl > 0:
         """
         If x is always positive, things are simple. We only need to worry about the sign of y.
         """
-        if yl < 0 and yu > 0:
+        if yl < 0 < yu:
```
jsiirola, line 94: This case's else branch is unreachable.  Consider changing to:
```python
else: # yu <= 0:
    lb = min(xu**yl, xu**yu)
    ub = max(xl**yl, xl**yu)
```

pyomo/contrib/fbbt/interval.py
```diff
@@ -39,35 +39,59 @@ def mul(xl, xu, yl, yu):
...
+    """
+    The case where xl is very slightly positive but should be very slightly negative (or xu is very slightly negative
+    but should be very slightly positive) should not be an issue. Suppose xu is 2 and xl is 1e-15 but should be -1e-15.
+    The bounds obtained from this function will be [0.5, 1e15] or [0.5, inf), depending on the value of
+    feasibility_tol. The true bounds are (-inf, -1e15] U [0.5, inf), where U is union. The exclusion of (-inf, -1e15]
+    should be acceptable. Additionally, it very important to return [0, 1/xu] when xl is 0.
+    """
+    if xu <= 0 <= xl:
+        if abs(xl - xu) <= feasibility_tol:
+            raise IntervalException(f'Division by zero in inv; xl: {xl}; xu: {xu}')
```
jsiirola, line 51: Is this correct?  This logic seems a bit strange: You allow for division by zero for the intervals `[-eps, eps]`, `[eps, n]`, and `[-n, eps]`, but not `[eps, -eps]`?

pyomo/contrib/fbbt/interval.py
```diff
@@ -39,35 +39,59 @@ def mul(xl, xu, yl, yu):
...
+        else:
+            raise InfeasibleConstraintException(f'lower bound is greater than upper bound in inv; xl: {xl}; xu: {xu}')
```
jsiirola, line 53: Why do you only raise this exception if `xu < xl` *and* spans 0?  For example, the interval `[3, 2]` would not trigger this exception?

pyomo/contrib/fbbt/interval.py
```diff
@@ -39,35 +39,59 @@ def mul(xl, xu, yl, yu):
...
+        lb = -inf
+        ub = 1.0 / xl
+    elif xu < -feasibility_tol:
+        # xl and xu must be strictly negative
         ub = 1.0 / xl
         lb = 1.0 / xu
+    else:
+        # everything else
+        lb = -inf
+        ub = inf
```
jsiirola, line 73: Why does `[-eps, 2]` return `[-inf, inf]`, but `[eps, 2]` returns `[0.5, inf]`?

pyomo/contrib/fbbt/interval.py
```diff
@@ -82,9 +106,13 @@ def power(xl, xu, yl, yu):
...
             ub = max(xu ** yl, xu ** yu)
+        elif yu <= 0:
+            lb, ub = inv(*power(xl, xu, *sub(0, 0, yl, yu), feasibility_tol), feasibility_tol)
         else:
-            lb = -inf
-            ub = inf
+            lb1, ub1 = power(xl, xu, 0, yu, feasibility_tol)
+            lb2, ub2 = power(xl, xu, yl, 0, feasibility_tol)
+            lb = min(lb1, lb2)
+            ub = max(ub1, ub2)
```
jsiirola, line 115: Why is the next clause not updated to be relaxed by `feasibility_tol`?

pyomo/contrib/fbbt/interval.py
```diff
@@ -140,7 +168,7 @@ def power(xl, xu, yl, yu):
             msg += 'The upper bound of a variable raised to the power of {0} is {1}'.format(yl, xu)
             raise InfeasibleConstraintException(msg)
         xl = 0
-        lb, ub = power(xl, xu, yl, yu)
+        lb, ub = power(xl, xu, yl, yu, feasibility_tol)
```
jsiirola, line 171: Why is this clause (including `yl == yu` above) not relaxed by `feasibility_tol`?

jsiirola (commented):
There are several nuances here that I am not following.  I understand why you are propagating the `feasibility_tol` into these methods, but I don't think you are covering all the cases, and I thing there are more of these methods that should also be "relaxed" by the `feasibility_tol`.

pyomo/contrib/fbbt/interval.py
michaelbynum: I believe this behavior is reasonable, but it might be safer to return `(-inf, inf)` for this case.

michaelbynum: This one is tricky. I will come back to it. We definitely need to be careful here.

pyomo/contrib/fbbt/interval.py
michaelbynum: The interval `[3, 2]` would raise an exception, but only when performing FBBT. That is where most of the infeasible constraint detection happens.

pyomo/contrib/fbbt/interval.py
michaelbynum: This is just me being careful. If the lower bound is negative at all, I return `-inf` for the lower bound (1/-1e-12 is -1e12). If `xl >= -eps`, it could reasonably be assumed that the result should positive. However, it is safer to return `-inf` for the resulting lower bound.

pyomo/contrib/fbbt/interval.py
michaelbynum: Good point.

pyomo/contrib/fbbt/interval.py
michaelbynum: Can you be more specific? Which part do you think needs relaxed?

michaelbynum: Nevermind. I did not see your comment below.

pyomo/contrib/fbbt/interval.py
michaelbynum: The purpose of this clause is to catch constant exponents only (e.g., `m.x**1.5`). We do not want to catch things like `m.x**m.y` where `1 <= m.y <= 1+1e-4`. In that case, the `m.y` is not guaranteed to be fractional, so we cannot assert that m.x has to be positive. Relaxing `yl == yu` with `feasibility_tol` is actually not a relaxation.

pyomo/contrib/fbbt/interval.py
michaelbynum: fixed

pyomo/contrib/fbbt/interval.py
jsiirola: I see.  That makes sense.  Would this code then be more clear if it was written as 
```python
if xu - xl <= -feasibility_tol:
    raise InfeasibleConstraintException
elif xu - xl <= feasibility_tol and min(xl, xu) <= 0 <= max(xl, xu):
    raise IntervalException(f'Division by zero in inv')
```

michaelbynum: Yes. I like that. I'll make that change shortly.

michaelbynum: Actually, I think it has to be 

```
if xu - xl <= -feasibility_tol:
    raise InfeasibleConstraintException
elif xu <= 0 <= xl:
    raise IntervalException(f'Division by zero in inv')
```

because I want `inv(-1e-7, 1e-7, feasibility_tol=1e-6)` to return `(-inf, inf)` and not raise an exception.

michaelbynum: I updated this part of the code.

michaelbynum:
I also updated the FBBT tests with long runtimes. I was able to utilize `np.split` to remove some of the `for` loops. On my machine this brought the total runtime of all FBBT tests from 12 seconds to 2 seconds.

jsiirola (approved):

pyomo/contrib/fbbt/interval.py
```diff
@@ -39,52 +39,77 @@ def mul(xl, xu, yl, yu):
...
+        ub = inf
     return lb, ub
 
 
 def div(xl, xu, yl, yu, feasibility_tol):
-    if yl <= feasibility_tol and yu >= -feasibility_tol:
-        # if the denominator (y) can include 0, then x/y is unbounded.
+    if xl <= 0 <= xu and yl <= 0 <= yu:
         lb = -inf
         ub = inf
```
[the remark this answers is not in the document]
michaelbynum: No, this is still needed. This special case indicates that we don't know anything if `0/0` is possible.

PR #1953, opened 2021-04-27
Update merged reader test to reduce failures in OSX

jsiirola:
## Summary/Motivation:
We are still observing intermittent test failures for the merged reader test.  This PR is a small change that reduces dependence on (arbitrary) delays in the test code to make the test more deterministic.

- confirm that `stderr` has been output before closing the TeeStream object down (instead of just delaying for a short time)

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/tests/test_tee.py
+++ b/pyomo/common/tests/test_tee.py
@@ -46,6 +46,8 @@ class TestTeeStream(unittest.TestCase):
         # _peek_available is True)
         a = StringIO()
         b = StringIO()
+        # make sure this doesn't accidentally become a very long wait
+        assert tee._poll_interval <= 0.1
         with tee.TeeStream(a,b) as t:
             # This is a slightly nondeterministic (on Windows), so a
             # flush() and short pause should help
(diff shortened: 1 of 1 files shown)


pyomo/common/tests/test_tee.py
```diff
@@ -51,10 +51,17 @@ def test_merge_out_and_err(self):
...
             t.STDERR.write("interrupting\ncow")
             t.STDERR.flush()
-            time.sleep(tee._poll_interval*30)
+            # For determinism, it is important that the STDERR message
+            # appears in hte output stream before we start shutting down
+            # the TeeStream (which will dump the OUT and ERR in an
+            # arbitrary order)
+            i = 0
+            while i < 1000 and 'cow' not in a.getvalue():
+                time.sleep(tee._poll_interval)
```
michaelbynum, line 63: Can we assert that `_poll_interval` has not been changed to a large value before starting this loop? I'm not sure why anyone would make it significantly larger, but it still concerns me a bit.

michaelbynum (changes requested):

pyomo/common/tests/test_tee.py
jsiirola: Fixed.

michaelbynum (approved):

PR #1954, opened 2021-04-27
Only cast constant expressions to bool

### LINKED ISSUE #1839: Converting expressions to bool
I'm not sure what the correct behavior of converting expressions to bool is, but the following is at least inconsistent. I expected to get an error when creating a constraint at the end. In general, I would expect to get an exception when converting an expression to bool unless the expression is constant.

```

In [3]: m.x = pe.Var()

In [4]: m.x.value = 1

In [5]: bool(m.x > 0)
WARNING: DEPRECATED: Chained inequalities are deprecated. Use the inequality()
    function to express ranged inequality expressions. (called from
    /.../repos/pyomo/pyomo/core/expr/logical_expr.py:203)
Out[5]: True

In [6]: bool(m.x < 0)
TypeError                                 Traceback (most recent call last)
<ipython-input-6-d9e44a7644a3> in <module>
----> 1 bool(m.x < 0)

~/repos/pyomo/pyomo/core/expr/numvalue.py in __lt__(self, other)
    706             other > self
    707         """
--> 708         return _generate_relational_expression(_lt, self, other)
    709
    710     def __gt__(self,other):

~/repos/pyomo/pyomo/core/expr/logical_expr.py in _generate_relational_expression(etype, lhs, rhs)
    392                           "identical upper and lower\n\tbounds using strict "
    393                           "inequalities: constraint infeasible:\n\t%s and "
--> 394                           "%s < %s" % ( prevExpr.to_string(), lhs, rhs ))
    395                 if match[0] == (0,0):
    396                     # This is a particularly weird case where someone

TypeError: Cannot create a compound inequality with identical upper and lower
	bounds using strict inequalities: constraint infeasible:
	0.0  <  x and x < 0.0

In [7]: bool(m.x == 0)
Out[7]: False

In [8]: bool(m.x == 1)
Out[8]: True

In [9]: def _rule(m):
   ...:     if m.x == 1:
   ...:         return m.x**2 <= 1
   ...:

In [10]: m.c = pe.Constraint(rule=_rule)

In [12]: m.c.pprint()
c : Size=1, Index=None, Active=True
    Key  : Lower : Body : Upper : Active
    None :  -Inf : x**2 :   1.0 :   True
```

michaelbynum:
- Only allow constant expressions to be cast to bool

### Legal Acknowledgement


jsiirola (changes requested):
I am pretty impressed that you only had to make these changes.  

There are a few changes that I think are necessary (in particular, I think `SetOf` comparison should be updated to still work).

I would also like to consider a (significant) change.  Should we be able to get a boolean value back if the equality expression is structurally equal (that is, `if lhs is rhs`)?  That way things like `m.x == m.x` would return `True` and operations like list equality would still be valid.

michaelbynum:
I like the idea of returning `True` if `lhs is rhs`. That alone should fix the `SetOf` problem, right?

pyomo/core/expr/logical_expr.py
```diff
@@ -127,6 +87,16 @@ def __getstate__(self):
         return state
 
     def __nonzero__(self):
+        if not self.is_constant():
```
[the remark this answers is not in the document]
michaelbynum: Agreed. Good catch.

pyomo/core/tests/unit/test_visitor.py
```diff
@@ -99,40 +99,40 @@ def test_identify_vars_vars(self):
...
-        self.assertEqual( list(identify_variables(m.a)), [m.a] )
-        self.assertEqual( list(identify_variables(m.b[1])), [m.b[1]] )
-        self.assertEqual( list(identify_variables(m.a+m.b[1])),
-                          [ m.a, m.b[1] ] )
-        self.assertEqual( list(identify_variables(m.a**m.b[1])),
-                          [ m.a, m.b[1] ] )
-        self.assertEqual( list(identify_variables(m.a**m.b[1] + m.b[2])),
-                          [ m.a, m.b[1], m.b[2] ] )
-        self.assertEqual( list(identify_variables(
+        self.assertEqual( ComponentSet(identify_variables(m.a)), ComponentSet([m.a]) )
```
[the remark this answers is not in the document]
michaelbynum: That is also a good point.

pyomo/core/tests/transform/test_add_slacks.py
```diff
@@ -357,7 +357,7 @@ def test_transformed_constraints_sumexpression_body(self):
 
     def test_transformed_constraint_scalar_body(self):
         m = self.makeModel()
-        m.p = Param(initialize=6)
+        m.p = Param(initialize=6, mutable=True)
```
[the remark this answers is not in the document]
michaelbynum: Because the expression given to the constraint resolves to a `bool` if the `Param` is not mutable. It did not seem like it was important for this test to use an immutable `Param`, but maybe that was an incorrect conclusion.

michaelbynum:
@jsiirola Are you thinking `m.x == m.x` generates a `bool` and not an `EqualityExpression`? Then `m.x == m.y` would return an `EqualityExpression`. I think that makes sense, but inequalities are a bit weird. Also, would this actually enable list comparisons?

jsiirola:
@michaelbynum I am still thinking through this one.  I was thinking that `m.x == m.x` would generate an EqualityExpression.  Unfortunately, I don't think this alone will fix the `SetOf` behavior:  `SetOf` leverages `__contains__` from the underlying container.  For the case of `list`, `__contains__` uses something like:
```python
for x in data:
   if val == x:
      return True
```
then we still get exceptions for something like `m.x in SetOf([1, 2, m.x])` because of evaluating the `m.x == 1` test.

jsiirola:
IDAES (@andrewlee94) is starting to get ready for this PR and hit an unfortunate side effect:  there are currently tests that looks like:
```python
 assert model.params.config.state_bounds == {
    "flow_mol": (0, 100, 1000, pyunits.mol/pyunits.s),
    "temperature": (10, 300, 350, pyunits.K),
    "pressure": (5e4, 1e5, 1e7, pyunits.Pa)
}
```
The problem is that as Units are mutable Params, they are _not_ constant, and generate the new `PyomoException`.  We should consider generalizing the `pyomo.common.unittest.TestCase.assertStructuredAlmostEqual()` to make a general structured comparison function that takes a callback that determines how to handle non-iterable / non-mapping data.

pyomo/core/tests/unit/test_numeric_expr.py
```diff
@@ -48,92 +49,64 @@
 from pyomo.core.base.var import _GeneralVarData
 
 from pyomo.repn import generate_standard_repn
+from pyomo.core.expr.numvalue import NumericValue
+
+
+class decompose_linear_term_wrapper(object):
```
michaelbynum, line 55: These functions currently use the `value` function. We should replace the use of `value` in these functions to utilize a function that can robustly compare expressions. I will create an issue for this.

michaelbynum:
@jsiirola Assuming tests pass, I think this is ready.

pyomo/core/expr/logical_expr.py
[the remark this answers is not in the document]
blnicho: @michaelbynum looks like you made this change in other parts of this file but maybe missed the `__nonzero__` method on line 89?

pyomo/core/expr/logical_expr.py
```diff
@@ -127,6 +87,16 @@ def __getstate__(self):
...
+                                 'using an expression in a boolean context '
+                                 'such as an if statement. For example, \n'
+                                 '    m.x = Var()\n'
+                                 '    e = inequality(-1, m.x, 1)'
+                                 '    if e:\n'
+                                 '        ...\n'
+                                 'would cause this exception.')
         return bool(self())
 
     __bool__ = __nonzero__
```
blnicho, line 102: Given @jsiirola's previous review comment about `__nonzero__` being a Python 2 thing, is this still needed?

pyomo/core/expr/logical_expr.py
```diff
@@ -198,19 +168,16 @@ def __getstate__(self):
...
-            #return bool(self())                # This is needed to apply simple evaluation of inequalities
-
-        return bool(self())
-
-    __bool__ = __nonzero__
+    def __bool__(self):
+        raise PyomoException('Cannot convert non-constant expression '
+                             'to bool. This error usually arises for '
+                             'one of two reasons. One possibility is '
+                             'using an expression in a boolean context '
```
blnicho, line 175: Error message states that there are two common reasons why this exception might be raised but then only describes one.

pyomo/core/expr/logical_expr.py
michaelbynum: @blnicho Good catch! Thanks!

pyomo/core/expr/logical_expr.py
michaelbynum: Fixed. Thanks!

blnicho:
@michaelbynum did the comment from @jsiirola on "side-effects" in IDAES tests get resolved?

michaelbynum:
Good question. It probably still fails when comparing an expression of units. I have discussed with @jsiirola a mechanism for comparing expressions correctly, but I don't think either of us have implemented it yet.

andrewlee94:
From the IDAES perspective, we could break these test up into smaller ones where we test each part of the dict and tuple structure individually, using `assert_units_equivalent` on the unit strings. However, there is a bit of work involved in doing so, and I am not sure how many instances of this we would have to fix.

michaelbynum:
Sorry - I'm confused. Did idaes/idaes-pse#342 not resolve this? At least for the short term? Long term, we definitely need a way to compare expressions robustly.

andrewlee94:
@michaelbynum No, https://github.com/IDAES/idaes-pse/pull/342 only fixed instances where `value` was missing when checking results. The issue @jsiirola mentioned has not been addressed yet.

pyomo/core/tests/unit/test_visitor.py
```diff
@@ -106,7 +106,7 @@ def test_identify_vars_vars(self):
         self.assertEqual( list(identify_variables(m.a**m.b[1])),
                           [ m.a, m.b[1] ] )
         self.assertEqual( list(identify_variables(m.a**m.b[1] + m.b[2])),
-                          [ m.a, m.b[1], m.b[2] ] )
+                          [ m.b[2], m.a, m.b[1]] )
```
jsiirola, line 109: Why did this test baseline change?

jsiirola: from @michaelbynum:
> If you look closely, all of the variables have the same value… The original test was incorrect, and the bool(==) behavior hid it.
Great!  It is very good that these tests are more robustly testing the result.

jsiirola (approved):

blnicho (approved):

PR #1959, opened 2021-05-03
Do not cache the picklability of function types

jsiirola:
## Summary/Motivation:
Function types are picklable based on how they were declared (module-level functions are picklable, but others are not).  This PR prevents the config system pickling scheme from caching the "picklability" of functions - and resolves an intermittent issue where the pickle tests passed / failed based on whether or not other modules had been imported / run before hand.

This also adds special handling for domains that are class types (not class instances).  In this case, we cache the actual type (and NOT the `type` type) for the same reasons that we do not cache the picklability of the `FunctionType`.

- do not cache the picklability of function types in the config system
- when pickling classes, do not cache the picklability of the `type` class and instead cache the picklability of the actual class type in question
- add a test to verify stateless behavior for functions
- add tests for pickling classes

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/tests/test_config.py
+++ b/pyomo/common/tests/test_config.py
@@ -56,6 +56,11 @@ def _display(obj, *args):
     return test.getvalue()
 
 
+class GlobalClass(object):
+    "test class for test_known_types"
+    pass
+
+
 class TestConfigDomains(unittest.TestCase):
     def test_PositiveInt(self):
         c = ConfigDict()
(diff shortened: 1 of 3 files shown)


mrmundt (approved):
Looks good to me. Neat patch.

pyomo/common/tests/test_config.py
```diff
@@ -2245,6 +2251,66 @@ def cast(x):
...
+        except:
+            local_picklable = False
+        if local_picklable:
+            self.assertIs(_picklable(local_class, obj), local_class)
+            self.assertEqual(_picklable.known.get(local_class, None), True)
+        else:
+            self.assertIsNot(_picklable(local_class, obj), local_class)
+            self.assertEqual(_picklable.known.get(local_class, None), False)
+
+        # Ensure that none of the added `type` to the known dict
```
mrmundt, line 2311: This comment is phrased oddly, so I'm not actually sure what it means.

mrmundt (changes requested):

pyomo/common/tests/test_config.py
jsiirola: Fixed.

mrmundt (approved):

PR #1960, opened 2021-05-04
Convert `Disjunct.indicator_var` from binary to Boolean variable

### LINKED ISSUE #1868: Certain attributes are not inherited by deepcopy making transformations fail
This error appeared when working on a new project for transforming GDPs with some collaborators.
We are trying to iteratively apply transformations to GDPs with logical constraints. Since we don't want to edit the original model, we deepcopy the model itself before performing the transformation, but this triggers the following error
`AttributeError: 'ExactlyExpression' object has no attribute '_nargs'`
An MWE can be found below:
```
import copy
import pyomo.environ as pe

m = pe.ConcreteModel() # Original model
m.N = pe.RangeSet(1, 3)
m.YF = pe.BooleanVar(m.N)

def _logical_constraint(m):
    return pe.exactly(1,m.YF)
m.logical_constraint = pe.LogicalConstraint(rule=_logical_constraint)

m2 = copy.deepcopy(m) # Deep copied model

pe.TransformationFactory('core.logical_to_linear').apply_to(m) # Works well

pe.TransformationFactory('core.logical_to_linear').apply_to(m2) # Fails with error above
```
I think that this error is related to the inheritance of `_nargs` attributes, but I'm not sure. I haven't encountered this error elsewhere, but it seems plausible that it might show up for other components in Pyomo.

jsiirola:
## Summary/Motivation:
The original implementation of GDP predated the availability of a proper "logical" expression in Pyomo.  As such, the `Disjunct.indicator_var` was defined as a *binary* variable.  This is both inconsistent with the literature and problematic when defining logical constraints involving disjunct indicator variables.  This PR switches the `Disjunct.indicator_var` from a binary (`Var(domain=Binary)`) to a Boolean (`BooleanVar`).

To maintain backwards compatibility, this PR implements this switch using two new variable classes (but not new ctypes): `AutoLinkedBooleanVar` and `AutoLinkedBinaryVar`.  Instances of these two classes are intended to be used in pairs and linked together to represent a single state: changes to one variable (either the value or the fixed status) are automatically propagated over to the other.  `Disjunct.indicator_var` is an instance of `AutoLinkedBooleanVar`, and we add a new automatic component, `Disjunct.binary_indicator_var` as an instance of `AutoLinkedBinaryVar`.

The other motivation for using a class derived from `BooleanVar` (and not an actual `BooleanVar`) is that we can provide a deprecation path whereby the new (Boolean) `indicator_var` can be implicitly cast back to the associated binary variable (`binary_indicator_var`) when it is used in a numeric context.

Open questions not addressed by this PR:
- Should we leverage this idea of "auto linked" variables more broadly within the logical expression system (potentially to avoid the need for the `update_boolean_vars_from_binary()` function?
- Should we add a convenience methods so that `Disjunct` instances can be used directly in logical expressions (instead of having to explicitly reference the indicator var)?  This would be a very similar idea to how `AutoLinkedBooleanVar` supports use in numeric contexts.  Adding this would support more concise notation: `m.d[1].implies(m.d[2])` instead of (or more properly, in addition to) `m.d[1].indicator_var.implies(m.d[2].indicator_var)`

- switch `Disjunct.indicator_var` from a binary `Var` to a `BooleanVar`
- implement `AutoLinkedBooleanVar` and `AutoLinkedBinaryVar` to provide linked Binary/Boolean variable interface
- update transformations and contributed packages
- resolve a bug deepcopying n-ary logical expressions (#1868)
- update the definition of `LogicalConstraint.Skip` to reflect the use of `IndexedComponent.Skip` (following `Constraint`)
  - replace locally-declared tuples with types (for `Infeasible`)
- update documentation

### Legal Acknowledgement


doc/OnlineDocs/modeling_extensions/gdp/modeling.rst
```diff
@@ -203,16 +203,16 @@ Integration with Disjunctions
 
 .. note::
 
-    This section is subject to change between this date and August 2020.
-    The current plan is to provide a deprecation path changing ``indicator_var`` from binary ``Var`` to ``BooleanVar``.
+    Historically, the ``indicator_var`` on ``Disjunct`` objects was
+    implemented as a binary ``Var``.  Beginning in Pyomo 6.0, that has
+    been changed to the more mathematically correct ``BooleanVar``, with
+    the associated binary variable available as
+    ``binary_indicator_var``.
```
emma58, line 210: Can someone create a `Disjunct` and then link the `BooleanVar` to a different binary? This is the difference between `disjunct.binary_indicator_var` and `disjunct.indicator_var.get_associated_binary()` right? But with the current implementation `disjunct.binary_indicator_var` would still always be right, even in this case, correct?

doc/OnlineDocs/modeling_extensions/gdp/solving.rst
```diff
@@ -48,14 +56,15 @@ Following solution of the GDP model, values of the Boolean variables may be upda
 Reformulation to MI(N)LP
 ------------------------
 
-To use standard commercial solvers, you must convert the disjunctive model to a standard MIP/MINLP model.
-The two classical strategies for doing so are the (included) Big-M and Hull reformulations.
+To use standard commercial solvers, you must convert the disjunctive
+model to a standard MILP/MINLP model.  The two classical strategies for
```
emma58, line 60: Maybe this should be MI(N)LP also, for consistency

emma58 (commented):
This is so exciting!

I have one big question: Does one have to be careful of when to use `disjunct.binary_indicator_var` and `disjunct.indicator_var.get_associated_binary()`? Because a user *could* associate the BooleanVar with some other binary, right? And if that happened then `disjunct.binary_indicator_var` is going to be out of sync, correct? 

And regarding your questions: 

- Partly because of the train of thought above, I think that using auto-linked variables more broadly sounds kind of lovely. It seems like it would save a lot of bugs where values hadn't been updated from boolean to binary or vice versa.
- I'm hesitant to have those convenience methods... I know it's more typing without them, but if the whole goal of this is to do the mathematically correct thing, I don't think it's mathematically correct to equate a Disjunct and its indicator var. I worry we have people confused enough as it is.

pyomo/gdp/disjunct.py
```diff
@@ -44,6 +46,260 @@ class GDP_Error(PyomoException):
...
+        if val is None:
+            if bool_var.value is not None:
+                bool_var.set_value(None)
+        else:
+            # Only update the associated boolean value if it is needed
+            # to match the current (potentially fractional) binary value.
+            # (This prevents infinite recursion.)
+            bool_val = bool(int(val + 0.5))
+            if bool_val != bool_var.value:
+                bool_var.set_value(bool_val)
```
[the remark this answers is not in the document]
emma58: Whoops, I thought these went away. I like it better that it's none!

pyomo/gdp/disjunct.py
```diff
@@ -44,6 +46,260 @@ class GDP_Error(PyomoException):
...
+        return self.as_binary().ub
+
+    @ub.setter
+    def ub(self, value):
+        self.as_binary().ub = value
+
+    def __abs__(self):
+        return self.as_binary().__abs__()
+    def __bool__(self):
+        return self.as_binary().__bool__()
```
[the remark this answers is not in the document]
jsiirola: This is temporary and will likely be removed by #1954.

pyomo/gdp/disjunct.py
```diff
@@ -44,6 +46,260 @@ class GDP_Error(PyomoException):
...
+                bin_var.set_value(None)
+        else:
+            # Only update the associated (potentially fractional) binary
+            # value if it is needed to match the current Boolean value.
+            # (This prevents infinite recursion.)
+            bin_val = bin_var.value
+            if bin_val is None:
+                bin_val = 2 # something not in [0..1]
+            if int(bin_val + 0.5) != int(self.value):
+                bin_var.set_value(int(self.value))
```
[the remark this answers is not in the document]
jsiirola: This has been updated, to be more streamlined, but to answer our question, the binary value may be (slightly) fractional due to a solver's integer tolerance.  In this case, we are seeing if the bool value of `self` (the `AutoLinkedBooleanVar`) is inconsistent with the current (potentially fracational) value in the binary.  If they are inconsistent, then update the binary value.  To determine consistency, we have to map the (fractional) binary to the Boolean space before comparing.

pyomo/gdp/basic_step.py
```diff
@@ -77,7 +78,8 @@ def apply_basic_step(disjunctions_or_constraints):
             tmp = _pseudo_clone(disjunctions[i].disjuncts[
                 idx[i] if isinstance(idx, tuple) else idx])
             for k,v in list(tmp.component_map().items()):
-                if k == 'indicator_var':
+                if v.parent_block() is not tmp:
+                    # Skip indicator_var and binary_indicator_var
```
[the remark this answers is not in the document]
jsiirola: fixed.

pyomo/core/expr/numeric_expr.py
```diff
@@ -1572,17 +1572,30 @@ def _decompose_linear_terms(expr, multiplier=1):
 def _process_arg(obj):
     # Note: caller is responsible for filtering out native types and
     # expressions.
-    if obj.is_numeric_type() and obj.is_constant():
+    if not obj.is_numeric_type():
+        if hasattr(obj, 'as_binary'):
+            # We assume non-numeric types that have an as_binary method
+            # are instances of AutoLinkedBooleanVar.  Calling as_binary
+            # will return a valif Binar Var (and issue the appropriate
```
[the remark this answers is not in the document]
jsiirola: fixed.

pyomo/gdp/plugins/gdp_var_mover.py
```diff
@@ -137,7 +137,7 @@ def _disjunct_on_active_block(self, disjunct):
             if parent_block.ctype is Block and not parent_block.active:
                 return False
             elif (parent_block.ctype is Disjunct and not parent_block.active
-                  and parent_block.indicator_var.value == 0
+                  and parent_block.indicator_var.value == False
```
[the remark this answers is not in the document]
jsiirola: I don't think so.  I believe that this wanted to distinguish between `None` and `False`.

pyomo/gdp/util.py
```diff
@@ -391,7 +391,7 @@ def _disjunct_on_active_block(disjunct):
             return False
         # properly deactivated Disjunct
         elif (parent_block.ctype is Disjunct and not parent_block.active
-              and parent_block.indicator_var.value == 0
+              and parent_block.indicator_var.value == False
```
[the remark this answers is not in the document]
jsiirola: I don't think so.  I believe that this wanted to distinguish between `None` and `False`.

pyomo/gdp/tests/test_hull.py
```diff
@@ -1799,12 +1805,12 @@ def test_infeasible_xor_because_all_disjuncts_deactivated(self):
...
         # constraint says that the disaggregated indicator variables of the
         # nested disjuncts sum to the indicator variable of the outer disjunct.
-        ct.check_linear_coef( self, repn,
-                              m.disjunction.disjuncts[0].indicator_var, -1)
-        ct.check_linear_coef(self, repn,
-                             disjunct1.disaggregatedVars.indicator_var, 1)
-        ct.check_linear_coef(self, repn,
-                             disjunct1.disaggregatedVars.indicator_var_4, 1)
+        ct.check_linear_coef(
+            self, repn, m.disjunction.disjuncts[0].indicator_var, -1)
```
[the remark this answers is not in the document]
jsiirola: Not necessary: `check_linear_coef()` will extract the binary variable.

jsiirola:
@emma58:
> Does one have to be careful of when to use disjunct.binary_indicator_var and disjunct.indicator_var.get_associated_binary()?

I don't think it is a huge concern: `Disjunct` creates both the `indicator_var` and the `binary_indicator_var`, and pairs them.  The `get_associated_binary()` method is part of the standard (unlinked) `BooleanVar`, and is therefore still available here in the `AutoLinked` variant.  It should always be a synonym for `binary_indicator_var`.  I suppose a user *could* reassign the linked variable, but they would have to go out of their way to do so and they would be asking for trouble.

doc/OnlineDocs/modeling_extensions/gdp/solving.rst
```diff
@@ -79,8 +88,9 @@ The Hull Reformulation requires a lifting into a higher-dimensional space and co
 
     - All variables that appear in disjuncts need upper and lower bounds.
 
-    - The hull reformulation is an exact reformulation at the solution points
-      even for nonconvex GDP models, but the resulting MINLP will also be nonconvex.
+    - The hull reformulation is an exact reformulation at the solution
+      points even for nonconvex GDP models, but the resulting MINLP will
```
bernalde, line 92: This is debatable. I would not go into this rabbit hole. I would write:
The Hull reformulation is based on a high-dimensional representation of the disjunctive sets, which yields the convex hull of the disjunctions when they represent convex sets.

pyomo/core/expr/numeric_expr.py
```diff
@@ -1572,17 +1572,30 @@ def _decompose_linear_terms(expr, multiplier=1):
 def _process_arg(obj):
     # Note: caller is responsible for filtering out native types and
     # expressions.
-    if obj.is_numeric_type() and obj.is_constant():
+    if not obj.is_numeric_type():
+        if hasattr(obj, 'as_binary'):
+            # We assume non-numeric types that have an as_binary method
+            # are instances of AutoLinkedBooleanVar.  Calling as_binary
+            # will return a valid Binar Var (and issue the appropriate
```
bernalde, line 1579: Typo, Binary

pyomo/gdp/basic_step.py
```diff
@@ -19,11 +19,12 @@
 logger = logging.getLogger('pyomo.gdp')
 
 
-def _pseudo_clone(self):
-    """Clone everything in a Disjunct except for the indicator_var"""
+def _clone_all_but_indicator_vars(self):
+    """Clone everything in a Disjunct except for the indicator_vars"""
```
bernalde, line 23: Can a Disjunct have more than one indicator_var ?

pyomo/gdp/disjunct.py
```diff
@@ -44,6 +47,272 @@ class GDP_Error(PyomoException):
     """Exception raised while processing GDP Models"""
 
 
+class AutoLinkedBinaryVar(SimpleVar):
+    """A binary variable implicitly linked to its equivalent Boolean variable.
+
+    Basic oprerations like setting values and fixing/unfixing this
+    variable are also automatically applied to the assiciated binary
```
bernalde, line 54: Typo, Associated

pyomo/gdp/disjunct.py
```diff
@@ -44,6 +47,272 @@ class GDP_Error(PyomoException):
...
+    transformations.  However, with the introduction of a proper logical
+    expression system, the mathematically correct approach is for the
+    Disjunct's indicator_var attribute to be a proper BooleanVar.  As
+    part of the transition, indicator_var attributes are instances of
+    AutoLinkedBooleanVar, which allow the indicator_var to be used in
+    logical expressions, but also implicitly converted (with deprecation
+    warning) into their equivalent binary varriable.
+
+    Basic oprerations like setting values and fixing/unfixing this
+    variable are also automatically applied to the assiciated binary
```
bernalde, line 131: Typo, Associated

pyomo/gdp/disjunct.py
```diff
@@ -44,6 +47,272 @@ class GDP_Error(PyomoException):
...
+    variable.
+
+    As this class is only intended to provide a deprecation path for
+    Disjunct.indicator_var, it only supports Scalar instances and does
+    not support indexing.
+    """
+
+    INTEGER_TOLERANCE = 0.001
```
bernalde, line 62: Is this the right place for such an integer tolerance? If so, should GDPOpt use this same value?

bernalde (approved):
As @emma58 said, this is pretty exciting PR!
My comments are mostly typos and open questions that I had regarding the wording/implementation. This looks very good. Something that came up to my attention is that some of the MPEC tests were modified while no mode inside of MPEC was, maybe I'm missing that the transformation MPEC -> MINLP happens via GDP, but I just thought it was interesting.

pyomo/gdp/basic_step.py
jsiirola: No, but with this PR, it has `binary_indicator_var` and `indicator_var`, which is why I made the description plural.

pyomo/gdp/disjunct.py
jsiirola: @bernalde Good question.  I put it here mostly because it governs how these two classes interact (so, e.g., pulling in a tolerance from `contrib` didn't make sense).  That said, I picked the number arbitrarily (I think the though process was "1e-2 seems too big, and 1e-4 is probably too close to a solver's integer tolerance.  Let's go with 1e-3 and see if anyone complains.") and am open to changing it.

bernalde (approved):

pyomo/gdp/disjunct.py
```diff
@@ -44,6 +47,272 @@ class GDP_Error(PyomoException):
     """Exception raised while processing GDP Models"""
 
 
+class AutoLinkedBinaryVar(SimpleVar):
+    """A binary variable implicitly linked to its equivalent Boolean variable.
+
+    Basic oprerations like setting values and fixing/unfixing this
```
blnicho, line 53: Typo: operations

pyomo/gdp/disjunct.py
```diff
@@ -44,6 +47,272 @@ class GDP_Error(PyomoException):
...
+
+    This class provides a deprecation path for GDP.  Originally,
+    Disjunct indicator_var was a binary variable.  This simplified early
+    transformations.  However, with the introduction of a proper logical
+    expression system, the mathematically correct approach is for the
+    Disjunct's indicator_var attribute to be a proper BooleanVar.  As
+    part of the transition, indicator_var attributes are instances of
+    AutoLinkedBooleanVar, which allow the indicator_var to be used in
+    logical expressions, but also implicitly converted (with deprecation
+    warning) into their equivalent binary varriable.
```
blnicho, line 128: Typo: variable

pyomo/gdp/disjunct.py
```diff
@@ -44,6 +47,272 @@ class GDP_Error(PyomoException):
...
+
+    Basic oprerations like setting values and fixing/unfixing this
```
blnicho, line 130: Typo: operations

pyomo/gdp/disjunct.py
```diff
@@ -44,6 +47,272 @@ class GDP_Error(PyomoException):
     """Exception raised while processing GDP Models"""
 
 
+class AutoLinkedBinaryVar(SimpleVar):
+    """A binary variable implicitly linked to its equivalent Boolean variable.
+
+    Basic oprerations like setting values and fixing/unfixing this
+    variable are also automatically applied to the associated binary
```
blnicho, line 54: Should this be "applied to the associated **boolean**" variable?

jsiirola:
@blnicho: I fixed those typos

blnicho (approved):

blnicho:
Looks like the failing check was due to an intermittent connection issue and not related to the changes in this PR so I'm going to merge.

PR #1969, opened 2021-05-10
Fix GDP error when using a disjunction with nested disjunctions as a target

### LINKED ISSUE #1933: pyomo.gdp fails to transform nested Disjunctions when the parent Disjunction is specified as the target
## Summary

If you specify a Disjunction (SimpleDisjunction or DisjunctionData) as a target of a `pyomo.gdp` transformation, we never touch `_transform_blockData`, which is what we use to start transforming from the inside out. As a result, we panic when we encounter an active `Disjunct`, even though it is our fault for not first descending into the Disjuncts of the target...

```
from pyomo.environ import *
from pyomo.gdp import *

m = ConcreteModel()
m.x = Var(bounds=(0, 10))
m.y = Var(bounds=(0, 20))

m.disj1 = Disjunct()
m.disj1.disj1 = Disjunct()
m.disj1.disj1.c = Constraint(expr=m.x + m.y <= 18)
m.disj1.disj2 = Disjunct()
m.disj1.disj2.c = Constraint(expr=m.x <= 2)
m.disj1.disjunction = Disjunction(expr=[m.disj1.disj1, m.disj1.disj2])
m.disj2 = Disjunct()
m.disj2.c = Constraint(expr=m.y <= 16)
m.disjunction = Disjunction(expr=[m.disj1, m.disj2])

m.obj = Objective(expr=m.y, sense=maximize)

TransformationFactory('gdp.bigm').apply_to(m, targets=[m.disjunction])
```

### Error Message

```
  File "multipleReferencesToNestedIndVars.py", line 21, in <module>
    TransformationFactory('gdp.bigm').apply_to(m, targets=[m.disjunction])
  File "/home/esjohn/src/pyomo/pyomo/core/base/plugin.py", line 243, in apply_to
    self._apply_to(model, **kwds)
  File "/home/esjohn/src/pyomo/pyomo/gdp/plugins/bigm.py", line 208, in _apply_to
    self._apply_to_impl(instance, **kwds)
  File "/home/esjohn/src/pyomo/pyomo/gdp/plugins/bigm.py", line 249, in _apply_to_impl
    self._transform_disjunctionData( t, bigM, t.index())
  File "/home/esjohn/src/pyomo/pyomo/gdp/plugins/bigm.py", line 392, in _transform_disjunctionData
    suffix_list)
  File "/home/esjohn/src/pyomo/pyomo/gdp/plugins/bigm.py", line 463, in _transform_disjunct
    self._transform_block_components(obj, obj, bigM, arg_list, suffix_list)
  File "/home/esjohn/src/pyomo/pyomo/gdp/plugins/bigm.py", line 520, in _transform_block_components
    handler(obj, disjunct, bigM, arg_list, suffix_list)
  File "/home/esjohn/src/pyomo/pyomo/gdp/plugins/bigm.py", line 554, in _warn_for_active_disjunct
    _warn_for_active_disjunct(innerdisjunct, outerdisjunct, NAME_BUFFER)
  File "/home/esjohn/src/pyomo/pyomo/gdp/util.py", line 292, in _warn_for_active_disjunct
    name_buffer=NAME_BUFFER)))
pyomo.gdp.disjunct.GDP_Error: Found active disjunct 'disj1.disj1' in disjunct 'disj1'! Either disj1.disj1 is not in a disjunction or the disjunction it is in has not been transformed. disj1.disj1 needs to be deactivated or its disjunction transformed before disj1 can be transformed.
```

Pyomo version: main branch
Python version: 3.6.5
Operating system: linux
Solver (if applicable):

emma58:
## Summary/Motivation:

GDP transformations usually get lucky because they transform objects in declaration order, and Disjuncts have to be declared before the Disjunctions they appear in. However, you can mess up this order in the `targets` list. In particular, in the referenced issue, you can give a Disjunction which contains nested Disjunctions as a target. This bypasses the depth-first-search transformation order, so GDP fails to transform the nested stuff and then blames you.

- Preprocess targets so that Disjuncts always appear before the Disjunctions that contain them in the list.
- Remove tests for the active Disjunction error, as that is now unreachable, I think
- Be more careful about the argument to the transformation: must be Block-like so that we are sure we catch this in targets

### Legal Acknowledgement

### DIFF
--- a/pyomo/gdp/plugins/bigm.py
+++ b/pyomo/gdp/plugins/bigm.py
@@ -247,16 +247,13 @@ class BigM_Transformation(Transformation):
                     if t.is_indexed():
                         for disjunction in t.values():
                             for disj in disjunction.disjuncts:
-                                print("adding Disjunct %s" % disj.name)
                                 preprocessed_targets.append(disj)
                     else:
                         for disj in t.disjuncts:
-                            print("adding Disjunct %s" % disj.name)
                             preprocessed_targets.append(disj)
                 # now we are safe to put the disjunction, and if the target was
                 # anything else, then we don't need to worry because disjuncts
                 # are declared before disjunctions they appear in
-                print("adding target %s of type %s" % (t.name, t.ctype))
                 preprocessed_targets.append(t)
             targets = preprocessed_targets
 
--- a/pyomo/gdp/plugins/bigm.py
+++ b/pyomo/gdp/plugins/bigm.py
@@ -213,6 +213,12 @@ class BigM_Transformation(Transformation):
             self.used_args.clear()
 
     def _apply_to_impl(self, instance, **kwds):
+        if not instance.ctype in (Block, Disjunct):
+            raise GDP_Error("Transformation called on %s of type %s. 'instance' "
+                            "must be a ConcreteModel, Block, or Disjunct (in "
+                            "the case of nested disjunctions)." %
+                            (instance.name, instance.ctype))
+
         config = self.CONFIG(kwds.pop('options', {}))
 
         # We will let args override suffixes and estimate as a last
--- a/pyomo/gdp/plugins/bigm.py
+++ b/pyomo/gdp/plugins/bigm.py
@@ -31,9 +31,8 @@ from pyomo.gdp.util import ( _warn_for_active_logical_constraint, target_list,
                              get_src_constraint, get_transformed_constraints,
                              _get_constraint_transBlock, get_src_disjunct,
                              _warn_for_active_disjunction,
-                             _warn_for_active_disjunct, )
+                             _warn_for_active_disjunct, preprocess_targets)
 from pyomo.repn import generate_standard_repn
-
 from functools import wraps
 from weakref import ref as weakref_ref
 
@@ -241,21 +240,7 @@ class BigM_Transformation(Transformation):
             # we need to preprocess targets to make sure that if there are any
             # disjunctions in targets that their disjuncts appear before them in
             # the list.
-            preprocessed_targets = []
-            for t in targets:
-                if t.ctype is Disjunction:
-                    if t.is_indexed():
-                        for disjunction in t.values():
-                            for disj in disjunction.disjuncts:
-                                preprocessed_targets.append(disj)
-                    else:
-                        for disj in t.disjuncts:
-                            preprocessed_targets.append(disj)
-                # now we are safe to put the disjunction, and if the target was
-                # anything else, then we don't need to worry because disjuncts
-                # are declared before disjunctions they appear in
-                preprocessed_targets.append(t)
-            targets = preprocessed_targets
+            targets = preprocess_targets(targets)
 
         #  We need to check that all the targets are in fact on
         # instance. As we do this, we will use the set below to cache components
--- a/pyomo/gdp/plugins/bigm.py
+++ b/pyomo/gdp/plugins/bigm.py
@@ -231,9 +231,32 @@ class BigM_Transformation(Transformation):
         targets = config.targets
         if targets is None:
             targets = (instance, )
-        # We need to check that all the targets are in fact on instance. As we
-        # do this, we will use the set below to cache components we know to be
-        # in the tree rooted at instance.
+        else:
+            # we need to preprocess targets to make sure that if there are any
+            # disjunctions in targets that their disjuncts appear before them in
+            # the list.
+            preprocessed_targets = []
+            for t in targets:
+                if t.ctype is Disjunction:
+                    if t.is_indexed():
+                        for disjunction in t.values():
+                            for disj in disjunction.disjuncts:
+                                print("adding Disjunct %s" % disj.name)
+                                preprocessed_targets.append(disj)
+                    else:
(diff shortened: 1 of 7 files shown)


pyomo/gdp/plugins/bigm.py
```diff
@@ -231,9 +237,32 @@ def _apply_to_impl(self, instance, **kwds):
...
+            # we need to preprocess targets to make sure that if there are any
+            # disjunctions in targets that their disjuncts appear before them in
+            # the list.
+            preprocessed_targets = []
+            for t in targets:
+                if t.ctype is Disjunction:
+                    if t.is_indexed():
+                        for disjunction in t.values():
+                            for disj in disjunction.disjuncts:
+                                print("adding Disjunct %s" % disj.name)
```
jsiirola, line 250: Is this debugging? If this is expected output, please convert to using the logger.

pyomo/gdp/plugins/bigm.py
```diff
@@ -231,9 +237,32 @@ def _apply_to_impl(self, instance, **kwds):
...
+                                preprocessed_targets.append(disj)
+                    else:
+                        for disj in t.disjuncts:
+                            print("adding Disjunct %s" % disj.name)
```
jsiirola, line 254: Is this debugging? If this is expected output, please convert to using the logger.

pyomo/gdp/plugins/bigm.py
```diff
@@ -231,9 +237,32 @@ def _apply_to_impl(self, instance, **kwds):
...
+                            preprocessed_targets.append(disj)
+                # now we are safe to put the disjunction, and if the target was
+                # anything else, then we don't need to worry because disjuncts
+                # are declared before disjunctions they appear in
+                print("adding target %s of type %s" % (t.name, t.ctype))
```
jsiirola, line 259: Is this debugging? If this is expected output, please convert to using the logger.

pyomo/gdp/plugins/hull.py
```diff
@@ -230,13 +230,41 @@ def _apply_to(self, instance, **kwds):
...
+                    else:
+                        for disj in t.disjuncts:
+                            print("adding Disjunct %s" % disj.name)
+                            preprocessed_targets.append(disj)
+                # now we are safe to put the disjunction, and if the target was
+                # anything else, then we don't need to worry because disjuncts
+                # are declared before disjunctions they appear in
+                print("adding target %s of type %s" % (t.name, t.ctype))
+                preprocessed_targets.append(t)
+            targets = preprocessed_targets
```
jsiirola, line 267: This seems identical to the implementation in bigm.  Should we create a utility method to prevent repeating code?

jsiirola (changes requested):
In addition to the specific comments, one general one:  what happens if given:
```python
m = ConcreteModel()
m.d1 = Disjunct()
m.d2 = Disjunct()
```
the user specifies something like:
```python
TransformationFactory('gdp.bigm').apply_to(m, targets=[m.d1, m.disj])
```
or
```python
TransformationFactory('gdp.bigm').apply_to(m, targets=[m, m.disj])
```
(that is, the same block appears "twice" in the targets list)?

pyomo/gdp/plugins/hull.py
```diff
@@ -230,13 +230,41 @@ def _apply_to(self, instance, **kwds):
...
+                                print("adding Disjunct %s" % disj.name)
+                                preprocessed_targets.append(disj)
+                    else:
+                        for disj in t.disjuncts:
+                            print("adding Disjunct %s" % disj.name)
+                            preprocessed_targets.append(disj)
+                # now we are safe to put the disjunction, and if the target was
+                # anything else, then we don't need to worry because disjuncts
+                # are declared before disjunctions they appear in
+                print("adding target %s of type %s" % (t.name, t.ctype))
```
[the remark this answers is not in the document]
emma58: Ooof, yes, it's debugging. Sorry, I apparently forgot to take *all* of them out...

pyomo/gdp/plugins/hull.py
emma58: Yes... This is not the only place we have this problem, but I may as well not add to it.

pyomo/gdp/plugins/bigm.py
```diff
@@ -213,6 +213,12 @@ def _apply_to(self, instance, **kwds):
             self.used_args.clear()
 
     def _apply_to_impl(self, instance, **kwds):
+        if not instance.ctype in (Block, Disjunct):
```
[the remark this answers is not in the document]
emma58: Perhaps? I'm not sure either. But I guess there is no reason we shouldn't go along with something's Blockiness. I think that means line 261 should change also, yes?

emma58:
@jsiirola regarding your question, it's fine if targets appear multiple times in the list because we just skip over already-transformed components (because there's nothing to stop me from transforming an inner disjunction with bigm and then the outer with hull or something like that). I added tests to that effect though... 

I addressed everything except your question about accepting all subclasses of Block, which I'm still thinking about. I *think* we should since it might save us from some of the incompatibilities with other packages. And it seems like the expected behavior, right?

jsiirola (approved):

PR #2001, opened 2021-06-04
Improve management of stdout in cyipopt solver

jsiirola:
## Summary/Motivation:
Cyipopt sends its output directly to file descriptor 1 (and not `sys.stdout`).  This forced users to resort to tools like `wurlitzer` to get proper IO redirection for cyipopt with `tee=True`.  This PR refines how the cyipopt solver interface manages output redirection, including cases where `sys.stdout` does not have a proper `fileno()`.

As part of this PR, the original file descriptor redirection logic in the cyipopt solver interface has been abstracted and refined into the `pyomo.common.tee.redirect_fd` context manager.  In the future we can use this with other C-bindings (like MC++) to prevent the unconditional output of those tools to the stdout / stderr file descriptors.

- Cyipopt solver: redirect the stdout file descriptor (fd=1) to `sys.stdout` when `tee=True` and to `os.devnull` when `tee=False`
- Implement a general `redirect_fd` context manager in `pyomo.common.tee`
- add a `capture_fd` boolean argument to `capture_output` to also capture output to the raw file descriptors
- remove the (deprecated) use of `import ipopt` to test for cyipopt availability
- remove the (deprecated) use of `setProblemScaling` and `addOption`
- Add tests

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/tee.py
+++ b/pyomo/common/tee.py
@@ -39,19 +39,102 @@ except ImportError:
 logger = logging.getLogger(__name__)
 
 
+class redirect_fd(object):
+    def __init__(self, fd=1, output=None, synchronize=True, enable=True):
+        if output is None:
+            # /dev/null is used just to discard what is being printed
+            output = os.devnull
+        self.fd = fd
+        self.std = {1: 'stdout', 2: 'stderr'}.get(self.fd, None)
+        self.target = output
+        self.target_file = None
+        self.synchronize = synchronize
+        self.enable = enable
+        self.original_file = None
+        self.original_fd = None
+
+    def __enter__(self):
+        if not self.enable:
+            return self
+        if self.std:
+            # important: flush the current file buffer when redirecting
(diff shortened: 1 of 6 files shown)


pyomo/common/tee.py
```diff
@@ -39,19 +39,121 @@
...
 
 
+class redirect_fd(object):
+    """Redirect a file descriptor to a new file or file descriptor.
+
+    This context manager will redirect the specified file descriptor to
+    a specified new output target (either file name or file descriptor).
+    For the special case of file descriptors 1 (stdout) and 2 (stderr),
+    we will also make sure that the Python `sys.stdout` or `sys.stderr`
+    reamin usable: in the case of synchronize=True, the `sys.stdout` /
```
blnicho, line 49: Typo: remain

blnicho (approved):
I found one typo but otherwise this looks fine.

PR #2005, opened 2021-06-04
gdp.hull: Only create one disaggregated variable for all Disjuncts a Var doesn't appear in

emma58:
## Summary/Motivation:

For a given disjunction and a variable which appears in at least one of its disjuncts, we need only create disaggregated copies of that variable for each disjunct it appears in plus one more. This is because, in disjuncts it doesn't appear in, it is free if any one of them is active and 0 otherwise. Implementing hull this way slightly reduces the number of added variables and constraints.

- Changes hull to only create disaggregated variables for the Vars it finds in the Disjuncts. Then, after that pass, if it didn't create number-of-disjuncts-in-the-disjunction Vars, it creates exactly one more.
- Moves the internal mappings for disaggregated Vars to their originals to the disjunct transformation blocks, which doesn't effect the external api.
- Updates tests... Note that FME tests changed because constraints changed order, and one was replaced by an equivalent one (alternative representation substituting from XOR equality). Note also that a lot of tests on two-term disjunctions changed because, if a Var doesn't appear in both Disjuncts, the copy for the one it doesn't appear in is now declared on the disjunction transformation block.

### Legal Acknowledgement


pyomo/gdp/plugins/hull.py
```diff
@@ -462,35 +489,86 @@ def _transform_disjunctionData(self, obj, index, transBlock=None):
...
+                    # Because we called _transform_disjunct above, we know that
+                    # if this isn't transformed it is because it was cleanly
+                    # deactivated, and we can just skip it.
+                    if disj._transformation_block is not None and \
+                       disj not in disjunctsVarAppearsIn[var]:
+                        relaxationBlock = disj._transformation_block().\
+                                          parent_block()
+                        relaxationBlock._bigMConstraintMap[
+                            disaggregated_var] = Reference(
+                                disaggregated_var_bounds[idx,...])
```
jsiirola, line 553: This should use "`:`" and not "`...`", correct?

pyomo/gdp/plugins/hull.py
```diff
@@ -462,35 +489,86 @@ def _transform_disjunctionData(self, obj, index, transBlock=None):
...
+                else:
+                    disaggregated_var.setub(ub)
+
+                # and we need the bounds constraints too
+                var_free = 1 - sum(disj.indicator_var.get_associated_binary()
+                                   for disj in disjunctsVarAppearsIn[var])
+                disaggregated_var_bounds[idx, 'lb'] = lb*var_free <= \
+                                                      disaggregated_var
+                disaggregated_var_bounds[idx, 'ub'] = disaggregated_var <= \
+                                                      ub*var_free
```
jsiirola, line 540: This is basically repeated elsewhere.  Can we combine this into a single helper function?

jsiirola (approved):
This looks good.  If you have a minute, it would be good to consolidate the repeated code.

pyomo/gdp/plugins/hull.py
jsiirola: (fixed)

pyomo/gdp/plugins/hull.py
emma58: OK this is done. It was a bigger change because it caught a couple bugs: 

1. Lines 535-540 above are wrong because the bounds constraints should only get added (respectively) if lb and ub are not `None` and nonzero. (That's what changed the hull test and the FME tests.)
2. In the check to see if a Disjunction was nested (if we should declare its disaggregated variables local), I was checking the parent Block of the Disjunct, not the Disjunction. (This is the wrong check, see the `makeNestedDisjunctions_FlatDisjuncts` model.) I moved this into a helper function as well, as it is needed for both the normal disaggregated variables and the ones representing multiple Disjuncts.

This broke a create_using test though, so we need to fix that before we can merge.

jsiirola:
PR #2024 will change how `pprint()` will format output so that the Suffix `pprint()` in the currently failing test will output component names instead of component `repr()`.  That is,

```
"[x, y[1]]"
```
instead of 
```
[<pyomo.core.base.var.ScalarVar object at 0x7fc44f3837d0>, <pyomo.core.base.var._GeneralVarData object at 0x7fc44f388520>]
```

pyomo/gdp/plugins/hull.py
```diff
@@ -714,7 +726,8 @@ def _transform_block_components( self, block, disjunct, var_substitute_map,
         # transformation blocks for nested disjunctions, so that we don't have
         # duplicate references.
         varRefBlock = disjunctBlock.localVarReferences
-        for v in block.component_objects(Var, descend_into=Block, active=None):
+        for v in block.component_data_objects(Var, descend_into=Block,
+                                              active=None):
             varRefBlock.add_component(unique_component_name(
                 varRefBlock, v.getname(fully_qualified=True,
                                        name_buffer=NAME_BUFFER)), Reference(v))
```
emma58, line 733: @jsiirola, this actually introduces a bug because `Reference(v)` fails when v is a `_GeneralVarData`, so it dies if you have an indexed variable declared on the Disjunct. But the reason I made the change was because without it, with the other changes, some transformed models don't pprint: They die in the `localVarReferences' block... I'm confused.

emma58: OK, it's hitting #2039. I think we can just check that we are actually making a Reference to *something* and skip it if not, since we know we aren't going to add to the containers after we make the Reference.

PR #2009, opened 2021-06-09
Methods for generating and solving strongly connected components of a square model

Robbybp:
## Summary/Motivation:
Strongly connected components, or diagonal blocks in a block triangularization, of a square model's incidence graph/matrix, are square subsets of equations and variables that may be solved independently to yield a feasible solution (with respect to equality constraints, at least). Generating and/or solving these subsets is useful for model initialization when (a) we know what inputs can/should be fixed to yield a nonsingular square model and (b) fixing these inputs yields a "decomposable" model. This is "decomposable" in the sense that the incidence graph is composed of more than one strongly connected component, which is not guaranteed, and in the opposite case these methods are useless.

An example use case:
```python
to_fix = input_vars
with TemporarySubsystemManager(to_fix=to_fix):
    solve_strongly_connected_components(model, solver=solver)
```

If you want more control over how the solve is performed, you can do something like:
```python
to_fix=input_vars
with TemporarySubsystemManager(to_fix=to_fix):
    for block in generate_strongly_connected_components(model):
        # Code to initialize block goes here!
        solver.solve(block)
```

- `generate_subsystem_blocks` generator in util/subsystems.py
- incidence_analysis/util.py module, which contains `generate_strongly_connected_components` and `solve_strongly_connected_components` generator/function

Note that this PR relies on #1998. I can combine these into one PR if that is convenient.

### Legal Acknowledgement


pyomo/contrib/incidence_analysis/util.py
```diff
@@ -0,0 +1,98 @@
...
+        ):
+    """ Performs a block triangularization of the variable-constraint
+    incidence matrix of the provided block, and yields a block that
+    contains the variables and constraints of each diagonal block
+    (strongly connected component).
+    """
+    variables = [var for var in block.component_data_objects(Var)
+            if not var.fixed]
+    constraints = [con for con in block.component_data_objects(Constraint)
+            if con.active]
```
jsiirola, line 35: This can potentially gather more constraints than you want: it will return an active constraint on a deactivated subblock.  To only return the active constraints (that the writers would output), use `block.component_data_objects(Constraint, active=True)`

pyomo/contrib/incidence_analysis/util.py
```diff
@@ -0,0 +1,98 @@
...
+def generate_strongly_connected_components(
+        block,
+        include_fixed=False,
+        fix_inputs=True,
+        ):
+    """ Performs a block triangularization of the variable-constraint
+    incidence matrix of the provided block, and yields a block that
+    contains the variables and constraints of each diagonal block
+    (strongly connected component).
+    """
```
jsiirola, line 31: Please document the `include_fixed` and `fix_inputs` parameters

pyomo/util/subsystems.py
```diff
@@ -0,0 +1,195 @@
...
+#  This software is distributed under the 3-clause BSD License.
+#  ___________________________________________________________________________
+
+from pyomo.core.base.PyomoModel import ConcreteModel
+from pyomo.core.base.reference import Reference
+from pyomo.core.expr.visitor import identify_variables
+from pyomo.common.collections import ComponentSet, ComponentMap
+
+
+def create_subsystem_block(constraints, variables=None, include_fixed=False):
```
jsiirola, line 17: Out of curiosity, what is the value of the `include_fixed` parameter?

pyomo/util/subsystems.py
```diff
@@ -0,0 +1,195 @@
...
+    """
+    for cons, vars in subsystems:
+        block = create_subsystem_block(cons, vars, include_fixed)
+        if fix_inputs:
+            to_fix = list(block.input_vars[:])
+            with TemporarySubsystemManager(to_fix=to_fix):
+                # On enter, we fix the "inputs" of the subsystem. When
+                # control returns to this code (iterator advance), we
+                # exit this context, and inputs are unfixed.
+                yield block
```
jsiirola, line 67: I like this: it looks pretty clean.  However, is this guaranteed to unfix the block if the iterator is not exhausted?  For example, if an exception is raised by a solver while iterating over `generate_subsystem_blocks`, is the last block solved unfixed?

pyomo/util/subsystems.py
```diff
@@ -0,0 +1,195 @@
...
+            yield block
+
+
+class TemporarySubsystemManager(object):
+    """ This class is a context manager for cases when we want to
+    temporarily fix or deactivate certain variables or constraints
+    in order to perform some solve or calculation with the resulting
+    subsystem.
+
+    """
```
jsiirola, line 78: Please document the `to_*` arguments

pyomo/util/subsystems.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+        Parameters
+        ----------
+        n_scenario: The number of different values we expect for each
+                    input variable
+        input_values: ComponentMap mapping each input variable to a list
+                      of values of length n_scenario
+        output_values: ComponentMap mapping each output variable to a list
+                       of values of length n_scenario
+        """
```
jsiirola, line 146: Plese document the `to_*` arguments

pyomo/util/subsystems.py
```diff
@@ -0,0 +1,195 @@
...
+            comp.set_value(val)
+
+
+class ParamSweeper(TemporarySubsystemManager):
+    """ This class enables setting values of variables/parameters
+    according to a provided sequence. Iterating over this object
+    sets values to the next in the sequence, at which point a
+    calculation may be performed and output values compared.
+    On exit, original values are restored.
+    """
```
jsiirola, line 127: I am a bit confused on the use of this class:
- It is named "ParamSweeper", but it seems to be focused on Vars (and not Params - can it be used with Params, too?).
- What is the distinction between "input" and "output"?  It seems like they are all just "values"

I think the documentation needs an example: I had to go to the tests to try and understand the intended functionality.

jsiirola (changes requested):
This is great!  A few questions / minor changes, and I think this can be merged.

pyomo/contrib/incidence_analysis/tests/test_util.py
```diff
@@ -0,0 +1,443 @@
...
+    discretizer = pyo.TransformationFactory("dae.finite_difference")
+    discretizer.apply_to(m, **default_disc_args)
+
+    return m
+
+
+class TestGenerateSCC(unittest.TestCase):
+
+    @unittest.skipUnless(scipy_available, "SciPy is not available")
+    @unittest.skipUnless(networkx_available, "NetworkX is not available")
```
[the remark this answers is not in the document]
Robbybp: Noted

pyomo/contrib/incidence_analysis/util.py
Robbybp: Changed

pyomo/contrib/incidence_analysis/util.py
Robbybp: Done

pyomo/util/subsystems.py
Robbybp: I believe so. `__exit__` is called before the program stops or the exception is handled. I have added a test to verify this behavior.

pyomo/util/subsystems.py
Robbybp: If a fixed variable may be unfixed at some point, then it needs to be attached to the subsystem block in order for that block to be a "valid model" according to the nl writer. We could want to solve this block after inputs are unfixed, perform a Dulmage-Mendelsohn decomposition, etc. I don't have a particular use case in mind, but it is conceivably possible that this would be useful/necessary at some point.

pyomo/util/subsystems.py
Robbybp: Done

pyomo/util/subsystems.py
Robbybp: Done

pyomo/util/subsystems.py
Robbybp: - Yes, mutable parameters may be included in `input_values`. It has additional functionality to handle the case when variables are used as "parameters" or "inputs" into a block of equations.
- The distinction is that inputs are set every time this iterator is advanced. Outputs are presumably solved for (by the user) in this object's context manager. Their expected values are provided by the user in a convenient data structure (component map from vars to lists), who can then write one loop to check actual vs. expected values using the data structure returned by the iterator (which only contains the values for the current scenario and is therefore hopefully more convenient).

Hopefully that did not make this too much more confusing. I also have added an example to the docstring.

pyomo/util/subsystems.py
Robbybp: There appears to be a pypy-only failure of exactly this test. Will try to debug tomorrow.

Robbybp: It appears that having the context manager inside of the generator function leads to implementation-specific behavior if an exception is raised. Since this can lead to an implementation-dependent state of the user's model as they process the exception, I removed the embedded `TemporarySubsystemManager` from the generator. If the user wants to fix the inputs to a subsystem block, they can add the context manager themselves.

pyomo/contrib/incidence_analysis/util.py
```diff
@@ -22,17 +22,33 @@
...
+    Arguments
+    ---------
+    block: Block whose strongly connected components will be generated
+    include_fixed: Bool indicating whether fixed variables will be
+                   attached as "input variables" on the subsystem blocks
+                   containing the strongly connected components
+    fix_inputs: Bool indicating whether "input variables" on a subsystem
+                block should be temporarily fixed when the block is
+                yielded. The default is True as this is necessary to solve
+                the strongly connected component as a square system.
```
jsiirola, line 40: This argument has been removed.

pyomo/util/subsystems.py
```diff
@@ -8,23 +8,38 @@
...
     parameters in the subsystem, but we do not fix them here as it is not
     obvious that this is desired.
+
+    Arguments
+    ---------
+    constraints: List of Pyomo constraint data objects
+    variables: List of Pyomo var data objects
+    include_fixed: Bool indicating whether fixed variables should be
+                   attached to the block. This is useful if they may
+                   be unfixed at some point.
```
jsiirola, line 32: Can you reformat this to follow numpydoc format?

(i.e.:
```
arg: type
    description
```
)

pyomo/util/subsystems.py
```diff
@@ -78,6 +83,21 @@ class TemporarySubsystemManager(object):
...
+        to_fix: List of var data objects that should be temporarily fixed.
+                These are restored to their original status on exit from
+                this object's context manager.
+        to_deactivate: List of constraint data objects that should be
+                       temporarily deactivated. These are restored to their
+                       original status on exit from this object's context
+                       manager.
+        to_reset: List of var data objects that should be reset to their
+                  original values on exit from this object's context
+                  context manager.
```
jsiirola, line 98: Can you reformat this to include the type?

pyomo/util/subsystems.py
```diff
@@ -143,22 +189,31 @@ def __init__(self,
                       of values of length n_scenario
         output_values: ComponentMap mapping each output variable to a list
                        of values of length n_scenario
+        to_fix: to_fix argument for base class
+        to_deactivate: to_deactivate argument for base class
+        to_reset: to_reset argument for base class. This list is extended
+                  with input variables.
```
jsiirola, line 195: Can you reformat this to include the argument type?

jsiirola (commented):
Some minor formatting changes to docstrings, and I think this is good to go.

pyomo/contrib/incidence_analysis/util.py
Robbybp: Updated the docstring

pyomo/util/subsystems.py
Robbybp: Updated

pyomo/util/subsystems.py
Robbybp: Updated

pyomo/util/subsystems.py
Robbybp: Updated

jsiirola (approved):

pyomo/contrib/incidence_analysis/util.py
```diff
@@ -97,8 +114,22 @@ def solve_strongly_connected_components(block, solver=None, solve_kwds=None):
     if solve_kwds is None:
         solve_kwds = {}
 
+    constraints = [con for con in
+            block.component_data_objects(Constraint, active=True)]
```
jsiirola, line 118: This can technically just be
```python
constraints = list(block.component_data_objects(Constraint, active=True))
```

Robbybp:
Just made a slight update to the API. `generate_strongly_connected_components` now accepts constraints and variables instead of a block. This is more general.

Testing failures appear unrelated.

jsiirola (approved):

blnicho (approved):

PR #2022, opened 2021-06-22
ExternalGreyBox subclass to embed implicit functions defined by Pyomo components

Robbybp:
## Summary/Motivation:
`ExternalGreyBoxModel` allows users to embed vector-valued external functions in their Pyomo models or blocks. One application of this is embedding one Pyomo model/block in another. This could be useful if we would like the "inner block" to include an implicit function from the perspective of the outer block. That is, the inner block performs a solve, computes derivatives via the implicit function theorem, and returns the results, either via outputs or residuals (here, I only implement residuals; outputs may be more general/useful).

This might be useful for certain types of NLPs where a subset of constraints form a "natural implicit function," e.g. index-1 DAEs.

There are several enhancements I would like to implement beyond what is in this PR:
- ~~Ability to create an ExternalGreyBoxBlock that does not add additional variables to the model (i.e. uses preexisting variables as inputs (and outputs if applicable)).~~
- ~~Fix IndexedExternalGreyBoxBlock. This didn't work when I tried to use it. Didn't spend long trying to debug.~~
- ~~Speed up tensor tensor arithmetic with Numpy. I currently store tensors as lists of Numpy arrays or Scipy matrices. This is so I could be very deliberate about how I take transposes and multiply/backsolve matrices. I can hopefully speed this up by using a single Numpy array for each tensor that I reshape and transpose-along-axes to fit my needs.~~
- Exploit potential sparsity in derivatives. Because derivatives are the results of matrix factorizations and backsolves, they are in general dense. Figuring out which entries _could possibly_ be nonzero seemed like a difficult/interesting task, so I left it out for now.
- Exploit potential decomposability in the inner solve. It may be the case that this solve can be performed as a sequence of smaller solves. If so, this should be exploited.

- **pynumero/interfaces/external_pyomo_model.py** module, which contains the `ExternalPyomoModel` class. 
- Tests. Mostly of the function/derivative evaluations that this class entails, but also that we can embed this properly in an `ExternalGreyBoxBlock`.

## Feedback requested:
This PR should be ready for review, but there are several things I could use feedback on for future development.
- Names. Should the name of this class indicate that its purpose is basically to implement implicit function theorem calculations?
- Division of classes. Would an ImplicitFunctionNLP-like object be useful as a general purpose construct? Should there be a class that implements the implicit function with only inputs and outputs, rather than inputs and residuals?

### Legal Acknowledgement


mrmundt:
@carldlaird , @michaelbynum - Are you able to review this PR? Thanks!

michaelbynum (approved):

PR #2027, opened 2021-06-23
Numpy expression integration (part 1)

### LINKED ISSUE #31: Bug using numpy float values in Pyomo expressions.

### LINKED ISSUE #685: Cannot do simple scalar variable multiplication...
I am unable to multiply numpy.float64 * SimpleVar, but I am able to multiply SimpleVar * numpy.float64. The error is as follows:

KeyError: "Cannot treat the scalar component 'x'as an indexed component"

where 'x' is of type pyomo.core.base.var.SimpleVar.

I am using Pyomo 5.5.1. Do we have support for numpy parameters in the models?

jsiirola:
## Summary/Motivation:
This is the initial step toward integrating numpy with the Pyomo expression system.  This provides basic support for generating Pyomo expressions contained in NumPy ndarrays.

It is by no means complete (there will be at least 3 more parts to finish the integration), but I felt breaking this up into chunks will make it easier to review.  However, this PR will support the following example:
```python
>>> import numpy as np
>>> m = ConcreteModel()
>>> m.x = Var([0,1,2,3])
>>> A = np.array([[1, 2, 3, 4], [5, 6, 7, 8]])
>>> b = np.array([10, 20])
>>> m.c = Constraint([0,1], expr=A @ m.x <= b)
>>> m.pprint()
2 Set Declarations
    c_index : Size=1, Index=None, Ordered=Insertion
        Key  : Dimen : Domain : Size : Members
        None :     1 :    Any :    2 : {0, 1}
    x_index : Size=1, Index=None, Ordered=Insertion
        Key  : Dimen : Domain : Size : Members
        None :     1 :    Any :    4 : {0, 1, 2, 3}

1 Var Declarations
    x : Size=4, Index=x_index
          0 :  None :  None :  None : False :  True :  Reals
          1 :  None :  None :  None : False :  True :  Reals
          2 :  None :  None :  None : False :  True :  Reals
          3 :  None :  None :  None : False :  True :  Reals

1 Constraint Declarations
    c : Size=2, Index=c_index, Active=True
        Key : Lower : Body                              : Upper : Active
          0 :  -Inf :   x[0] + 2*x[1] + 3*x[2] + 4*x[3] :  10.0 :   True
          1 :  -Inf : 5*x[0] + 6*x[1] + 7*x[2] + 8*x[3] :  20.0 :   True

4 Declarations: x_index x c_index c
```
Please note that this implementation is NOT intended to be "numpy-fast": we still generate the same expanded Pyomo expression trees that we always have.  However, this is intended to lay the groundwork for a fast "matrix-based" expression system later.

- add support for casting some `IndexedComponents` into `ndarray` objects
- add recognition for `ndarray` in `Initializer()`

### Legal Acknowledgement


pyomo/core/tests/unit/test_numpy_expr.py
```diff
@@ -0,0 +1,78 @@
...
+
+@unittest.skipUnless(numpy_available, 'numpy is not available')
+class TestNumPy(unittest.TestCase):
+    def test_numpy_scalar_times_scalar_var(self):
+        # Test issue #685
+        m = ConcreteModel()
+        m.x = Var()
+        e = np.float64(5) * m.x
+        self.assertIs(type(e), MonomialTermExpression)
+        self.assertEqual(str(e), "5.0*x")
```
michaelbynum, line 32: I would recommend using the `compare_expressions` function from `pyomo/core/expr/compare.py` throughout these tests.

michaelbynum (commented):
Overall, this looks great. I had one suggestion for the tests.

pyomo/core/tests/unit/test_numpy_expr.py
jsiirola: Good call.  Done.

michaelbynum (approved):

PR #2036, opened 2021-06-30
Improve management of tempfiles in tests

jsiirola:
## Summary/Motivation:
Recently infrastructure changes caused a number of tests to no longer clean up temporary files that they created.  This PR is the first step toward cleaning up the management of temporary files within tests

- remove use of the `sequential_files()` mode (make that mode a flag local to the current "tempfile context")
- create test files in the tempdir (and not the current working directory or the test file directory)
- improve management of some written files to ensure that they are registered to (and deleted by) a TempfileManager
- only keep temporary files for failed tests when the test module is run directly (i.e., by `python` and not through `nosetests` or `pytests`)

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/unittest.py
+++ b/pyomo/common/unittest.py
@@ -500,6 +500,11 @@ class TestCase(_unittest.TestCase):
         # Disable nose's use of test docstrings for the test description.
         return None
 
+    def currentTestPassed(self):
+        # Note: this only works for Python 3.4+
+        return not (self._outcome and any(
+            test is self and err for test, err in self._outcome.errors))
+
     def assertStructuredAlmostEqual(self, first, second,
                                     places=None, msg=None, delta=None,
                                     reltol=None, abstol=None,
--- a/pyomo/solvers/tests/mip/test_ipopt.py
+++ b/pyomo/solvers/tests/mip/test_ipopt.py
@@ -11,22 +11,21 @@
 import json
 import os
 from os.path import abspath, dirname, join
-currdir = dirname(abspath(__file__))
 
 import pyomo.common.unittest as unittest
+
+from pyomo.common.fileutils import this_file_dir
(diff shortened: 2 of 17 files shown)


pyomo/common/unittest.py
```diff
@@ -500,6 +500,11 @@ def shortDescription(self):
         # Disable nose's use of test docstrings for the test description.
         return None
 
+    def currentTestPassed(self):
+        # Note: this only works for Python 3.4+
```
blnicho, line 504: Do we need this comment considering we're only supporting 3.6+ now?

pyomo/solvers/tests/mip/test_ipopt.py
```diff
@@ -204,11 +214,12 @@ def test_bad_dof(self):
         m.c.add(m.x + m.y == 1)
         m.c.add(m.x - m.y == 0)
         m.c.add(2*m.x - 3*m.y == 1)
-        m.write('j.nl')
+        #m.write('j.nl')
```
blnicho, line 217: Can this commented line be removed?

blnicho (approved):

pyomo/common/unittest.py
jsiirola: It is not strictly necessary, but I have taken to leaving breadcrumbs when I have to dig around for a solution, especially if web searches turn up a number of alternative "solutions" (and the first 3 are all for Python<3.3).

pyomo/solvers/tests/mip/test_ipopt.py
jsiirola: Yes.  This file was written by the original test and not deleted.  Given that no part of the test checked or did anything with the file, I commented it out, but wanted to leave it in case someone knew why we were generating the file in the first place.  It can be deleted outright if no one objects.

PR #2059, opened 2021-07-26
Resolving tee issues with Xpress 8.9+

bknueven:
## Summary/Motivation:
Starting with Xpress 8.9 / 36, then behavior of console logging changed in Xpress. This PR keeps xpress_direct / xpress_persistent current with this change.

- Only create a "cbmessage" callback when tee is True for Xpress <8.9
- Capture the solver output in any case when tee is False

### Legal Acknowledgement

### DIFF
--- a/pyomo/solvers/plugins/solvers/xpress_direct.py
+++ b/pyomo/solvers/plugins/solvers/xpress_direct.py
@@ -170,11 +170,13 @@ class XpressDirect(DirectSolver):
 
         self._solver_model.setlogfile(self._log_file)
         if self._keepfiles:
-            print("Solver log file: "+self.log_file)
+            print("Solver log file: "+self._log_file)
 
-        # setting a log file in xpress disables all output
-        # this callback prints all messages to stdout
-        if self._tee:
+        # Setting a log file in xpress disables all output
+        # in xpress versions less than 36.
+        # This callback prints all messages to stdout
+        # when using those xpress versions.
+        if self._tee and XpressDirect._version[0] < 36:
             self._solver_model.addcbmessage(_print_message, None, 0)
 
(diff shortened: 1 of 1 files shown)


pyomo/solvers/plugins/solvers/xpress_direct.py
```diff
@@ -203,11 +205,23 @@ def _apply_solver(self):
...
+            self._solver_model.solve()
+        else:
+            # In xpress versions greater than or equal 36,
+            # it seems difficult to completely suppress console
+            # output without disabling logging altogether.
+            # As a work around, we capature all screen output
+            # when tee is False.
+            try:
+                with capture_output() as OUT:
+                    self._solver_model.solve()
```
jsiirola, line 218: Should this be sent to the `self._log_file`?  Also, why is this in a try-except block?

bknueven: The output will be sent to `self._log_file` via the method call on line 171: `self._solver_model.setlogfile(self._log_file)`.

As for the try/except, I wanted to be sure if the `solve` method raised a python Exception that it would be printed to the screen. Perhaps `capture_output()` handles this already, or this shouldn't be the behavior expected from a Pyomo solver with `tee=False`.

jsiirola: If the solver raises an exception, the error is printed to the screen at the very top of the call stack (that is, the exception passes out of the `capture_output` triggering the context manager's `__exit__`, so the output will be restored before the exception is ever printed to the scrieen.

jsiirola (approved):

mrmundt (approved):

PR #2061, opened 2021-07-27
Adding Coefficient Matching to PyROS

natalieisenberg:
## Summary/Motivation:
PyROS modelers may encounter uncertain equality constraints wherein no state variable is participating. Currently, we do not support these constraints and instead return an error. With the changes in this PR, we can support these types of constraints, which now means PyROS can handle any equality constraints in robust optimization. There were additional features and a bug fix made, outlined below. 

- Additional capability to now handle uncertain equality constraints of the form h(x,q) = 0, e.g. those not referencing state variables. With the coefficient matching capability, we now can handle any uncertain equality constraints. In order to achieve this, changes were made to standard_repn.py to support mutable Params. 
- Efficiency for separation under DiscreteScenarioSets to only consider scenarios not already explicitly added to master problem.
- A bug was fixed for how decision rule functions are generated under quadratic decision rules to ensure correct ordering of coefficients.
- UncertaintySet classes have new boundedness and non-emptiness properties.
- User interface improvements via renaming.

### Legal Acknowledgement


pyomo/contrib/pyros/util.py
```diff
@@ -459,6 +657,11 @@ def partition_powers(n, v):
         starts = (0,) + starts + (n+v,)
         yield [starts[i+1] - starts[i] - 1 for i in range(v)]
 
+def sort_partitioned_powers(powers_list):
+    powers_list = sorted(powers_list, reverse=True)
+    powers_list = sorted(powers_list, key=lambda elem: max(elem))
```
jsiirola, line 662: Out of curiosity, why sort this list twice?

jsiirola (commented):
This looks pretty good: just a few minor questions / comments.

pyomo/contrib/pyros/master_problem_methods.py
```diff
@@ -52,9 +51,8 @@ def solve_master_feasibility_problem(model_data, config):
     if not solver.available():
         raise RuntimeError("NLP solver %s is not available." %
                            config.solver)
-
     try:
-        results = solver.solve(model, tee=config.tee)
+        results = solver.solve(model, tee=False)
```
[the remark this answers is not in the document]
natalieisenberg: This is a debugging switch, I will revert that to the original.

pyomo/contrib/pyros/master_problem_methods.py
```diff
@@ -194,10 +192,10 @@ def minimize_dr_vars(model_data, config):
         raise RuntimeError("NLP solver %s is not available." %
                            config.solver)
     try:
-        results = solver.solve(polishing_model, tee=False)
+        results = solver.solve(polishing_model, tee=config.tee)
```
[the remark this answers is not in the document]
natalieisenberg: I think the current protocol will be that any call to a subsolver has tee=config.tee.

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -95,6 +91,55 @@ def parameter_bounds(self):
...
+        return True
+
+    def is_nonempty(self, config):
+        """
+        Return True if the uncertainty set is nonempty, else False.
+        """
+        if self.is_bounded(config):
+            return True
+        else:
+            return False
```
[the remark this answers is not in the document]
natalieisenberg: Indeed it is! An oversight.

pyomo/contrib/pyros/util.py
natalieisenberg: This sorting is to achieve the same ordering of powers as is generated by itertools.combinations, because I need to know the decision rule coefficient corresponding to each term for when we do the polishing step. The sorts are first lexicographic and then by max power.

pyomo/contrib/pyros/separation_problem_methods.py
```diff
@@ -237,29 +242,33 @@ def solve_separation_problem(model_data, config):
...
-    sep_priority_dict = config.separation_priority_order
+    config_sep_priority_dict = config.separation_priority_order
+    actual_sep_priority_dict = ComponentMap()
     for perf_con in model_data.separation_model.util.performance_constraints:
-        if perf_con.name not in sep_priority_dict.keys():
-            sep_priority_dict[perf_con.name] = 0
+        if perf_con.name not in config_sep_priority_dict.keys():
+            actual_sep_priority_dict[perf_con] = 0
+        else:
+            actual_sep_priority_dict[perf_con] = config_sep_priority_dict[perf_con.name]
```
jsiirola, line 256: FWIW, I was surprised to learn that this is efficient: I used to think of `keys()` as a generator, so would have expected this to either not work or be a linear time operation.  It turns out that this is only slightly less efficient than `perf_con.name not in config_sep_priority_dict`.

That said, this is really just:
```python
actual_sep_priority_dict[perf_con] = config_sep_priority_dict.get(perf_con.name, 0)
```
right?

jsiirola (commented):
One pair of bugs, then I think this is good to go.

jsiirola (approved):

pyomo/contrib/pyros/separation_problem_methods.py
natalieisenberg: I actually did not know about the dict get() function, I think this is a nicer way to do this...I will make the change.

jsiirola (approved):

blnicho (approved):

PR #2072, opened 2021-08-05
Resolve TeeStream deadlock / libgcc_s.so.1 loader issues

### LINKED ISSUE #2054: Error in solver tee output on Windows
## Summary

When I an solving a pyomo problem and use the "tee=True" option, and I scroll up in the CMD window on Windows to see some output displayed above, I often get the error below.

```sh
Number of objective function evaluations             = 30
Number of objective gradient evaluations             = 6
Number of equality constraint evaluations            = 31
Number of inequality constraint evaluations          = 0
Number of equality constraint Jacobian evaluations   = 6
Number of inequality constraint Jacobian evaluations = 0
Number of Lagrangian Hessian evaluations             = 5
Total CPU secs in IPOPT (w/o function evaluations)   =      0.668
Total CPU secs in NLP function evaluations           =     25.060

EXIT: Optimal Solution Found.
    solver.solve(m, tee=True)
  File "c:\users\eslickj\git\pyomo\pyomo\opt\base\solvers.py", line 578, in solve
    _status = self._apply_solver()
  File "c:\users\eslickj\git\pyomo\pyomo\opt\solver\shellcmd.py", line 246, in _apply_solver
    self._rc, self._log = self._execute_command(self._command)
  File "c:\users\eslickj\git\pyomo\pyomo\opt\solver\shellcmd.py", line 324, in _execute_command
    t.STDERR.flush()
  File "c:\users\eslickj\git\pyomo\pyomo\common\tee.py", line 390, in __exit__
    self.close(et is not None)
  File "c:\users\eslickj\git\pyomo\pyomo\common\tee.py", line 378, in close
    raise RuntimeError(
RuntimeError: TeeStream: deadlock observed joining reader threads
```

I'm not sure this happens 100% of the time, but you just need to run a pyomo problem on windows in a cmd window and be scrolling up when the solver completes.

It's not with a specific script.  It happens with with any problem I run.  Should be reproducible with any problem using ipopt and "tee=True."   

### Error Message

```
  File "c:\users\eslickj\git\pyomo\pyomo\common\tee.py", line 378, in close
    raise RuntimeError(
RuntimeError: TeeStream: deadlock observed joining reader threads
```

Pyomo version: Pyomo 6.0.2 (devel {string_args}) (CPython 3.9.4 on Windows 10)
Python version:  3.9.4
Operating system: Windows 10
How Pyomo was installed (PyPI, conda, source): source (pip with -e option)
Solver (if applicable): Ipopt

jsiirola:
## Summary/Motivation:
@eslickj noted that the recent rework of `TeeStream` introduced deadlock errors on Windows.  As part of debugging that issue, we ran across errors where `TeeStream` would cease working and abort with the error:
```
libgcc_s.so.1 must be installed for pthread_cancel to work
```

This appears to all be due to closing files in different threads from the thread that opened them.  This PR resolves that issue and adds a test that the deadlock error is correctly generated.  It also lengthens the timeout before we quit trying to wait for reader threads to terminate (to roughly 200 seconds) and introduces a separate timeout where we issue a warning about possible deadlock (after ~3 seconds).

- Resolve `TeeStream` handle errors where files were closed by a thread that didn't open them
- Lengthen `TeeStream` timeouts (and issue a warning before the deadlock triggers an exception)
- Add tests exercising the deadlock logic
- (unrelated) resolve a test that would fail if the user's HOME directory was accessed through a symlink

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/tee.py
+++ b/pyomo/common/tee.py
@@ -23,8 +23,18 @@ import threading
 import time
 from io import StringIO
 
-_mswindows = sys.platform.startswith('win')
 _poll_interval = 0.0001
+_poll_rampup_limit = 0.099
+# reader polling: number of timeouts with no data before increasing the
+# polling interval
+_poll_rampup = 10
+# polling timeout when waiting to close threads.  This will bail on
+# closing threast after a minimum of 13.1 seconds and a worst case of
+# ~(13.1 * #threads) seconds
+_poll_timeout = 3  # 15 rounds: 0.0001 * 2**15 == 3.2768
+_poll_timeout_deadlock = 200  # 21 rounds: 0.0001 * 2**21 == 209.7152
+
+_mswindows = sys.platform.startswith('win')
 try:
     if _mswindows:
         from msvcrt import get_osfhandle
@@ -370,11 +380,15 @@ class TeeStream(object):
             if not self._threads:
                 break
             _poll *= 2
-            if _poll >= 2:  # bail after a total of 2 seconds * #threads
+            if _poll_timeout <= _poll < 2*_poll_timeout:
                 if in_exception:
                     # We are already processing an exception: no reason
-                    # to trigger another
+                    # to trigger another, nor to deadlock for an extended time
                     break
+                logger.warning(
+                    "Significant delay observed waiting to join reader "
+                    "threads, possible output stream deadlock")
+            elif _poll >= _poll_timeout_deadlock:
                 raise RuntimeError(
                     "TeeStream: deadlock observed joining reader threads")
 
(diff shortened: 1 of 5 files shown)


pyomo/common/tee.py
```diff
@@ -345,12 +355,22 @@ def open(self, mode='w', buffering=-1, encoding=None, newline=None):
         if encoding is None:
             encoding = self.encoding
         handle = _StreamHandle(mode, buffering, encoding, newline)
+        # Note that is it VERY important to close file handles in the
+        # same thread that opens it.  If you don't you can see deadlocks
+        # and a peculiar error ("libgcc_s.so.1 must be installed for
+        # pthread_cancel to work"; see https://bugs.python.org/issue18748)
+        #
+        # To accomplish this, er will keep two handle lists: one is the
```
mrmundt, line 363: `er`? I assume that's a typo.

pyomo/opt/tests/solver/test_shellcmd.py
```diff
@@ -281,8 +292,13 @@ def test_executable_isexe_abspath_user(self):
             with SolverFactory(name, executable=isexe_abspath_user) as opt:
                 if isinstance(opt, UnknownSolver):
                     continue
-                self.assertEqual(opt._user_executable, isexe_abspath)
-                self.assertEqual(opt.executable(), isexe_abspath)
+                # Note the user path could be different from our
```
mrmundt, line 295: This isn't related, correct? It's just an extra test that is nice to have?

mrmundt (changes requested):

pyomo/opt/tests/solver/test_shellcmd.py
jsiirola: Worse: those assertions are not valid and will fail if we are running in a directory (like a user's HOME directory) that is accessed through symbolic links.

mrmundt (approved):

PR #2073, opened 2021-08-09
Allow native numeric types in relational expressions

jsiirola:
## Summary/Motivation:
While Pyomo5 Expressions allow native numeric types to appear in the expression tree, the relational operators (and by extension, the `Constraint.lower` and `Constraint.upper`) have explicitly wrapped these in `NumericConstant` objects.  This PR removes that behavior, making relational expression nodes (Equality, Inequality, and Ranged expressions) allow native types to appear as arguments.

For backwards compatibility, `Constraint.body`, `Constraint.lower` and `Constraint.upper` still return numeric types (i.e., PyomoObject derivatives obtained by passing the value through `as_numeric()` before returning it).  To avoid this, Constraint data objects now also support `lb` and `ub` properties for obtaining the raw bound value.  This API was chosen to be consistent with the Kernel constraint API.

This PR also relaxes the set of allowable constraint expressions: rules may now return ranged expressions with potentially variable bounds: the exception for variable bounds is only emitted when interrogating the Constraint (through `lower` / `upper` / `lb` / `ub`).  This enables `Expression` objects to be used as constraint bounds (as long as the resulting expressions are fixed by the time the bounds are interrogated). 

Finally, these changes set the stage for improving the writer performance (a draft updated LP writer shows ~20% improvement on p-median test8) and allowing for alternative storage models for Constraint types (notably templated constraints).

- allow native numeric types to appear as arguments in relational expressions
- add new `lb` and `ub` properties to ConstraintData to return the "raw" bound values
- allow ranged inequalities with potentially variable expressions in lower/upper bounds
- update relevant tests

### Legal Acknowledgement


mrmundt (changes requested):

mrmundt (approved):
Nice additions!

PR #1613, opened 2020-09-05
Add kaug dsdp mode into sens.py 

JanghoPark-LBL:
## Summary/Motivation:
The current sens.py uses the only sipopt. kaug dsdp mode has added as another option for sensitivity. 

## Changes proposed in this PR:
- Add ipopt solver option as an input, optarg=None (`line 229, 375-376`)
- kaug requires variable initialization (`line 311-316`). 
- k_aug doesn’t support inequalities. Raise exception error (`line 363`). 
- This function requires ipopt, k_aug, dotsens (`line 374-384`). 
- Declare Suffixes(`line 386-412`). 
- ipopt.solve -> kaug.solve -> dotsens.solve (`line 415-428`). 
- fixes #2047

### Legal Acknowledgement

By contributing to this software project, I have read the [contribution guide](https://pyomo.readthedocs.io/en/stable/contribution_guide.html) and agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #2082, opened 2021-08-16
fix an error in the doc string of get_dfds_dcds()

jialuw96:
@adowling2 Fixed an error in the doc string of get_dfds_dcds(): each row in gradient_c should be [row number, col number, and value], and the index starts from 0.

### DIFF
--- a/pyomo/contrib/sensitivity_toolbox/sens.py
+++ b/pyomo/contrib/sensitivity_toolbox/sens.py
@@ -301,9 +301,9 @@ def get_dfds_dcds(model, theta_names, tee=False, solver_options=None):
     gradient_c: scipy.sparse.csr.csr_matrix
         Ncon by Nvar size sparse matrix. A Jacobian matrix of the
         constraints with respect to the (decision variables, parameters)
-        at the optimal solution. Each row contains [column number,
-        row number, and value], column order follows variable order in col
-        and index starts from 1. Note that it follows k_aug.
+        at the optimal solution. Each row contains [row number,
+        column number, and value], column order follows variable order in col
+        and index starts from 0. Note that it follows k_aug.
         If no constraint exists, return []
     col: list
         Size Nvar list of variable names


adowling2:
@jialuw96 Per our conversation last week, were you also going to migrate the changes in `sens.py` from IDAES to Pyomo? I am only seeing one change here (fix the doc string).

jialuw96:
> @jialuw96 Per our conversation last week, were you also going to migrate the changes in `sens.py` from IDAES to Pyomo? I am only seeing one change here (fix the doc string).

@adowling2 I compared the date of sens.py of IDAES and Pyomo. The Pyomo one has a more recent date so I am not sure if we should replace that.

blnicho:
@Robbybp I think you made sure the changes to `sens.py` in IDAES made it into #1613. Could you confirm that's the case?

Robbybp:
@blnicho I wasn't aware there were any changes in the IDAES instance. I think I made changes to the Pyomo version since the IDAES one got merged (K_augInterface that uses a temporary directory), but am not aware of any changes to the IDAES version.

pyomo/contrib/sensitivity_toolbox/sens.py
```diff
@@ -301,9 +301,9 @@ def get_dfds_dcds(model, theta_names, tee=False, solver_options=None):
     gradient_c: scipy.sparse.csr.csr_matrix
         Ncon by Nvar size sparse matrix. A Jacobian matrix of the
         constraints with respect to the (decision variables, parameters)
-        at the optimal solution. Each row contains [column number,
-        row number, and value], column order follows variable order in col
-        and index starts from 1. Note that it follows k_aug.
+        at the optimal solution. Each row contains [row number,
+        column number, and value], column order follows variable order in col
+        and index starts from 0. Note that it follows k_aug.
         If no constraint exists, return []
```
Robbybp, line 307: Is this docstring describing the structure of a CSR matrix? What does "row" refer to for a CSR matrix? The sub-arrays in the `indices` and `data` arrays corresponding to a particular row? Unless this is documenting something I'm missing, I would just cut out everything after `at the optimal solution.` Also, if no constraints exist, this docstring is incorrect, because we return an empty array rather than an empty list. In this case we should probably just return an empty CSR matrix...

Robbybp (changes requested):

pyomo/contrib/sensitivity_toolbox/sens.py
adowling2: I think this is describing the structure of a CSR matrix. @jialuw96 take a look at this: https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.csr_matrix.html

I agree with the suggestion to return an empty CSR matrix.

adowling2: The docstring is saying (I think) that the rows and columns of gradient_c correspond to the equations and variables, respectively. @Robbybp Do you agree with this? I agree the description of the CSR matrix is confusing. I'd say just remove the [ ] and reword the sentence after "optimal".

Robbybp: Yes, I agree that rows correspond to equations and columns correspond to variables. It is a little confusing because k_aug appears to print out the (row, col, data) of the transpose of this matrix, then this method switches the row and col entries to return this gradient_c, which I think is the more intuitive representation.

adowling2: @jialuw96 We should update the docstring (or comments or both) to reflect this nuance.

Robbybp, line 307: I would recommend replacing this portion of the docstring with:

```python
"""
Jacobian matrix of constraints with respect to variables and parameters.
Each row corresponds to a constraint and each column corresponds to a
variable or parameter.
"""
```

adowling2:
@jialuw96 Please update `sens.py` in PR #2124 to incorporate the suggestion above.

This PR is can be closed; it is superseded by PR #2124.

adowling2:
We found that the Pyomo/contrib version of `sens.py` has a bug that was fixed in the IDAES version of `sens.py`. See Issue #2123 and PR #2124.

PR #2097, opened 2021-08-19
Correct calculation of Constraint.equality for RangedExpression with constant bounds

### LINKED ISSUE #2093: Constraint's equality attribute sometimes fails to properly identify equality constraints
## Summary
Occasionally constraint's new equality attribute will fail in ways that are difficult to reproduce.

This happened while I was running the sequential decomposition tool, but I can't seem to create a small example which reproduces the issue on my system. Instead I've included some hopefully useful output from the debugger:
```python
ipdb> type(con)
<class 'pyomo.core.base.constraint._GeneralConstraintData'>
ipdb> type(con._expr)
<class 'pyomo.core.expr.logical_expr.RangedExpression'>
ipdb> con.pprint()
{Member of flow_mass_phase_comp_equality} : Size=2, Index=fs.feed._flow_mass_phase_comp_outlet_ref_index, Active=True
    Key                 : Lower : Body                                                                                                                                       : Upper : Active
    (0.0, 'Liq', 'H2O') :   0.0 : fs.feed.properties[0.0].flow_mass_phase_comp[Liq,H2O] - fs.PrimaryPumps[1].control_volume.properties_in[0.0].flow_mass_phase_comp[Liq,H2O] :   0.0 :   True
ipdb> con.lb == con.ub
True
ipdb> con.lb is con.ub
False
ipdb> type(con.lb)
<class 'float'>
ipdb> type(con.ub)
<class 'float'>
ipdb> id(con.lb)
140324945312432
ipdb> id(con.ub)
140324945312400
ipdb> 
```

Pyomo version: 6.1.1
Python version: 3.7.9
Operating system: Ubuntu 20.04.2
Solver (if applicable): N/A

### Additional information
I believe the issue is with the `is` check on bounds in the `equality` attribute on line 416:
https://github.com/Pyomo/pyomo/blob/3cf310ac1d52b7e362dc2eaf6b8a59ac4c896fe4/pyomo/core/base/constraint.py#L408-L418

For `RangedExpressions` with floating point bounds, CPython does not guarantee that the same valued `float` objects will occupy the same location in memory (as shown by the debugging output above). However, it is usually the case that identical floating point values *do* occupy the same location in memory, making this difficult to reproduce. There was some discussion around resolving this issue in Python itself: https://discuss.python.org/t/demoting-the-is-operator-to-avoid-an-identity-crisis/86, but the CPython team went a different route: https://bugs.python.org/issue34850.

jsiirola:
## Summary/Motivation:
As was documented in #2093, the new implementation of `ConstraintData.equality` would occasionally fail to identify equality constraints for `RangedExpression` objects where the bounds were constants.  This resolves that issue by using `==` instead of `is` when both bounds are constants (native types or constant expressions).

- Use `==` instead of `is` to compare constraint bounds and determine `equality` when both bounds are constants.

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/constraint.py
+++ b/pyomo/core/base/constraint.py
@@ -25,6 +25,7 @@ from pyomo.common.timing import ConstructionTimer
 from pyomo.core.expr import logical_expr
 from pyomo.core.expr.numvalue import (
     NumericValue, value, as_numeric, is_fixed, native_numeric_types,
+    is_constant,
 )
 from pyomo.core.base.component import (
     ActiveComponentData, ModelComponentFactory,
@@ -411,9 +412,13 @@ class _GeneralConstraintData(_ConstraintData):
         if self._expr.__class__ is logical_expr.EqualityExpression:
             return True
         elif self._expr.__class__ is logical_expr.RangedExpression:
-            # TODO: this is a very restrictive form of structural equality.
-            lb = self._expr.arg(0)
-            if lb is not None and lb is self._expr.arg(2):
+            lb, _, ub = self._expr.args
+            if lb is None:
+                return False  # unbounded
+            if lb is ub:
+                return True  # structurally equal
+            if is_constant(lb) and is_constant(ub) and value(lb) == value(ub):
+                # Should we consider an equality tolerance?
                 return True
         return False
 
--- a/pyomo/core/tests/unit/test_con.py
+++ b/pyomo/core/tests/unit/test_con.py
@@ -170,10 +170,105 @@ class TestConstraintCreation(unittest.TestCase):
             return (0, model.y, 1)
         model.c = Constraint(rule=rule)
 
-        self.assertEqual(model.c.equality,         False)
-        self.assertEqual(model.c.lower,             0)
-        self.assertIs   (model.c.body,              model.y)
-        self.assertEqual(model.c.upper,             1)
+        self.assertEqual(model.c.equality, False)
+        self.assertEqual(model.c.lower, 0)
+        self.assertIs   (model.c.body, model.y)
+        self.assertEqual(model.c.upper, 1)
+
+        model = self.create_model()
+        def rule(model):
+            return (0, model.y, 0)
+        model.c = Constraint(rule=rule)
+
+        self.assertEqual(model.c.equality, True)
+        self.assertEqual(model.c.lower, 0)
+        self.assertIs   (model.c.body, model.y)
+        self.assertEqual(model.c.upper, 0)
+
+        # Test mixed bound types (int / float)
+        model = self.create_model()
+        def rule(model):
+            return (0, model.y, 0.)
+        model.c = Constraint(rule=rule)
+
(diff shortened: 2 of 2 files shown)


ghackebeil:
This feels like a move in the wrong direction. The use of two inequalities to express a constraint may be intentional, for reasons outside of Pyomo.

jsiirola:
@ghackebeil I spent some time thinking about exactly this.  Some discussion points included should we relax this more (that is, allow an 'equality tolerance'), or leave it more strict (leaning toward a very narrow relaxation of the `is` test, e.g., just to cover the originally reported issue with `0. is 0.` occasionally not being `True` and comparison of `0` with `0.`).  The final argument for this approach centered on:
- this removes a regression and restores behavior seen pre-6.1
- this is restricted to structural equality:  it will only return True if the lower bound and upper bound are *constants*  and equal
- if a user really wants to express this as two inequalities, they still can.

blnicho (approved):

ghackebeil:
I'm simply saying that ranged inequality expressions may be important in whatever form the Pyomo model is translated into, whether or not the upper and lower bounds are constant, equal, parameters, or whatever. A user may want that structure retained (in, say, the LP or NL file). It seems inconsistent to turn a constraint into equality form, simply because the bound types or mutability changes.

My vote is for the entire `elif self._expr.__class__ is logical_expr.RangedExpression:` block to just be removed, but it's just a suggestions.  `¯\_(ツ)_/¯`

michaelbynum:
Should we have an explicit mechanism for users to have control over this behavior? This could be done a few different ways. We could have special constraint types or we could have functions that construct "the right thing" based on user input. I think it is best to make the default behavior something reasonable (probably what is implemented in this PR) but also allow users to have greater control if necessary.

jsiirola:
@blnicho and I were talking about this, too.  I think there are three possible interpretations of `ConstraintData.equality`:
1. Was the Constraint initialized with an `EqualityExpression` object (as opposed to an `InequalityExpression` or `RangedExpression`)?
2. Does the `ConstraintData` represent a "structural equality"; that is (1) plus `RangedExpression` where the upper and lower bounds are symbolically equal?
3. Does the `ConstraintData` represent a "numerical equality"; that is (1) plus `RangedExpression` where the upper and lower bounds are have equal `value()`?

Looking through the code base, there is not real consistency ... some places assume `equality` means case (1) (primarily transformations), and others (primarily writers) assume something closer to (3).  @ghackebeil makes a good point that users may have had a good reason for expressing an equality as a ranged expression.  Similarly, some solvers (like PyROS) would be much more interested in (3) (that is, the algorithm treats equality constraints very differently from inequalities).

We should probably document which one `equality` refers to (the property has never been documented), as well as how clients that care can get the other two versions if they care.  One possible path forward is to deprecate `equality` and replace it with 1 or more new methods / properties that have more explicit names.

jsiirola:
Closing this PR and replacing it with PEP #2105

PR #2109, opened 2021-08-30
Overhaul of Pyomo TempfileManager

jsiirola:
## Summary/Motivation:
Discussions with DISPATCHES / IDAES developers (notably, @andrewlee94) highlighted several shortcomings and weaknesses in the `TempfileManager` class (originally developed for PyUtilib)

This PR overhauls the internal implementation of the `TempfileManager` class.  This update is backwards compatible with the previous implementation, but adds several features:
- handles both `str` and `bytes`, using the same logic as the Python `tempfile` module
- adds several new methods for consistency with the `tempfile` module (`mkstemp`, `mkdtemp`, `gettemptdir`, `gettemptdirb`, `gettemptprefix`, `gettemptprefixb`)
- defines a proper "context" object that can be used to create / remove temporary files.  This allows users to create and manipulate temporary file contexts independently of the normal `TempfileManager` "context stack"
- improved robustness for deleting temporary files when either the context or a local `TempfileManager` instance fall out of scope.
- incorporate documentation into Pyomo online docs

- (see above)

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/tempfiles.py
+++ b/pyomo/common/tempfiles.py
@@ -40,11 +40,6 @@ class TempfileManagerClass(object):
     Pyomo declares a global instance of this class as ``TempfileManager``:
 
     .. doctest::
-       :hide:
-
-       >>> import os
-
-    .. doctest::
 
        >>> from pyomo.common.tempfiles import TempfileManager
 
@@ -59,6 +54,7 @@ class TempfileManagerClass(object):
 
     .. doctest::
 
+       >>> import os
        >>> with TempfileManager.new_context() as tempfile:
        ...     fd, fname = tempfile.mkstemp()
        ...     dname = tempfile.mkdtemp()
--- a/pyomo/common/tempfiles.py
+++ b/pyomo/common/tempfiles.py
@@ -264,6 +264,7 @@ class TempfileContext:
 
         """
         dir = self._resolve_tempdir(dir)
+        # Note: ans == (fd, fname)
         ans = tempfile.mkstemp(suffix=suffix, prefix=prefix, dir=dir, text=text)
         self.tempfiles.append(ans)
         return ans
@@ -336,10 +337,7 @@ class TempfileContext:
     def create_tempfile(self, suffix=None, prefix=None, text=False, dir=None):
         """Create a unique temporary file.
 
-        If this context is generating sequential files (see
-        :meth:`sequential_files()`), the directory name will be
-        generated as ``prefix + counter + suffix``; otherwise, the name
-        is generated as in :func:`tempfile.mkstemp()`.
+        The file name is generated as in :func:`tempfile.mkstemp()`.
 
         Any file handles to the new file (e.g., from :meth:`mkstemp`)
         are closed.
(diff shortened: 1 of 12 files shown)


pyomo/opt/tests/base/test_convert.py
```diff
@@ -71,6 +71,7 @@ def write(self,filename="", format=None):
 class OptConvertDebug(unittest.TestCase):
 
     def setUp(self):
+        TempfileManager.push()
```
carldlaird, line 74: @jsiirola maybe consider push_context() instead of push()?

carldlaird (approved):
Please address the comment around DELETE_COUNTED_CONFLICTS. Rest are all minor.

pyomo/opt/tests/base/test_convert.py
jsiirola: That would definitely be a better name.  Unfortunately, that is a (very)old API.  We could change it, but then would have to add the deprecation paths, etc...  Maybe something to do in a subsequent PR?

pyomo/core/tests/unit/test_param.py
```diff
@@ -1562,7 +1562,6 @@ def test_mutable_self(self):
 
     # Test that display actually displays the correct param value
     def test_mutable_display(self):
-        tmp_stream = TempfileManager.create_tempfile(suffix = '.param_display.test')
         model = ConcreteModel()
```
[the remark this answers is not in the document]
jsiirola: Best I can tell, nothing.  It looks like it was code rot left over from before we started using StringIO to capture output for testing.

pyomo/common/tempfiles.py
```diff
@@ -29,174 +30,497 @@
...
+                prefix + next_id + suffix
+            )
+            if not os.path.exists(new_target):
+                break
+            if DELETE_COUNTED_CONFLICTS:
+                # Delete any file having the sequential name and then rename
+                logger.warning("Deleting a pre-existing (sequential) "
+                               "temporary file: %s" % new_target)
+                self._remove_filesystem_object(new_target)
+                break
```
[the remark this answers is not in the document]
jsiirola: It absolutely will delete files we didn't create.  This was put in to preserve previous (bad?) behavior.  I think a better solution would be to remove the `sequential_files` logic entirely.  I just looked and cannot find any reference to it in either Pyomo nor PySP.  I think I will put in a separate PR to remove that functionality entirely.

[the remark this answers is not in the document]
jsiirola: Actually, I went ahead and removed the sequential_files functionality.  Assuming tests pass, that is the better solution.

pyomo/common/tempfiles.py
```diff
@@ -30,173 +31,432 @@
...
-        self.files.append(filename)
+       >>> from pyomo.common.tempfiles import TempfileManager
 
-    def __iter__(self):
-        return iter(self.files)
+    This class provides an interface for managing
+    :class:`TempfileContext` contexts.  It implements a basic stack,
+    where users can :meth:`push()` a new context (causing it to become
+    the current "active" context) and :meth:`pop()` contexts off
+    (optionally deleting all files associated with the context.  In
```
blnicho, line 55: Typo: Missing " ) "

pyomo/common/tempfiles.py
```diff
@@ -30,173 +31,432 @@
...
+    general usage, users will either use the this class to create new
```
blnicho, line 56: Typo: "use the this class"

blnicho (approved):
I found a couple typos but otherwise this looks fine

Issue #1601, opened 2020-08-19
XPRESS Solver plugin has some oddities


mrmundt:
When completing the work for #1517 , I ran across an oddity in the `XPRESS` plugin. Namely, in `pyomo/solvers/plugins/solvers/XPRESS.py `, starting at line `372`, there are a lot of undefined variable names and references. 

I am sure this would cause issues once we we are able to test XPRESS and want to keep the oddities on our radar.

The note from 1517:

> ### TODO: What is going on here with field_name, and shortly thereafter, with variable_status and whatnot? They've never been defined.
> ### It seems like this is trying to mimic the CPLEX solver but has some issues

bknueven:
I have noticed this too, and in my own testing, the XPRESS plugin is broken for modern versions of XPRESS. Until somebody re-writes it, it may be better if `SolverFactory(“xpress”)` returned an XpressDirect solver instance.

Issue #1429, opened 2020-05-08
XPRESS.py failed to recognize model as MIP


bbrunaud:
Hello, 
I have a model that I am solving using Xpress for which I was getting the wrong result. (Unfortunately I don't have a minimum working example I can provide).

After doing some digging, I realized the model was not being recognized as MIP and was being solved as LP instead. I fixed the issue by manually setting: `solver.is_mip = True` before solving. 

Would it be possible to get rid of the `lpoptimize` command and use `mipoptimize` for all models?

This is a snippet of my code
```python3
    import pyomo.environ as pyomo

    m = ConcreteModel()

    m.w = pyomo.Var(items, intervals, within=pyomo.Binary)
    m.wn = pyomo.Var(superset, within=pyomo.Binary)
    m.sh = pyomo.Var(itemshift, within=pyomo.Binary)
    m.v = pyomo.Var(brands,intervals, within=pyomo.Binary)
    m.V = pyomo.Var(intervals, within=pyomo.NonNegativeReals)
    m.P = pyomo.Var(intervals, within=pyomo.NonNegativeReals)
    m.t = pyomo.Var(items,intervals, within=pyomo.NonNegativeReals)
    m.u = pyomo.Var(intervals, within=pyomo.NonNegativeReals)
    m.mu = pyomo.Var(within=pyomo.NonNegativeReals)
```

https://github.com/Pyomo/pyomo/blob/961facde258314d7168b5d578d8ca3eca1085278/pyomo/solvers/plugins/solvers/XPRESS.py#L160-L165

Issue #69, opened 2016-11-14
Unable to get slack with XPRESS.py


kdheepak:
``` python
from pyomo.environ import *

m = ConcreteModel()
m.x = Var()
m.y = Var(bounds=(0,1))
m.o = Objective(expr= m.x + m.y)
m.c = Constraint(expr= m.x >= 1)
m.dual = Suffix(direction=Suffix.IMPORT)
m.rc = Suffix(direction=Suffix.IMPORT)
m.slack = Suffix(direction=Suffix.IMPORT)

solver = SolverFactory("xpress",
                       solver_io="lp")
results = solver.solve(m)

print(m.o())
print(m.x())
print(m.y())
print(m.dual[m.c])
print(m.rc[m.y])
print(m.slack[m.c])

# should print                                                                                 
# 1                                                                                            
# 1                                                                                            
# 0                                                                                            
# 1                                                                                            
# 1                                                                                            
# 0  
```

The output is below.

```
1.0
1.0
0.0
1.0
1.0
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-1-090dcfa991bb> in <module>()
     19 print(m.dual[m.c])
     20 print(m.rc[m.y])
---> 21 print(m.slack[m.c])
     22 
     23 # should print

/Users/$USER/miniconda2/lib/python2.7/site-packages/pyomo/core/base/suffix.pyc in __getitem__(self, component)
    197             raise KeyError("Component with name: "
    198                            +cname+
--> 199                            " (id=%s)" % id(component))
    200 
    201     def __setitem__(self, component, value):

KeyError: 'Component with name: c (id=4484422800)'
```

The above uses `lpoptimize` in XPRESS.py. Any thoughts on what the problem might be?

jotaigna:
I have an idea about this issue, I faced a similar problem extracting duals. The problem in question had several equalities whose dual had a value of zero. The current XPRESS plugin extracts variables by detecting the type `C` in the second column of a wrtsol solution file. 

All other variables (i.e. `L` for less than constraints, `G` for greater than constraints and `E` for equality constraints) are treated the same: not scanning their type and extracting slacks and duals only if their value is different from zero: https://github.com/Pyomo/pyomo/blob/7beec1ba8b37c4f7b0b666db3d20c8675b0784ae/pyomo/solvers/plugins/solvers/XPRESS.py#L389-L392

They will show in the component map as `None` and return a `KeyError` when read.

[...]

Issue #59, opened 2016-10-22
XPRESS: NameError: global name 'field_name' is not defined


kdheepak:
[XPRESS.py](https://github.com/Pyomo/pyomo/blob/master/pyomo/solvers/plugins/solvers/XPRESS.py#L354) does not define `field_name` before using it. The following is the traceback from running the example in #58 

```
$ python test.py
Traceback (most recent call last):
  File "test.py", line 14, in <module>
    results = solver.solve(m)
  File "/Users/$USER/anaconda/lib/python2.7/site-packages/pyomo/opt/base/solvers.py", line 587, in solve
    result = self._postsolve()
  File "/Users/$USER/anaconda/lib/python2.7/site-packages/pyomo/opt/solver/shellcmd.py", line 267, in _postsolve
    results = self.process_output(self._rc)
  File "/Users/$USER/anaconda/lib/python2.7/site-packages/pyomo/opt/solver/shellcmd.py", line 329, in process_output
    self.process_soln_file(results)
  File "/Users/$USER/anaconda/lib/python2.7/site-packages/pyomo/solvers/plugins/solvers/XPRESS.py", line 354, in process_soln_file
    if (extract_reduced_costs is True) and (field_name == "reducedCost"):
NameError: global name 'field_name' is not defined
```

PR #2125, opened 2021-09-08
Issue warning when using the Xpress LP/MPS interface

bknueven:
## Fixes:
There are lots of open issues around the Xpress LP/MPS interface (#58, #59, #69, #1429, #1601).

## Summary/Motivation:
The Xpress LP/MPS interface only partially works with current versions of Xpress. This PR would add a warning directing users to the direct or persistent interface.

- Add warning when `_xpress_shell` is created
- Remove `xpress` from solver factory tests as the new warning causes this test to fail

### Legal Acknowledgement

### DIFF
--- a/pyomo/environ/tests/standalone_minimal_pyomo_driver.py
+++ b/pyomo/environ/tests/standalone_minimal_pyomo_driver.py
@@ -98,6 +98,9 @@ def run_writer_test():
 def run_solverfactory_test():
     skip_solvers = {
         'py',
+        'xpress',
+        '_xpress_shell',
+        '_mock_xpress',
     }
 
     with LoggingIntercept() as LOG, capture_output(capture_fd=True) as OUT:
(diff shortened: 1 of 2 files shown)


ghackebeil:
I believe the XPRESS folks agreed we should just make the Python interface the default, and just remove the legacy file based interface altogether. Can we just do that in this PR?

bknueven:
> I believe the XPRESS folks agreed we should just make the Python interface the default, and just remove the legacy file based interface altogether. Can we just do that in this PR?

I would be very happy to implement that solution. This PR, as it is, is my implementation @jsiirola's  suggestion to point users through a warning to the python interface.

blnicho (approved):

jsiirola (approved):
Approving this as a first step.  We are likely to revisit the meaning of the "generic" solver interface names in the near future.

blnicho:
Merging this as-is but the longer term plan is to develop a PEP for cleaner handling of the default solver interface that gets returned when multiple options are available.

PR #2142, opened 2021-10-05
Add external functions to block in `create_subsystem_block`

Robbybp:
## Summary/Motivation:
`create_subsystem_block` adds variables and constraints to a block, presumably to be solved as a subsystem. However, if these constraints contained external functions, the nl writer would not work with one of these subsystem blocks because it needs to be able to find, in the block hierarchy it is sent, a function of the same name for every external function present in the constraints it parses. This PR adds `ExternalFunction` components to the "subsystem block" for every unique function present in the constraints, allowing blocks created from `create_subsystem_block` to be solved with "nl solvers."

- Add `ExternalFunction` components to block returned by `create_subsystem_block`

### Legal Acknowledgement

### DIFF
--- a/.coin-or/projDesc.xml
+++ b/.coin-or/projDesc.xml
@@ -227,8 +227,8 @@ Carl D. Laird, Chair, Pyomo Management Committee, claird at andrew dot cmu dot e
 			     Use explicit overrides to disable use of automated
 			     version reporting.
 			-->
-			<stableVersionNumber>6.1.1</stableVersionNumber>
-			<releaseNumber>6.1.1</releaseNumber>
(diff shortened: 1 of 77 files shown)


Robbybp:
Looks like I accidentally did this on a branch downstream of https://github.com/Pyomo/pyomo/pull/2128

Robbybp:
> My understanding is that this is needed because the current NL writer only handles external functions whose ExternalFunction component is in Block tree being solved, correct? Is a better (long-term?) solution to relax that requirement?

The current NL writer actually seems not to care where the ExternalFunction component lives, as long as it can find an ExternalFunction with the same library and function names on the Block tree being solved. I believe a better solution would be to relax this requirement.

PR #2155, opened 2021-10-15
`ExternalPyomoModel` performance improvements

Robbybp:
## Summary/Motivation:
The implicit function implementation in `ExternalPyomoModel` was pretty slow, so I spent some time debottlenecking it, and made some good progress. A distillation column example that took 242s to solve previously now takes 15s to solve.

- Don't write more nl files than necessary
- Evaluate the implicit function using an SCC decomposition
- Calculate the reduced-space Hessian-of-Lagrangian all at once instead of constructing from individual constraint Hessians

Before:
imp-fcn-old
and after:
imp-fcn-new

### Legal Acknowledgement


Robbybp:
This PR now modifies PyomoNLP, PyomoNLPwithGreyBoxBlocks, and _ExternalGreyBoxAsNLP to use a name buffer, which gives another significant speedup:
image

PR #2159, opened 2021-10-21
Add uncertain variable bounds detection to PyROS

shermanjasonaf:
## Fixes # .
Add uncertain variable bounds detection to PyROS.

## Summary/Motivation:
Whenever PyROS replaces variable bounds with explicit inequality constraints through the `pyomo.contrib.pyros.util.turn_bounds_to_constraints` function, the resulting constraint expressions do not contain any mutable (and therefore, any uncertain) parameters on which the bounds may depend, but the values of these parameters at the point of bound/constraint replacement. Hence, the `make_separation_problem` and `make_separation_objectives` functions in `pyomo.contrib.pyros.separation_problem_methods` generate the corresponding separation problem objective function expressions incorrectly for second-stage variables whose bounds are uncertain.

To resolve, obtain the variable bound expressions through the protected `_lb` and `_ub` attributes, and use these expressions to generate explicit inequality constraints for all variables whose bounds are contingent on uncertain parameters. This constitutes a temporary fix, to be addressed further with the use of a more secure getter/API in the future.

-
-

### Legal Acknowledgement


pyomo/contrib/pyros/util.py
```diff
@@ -279,6 +280,56 @@ def transform_to_standard_form(model):
...
+
+        # add explicit inequality constraint(s), remove variable bound(s)
+        if mutable_params_ub & uncertain_param_set:
+            uncertain_var_bound_constrs.add(v - v._ub <= 0)
+            v.setub(None)
+        if mutable_params_lb & uncertain_param_set:
+            uncertain_var_bound_constrs.add(v._lb - v <= 0)
+            v.setlb(None)
+
+    return
```
jsiirola, line 330: Technically, this "return" is not needed.

jsiirola (approved):

pyomo/contrib/pyros/tests/test_grcs.py
```diff
@@ -271,6 +273,83 @@ def test_bounds_to_constraints(self):
...
+        replace_uncertain_bounds_with_constraints(mod_2, uncertain_params)
+
+        # check that vars participating in
+        # activated constraints correctly determined
+        svars = ComponentSet([v for v in get_vars_from_constraints(mod_2)])
+        vars_in_active_constraints = ComponentSet([mod_2.z, mod_2.w, mod_2.y,
+                                                   mod_2.x, mod_2.v])
+        self.assertEqual(svars, vars_in_active_constraints,
+                         msg='Mismatch in variables participating in '
+                             'activated constraints.')
```
Robbybp, line 344: Do you check that the constraints you add are correct? You probably want to check that `mod_2.uncertain_var_bound_cons` contains the variables and parameters you expect it to. Right now it looks like you're just checking that model (which contains other constraints) contains the variables you expect it to.

pyomo/contrib/pyros/util.py
```diff
@@ -279,6 +280,56 @@ def transform_to_standard_form(model):
...
+    """
+    uncertain_param_set = ComponentSet(uncertain_params)
+
+    # component for explicit inequality constraints
+    uncertain_var_bound_constrs = ConstraintList()
+    model.add_component(unique_component_name(model,
+                                              'uncertain_var_bound_cons'),
+                        uncertain_var_bound_constrs)
+
+    for v in get_vars_from_constraints(model):
```
Robbybp, line 316: Do you want to add constraints for vars in the objective as well? Or just in constraints?

pyomo/contrib/pyros/tests/test_grcs.py
emma58: I agree with @Robbybp that this would be a good idea. I like to use `generate_standard_repn` on the constraint body for this--it's pretty easy since your constraints are linear. You could look at [these tests](https://github.com/Pyomo/pyomo/blob/main/pyomo/contrib/fme/tests/test_fourier_motzkin_elimination.py#L82) for a fairly self-contained example.

pyomo/contrib/pyros/util.py
shermanjasonaf: Addressed in the subsequent commits.

pyomo/contrib/pyros/util.py
```diff
@@ -280,8 +280,21 @@ def transform_to_standard_form(model):
     return
 
 
+def get_vars_from_objective(block):
```
jsiirola, line 283: This is fine,  However, this method and `get_vars_from_constraints` are basically identical.  You might consider replacing both with a common method,
```python
def get_vars_from_component(block, ctype):
    """Determine all variables used in active components within a block.

    Parameters
    ----------
    block: Block
        The block to search for components.  This is a recursive
        generator and will descend into any active sub-Blocks as well.
    ctype:  class
        The component type (typically either :py:class:`Constraint` or
        :py:class:`Objective` to search for).

    """

    seen = set()
    for compdata in block.component_data_objects(
            ctype,
            descend_into=True,
            active=True):
        for var in EXPR.identify_variables(compdata.expr):
            if id(var) not in seen:
                seen.add(id(var))
                yield var
```

Alternatively, `component_data_objects()` accepts either a single ctype, or an iterable of ctypes to look for.  Seeing as you combine the results for objectives and constraints anyways, you could get it all in one pass with
```python
ComponentSet(get_vars_from_component(model, (Objective, Constraint))
```

jsiirola (approved):
I think this is fine to merge, however you might consider removing redundant code (see comment below).

pyomo/contrib/pyros/util.py
shermanjasonaf: See subsequent commit.

jsiirola (approved):

PR #2177, opened 2021-11-03
 `calculate_variable_from_constraint()` should ignore bounds/domain when updating variable value

### LINKED ISSUE #2176: calculate_variable_from_constraint can fail for simple constraints due to domain violation
## Summary

Using the `calculate_variable_from_constraint` method can result in unexpected failures due to trying to set the value of a `Var` outside of its domain, as shown in the example below. In this case, a simple constraint of `v == 0` can be successfully solved, but fails if the constraint is scaled by a factor of 10 (ie. `10*v == 0`).

```
# example.py
import pyomo.environ as pyo
from pyomo.util.calc_var_value import calculate_variable_from_constraint

# Create basic model
m = pyo.ConcreteModel()
m.v1 = pyo.Var(initialize=1, domain=pyo.NonNegativeReals)

m.c1 = pyo.Constraint(expr=m.v1 == 0)

# Calculate value of v1 using constraint c1
calculate_variable_from_constraint(m.v1, m.c1)

# This works as expected
assert pyo.value(m.v1) == 0

# However, if we multiply v1 by 10:
m.c2 = pyo.Constraint(expr=m.v1*10 == 0)
m.v1.set_value(1)

calculate_variable_from_constraint(m.v1, m.c2)

>>> ValueError: Numeric value `-9` (<class 'int'>) is not in domain NonNegativeReals for Var v1
```

This issue arises due to line 113 in `calculate_variable_from_constraint` which assumes that the constraint is linear with a coefficient of 1 w.r.t. x to quickly solve for trivial cases (such as the initial problem). However, in the case of the scaled problem this violates the domain on the `Var` causing `calculate_variable_from_constraint` to fail when ideally it would continue.

I think the first part of the solution is to wrap line 113 in a `try/except` to handle this case, but this leads to the question of what to do in the except. However, the next step in the method assumes that line 113 set the value of the variable to a new value such that there is also a new residual for the constraint, which is used to estimate the derivative w.r.t. the variable. I the except is just `pass`, then this next step fails due to a division by zero error (as the value of the variable is the same).

jsiirola:
## Summary/Motivation:
@andrewlee94 noted that `calculate_variable_from_constraint()` would raise exceptions if the intermediate values encountered by the solver fell outside the variable bounds (or were not admitted by the variable domain).  This PR updates the solver to avoid bounds/domain checking when setting the variable value.  As the method implements an unconstrined solver (which is documented in the function docstring), it should not surprise the user if the resulting variable values is outside the variable domain.

- Ignore variable domain/bound when setting the variable value in `calculate_variable_from_constraint()`
- Add tests for behavior reported in #2176

### Legal Acknowledgement

### DIFF
--- a/pyomo/util/calc_var_value.py
+++ b/pyomo/util/calc_var_value.py
@@ -28,6 +28,12 @@ def calculate_variable_from_constraint(variable, constraint,
     residual, it falls back on Newton's method using exact (symbolic)
     derivatives.
 
+    Notes
+    -----
+    This is an unconstrained solver and is NOT guaranteed to respect the
+    variable bounds or domain.  The solver may leave the variable value
+    in an infeasible state (outside the declared bounds or domain bounds).
+
     Parameters:
     -----------
     variable: :py:class:`_VarData`
@@ -52,9 +58,6 @@ def calculate_variable_from_constraint(variable, constraint,
     Returns:
     --------
     None
-
-    Note: this is an unconstrained solver and is NOT guaranteed to
-    respect the variable bounds.
 
     """
     # Leverage all the Constraint logic to process the incoming tuple/expression
--- a/pyomo/util/calc_var_value.py
+++ b/pyomo/util/calc_var_value.py
@@ -73,27 +73,30 @@ def calculate_variable_from_constraint(variable, constraint,
         raise ValueError("Constraint must be an equality constraint")
 
     if variable.value is None:
+        # Note that we use "valid=True" here as well, as the variable
+        # domain may not admit the calculated initial guesses, and we
+        # want to bypass that check.
         if variable.lb is None:
             if variable.ub is None:
                 # no variable values, and no lower or upper bound - set
                 # initial value to 0.0
-                variable.set_value(0)
+                variable.set_value(0, valid=True)
             else:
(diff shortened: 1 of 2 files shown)


andrewlee94:
Would it be a good idea to log a warning if the final result is outside the variable bounds/domain? Whilst I am fine with the solver returning an infeasible value (the documentation clearly warns about this), it would probably nice to warn the user.

At the very least, this might save you from having to field questions/complaints from users about "my model won't solve" because an invalid value was returned here that the user was unaware of.

mrmundt (approved):

Robbybp:
> Would it be a good idea to log a warning if the final result is outside the variable bounds/domain?
I would go as far as saying we should raise an exception in this case, and a knowledgeable user can catch and ignore/process it if they still want to do something with the result.

jsiirola:
@Robbybp: I am tempted to agree.  However, we do not do the same with other solvers.  For example, if a MIP solver returns 0.9999999 for a binary variable value, we (silently) accept that result.  We only raise an exception when the solver returns an "error" state.  There are a number of questions around this -- I have opened PEP #2180 to hopefully get some discussion and consensus.

blnicho (approved):

PR #2178, opened 2021-11-03
Allow `find_component` to work with `ComponentUID`

Robbybp:
## Fixes:
The following code did not work:
```python
import pyomo.environ as pyo
m = pyo.ConcreteModel()
m.v = pyo.Var()
cuid = pyo.ComponentUID("v")
m.find_component(cuid)
```
Of course, `cuid.find_component_on(m)` would still work in this example.

## Summary/Motivation:
It's convenient to use the same syntax for finding a component from either a string or CUID.

- `find_component` checks if its argument is a CUID and calls `find_component_on` appropriately
- Tests

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/block.py
+++ b/pyomo/core/base/block.py
@@ -901,7 +901,11 @@ class _BlockData(ActiveComponentData):
         """
         Return a block component given a name.
         """
-        return ComponentUID(label_or_component).find_component_on(self)
+        if type(label_or_component) is ComponentUID:
+            cuid = label_or_component
+        else:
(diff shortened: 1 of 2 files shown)


pyomo/core/base/block.py
```diff
@@ -901,7 +901,11 @@ def find_component(self, label_or_component):
         """
         Return a block component given a name.
```
jsiirola, line 902: Can you also update the docstring (and bring it to numpydoc format)?

jsiirola (approved):

mrmundt (approved):

PR #2195, opened 2021-11-16
Remove `optional_packages` from `pyomo.environ`

mrmundt:
## Summary/Motivation:
This PR removes the separation of Pyomo modules/packages into `optional` and `required.` All packages are now `required`. The existence of `optional_packages` resulted in the swallowing of errors which causes debugging concerns in those packages.

- Removal of `optional_packages`
- Fixing import statements in `TrustRegion`

### Legal Acknowledgement


pyomo/contrib/trustregion/GeometryGenerator.py
```diff
@@ -11,7 +11,10 @@
 import logging
 
 # This is an auto geometry generator for quadratic ROM
-import numpy as np
+from pyomo.common.dependencies import numpy_available
+if numpy_available:
+    import numpy as np
```
jsiirola, line 16: The simpler thing to do here is just:
```python
```
(numpy will be automatically imported the first time it is used).  That also avoids importing the dependency unless someone is actually using trustregion.

pyomo/contrib/trustregion/Logger.py
```diff
@@ -8,8 +8,11 @@
 #  This software is distributed under the 3-clause BSD License.
 #  ___________________________________________________________________________
 
-import numpy as np
-from numpy.linalg import norm
+from pyomo.common.dependencies import numpy_available
+if numpy_available:
+    import numpy as np
+    from numpy.linalg import norm
```
jsiirola, line 14: I would replace this with 
```python
```
and then used `np.linalg.norm` explicitly in the module

jsiirola (approved):
This is fine, although I would change how you are importing optional dependencies in the `trustregion` code (see comments below, but the same ideas apply to all 5 files).

blnicho (approved):

PR #2198, opened 2021-11-17
Adding option to test driver for turning off log capture

mrmundt:
## Summary/Motivation:
As part of the release process, we now check the logs to make sure that deprecation warnings are not being thrown in the core. This adds the option to show all logs rather than the default where they are swallowed.

This PR also enables any extra values that are passed to be tacked on directly to the nosetests call.

- Add `--show-log` option for Pyomo test driver
- Allow unknown arguments to be passed and added onto nosetests call

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/tests/test_unittest.py
+++ b/pyomo/common/tests/test_unittest.py
@@ -10,6 +10,7 @@
 
 import datetime
 import multiprocessing
+import os
 from io import StringIO
 import time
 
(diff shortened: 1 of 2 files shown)


blnicho (approved):

jsiirola (approved):
This is fine.

That said, I am starting to think that we should change the parser to just `parse_known_args` and pass everything that is not recognized through to nosetests.  That way we wouldn't have to update this driver every time that we wanted to use another nose argument.

mrmundt:
> This is fine.
> 
> That said, I am starting to think that we should change the parser to just `parse_known_args` and pass everything that is not recognized through to nosetests. That way we wouldn't have to update this driver every time that we wanted to use another nose argument.

I added that, too - did you not notice the update?: https://github.com/Pyomo/pyomo/pull/2198/files#diff-e212245797a79af4ba8e58cc9073a2a78210208ef622ae70dec1b277087733a1R609

PR #2209, opened 2021-11-22
NPV_SumExpression needs to inherit from NPV_Mixin

### LINKED ISSUE #2208: NPV_SumExpression does not inherit from NPV_Mixin
## Summary

`NPV_SumExpression` does not inherit form `NPV_Mixin`. As a result replacing mutable parameters in an `NPV_SumExpression` with variables does not work correctly.

```
import pyomo.environ as pe
from pyomo.core.expr.visitor import replace_expressions

m = pe.ConcreteModel()
m.x = pe.Var()
m.p = pe.Param(mutable=True)
e1 = 2 + m.p
print(type(e1))
print(e1)
e2 = replace_expressions(e1, substitution_map={id(m.p): m.x})
print(type(e2))
print(e2)
```

outputs

```
<class 'pyomo.core.expr.numeric_expr.NPV_SumExpression'>
2 + p
<class 'pyomo.core.expr.numeric_expr.NPV_SumExpression'>
2 + x
```

michaelbynum:
- `NPV_SumExpression` inherits from `NPV_Mixin`
- Unfortunately, the implementation of `create_node_with_local_data` in `NPV_Mixin` does not work for `NPV_SumExpression`. You end up with an instance of `SumExpressionBase` instead of `SumExpression` when some of `args` are potentially variable. Therefore, I reimplemented `create_node_with_local_data` for `NPV_SumExpression`. Maybe there is a way to make the `NPV_Mixin` implementation work?

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/expr/numeric_expr.py
+++ b/pyomo/core/expr/numeric_expr.py
@@ -1246,11 +1246,8 @@ class UnaryFunctionExpression(ExpressionBase):
         return self._fcn(result[0])
 
 
-class NPV_UnaryFunctionExpression(UnaryFunctionExpression):
+class NPV_UnaryFunctionExpression(NPV_Mixin, UnaryFunctionExpression):
     __slots__ = ()
-
-    def is_potentially_variable(self):
-        return False
 
 
 # NOTE: This should be a special class, since the expression generation relies
@@ -1266,6 +1263,11 @@ class AbsExpression(UnaryFunctionExpression):
 
     def __init__(self, arg):
         super(AbsExpression, self).__init__(arg, 'abs', abs)
+
+    def create_node_with_local_data(self, args, classtype=None):
+        if classtype is None:
+            classtype = self.__class__
+        return classtype(args)
 
 
 class NPV_AbsExpression(NPV_Mixin, AbsExpression):
--- a/pyomo/core/expr/numeric_expr.py
+++ b/pyomo/core/expr/numeric_expr.py
@@ -958,7 +958,7 @@ class SumExpressionBase(_LinearOperatorExpression):
         return 'sum'
 
 
-class NPV_SumExpression(SumExpressionBase):
+class NPV_SumExpression(NPV_Mixin, SumExpressionBase):
     __slots__ = ()
 
     def create_potentially_variable_object(self):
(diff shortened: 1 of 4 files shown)


pyomo/core/expr/numeric_expr.py
```diff
@@ -975,8 +975,22 @@ def _to_string(self, values, verbose, smap, compute_values):
...
+        try:
+            npv_args = all(
+                type(arg) in native_types or not arg.is_potentially_variable()
+                for arg in args
+            )
+        except AttributeError:
+            # We can hit this during expression replacement when the new
+            # type is not a PyomoObject type, but is not in the
+            # native_types set.  We will play it safe and clear the NPV flag
+            npv_args = False
```
jsiirola, line 989: I understand the logic of this, but throughout the expression code we assume that (apart from in the `generate_*` methods) if the argument is not in `native_types`, then it adheres to the `NumericValue` API.  Instead of trapping for this here, would it be better to process the argument to the expression replacement visitor method?

jsiirola (commented):
`NPV_Mixin` was designed for the NPV version of an expression node to derive from the "regular" version.  `SumExpression` and `NPV_SumExpression` don't follow that convention (and until that chances, your custom implementation of `create_node_with_local_data` will be necessary.  

I think it makes sense to accept this for the time being with the intention to eventually rework `SumExpression` so that it follows the convention of the other `NPV_*` classes.

pyomo/core/expr/numeric_expr.py
michaelbynum: I honestly just copied this from the `NPV_Mixin` class. I am certainly fine with this change. Should I change the implementation in `NPV_Mixin`?

michaelbynum:
Thanks @jsiirola. That makes sense. I completely agree.

michaelbynum:
I'm going to add some tests too.

michaelbynum:
While adding tests, I also modified many of the tests in `test_visitor` to use `compare_expressions` instead of comparing the strings of expressions.

jsiirola (approved):

pyomo/core/expr/numeric_expr.py
jsiirola: Ahh yes.  I remember that logic now.  I vote we leave this as-is, although it might be worthwhile to add a note that in a future rework of the `SumExpression` internals we should reorder the class hierarchy to remove the need for that override.

PR #2212, opened 2021-11-24
Overload initializer of basic component classes with named keyword params to enable code hinting.

### LINKED ISSUE #2145: Stop using *args and **kwargs in function definition and add type annotations for the code
## Summary

1. Most basic classes in Pyomo, including Var, Set, Param, Block etc.,  use  `*args` and `**kwargs` to pass all parameters in their constructor function. This prevents the IDE from giving hints and checking for error.  Please replace `*args` and `**kwargs` with named positional and optional parameters.

2. Static type checking tools are popular and useful, but they cannot work properly with Pyomo now. Please add type annotations for Pyomo.

### Rationale
The parameters for the basic Pyomo classes do not change frequently. It seems unnecessary to use `**kwargs` to pass basic parameters and unnecessary to use `*args` except for multiple indexes.

**Example for `**kwargs`**. In class Var, the total number of parameters is not large and these parameters are frequently used. I think it's better to replace **kwd with named optional parameters.
``` python
class Var(IndexedComponent, IndexedComponent_NDArrayMixin):
    def __init__(self, *args, **kwd):
        #
        # Default keyword values
        #
        initialize = kwd.pop('initialize', None)
        initialize = kwd.pop('rule', initialize)
        domain = kwd.pop('within', Reals)
        domain = kwd.pop('domain', domain)
        bounds = kwd.pop('bounds', None)
        self._dense = kwd.pop('dense', True)
        self._units = kwd.pop('units', None)
```
**Example for `*args`.** In `RangeSet`, `start`, `end`, `step` are passed with `*args`. I think overloading the constructor may be a better way. Take the Python basic class  `range`  for Example.
``` python
class RangeSet(Component):
    Parameters
    ----------
    *args: int | float | None
        The range defined by ([start=1], end, [step=1]).  If only a
        single positional parameter, `end` is supplied, then the
        RangeSet will be the integers starting at 1 up through and
        including end.  Providing two positional arguments, `x` and `y`,
        will result in a range starting at x up to and including y,
        incrementing by 1.  Providing a 3-tuple enables the
        specification of a step other than 1.

    finite: bool, optional
        This sets if this range is finite (discrete and bounded) or infinite

    ranges: iterable, optional
        The list of range objects that compose this RangeSet

    bounds: tuple, optional
        The lower and upper bounds of values that are admissible in this
        RangeSet

    filter: function, optional
        Function (rule) that returns True if the specified value is in
        the RangeSet or False if it is not.

    validate: function, optional
        Data validation function (rule).  The function will be called
        for every data member of the set, and if it returns False, a
        ValueError will be raised.
    def __init__(self, *args, **kwds):
        # Finite was processed by __new__
        kwds.setdefault('ctype', RangeSet)
        if len(args) > 3:
            raise ValueError("RangeSet expects 3 or fewer positional "
                             "arguments (received %s)" % (len(args),))
        kwds.pop('finite', None)
        self._init_data = (
            args,
            kwds.pop('ranges', ()),
        )
        self._init_validate = Initializer(kwds.pop('validate', None))
        self._init_filter = Initializer(kwds.pop('filter', None))
        self._init_bounds = kwds.pop('bounds', None)
        if self._init_bounds is not None:
            self._init_bounds = BoundsInitializer(self._init_bounds)
````

### Description

1. Replace `*args` and `**kwargs` with named positional and optional parameters.
2. Add type annotations for Pyomo.

### Additional information
Thank you for building Pyomo.

### PR DESCRIPTION
**Pull request by shishouyuan**
## Summary/Motivation:
Most basic classes in Pyomo, like `Var`, `Set`, `Param`, etc., use `*args` and `**kwargs` to pass all parameters in their initializer. There are several disadvantages with this way.
- Users cannot take the advantage of code hints provided by the IDE to make the coding process more fluently.
- Users have to pay attention to the correctness of keyword spelling, thus got distracted from model building. 
- Keyword name spelling errors can only be found at runtime, thus it may be tedious to locate all of them.

I think this issue is a legacy from the early version of Pyomo with Python 2. As stated in [PEP 3102 -- Keyword-Only Arguments](https://www.python.org/dev/peps/pep-3102/) that, except `**kwargs`, keyword arguments after `*args` were not supported at that time. But now since Pyomo only supports Python 3.6 or newer , the issue can be addressed.

To avoid introducing bugs into the code,  I choose to add overloads for the initializer of basic component classes. Because the overloads will be replaced by the original initializer right behind, the runtime behavior will not be affected. Here are the examples.
### Class `Var`
```python
@overload
def __init__(self, *indexes, domain=..., within=..., bounds=...,
             initialize=..., rule=..., dense=..., units=...): ...    

# the original initializer 
def __init__(self, *args, **kwargs):
```
### Class `RangeSet`
```python
# `start`, `end`, `step` in `*args` are positional-only that cannot be filled with keywords. 
# But positional-only params syntax are not supported before python 3.8.
# To emphasize they are positional-only, an underscore is added before their name. 
@overload
def __init__(self, _end, *, finite=..., ranges=...,
             bounds=..., filter=..., validate=...): ...

@overload
def __init__(self, _start, _end, _step=1, *, finite=..., ranges=...,
             bounds=..., filter=..., validate=...): ...

@overload
def __init__(self, *, finite=..., ranges=...,
             bounds=..., filter=..., validate=...): ...

# the original initializer 
def __init__(self, *args, **kwds):
```

With these overloads, hinting for parameter names becomes available, and error messages will be displayed by the editor if keywords are not valid (tested with Visual Studio Code).

- Overloaded the initializers of class `Set`, `RangeSet`,  `Var`, `Param`, `Constraint`, `Expression`, `Objective`.
- Deleted `index` item in the docstring of class `Param` for 3 reasion. 1) The `index` item was put at 2nd position which may be perceived as a keyword param. 2) There is no keyword param called `index`  for `Param`. 3) Other classes do not doc the `index` param.

### Legal Acknowledgement


pyomo/core/base/expression.py
```diff
@@ -254,6 +255,9 @@ def __new__(cls, *args, **kwds):
         else:
             return IndexedExpression.__new__(IndexedExpression)
 
+    @overload
+    def __init__(self, *indexes, rule=..., expr=..., initialize=...): ...
```
jsiirola, line 259: Missing `name` and `doc`

pyomo/core/base/param.py
```diff
@@ -259,6 +257,11 @@ def __new__(cls, *args, **kwds):
         else:
             return super(Param, cls).__new__(IndexedParam)
 
+    @overload
+    def __init__(self, *indexes, rule=..., initialize=...,
+                 domain=..., within=..., validate=..., mutable=...,
+                 default=..., initialize_as_dense=..., units=..., name=...): ...
```
jsiirola, line 263: Missing `doc`

pyomo/core/base/var.py
```diff
@@ -594,6 +595,10 @@ def __new__(cls, *args, **kwargs):
         else:
             return super(Var, cls).__new__(IndexedVar)
 
+    @overload
+    def __init__(self, *indexes, domain=..., within=..., bounds=...,
+                 initialize=..., rule=..., dense=..., units=...): ...    
+
```
jsiirola, line 601: Missing `name` and `doc`

jsiirola (changes requested):
This is a really nice solution.

There are a couple missing keywords (mostly `name` and `doc`; see inline comments).  `ctype` is a valid keyword that is universally missing; however, that is not really a "user option", so I agree that it should knot be listed in the `@overload` (which is really there for user documentation).

In addition, I think we should make this complete and include the default values and type hints where we can (instead of having the default values be `Ellipsis`).

shishouyuan:
@jsiirola Thanks for the advices. Please review the new commits.

- I have added `name` and `doc` for all of the classes. 
- Default values are included in the overloads.
- New overloads for `Block` and `Suffix` are added.

I agree that it is a good idea to add type hints for Pyomo, but it may be a big project. Since improper type hints may raise a number of warnings to the user and I don't have a thorough understand of  Pyomo🤣, I didn't add type annotations in the above commits. Maybe we can start a new project called something like `pyomo-stubs` to provide `pyi` files as Gurobi does.

pyomo/core/base/expression.py
shishouyuan: finised

pyomo/core/base/param.py
shishouyuan: finished

pyomo/core/base/var.py
shishouyuan: finished

jsiirola (changes requested):
This is really close.  I realize that type hints are tricky, and for this PR, let's just leave them out.

One general comment:
- We should probably have the default `name` be `None`, as the default name for unattached components is the component type (e.g., `IndexedBlock` or `ScalarBlock`).

shishouyuan:
@jsiirola Thanks. All the default values of `name` have been changed to `None`.

jsiirola (approved):

blnicho (approved):

PR #2214, opened 2021-11-30
Switch `Var.value` setter to validate values

### LINKED ISSUE #2180: [PEP] Define standards for enforcing variable bounds / domains
## Summary

Currently, Pyomo is inconsistent in how strictly it applies / enforces variable domains and bounds:
- Variable domains are checked when the value is set using either assignment (`m.x = 5`) or `set_value()` (`m.x.set_value(5)`)
  - Failed checks raise an exception
- Variable domains are NOT checked when directly setting the `value` attribute (`m.x.value = 5`)
- Variable bounds are *never* checked when setting a value
   - except if the variable bounds are set by the domain (and then the bounds are implicitly checked when checking the domain)

Further, different clients use different APIs:
- Variable construction uses `set_value()` (and thus checks the variable domain)
- Solvers generally return the values from the solver using `.value` (and thus do NOT check the domain)

## Proposal:

We should be more consistent when checking variable domains / bounds:
- **Proposal 1**: when we check either the domain or the bounds, we should check both the domain *and* the bounds
- Rationale: as bounds information comes from the tighter of the domain bounds and the user bounds, it is misleading to check one without checking the other

We should decide how "egregious" violating the domain or bounds is (generally, one of {*ignored*, *info*, *warning*, or *exception*})
- **Proposal 2**: Violating the domain or bounds should always generate a user warning (through the local logger) whenever the value will be left in the variable and control returned to the user.
- Rationale: There are perfectly rational algorithmic reasons to take an infeasible path (e.g., Newton's method when solving for a variable value from a constraint, or an interior point algorithm may take steps outside the domain / bounds.  That should not cause the algorithm to halt, nor should it pollute the user's screen with warnings.  However, if an algorithm (e.g., solver) leaves a variable in a state outside its bounds / domain, then a user warning should be emitted.  This should not generate an exception, either, as there are perfectly rational reasons for an algorithm to terminate at a point not in the bounds (terminating at point of local infeasibility, using an unconstrained solver for a bounded problem, using a continuous solver for a discrete problem).  Raising an exception when setting the variable value could cause control to return to the user prematurely.

We should be more consistent with our checking, but allow users that "know better" to bypass it.
- **Proposal 3**: Both `m.x.value = 5` and `m.x.set_value(5)` should validate the domain and bounds.  Users that wish to bypass validation would call `m.x.set_value(5, skip_validation=True)`
- Rationale: We should allow users who know that they are going to adhere to the principle of the previous proposal to bypass the domain and bound checking in the name of code performance.  However, that should be a *conscious and intentional decision* (i.e., they need to make it explicit and shouldn't end up bypassing validation just because they accidentally copied an example that was setting `.value`).

## Changelog

- 8 Nov: Added third proposal to make the interface for bypassing validation require an explicit argument
- 10 Nov: Labeled the three proposals to make discussion easier
- 16 Nov: change the validation flag to `skip_validation` to match the sense of the current `valid` flag.

jsiirola:
## Summary/Motivation:
This completes part 3 in PEP #2180 by switching the `Var` `value` setter to perform validation of incoming values.  To preserve existing behavior, previous uses of the `value` setter (e.g., in solver interfaces) have been switched over to use `Var.set_value(skip_validation=True)`.

This also adds functionality to the Pyomo logger so that messages can be logged with an identifier that references extended documentation in the Online Docs.

- Switch `Var.value` and `BooleanVar.value` setters to perform value validation
- Convert uses of the `value` setter in performance-sensitive portions of Pyomo (e.g., solver interfaces) to use `set_value(skip_validation=True)`
- Add processing of an `extra={'id': 'Wxxxx'}` argument for logged Pyomo warnings and errors, which adds a link to extended descriptions in RTD

### Legal Acknowledgement


pyomo/core/kernel/variable.py
```diff
@@ -385,6 +385,9 @@ def value(self):
     def value(self, value):
         self._value = value
 
+    def set_value(self, value, skip_validation=True):
+        self.value = value
```
michaelbynum, line 389: This is needed so the solver interfaces work with kernel?

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -1306,7 +1306,7 @@ def check_expr_evaluation(model, symbolMap, solver_io):
     finally:
         # Return uninitialized variables to None
         for var in uninit_vars:
-            var.value = None
+            var.set_value(None)
```
michaelbynum, line 1309: Should this not use `skip_validation=True`?

michaelbynum (commented):
The PR looks good. I have a couple comments/questions below.

This PR causes 

```
import pyomo.environ as pe
m = pe.ConcreteModel()
m.x = pe.Var(domain=pe.Binary)
m.x.value = 0.5
```

to generate a warning. Is that desired behavior? I think so, but I want to make sure. If someone does not want the warning, they should use `set_value`. Although, maybe we should add a note in the warning message that `set_value` is an option. It may not be obvious that there is another mechanism for setting values.

pyomo/solvers/plugins/solvers/GAMS.py
michaelbynum: Got it.

jsiirola:
@michaelbynum, I think updating the warning with a note about using `set_value` would be fine.  An alternative is to have a shorter warning message with a reference to the online docs where we can have a longer discussion (this is an idea I have been mulling over, in part to address the ever-growing length of our warnings/error messages).

I have also toyed with the idea of having a global flag that controls whether you get a warning or the previous behavior of an exception.  I have been fighting with Param (to do the same kind of thing and centralize the processing of new values), and noted that Param still tosses exceptions (and in general, I am OK with the Param tossing an exception, even if Var moves to warnings; but, it might be nice if this were configurable).  Thoughts?

michaelbynum:
I think a reference to online docs is a great idea. That would enable us to give more context.

A global flag could be problematic. Library A might set the flag one way while library B might set the flag another way. If both libraries were used in library C, unexpected things might happen. If we had a way to ensure the flag could only be set once, then it might work. It could certainly be useful.

michaelbynum (approved):

michaelbynum:
This is awesome, @jsiirola. Thank you. I love this solution.

PR #2222, opened 2021-12-04
Define an `as_quantity()` to evaluate a Pyomo expression to a pint `Quantity`

jsiirola:
## Summary/Motivation:
This PR adds a `as_quantity()` function that will accept numeric-like objects (int / float / None / Var / Param / expression / etc) and return the value as a native `pint` `Quantity` (value + units).

This is useful for applications like user interfaces, and was requested by IDAES (@andrewlee94 and @dangunter)

- add `as_quantity()` function to evaluate an expression to a pint Quantity
- add tests (including expanded coverage of `units_container`)

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/units_container.py
+++ b/pyomo/core/base/units_container.py
@@ -117,6 +117,7 @@ from pyomo.core.expr.numvalue import (
     native_numeric_types, pyomo_constant_types,
 )
 from pyomo.core.expr.template_expr import IndexTemplate
+from pyomo.core.expr.visitor import ExpressionValueVisitor
 from pyomo.core.expr import current as EXPR
 
 pint_module, pint_available = attempt_import(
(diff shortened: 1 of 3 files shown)


blnicho (approved):

pyomo/core/base/units_container.py
```diff
@@ -1300,6 +1301,87 @@ def pint_registry(self):
...
+
+_QuantityVisitor.handlers = {
+    EXPR.UnaryFunctionExpression: _QuantityVisitor._handle_unary_function,
+    EXPR.NPV_UnaryFunctionExpression: _QuantityVisitor._handle_unary_function,
+    EXPR.ExternalFunctionExpression: _QuantityVisitor._handle_external,
+    EXPR.NPV_ExternalFunctionExpression: _QuantityVisitor._handle_external,
+}
+
+def as_quantity(expr):
+    return _QuantityVisitor().dfs_postorder_stack(expr)
```
mrmundt, line 1382: Just a comment, not a required change - should we standardize this at some point to change all instances that use `dfs_postorder_stack` to `walk_expression`?

mrmundt (approved):

pyomo/core/base/units_container.py
jsiirola: Eventually? Yes.  But that will require reworking the `ExpressionValueVisitor` to be based on the `StreamBasedExpressionVisitor`.

pyomo/core/base/units_container.py
```diff
@@ -359,11 +360,11 @@ def to_string(self, verbose=None, labeler=None, smap=None,
...
         """
-        if len(self._pint_unit.dimensionality) > 1 or any(
-                i < 0 for i in self._pint_unit.dimensionality.values()):
-            return "("+str(self)+")"
+        _str = str(self)
+        if any(map(_str.__contains__, ' */')):
+            return "(" + _str + ")"
         else:
-            return str(self)
+            return _str
```
carldlaird, line 367: Is this mostly for debugging, or is it used somewhere else?

pyomo/core/base/units_container.py
```diff
@@ -1300,6 +1301,87 @@ def pint_registry(self):
...
+            return True, node
+
+        if node.is_expression_type():
+            return False, None
+
+        if node.is_numeric_type():
+            if hasattr(node, 'get_units'):
+                unit = node.get_units()
+                if unit is not None:
+                    return True, value(node) * unit._pint_unit
```
carldlaird, line 1335: I don't see how this works. If a Var has units that are themselves an expression (e.g., kg*m), then that expression won't have "_pint_unit"?

carldlaird: After discussions with @jsiirola I guess we create compound units instead of keeping the expression. Let's make sure there are tests that include Vars with units that are expressions.

carldlaird (approved):
This looks reasonable to me.

carldlaird (commented):
This causes a broader question for the team. This PR is officially exposing pint from behind the Pyomo units. Do we want to do this?

I have no real concerns, however, we did quite a bit of work to make sure that pint was abstracted away (so we could replace pint with something else later if we wanted to). This decision may have been more about being unfamiliar with pint and not "ready to commit" to it. If we are ready to commit to pint, then there may be other simplifications we can make.

Just thoughts for discussion.

blnicho:
@carldlaird we discussed your question at the developers meeting this week and the consensus was that we didn't have any concerns with exposing more of pint. Given the currently available alternatives to pint we think the likelihood of replacing it with something else is pretty slim.

PR #2223, opened 2021-12-06
Construct slices with normalized indices in slicing utility

Robbybp:
## Summary/Motivation:
`pyomo.util.slices` contains a function for slicing a component data object along sets in itself or any of its parent blocks. The implementation leads to "unexpected" CUID representations of the slices, as indices are always converted to tuples. For instance, the following code:
```python
import pyomo.environ as pyo
from pyomo.util.slices import slice_component_along_sets
m = pyo.ConcreteModel()
m.s1 = pyo.Set(initialize=["a", "b"])
m.s2 = pyo.Set(initialize=["c", "d"])
m.b = pyo.Block(m.s1)
for i in m.s1:
    m.b[i].v = pyo.Var(m.s2)
slice_ = slice_component_along_sets(m.b["a"].v["c"], ComponentSet((m.s1,)))
print(cuid)
```
prints `b[*].v[(c,)]`. This is nonintuitive and makes these slices incompatible with those generated by other means, e.g. `flatten_dae_components`.

- Normalize indices given to the slice in `slice_component_along_sets`. Note that this function is currently not compatible with `normalize_index.flatten == False`
- Allow `slice_component_along_sets` to safely accept a list or tuple of sets as an argument. Previously, this could silently do the wrong thing.

### Legal Acknowledgement


jsiirola (approved):
This looks reasonable.  One question: if the function is not compatible with `normalize_indices == False`, should you check and raise an error if that is the case?

Robbybp:
@jsiirola For better or worse, the error is raised in `get_location_set_map` rather than `slice_component_along_sets`, as the former function is where the assumption that `normalize_index.flatten == False` violates is made.

PR #2225, opened 2021-12-07
Rework coverage uploads to use GitHub Action

mrmundt:
## Summary/Motivation:
The bash uploaded for coverage reports is deprecated. After a break in the system yesterday, we resolved to switch to using the better supposed GitHub action `codecov-action`.

- Switch from codecov bash uploaded to `codecov-action` uploader

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/test_branches.yml
+++ b/.github/workflows/test_branches.yml
@@ -697,17 +697,15 @@ jobs:
             cp -v $ARTIFACT/.coverage .coverage-$NAME
         done
         rm -vf .coverage coverage.xml
-        if [ ${{ matrix.TARGET }} == 'linux' ]; then
-          echo "Build: linux/other"
-          echo ""
-          FILES=.coverage-*_other-*
-          coverage combine --debug=dataio $FILES
-          if test ! -e .coverage; then
-              echo "No coverage to upload."
-          else
-              coverage xml || coverage xml -i
-              mv -v coverage.xml coverage-other.xml
-          fi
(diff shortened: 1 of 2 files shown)


.github/workflows/test_branches.yml
```diff
@@ -692,71 +692,49 @@ jobs:
...
+          echo "Build: linux/other"
+          echo ""
+          FILES=.coverage-*_other-*
+          coverage combine --debug=dataio $FILES
+          if test ! -e .coverage; then
+              echo "No coverage to upload."
+          else
+              coverage xml || coverage xml -i
+              mv -v coverage.xml coverage-other.xml
+          fi
```
jsiirola, line 710: Can we do this for all platforms?  Right now we only use Linux for "other", but that may not always be the case.

.github/workflows/test_branches.yml
```diff
@@ -692,71 +692,47 @@ jobs:
...
+      uses: codecov/codecov-action@v2
+      with:
+        files: coverage.xml
+        name: ${{ matrix.TARGET }}
+        flags: ${{ matrix.TARGET }}
+
+    - name: Upload other coverage reports
+      if: |
+        matrix.TARGET  == 'linux' &&
+        (github.repository_owner == 'Pyomo' || github.ref != 'refs/heads/main')
```
jsiirola, line 733: I think a cleaner implementation of this might be:
```
if: hashFiles('coverage-other.xml') != ''
```

jsiirola (approved):

PR #2232, opened 2021-12-14
Adding support for abs in numeric differentiation

michaelbynum:
## Summary/Motivation:
This PR adds support for `abs` in Pyomo's numeric differentiation. Attempting symbolic differentiation on an `abs` expression with a non-constant argument will still raise an exception.

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/tests/unit/test_derivs.py
+++ b/pyomo/core/tests/unit/test_derivs.py
@@ -10,7 +10,7 @@
 
 import pyomo.common.unittest as unittest
 import pyomo.environ as pyo
-from pyomo.core.expr.calculus.diff_with_pyomo import reverse_ad, reverse_sd
+from pyomo.core.expr.calculus.diff_with_pyomo import reverse_ad, reverse_sd, DifferentiationException
 from pyomo.common.getGSL import find_GSL
 from pyomo.core.expr.numeric_expr import LinearExpression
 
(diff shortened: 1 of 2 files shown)


pyomo/core/tests/unit/test_derivs.py
```diff
@@ -172,6 +172,21 @@ def test_atan(self):
         self.assertAlmostEqual(derivs[m.x], pyo.value(symbolic[m.x]), tol+3)
         self.assertAlmostEqual(derivs[m.x], approx_deriv(e, m.x), tol)
 
+    def test_abs(self):
+        m = pyo.ConcreteModel()
+        m.x = pyo.Var(initialize=2.0)
+        e = 2 * abs(m.x)
+        with self.assertRaises(DifferentiationException):
```
jsiirola, line 179: In general, I prefer `assertRaisesRegex()` so that we check that the correct message is being raised (I have run across bugs where we thought we were testing that Pyomo correctly handled an exception only to find that there was a bug that happened to raise the same exception).

michaelbynum: Good point. Fixed.

pyomo/core/tests/unit/test_derivs.py
```diff
@@ -176,15 +176,15 @@ def test_abs(self):
         m = pyo.ConcreteModel()
         m.x = pyo.Var(initialize=2.0)
         e = 2 * abs(m.x)
-        with self.assertRaises(DifferentiationException):
+        with self.assertRaisesRegexp(DifferentiationException, 'Cannot perform symbolic differentiation of abs\(x\)'):
```
jsiirola, line 179: The reference string needs to be declared as a raw string to avoid errors in recent Python versions.  (I will fix that shortly)

jsiirola (approved):

Issue #2160, opened 2021-10-21
Add API/Getter for Variable Bounds Expressions


shermanjasonaf:
## Summary

API/Getter for variable bounds expression to circumvent use of the protected `._lb` and `._ub` attributes of variable objects.

### Rationale

The module `pyomo.contrib.pyros` may require access to variable bounds expressions to determine whether these bounds are contingent on uncertain parameters, and consequently, to construct separation problem objective expressions which contain variable objects meant to simulate these uncertain parameters (as opposed to just the numerical values of the parameters). Currently, the only known way to access these variable bound expressions is through the protected `._lb` and `._ub` attributes.

### Description

See summary.

PR #2236, opened 2021-12-16
Update kernel.constraint to match Constraint (lb/lower, ub/upper) API

jsiirola:
## Summary/Motivation:
#2073 updates the `Constraint` API to provide two sets of methods to retrieve constraint bounds:
- `lb`, `ub` to retrieve the numeric value of the bound
- `lower`, `upper` to retrieve the bound expression

This PR updates kernel's `constraint` to match this API

- add `lower` and `upper` attributes
- switch `lb` and `ub` to return the current numeric value of the bound
- update tests

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/kernel/conic.py
+++ b/pyomo/core/kernel/conic.py
@@ -97,13 +97,23 @@ class _ConicBase(IConstraint):
         return self._body
 
     @property
+    def lower(self):
+        """The expression for the lower bound of the constraint"""
+        return None
+
+    @property
+    def upper(self):
+        """The expression for the upper bound of the constraint"""
+        return 0.0
+
+    @property
     def lb(self):
-        """The lower bound of the constraint"""
+        """The value of the lower lower bound of the constraint"""
         return None
 
     @property
     def ub(self):
-        """The upper bound of the constraint"""
+        """The value of the lower upper bound of the constraint"""
         return 0.0
 
     @property
--- a/pyomo/core/kernel/conic.py
+++ b/pyomo/core/kernel/conic.py
@@ -108,12 +108,12 @@ class _ConicBase(IConstraint):
 
     @property
     def lb(self):
-        """The value of the lower lower bound of the constraint"""
+        """The value of the lower bound of the constraint"""
         return None
 
     @property
     def ub(self):
-        """The value of the lower upper bound of the constraint"""
+        """The value of the upper bound of the constraint"""
         return 0.0
 
     @property
--- a/pyomo/core/kernel/constraint.py
+++ b/pyomo/core/kernel/constraint.py
@@ -199,7 +199,7 @@ class _MutableBoundsConstraintMixin(object):
     @property
     def lb(self):
         """The value of the lower bound of the constraint"""
-        return None if self._lb is None else value(self._lb)
+        return value(self._lb)
     @lb.setter
     def lb(self, lb):
         self.lower = lb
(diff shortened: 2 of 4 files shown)


pyomo/core/kernel/conic.py
```diff
@@ -96,14 +96,24 @@ def body(self):
...
+
+    @property
+    def upper(self):
+        """The expression for the upper bound of the constraint"""
+        return 0.0
+
     @property
     def lb(self):
-        """The lower bound of the constraint"""
+        """The value of the lower lower bound of the constraint"""
```
blnicho, line 111: Typo: "lower lower"

pyomo/core/kernel/conic.py
```diff
@@ -96,14 +96,24 @@ def body(self):
...
         return None
 
     @property
     def ub(self):
-        """The upper bound of the constraint"""
+        """The value of the lower upper bound of the constraint"""
```
blnicho, line 116: Typo: "lower upper"

blnicho (approved):
I found 2 typos but otherwise this looks fine.

ghackebeil:
This seems like a pretty significant and confusing change. What value does this add over calling `pmo.value(constraint.lb)`?

jsiirola:
It is a significant change.  The reason is to bring the `kernel.constraint` API back in sync with the `base.constraint` API.  Originally, `base.constraint` implemented `lower` and `upper`, which returned the expressions for the lower/upper bounds.  `kernel.constraint` did not implement `lower` and `upper`, and instead implemented `lb` an `ub` to do the same thing.  The `base` implementation was confusing with `Var`, where `Var` did not implement `lower` and `upper`, and instead implemented `lb` and `ub`, which returned the current bound value (as a float).  Two PRs (#2073 and #2160) in `core.base` attempted to reconcile the `Var` and `Constraint` APIs by preserving `lb` and `ub` as the way to get the bound value and `lower` and `upper` to get the bound expression.  This had the benefit that it was backwards compatible with both `Var` and `Constraint`'s original implementations.

Unfortunately, that API was never propagated into kernel, and the inconsistency was starting to lead to issues with other core Pyomo code that was build against the `base` API.  This PR resolves that inconsistency so that model manipulation code does not need to have special checks for `kernel` components (or conversely, does not need unnecessary calls to `value()` to resolve expressions)

blnicho:
@ghackebeil given John's reply to your question, are you alright with the changes in this PR?

ghackebeil:
The kernel interface was written so that expressions/values/whatever you assign to `.lb` / `.ub` are what you get when you access them (for variables or constraints), and an exception will be raised if you try to assign something that is not valid. The `.lower` and `.upper` properties were put on constraint simply for backwards compatibility with the solver interfaces. E.g.,
```python
x = pyomo.constraint() # pmo.variable()
lb = x.lb = pmo.parameter(1)
print(x.lb) # <parameter>
print(lb) # <parameter>
```
It looks like I missed someone changing the behavior for kernel variable in the 6.2 release, so I already have simple functions that are broken due to assuming the above logic holds for both variable and constraint. My preference would be to revert the change to kernel variable so that both it and constraint retain the above behavior (from the 6.1 release).

Does no one else find it confusing that if I assign something to `.lb`, I have to access `.lower` to get that thing back?

michaelbynum:
I agree that it is confusing. However, I think backwards compatibility has backed us into a corner on the non-kernel side because the domain and bounds are not independent. I think the approach taken on the non-kernel side is a reasonable solution. The kernel design where the domain is only used at construction time makes a lot of sense to me. Unfortunately, I don't think that is possible for non-kernel variables. I guess the question is whether or not it is okay for the non-kernel and kernel interfaces to drift. My guess is that we will encounter problems in the solver interfaces if we let them drift too much. Thoughts?

mrmundt:
This PR has been tracked in issue #2402

PR #2237, opened 2021-12-17
Remove hard-coded project name from default `attempt_import()` message

jsiirola:
## Summary/Motivation:
This removes a hard-coded package name (`Pyomo`) from the error message generated by unavailable modules imported through `attempt_import()`.  The package name is now automatically generated.  This produces more informative user messages when projects outside of Pyomo leverage `attempt_import()` (like IDAES).

- calculate package name from `attempt_import()` calling frame 
- Cache the import of the version parser (performance improvement)
- Expand test coverage

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/tests/test_dependencies.py
+++ b/pyomo/common/tests/test_dependencies.py
@@ -221,7 +221,8 @@ class TestDependencies(unittest.TestCase):
 
     def test_import_exceptions(self):
         mod, avail = attempt_import('pyomo.common.tests.dep_mod_except',
-                                    defer_check=True)
+                                    defer_check=True,
+                                    only_catch_importerror=True)
         with self.assertRaisesRegex(ValueError, "cannot import module"):
             bool(avail)
         # second test will not re-trigger the exception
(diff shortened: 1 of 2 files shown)


pyomo/common/tests/test_dependencies.py
```diff
@@ -233,6 +234,20 @@ def test_import_exceptions(self):
         self.assertFalse(avail)
         self.assertFalse(avail)
 
+        mod, avail = attempt_import('pyomo.common.tests.dep_mod_except',
+                                    defer_check=True,
+                                    catch_exceptions=(ImportError, ValueError))
+        self.assertFalse(avail)
+        self.assertFalse(avail)
```
mrmundt, line 241: Repeated assert?

mrmundt (approved):

pyomo/common/tests/test_dependencies.py
jsiirola: Yes.  One of the magic things that attempt_import does is that the returned objects actually replace themselves with the resolved values when the deferred resolution actually happens.  This implicitly checks both the original return value, and the repeated return value.

PR #2240, opened 2021-12-31
Add `remove_nodes` method to IncidenceGraphInterface

Robbybp:
## Summary/Motivation:
When debugging a model that has an unexpected overconstrained or underconstrained subsystem, a common task we want to do is remove a particular variable and constraint from the incidence graph/matrix, the equivalent of fixing a variable and deactivating a constraint. This can help narrow down the "cause" of the unexpected overconstrained/underconstrained system. This is currently possible by fixing/deactivating, then re-constructing the incidence graph, but would be nice to do this without altering the Pyomo model. This motivates the addition of a method on `IncidenceGraphInterface` for removing nodes from an incidence graph. Note this is "projecting" variables and constraints out of the model, rather than doing some sort of vertex elimination (where new edges/nonzeros would be added to the graph). 

- Add a `remove` method to `IncidenceGraphInterface` that removes columns and rows corresponding to provided variables and constraints from the cached incidence matrix
- Basic tests for the new method

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/incidence_analysis/interface.py
+++ b/pyomo/contrib/incidence_analysis/interface.py
@@ -353,7 +353,7 @@ class IncidenceGraphInterface(object):
         rows) from the cached incidence matrix. This is a "projection"
         of the variable and constraint vectors, rather than something
         like a vertex elimination.
-        For the puropse of this method, there is not need to distinguish
+        For the puropse of this method, there is no need to distinguish
         between variables and constraints. However, we provide the
         "constraints" argument so a call signature similar to other methods
         in this class is still valid.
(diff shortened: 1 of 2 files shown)


pyomo/contrib/incidence_analysis/interface.py
```diff
@@ -346,3 +346,44 @@ def dulmage_mendelsohn(self, variables=None, constraints=None):
...
+            ConstraintData objects whose rows will be removed from the
+            incidence matrix.
+
+        """
+        if self.cached is IncidenceMatrixType.NONE:
+            raise RuntimeError(
+                "Attempting to remove variables and constraints from cached "
+                "incidence matrix,\nbut no incidence matrix has been cached."
+            )
+        to_exclude = ComponentSet(variables + constraints)
```
jsiirola, line 372: This assumes that `variables` and `constraints` are "addable" sequences.  If one is a tuple and one is a list, this will error.  If either is a generator, then it will error.  Consider:
```python
to_exclude = ComponentSet(variables)
to_exclude.update(constraints)
```

jsiirola (commented):
I think this is OK, but there is a small change that you can make to make the API a little more flexible,

jsiirola (approved):

PR #2245, opened 2022-01-07
Appsi Updates

### LINKED ISSUE #2094: Add support for the HiGHS solvers via scipy
## Summary

scipy has started to vendor the HiGHS solvers. Providing a Pyomo solver based on this interface would give clean access to high-performance solvers, unencumbered by problematic licenses.

### Description

See https://github.com/scipy/scipy/issues/14455

michaelbynum:
- Improvements to the "persistent" part of Appsi
    - Better handling of fixed variables. Previously, appsi did not allow fixed variables to change the structure of the constraint. For example, the constraint `z==x + log(y)` could not be written to an lp file even if `y` was fixed. This now works. There is also an option to control this behavior for performance. By default, when a fixed variable is unfixed, any constraints it is involved in gets removed and added again. If you know the structure of the problem will not change (e.g., `z==2*x + 3*y` where `y` is fixed), then there is an option to keep the variable in the problem and fix it (e.g., with bounds - but that is solver dependent).
    - Performance improvements
    - Adds support for updating variable bounds with mutable parameters or other fixed expressions
    - Adds support for trig and log10 operators
- Added an appsi interface to HiGHS. I created a thin wrapper around the `Highs` c++ class to enable efficiently modifying the model. I also added HiGHS to GHA. It takes about 6 minutes to build, so we may only want it for one OS and one Python version. Currently, the appsi interface does not support quadratic objectives, but I'll add that soon.
- Reimplemented `pyomo.contrib.fbbt` in c++ with a persistent interface through appsi. Based on preliminary testing, this is about 30 times faster. I added lots of tests to ensure that `pyomo.contrib.fbbt` and `pyomo.contrib.appsi.fbbt` produce the same results.

### Legal Acknowledgement


michaelbynum:
I am going to close this and split it into 3 pull requests.

Issue #1032, opened 2019-05-23
[PEP] Solving Blocks with variables that do not live on the Block


michaelbynum:
I would like to be able to solve Pyomo Blocks with variables that do not live on the Block. I will describe my use case, but there are others. 

I have a structured MINLP. In fact, it looks like a binary tree of Blocks. I have a transformation to relax this MINLP. The the root block has a constraint with a bilinear term (for example), ```x*y```, then I will replace that bilinear term with an auxiliary variable, ```aux```, and add  constraints for the McCormick envelopes relating ```aux``` to ```x``` and ```y```. If this same bilinear term, ```x*y``` appears in other constraints, I want to replace it with the same auxiliary variable ```aux```. I do not want to create a new variable for the same term. Now, suppose I encounter the same bilinear term in a constraint on a child Block. I already have the auxiliary variable on the root Block. It is not easy to move the auxiliary variable so that it has a new parent. I also have a need to solve individual child Blocks.

Another example I am aware of arises in GDP transformations. Maybe @jsiirola can elaborate.

carldlaird:
I see the use case you describe. Two comments:

1) Not that I think this is better, but would it make sense to use a transformation to flatten the model first? Then this kind of issue would go away? (I suspect this would bring a host of other issues, not the least of which is the required name mangling when flattening.)

2) Is this (and other use cases like this) implying that variables are not really tied to their parent?

emma58:
I think that besides the naming issues, flattening the model also obscures the structure which was likely why you were using Blocks in the first place. I don't think there is any reason to assume that you are going to stop messing with a model after just one solve. If you have to flatten to do that solve, you've just made it much harder to make further changes. And I think if individual Blocks on the original model were solvable before the transformation, it seems reasonable to expect them to be after it as well. Granted, transformations could be designed so that this was still true, but in the GDP transformations at least, we are using Blocks so that the transformed model contains a completely intact version of the original model (possibly with some components deactivated). Everything the transformation adds is on a new Block it creates. This helps map the original model to what the transformation did and makes the names of components created by transformations predictable.

qtothec:
I'll add that flattening also obscures the structure that the solvers I am involved with rely upon.

michaelbynum:
The only reason I can think of to require that variables live on the block bening solved is for model validation (which certainly depends on how you conceptually define a model). However, model validation is not really what is currently happening. Consider the following:

```
>>> b = pe.Block(concrete=True)
[...]

PR #2249, opened 2022-01-08
Redesign `Var.stale` handling to use a global state flag

jsiirola:
## Summary/Motivation:
The current implementation of the `Var.stale` indicator results in inconsistent updates to the "staleness" of a VarData when solving sub-trees in a block-hierarchical model.  Consider a model with a sub-block with variables and constraints on both the outer and inner blocks.  Constraints only reference variables within their current sub-tree: constraints on the sub-block only reference variables on the sub-block, and constraints on the model reference variables on both the model and the sub-block.

If I solve the model, then all variable `stale` flags will be `False`.  If I then go and solve just the sub-block, then the variables on the sub-block will still have their "stale" flags set to `False`, but then the variables on the model will also still be `False`, even though those variables are very much stale.

This issue arises because the `stale` flag is a simple `bool`, and updating the flag is the responsibility of the solution loader.  As the solution loader only works in the context of the sub-block being solved, it does not know about the larger model and does not set the `stale` flag on the upper-level model:
```python
>>> m = ConcreteModel()
>>> m.x = Var()
>>> m.b = Block()
>>> m.b.y = Var()
>>> m.b.o = Objective(expr=m.b.y**2)
>>> m.o = Objective(expr=(m.x-1)**2 + (m.b.y-3)**2)
>>> m.b.o.deactivate()
>>> s = SolverFactory('ipopt')
>>> s.solve(m)
{'Problem': [{'Lower bound': -inf, 'Upper bound': inf, 'Number of objectives': 1, 'Number of constraints': 0, 'Number of variables': 2, 'Sense': 'unknown'}], 'Solver': [{'Status': 'ok', 'Message': 'Ipopt 3.13.2\\x3a Optimal Solution Found', 'Termination condition': 'optimal', 'Id': 0, 'Error rc': 0, 'Time': 0.01160120964050293}], 'Solution': [OrderedDict([('number of solutions', 0), ('number of solutions displayed', 0)])]}
>>> m.pprint()
1 Var Declarations
    x : Size=1, Index=None
        None :  None :   1.0 :  None : False : False :  Reals

1 Objective Declarations
    o : Size=1, Index=None, Active=True
        Key  : Active : Sense    : Expression
        None :   True : minimize : (x - 1)**2 + (b.y - 3)**2

1 Block Declarations
    b : Size=1, Index=None, Active=True
        1 Var Declarations
            y : Size=1, Index=None
                None :  None :   3.0 :  None : False : False :  Reals

        1 Objective Declarations
            o : Size=1, Index=None, Active=False
                Key  : Active : Sense    : Expression
                None :  False : minimize :     b.y**2

        2 Declarations: y o

3 Declarations: x b o
>>> m.b.o.activate()
>>> s.solve(m.b)
{'Problem': [{'Lower bound': -inf, 'Upper bound': inf, 'Number of objectives': 1, 'Number of constraints': 0, 'Number of variables': 1, 'Sense': 'unknown'}], 'Solver': [{'Status': 'ok', 'Message': 'Ipopt 3.13.2\\x3a Optimal Solution Found', 'Termination condition': 'optimal', 'Id': 0, 'Error rc': 0, 'Time': 0.012385129928588867}], 'Solution': [OrderedDict([('number of solutions', 0), ('number of solutions displayed', 0)])]}
>>> m.pprint()
1 Var Declarations
    x : Size=1, Index=None
        None :  None :   1.0 :  None : False : False :  Reals

1 Objective Declarations
    o : Size=1, Index=None, Active=True
        Key  : Active : Sense    : Expression
        None :   True : minimize : (x - 1)**2 + (b.y - 3)**2

1 Block Declarations
    b : Size=1, Index=None, Active=True
        1 Var Declarations
            y : Size=1, Index=None
                None :  None :   0.0 :  None : False : False :  Reals

        1 Objective Declarations
            o : Size=1, Index=None, Active=True
                Key  : Active : Sense    : Expression
                None :   True : minimize :     b.y**2

        2 Declarations: y o

3 Declarations: x b o
```

This PR proposes to change how we calculate "staleness" to use a global, monotonically increasing "stale flag".  The Var's stale attribute is no longer a bool, and is instead the value of the global stale flag at the last time the Var value was updated.  A stale variable is then a variable whose stale flag is older than the current global stale flag.  This makes "updating" staleness simpler and more well-defined: incrementing the global stale flag marks all variables as stale.  At present, we only need to update the stale flag when loading a solution (either from a solver or from a solutions database), and whenever a non-stale variable's value is changed.

One consequence of this change is that if there are multiple models, solving any one model will mark the variables in all other models as stale.  While this can initially appear to be inconvenient and a strange side effect, the behavior is more deterministic (and defensible) compared to the current behavior of having to "find" all variables that have been declared anywhere.

This PR is compatible with the concept of allowing models to reference variables outside the current subblock (PEP #1032)

- switch `stale` attributes from `bool` to properties that calculate the current staleness by comparing a local integer to a global integer.
- switch Var `fixed` from attribute to property and disable in AbstractVars

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/PyomoModel.py
+++ b/pyomo/core/base/PyomoModel.py
@@ -541,9 +541,9 @@ class ModelSolutions(object):
                 if attr_key in valid_import_suffixes:
                     valid_import_suffixes[attr_key][cdata] = attr_value
 
-        # Advance the stale flag again so that any subsequent variable
-        # updates will "set" the stale flag for these variables
-        StaleFlagManager.advance_flag()
+        # Set the state flag to "delayed advance": it will auto-advance
+        # if a non-stale variable is updated.
+        StaleFlagManager.advance_flag(delayed=True)
 
 @ModelComponentFactory.register('Model objects can be used as a component of other models.')
 class Model(ScalarBlock):
--- a/pyomo/core/base/PyomoModel.py
+++ b/pyomo/core/base/PyomoModel.py
@@ -463,7 +463,7 @@ class ModelSolutions(object):
         # loading the solution, so you known which variables have "real"
         # values and which ones don't.
         #
-        StaleFlagManager.advance_flag()
+        StaleFlagManager.mark_all_as_stale()
 
         if index is not None:
             self.index = index
@@ -542,8 +542,9 @@ class ModelSolutions(object):
                     valid_import_suffixes[attr_key][cdata] = attr_value
 
         # Set the state flag to "delayed advance": it will auto-advance
-        # if a non-stale variable is updated.
-        StaleFlagManager.advance_flag(delayed=True)
+        # if a non-stale variable is updated (causing all non-stale
+        # variables to be marked as stale).
+        StaleFlagManager.mark_all_as_stale(delayed=True)
 
 @ModelComponentFactory.register('Model objects can be used as a component of other models.')
 class Model(ScalarBlock):
--- a/pyomo/core/base/PyomoModel.py
+++ b/pyomo/core/base/PyomoModel.py
@@ -22,6 +22,7 @@ from pyomo.common.dependencies import pympler, pympler_available
 from pyomo.common.deprecation import deprecated, deprecation_warning
 from pyomo.common.gc_manager import PauseGC
 from pyomo.common.log import is_debug_set
+from pyomo.core.staleflag import StaleFlagManager
 from pyomo.core.expr.symbol_map import SymbolMap
 from pyomo.core.base.component import ModelComponentFactory
 from pyomo.core.base.var import Var
@@ -458,11 +459,13 @@ class ModelSolutions(object):
         """
         instance = self._instance()
         #
-        # Set the "stale" flag of each variable in the model prior to loading the
-        # solution, so you known which variables have "real" values and which ones don't.
+        # Set the "stale" flag of each variable in the model prior to
+        # loading the solution, so you known which variables have "real"
+        # values and which ones don't.
         #
-        instance._flag_vars_as_stale()
-        if not index is None:
+        StaleFlagManager.advance_flag()
+
+        if index is not None:
             self.index = index
         soln = self.solutions[self.index]
 
@@ -521,7 +524,6 @@ class ModelSolutions(object):
                                        str(vdata.value)))
 
             vdata.set_value(val, skip_validation=True)
-            vdata.stale = False
 
             for _attr_key, attr_value in entry.items():
                 attr_key = _attr_key[0].lower() + _attr_key[1:]
@@ -539,6 +541,9 @@ class ModelSolutions(object):
                 if attr_key in valid_import_suffixes:
                     valid_import_suffixes[attr_key][cdata] = attr_value
 
+        # Advance the stale flag again so that any subsequent variable
+        # updates will "set" the stale flag for these variables
+        StaleFlagManager.advance_flag()
 
 @ModelComponentFactory.register('Model objects can be used as a component of other models.')
 class Model(ScalarBlock):
--- a/pyomo/core/base/var.py
+++ b/pyomo/core/base/var.py
@@ -324,11 +324,15 @@ class _GeneralVarData(_VarData):
         state = super(_GeneralVarData, self).__getstate__()
         for i in _GeneralVarData.__slots__:
             state[i] = getattr(self, i)
+        state['_stale'] = self.stale
         return state
 
-    # Note: None of the slots on this class need to be edited, so we
-    # don't need to implement a specialized __setstate__ method, and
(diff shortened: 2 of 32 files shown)


michaelbynum (commented):
This is awesome! 

Why did `fixed` need to become a property?

This is not important for this PR, but I think we need to brainstorm a little bit about the stale concept with persistent solver interfaces. One of the nice features of some of the persistent solver interfaces is that users can request to only load a subset of the variable values, which can help with performance for some applications. However, this makes updating the stale flag problematic. I've sort of ignored this problem so far. The existing persistent solver interfaces only set the `stale` flag to `False` if the user asks for the value of that variable to be loaded. Maybe this is the correct behavior? I think it is the correct behavior if the `stale` flag is supposed to indicate whether or not the `value` attribute is stale or not (rather than whether or not the variable was used in the last solve). I think that aligns with this PR. Thoughts?

pyomo/core/base/PyomoModel.py
```diff
@@ -458,11 +459,13 @@ def select(self,
         """
         instance = self._instance()
         #
-        # Set the "stale" flag of each variable in the model prior to loading the
-        # solution, so you known which variables have "real" values and which ones don't.
+        # Set the "stale" flag of each variable in the model prior to
+        # loading the solution, so you known which variables have "real"
+        # values and which ones don't.
         #
-        instance._flag_vars_as_stale()
```
michaelbynum, line 464: Should the `_flag_vars_as_stale` method get deprecated?

michaelbynum: Ignore this comment. I did not see that the method was deleted.

pyomo/core/base/var.py
```diff
@@ -408,20 +385,22 @@ def set_value(self, val, skip_validation=False):
                     extra={'id':'W1002'},
                 )
 
+        if self._value == val and not self.stale:
+            # do not update the value / set the stale flag if we are not
+            # changing the value and the variable is currently not stale
+            return
```
michaelbynum, line 391: Is this desired? I would think the stale flag should always get updated when calling `set_value`, even if the the value is the same. What if the value just happens to be the same?

```
>>> StaleFlagManager.mark_all_as_stale(delayed=False)
>>> m.x.value = 0
>>> m.y.value = 1
>>> StaleFlagManager.mark_all_as_stale(delayed=True)
>>> m.x.stale, m.y.stale
(False, False)
>>> m.x.value = 0; m.y.value = 0
>>> m.x.stale, m.y.stale
(True, False)
```

I don't think this is correct???

pyomo/solvers/tests/models/LP_unused_vars.json
```diff
@@ -2,35 +2,35 @@
   "B[1]": {}, 
   "B[1].b": {}, 
   "B[1].b.X[1]": {
-    "stale": false, 
+    "stale": true,
```
michaelbynum, line 5: Why did these need changed?

michaelbynum (approved):

pyomo/core/base/var.py
jsiirola: Desired is of course open for discussion, but it is intentional given this implementation.  The idea is that "stale" indicates that the variable value is potentially inconsistent with the results from the most recent "batch" model update (e.g., from running a solver or loading a solution).  The implicit assumption is that after a solve, a user / algorithm may update variables: if the variable is currently stale, the assumption is that the new value is to update the variable to be consistent with the solve result.  However, if the variable is not stale, then if the variable value is being changed, then it will (almost by definition) *not* be consistent with the previous solve, so the current stale flag gets advanced (marking all variables as stale) before assigning the new value.  However, if the variable value is not changing, then it is still consistent with the previous solve, and since it is still consistent, we don't have to advance the stale flag.

The more I think about it, I could be convinced that a user updating a variable value that is currently not stale is *intentionally* updating the model, so we could / should remove this case (and make it so that *any* update to a non-stale variable would trigger the delayed "mark as stale" action).  The workaround is of course for the user to mark the variable as stale before updating the value (something that we do as part of solution updates in MindtPy).

pyomo/solvers/tests/models/LP_unused_vars.json
michaelbynum: Got it. Thanks.

pyomo/core/base/var.py
jsiirola: At the Dev call we decided that the expected behavior for `stale` should follow @michaelbynum's original intuition.  The PR has been updated (including tests for the expected behavior).

michaelbynum (approved):
Looks good. Can we go ahead and merge this?

PR #2263, opened 2022-01-19
gdp.fix_disjuncts transformation should create a fully algebraic model

### LINKED ISSUE #2261: gdp.fix_disjuncts assumes all Disjuncts are set and doesn't die very gracefully if they aren't
## Summary

This is issue is mainly a question, which is: Should the `gdp.fix_disjuncts` transformation be a transformation from a GDP to a MI(N)LP? Historically, this seems to have been assumed: at least in `gdpopt`, after `fix_disjuncts` is called, it is assumed that the NLP subproblem can be solved with a MI(N)LP solver. The alternative would be for it to support `Disjunction` targets, in which case it should complain if not all the Disjuncts' indicator_vars in the Disjunction are set, but it could leave unspecified Disjunctions alone. In the first case, I think it would have a responsibility to transform logical stuff on the model. In the second case, probably not. In either case, it needs an error message that tells the user something more informative than the `ValueError` below.

```
from pyomo.environ import *
from pyomo.gdp import *

m = ConcreteModel()
m.d1 = Disjunct()
m.d2 = Disjunct()
m.d = Disjunction(expr=[m.d1, m.d2])
m.another1 = Disjunct()
m.another2 = Disjunct()
m.another = Disjunction(expr=[m.another1, m.another2])

m.d1.indicator_var.set_value(True)
m.d2.indicator_var.set_value(False)
        
TransformationFactory('gdp.fix_disjuncts').apply_to(m)
```

### Error Message

```
ERROR: evaluating object as numeric value: another1.binary_indicator_var
        (object: <class 'pyomo.gdp.disjunct.AutoLinkedBinaryVar'>)
    No value for uninitialized NumericValue object
    another1.binary_indicator_var
  File "partially_fixed.py", line 15, in <module>
    TransformationFactory('gdp.fix_disjuncts').apply_to(m)
  File "/home/esjohn/src/pyomo/pyomo/core/base/transformation.py", line 69, in apply_to
    self._apply_to(model, **kwds)
  File "/home/esjohn/src/pyomo/pyomo/gdp/plugins/fix_disjuncts.py", line 49, in _apply_to
    self._transformContainer(model)
  File "/home/esjohn/src/pyomo/pyomo/gdp/plugins/fix_disjuncts.py", line 66, in _transformContainer
    if fabs(value(_binary) - 1) <= self.config.integer_tolerance:
  File "/home/esjohn/src/pyomo/pyomo/core/expr/numvalue.py", line 143, in value
    raise ValueError(
ValueError: No value for uninitialized NumericValue object another1.binary_indicator_var
```

Pyomo version: main
Python version: 3.8
Operating system: centos7
Solver (if applicable):

emma58:
## Summary/Motivation:

I think that the `gdp.fix_disjuncts` transformation should be guaranteed to transform a GDP to a fully algebraic model. The cases where this might not happen are:
1) If there are active LogicalConstraints on the model
2) If not all of the Disjuncts' indicator_vars have values.

In the case of 2, the transformation gives a `ValueError` (see #2261). The current behavior for 1 is to ignore the LogicalConstraints. This PR changes that to calling `logical_to_linear` at the end so that they are transformed.

- Change the error when not all Disjuncts have a value to have a useful message
- Include a call to `logical_to_linear` after the Disjuncts have been reclassified, to transform any global or local-to-a-selected-Disjunct LogicalConstraints

### Legal Acknowledgement

### DIFF
--- a/pyomo/gdp/plugins/fix_disjuncts.py
+++ b/pyomo/gdp/plugins/fix_disjuncts.py
@@ -20,8 +20,8 @@ logger = logging.getLogger('pyomo.gdp.fix_disjuncts')
 class GDP_Disjunct_Fixer(Transformation):
     """Fix disjuncts to their current Boolean values.
 
-    This reclassifies all disjuncts in the passed model instance as ctype Block and deactivates the
-    constraints and disjunctions within inactive disjuncts.
+    This reclassifies all disjuncts in the passed model instance as ctype Block
+    and deactivates the constraints and disjunctions within inactive disjuncts.
 
     """
 
@@ -38,19 +38,24 @@ class GDP_Disjunct_Fixer(Transformation):
     ))
 
     def _apply_to(self, model, **kwds):
-        """Fix all disjuncts in the given model and reclassify them to Blocks."""
+        """Fix all disjuncts in the given model and reclassify them to 
+        Blocks."""
         config = self.config = self.CONFIG(kwds.pop('options', {}))
         config.set_value(kwds)
 
         self._transformContainer(model)
 
         # Reclassify all disjuncts
-        for disjunct_object in model.component_objects(Disjunct, descend_into=(Block, Disjunct)):
-            disjunct_object.parent_block().reclassify_component_type(disjunct_object, Block)
+        for disjunct_object in model.component_objects(Disjunct,
+                                                       descend_into=(Block,
+                                                                     Disjunct)):
+            disjunct_object.parent_block().reclassify_component_type(
+                disjunct_object, Block)
 
     def _transformContainer(self, obj):
         """Find all disjuncts in the container and transform them."""
-        for disjunct in obj.component_data_objects(ctype=Disjunct, active=True, descend_into=True):
+        for disjunct in obj.component_data_objects(ctype=Disjunct, active=True,
+                                                   descend_into=True):
             _binary = disjunct.binary_indicator_var
             if fabs(value(_binary) - 1) <= self.config.integer_tolerance:
                 disjunct.indicator_var.fix(True)
@@ -62,7 +67,9 @@ class GDP_Disjunct_Fixer(Transformation):
                     'Non-binary indicator variable value %s for disjunct %s'
                     % (disjunct.name, value(_binary)))
 
-        for disjunction in obj.component_data_objects(ctype=Disjunction, active=True, descend_into=True):
+        for disjunction in obj.component_data_objects(ctype=Disjunction,
+                                                      active=True,
+                                                      descend_into=True):
             self._transformDisjunctionData(disjunction)
 
     def _transformDisjunctionData(self, disjunction):
--- a/pyomo/gdp/plugins/fix_disjuncts.py
+++ b/pyomo/gdp/plugins/fix_disjuncts.py
@@ -4,7 +4,7 @@
 import logging
 from math import fabs
 
-from pyomo.common.config import ConfigBlock, ConfigValue, NonNegativeFloat
+from pyomo.common.config import ConfigBlock, NonNegativeFloat
 from pyomo.core.base import Transformation, TransformationFactory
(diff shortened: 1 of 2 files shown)


bernalde (approved):

pyomo/gdp/plugins/fix_disjuncts.py
```diff
@@ -38,20 +43,33 @@ def __init__(self, **kwargs):
...
-        for disjunct_object in model.component_objects(Disjunct, descend_into=(Block, Disjunct)):
-            disjunct_object.parent_block().reclassify_component_type(disjunct_object, Block)
+        for disjunct_object in model.component_objects(Disjunct,
+                                                       descend_into=(Block,
+                                                                     Disjunct)):
+            disjunct_object.parent_block().reclassify_component_type(
+                disjunct_object, Block)
+
+        # Transform any remaining logical stuff
+        TransformationFactory('core.logical_to_linear').apply_to(model)
```
michaelbynum, line 61: Is it worth having an option to control whether or not this happens?

bernalde: I thought about this, but I cannot think of any application where you would like to have lingering logical constraints after fixing disjunct. Do you have any examples in mind?

pyomo/gdp/plugins/fix_disjuncts.py
```diff
@@ -38,20 +43,33 @@ def __init__(self, **kwargs):
...
         config = self.config = self.CONFIG(kwds.pop('options', {}))
         config.set_value(kwds)
 
         self._transformContainer(model)
 
         # Reclassify all disjuncts
-        for disjunct_object in model.component_objects(Disjunct, descend_into=(Block, Disjunct)):
-            disjunct_object.parent_block().reclassify_component_type(disjunct_object, Block)
+        for disjunct_object in model.component_objects(Disjunct,
+                                                       descend_into=(Block,
```
jsiirola, line 55: Should this specify `active=True`?

pyomo/gdp/plugins/fix_disjuncts.py
```diff
@@ -38,20 +43,33 @@ def __init__(self, **kwargs):
...
+
+        # Transform any remaining logical stuff
+        TransformationFactory('core.logical_to_linear').apply_to(model)
 
     def _transformContainer(self, obj):
         """Find all disjuncts in the container and transform them."""
-        for disjunct in obj.component_data_objects(ctype=Disjunct, active=True, descend_into=True):
+        for disjunct in obj.component_data_objects(ctype=Disjunct, active=True,
+                                                   descend_into=True):
             _binary = disjunct.binary_indicator_var
```
jsiirola, line 67: This transformation predates the conversion of `indicator_var` to being a `BooleanVar`.  I believe that this logic can be simplified to use `indicator_var` and only check for `None`, `True`, and `False`.  Fractional `binary_indicator_var` values are automatically mapped to `None` values for `indicator_var`.  I would recommend switching to `indicator_var` and removing the tolerance logic below.

jsiirola (changes requested):

pyomo/gdp/plugins/fix_disjuncts.py
emma58: I'm inclined to agree with @bernalde. I think you may as well head all the way to algebraic-land (assuming the only stuff that needed transformation on the model *was* GDP stuff) because there's not really any way to go back to something logical after calling this.

pyomo/gdp/plugins/fix_disjuncts.py
emma58: OK, I changed this. I removed the integer_tolerance option completely because there's no way to change that tolerance for the BooleanVars right now, I don't think.

jsiirola (approved):

PR #2283, opened 2022-02-09
Use projected NLPs for the inner problem solves in ExternalPyomoModel

Robbybp:
## Summary/Motivation:
There is a significant performance improvement for some problems if we create ProjectedNLPs that we solve repeatedly with CyIpopt, rather than solve Pyomo blocks with Ipopt, for the inner problems in the implicit function formulation.

- Add the data structures necessary to repeatedly solve a sequence of square problems as PyomoNLPs, where variables are updated before each solve
- Change the default behavior of ExternalPyomoModel to solve ProjectedNLPs with CyIpopt, and only use Ipopt (or another Pyomo solver) if specified, or if CyIpopt is unavailable.

### A comparison comparison on a chemical looping steady state optimization problem:
- Current main:
2022-02-08_before_patch
- This branch
2022-02-08_after_patch

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/pynumero/interfaces/external_pyomo_model.py
+++ b/pyomo/contrib/pynumero/interfaces/external_pyomo_model.py
@@ -15,6 +15,7 @@ from pyomo.core.base.constraint import Constraint
 from pyomo.core.base.objective import Objective
 from pyomo.core.expr.visitor import identify_variables
 from pyomo.common.collections import ComponentSet
+from pyomo.core.base.suffix import Suffix
 from pyomo.util.calc_var_value import calculate_variable_from_constraint
 from pyomo.util.subsystems import (
     create_subsystem_block,
@@ -23,6 +24,12 @@ from pyomo.util.subsystems import (
 from pyomo.contrib.pynumero.interfaces.pyomo_nlp import PyomoNLP
 from pyomo.contrib.pynumero.interfaces.external_grey_box import (
     ExternalGreyBoxModel,
+)
+from pyomo.contrib.pynumero.interfaces.nlp_projections import ProjectedNLP
+from pyomo.contrib.pynumero.algorithms.solvers.cyipopt_solver import (
+    cyipopt_available,
+    CyIpoptNLP,
+    CyIpoptSolver,
 )
 from pyomo.contrib.incidence_analysis.util import (
     generate_strongly_connected_components,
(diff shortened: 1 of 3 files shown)


pyomo/contrib/pynumero/interfaces/external_pyomo_model.py
```diff
@@ -190,14 +283,57 @@ def set_input_values(self, input_values):
...
+        vector_scc_idx = 0
         for block, inputs in self._scc_list:
             if len(block.vars) == 1:
                 calculate_variable_from_constraint(
                     block.vars[0], block.cons[0]
                 )
             else:
-                with TemporarySubsystemManager(to_fix=inputs):
-                    solver.solve(block)
+                if self._use_cyipopt:
```
jsiirola, line 293: `elif` to avoid the extra indentation?

pyomo/contrib/pynumero/interfaces/external_pyomo_model.py
```diff
@@ -190,14 +283,57 @@ def set_input_values(self, input_values):
...
+                    for var, val in zip(variables, new_primals):
+                        var.set_value(val)
+
+                else:
+                    # Use a Pyomo solver to solve this strongly connected
+                    # component.
+                    with TemporarySubsystemManager(to_fix=inputs):
+                        solver.solve(block)
+
+                vector_scc_idx += 1
```
jsiirola, line 336: I think this is technically correct as written, but it would be clearer if it was done inside the `if self._use_cyipopt` block.

Robbybp: I think I like having this outside the `if self._use_cyipopt` tree, as otherwise, in the case where we use a regular Pyomo solver, `vector_scc_idx` will always be zero, even though we may not be at the zeroth "vector-valued" SCC. Having this information may be useful, e.g. for debugging, even if it's not strictly necessary for the algorithm.

pyomo/contrib/pynumero/interfaces/external_pyomo_model.py
Robbybp: See comment below. I think I want to have this extra if statement so that the increment of `vector_scc_idx` happens regardless of whether we're using CyIpopt.

jsiirola (approved):

blnicho (approved):

PR #2295, opened 2022-02-21
Support string arguments in NL external functions

### LINKED ISSUE #1903: NL-writer External Function Issues

jsiirola:
## Summary/Motivation:
This is a partial rework, building off of #1904.

This adds support for passing strings to external AMPL functions and fixes issues in #1903. The ASL cannot handle Windows-style (`\r\n`) line endings for the "h" lines in the NL file (the lines that mark string arguments).  To generate a parseable NL file, "h"-lineas must be terminated with UNIX-style (`\n`) line endings.

#1904 originally proposed disabling universal line endings for the NL file, which works, but requires updating all the NL baseline tests to be line ending agnostic (which is further complicated as git by default will convert line endings to match the target platform).  It is an open question as to if we want to make that move in the long term.

This PR also refactors how we locate / build the ASL.  Instead of duplicating the build harness from pynumero, this PR extracts the core logic for building / linking against the ASL into a separate cmake module.  Both the `CMakeLists.txt` here and in Pynumero now link in the same common module.

This also switches pynumero/ampl_function_demo to build against https://github.com/AMPL/asl instead of https://github.com/AMPL/mp.  This significantly simplifies the build process and opens the possibility of distributing these libraries in binary form (AMPL/asl is *just* the public domain ASL code).

- Abstract the Disutils extension for building CMake projects into `pyomo/common/cmake_builder.py`
- Abstract the logic for finding/building the ASL into `FindASL.cmake`
- Bypass universal newlines for NL "h"-lines
- Add support for evaluating external functions wtith string arguments

### Legal Acknowledgement


michaelbynum (approved):
Looks good.

blnicho (approved):

PR #2296, opened 2022-02-22
appsi: making time limit test more robust

michaelbynum:
- Make the time limit test in appsi more robust by making the model more complex.

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/appsi/solvers/tests/test_persistent_solvers.py
+++ b/pyomo/contrib/appsi/solvers/tests/test_persistent_solvers.py
@@ -824,10 +824,7 @@ class TestSolvers(unittest.TestCase):
         if platform == 'win32':
             raise unittest.SkipTest
 
-        if opt.available() == opt.Availability.LimitedLicense:
-            N = 30
-        else:
-            N = 100
+        N = 30
         m = pe.ConcreteModel()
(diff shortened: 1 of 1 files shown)


jsiirola (approved):

pyomo/contrib/appsi/solvers/tests/test_persistent_solvers.py
```diff
@@ -819,28 +820,47 @@ def test_time_limit(self, name: str, opt_class: Type[PersistentSolver]):
...
-        if type(opt) is Cbc:
-            raise unittest.SkipTest
         from sys import platform
         if platform == 'win32':
             raise unittest.SkipTest
+
+        if opt.available() == opt.Availability.LimitedLicense:
+            N = 30
+        else:
+            N = 100
```
jsiirola, line 830: Is there a reason to make the model bigger when a license is present?  If the test reliably times out with the smaller model, it should do the same with the larger one, right?

michaelbynum: It seems to me that different solvers check the time limit at different times. I think you are still correct though. I think this should work with `N=30` for all solvers.

PR #2299, opened 2022-02-22
Appsi: correctly identify changes to constraints

michaelbynum:
## Summary/Motivation:
Currently, Appsi does more work than necessary when the `_lower` or `_upper` attributes are native numeric types. This happens because the `lower` and `upper` properties return a new instance of `NumeriConstant` every call.

- Add extra checks to avoid processing constraints that have not changed.

### Legal Acknowledgement


pyomo/contrib/appsi/base.py
```diff
@@ -1052,8 +1053,22 @@ def update(self, timer: HierarchicalTimer = None):
...
             for c in cons_to_update:
                 lower, body, upper = self._active_constraints[c]
-                if c.lower is not lower or c.body is not body or c.upper is not upper:
+                new_lower, new_body, new_upper = c.lower, c.body, c.upper
+                if new_body is not body:
                     cons_to_remove_and_add[c] = None
+                    continue
+                if new_lower is not lower:
+                    if type(new_lower) is NumericConstant and type(lower) is NumericConstant and new_lower.value == lower.value:
+                        pass
```
jsiirola, line 1062: This seems like a very specific edge case.  Isn't is sufficient to just have:
```python
if value(new_lower) == value(lower):
    pass
```

michaelbynum: It is not. It is possible that the `lower` or `upper` changed from a constant to a mutable parameter with the same value. If that happens, we need to reprocess the constraint in case the value of the mutable parameter changes in the future.

jsiirola: That makes sense.

jsiirola (approved):

PR #2322, opened 2022-03-01
Solve square problem to initialize regression problem using parmest

kanishka-ghosh:
## Summary/Motivation: 
Solve a square instance of the parameter estimation problem in parmest for complex Pyomo models before solving the actual regression problem

- Add code and tests to solve the square problem in order to initialize the regression problem in parmest (@adowling2)

### Legal Acknowledgement


pyomo/contrib/parmest/parmest.py
```diff
@@ -433,14 +434,69 @@ def _Q_opt(self, ThetaVals=None, solver="ef_ipopt",
...
+                    try:
+                        # If the component that was found is not a variable,
+                        # this will generate an exception (and the warning
+                        # in the 'except')
+                        var_validate.unfix()
+                        # We want to standardize on the CUID string
+                        # representation
+                        # self.theta_names[i] = repr(var_cuid)
+                    except:
+                        logger.warning(theta + ' is not a variable')
```
blnicho, line 490: Instead of repeating this logic I think it would make more sense to just keep a list of the vars being fixed above and then just iterate over that list to unfix them. Going through the process of searching for them again is inefficient and there is no need to redo the validation logic since the error would have already been thrown above.

pyomo/contrib/parmest/parmest.py
```diff
@@ -662,9 +719,12 @@ def theta_est(self, solver="ef_ipopt", return_values=[], calc_cov=False, cov_n=N
...
             If True, calculate and return the covariance matrix (only for "ef_ipopt" solver)
         cov_n: int, optional
-            If calc_cov=True, then the user needs to supply the number of datapoints 
+            If calc_cov=True, then the user needs to supply the number of datapoints
             that are used in the objective function
-            
+        pre_solve: boolean, optional
+            If True, solve square problem with parameters fixed before solving parameter estimation
+        bound_push: float, optional
+            If True, set ipopt solver option 'bound_push' to value provided
```
blnicho, line 727: The `bound_push` option is too specific to IPOPT. I recommend generalizing this to something like `solver_options` instead of assuming the solver will always be IPOPT

blnicho:
@kanishka-ghosh @adowling2 this has been open as a WIP for a while. Is this something you're still actively working on? Are you waiting on input from us?

 As a general rule we tend to not focus any of our PR review resources on WIP PRs unless specifically asked.

pyomo/contrib/parmest/parmest.py
kanishka-ghosh: @blnicho yes, I appreciate that `bound_push` is IPOPT-specific so, to avoid confusion, I have updated the name of the optional argument to `initialization_with_bound_push`.  I am adding this as a separate argument to `solver_options` since IPOPT is hard-coded as the solver in the method `_Q_opt()` in parmest.py and in the method `inv_reduced_hessian_barrier()` in inverse_reduced_hessian.py

mrmundt:
@kaklise - Would you mind taking a look at this when you get a chance?

kaklise:
I believe the ability to pre_solve duplicates functionality in `pest.objective_at_theta`.  The difference is that the user can provide multiple input values to `pest.objective_at_theta` and the `pre_solve` input argument to `pest.theta_est` fixes the values from the model.  I'm not sure if both options are needed.  

As for `initialization_with_bound_push`, can we instead use `solver_options` when the Estimator class is created?

adowling2:
@kaklise This is good to know. I think what you are saying is that the initialization capability can be provided as follows:
- Create the estimator object `pest`
- Call the `pest.objective_at_theta` function to effectively initialize the model at the given value of theta
- Adjust the solver options such as `bound_push` as one sees fits
- Call the parameter estimation problem by calling `pest.theta_est`

Does this approach make sense? @kanishka-ghosh What are your thoughts?

kaklise:
That's right.  It's a little cumbersome to create a DataFrame with the theta values if you just want to fix the variables and solve.  We could update `pest.objective_at_theta` to use the initialized Var values if `theta_values=None` is passed into the method.

adowling2:
I like that idea.

adowling2:
I like that idea.

kanishka-ghosh:
@adowling2 Yes, modifying `pest.objective_at_theta` to set parameters to initialization values when `theta_values=None` will do what `pre_solve=True` is supposed to do in `theta_est()`.  As you mentioned, this will require the user to call 2 functions: `pest.objective_at_theta()` followed by `pest.theta_est()` to perform parameter estimation with square problem initialization

kanishka-ghosh:
Feature being added following Issue #2377 and new PR will be opened once the feature is implemented

PR #2323, opened 2022-03-01
Skip bound/domain validation in `ExternalPyomoModel`

Robbybp:
## Fixes # .
Warnings in implicit function solves when we set a variable to a value outside its bounds. 

## Summary/Motivation:
Sometimes IPOPT sets a variable with a value that violates its bound slightly (for some reason). This is not really a problem, and is pretty ugly when a bound violation warning pops up in the middle of your IPOPT trace complaining about a violation of $1e-4$ or something, so I'd rather use the `skip_validation` option.

### Legal Acknowledgement


Robbybp:
Maybe this should be a configurable option. I could see wanting to know about this.

jsiirola (approved):

jsiirola:
I like the idea of making this a configurable option, though.

PR #2331, opened 2022-03-07
Fix PyROS Variable Default Initialization

shermanjasonaf:
## Summary/Motivation:
PyROS does not gracefully handle models with:
- uninitialized variables on which the objective function is contingent
- uninitialized second stage variables

as auxiliary variables are initialized to values using the `value()` method. In either case, an exception is raised by the `value()` method.

- add `exception=False` option to the `value()` method calls used for initializing epigraph and decision rule variables. Consequently, by default, auxiliary variables are not initialized if any of the variables used to
initialize the auxiliary variables are not initialized.
- add a corresponding unit test

### Legal Acknowledgement


jsiirola (approved):
This looks good.

michaelbynum (approved):

PR #2347, opened 2022-03-17
Better support for absolute value in differentiation and FBBT

michaelbynum:
## Summary/Motivation:
While trying to check the scaling of a model, I realized we did not support symbolic differentiation or interval arithmetic for `AbsExpression`.  This PR fixes that.

- Support symbolic differentiation of `AbsExpression`
- Support interval arithmetic with `AbsExpression` both in the pure Python version and in Appsi

### Legal Acknowledgement


pyomo/contrib/fbbt/fbbt.py
```diff
@@ -1035,13 +1059,15 @@ def _prop_bnds_root_to_leaf_GeneralExpression(node, bnds_dict, feasibility_tol):
...
 _prop_bnds_root_to_leaf_map[numeric_expr.LinearExpression] = _prop_bnds_root_to_leaf_LinearExpression
+_prop_bnds_root_to_leaf_map[numeric_expr.AbsExpression] = _prop_bnds_root_to_leaf_abs
 
 _prop_bnds_root_to_leaf_map[numeric_expr.NPV_ProductExpression] = _prop_bnds_root_to_leaf_ProductExpression
 _prop_bnds_root_to_leaf_map[numeric_expr.NPV_DivisionExpression] = _prop_bnds_root_to_leaf_DivisionExpression
 _prop_bnds_root_to_leaf_map[numeric_expr.NPV_PowExpression] = _prop_bnds_root_to_leaf_PowExpression
 _prop_bnds_root_to_leaf_map[numeric_expr.NPV_SumExpression] = _prop_bnds_root_to_leaf_SumExpression
 _prop_bnds_root_to_leaf_map[numeric_expr.NPV_NegationExpression] = _prop_bnds_root_to_leaf_NegationExpression
 _prop_bnds_root_to_leaf_map[numeric_expr.NPV_UnaryFunctionExpression] = _prop_bnds_root_to_leaf_UnaryFunctionExpression
+_prop_bnds_root_to_leaf_map[numeric_expr.NPV_AbsExpression] = _prop_bnds_root_to_leaf_abs
```
jsiirola, line 1070: General question about handling NPV nodes: as they *cannot* (by definition) contain Variables, then their bounds should always be the constant that that subexpression evaluates to, right?  In that case, shouldn't all NPV nodes be able to be serviced by the same callback?

michaelbynum: Yes, you are correct. Either way works.

michaelbynum: I think the appropriate way to handle this is to avoid descending into NPV expressions entirely. I'll do this in a separate PR.

jsiirola (approved):

pyomo/contrib/fbbt/fbbt.py
jsiirola: Sounds good.  Is it worth adding an issue as a reminder?

michaelbynum: Done! Good suggestion.

PR #2351, opened 2022-03-29
Adding indices to ComponentData objects (take 2)

### LINKED ISSUE #1130: [PEP] Add _index slot to _ComponentData

emma58:
## Summary/Motivation:

This supersedes #1228--it turned out to be easier to make the changes starting from the modern era than it was to revive the original PR.

- Renames `_index` to `_index_set` on all indexed components
- Adds `_index` to `_ComponentData` to store the ComponentData's key in its parent's `_data` dictionary
- Sets `_index=None` in the `__init__` methods of all ScalarComponents
- `getname` no longer means a linear scan *in the absence of a name buffer.* However, the behavior is the same as before when a name buffer is provided. This is because there are bugs with the `name_buffer` arg if the user uses the same dictionary but switches between fully-qualified and local names or switches what component the name is relative to. I will address this in a subsequent PR.

Things to look for when reviewing:
- Keeping the `_index` in sync with the parent's `_data` dictionary relies on the assumption that the only time key-value pairs are updated in this dictionary is in `_getitem_when_not_present` and `_setitem_when_not_present`. It turns out that's not true for `Var` and `BooleanVar`, and if there are other places where the assumption doesn't hold, the `_index` needs to be updated when the value is changed or when a key-value pair is added.
- `_ComponentData.__init__` is in-lined in *many* classes... Any `__init__` method that in-lines it needs to set `_index = NoArgumentGiven`.

## Performance impact

Performance results comparing main and this branch with 10 replicate each are here:
image
I'm not completely sure what's happening with the last test or if we should be concerned, but overall it looks like this is okay.

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/plugins/transform/util.py
+++ b/pyomo/core/plugins/transform/util.py
@@ -176,9 +176,9 @@ def _getAbstractIndices(comp):
     """
     Returns the index or index set of this component
     """
-    if type(comp._index_set) != type({}):
+    if type(comp.index_set()) != type({}):
         # Singly indexed component
-        return comp._index_set
+        return comp.index_set()
     else:
         # Unindexed constraint
         return {None: None}
--- a/pyomo/core/plugins/transform/util.py
+++ b/pyomo/core/plugins/transform/util.py
@@ -176,9 +176,9 @@ def _getAbstractIndices(comp):
     """
     Returns the index or index set of this component
     """
-    if type(comp._index) != type({}):
+    if type(comp._index_set) != type({}):
         # Singly indexed component
-        return comp._index
+        return comp._index_set
     else:
         # Unindexed constraint
         return {None: None}
--- a/pyomo/core/tests/transform/test_transform.py
+++ b/pyomo/core/tests/transform/test_transform.py
@@ -327,7 +327,7 @@ class Test(unittest.TestCase):
         for c in ('x', 'y', 'z'):
             for n in ('1', '2', '3', '4'):
                 var = transformed.__getattribute__(c+n)
-                for ndx in var._index:
+                for ndx in var._index_set:
                     self.assertTrue(self.nonnegativeBounds(var[ndx]))
 
     @unittest.skipIf(not 'glpk' in solvers, "glpk solver is not available")
--- a/pyomo/core/tests/transform/test_transform.py
+++ b/pyomo/core/tests/transform/test_transform.py
@@ -327,7 +327,7 @@ class Test(unittest.TestCase):
         for c in ('x', 'y', 'z'):
             for n in ('1', '2', '3', '4'):
                 var = transformed.__getattribute__(c+n)
-                for ndx in var._index_set:
+                for ndx in var.index_set():
                     self.assertTrue(self.nonnegativeBounds(var[ndx]))
 
     @unittest.skipIf(not 'glpk' in solvers, "glpk solver is not available")
--- a/pyomo/core/tests/unit/test_param.py
+++ b/pyomo/core/tests/unit/test_param.py
@@ -787,7 +787,6 @@ class ArrayParam6(unittest.TestCase):
         model.b = Set(initialize=['a','b','c'])
         model.c = model.b * model.b
         model.p = Param(model.a, model.c, within=NonNegativeIntegers, default=0, mutable=True)
-        #print(model.p._index_set.keys())
         model.p[1,'a','b'] = 1
         model.p[1,('a','b')] = 1
         model.p[(1,'b'),'b'] = 1
--- a/pyomo/core/tests/unit/test_param.py
+++ b/pyomo/core/tests/unit/test_param.py
@@ -787,7 +787,7 @@ class ArrayParam6(unittest.TestCase):
         model.b = Set(initialize=['a','b','c'])
         model.c = model.b * model.b
         model.p = Param(model.a, model.c, within=NonNegativeIntegers, default=0, mutable=True)
-        #print(model.p._index.keys())
+        #print(model.p._index_set.keys())
         model.p[1,'a','b'] = 1
         model.p[1,('a','b')] = 1
         model.p[(1,'b'),'b'] = 1
--- a/pyomo/dae/misc.py
+++ b/pyomo/dae/misc.py
@@ -119,7 +119,7 @@ def expand_components(block):
     # BlockData will be added to the indexed Block but will not be
     # constructed correctly.
     for blk in block.component_objects(Block, descend_into=True):
-        missing_idx = set(blk._index) - set(blk._data.keys())
+        missing_idx = set(blk._index_set) - set(blk._data.keys())
         if missing_idx:
             blk._dae_missing_idx = missing_idx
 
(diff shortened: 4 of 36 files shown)


pyomo/core/plugins/transform/util.py
```diff
@@ -176,9 +176,9 @@ def _getAbstractIndices(comp):
     """
     Returns the index or index set of this component
     """
-    if type(comp._index) != type({}):
+    if type(comp._index_set) != type({}):
         # Singly indexed component
-        return comp._index
+        return comp._index_set
```
jsiirola, line 181: This should really use the public `index_set()` method.

pyomo/core/tests/unit/test_param.py
```diff
@@ -787,7 +787,7 @@ def test_setitem(self):
         model.b = Set(initialize=['a','b','c'])
         model.c = model.b * model.b
         model.p = Param(model.a, model.c, within=NonNegativeIntegers, default=0, mutable=True)
-        #print(model.p._index.keys())
+        #print(model.p._index_set.keys())
```
jsiirola, line 790: This should just be deleted.

pyomo/core/tests/transform/test_transform.py
```diff
@@ -327,7 +327,7 @@ def domainRule(*args):
         for c in ('x', 'y', 'z'):
             for n in ('1', '2', '3', '4'):
                 var = transformed.__getattribute__(c+n)
-                for ndx in var._index:
+                for ndx in var._index_set:
```
jsiirola, line 330: Shouldn't this use the public `index_set()` method?

pyomo/dae/misc.py
```diff
@@ -119,7 +119,7 @@ def expand_components(block):
     # BlockData will be added to the indexed Block but will not be
     # constructed correctly.
     for blk in block.component_objects(Block, descend_into=True):
-        missing_idx = set(blk._index) - set(blk._data.keys())
+        missing_idx = set(blk._index_set) - set(blk._data.keys())
```
jsiirola, line 122: Shouldn't this use the public `index_set()` method?

pyomo/dae/misc.py
```diff
@@ -259,7 +259,7 @@ def _update_var(v):
     #       Var (which is now a IndexedComponent). However, it
     #       would be much slower to rely on that method to generate new
     #       _VarData for a large number of new indices.
-    new_indices = set(v._index) - set(v._data.keys())
+    new_indices = set(v._index_set) - set(v._data.keys())
```
jsiirola, line 262: Shouldn't this use the public `index_set()` method?

pyomo/dae/tests/test_diffvar.py
```diff
@@ -42,7 +42,7 @@ def test_valid(self):
         self.assertTrue(m.dv._sVar is m.v)
         self.assertTrue(m.v._derivative[('t',)]() is m.dv)
         self.assertTrue(m.dv.ctype is DerivativeVar)
-        self.assertTrue(m.dv._index is m.t)
+        self.assertTrue(m.dv._index_set is m.t)
```
jsiirola, line 45: Shouldn't this use the public `index_set()` method?

pyomo/dae/tests/test_flatten.py
```diff
@@ -586,13 +586,13 @@ def test_flatten_m2_2d(self):
 
         ref1 = Reference(m.v_2n[:,('c',3)])
 
-        ref_set = ref1._index._ref # _index is a _ReferenceSet
+        ref_set = ref1._index_set._ref # _index is a _ReferenceSet
```
jsiirola, line 589: Shouldn't this use the public `index_set()` method?

pyomo/dae/tests/test_flatten.py
```diff
@@ -586,13 +586,13 @@ def test_flatten_m2_2d(self):
...
 
         #next(ref_set._get_iter(ref_set._slice, ('a',1)))
         # ^ This raises a somewhat cryptic error message.
         # _ReferenceSet.__contains__ seems to be the culprit,
         # which is called in validate_index for every __getitem__.
-        self.assertNotIn(('a',1), ref1._index._ref)
+        self.assertNotIn(('a',1), ref1._index_set._ref)
```
jsiirola, line 595: Shouldn't this use the public `index_set()` method?

pyomo/dae/tests/test_misc.py
```diff
@@ -484,7 +484,7 @@ def _con2(_b, i, ti):
 
         generate_finite_elements(model.t, 5)
 
-        missing_idx = set(model.blk._index) - set(model.blk._data.keys())
+        missing_idx = set(model.blk._index_set) - set(model.blk._data.keys())
```
jsiirola, line 487: Shouldn't this (and the subsequent 3 instances) all use the public `index_set()` method?

pyomo/duality/lagrangian_dual.py
```diff
@@ -89,7 +89,7 @@ def _create_using(self, model, **kwds):
 
         # Walk constaints
         for (con_name, con_array) in sf.component_map(Constraint, active=True).items():
-            for con in (con_array[ndx] for ndx in con_array._index):
+            for con in (con_array[ndx] for ndx in con_array._index_set):
```
jsiirola, line 92: Shouldn't this (and the subsequent 3 instances) all use the public `index_set()` method?

jsiirola (changes requested):
In addition to the inline comments, it looks like there may be three files that may have been missed:
- `pyomo/repn/beta/matrix.py`
- `pyomo/core/base/matrix_constraint.py`
- `pyomo/core/expr/template_expr.py`

Also, as a stylistic thing, I generally prefer in this situation to use `NOTSET` instead of `NoArgumentGiven`

Finally, we should consider making a global `UnindexedComponent_index` (analogous to `pyomo.core.base.global_set.UnindexedComponent_set`) instead of all the hard-coded `None` (e.g., `self._index = UnindexedComponent_index` instead of `self._index = None`)

emma58:
@jsiirola I think everything is addressed (assuming the tests pass)

jsiirola (approved):

mrmundt (approved):

PR #2360, opened 2022-04-14
Improved processing of strings passed to `ConfigList`, `ConfigDict`, and `ListOf`

jsiirola:
## Summary/Motivation:
@DLWoodruff pointed out that `ListOf` produces nonintuitive behavior when passed strings: Given a `ListOf(int, PositiveInt)`, passing it the string `'42'` results in the list `[4, 2]`.  This is particularly problematic when using the coupling between the Config system and argparse (where all data comes in as strings).

This PR adds a reasonable parser for interpreting string data: tokens are split using either commas or whitespace, and quotes (single or double) are honored for embedding spaces, quotes, and commas into strings.  An extension of this parser also works for `key=value` (or `key:value`).  

Note that this changes behavior slightly: `ListOf(str)` used to ignore all whitespace or commas and would create a list with a single entry.  After this PR, the string will be parsed and broken up into "words".  The previous behavior can be restored by enclosing the string in quotes.  To my knowledge, no part of Pyomo (apart from one test) relied on the old behavior.

- Parse string data passed to `ListOf`, `ConfigList`, and `ConfigDict` objects
- Improve reporting of domain names for complex domain validators
- Update testing

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/config.py
+++ b/pyomo/common/config.py
@@ -24,6 +24,7 @@ import io
 import logging
 import os
 import pickle
+import ply.lex
 import re
 import sys
 from textwrap import wrap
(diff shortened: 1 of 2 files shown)


DLWoodruff (approved):
This is very clever.

blnicho (approved):

PR #2364, opened 2022-04-19
Update Actions and PyPy Version

mrmundt:
## Summary/Motivation:
There are newer versions of the `checkout` and `setup-python`, the latter of which allows use of newer `pypy` versions.

- Update `setup-python` and `checkout` to `v3`
- Update `pypy3` to `pypy-3.9`

### Legal Acknowledgement


blnicho (approved):

jsiirola (approved):

mrmundt:
@jsiirola @blnicho 

Alright, so there are two failures. One is a test that isn't supposed to be running on `pypy` but is now triggering. The other, however...
```
> self.assertIs(config.init_strategy, 'FP') 
E AssertionError: 'FP' is not 'FP' 

pyomo/contrib/mindtpy/tests/unit_test.py:366: AssertionError
```
Seems like `FP` is `FP` me, but what do I know.

jsiirola:
> Seems like `FP` is `FP` me, but what do I know.

That should be `assertEqual` and not `assertIs` (the test is failing because PyPy is not as good about reusing string objects and there are two *diferent* "FP" string objects).

jsiirola:
The other failing test needs to be skipped on PyPy (`-X importtime` is a CPython thing that PyPy does not support).  I went ahead and pushed these changes to the PR.  Hopefully tests will start passing.

jsiirola:
It looks like we are down to just the MPI failure...

mrmundt:
Is that something we wait on, or knowing that it's currently problematic, do we merge?

mrmundt:
@jsiirola @blnicho @michaelbynum - I created an issue with `openmpi` about the breakage: https://github.com/open-mpi/ompi/issues/10300

.github/workflows/test_branches.yml
```diff
@@ -66,7 +66,7 @@ jobs:
           skip_doctest: 1
           TARGET: linux
           PYENV: conda
-          PACKAGES: mpi4py openmpi
+          PACKAGES: mpi4py
```
jsiirola, line 69: I think if you are removing openmpi, you can / should also remote mpi4py.

jsiirola: Correction: we need to leave openmpi and mpi4py here.  It can be removed from the `ubuntu-latest/3.9/conda` above.

PR #2370, opened 2022-04-21
Deprecate name_buffer argument to getname

emma58:
## Summary/Motivation:

Now that ComponentData objects know their own indices, the name_buffer argument to getname is no longer necessary to get good performance. In addition, it returns the wrong name if the same dictionary was specified as the name_buffer in queries that switched where the the model the names where relative to or switched between local and fully-qualified names.

- Deprecates the `name_buffer` argument to `getname`
- Deprecates the `remove_obj` function in the Labeler classes since all it did was update the name_buffer
- Adds a check in the `index` method that the `_index` attribute and the `_data` dictionary are in sync and raises an error if not
- Fixes bugs that the addition above exposed
- Fixes tests that wrongly relied on the bugs in name_buffer, mainly in the `hull` transformation in pyomo.gdp

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/component.py
+++ b/pyomo/core/base/component.py
@@ -851,31 +851,17 @@ class ComponentData(_ComponentBase):
         to the parent component's index set.
         """
         parent = self.parent_component()
-        if parent is None:
-            return None
-        idx = self._index
-        # ComponentList objects implement _data as a list rather than a dict so
-        # we have to account for both here
-        if isinstance(parent._data, dict):
-            parent_idx = parent._data.get(idx, None)
-        elif isinstance(parent._data, list):
-            try:
-                parent_idx = parent._data[idx]
-            except IndexError:
-                parent_idx = None
-        else:
-            raise DeveloperError("Unrecognized type for '_data' dictionary")
-        if ((idx is NOTSET and parent_idx is not None) or
-            (parent_idx is not self)):
-            parent_idx_name = parent_idx.name if parent_idx is not None \
-                              else 'None'
+        if ( parent is not None and
+             self._index is not NOTSET and
+             parent[self._index] is not self ):
             # This error message is a bit goofy, but we can't call self.name
             # here--it's an infinite loop!
             raise DeveloperError(
                 "The '_data' dictionary and '_index' attribute are out of "
                 "sync for index '%s' on indexed component '%s': The '_data' "
                 "dictionary contains '%s' as the value corresponding to '%s'."
-                % (idx, parent.name, parent_idx_name, idx))
+                % (self._index, parent.name, parent[self._index].name,
+                   self._index))
         return self._index
 
     def __str__(self):
--- a/pyomo/core/base/component.py
+++ b/pyomo/core/base/component.py
@@ -853,6 +853,13 @@ class ComponentData(_ComponentBase):
         if parent is None:
             return None
         idx = self._index
+        if idx is NOTSET:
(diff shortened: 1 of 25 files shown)


pyomo/core/base/component.py
```diff
@@ -846,8 +850,32 @@ def index(self):
...
         to the parent component's index set.
         """
-        if self.parent_component() is None:
+        parent = self.parent_component()
+        if parent is None:
             return None
+        idx = self._index
+        # ComponentList objects implement _data as a list rather than a dict so
+        # we have to account for both here
+        if isinstance(parent._data, dict):
```
jsiirola, line 859: Would a simpler test just be to:
```python
if parent[self._index] is not self:
    raise DeveloperError(...)
return self._index
```

jsiirola (changes requested):
A couple questions / changes.  Otherwise this looks good.

pyomo/core/base/component.py
emma58: You get a KeyError if self._index is NOTSET. So, it serves the purpose, but the error message isn't very good.

jsiirola: Ok, but that should be easy to correct, right?
```python
if self._index is not NOTSET and parent[self._index] is not self:
    raise DeveloperError(...)
return self._index
```
That does raise the question: should `NOTSET` raise an exception (which is the current implementation), or should it just be returned?  I think I am leaning toward the latter (and is proposed in this comment).

emma58: OK, you convinced me, but just documenting here for my future self: We decided that `NOTSET` should not raise an exception, and neither should not having a parent: both cases should just return `self._index`. The only case that is necessarily an error is if the `_index` on the ComponentData is mapped to something different in the parent's `_data` dictionary. (Though more restrictive checks might be useful for debugging.)

emma58:
@jsiirola I think I addressed all your comments.

blnicho (approved):

jsiirola (approved):

pyomo/core/tests/unit/test_indexed.py
```diff
@@ -200,6 +201,21 @@ def test_ordered_keys(self):
...
+        for i in [1, 2, 3]:
+            self.assertEqual(m.x[i].index(), i)
+        # now mess it up
+        m.x[3]._index = 2
+        with self.assertRaisesRegex(
+                DeveloperError,
+                ".*The '_data' dictionary and '_index' attribute are out of "
+                "sync for index '2' on indexed component 'x': The '_data' "
+                "dictionary contains 'x\[2\]' as the value corresponding to "
+                "'2'"):
```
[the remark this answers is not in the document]
jsiirola: I think this is correct, but it is a good point that the message is confusing.  @emma58, what do you think about a message more like:

```python
"The '_data' dictionary and '_index' attribute are out of sync for indexed "
"%s '%s': The %s entry in the '_data' dictionary does not map back to this "
"component data object' % (obj.ctype.__name__, obj.parent_component().name, obj._index) 
```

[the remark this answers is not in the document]
emma58: Yes, @blnicho, they are both `2`'s because you would have to iterate over the `_data` dictionary to get the `3` (returning to the old inefficiency). But I like @jsiirola's suggestion--I'll make the change.

jsiirola (approved):

PR #2400, opened 2022-05-13
Updating MC++ Interface Tests

michaelbynum:
## Summary/Motivation:
This PR updates a couple MC++ interface tests. In `test_lmtd`, instead of ensuring the convex underestimator does not change, I test that the convex underestimator is an underestimator at several points. I also check that the convex underestimator matches the original function at the lower and upper bounds of the variables. I do the same for the overestimators. 

### Legal Acknowledgement


jsiirola (approved):

michaelbynum:
That is unfortunate. Yes - it should be resolvable. Let me take a look.

PR #2406, opened 2022-05-17
GDPopt rewrite!

emma58:
## Summary/Motivation:

This is a rewrite of GDPopt with the goal of improving extensibility and maintainability.

- Removes gdpbb--it has been deprecated since 5.6.9
- Centralizes common functionality between the GDPopt algorithms in one class, but has each algorithm implement its own main loop and other unique methods (usually cut generation).
- Stores algorithm state (bounds, timing info, etc) on the GDPoptSolver class in lieu of the former `GDPoptSolveData' object.
    - Moves some util functions to `mindtpy` as a result, as gdpopt no longer uses them
 - Significantly changes the behavior of `tee=True` similarly to recent changes in mindtpy: simplifies output at the `INFO` level to log bounds by iteration, with some info before and after the solve. The old detailed output (and more) is now at the `DEBUG` level.
 - Makes the function to initialize NLP subproblems a callback. The default is still the former behavior of using the solution originally in the model passed to the solver.
 - Deprecates the 'strategy' argument in favor of 'algorithm'

NOTE: The changes to LBB are fairly minimal--just divorcing it from the `solve_data` object. But otherwise it is largely exempt from this rewrite.

### Legal Acknowledgement


pyomo/contrib/gdpopt/GDPopt.py
```diff
@@ -154,10 +590,6 @@ def __exit__(self, t, v, traceback):
 
     def available(self, exception_flag=True):
         """Check if solver is available.
```
bernalde, line 592: Why is this comment removed? Should the solver be available even if no subsolvers are? Is there ever going to be a case where GDPOpt is unavailable (as it comes directly with Pyomo)?

pyomo/contrib/gdpopt/branch_and_bound.py
```diff
@@ -37,442 +45,470 @@
...
+                    root_util_blk.disjunct_to_nonlinear_constraints[
+                        disjunct] = nonlinear_constraints_in_disjunct
+
+            root_util_blk.disjunction_to_unfixed_disjuncts[
+                disjunction] = unfixed_disjuncts
+            pass
+
+        # Add the BigM suffix if it does not already exist. Used later during
+        # nonlinear constraint activation.
+        # TODO is this still necessary?
```
bernalde, line 149: Would this rewrite answer this question?

pyomo/contrib/gdpopt/branch_and_bound.py
```diff
@@ -37,442 +45,470 @@
...
+            node_count=0,
+            unbranched_disjunction_indices=unbranched_disjunction_indices,
+        )
+        heappush(queue, (sort_tuple, root_node))
+
+        # Do the branch and bound
+        while len(queue) > 0:
+            # visit the top node on the heap
+            # from pprint import pprint
+            # pprint([(
```
bernalde, line 174: I think these comments could be removed

pyomo/contrib/gdpopt/branch_and_bound.py
```diff
@@ -37,442 +45,470 @@
...
+            copy_var_list_values(
+                from_list=subprob_utils.algebraic_variable_list,
+                to_list=model_utils.algebraic_variable_list,
+                config=config, ignore_integrality=True
+            )
+            return float('inf'), float('inf')
+
+
+    def _solve_local_rnGDP_subproblem(self, model, config):
+        # TODO for now, return (LB, UB) = (-inf, inf) (for minimize)
```
bernalde, line 444: This could be addressed now

pyomo/contrib/gdpopt/config_options.py
```diff
@@ -9,60 +9,23 @@
...
-
+from pyomo.contrib.gdpopt.nlp_initialization import (
+    restore_vars_to_original_values)
+from pyomo.contrib.gdpopt.util import _DoNothing
+from pyomo.opt import SolverFactory, UnknownSolver
+from pyomo.solvers.plugins.solvers.persistent_solver import PersistentSolver
 
 def _add_OA_configs(CONFIG):
     CONFIG.declare("init_strategy", ConfigValue(
         default="set_covering", domain=In(valid_init_strategies.keys()),
```
bernalde, line 24: Since the `strategies` were rebranded as `algorithms`, should the initialization ones be renamed as well?

pyomo/contrib/gdpopt/loa.py
```diff
@@ -0,0 +1,266 @@
...
+            if self.parent.any_termination_criterion_met(config):
+                break
+
+    def _setup_augmented_Lagrangian_objective(self, master_util_block):
+        m = master_util_block.model()
+        main_objective = next(m.component_data_objects(Objective, active=True))
+
+        # Set up augmented Lagrangean penalty objective
+        main_objective.deactivate()
+        # placeholder for oa objective
```
bernalde, line 119: OA

pyomo/contrib/gdpopt/nlp_initialization.py
```diff
@@ -0,0 +1,34 @@
...
+#  rights in this software.
+#  This software is distributed under the 3-clause BSD License.
+#  ___________________________________________________________________________
+"""A library of possible callbacks to use for initializing the NLP subproblems.
+However, it seems likely that problem-specific routines might be better, in
+which case you can write your own, and specify it in the
+'subproblem_initialization_method' argument."""
+
+# This is the original GDPopt behavior:
+def restore_vars_to_original_values(nlp_util_block, mip_util_block):
```
bernalde, line 17: May I suggest another name besides `original`? I would go use `initial` instead. It just sounds odd to me

pyomo/contrib/gdpopt/nlp_solve.py
```diff
@@ -10,130 +10,37 @@
 #  ___________________________________________________________________________
 
 """Functions for solving the nonlinear subproblem."""
-from __future__ import division
-
-from math import fabs
+from contextlib import contextmanager
```
bernalde, line 13: Since this file has functions to solve linear subproblems I would suggest to rename it to `solve_subproblem.py` or `solve_fixed_subproblem.py`

pyomo/contrib/gdpopt/nlp_solve.py
```diff
@@ -310,288 +193,135 @@ def detect_unfixed_discrete_vars(model):
...
+        fbbt(m, integer_tol=config.integer_tolerance,
+             feasibility_tol=config.constraint_tolerance,
+             max_iter=config.max_fbbt_iterations)
+        xfrm = TransformationFactory
+        # Now that we've tightened bounds, see if any variables are fixed
+        # because their lb is equal to the ub (within tolerance)
+        xfrm('contrib.detect_fixed_vars').apply_to(
+            m, tolerance=config.variable_tolerance)
+
+        # Restore the original bounds because the NLP solver might like that
```
bernalde, line 229: NLP -> Subproblem

pyomo/contrib/gdpopt/nlp_solve.py
```diff
@@ -310,288 +193,135 @@ def detect_unfixed_discrete_vars(model):
...
-    # Last, check if any constraints are now trivial and deactivate them
-    xfrm('contrib.deactivate_trivial_constraints').apply_to(
-        m, tolerance=config.constraint_tolerance)
-
-def initialize_subproblem(model, solve_data):
-    """Perform initialization of the subproblem.
-
-    Presently, this just restores the continuous variables to the original 
-    model values.
+    # A bit counter-intuitively (but I assume so that it can propogate those
```
bernalde, line 259: Typo: propagate

pyomo/contrib/gdpopt/util.py
```diff
@@ -244,48 +145,122 @@ def a_logger(str_or_logger):
...
+        fixed_discrete = []
+        for master_var, subprob_var in zip(
+                master_util_block.discrete_variable_list,
+                subproblem_util_block.discrete_variable_list):
+            # [ESJ 1/24/21]: We don't check if master_var actually has a value
+            # here because we are going to have to do that error checking
+            # later. This is because the subproblem could have discrete
+            # variables that aren't in the master and vice versa since master is
+            # linearized, but subproblem is a specific realization of the
+            # disjuncts. All this means we don't have enough info to do it here.
```
bernalde, line 234: Agreed

pyomo/contrib/gdpopt/util.py
```diff
@@ -244,48 +145,122 @@ def a_logger(str_or_logger):
...
+        bool_var.get_associated_binary().unfix()
+
+    # unfix all discrete variables and restore them to their original values
+    if make_subproblem_continuous:
+        for var in subproblem_util_block.discrete_variable_list:
+            subprob_var.fixed = False
+
+    # [ESJ 2/25/22] I think we don't need to reset the values of the continuous
+    # variables because we will initialize them based on the master solution
+    # before we solve again.
```
bernalde, line 263: I agree as well

bernalde (changes requested):
Very exciting work Emma! This is my first pass and the comments I left behind are mainly esthetic. It would be fantastic if @ZedongPeng (and potentially @qtothec ) could review the PR, as well as other Pyomo maintainers.
Many of these things can be used in MindtPy so after this is merged we can expect a similar PR for MindtPy coming in.

pyomo/contrib/gdpopt/GDPopt.py
emma58: I think it should always be available, rather than being a function of if the subsolvers are. Without knowing the algorithm in advance of the call to `solve` (which we allow), it's not possible to know *which* subsolvers are necessary. We do fail gracefully when the subsolvers aren't available, so this seems okay to me.

pyomo/contrib/gdpopt/branch_and_bound.py
emma58: This I think could be necessary, though probably rarely. The bigm transformation now uses fbbt to calculate M values, so it is successful more often than not now, but this implementation makes sense to me since there is no reason to let that fail (or even run) for a constraint where we know that the `indicator_var` is fixed to `True`.

pyomo/contrib/gdpopt/branch_and_bound.py
emma58: I don't actually understand this TODO... What are you thinking should be changed? In general I think we should save a rewrite of this algorithm for a separate PR, but I'm happy to make some small changes.

pyomo/contrib/gdpopt/nlp_initialization.py
emma58: I'm not sure I like using `initial` because we could set the values to something other than what was in the original (given) model and we would still call whatever we set them to the "initial value." Would `given` or `user_given` sound better to you?

pyomo/contrib/gdpopt/nlp_solve.py
emma58: Sure, good point. I think I'll rename `mip_solve` too then, to match.

pyomo/contrib/gdpopt/config_options.py
emma58: Yes, I think this makes sense.

ZedongPeng:
Excellent work!

pyomo/contrib/gdpopt/branch_and_bound.py
emma58: Ah, okay, I will clarify the comment then. Thanks!

pyomo/contrib/gdpopt/tests/test_gdpopt.py
```diff
@@ -826,27 +1195,93 @@ def test_RIC_custom_disjuncts(self):
              eight_process.use_unit_8ornot.disjuncts[0]]
         ]
 
-        def assert_correct_disjuncts_active(nlp_model, solve_data):
-            if solve_data.master_iteration >= 1:
+        def assert_correct_disjuncts_active(subprob_util_block,
+                                            master_util_block):
+            # I can get the iteration based on the number of no-good
+            # cuts in this case...
+            iteration = len(master_util_block.no_good_cuts)
```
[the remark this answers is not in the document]
emma58: This is a good point. I just made a change so that all the callbacks accept the solver object as an argument, and then you can get into whatever trouble you want... :)

pyomo/contrib/gdpopt/branch_and_bound.py
```diff
@@ -37,442 +46,483 @@
...
+                config=config,
+            )
+            return lb, ub
+        elif term_cond == tc.locallyOptimal or term_cond == tc.feasible:
+            assert result.solver.status is SolverStatus.ok
+            lb = result.problem.lower_bound if not obj_sense_correction else \
+                 -result.problem.upper_bound
+            ub = result.problem.upper_bound if not obj_sense_correction else \
+                 -result.problem.lower_bound
+            # TODO handle LB absent
```
bernalde, line 419: Is this TODO still relevant?

pyomo/contrib/gdpopt/branch_and_bound.py
```diff
@@ -37,442 +46,483 @@
...
+                config=config, ignore_integrality=True
+            )
+            return float('inf'), float('inf')
+
+
+    def _solve_local_rnGDP_subproblem(self, model, config):
+        # TODO: The returns of this method should be improved. Currently, it
+        # returns trivial bounds (LB, UB) = (-inf, inf) if there is an error in
+        # the solve, if the problem is infeasible, or if the problem is
+        # unbounded.
```
bernalde, line 455: Is there anything better we could do in these cases? I think we could do better in the infeasible case such that the branch is "closed" but besides that, the unbounded and error behaviors is what I would expect.

pyomo/contrib/gdpopt/config_options.py
```diff
@@ -9,21 +9,45 @@
...
-    _supported_strategies = {
-        'LOA',  # Logic-based outer approximation
-        'GLOA',  # Global logic-based outer approximation
-        'LBB',  # Logic-based branch-and-bound
-        'RIC',  # Relaxation with Integer Cuts
-    }
-    CONFIG = ConfigBlock("GDPopt")
+def _strategy_deprecation(strategy):
+    deprecation_warning("The argument 'strategy' has been deprecated "
+                        "in favor of 'algorithm.'", version="TBD")
```
bernalde, line 30: TBD can be changed to this PR version

pyomo/contrib/gdpopt/config_options.py
```diff
@@ -9,21 +9,45 @@
...
+    return In(_supported_algorithms.keys())(strategy)
+
+def _init_strategy_deprecation(strategy):
+    deprecation_warning("The argument 'init_strategy' has been deprecated "
+                        "in favor of 'init_algorithm.'", version="TBD")
```
bernalde, line 35: Same as above

pyomo/contrib/gdpopt/config_options.py
```diff
@@ -32,13 +56,10 @@ def _get_GDPopt_config():
...
-            "currently only be enforced between subsolver invocations. You may "
-            "need to set subsolver time limits as well."
-    ))
-    CONFIG.declare("strategy", ConfigValue(
-        default=None, domain=In(_supported_strategies),
-        description="Decomposition strategy to use."
+        doc="""
+        Seconds allowed until terminated. Note that the time limit can
+        currently only be enforced between subsolver invocations. You may
+        need to set subsolver time limits as well."""
```
bernalde, line 62: This does not seem right. Usually because subsolver calls can be really long. In mindtpy we keep computing remaining times that are passed down to subsolvers.

pyomo/contrib/gdpopt/config_options.py
```diff
@@ -75,136 +96,228 @@ def _add_OA_configs(CONFIG):
...
+        Note that unless you are *very* confident in what you are doing, the
+        subproblem should not be modified in this callback: it should be used
+        to interrogate the problem only.
+        """
     ))
     CONFIG.declare("round_discrete_vars", ConfigValue(
         default=True,
-        description="flag to round subproblem discrete variable values to the "
-        "nearest integer. Rounding is done before fixing disjuncts."
+        description="""flag to round subproblem discrete variable values to the
```
bernalde, line 206: Uppercase: Flag

pyomo/contrib/gdpopt/config_options.py
```diff
@@ -75,136 +96,228 @@ def _add_OA_configs(CONFIG):
...
     ))
 
 
-def _add_subsolver_configs(CONFIG):
+def _add_mip_solver_configs(CONFIG):
     CONFIG.declare("mip_solver", ConfigValue(
         default="gurobi",
-        description="Mixed integer linear solver to use."
+        description="""
+        Mixed integer linear solver to use. Note that no persisent solvers
```
bernalde, line 278: Typo: Mixed-integer

pyomo/contrib/gdpopt/config_options.py
```diff
@@ -215,17 +328,19 @@ def _add_tolerance_configs(CONFIG):
     ))
     CONFIG.declare("small_dual_tolerance", ConfigValue(
         default=1E-8,
-        description="When generating cuts, small duals multiplied "
-                    "by expressions can cause problems. Exclude all duals "
-                    "smaller in absolue value than the following."
+        description="""
+        When generating cuts, small duals multiplied by expressions can
+        cause problems. Exclude all duals  smaller in absolue value than the
```
bernalde, line 333: Double space here

pyomo/contrib/gdpopt/config_options.py
```diff
@@ -75,136 +96,228 @@ def _add_OA_configs(CONFIG):
...
-    ))
-    CONFIG.declare("algorithm_stall_after", ConfigValue(
-        default=2,
-        description="number of non-improving master iterations after which "
-                    "the algorithm will stall and exit."
+        description="""
+        callback hook after feasible solution of
+        the nonlinear subproblem""",
+        doc="""
+        Callback called right after the (MI)NLP subproblem is solved,
```
bernalde, line 195: Do we like acronyms in the option descriptions? iF not, edit NLP, MILP, MINLP, ... in other options.

pyomo/contrib/gdpopt/gloa.py
```diff
@@ -0,0 +1,225 @@
...
+            if self.any_termination_criterion_met(config):
+                break
+
+    # Utility used in cut_generation: We saved a map of Disjuncts to the
+    # active constraints they contain on the master problem util_block, and use
+    # it here to find the active constraints under the current discrete
+    # solution. Note that this preprocesses not just to be efficient, but
+    # because everything on the Disjuncts is deactivated at this point, since
+    # we've already transformed the master problem to a MILP
+    def _get_active_untransformed_constraints(self, util_block, config):
```
bernalde, line 127: Is this utility only used here? Maybe moving this to utils.py might make sense as it can be useful for LOA as well.

pyomo/contrib/gdpopt/loa.py
```diff
@@ -0,0 +1,281 @@
...
+    'gdpopt.loa',
+    doc="The LOA (logic-based outer approximation) Generalized Disjunctive "
+    "Programming (GDP) solver")
+class GDP_LOA_Solver(_GDPoptAlgorithm, _OAAlgorithmMixIn):
+    """The GDPopt (Generalized Disjunctive Programming optimizer) logic-based
+    outer approximation (LOA) solver.
+
+    Accepts models that can include nonlinear, continuous variables and
+    constraints, as well as logical conditions. For nonconvex problems, LOA
+    may not report rigorous lower/upper bounds.
```
bernalde, line 52: 'optimistic' bounds, as the 'feasible' bound is going to be rigorous

pyomo/contrib/gdpopt/loa.py
```diff
@@ -0,0 +1,281 @@
...
+    """
+    CONFIG = _GDPoptAlgorithm.CONFIG()
+    _add_OA_configs(CONFIG)
```
bernalde, line 55: This is just being picky, but all the other configs are lower case. Why is 'OA' not?

pyomo/contrib/gdpopt/loa.py
```diff
@@ -0,0 +1,281 @@
...
+
+            # Add integer cut
+            with time_code(self.timing, "integer cut generation"):
+                add_no_good_cut(master_util_block, config)
+
+            # Check termination conditions
+            if self.any_termination_criterion_met(config):
+                break
+
+    def _setup_augmented_Lagrangian_objective(self, master_util_block):
```
bernalde, line 125: I think this is called augmented penalty in the outer-approximation literature as the we are not moving the actual constraint to the objective, but I can see this being an augmented Lagrangian (or Lagrangean according to the comments).

pyomo/contrib/gdpopt/solve_master_problem.py
```diff
@@ -0,0 +1,185 @@
...
+                                             remaining)
+            results = SolverFactory(config.mip_solver).solve(m, **mip_args)
+
+    except RuntimeError as e:
+        # ESJ: I don't know enough about GAMS to know if this is a good
+        # idea. Does this always mean it's infeasible?
+        if 'GAMS encountered an error during solve.' in str(e):
+            config.logger.warning(
+                "GAMS encountered an error in solve. Treating as infeasible.")
+            return tc.infeasible
```
bernalde, line 81: This is making a big assumption that a runtime error in GAMS is only coming from a trivially infeasible problem (there might be other causes, but if we assume perfect translation of the model from GAMS.py, and perfectly working GAMS installation) we can say that the problem is trivially infeasible. At this point, FBBT should've told you that I guess, I would be in favor of changing tc.infeasible to something else. tc.error maybe?

pyomo/contrib/gdpopt/solve_master_problem.py
```diff
@@ -0,0 +1,185 @@
...
+                               "feasible solution.")
+
+    if terminate_cond is tc.optimal:
+        return tc.optimal
+    elif terminate_cond in {tc.locallyOptimal, tc.feasible}:
+        return tc.feasible
+    elif terminate_cond is tc.infeasible:
+        config.logger.info(
+            'MILP master problem is now infeasible. '
+            'GDPopt has finished exploring feasible discrete configurations.')
```
bernalde, line 135: *might* have finished exploring feasible discrete configurations. Invalid cuts coming from LOA might lead to infeasible master without enumerating.

pyomo/contrib/gdpopt/solve_master_problem.py
```diff
@@ -0,0 +1,185 @@
...
+
+    """
+    tmp_args = deepcopy(config.mip_solver_args)
+    if config.mip_solver == 'gurobi':
+        # This solver option is specific to Gurobi.
+        tmp_args['options'] = tmp_args.get('options', {})
+        tmp_args['options']['DualReductions'] = 0
+    mipopt = SolverFactory(config.mip_solver)
+    # gdpopt no longer supports non-auto persistent solvers, but mindtpy does,
+    # and it uses this function.
```
bernalde, line 179: Why not? Would it fail is I call a persistent solver as a MIP solver in GDPOpt?

bernalde (commented):
I'm just leaving some comments after checking what has been changed here.
Great PR!

pyomo/contrib/gdpopt/branch_and_bound.py
emma58: I'm honestly not sure--It doesn't seem like it can be because the lower bound isn't used in that `elif` even if the solver does return one. I'll remove it (and the useless `lb` variable).

pyomo/contrib/gdpopt/branch_and_bound.py
emma58: It's apparently not tested... I just changed the infeasible return to be `float('inf'), float('inf')` and nothing broke. I agree with you that that seems more reasonable (since the problem is always corrected to be a minimization). But I'm a little nervous to changes here without taking the time to figure out why it was this way in the first place. Which is an exercise I would really like to defer to a future PR.

pyomo/contrib/gdpopt/config_options.py
emma58: Good point, I will make that change.

pyomo/contrib/gdpopt/solve_master_problem.py
emma58: Yes, it will fail because it will never call `set_instance`. There has never been any real support for using a persistent solver--the previous implementation just checked if the solver was persistent, then called `set_instance` every iteration if it was. I took that out because there's no point--you can use a direct solver if that's the behavior you want.

pyomo/contrib/gdpopt/solve_master_problem.py
emma58: OK, thank you for your input here. I think I'll change to tc.error.

pyomo/contrib/gdpopt/config_options.py
emma58: Sorry, @bernalde, I wasn't thinking very clearly the other day: How do you do this when you don't know what solver is being used? Could you point me to where that happens in mindtpy?

pyomo/contrib/gdpopt/config_options.py
emma58: I'm neutral. I kind of like it on the assumption that anyone using advanced configurations in GDPopt will probably know what we mean, and the paren thing actually makes the algorithm choices clearer. But if you think writing things out is better, I'm happy to.

pyomo/contrib/gdpopt/loa.py
emma58: Hehe, beats me!

pyomo/contrib/gdpopt/loa.py
emma58: Oh, good to know, thank you. I think I'll change to match the OA literature. That seems clearer.

pyomo/contrib/gdpopt/branch_and_bound.py
bernalde: SOunds good!

pyomo/contrib/gdpopt/branch_and_bound.py
bernalde: Perfect. Maybe then add a comment for future developers or an issue here in GitHub for us to remember.

pyomo/contrib/gdpopt/config_options.py
bernalde: I'm neutral as well. It just seemed a little mixed-up. Feel free to ignore this comment.

pyomo/contrib/gdpopt/gloa.py
bernalde: I think that this can wait for a future PR, maybe even consider having the same util for MindtPy. There is going to be a MindtPy rewrite coming after this is merged where we can see where things make sense together.

pyomo/contrib/gdpopt/config_options.py
bernalde: It is actually a GDPOpt util https://github.com/Pyomo/pyomo/blob/6ecb587581bd64bc5756b71662c7a66b1211632a/pyomo/contrib/gdpopt/util.py#L517

bernalde:
This might sound not too relevant, but recently in a review, I was asked to refer to the master problem as the main problem. This is something that has already been implemented in the GitHub branches and that we included in MindtPy. Since you are relabeling things, we can put this there as well.

pyomo/contrib/gdpopt/config_options.py
emma58: Yes, but I couldn't find in Mindtpy where the time remaining is passed to the subsolver. I'm unsure (short of a huge string of `if` statements) of how to do this in a general way because the option names are solver-specific. For example, the option is `timelimit` in cplex but `tmlim` in glpk, I think.

bernalde: You are right! We have discussed this briefly with @jsiirola. Whether it makes sense to have an alias for the important options for all solvers, but for the time being, I guess that we cannot consider the remaining time and just hope for the best with each suproblem.

pyomo/contrib/gdpopt/solve_master_problem.py
emma58: Actually, on second thought, I'm going to let it raise the error. If we think it really is one, then the try-except is just kicking the can down the road--this will fail when updating the dual bound anyway, if we return tc.error.

pyomo/contrib/gdpopt/util.py
```diff
@@ -253,41 +153,126 @@ def copy_var_list_values(from_list, to_list, config,
...
+            "is not within the specified integer tolerance of %s." %
+            (var.name, val, config.integer_tolerance))
+    else:
+        # variable is integer and within tolerance
+        if config.round_discrete_vars:
+            var.fix(int(round(val)))
+        else:
+            var.fix(val, skip_validation=True)
+
+@contextmanager
```
jsiirola, line 190: This is a huge context manager.  We should make it an explicit class with `__enter__` / `__exit__` methods to make this more readable.

jsiirola (commented):
In addition to specific comments, we need to go through and ensure the Pyomo copyright header is in all `.py` files.

pyomo/contrib/gdpopt/algorithm_base_class.py
```diff
@@ -0,0 +1,470 @@
...
+            if original_binary is not None:
+                original_binary.set_value(soln)
+        StaleFlagManager.mark_all_as_stale(delayed=True)
+
+    def _delete_original_model_util_block(self):
+        """For cleaning up after a solve--we want the original model to be
+        untouched except for the solution being loaded"""
+        blk = self.original_util_block
+        m = blk.model()
+        m.del_component(blk)
```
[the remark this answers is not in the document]
emma58: Eeeek, good catch!

pyomo/contrib/gdpopt/branch_and_bound.py
```diff
@@ -9,13 +9,22 @@
 #  This software is distributed under the 3-clause BSD License.
 #  ___________________________________________________________________________
 
-import traceback
 from collections import namedtuple
 from heapq import heappush, heappop
+import traceback
```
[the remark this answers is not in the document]
emma58: Ummm, I guess not? It gets used in some of the logger warnings, though I'm not really sure why?

pyomo/contrib/gdpopt/util.py
```diff
@@ -79,163 +73,70 @@ def __exit__(self, exception_type, exception_value, traceback):
...
+        logger.info("Your model is an LP (linear program). "
+                    "Using LP solver %s to solve." % config.mip_solver)
+        results = SolverFactory(config.mip_solver).solve(
+            m, **config.mip_solver_args)
+        return results
+
+def move_nonlinear_objective_to_constraints(util_block, logger):
+    m = util_block.model()
+    principal_obj = next(m.component_data_objects(Objective, descend_into=True,
+                                                  active=True))
```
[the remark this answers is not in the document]
emma58: This is a really good point. This was everywhere--I don't think we ever want to ask for `model()`--we just want to pretend that what we were given is the model, and we can find it by asking for the `parent_block` of the util block.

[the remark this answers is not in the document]
jsiirola: I wonder if we should ever consider a mode where we take the incoming block to be solved and, if it was not the root model(), temporarily clear its parent pointer — effectively making it a temporary “model()”?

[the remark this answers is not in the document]
emma58: Oooo, that would be a cool context manager. It would be useful for things like this. Because it's nice to pretend...

pyomo/contrib/gdpopt/GDPopt.py
```diff
@@ -88,76 +95,55 @@ class GDPoptSolver(object):
     - Logic-to-linear transformation: Romeo Valentin
 
     """
-
-    # Declare configuration options for the GDPopt solver
-    CONFIG = _get_GDPopt_config()
+    CONFIG = ConfigBlock("GDPopt")
+    _add_common_configs(CONFIG)
```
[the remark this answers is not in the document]
emma58: They almost could, but `GDPoptSolver` and `_GDPoptAlgorithm` have different `solve` methods: `GDPoptSolver` only exists for backwards compatibility--to support instantiating a solver without specifying the algorithm. Then its `solve` method creates an ephemeral instance of whatever actual solver it needs depending on the algorithm. So I guess we could do a mixin or something, but I kind of like keeping `GDPoptSolver` really minimal because eventually we might just deprecate it and just force people to choose an algorithm?

pyomo/contrib/gdpopt/algorithm_base_class.py
```diff
@@ -0,0 +1,470 @@
...
+                results = self._gather_problem_info_and_solve_non_gdps(model,
+                                                                       config)
+                # If it wasn't disjunctive, we solved it
+                if results:
+                    return results
+                else:
+                    # main loop implemented by each algorithm
+                    self._solve_gdp(model, config)
+
+            self._get_final_pyomo_results_object()
```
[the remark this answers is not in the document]
emma58: No good reason, I don't think, but because the partially-filled-in results object is already stored on the solver. It gets created early with the `_get_pyomo_results_object_with_problem_info` method. I'm confused about the model references though--there's nothing on that object that references the model, is there?

pyomo/contrib/gdpopt/algorithm_base_class.py
```diff
@@ -0,0 +1,470 @@
...
+            var.set_value(soln, skip_validation=True)
+        for var, soln in zip(
+                self.original_util_block.boolean_variable_list,
+                self.incumbent_boolean_soln):
+            # If there is an associated binary, make sure it's in sync too
+            original_binary = var.get_associated_binary()
+            if soln is None:
+                var.set_value(soln, skip_validation=True)
+                if original_binary is not None:
+                    original_binary.set_value(None, skip_validation=True)
```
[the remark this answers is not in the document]
emma58: Nope, that was just me being insane, whoops.

pyomo/contrib/gdpopt/branch_and_bound.py
```diff
@@ -37,442 +46,481 @@
...
-            # We have reached a leaf node, or the best available node is
-            # infeasible.
-            original_model = solve_data.original_model
-            copy_var_list_values(
-                from_list=node_model.GDPopt_utils.variable_list,
-                to_list=original_model.GDPopt_utils.variable_list,
-                config=config,
+            # Make sure disjuncts fixed to False are properly deactivated.
+            for disjunct in disjuncts_fixed_False:
+                disjunct.deactivate()
```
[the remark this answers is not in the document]
emma58: Yeah, it is. We fix Disjuncts to False when we deactivate them, but the converse isn't true. So this model would hit it:
```
m = ConcreteModel()
m.x = Var(bounds=(0, 3))
m.d = Disjunction(expr=[[m.x ** 2 >= 3, m.x >= 3], [m.x ** 2 <= -1, m.x <= -1]])
m.o = Objective(expr=m.x)
m.d.disjuncts[1].indicator_var.fix(False)
```
Unfortunately, this model also doesn't solve, so I'm going to open an issue and defer fixing that to another PR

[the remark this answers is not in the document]
emma58: OK, that's in #2483 now

pyomo/contrib/gdpopt/branch_and_bound.py
```diff
@@ -37,442 +46,481 @@
...
+                               num_unfixed_disjuncts,
+                               node_data.num_unbranched_disjunctions - 1,
+                               len(self.bb_queue)))
+
+    def _prescreen_node(self, node_data, node_model, config):
+        # Check node for satisfiability if sat-solver is enabled
+        if config.check_sat and satisfiable(node_model, config.logger) is False:
+            if node_data.node_count == 0:
+                config.logger.info("Root node is not satisfiable. Problem is "
+                                   "infeasible.")
```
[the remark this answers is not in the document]
emma58: Yep, this one is both reachable and the termination condition is correct. I added a test (well played, John. :P)

pyomo/contrib/gdpopt/config_options.py
```diff
@@ -32,13 +57,10 @@ def _get_GDPopt_config():
         default=600,
```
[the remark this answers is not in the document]
emma58: Agreed. I like `None` better than `0` as the default.

pyomo/contrib/gdpopt/branch_and_bound.py
```diff
@@ -37,442 +46,481 @@
...
+                except InfeasibleConstraintException:
+                    # copy variable values, even if errored
+                    copy_var_list_values(
+                        from_list=subprob_utils.algebraic_variable_list,
+                        to_list=model_utils.algebraic_variable_list,
+                        config=config, ignore_integrality=True
+                    )
+                    return float('inf'), float('inf')
+                minlp_args = dict(config.minlp_solver_args)
+                if config.minlp_solver == 'gams':
```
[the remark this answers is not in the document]
emma58: Yes, the inconsistency of doing this sometimes is really annoying. I like the open-a-ticket plan! #2484

pyomo/contrib/gdpopt/solve_subproblem.py
```diff
@@ -0,0 +1,329 @@
...
+    try:
+        # First do FBBT
+        fbbt(m, integer_tol=config.integer_tolerance,
+             feasibility_tol=config.constraint_tolerance,
+             max_iter=config.max_fbbt_iterations)
+        xfrm = TransformationFactory
+        # Now that we've tightened bounds, see if any variables are fixed
+        # because their lb is equal to the ub (within tolerance)
+        xfrm('contrib.detect_fixed_vars').apply_to(
+            m, tolerance=config.variable_tolerance)
```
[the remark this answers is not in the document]
emma58: Yes, this would be a nice option in FBBT. What do you think @michaelbynum?

pyomo/contrib/gdpopt/solve_subproblem.py
```diff
@@ -0,0 +1,329 @@
...
+                v.setlb(lb)
+                v.setub(ub)
+
+        # Now, if something got fixed to 0, we might have 0*var terms to remove
+        xfrm('contrib.remove_zero_terms').apply_to(
+            m, constraints_modified=constraints_modified)
+        # Last, check if any constraints are now trivial and deactivate them
+        xfrm('contrib.deactivate_trivial_constraints').apply_to(
+            m, tolerance=config.constraint_tolerance,
+            return_trivial=constraints_deactivated)
```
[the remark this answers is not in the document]
emma58: Yes, right now we need this, but I agree that it would be great if the writers and solver interfaces could take care of it. #2485

emma58:
OK, I'm confused what's going on with Jenkins, but this is passing everything else and is ready for you @jsiirola

pyomo/contrib/gdpopt/solve_subproblem.py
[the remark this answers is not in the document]
michaelbynum: Yes. I can certainly add an option to FBBT.

pyomo/contrib/gdpopt/solve_subproblem.py
```diff
@@ -0,0 +1,329 @@
...
+    # that's the case.
+    constraints_modified = {}
+    constraints_deactivated = []
+    not_infeas = True
+
+    try:
+        # First do FBBT
+        fbbt(m, integer_tol=config.integer_tolerance,
+             feasibility_tol=config.constraint_tolerance,
+             max_iter=config.max_fbbt_iterations)
```
michaelbynum: FBBT can be quite slow for some problems (especially if `max_fbbt_iterations` is above 2 or 3). I would try use the implementation in Appsi if it is available. I probably need to add an `available` method first though.

pyomo/contrib/gdpopt/tests/test_gdpopt.py
```diff
@@ -143,103 +163,289 @@ def test_multiple_objectives(self):
...
+        m.GDPopt_utils.variable_list = [m.x, m.y, m.z]
+        m.GDPopt_utils.disjunct_list = [m.d._autodisjuncts[0],
+                                        m.d._autodisjuncts[1]]
+        results = SolverFactory('gdpopt.loa').solve(
+            m, mip_solver=mip_solver, nlp_solver=nlp_solver)
+        self.assertEqual(results.solver.termination_condition,
+                         TerminationCondition.unbounded)
+
+    # [ESJ 5/16/22]: Using Gurobi for this test because glpk seems to get angry
+    # on Windows when the MIP is arbitarily bounded with the large bounds. And I
```
blnicho, line 412: Typo: arbitrarily

blnicho (approved):

jsiirola (approved):

PR #2410, opened 2022-05-18
Replace variables in a LinearExpression with constants

michaelbynum:
## Summary/Motivation:
Currently, 

```
import pyomo.environ as pe
from pyomo.core.expr.numeric_expr import LinearExpression
from pyomo.core.expr.visitor import replace_expressions

m = pe.ConcreteModel()
m.x = pe.Var()
m.y = pe.Var()
e = LinearExpression(linear_coefs=[2.5], linear_vars=[m.x])
e += m.y
sub_map = dict()
sub_map[id(m.x)] = 0.5
e2 = replace_expressions(e, sub_map)
```

results in an error:

```
  File "/.../pyomo/pyomo/core/linear_replacement.py", line 12, in <module>
    e2 = replace_expressions(e, sub_map)
  File "/.../pyomo/pyomo/core/expr/visitor.py", line 625, in replace_expressions
    return ExpressionReplacementVisitor(
  File "/.../pyomo/pyomo/core/expr/visitor.py", line 313, in walk_expression
    node_result = self.exitNode(node, data)
  File "/.../pyomo/pyomo/core/expr/visitor.py", line 703, in exitNode
    return node.create_node_with_local_data(tuple(data[1]))
  File "/.../pyomo/pyomo/core/expr/numeric_expr.py", line 1384, in create_node_with_local_data
    return classtype(args)
  File "/.../pyomo/pyomo/core/expr/numeric_expr.py", line 1323, in __init__
    self._args_ = args
  File "/.../pyomo/pyomo/core/expr/numeric_expr.py", line 1361, in _args_
    self.linear_coefs, self.linear_vars = zip(
ValueError: not enough values to unpack (expected 2, got 0)
```

This PR corrects that by checking to see if there is only one argument and it is a constant.

### Legal Acknowledgement


michaelbynum:
On second thought, it may be better to check for this in `create_node_with_local_data`.

pyomo/core/expr/numeric_expr.py
```diff
@@ -1376,7 +1376,7 @@ def create_node_with_local_data(self, args, classtype=None):
                 classtype = self.__class__
             elif ( args[0].__class__ is MonomialTermExpression or
                    (args[0].__class__ in native_types or args[0].is_constant()
-                ) and all(arg.__class__ is MonomialTermExpression
+                ) and len(args) > 1 and all(arg.__class__ is MonomialTermExpression
                           for arg in args[1:])):
                 classtype = self.__class__
```
jsiirola, line 1381: This proposed change is IMHO to aggressive at turning valid a `LinearExpression` into a general `SumExpression`.  I think a better solution is:
```python
if classtype is None:
    const = 0
    new_args = []
    for arg in args:
            new_args.append(arg)
        elif arg.__class__ in native_types or arg.is_constant():
            const += arg
        else:
            return SumExpression(args)
    return self.__class__([const] + new_args)
```

michaelbynum: Does it make sense to have a `LinearExpression` with no variables?

jsiirola: Probably not.  So, after the loop might look like:
```python
if not new_args:
    return const
if const:
    new_args.insert(0, const)
return self.__class__(new_args)
```

jsiirola: @michaelbynum Following up on today's discussion:  what do you think about the following implementation:
```python
def create_node_with_local_data(self, args, classtype=None):
    if classtype is not None:
        return classtype(args)
    else:
        const = 0
        new_args = []
        for arg in args:
                new_args.append(arg)
            elif arg.__class__ in native_types or arg.is_constant():
                const += arg
            else:
                return SumExpression(args)
        if not new_args:
            return const
        if const:
            new_args.insert(0, const)
        return self.__class__(new_args)
```

michaelbynum: @jsiirola, I like this proposal a lot. I'm on board.

michaelbynum:
@jsiirola I finally took care of this one. Assuming tests pass, it is ready for you to look at.

michaelbynum:
Thanks, @jsiirola!

PR #2421, opened 2022-05-24
Support cython functions through `Initializer()`

### LINKED ISSUE #2374: Support cython function to initialize set
## Summary

Currently it's impossible to pass a "cython" function to initialize a Set. 

The point is that, in the "Initializer" function of pyomo.core.base.util, the cython function is not correctly recognized and a "ConstantInitializer" is returned. Ideally a "ScalarCallInitializer" or "IndexedCallInitializer" should be returned. 

It would be nice to support passing "cython" function.

### Rationale
The main reason to support cython function is that there is no good reason to not accept them. They are behaving mostly like normal python function.

 In particular I think that the only modification needed will be in the Initializer function of pyomo.core.base.util.

Pyomo is currently relying on "inspect.isfunction" which does not recognized cython function. Note that other methods of "inspect" such as "inspect.getfullargspec(init)" and "inspect.ismethod(init)" are working correctly on cython function.

### Description

There are many discussions around the question whether "inspect.isfunction" should return true or false for cython function. It seems that it will not change soon. The reason is that cython function are not supporting ALL the features expected from types.FunctionType. But these features are not needed by pyomo. 

As workaround in my code I'm currently using the following patching:

```
from unittest.mock import patch

def extended_isfunction(f):
    return isinstance(f, types.FunctionType) or (type(f).__name__ == 'cython_function_or_method')

with patch('inspect.isfunction', new=extended_isfunction):
    my_pyomo_code
```
It would be nicer if this logic could be implemented in pyomo directly.

jsiirola:
## Summary/Motivation:
This resolves issues where `Initializer` did not correctly identify all functions (notably cythonized functions).  This updates the logic for processing initialization functions and expands the `Initializer()` unit tests.

- Correctly detect cythonized functions as callable functions
- update `Initializer` unit tests
- Fix issue in `pyomo.core.__init__.py` that was preventing the import of `pyomo.core.util`

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/initializer.py
+++ b/pyomo/core/base/initializer.py
@@ -66,8 +66,8 @@ def Initializer(init,
         # indexed components).  We will preserve that functionality
         # here.
         #
-        # I was concerned that some builting aren't compatible with
-        # getfullargspec (and woulf need the same try-except logic as in
+        # I was concerned that some builtins aren't compatible with
+        # getfullargspec (and would need the same try-except logic as in
         # the partial handling), but I have been unable to come up with
         # an example.  The closest was getattr(), but that falls back on
         # getattr.__call__, which does support getfullargspec.
--- a/pyomo/core/base/initializer.py
+++ b/pyomo/core/base/initializer.py
@@ -22,6 +22,12 @@ from pyomo.core.pyomoobject import PyomoObject
 
 initializer_map = {}
 sequence_types = set()
+# initialize with function, method, and method-wrapper types.
+function_types = set([
+    type(PyomoObject.is_expression_type),
+    type(PyomoObject().is_expression_type),
+    type(PyomoObject.is_expression_type.__call__),
+])
 
 #
 # The following set of "Initializer" classes are a general functionality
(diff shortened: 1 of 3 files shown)


pyomo/core/base/initializer.py
```diff
@@ -49,13 +55,22 @@ def Initializer(init,
...
+        # because some function-like things (notably cythonized
+        # functions) return False
         if not allow_generators and inspect.isgeneratorfunction(init):
             raise ValueError("Generator functions are not allowed")
         # Historically pyomo.core.base.misc.apply_indexed_rule
         # accepted rules that took only the parent block (even for
         # indexed components).  We will preserve that functionality
         # here.
+        #
+        # I was concerned that some builting aren't compatible with
```
mrmundt, line 69: Typo: `builting`

pyomo/core/base/initializer.py
```diff
@@ -49,13 +55,22 @@ def Initializer(init,
...
+        # getfullargspec (and woulf need the same try-except logic as in
```
mrmundt, line 70: Typo: `would`

pyomo/core/base/initializer.py
```diff
@@ -111,7 +126,13 @@ def Initializer(init,
         # generator into a tuple and then store it as a constant.
         return ConstantInitializer(tuple(init))
     if type(init) is functools.partial:
-        _args = inspect.getfullargspec(init.func)
+        try:
+            _args = inspect.getfullargspec(init.func)
+        except:
+            # Inspect doesn't work for some built-in callables (notably
```
mrmundt, line 132: Does one of the new tests check against this assumption?

pyomo/core/tests/unit/test_initializer.py
```diff
@@ -462,6 +618,52 @@ def test_dataframe(self):
...
+        self.assertIs(type(a), ItemInitializer)
+        self.assertFalse(a.constant())
+        self.assertFalse(a.verified)
+        self.assertTrue(a.contains_indices())
+        self.assertEqual(list(a.indices()), [0,1,2])
+        self.assertEqual(a(None, 0), 1)
+        self.assertEqual(a(None, 1), 2)
+        self.assertEqual(a(None, 2), 4)
+
+        # TODO: How should we handle ndarray matricies?
```
mrmundt, line 647: Do you actually want to leave this part it? (If so, typo: `matrices`)

mrmundt (changes requested):

pyomo/core/base/initializer.py
jsiirola: Yes: test_initializer.py:531 (`test_no_argspec`)

mrmundt (approved):

PR #2425, opened 2022-05-31
Fixes to PyROS Ellipsoidal Sets

shermanjasonaf:
## Summary/Motivation:
Fixes to the PyROS `EllipsoidalSet` and `AxisAlignedEllipsoidalSet` objects.

1) Fix the `AxisAlignedEllipsoidalSet` bounds calculation. Given a central point $q^0$, a nonnegative scalar $s$, and a positive-definite square matrix $P$, the bounds for the Cartesian coordinates in the ellipsoid defined by $$\mathcal{Q}(q^0, P, s) = \\{ q \in \mathbb{R}^n \mid q = q^0 + P^{1/2} \xi, \xi \in \mathbb{R}^n, \xi^\intercal \xi \leq s \\}$$ are $$q_i^0 \pm (sP_{ii})^{1/2}, \qquad i = 1, 2, \dots, n;$$ but currently, the bounds provided through the `EllipsoidalSet.parameter_bounds` method are $$q_i^0 \pm s(P_{ii})^{1/2}, \qquad i = 1, 2, \dots, n;$$ Consequently, the bounds provided by PyROS are overly restrictive when $0 < s < 1$, and lax for $s > 1$. Since these bounds are imposed on the uncertain parameter variables in PyROS separation problems (through `pyros.separation_problem_methods.add_uncertainty_set_constraints`), PyROS only solves a restriction of the actual separation problem(s) for the case $0 < s < 1$.

2) Add support for `AxisAlignedEllipsoidalSet` instances with half-lengths of zero. This makes the implementation of the `AxisAlignedEllipsoidalSet` class consistent with the documentation, which provides a generic definition to account for half-lengths of zero.

### Legal Acknowledgement


pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -767,31 +767,58 @@ def parameter_bounds(self):
 
     def set_as_constraint(self, uncertain_params, **kwargs):
```
blnicho, line 768: Can you remove `**kwargs` here? Doesn't look like it is being used at all in this method.

blnicho (approved):

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -765,33 +765,60 @@ def parameter_bounds(self):
...
+        # expand all uncertain parameters to a list.
+        # this accounts for the cases in which `uncertain_params`
+        # consists of indexed model components,
+        # or is itself a single indexed component
+        if isinstance(uncertain_params, list):
+            for param in uncertain_params:
+                if param.is_indexed():
+                    all_params.extend(list(param.values()))
+                else:
+                    all_params.append(param)
```
jsiirola, line 789: Do you need the `is_indexed()` test here?  I think `.values()` should work for both scalar and indexed.  In addition, you don't need the `list(...)`, as `extend()` accepts iterables / generators.  I think the following should work:
```python
if not isinstance(uncertain_params, (tuple, list)):
    uncertain_params = [uncertain_params]
all_params = []
for uparam in uncertain_params:
    all_params.extend(uparam.values())
```

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -765,33 +765,60 @@ def parameter_bounds(self):
...
+
+        for idx, (param, ctr, half_len) in enumerate(zip_all):
+            if half_len > 0:
+                diffs_squared.append((param - ctr) ** 2 / (half_len) ** 2)
+            else:
+                certain_params.append((param, ctr))
 
-        # Calculate inner product of difference vector and variance matrix
-        constraint = sum([x * y for x, y in zip(inverse_squared_half_lengths, diff_squared)])
+        con_lhs = sum(diff for diff in diffs_squared)
```
jsiirola, line 811: this can just be `con_lhs = sum(diffs_squared)`

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -765,33 +765,60 @@ def parameter_bounds(self):
...
+                f"Center of ellipsoid is of dimension {len(self.center)},"
+                f" but vector of uncertain parameters is of dimension"
+                f" {len(all_params)}"
+            )
+
+        zip_all = zip(all_params, self.center, self.half_lengths)
+        diffs_squared = list()
+        certain_params = list()
+
+        for idx, (param, ctr, half_len) in enumerate(zip_all):
```
jsiirola, line 805: General question: why create the `certain_params` list?  Wouldn't the following be simpler to just:
```python
for idx, (param, ctr, half_len) in enumerate(zip_all):
    if half_len > 0:
        diffs_squared.append((param - ctr) ** 2 / (half_len) ** 2)
    else:
        # If the half length is zero, fix the uncertain param
        conlist.add(param == ctr)
```

jsiirola (commented):
Some minor comments / changes, then I think this is good.

pyomo/contrib/pyros/uncertainty_sets.py
shermanjasonaf: See subsequent commits.

pyomo/contrib/pyros/uncertainty_sets.py
shermanjasonaf: See subsequent commits.

pyomo/contrib/pyros/uncertainty_sets.py
shermanjasonaf: See subsequent commits.

pyomo/contrib/pyros/uncertainty_sets.py
shermanjasonaf: @blnicho I've noticed that removing the `**kwargs` may result in a `TypeError` during the course of a PyROS solve attempt in which the uncertainty set is an `AxisAlignedEllipsoidalSet`. In particular, this issue arises from [lines 107--109 of uncertainty_sets.py](https://github.com/shermanjasonaf/pyomo/blob/280e4c76e49e02f613444a8d33c8f381e90c825b/pyomo/contrib/pyros/uncertainty_sets.py#L107), and [lines 30--32 of separation_problem_methods.py](https://github.com/shermanjasonaf/pyomo/blob/280e4c76e49e02f613444a8d33c8f381e90c825b/pyomo/contrib/pyros/separation_problem_methods.py#L30), in which the `set_as_constraint` method is invoked generically for any uncertainty set, without accounting for the differences between the arguments accepted by the `set_as_constraint` methods for different `UncertaintySet` subclasses. Currently, the presence of the `**kwargs` argument(s) seems to "guard" against this issue. For now, I've restored the `**kwargs` argument and added a test case in a subsequent commit.

jsiirola: I think the correct solution is to update the signature here to be:
```
def set_as_constraint(self, uncertain_params, model, config):
```
to match the other APIs.

shermanjasonaf: See subsequent commits

jsiirola (approved):

PR #2432, opened 2022-06-14
Mindtpy bug fix

ZedongPeng:
## Summary/Motivation:
In MindtPy, after the OA algorithm converges, the optimal solution will be copied to the original model using `copy_var_list_values` function as below. However, `MindtPy.variable_list` does not include the unused variables, while `solve_data.original_model.component_data_objects(Var)` does. Now, we use `component_data_objects(Var)` for both working_model and original_model. 
The length check of `from_list` and `to_list` is also added in `copy_var_list_values` function.

https://github.com/Pyomo/pyomo/blob/aaed8923564d0198725c3e02fd56ccdcf4fd955a/pyomo/contrib/mindtpy/MindtPy.py#L242-L246

### Legal Acknowledgement


ZedongPeng:
Hi @emma58 . I added a length check in the `copy_var_list_values` function. There are several tests in GDPOpt failing because the lengths of `from_list` and `to_list` do not match. Could you look into it?

bernalde (approved):

pyomo/contrib/mindtpy/MindtPy.py
```diff
@@ -239,10 +242,15 @@ def solve(self, model, **kwds):
...
+                solve_data.working_model.MindtPy_utils.deactivate()
+                if solve_data.working_model.find_component("_int_to_binary_reform") is not None:
+                    solve_data.working_model._int_to_binary_reform.deactivate()
                 copy_var_list_values(
-                    MindtPy.variable_list,
+                    [i for i in solve_data.working_model.component_data_objects(
+                        Var, active=True, descend_into=True) if not i.fixed],
                     [i for i in solve_data.original_model.component_data_objects(
-                        Var) if not i.fixed],
+                        Var, active=True, descend_into=True) if not i.fixed],
```
emma58, line 253: This won't work if a Var that's used in an active Constraint or Objective is on a deactivated Block, which will be a problem as #1032 permeates. I think it would be better to use the `get_vars_from_components` utility to collects the Vars. (But if I'm missing something and you really do need the unused variables, then you will have to go find them with `component_data_objects`. Though it still seems better to find the used ones not on the active sub-tree as well.)

pyomo/contrib/mindtpy/MindtPy.py
```diff
@@ -239,10 +242,15 @@ def solve(self, model, **kwds):
                     from_list=solve_data.best_solution_found.MindtPy_utils.variable_list,
                     to_list=MindtPy.variable_list,
                     config=config)
+                # MindtPy.variable_list does not include the unused variables, but component_data_objects(Var) does. component_data_objects(Var) should be used for both working_model and original_model.
```
emma58, line 245: Why do you need the unused variables? If they're unused in the original model, you're never going to need to do anything with them, right?

emma58 (commented):
@ZedongPeng, I'm a little confused why you need to find the unused variables at all? Maybe this is a little philosophical, but ignoring them makes sense to me: They're not really part of the model if they're not in any constraints or objectives. In addition, I don't think using `component_data_objects`  to find the Vars is safe, as it won't find used variables that are hidden on deactivated blocks.

bernalde:
For context, this PR comes from addressing a comment someone made to @ZedongPeng and me privately over email which I copy here:

> 
> Dear Bernal and Peng,
> 
>  
> 
> I was testing pyomo using mindtpy and it was working perfectly in a simple NPL problem. Then I tried to add a binary variable to check if it worked as expected and I got really odd results (the expected results were the same as in the NLP). Can you please take a look and give me your best opinion?
> 
> Here is the code:
> ```
> 
>  
> 
> from pyomo.environ import *
> 
> Cmn=18.06 #Weymouth constant
> Nodes=[1,2,3]
> 
> m = ConcreteModel()
> 
> m.flow_gas = Var(Nodes,Nodes, bounds=(-1000,1000))
> m.gas_output = Var(Nodes, bounds=(-20,0), initialize=0)
> m.gas_input = Var(Nodes, initialize=1)
> m.gas_pressure = Var(Nodes, bounds=(100,110))
> m.binary_test = Var(within=Binary,initialize=0)
> 
> m.c1 = Constraint(expr = m.flow_gas[1,2]**2 == (Cmn**2)*abs((m.gas_pressure[1]**2)-(m.gas_pressure[2]**2)))
> m.c2 = Constraint(expr = m.flow_gas[2,3]**2 == (Cmn**2)*abs((m.gas_pressure[2]**2)-(m.gas_pressure[3]**2)))
> 
> m.c7 = Constraint(expr = m.gas_input[1] == m.flow_gas[1,2])
> m.c8 = Constraint(expr = m.gas_input[2] + m.flow_gas[1,2] == m.flow_gas[2,3])
> m.c39 = Constraint(expr = -m.gas_output[3] == m.flow_gas[2,3])
> 
> #Gas input
> m.c15 = Constraint(expr = m.gas_input[1] >= 0)
> m.c16= Constraint(expr = m.gas_input[1] <= 10)
> 
> m.c17 = Constraint(expr = m.gas_input[2] >= 0)
> m.c18= Constraint(expr = m.gas_input[2] <= 5)
> 
> m.c19 = Constraint(expr = m.gas_input[3] == 0)
> #
> # #Gas output
> m.c21 = Constraint(expr = m.gas_output[1] == 0)
> 
> m.c23 = Constraint(expr = m.gas_output[2] == 0)
> #
> m.c25 = Constraint(expr = m.gas_output[3] <= 0)
> m.c26 = Constraint(expr = m.gas_output[3] >= -15*(m.binary_test))
> 
> m.objective = Objective(expr = m.gas_input[1]*10+m.gas_input[2]*10+m.gas_output[3]*11, sense = minimize)
> 
> results=SolverFactory('mindtpy').solve(m, mip_solver='gurobi', nlp_solver='ipopt', tee=True)
> 
> 
> m.objective.display()
> 
> 
> print(m.gas_input[1].value)
> print(m.gas_input[2].value)
> print(m.gas_input[3].value)
> 
> print(m.gas_output[1].value)
> print(m.gas_output[2].value)
> print(m.gas_output[3].value)
> print(m.binary_test.value)
> ```

pyomo/contrib/mindtpy/MindtPy.py
ZedongPeng: Yep. Loading the optimal solution has nothing to do with the unused variables. In the working_model, the unused variables have already been filtered out in the variable_list. However, the original_model has not. I am not sure if we need to take efforts to check the unused variables in the original model here. Or just load the optimal solution to both used and unused variables.

pyomo/contrib/mindtpy/MindtPy.py
ZedongPeng: I made a simple test and as you said, `component_data_objects` doesn't work if a Var that's used in an active Constraint or Objective is on a deactivated Block. I'll use `get_vars_from_components` here. Thanks for your advice.

pyomo/contrib/mindtpy/MindtPy.py
emma58: OK, this makes sense. Then yes: I think you should never use `component_data_objects` to collect the Vars. Then you'll never see unused variables, which is fine because it will make mindtpy match the behavior of other solvers. That is, unused variables are never sent to the solver, and no solution is filled in for them after the solve. (This means unused variables will be stale after the solve, which they should be.)

emma58:
@ZedongPeng, perhaps it would be worth adding the model @bernalde posted as a test in mindtpy, as well. That way we never reintroduce this bug! :)

ZedongPeng:
> @ZedongPeng, perhaps it would be worth adding the model @bernalde posted as a test in mindtpy, as well. That way we never reintroduce this bug! :)

I added the optimal solution check for all the tests in MindtPy. With the length check in `copy_var_list_values` function, I think this will be enough.

pyomo/contrib/mindtpy/MindtPy.py
```diff
@@ -239,11 +242,25 @@ def solve(self, model, **kwds):
...
-                    config)
+                # The original does not have variable list. Use get_vars_from_components() should be used for both working_model and original_model to exclude the unused variables.
+                solve_data.working_model.MindtPy_utils.deactivate()
+                if solve_data.working_model.find_component("_int_to_binary_reform") is not None:
+                    solve_data.working_model._int_to_binary_reform.deactivate()
+                copy_var_list_values(list(get_vars_from_components(block=solve_data.working_model, 
+                                         ctype=(Constraint, Objective), 
+                                         include_fixed=False, 
+                                         active=True,
+                                         sort=True,
```
ZedongPeng, line 253: Here, I need to set `sort=Ture`. Otherwise, the order of variables in `list(get_vars_from_components(block=solve_data.working_model, ...)` and `list(get_vars_from_components(block=solve_data.original_model, ...)` will be different.
@emma58

emma58:
@ZedongPeng, sorry for the delay. I had a chance to look into the failing GDPopt tests. They're a bit hard to fix in the current version of GDPopt, so why don't you just skip those tests (I believe it's only two) in this PR, and I will fix them (and un-skip them) in the GDPopt rewrite shortly.

pyomo/contrib/mindtpy/MindtPy.py
```diff
@@ -60,10 +60,13 @@
...
 from pyomo.contrib.mindtpy.util import model_is_valid, set_up_solve_data, set_up_logger, get_primal_integral, get_dual_integral
 from pyomo.core import (Block, ConstraintList, NonNegativeReals,
-                        Var, VarList, TransformationFactory, RangeSet, minimize)
+                        Var, VarList, TransformationFactory, RangeSet, minimize, Constraint, Objective)
 from pyomo.opt import SolverFactory
 from pyomo.contrib.mindtpy.config_options import _get_MindtPy_config, check_config
 from pyomo.common.config import add_docstring_list
+from pyomo.util.vars_from_expressions import get_vars_from_components
+from pyomo.gdp import Disjunct
+from pyomo.common.collections import ComponentSet
```
emma58, line 69: Do you use these two imports?

emma58 (approved):
This looks good pending tests passing (though I think you might have a couple unused imports?)

pyomo/contrib/gdpopt/tests/test_gdpopt.py
```diff
@@ -590,34 +590,34 @@ def test_infeasible_GDP(self):
...
-            [m.x + m.y >= 5], [m.x - m.y <= 3]
-        ])
-        m.o = Objective(expr=-m.x ** 2, sense=maximize)
-        SolverFactory('gdpopt').solve(
-            m, strategy='RIC',
-            mip_solver=mip_solver,
-            nlp_solver=nlp_solver
-        )
-        self.assertAlmostEqual(value(m.o), 0)
+    # def test_GDP_nonlinear_objective(self):
```
bernalde, line 593: Please add a comment here referring to the PR and that it will get fixed in a future PR to not loose track of this

pyomo/contrib/gdpopt/util.py
```diff
@@ -253,6 +253,8 @@ def copy_var_list_values(from_list, to_list, config,
     Rounds to Binary/Integer if necessary
     Sets to zero for NonNegativeReals if necessary
     """
+    if len(from_list) != len(to_list):
+        raise ValueError('The lengths of from_list and to_list do not match.')
```
bernalde, line 257: Can we come up with a simple test for this line? Codecov is crying about it and it refers to the commented tests above.

bernalde (approved):
This PR is ready to merge. If possible we can include a comment to refer to some future PR which enables some commented out tests regarding the length check for GDPOpt. Besides, if we can increase the test coverage by a small % it should be ok with codecov, any comment in this regard is well appreciated.

pyomo/contrib/gdpopt/tests/test_gdpopt.py
emma58: It is actually already fixed in #2406

PR #2445, opened 2022-06-23
Add Tolerances to FBBT Tests

michaelbynum:
- Add tolerances to FBBT tests
-

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/fbbt/tests/test_fbbt.py
+++ b/pyomo/contrib/fbbt/tests/test_fbbt.py
@@ -469,8 +469,8 @@ class FbbtTestBase(object):
                 else:
                     xu = m.x.ub
                 _x = np.exp(np.log(y) / _exp_val)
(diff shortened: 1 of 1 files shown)


blnicho (approved):

jsiirola:
This seems reasonable, but is there any indication as to why tolerances are necessary now?

michaelbynum:
@jsiirola It is not clear to me. The tolerance comes into play when the upper bound on `y` is `2.8` and the exponent, `_exp_val` is `0.5`. The computed upper bound on `x` is `7.839999999999998` both on my machine and on GHA (the exact value should be `7.84`). However, the test passes on my machine without the tolerance. The value it gets compared against on GHA is `7.839999999999999`. I guess I am surprised this test passed so consistently before.

jsiirola:
If this is a machine precision issue, should we make the tolerance 1e-14 instead of 1e-12?

PR #2456, opened 2022-07-07
Do not return duplicate data from `component_data_objects` or `component_data_iterindex`

jsiirola:
## Summary/Motivation:
This is the first step toward resolving issues with references where the model component iterators can return the same data multiple times.  The implementation is aware that the only way to see the same component twice is through `Reference`s, and makes use of this to minimize the overhead of deduplication (in particular, `id()`s are only cached for component containers, _unless_ a `Reference` is encountered).

This implementation has only minimal impact on performance (<1% overhead):
image

Note that this does NOT resolve the issues in the writers, etc.  That will be implemented in a subsequent PR

- `component_data_objects()` and `component_data_iterindex()` will not yield the same data object more than once
- clean up GDP handling of deactivated blocks / disjuncts 
- updated documentation

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/block.py
+++ b/pyomo/core/base/block.py
@@ -1337,9 +1337,14 @@ Components must now specify their rules explicitly using 'rule=' keywords.""" %
             return PseudoMap(self, ctype, active, sort)
 
     def _component_data_iteritems(self, ctype=None, active=None, sort=False):
-        """
-        Generator that returns a 3-tuple of (component name, index value,
-        and _ComponentData) for every component data in the block.
+        """return the name, index, and component data for matching ctypes
+
+        Generator that returns a nested 2-tuple of
+
+            ((component name, index value), _ComponentData)
+
+        for every component data in the block matching the specified
+        ctype(s).
         """
         _sort_indices = SortComponents.sort_indices(sort)
         _subcomp = PseudoMap(self, ctype, active, sort)
--- a/pyomo/core/base/block.py
+++ b/pyomo/core/base/block.py
@@ -148,6 +148,23 @@ class _DeduplicateInfo(object):
         self.seen_data = set()
 
     def unique(self, comp, items, are_values):
+        """Generator that filters duplicate _ComponentData objects from items
+
+        Parameters
+        ----------
+        comp: ComponentBase
+           The Component (indexed or scalar) that contains all
+           _ComponentData returned by the `items` generator.
+
+        items: generator
+            Generator yielding either the values or the items from the
+            `comp` Component.
+
+        are_values: bool
+            If `True`, `items` yields _ComponentData objects, otherwise,
+            `items` yields `(index, _ComponentData)` tuples.
+
+        """
         if comp.is_reference():
             seen_components_contains = self.seen_components.__contains__
             seen_comp_thru_reference_contains \
@@ -160,7 +177,7 @@ class _DeduplicateInfo(object):
                 _data = _item if are_values else _item[1]
                 # If the data is contained in a component we have
                 # already processed, then it is a duplicate and we can
-                # bypass forther checks.
+                # bypass further checks.
                 _id = id(_data.parent_component())
                 if seen_components_contains(_id):
                     continue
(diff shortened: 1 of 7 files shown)


pyomo/core/base/block.py
```diff
@@ -1334,10 +1414,15 @@ def _component_typemap(self, ctype=None, active=None, sort=False):
         else:
             return PseudoMap(self, ctype, active, sort)
 
-    def _component_data_iter(self, ctype=None, active=None, sort=False):
-        """
-        Generator that returns a 3-tuple of (component name, index value,
-        and _ComponentData) for every component data in the block.
+    def _component_data_iteritems(self, ctype, active, sort, dedup):
```
michaelbynum, line 1417: Can you please add the expected types of the arguments (either as type hints or in the doc string)? I believe deduce is expected to be an instance of `_DeduplicateInfo`, but future me (if I ever look at this code again) will not remember that.

michaelbynum (commented):
I am super excited about this PR. I also really like this implementation. It should add very little overhead. I do have a few minor comments/questions.

michaelbynum:
P.s. Thanks for checking the performance.

pyomo/core/base/block.py
```diff
@@ -120,6 +120,84 @@ def __len__(self):
...
+                = self.seen_comp_thru_reference.__contains__
+            seen_comp_thru_reference_add = self.seen_comp_thru_reference.add
+            seen_data_contains = self.seen_data.__contains__
+            seen_data_add = self.seen_data.add
+
+            for _item in items:
+                _data = _item if are_values else _item[1]
+                # If the data is contained in a component we have
+                # already processed, then it is a duplicate and we can
+                # bypass forther checks.
```
blnicho, line 163: Typo: further

pyomo/core/base/block.py
```diff
@@ -1482,72 +1592,56 @@ def block_data_objects(self,
                            sort=False,
                            descend_into=True,
                            descent_order=None):
+        """Generator returning this block and an any matching sub-blocks.
```
blnicho, line 1595: Typo: "and an any"

pyomo/core/base/block.py
```diff
@@ -1482,72 +1592,56 @@ def block_data_objects(self,
...
+        -----
+        The `self` block is *always* returned, regardless of the types
+        indicated by descend_into.
 
-        # TODO: merge into block_data_objects
-        if ctype is None:
-            ctype = (Block,)
-        elif isclass(ctype):
-            ctype = (ctype,)
+        The active flag is enforced on *all* blocks, inclufing self.
```
blnicho, line 1609: Typo: including

emma58 (commented):
This is really cool! I don't have any questions Michael didn't already ask...

pyomo/core/base/block.py
```diff
@@ -1482,72 +1592,56 @@ def block_data_objects(self,
...
 
-        """
-        This method returns a generator that iterates
-        through the current block and recursively all
-        sub-blocks.  This is semantically equivalent to
+        This is roughly equivalent to
 
-            component_data_objects(Block, ...)
+            iter(block for block in itertools.chain(
+                [self], self.component_data_objects(descend_into, ...))
```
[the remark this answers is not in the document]
jsiirola: Not really - `descend_into` provides the `ctype`; if it is `True`, it is mapped to `(Block,)`.  This is also why I introduced it as "roughly equivalent to".  To make it exact for all cases, I would end up replicating the entire body of the function.  I

pyomo/core/base/indexed_component.py
```diff
@@ -403,8 +403,7 @@ def keys(self, ordered=False):
 
         """
         sort_needed = ordered
-        if hasattr(self._index_set, 'isfinite') and not \
-           self._index_set.isfinite():
+        if not self._index_set.isfinite():
```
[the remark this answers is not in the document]
jsiirola: I think this was legacy code from before the Set rewrite.  I could not see any path / test / case where the `hasattr` test was needed.

blnicho (approved):

michaelbynum (approved):

PR #2458, opened 2022-07-11
Fix exception handling logic bug in `quicksum`

### LINKED ISSUE #2457: 'quicksum' hides exceptions raised in generator from the user, leading to a wrong model
## Summary
If the generator passed to `quicksum` raises an exception at the first call to __next__, the exception handling logic of `quicksum` will hide the exception from the user and return an expression as `0`.  This behavior of `quicksum` may produce a wrong model unintentionally.

The example below shows that the user misspells `param.aaa` as `param.bbb` in the expression,  in fact it is a non-exsiting attribute of the class. However, the code can run fluently without any exception occurs.  

```python
import pyomo.environ as pyo

class Param:
    def __init__(self) -> None:
        self.aaa = [1.1, 2.2, 3.3]

param = Param()

m = pyo.ConcreteModel()
m.i = pyo.RangeSet(0, 3 - 1)
m.x = pyo.Var(m.i)

m.expr = pyo.Expression(expr=pyo.quicksum(m.x[i] * param.aaa[i] for i in m.i))
m.expr.pprint()

m.expr2 = pyo.Expression(expr=pyo.quicksum(m.x[i] * param.bbb[i] for i in m.i))
m.expr2.pprint()
```
### Error Message
```console
expr : Size=1, Index=None
    Key  : Expression
    None : 1.1*x[0] + 2.2*x[1] + 3.3*x[2]
expr2 : Size=1, Index=None
    Key  : Expression
    None :        0.0
```

Pyomo version: 6.4.1
Python version: 3.9.12
Operating system: Windows 11
How Pyomo was installed (PyPI, conda, source): conda

shishouyuan:
## Summary/Motivation:

If the generator passed to `quicksum` raises an exception at the first call to next, the exception handling logic of quicksum will hide the exception from the user and return an expression as 0. This behavior of quicksum may produce a wrong model unintentionally.  See example in #2457 

This PR solves this problem.

- judge iterable by its interface instead of by `try...except...`
- removed `try...except...` when getting the first element of the iterable to prevent hiding exception from the user.

### Legal Acknowledgement


pyomo/core/util.py
```diff
@@ -76,14 +76,18 @@ def quicksum(args, start=0, linear=None):
...
+                # `IndexedComponent_slice` from `_IndexedComponent_slice_iter'
+                if hasattr(args, '__iter__'):
                     args = args.__iter__()
-                    first = next(args, None)
-                except:
+                else:
                     raise RuntimeError("The argument to quicksum() is not iterable!")
+            # No `try...except...`
+            # to prevent hiding exceptions raised in the iterator from the user
+            first = next(args, None)
```
jsiirola, line 89: Wouldn't it be simpler to just do the following:
```python
# Ensure that args is an iterator (this manages things like IndexedComponent_slice objects)
args = iter(args)
first = next(args, None)
```

jsiirola (changes requested):
Thanks for catching this!  I completely agree with removing the try-except.  That said, I think there is a simpler (and more pythonic) approach to consider.

Also, please add a test to the quicksum test suite (`pyomo/core/tests/unit/test_numeric_expr.py` is as good a place as any) that exercises the behavior you are correcting.

pyomo/core/util.py
shishouyuan: Thank you for the suggestion!

- I have put this line of code to the beginning of the `quicksum` function since `args` is required to be iterable for all conditions.
- 2 test for `quicksum` was added to `TestSumExpression` in `test_numeric_expr.py`.

pyomo/core/tests/unit/test_numeric_expr.py
```diff
@@ -3677,6 +3677,45 @@ def test_summation7(self):
...
         self.assertEqual( e(), 15 )
         self.assertIs(type(e), SumExpression)
+        
+    def test_quicksum_reject_noniterable(self):
+        msg = 'Failed to recognize non-iterable args.'
+        try:
+            e = quicksum(1)
+            self.fail(msg)
+        except Exception as ex:
+            self.assertIsInstance(ex, TypeError, msg)
```
jsiirola, line 3687: We should use the standard assertion here:
```python
with self.assertRaisesRegex(TypeError, "'int' object is not iterable"):
    quicksum(1)
```

If you also adopt the suggestion below of logging an error message, we could test that as well:
```python
with LoggingIntercept() as LOG:
    with self.assertRaisesRegex(TypeError, "'int' object is not iterable"):
        quicksum(1)
self.assertEqual(LOG.getvalue(), 'The argument `args` to quicksum() is not iterable!\n')
```

pyomo/core/util.py
```diff
@@ -66,6 +66,13 @@ def quicksum(args, start=0, linear=None):
     Returns:
         The value of the sum, which may be a Pyomo expression object.
     """
+
+    # Ensure that args is an iterator (this manages things like IndexedComponent_slice objects)
+    try:
+        args = iter(args)
+    except:
+        raise TypeError('The argument `args` to quicksum() is not iterable!')
```
jsiirola, line 74: Instead of swallowing the `TypeError` and raising another, I think I would prefer that we log an error and re-raise (this makes the stack trace that the user sees less confusing by avoiding the "During handling of the above exception another exception occurred" message):
```python
try:
    args = iter(args)
except:
    logger.error('The argument `args` to quicksum() is not iterable!')
    raise
```
(and at the top of the module)
```python
import logging
```

jsiirola (commented):
This is great!  A couple more suggestions for cleaning up the code (try not to unnecessarily replace exceptions, and leverage standard unittest assertions), and I think this is good to go!

pyomo/core/tests/unit/test_numeric_expr.py
shishouyuan: Thank you. All changes have been done.

jsiirola (approved):

jsiirola:
@shishouyuan, Thank you for both your work in identifying this bug and your patience with iterating over the PR.  This is great work and much appreciated.

shishouyuan:
> @shishouyuan, Thank you for both your work in identifying this bug and your patience with iterating over the PR. This is great work and much appreciated.

@jsiirola Thank you for your good advice.

Issue #1950, opened 2021-04-26
Fix non-doctest documentation tests


blnicho:
## Summary

Several of our online documentation pages include code snippets that are "literal included" from `doc/OnlineDocs/tests`. This directory has a test driver script, `test_examples.py`, which is meant to test these code snippets that aren't being tested using the standard Sphinx doctest infrastructure. However, it doesn't look like we're running these tests in any of our automated testing infrastructure. Furthermore, the `test_examples.py` script is a modified copy of the book test driver which was rewritten in #1914. 

We need to:

- [ ] Update `test_examples.py` following the changes in `examples/pyomobook/test_book_examples.py`
- [ ] Get the tests passing again
- [ ] Make sure the tests are being run on GitHub Actions

### Information on your system

Pyomo version: main branch
Python version: all
Operating system: all
How Pyomo was installed (PyPI, conda, source): source
Solver (if applicable):

PR #2502, opened 2022-08-21
Fixed examples in working_models.html to Run

MrGorbunov:
## Fixes # .
Code examples in docs now run verbatim

## Summary/Motivation:
Two examples from the docs did not run when copy pasted

In working_models.rst:
 - 'summation' was used instead of 'pyo.summation'
 - An example script instantiated an optimizer twice. First with glpk and then with cplex. Removed the cplex.

### Legal Acknowledgement


blnicho (approved):

jsiirola (approved):
This is good, but why didn't the doc tests catch the error?

blnicho:
> This is good, but why didn't the doc tests catch the error?

I don't think we're running the `test_examples.py` script found here: https://github.com/Pyomo/pyomo/blob/main/doc/OnlineDocs/tests/test_examples.py. See issue https://github.com/Pyomo/pyomo/issues/1950

PR #2504, opened 2022-08-22
Support cloning individual indexed blocks

### LINKED ISSUE #2008: Cannot clone individual blocks within an indexed block
## Summary

`_BlockData` does not have a `_parent` attribute, so it cannot be cloned.

```

In [3]: m.b = pe.Block([1,2,3])

In [4]: m.b[1].clone()
AttributeError                            Traceback (most recent call last)
<ipython-input-4-bb20e18a9495> in <module>
----> 1 m.b[1].clone()

~/repos/pyomo/pyomo/core/base/block.py in clone(self)
   1232         # should be preserved as singletons.
   1233         #
-> 1234         save_parent, self._parent = self._parent, None
   1235         try:
   1236             new_block = copy.deepcopy(

~/repos/pyomo/pyomo/core/base/block.py in __getattr__(self, val)
    522         # throw the "normal" AttributeError
    523         raise AttributeError("'%s' object has no attribute '%s'"
--> 524                              % (self.__class__.__name__, val))
    525
    526     def __setattr__(self, name, val):

AttributeError: '_BlockData' object has no attribute '_parent'
```

jsiirola:
## Summary/Motivation:
This resolves #2008 and allows individual `_BlockData` to be `cloned()`.  Note that the result is itself a `_BlockData` and should not be directly attached to a model.  Instead, the result should be passed to another block initializer / rule.  For example:

```python
m = ConcreteModel()
@m.Block([1,2,3])
def blk(b, i):
    b.IDX = RangeSet(i)
    b.x = Var(b.IDX)

m.c = Block(rule=m.blk[2].clone())

@m.Block([1,2,3])
def d(b, i):
    return b.model().blk[i].clone()
```

- Resolve exception when cloning individual `_BlockData`
- Add tests for expected behavior

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/block.py
+++ b/pyomo/core/base/block.py
@@ -1329,7 +1329,6 @@ Components must now specify their rules explicitly using 'rule=' keywords.""" %
         # NonNegativeReals, etc) that are not "owned" by any blocks and
         # should be preserved as singletons.
         #
-        save_parent, self._parent = self._parent, None
         try:
             new_block = copy.deepcopy(
                 self, {
@@ -1342,8 +1341,13 @@ Components must now specify their rules explicitly using 'rule=' keywords.""" %
                     '__block_scope__': {id(self): True, id(None): False},
                     '__paranoid__': True,
                     })
-        finally:
-            self._parent = save_parent
+
+        # We need to "detangle" the new block from the original block
+        # hierarchy
+        if self.parent_component() is self:
+            new_block._parent = None
+        else:
+            new_block._component = None
 
         return new_block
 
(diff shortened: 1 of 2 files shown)


blnicho (approved):

michaelbynum:
I realize something like `copy_from` is possible with rules like in the example you gave. Sometimes it is just nice to not need a rule.

michaelbynum (approved):

michaelbynum:
I guess `m.c = Block(rule=m.blk[2].clone())` did not need a rule. Okay, forget I said anything.

PR #2506, opened 2022-08-24
Mindtpy rewrite

ZedongPeng:
## Summary/Motivation:
Following #2406 , MindtPy rewrite will include
- change the integrated main loop into a base class and several child class for each algorithm.
- put the utility functions in the base class and put the specific functions in the corresponding child class.
- restructure `solve_data`.

## TBD
For some functions like `_get_discrete_problem_and_subproblem()`, I not sure if we should define them as an independent function or a method of the base class. It can absolutely be defined as a method. 
Can we define all the functions as methods?

Notion notes: https://uncovered-plantain-7a1.notion.site/MindtPy-write-06ff9fccab8f43c98e6bc6b2abec8aeb

[MindtPy_write.pdf](https://github.com/Pyomo/pyomo/files/9417227/MindtPy_write.pdf)

### Legal Acknowledgement


bernalde:
I'm tagging in this (currently) blank PR @michaelbynum @emma58 @carldlaird @jsiirola (and any other interested @Pyomo/pyomo-developers ) to chime in with opinions on how could we better integrate the MINLP/decomposition tools in Pyomo. We currently have efforts such as [Coramin](https://github.com/Coramin/Coramin) which might make sense to integrate in this rewrite (or at least make it more amenable to use together with MindtPy or whatever comes out of this).
Our main idea is to follow the GDPOpt rewrite PR #2406 and use a similar class structure for this case, with the main exception of the "single-tree" approaches which rely on direct solver interfaces and might benefit more from APPSI.
If you have the time please review the PDF attached above and leave any comments here to discuss.
One final thought is that GDPOpt is able to work with external function models. Can we modify the functionality of MindtPy to work similarly (I can imagine this should be restricted by forcing the integer variables to only appear in explicit/non-external constraints). This is motivated by some DISPATCHES examples.

blnicho:
@bernalde @ZedongPeng closing this, please open an issue to track discussion for the MindtPy rewrite.

PR #2512, opened 2022-08-26
Add IIS Tool contrib package

bknueven:
## Summary/Motivation:
This formalizes a workflow I've used for a while to compute the IIS of an infeasible Pyomo model using the persistent solver interface of a commercial solver.

- Add Infeasible Irreducible System (IIS) Tool to contrib.

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/iis/iis.py
+++ b/pyomo/contrib/iis/iis.py
@@ -5,10 +5,14 @@ Gurobi, or Xpress.
 """
 
 import abc
+import logging
 from pyomo.environ import SolverFactory
 
+logger = logging.getLogger("pyomo.contrib.iis")
+logger.setLevel(logging.INFO)
 
-def write_iis(pyomo_model, iis_file_name, solver=None):
+
+def write_iis(pyomo_model, iis_file_name, solver=None, logger=logger):
     """
     Write an irreducible infeasible set for a Pyomo MILP or LP
     using the specified commerical solver.
@@ -22,6 +26,8 @@ def write_iis(pyomo_model, iis_file_name, solver=None):
         solver:str
             Specify the solver to use, one of "cplex", "gurobi", or "xpress".
             If None, the tool will use the first solver available.
+        logger:logging.Logger
+            A logger for messages. Uses pyomo.contrib.iis logger by default.
 
     Returns
     -------
@@ -37,9 +43,11 @@ def write_iis(pyomo_model, iis_file_name, solver=None):
 
     if solver is None:
         if len(available_solvers) == 0:
-            raise RuntimeError("Could not find a solver to use")
-        solver = _available_solvers[0]
-        print(f"Using solver {solver}")
+            raise RuntimeError(
+                f"Could not find a solver to use, supported solvers are {_supported_solvers}"
+            )
+        solver = available_solvers[0]
+        logger.info(f"Using solver {solver}")
     else:
         # validate
         solver = solver.lower()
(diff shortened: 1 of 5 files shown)


pyomo/contrib/iis/iis.py
```diff
@@ -0,0 +1,136 @@
...
+    available_solvers = [
+        s
+        for s, sp in zip(_supported_solvers, _supported_solvers_persistent)
+        if SolverFactory(sp).available(exception_flag=False)
+    ]
+
+    if solver is None:
+        if len(available_solvers) == 0:
+            raise RuntimeError("Could not find a solver to use")
+        solver = _available_solvers[0]
```
michaelbynum, line 41: is _available_solvers defined? It looks like there is no underscore above.

pyomo/contrib/iis/iis.py
```diff
@@ -0,0 +1,136 @@
...
+        print(f"Using solver {solver}")
```
michaelbynum, line 42: I would replace all print statements with logging statements so they can be turned off if needed.

pyomo/contrib/iis/tests/test_iis.py
```diff
@@ -0,0 +1,75 @@
...
+    m.c2 = pyo.Constraint(expr=m.y <= -100.0 * m.x)
+    m.c3 = pyo.Constraint(expr=m.x >= 0.5)
+
+    m.o = pyo.Objective(expr=-m.y)
+
+    return m
+
+
+class TestIIS(unittest.TestCase):
+    @unittest.skipUnless(
```
michaelbynum, line 24: The tests look great. I would just recommend adding one more test for the case where the solver is not specified:

```
  @unittest.skipUnless(
         (pyo.SolverFactory("cplex_persistent").available(exception_flag=False) or
          pyo.SolverFactory("gurobi_persistent").available(exception_flag=False) or
          pyo.SolverFactory("xpress_persistent").available(exception_flag=False)),
         "Persistent solver not available",
     )
     def test_write_iis_any_solver(self):
         _test_iis(None)
```

I think this should work without changing any of the other test code???

michaelbynum (changes requested):
This is awesome! Thanks, @bknueven!

pyomo/contrib/iis/iis.py
```diff
@@ -0,0 +1,136 @@
...
+
+    iis = IISFactory(solver)
+    iis.compute()
+    iis_file_name = iis.write(iis_file_name)
+    print(f"IIS written to {iis_file_name}")
+    return iis_file_name
+
+
+def _remove_suffix(string, suffix):
+    if suffix in string:
```
jsiirola, line 65: This should be `if string.ends_with(suffix):`

pyomo/contrib/iis/iis.py
```diff
@@ -0,0 +1,136 @@
...
+            return file_name
+        else:
+            return _remove_suffix(file_name, ".lp") + ".lp"
+
+
+_solver_map = {
+    "cplex_persistent": CplexConflict,
+    "gurobi_persistent": GurobiIIS,
+    "xpress_persistent": XpressIIS,
+}
```
jsiirola, line 126: @michaelbynum, How challenging would it be to add support for the APPSI interfaces as well?

jsiirola (commented):
Thanks, @bknueven! Some questions / requests, but overall, this is great!

pyomo/contrib/iis/iis.py
michaelbynum: Not hard for Gurobi. I still have not refactored Appsi's interface to cplex to use the python bindings.

bknueven:
Thanks for the thorough review @michaelbynum and @jsiirola. I believe I have addressed all of your comments.

I noticed incidentally that this does not quite work for Cplex 12.10 (and probably not the whole 12.x series). I may have a go at correcting that in the next few days.

I've tested locally the oldest release of Xpress available on pypi, and as for Gurobi I think its IIS API has been the same since at least version 5.6.

michaelbynum (approved):
I think it is okay if we do not support this functionality for older versions of solvers.

jsiirola (approved):

PR #2514, opened 2022-08-28
Fix #2475 - Implement "thread safe" proxies for singleton instances (PauseGC, TempFileManager)

### LINKED ISSUE #1125: Problem solving in multithreading
Hi, I am trying to solve a stochastic problem. The problem for 24 steps and in each step I solve it for 450 different instances.
Now it is possible to solve the problem serially, but that takes a lot of time, so i wanted to solve those 450 instances parallely. 
I approached this in 2 ways:
1) using solver_manager and multiple pyro mip servers. Here I just have to queue the insances and the solver manager would assign the problems to pyro mip servers. But the pyro mip servers occupy a lot of memory and retain it even after the final result is computed. 
2) using opt solver.
a) I can solve using the opt solver directly, but it is slow. 
b) using multi threading. I tried to create a threadPoolExecutor with 5 threads and then solve those 450 instances in different threads. But just shifting to multithreading causes the pyomo to throw errors. 

Code for this approach - 
`
def optimize():

    optsolver = SolverFactory(self.solver_name)
    value = self.initialize_all_values()
    for timestep in range(24):
        instance_list = self.create_instance_list(value)
        futures = {}
        with concurrent.futures.ThreadPoolExecutor(max_workers=1) as executor:
            for instance_object in instance_list:
                futures = {executor.submit(self.thread_solver, instance_object["instance"], optsolver, timestep)}
            for future in concurrent.futures.as_completed(futures):
                try:
                    result = future.result()
                    value.update(result)
                except Exception as e:
                    self.logger.error(e)
`
`
def thread_solver(instance, optsolver, timestep):

    result = optsolver.solve(instance)
    if (result.solver.status == SolverStatus.ok) and (
            result.solver.termination_condition == TerminationCondition.optimal):
        instance.solutions.load_from(result)\
        my_dict = {}
        for v in instance.component_objects(Var, active=True):
            varobject = getattr(instance, str(v))
            var_list = []
            try:
                for index in varobject:
                    var_list.append(varobject[index].value)
                my_dict[str(v)] = var_list
            except Exception as e:
                self.logger.error("error reading result " + str(e))
        value = {key: my_dict[k][0]}
        return value
    elif result.solver.termination_condition == TerminationCondition.infeasible:
        self.logger.info("Termination condition is infeasible")
        return {}
    else:
        self.logger.info("Nothing fits")
        return {}
`
I get the following errors/outputs:
1) Nothing fits
2) Solver failed to locate input problem file: /usr/src/app/temp/pyomo/tmpn23yhmz_.pyomo.lp
I have set the pyomo temp file location to "/usr/src/app/temp/pyomo"

The same code runs perfectly fine without multithreading. 
Any reason for the pyomo to not solve the problem?

### LINKED ISSUE #2475: Can't run solve() inside a ThreadPool with GLPK
## Summary

From https://stackoverflow.com/questions/73064859/pyomo-how-do-i-run-multiple-solve-concurrently-in-parallel-on-the-same-model:

> I'm using opt = pyo.SolverFactory('glpk'), and I'd like to call opt.solve(), multiple times, inside a map function of a ThreadPool, but it gives error caused by race conditions.
>
> 
> It doesn't work even if I instanciate opt inside the map function.
> 
> The errors I get are random, but in many cases they refer to a log_file which can't be opened.
> 
> I'm even cloning the model for each map function call, to avoid race conditions on it.
> 
> I can't use MPI, I need to use a ThreadPool (which parallelizes only I/O ops, like waiting for the solver to end the task, which is good enough for me)

```python
import pyomo.environ as pyo
from multiprocessing.dummy import Pool as ThreadPool

model = pyo.ConcreteModel()
model.nVars = pyo.Param(initialize=4)
model.N = pyo.RangeSet(model.nVars)
model.x = pyo.Var(model.N, within=pyo.Binary)
model.obj = pyo.Objective(expr=pyo.summation(model.x))
model.cuts = pyo.ConstraintList()

def test(_):
    opt = SolverFactory('glpk')
    opt.solve(model) 

    # Iterate, adding a cut to exclude the previously found solution
    for i in range(5):
        expr = 0
        for j in model.x:
            if pyo.value(model.x[j]) < 0.5:
                expr += model.x[j]
            else:
                expr += (1 - model.x[j])
        model.cuts.add( expr >= 1 )
        results = opt.solve(model)
        return results

tp = ThreadPool(4)
results = tp.map(test, range(5))
print(results)
```
or:
```python
import pyomo.environ as pyo
from multiprocessing.dummy import Pool as ThreadPool

model = pyo.ConcreteModel()
model.nVars = pyo.Param(initialize=4)
model.N = pyo.RangeSet(model.nVars)
model.x = pyo.Var(model.N, within=pyo.Binary)
model.obj = pyo.Objective(expr=pyo.summation(model.x))
model.cuts = pyo.ConstraintList()

def test(model):
    opt = SolverFactory('glpk')
    opt.solve(model) 

    # Iterate, adding a cut to exclude the previously found solution
    for i in range(5):
        expr = 0
        for j in model.x:
            if pyo.value(model.x[j]) < 0.5:
                expr += model.x[j]
            else:
                expr += (1 - model.x[j])
        model.cuts.add( expr >= 1 )
        results = opt.solve(model)
        return results

tp = ThreadPool(4)
results = tp.map(test, [model.clone() for i in range(4)])
print(results)
```

### Error Message

```console
WARNING: Empty constraint block written in LP format - solver may error
WARNING: Empty constraint block written in LP format - solver may error
WARNING: Empty constraint block written in LP format - solver may error
WARNING: Empty constraint block written in LP format - solver may error
  File "c:\Users\_\Desktop\problem.py", line 29, in <module>
    results = tp.map(test, [model.clone() for i in range(4)])
  File "D:\Programs\Python310\lib\multiprocessing\pool.py", line 364, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "D:\Programs\Python310\lib\multiprocessing\pool.py", line 771, in get
    raise self._value
  File "D:\Programs\Python310\lib\multiprocessing\pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "D:\Programs\Python310\lib\multiprocessing\pool.py", line 48, in mapstar
    return list(map(*args))
  File "c:\Users\_\Desktop\problem.py", line 14, in test
    opt.solve(model)
  File "D:\Programs\Python310\lib\site-packages\pyomo\opt\base\solvers.py", line 602, in solve
    result = self._postsolve()
  File "D:\Programs\Python310\lib\site-packages\pyomo\opt\solver\shellcmd.py", line 281, in _postsolve
    results = self.process_output(self._rc)
  File "D:\Programs\Python310\lib\site-packages\pyomo\opt\solver\shellcmd.py", line 354, in process_output
    results = self.process_logfile()
  File "D:\Programs\Python310\lib\site-packages\pyomo\solvers\plugins\solvers\GLPK.py", line 213, in process_logfile
    with open(self._log_file, 'r') as output:
OSError: [WinError 6] The handle is invalid
Exception ignored in: <function TeeStream.__del__ at 0x00000237C62AE830>
  File "D:\Programs\Python310\lib\site-packages\pyomo\common\tee.py", line 425, in __del__
    self.close()
  File "D:\Programs\Python310\lib\site-packages\pyomo\common\tee.py", line 407, in close
    h.finalize(self.ostreams)
  File "D:\Programs\Python310\lib\site-packages\pyomo\common\tee.py", line 267, in finalize
    os.close(self.read_pipe)
OSError: [Errno 9] Bad file descriptor
```

Pyomo version: 6.4.1
Python version: 3.10.5
Operating system: Windows
Solver (if applicable): GLPK

### Additional information

I posted on the forum too, but none responded:
https://groups.google.com/g/pyomo-forum/c/CovfenBC0wg

ldeluigi:
# Fixes #2475
# Fixes #1125
## Summary/Motivation:
The purpose is to support thread based parallelism of solve calls in python. This is possible despite the GIL because the solve calls wait for the solver child processes to end, and this is a I/O blocking operation.

In order to avoid major code refactors and problems, I implemented a proxy that turns global instances (aka singletons) into thread-specific objects. For the usual use-case, this change should be almost unnoticeable, while for a multithreaded use of these singletons, now each of them is a different and totally independent instance of the same type for each thread. This avoids race conditions. The idea is inspired by process oriented parallelism, which is already supported with MPI.

- Added a new file in pyomo.common: multithread.py.  
  This file contains the boilerplate code that is common to all proxies for singletons.
- Replaced these singletons with refactored+proxied versions of themselves:
  - PauseGC class fields, using the Companion pattern
  - TempFileManager
  
The changes pass the tests and should solve the [issue above](https://github.com/Pyomo/pyomo/issues/2475): I tested in local, and I have created unit tests.

As asked in the [issue](https://github.com/Pyomo/pyomo/issues/2475), I implemented a version of the MultiThread proxy that considers the main thread instance of the singleton as special, exposing it in a ".main" field. This could create problems if any proxied class has its own "main" field definition: I'm open for suggestions for an improvement of this, but I think it's an unavoidable issue.

I used type hints for proxied singletons otherwise autocompletion would not work anymore on those.

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/multithread.py
+++ b/pyomo/common/multithread.py
@@ -3,6 +3,16 @@ from threading import get_ident, current_thread, main_thread
 
 
 class MultiThreadWrapper():
+    """A python object proxy that wraps different instances for each thread.
+
+    This is useful for handling thread-safe access to singleton objects without
+    having to refactor the code that depends on them.
+
+    Note that instances of the wrapped object are reused if two threads share the same
+    identifier, because identifiers could be reused and are unique only for any given moment.
+    See [get_ident()](https://docs.python.org/3/library/threading.html#threading.get_ident) for more information.
+    """
+
     def __init__(self, base):
         self.__mtdict = defaultdict(base)
 
@@ -27,6 +37,13 @@ class MultiThreadWrapper():
 
 
 class MultiThreadWrapperWithMain(MultiThreadWrapper):
+    """An extension of `MultiThreadWrapper` that exposes the wrapped instance
+    corresponding to the [main_thread()](https://docs.python.org/3/library/threading.html#threading.main_thread)
+    under the `.main_thread` field.
+
+    This is useful for a falling back to a main instance when needed, but results
+    in race conditions if used improperly.
+    """
     def __init__(self, base):
         super().__init__(base)
         self.main_thread = base()
--- a/pyomo/common/multithread.py
+++ b/pyomo/common/multithread.py
@@ -1,5 +1,5 @@
 from collections import defaultdict
-from threading import get_ident, current_thread, main_thread
+from threading import get_ident, main_thread
 
 
 class MultiThreadWrapper():
@@ -46,29 +46,16 @@ class MultiThreadWrapperWithMain(MultiThreadWrapper):
     """
     def __init__(self, base):
         super().__init__(base)
-        self.main_thread = base()
 
     def __getattr__(self, attr):
-        if current_thread() is main_thread():
-            return getattr(self.main_thread, attr)
-        return super().__getattr__(attr)
-    
-    def __setattr__(self, attr, value):
-        if attr == '_MultiThreadWrapper__mtdict':
-            super().__setattr__(attr, value)
+        if attr == '_MultiThreadWrapperWithMain__mtdict':
+            return object.__getattribute__(self, '_MultiThreadWrapper__mtdict')
         elif attr == 'main_thread':
-            object.__setattr__(self, attr, value)
-        elif current_thread() is main_thread():
-            setattr(self.main_thread, attr, value)
+            return self.__mtdict[main_thread().ident]
+        return super().__getattr__(attr)
+
+    def __setattr__(self, attr, value):
+        if attr == 'main_thread':
+            raise ValueError('Setting `main_thread` attribute is not allowed')
         else:
             super().__setattr__(attr, value)
-    
-    def __enter__(self):
-        if current_thread() is main_thread():
-            return self.main_thread.__enter__()
-        return super().__enter__()
-    
-    def __exit__(self, exc_type, exc_value, traceback):
-        if current_thread() is main_thread():
-            return self.main_thread.__exit__(exc_type, exc_value, traceback)
-        return super().__exit__(exc_type, exc_value, traceback)
--- a/pyomo/common/multithread.py
+++ b/pyomo/common/multithread.py
@@ -1,0 +1,14 @@
+from collections import defaultdict
+from threading import get_ident
+
+
+class MultiThreadWrapper(object):
+    def __init__(self, base):
+        self.__mtdict = defaultdict(base)
+
+    def __getattr__(self, attr):
+        try:
+            return super(MultiThreadWrapper, self).__getattr__(attr)
+        except AttributeError:
+            id = get_ident()
+            return getattr(self.__mtdict[id], attr)
--- a/pyomo/common/multithread.py
+++ b/pyomo/common/multithread.py
@@ -39,6 +39,8 @@ class MultiThreadWrapper():
     def __str__(self):
         return self.mtdict[get_ident()].__str__()
 
+    def __new__(cls, wrapped):
+        return super().__new__(type('MultiThreadMeta' + wrapped.__name__, (cls,), {'__doc__': wrapped.__doc__}))
 
 
 class MultiThreadWrapperWithMain(MultiThreadWrapper):
@@ -49,6 +51,7 @@ class MultiThreadWrapperWithMain(MultiThreadWrapper):
     This is useful for a falling back to a main instance when needed, but results
     in race conditions if used improperly.
     """
+
     def __init__(self, base):
         super().__init__(base)
 
--- a/pyomo/common/multithread.py
+++ b/pyomo/common/multithread.py
@@ -13,31 +13,31 @@ class MultiThreadWrapper():
     See [get_ident()](https://docs.python.org/3/library/threading.html#threading.get_ident) for more information.
     """
 
-    __slots__ = 'mtdict'
+    __slots__ = '_mtdict'
 
     def __init__(self, base):
-        object.__setattr__(self, 'mtdict', defaultdict(base))
+        object.__setattr__(self, '_mtdict', defaultdict(base))
 
     def __getattr__(self, attr):
-        return getattr(self.mtdict[get_ident()], attr)
+        return getattr(self._mtdict[get_ident()], attr)
     
     def __setattr__(self, attr, value):
-        setattr(self.mtdict[get_ident()], attr, value)
+        setattr(self._mtdict[get_ident()], attr, value)
     
     def __delattr__(self, attr):
-        delattr(self.mtdict[get_ident()], attr)
+        delattr(self._mtdict[get_ident()], attr)
     
     def __enter__(self):
-        return self.mtdict[get_ident()].__enter__()
+        return self._mtdict[get_ident()].__enter__()
     
     def __exit__(self, exc_type, exc_value, traceback):
-        return self.mtdict[get_ident()].__exit__(exc_type, exc_value, traceback)
+        return self._mtdict[get_ident()].__exit__(exc_type, exc_value, traceback)
     
     def __dir__(self):
-        return list(object.__dir__(self)) + list(self.mtdict[get_ident()].__dir__())
+        return list(object.__dir__(self)) + list(self._mtdict[get_ident()].__dir__())
     
     def __str__(self):
-        return self.mtdict[get_ident()].__str__()
+        return self._mtdict[get_ident()].__str__()
 
     def __new__(cls, wrapped):
         return super().__new__(type('MultiThreadMeta' + wrapped.__name__, (cls,), {'__doc__': wrapped.__doc__}))
@@ -57,7 +57,7 @@ class MultiThreadWrapperWithMain(MultiThreadWrapper):
 
     def __getattr__(self, attr):
         if attr == 'main_thread':
-            return self.mtdict[main_thread().ident]
+            return self._mtdict[main_thread().ident]
         return super().__getattr__(attr)
 
     def __setattr__(self, attr, value):
--- a/pyomo/common/multithread.py
+++ b/pyomo/common/multithread.py
@@ -29,29 +29,29 @@ class MultiThreadWrapper():
 class MultiThreadWrapperWithMain(MultiThreadWrapper):
     def __init__(self, base):
         super().__init__(base)
-        self.main = base()
+        self.main_thread = base()
 
     def __getattr__(self, attr):
         if current_thread() is main_thread():
-            return getattr(self.main, attr)
+            return getattr(self.main_thread, attr)
         return super().__getattr__(attr)
     
     def __setattr__(self, attr, value):
         if attr == '_MultiThreadWrapper__mtdict':
             super().__setattr__(attr, value)
-        elif attr == 'main':
+        elif attr == 'main_thread':
             object.__setattr__(self, attr, value)
         elif current_thread() is main_thread():
-            setattr(self.main, attr, value)
+            setattr(self.main_thread, attr, value)
         else:
             super().__setattr__(attr, value)
     
     def __enter__(self):
         if current_thread() is main_thread():
-            return self.main.__enter__()
+            return self.main_thread.__enter__()
         return super().__enter__()
     
     def __exit__(self, exc_type, exc_value, traceback):
         if current_thread() is main_thread():
-            return self.main.__exit__(exc_type, exc_value, traceback)
+            return self.main_thread.__exit__(exc_type, exc_value, traceback)
         return super().__exit__(exc_type, exc_value, traceback)
--- a/pyomo/common/multithread.py
+++ b/pyomo/common/multithread.py
@@ -2,16 +2,13 @@ from collections import defaultdict
 from threading import get_ident, current_thread, main_thread
 
 
-class MultiThreadWrapper(object):
+class MultiThreadWrapper():
     def __init__(self, base):
         self.__mtdict = defaultdict(base)
 
     def __getattr__(self, attr):
-        try:
-            return super(MultiThreadWrapper, self).__getattr__(attr)
-        except AttributeError:
-            id = get_ident()
-            return getattr(self.__mtdict[id], attr)
+        id = get_ident()
+        return getattr(self.__mtdict[id], attr)
 
 
 class MultiThreadWrapperWithMain(MultiThreadWrapper):
--- a/pyomo/common/multithread.py
+++ b/pyomo/common/multithread.py
@@ -1,5 +1,5 @@
 from collections import defaultdict
-from threading import get_ident
+from threading import get_ident, current_thread, main_thread
 
 
 class MultiThreadWrapper(object):
@@ -12,3 +12,14 @@ class MultiThreadWrapper(object):
         except AttributeError:
             id = get_ident()
             return getattr(self.__mtdict[id], attr)
+
+
+class MultiThreadWrapperWithMain(MultiThreadWrapper):
+    def __init__(self, base):
+        super().__init__(base)
+        self.main = base()
+
+    def __getattr__(self, attr):
+        if current_thread() is main_thread():
+            return getattr(self.main, attr)
+        return super().__getattr__(attr)
--- a/pyomo/common/multithread.py
+++ b/pyomo/common/multithread.py
@@ -16,6 +16,14 @@ class MultiThreadWrapper():
         else:
             id = get_ident()
             setattr(self.__mtdict[id], attr, value)
+    
+    def __enter__(self):
+        id = get_ident()
+        return self.__mtdict[id].__enter__()
+    
+    def __exit__(self, exc_type, exc_value, traceback):
+        id = get_ident()
+        return self.__mtdict[id].__exit__(exc_type, exc_value, traceback)
 
 
 class MultiThreadWrapperWithMain(MultiThreadWrapper):
@@ -37,3 +45,13 @@ class MultiThreadWrapperWithMain(MultiThreadWrapper):
             setattr(self.main, attr, value)
         else:
             super().__setattr__(attr, value)
+    
+    def __enter__(self):
+        if current_thread() is main_thread():
+            return self.main.__enter__()
+        return super().__enter__()
+    
+    def __exit__(self, exc_type, exc_value, traceback):
+        if current_thread() is main_thread():
+            return self.main.__exit__(exc_type, exc_value, traceback)
+        return super().__exit__(exc_type, exc_value, traceback)
--- a/pyomo/common/multithread.py
+++ b/pyomo/common/multithread.py
@@ -17,23 +17,19 @@ class MultiThreadWrapper():
         self.__mtdict = defaultdict(base)
 
     def __getattr__(self, attr):
-        id = get_ident()
-        return getattr(self.__mtdict[id], attr)
+        return getattr(self.__mtdict[get_ident()], attr)
     
     def __setattr__(self, attr, value):
         if attr == '_MultiThreadWrapper__mtdict':
             object.__setattr__(self, attr, value)
         else:
-            id = get_ident()
-            setattr(self.__mtdict[id], attr, value)
+            setattr(self.__mtdict[get_ident()], attr, value)
     
     def __enter__(self):
-        id = get_ident()
-        return self.__mtdict[id].__enter__()
+        return self.__mtdict[get_ident()].__enter__()
     
     def __exit__(self, exc_type, exc_value, traceback):
-        id = get_ident()
-        return self.__mtdict[id].__exit__(exc_type, exc_value, traceback)
+        return self.__mtdict[get_ident()].__exit__(exc_type, exc_value, traceback)
 
 
 class MultiThreadWrapperWithMain(MultiThreadWrapper):
--- a/pyomo/common/multithread.py
+++ b/pyomo/common/multithread.py
@@ -14,22 +14,19 @@ class MultiThreadWrapper():
     """
 
     def __init__(self, base):
-        self.__mtdict = defaultdict(base)
+        object.__setattr__(self, 'mtdict', defaultdict(base))
 
     def __getattr__(self, attr):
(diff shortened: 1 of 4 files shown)


pyomo/common/multithread.py
```diff
@@ -0,0 +1,57 @@
...
+    def __getattr__(self, attr):
+        id = get_ident()
+        return getattr(self.__mtdict[id], attr)
+    
+    def __setattr__(self, attr, value):
+        if attr == '_MultiThreadWrapper__mtdict':
+            object.__setattr__(self, attr, value)
+        else:
+            id = get_ident()
+            setattr(self.__mtdict[id], attr, value)
```
jsiirola, line 18: I would suggest reimplementing this as:
```python
def __getattr__(self, attr):
    if attr == 'main_thread':
        return self.__mtdict[main_thread().ident]
    else:
        id = get_ident()
        return getattr(self.__mtdict[id], attr)

def __setattr__(self, attr, value):
    if attr == '_MultiThreadWrapper__mtdict':
        object.__setattr__(self, attr, value)
    elif attr == 'main_thread':
        raise ValueError('Setting `main_thread` attribute is not allowed')
    else:
        id = get_ident()
        setattr(self.__mtdict[id], attr, value)
```

pyomo/common/multithread.py
```diff
@@ -0,0 +1,57 @@
+from collections import defaultdict
+from threading import get_ident, current_thread, main_thread
+
+
+class MultiThreadWrapper():
```
jsiirola, line 5: Some documentation would be appreciated.  :)

pyomo/common/tempfiles.py
```diff
@@ -428,6 +429,8 @@ def _resolve_tempdir(self, dir=None):
             return self.tempdir
         elif self.manager().tempdir is not None:
             return self.manager().tempdir
+        elif TempfileManager is not None and TempfileManager.main.tempdir is not None:
```
jsiirola, line 432: I don't think you need the test for `TempfileManager is not None`, do you?

jsiirola (commented):
This is really nice!

I would like to suggest one change, though.  I do not think that you need two classes.  Plus, I think that for the current implementation of `MultiThreadWrapperWithMain`, you are ending up with two instances of the wrapper class for the main thread.  Instead, I would implement a single `MultiThreadWrapper` class that does special handling for the `main_thread` attribute, returning the object registered for the main thread.

(I also think that `main_thread` may be more specific / clear than just `main`, but could be convinced otherwise).

pyomo/common/multithread.py
ldeluigi: The idea behind having two classes is that in some cases (i.e. PauseGC) the main thread instance reference is not needed (in fact I used `MultiThreadWrapper` instead of `MultiThreadWrapperWithMain`). I like the clear, almost self-explanatory way `MultiThreadWrapper` is implemented. MultiThreadWrapperWithMain is just a different way of treating the instance corresponding to the main_thread() id.

In general, I like to separate responsibilities of classes as a principle.

Nevertheless, I like the name `main_thread` instead of just main.

> I think that for the current implementation of `MultiThreadWrapperWithMain`, you are ending up with two instances of the wrapper class for the main thread.

How could it happen? I think this would be a fixable bug if so.

pyomo/common/tempfiles.py
ldeluigi: You are right, I'm removing that check now

pyomo/common/multithread.py
ldeluigi: I just thought of a simpler implementation of MultiThreadWrapperWithMain.. let me know what you think

ldeluigi: I'll try to write a simpler implementation of the wrapper with main

ldeluigi: Done. Please, review the newer one,

pyomo/common/multithread.py
```diff
@@ -0,0 +1,61 @@
...
+    in race conditions if used improperly.
+    """
+    def __init__(self, base):
+        super().__init__(base)
+
+    def __getattr__(self, attr):
+        if attr == '_MultiThreadWrapperWithMain__mtdict':
+            return object.__getattribute__(self, '_MultiThreadWrapper__mtdict')
+        elif attr == 'main_thread':
+            return self.__mtdict[main_thread().ident]
```
jsiirola, line 54: I know I suggested this approach, but now that I am thinking more clearly, should this be a `@property`?  That way we can document it more easily in a docstring.

pyomo/common/multithread.py
```diff
@@ -0,0 +1,61 @@
...
+    def __init__(self, base):
+        self.__mtdict = defaultdict(base)
+
+    def __getattr__(self, attr):
+        id = get_ident()
+        return getattr(self.__mtdict[id], attr)
+    
+    def __setattr__(self, attr, value):
+        if attr == '_MultiThreadWrapper__mtdict':
+            object.__setattr__(self, attr, value)
```
jsiirola, line 25: Why do you allow overwriting the `__mtdict` (but prevent it for the `main_thread`)?

pyomo/common/multithread.py
```diff
@@ -0,0 +1,61 @@
...
+
+    This is useful for a falling back to a main instance when needed, but results
+    in race conditions if used improperly.
+    """
+    def __init__(self, base):
+        super().__init__(base)
+
+    def __getattr__(self, attr):
+        if attr == '_MultiThreadWrapperWithMain__mtdict':
+            return object.__getattribute__(self, '_MultiThreadWrapper__mtdict')
```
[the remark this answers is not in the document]
ldeluigi: The idea was to avoid as much as possible collisions with name inside the wrapped classes

pyomo/common/multithread.py
```diff
@@ -0,0 +1,61 @@
...
+    Note that instances of the wrapped object are reused if two threads share the same
+    identifier, because identifiers could be reused and are unique only for any given moment.
+    See [get_ident()](https://docs.python.org/3/library/threading.html#threading.get_ident) for more information.
+    """
+
+    def __init__(self, base):
+        self.__mtdict = defaultdict(base)
+
+    def __getattr__(self, attr):
+        id = get_ident()
```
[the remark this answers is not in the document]
ldeluigi: ok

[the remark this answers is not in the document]
ldeluigi: I forgot id was a builtin

pyomo/common/multithread.py
ldeluigi: I think that, under the hood, property don't go well with setattr and getattr overrides... I'd avoid risking with a different implementation... can we document it as @property without changing the implementation?

pyomo/common/multithread.py
ldeluigi: Because, inside the __init__ I'm doing `self.__mtdict = ...` which calls the __setattr__

pyomo/common/multithread.py
```diff
@@ -0,0 +1,61 @@
...
+    This is useful for handling thread-safe access to singleton objects without
+    having to refactor the code that depends on them.
+
+    Note that instances of the wrapped object are reused if two threads share the same
+    identifier, because identifiers could be reused and are unique only for any given moment.
+    See [get_ident()](https://docs.python.org/3/library/threading.html#threading.get_ident) for more information.
+    """
+
+    def __init__(self, base):
+        self.__mtdict = defaultdict(base)
```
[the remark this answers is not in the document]
ldeluigi: At first I implemented it like that, but with that implementation I need to inherit from object in order to call super().__setattr__

Furthermore, with that implementation the visual studio debugger doesn't show __mtdict between the fields of the object... so I avoided doing that

pyomo/common/multithread.py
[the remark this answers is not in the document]
ldeluigi: But if 'mtdict' and 'main_thread' feel safe enough for you there's no problem for me, you say it and I remove the underscores

pyomo/common/multithread.py
jsiirola: Right - that was my point above: you could use `object.__setattr__` in `__init__` and then make the attribute not settable here (following what you do with `main_thread`).

pyomo/common/multithread.py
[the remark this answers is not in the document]
jsiirola: Everything inherits from object (it is just a Python 3-ism that you don't *have* to make it explicit):
```python
>>> from pyomo.common.multithread import *
>>> MultiThreadWrapper.__mro__
(<class 'pyomo.common.multithread.MultiThreadWrapper'>, <class 'object'>)
```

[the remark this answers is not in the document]
ldeluigi: I could do `object.__setattr__(self, ...` instead

jsiirola:
So this has got me digging into things...  I think we should also consider implementing:
```python
def __dir__(self):
    return self.__mtdict[get_ident()].__dir__()
```
do that autocompletion on the command line works.  I would also like to figure out how to get `help(TempfileManager)` to display the help for the wrapped (`base` / `TempfileManagerClass`) class and not the help for the `MultiThreadWrapperWithMain` class.  My initial forays into overloading the `__doc__` on the instance do not work.  I wonder if we need to go down the wrote of a metaclass / class factory pattern to get that to work?

ldeluigi:
I implemented (some) magic features of python proxies, but not all, in order to allow debugging MultiThreadWrapper.

See this package implementation: https://github.com/zyga/padme/blob/master/padme/__init__.py
It has interesting usage of python metaclasses.

I'll try to override the help string and then I'll push the changes

ldeluigi:
> So this has got me digging into things... I think we should also consider implementing:
> 
> ```python
> def __dir__(self):
>     return self.__mtdict[get_ident()].__dir__()
> ```
> 
> so that autocompletion on the command line works. I would also like to figure out how to get `help(TempfileManager)` to display the help for the wrapped (`base` / `TempfileManagerClass`) class and not the help for the `MultiThreadWrapperWithMain` class. My initial forays into overloading the `__doc__` on the instance do not work. I wonder if we need to go down the wrote of a metaclass / class factory pattern to get that to work?

I managed to do it in my last commit! 🎉

pyomo/common/multithread.py
```diff
@@ -0,0 +1,70 @@
...
+        return self.mtdict[get_ident()].__exit__(exc_type, exc_value, traceback)
+    
+    def __dir__(self):
+        return list(object.__dir__(self)) + list(self.mtdict[get_ident()].__dir__())
+    
+    def __str__(self):
+        return self.mtdict[get_ident()].__str__()
+
+    def __new__(cls, wrapped):
+        return super().__new__(type('MultiThreadMeta' + wrapped.__name__, (cls,), {'__doc__': wrapped.__doc__}))
```
jsiirola, line 43: Clever!

jsiirola:
I think this is good to go (pending the tests finishing).  @michaelbynum can you provide the 2nd review?

The only thing I saw that we could discuss / debate moving forward is if `mtdict` should be made "private" (i.e. `_mtdict`).  My knee-jerk reaction is "yes."  But, I also keep coming back to if it makes sense to have both `MultiThreadWrapper` and `MultiThreadWrapperWithMain`: we do have a use case (`TempfileManager`), where as part of attribute resolution it makes sense for threads to "inherit" the state from the main thread:
```python
        if dir is not None:
            return dir
        elif self.tempdir is not None:
            return self.tempdir
        elif self.manager().tempdir is not None:
            return self.manager().tempdir
        elif TempfileManager.main_thread.tempdir is not None:
            return TempfileManager.main_thread.tempdir
```
Right now (and this was my questionable idea), we support that through a public attribute `main_thread`, which resolves to the value in `mtdict` associated with the main thread.  I wonder if this is both redundant (with `mtdict`) and unnecessary.  An alternative could be to leave `mtdict` public, but potentially rename it to something more descriptive (`thread`?).  This would remove the need for the `main_thread` attribute (and the derived class).  `TempfileManager` could just do:
```python
        if dir is not None:
            return dir
        elif self.tempdir is not None:
            return self.tempdir
        elif self.manager().tempdir is not None:
            return self.manager().tempdir
        elif TempfileManager.thread[threading.main_thread()].tempdir is not None:
            return TempfileManager.thread[threading.main_thread()].tempdir
```
Thoughts?

ldeluigi:
> I think this is good to go (pending the tests finishing). @michaelbynum can you provide the 2nd review?
> 
> The only thing I saw that we could discuss / debate moving forward is if `mtdict` should be made "private" (i.e. `_mtdict`). My knee-jerk reaction is "yes." But, I also keep coming back to if it makes sense to have both `MultiThreadWrapper` and `MultiThreadWrapperWithMain`: we do have a use case (`TempfileManager`), where as part of attribute resolution it makes sense for threads to "inherit" the state from the main thread:
> 
> ```python
>         if dir is not None:
>             return dir
>         elif self.tempdir is not None:
>             return self.tempdir
>         elif self.manager().tempdir is not None:
>             return self.manager().tempdir
>         elif TempfileManager.main_thread.tempdir is not None:
>             return TempfileManager.main_thread.tempdir
> ```
> 
> Right now (and this was my questionable idea), we support that through a public attribute `main_thread`, which resolves to the value in `mtdict` associated with the main thread. I wonder if this is both redundant (with `mtdict`) and unnecessary. An alternative could be to leave `mtdict` public, but potentially rename it to something more descriptive (`thread`?). This would remove the need for the `main_thread` attribute (and the derived class). `TempfileManager` could just do:
> 
> ```python
>         if dir is not None:
>             return dir
>         elif self.tempdir is not None:
>             return self.tempdir
>         elif self.manager().tempdir is not None:
>             return self.manager().tempdir
>         elif TempfileManager.thread[threading.main_thread()].tempdir is not None:
>             return TempfileManager.thread[threading.main_thread()].tempdir
> ```
> 
> Thoughts?

#### On removing the specialized class for accessing the main thread instance
What if you want to add a mutex lock around concurrent accesses to mtdict items from other threads? What if we want the mutex only for the main_thread instance (which is a singleton in practice)? If you don't keep a specialized class implementation that encapsulates the behavior you need, you end up with technical debt.

#### On mtdict private vs. public:
In order to avoid race conditions (which is the purpose of the whole wrapper) I would go with private. I would discourage in any way possible directly interacting with such a data structure.
But we've seen that this comes with ugly code that needs to handle it.

#### On naming
##### mtdict
if mtdict gets to be private, the name isn't an issue. If it stays public, you can come up with any name you think won't collide with wrapped classes.
##### main_thread
idk, I'm open for better ideas

ldeluigi:
@jsiirola should I make mtdict back to private?

michaelbynum (approved):
Looks good pending tests.

mrmundt:
@ldeluigi - I am okay with this not holding up the PR, but if you are able, can you please add some minimal tests? I'd like there to be at least some code coverage on the new file.

ldeluigi:
I'll try to make some miningful test in my next spare time, sorry but I'm busy these days

pyomo/common/tests/test_multithread.py
```diff
@@ -0,0 +1,102 @@
...
+        sut.number = 5
+        def thread_func():
+            self.assertEqual(sut.number, 1)
+            self.assertIsNot(sut.main_thread.rnd, sut.rnd)
+            del sut.number
+        t = Thread(target=thread_func)
+        t.start()
+        t.join()
+        self.assertEqual(sut.number, 5)
+    def test_solve(self):
```
jsiirola, line 69: This test will need to be skipped if `glpk` is not available.

pyomo/common/tests/test_multithread.py
```diff
@@ -0,0 +1,102 @@
...
+class Dummy():
+    """asdfg"""
+    def __init__(self):
+        self.number = 1
+        self.rnd = threading.get_ident()
+    def __str__(self):
+        return 'asd'
+
+class TestMultithreading(unittest.TestCase):
+    def test_wrapper_docs(self):
```
jsiirola, line 15: PEP8: blank line between functions

pyomo/common/tests/test_multithread.py
```diff
@@ -0,0 +1,102 @@
...
+        model.x = pyo.Var(model.N, within=pyo.Binary)
+        model.obj = pyo.Objective(expr=pyo.summation(model.x))
+        model.cuts = pyo.ConstraintList()
+
+        def test(model):
+            opt = SolverFactory('glpk')
+            opt.solve(model) 
+
+            # Iterate, adding a cut to exclude the previously found solution
+            for i in range(5):
```
jsiirola, line 87: One question: this only executed a single iteration of this loop...  Should the return be unindented?

pyomo/common/tests/test_multithread.py
```diff
@@ -0,0 +1,102 @@
...
+                        expr += (1 - model.x[j])
+                model.cuts.add( expr >= 1 )
+                results = opt.solve(model)
+                return results
+
+        tp = ThreadPool(4)
+        results = tp.map(test, [model.clone() for i in range(4)])
+        tp.close()
+        for result in results:
+            self.assertEqual(result.solver.termination_condition, pyo.TerminationCondition.optimal)
```
jsiirola, line 102: Should we also verify that each thread returned the same answer?

jsiirola (commented):
This is great!  Thanks!  Some minor nits, passing tests, and we should merge this.

pyomo/common/tests/test_multithread.py
ldeluigi: Yes, I've just added the skipIf annotation. I hope it works, we will see when tests finish

pyomo/common/tests/test_multithread.py
ldeluigi: yes, my bad. I'm unindenting the return

pyomo/common/tests/test_multithread.py
ldeluigi: Ok

ldeluigi: Nvm, threads are solving different problems because of cuts

ldeluigi: Should I make a different test where each thread solves the same problem?

pyomo/common/tests/test_multithread.py
ldeluigi: Added

pyomo/common/tests/test_multithread.py
ldeluigi: Added

ldeluigi:
I should have fixed the tests

PR #2515, opened 2022-08-29
Modify PyROS Subproblem Initialization and Solver Call Routines

shermanjasonaf:
## Summary/Motivation:
Modify subproblem initialization and solver call routines such that:
- subproblem models are initialized to a feasible point, where possible
- non-optimal or exceptional solver termination statuses are handled more gracefully
- solutions returned by the appropriate solver(s) are properly loaded to the appropriate model(s) where appropriate.

**General**
- More comprehensive documentation for all subproblem solver call routines
- alter exception handling for subordinate optimizer `solve()` calls

**Master Feasibility Problem**
- ensure initialized to feasible point
- if solver does not succesfully converge to at least a feasible solution, load the initial point to the master model

**Master Problem**
- ensure initial point provided to each backup solver (if used) is the same, and not the final iterate returned by the previous backup solver

**DR Polishing Problem**
- use local subsolver if `solve_master_globally=False`, and global subsolver otherwise.
- load second-stage variable and state variable values from polished DR problem (to ensure polished master solution is feasible for the master model, and separation problems may therefore be initialized to a feasible point).
- ensure polished DR solution is loaded in event of `globallyOptimal` solver termination (resolve bug)

**Separation Problem**
- initialize to feasible point from either the nominal or most recently added master block (depending on the objective focus)
- ensure initial point provided to each backup solver (if used) is the same, and not final iterate returned by previous backup solver
- check for PyROS timeout before checking for optimality

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/pyros/master_problem_methods.py
+++ b/pyomo/contrib/pyros/master_problem_methods.py
@@ -197,7 +197,22 @@ def construct_master_feasibility_problem(model_data, config):
 
 def solve_master_feasibility_problem(model_data, config):
     """
-    Solve a slack variable based feasibility model for the master problem
+    Solve a slack variable-based feasibility model derived
+    from the master problem. Initialize the master problem
+    to the  solution found by the optimizer if solved successfully,
+    or to the initial point provided to the solver otherwise.
+
+    Parameters
+    ----------
+    model_data : MasterProblemData
+        Master problem data.
+    config : ConfigDict
+        PyROS solver settings.
+
+    Returns
+    -------
+    results : SolverResults
+        Solver results.
     """
     model = construct_master_feasibility_problem(model_data, config)
 
@@ -206,24 +221,28 @@ def solve_master_feasibility_problem(model_data, config):
     else:
         solver = config.local_solver
 
-    if not solver.available():
-        raise RuntimeError("NLP solver %s is not available." %
-                           config.solver)
-
-    results = solver.solve(model, tee=config.tee, load_solutions=False)
+    try:
+        results = solver.solve(model, tee=config.tee, load_solutions=False)
+    except ApplicationError:
+        # account for possible external subsolver errors
+        # (such as segmentation faults, function evaluation
+        # errors, etc.)
+        config.progress_logger.error(
+            f"Solver {repr(solver)} encountered exception attempting to "
+            "optimize master feasibility problem in iteration "
+            f"{model_data.iteration}"
+        )
+        raise
 
     if check_optimal_termination(results):
         model.solutions.load_from(results)
 
-        # load solution to master model
-        for v in model.component_data_objects(Var):
-            master_v = model_data.master_model.find_component(v)
-            if master_v is not None:
-                master_v.set_value(v.value, skip_validation=True)
-    else:
-        results.solver.termination_condition = tc.error
-        results.solver.message = ("Cannot load a SolverResults object with "
-                                  "bad status: error")
+    # load master feasibility point to master model
+    for v in model.component_data_objects(Var):
+        master_v = model_data.master_model.find_component(v)
+        if master_v is not None:
+            master_v.set_value(v.value, skip_validation=True)
+
     return results
 
 
--- a/pyomo/contrib/pyros/master_problem_methods.py
+++ b/pyomo/contrib/pyros/master_problem_methods.py
@@ -512,7 +512,7 @@ def solver_call_master(model_data, config, solver, solve_data):
 
     higher_order_decision_rule_efficiency(config, model_data)
 
-    for idx, opt in enumerate(backup_solvers):
+    for opt in backup_solvers:
         try:
             results = opt.solve(
                 nlp_model,
--- a/pyomo/contrib/pyros/master_problem_methods.py
+++ b/pyomo/contrib/pyros/master_problem_methods.py
@@ -227,9 +227,21 @@ def solve_master_feasibility_problem(model_data, config):
 
 def minimize_dr_vars(model_data, config):
     """
-    Decision rule polishing: For a given optimal design (x) determined in separation,
-    and the optimal value for control vars (z), choose min magnitude decision_rule_var
-    values.
+    Polish the PyROS decision rule determined for the most
+    recently solved master problem by minimizing the collective
+    L1 norm of the vector of all decision rule variables.
+
+    Parameters
+    ----------
+    model_data : MasterProblemData
+        Master problem data.
+    config : ConfigDict
+        PyROS solver settings.
+
+    Returns
+    -------
+    results : SolverResults
+        Subordinate solver results for the polishing problem.
     """
     #config.progress_logger.info("Executing decision rule variable polishing solve.")
     model = model_data.master_model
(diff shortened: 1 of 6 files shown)


pyomo/contrib/pyros/tests/test_grcs.py
```diff
@@ -2126,6 +2129,90 @@ def test_terminate_with_time_limit(self):
...
+                r"Solver \(ipopt\) did not exit normally",
+                ):
+            pyros_solver.solve(
+                m,
+                [m.x1],
+                [],
+                [m.p],
+                box_set,
+                solver,
+                baron,
```
jsiirola, line 2162: With this many arguments, we should consider specifying them all as keyword arguments.

jsiirola (approved):
Some minor comments, but nothing that would prevent merging.

jsiirola:
One note: patch coverage is a bit low.  Most of the uncovered lines relate to handling solver errors.  We should consider adding a test case that exercises (some of) the solver failure cases.

pyomo/contrib/pyros/master_problem_methods.py
```diff
@@ -202,32 +221,51 @@ def solve_master_feasibility_problem(model_data, config):
...
-                master_v.set_value(v.value, skip_validation=True)
-    else:
-        results.solver.termination_condition = tc.error
-        results.solver.message = ("Cannot load a SolverResults object with "
-                                  "bad status: error")
+    # load master feasibility point to master model
+    for v in model.component_data_objects(Var):
+        master_v = model_data.master_model.find_component(v)
+        if master_v is not None:
+            master_v.set_value(v.value, skip_validation=True)
```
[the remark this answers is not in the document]
shermanjasonaf: Understood. What would be the most efficient way to address this? I am currently thinking of adding:
```
var_map = ComponentMap(
    zip(
        model_data.master_model.component_data_objects(pyo.Var, sort=True),
        model.component_data_objects(pyo.Var, sort=True),
    )
)
```
immediately after cloning the master model (in `construct_master_feasibility_problem`---and returning this mapping along with the feasibility model), and then changing the loop here to:
```
for master_var, feas_var in var_map.items():
    master_var.set_value(feas_var.value, skip_validation=True)
```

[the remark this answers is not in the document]
jsiirola: The challenge with what you propose is that it relies on iterating over the variables in the same order on both models.  This *should* be the case, but may not be guaranteed if a user uses an explicitly unordered set.  My favorite way to map variables from a model to the clone of the model is to leverage the logic in `clone()` (nee `deepcopy()`) to do the hard work for me.  In this case, in `construct_master_feasibility_problem`:
```python
varmap_name = unique_component_name(model_data.master_model, 'pyros_var_map')
setattr(model_data.master_model, varmap_name,
        list(model_data.master_model.component_data_objects(Var)))
model = model_data.master_model.clone()
model_data.feasibility_problem_varmap = list(zip(
    getattr(model_data.master_model, varmap_name),
    getattr(model, varmap_name)
))
delattr(model_data.master_model, varmap_name)
delattr(model, varmap_name)
```
then in `solve_master_feasibility_problem`:
```python
for master_var, feas_var in model_data.feasibility_problem_varmap:
    master_var.set_value(feas_var.value, skip_validation=True)
```

pyomo/contrib/pyros/master_problem_methods.py
```diff
@@ -142,7 +144,9 @@ def construct_master_feasibility_problem(model_data, config):
             if con not in dr_eqs])
 
     # retain original constraint exprs (for slack initialization and scaling)
-    pre_slack_con_exprs = ComponentMap([(con, con.body) for con in targets])
+    pre_slack_con_exprs = ComponentMap(
+        [(con, con.body - con.upper) for con in targets]
```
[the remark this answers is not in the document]
shermanjasonaf: See subsequent commits

pyomo/contrib/pyros/tests/test_grcs.py
shermanjasonaf: Agreed

shermanjasonaf:
Can you add @natalieisenberg as a reviewer? I'm still unable to request reviewers.

jsiirola (approved):
Looks good.

pyomo/contrib/pyros/separation_problem_methods.py
```diff
@@ -419,68 +500,139 @@ def initialize_separation(model_data, config):
...
     if any(c.active for c in model_data.separation_model.util.h_x_q_constraints):
-        raise AttributeError("All h(x,q) type constraints must be deactivated in separation.")
+        raise AttributeError(
+            "All h(x,q) type constraints must be deactivated in separation."
+        )
+
+    # check: initial point feasible?
+    for con in sep_model.component_data_objects(Constraint, active=True):
+        lb, val, ub = value(con.lb), value(con.body), value(con.ub)
+        lb_viol = val < lb - 1e-5 if lb is not None else False
```
natalieisenberg, line 510: There's a hard-coded tolerance here of 1e-5...I assume that was agreed upon at some point, but should it be made into a global variable somewhere so we don't loose track of it? Like `initial_point_feas_tol` or something? Also out of curiosity, why 1e-5?

natalieisenberg (commented):
Overall it looks good, I just had one comment/question.

pyomo/contrib/pyros/separation_problem_methods.py
shermanjasonaf: My choice of `1E-5` is somewhat arbitrary; I chose this initially as this is the default absolute constraint violation tolerance for BARON, a solver I use frequently with PyROS. I could make this a global variable (perhaps defined in `util.py`), or opt for `config.robust_feasibility_tolerance` instead.

pyomo/contrib/pyros/master_problem_methods.py
```diff
@@ -337,34 +390,60 @@ def minimize_dr_vars(model_data, config):
...
+            master_sv.set_value(value(polish_sv))
+
+        # update master problem decision rule variables
+        dr_var_zip = zip(
+            blk.util.decision_rule_vars,
+            polishing_model.scenarios[idx].util.decision_rule_vars,
+        )
+        for master_dr, polish_dr in dr_var_zip:
+            for mvar, pvar in zip(master_dr.values(), polish_dr.values()):
+                mvar.set_value(value(pvar))
```
emma58, line 446: Should `skip_validation=True` here? Just noticing since it looks like this is change from before.

emma58 (approved):
One question (that might be silly), but otherwise I think this looks good!

pyomo/contrib/pyros/master_problem_methods.py
shermanjasonaf: Restored `skip_validation=True` in subsequent commits.

pyomo/contrib/pyros/master_problem_methods.py
```diff
@@ -100,6 +102,23 @@ def construct_master_feasibility_problem(model_data, config):
...
     model = model_data.master_model.clone()
+
+    # obtain mapping from master problem to master feasibility
+    # problem variables
+    varmap_name = unique_component_name(
+        model_data.master_model,
+        'pyros_var_map',
+    )
+    setattr(model_data.master_model, varmap_name,
+            list(model_data.master_model.component_data_objects(Var)))
```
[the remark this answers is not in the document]
shermanjasonaf: Noted. Should I make this change here? If I understand correctly, `master_model` has no parent, such that all variables are in the active tree rooted here.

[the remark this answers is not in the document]
emma58: Oh, oops, I may have mislead you. The fact that it has no parent doesn't necessarily mean that all the variables are in the active tree (they could be on deactivated Blocks, for example), but pyros *makes* `master_model`, right? So this is actually in your control at this point, I think. I just brought it up because when you initially collect the variables from the user-defined model, you likely want to use `get_vars_from_components`.

PR #2520, opened 2022-08-31
Adding utility for gathering BigM values after transformation

emma58:
## Summary/Motivation:

Calculating M values can be time-consuming, so it may make sense to run the model once, allowing for automatic calculating, and then pull out the calculated values and specify either all of them or the largest one to the transformation thereafter. This adds a couple utility methods to the bigm transformation to ease that process.

- Adds two methods to the bigm transformation: `get_all_M_values_by_constraint` to get M values from a transformed model in a dictionary with the original constraints as the keys and `get_largest_M_value` to grab the largest (in absolute value) of these

### Legal Acknowledgement

### DIFF
--- a/pyomo/gdp/plugins/bigm.py
+++ b/pyomo/gdp/plugins/bigm.py
@@ -1004,3 +1004,33 @@ class BigM_Transformation(Transformation):
         # fails... (That is, it's a bug in the mapping.)
         lower, upper = transBlock.bigm_src[constraint]
         return (lower[0], upper[0])
+
+    def get_all_M_values_by_constraint(self, model):
+        """Returns a dictionary mapping each constraint to a tuple:
+        (lower_M_value, upper_M_value), where either can be None if the
+        constraint does not have a lower or upper bound (respectively).
+
+        Parameters
+        ----------
+        model: A GDP model that has been transformed with BigM
+        """
+        m_values = {}
+        for disj in model.component_data_objects(
+                Disjunct,
+                active=None,
+                descend_into=(Block, Disjunct)):
+            if disj.transformation_block is not None:
+                transBlock = disj.transformation_block()
+                for cons, (lower, upper) in transBlock.bigm_src.items():
+                    m_values[cons] = self.get_M_value(cons)
+        return m_values
+
+    def get_largest_M_value(self, model):
+        """Returns the largest M value for any constraint on the model.
+
+        Parameters
+        ----------
+        model: A GDP model that has been transformed with BigM
+        """
+        return max(max(abs(m) for m in m_values if m is not None) for m_values
+                   in self.get_all_M_values_by_constraint(model).values())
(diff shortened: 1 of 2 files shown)


pyomo/gdp/plugins/bigm.py
```diff
@@ -1004,3 +1004,33 @@ def get_M_value(self, constraint):
...
+        """
+        m_values = {}
+        for disj in model.component_data_objects(
+                Disjunct,
+                active=None,
+                descend_into=(Block, Disjunct)):
+            if disj.transformation_block is not None:
+                transBlock = disj.transformation_block()
+                for cons, (lower, upper) in transBlock.bigm_src.items():
+                    m_values[cons] = self.get_M_value(cons)
```
jsiirola, line 1025: This assumes that all Disjuncts were transformed by BigM.  I think you need a check to make sure bigm_src is there...

pyomo/gdp/plugins/bigm.py
```diff
@@ -1004,3 +1004,33 @@ def get_M_value(self, constraint):
...
+        return m_values
+
+    def get_largest_M_value(self, model):
+        """Returns the largest M value for any constraint on the model.
```
jsiirola, line 1029: Why does this return a scalar and not a 2-tuple of `(biggest_lower_M, biggest_upper_M)`?

jsiirola (changes requested):
This is a great utility.  Just two questions...

pyomo/gdp/plugins/bigm.py
emma58: Good point.

emma58: Oh, wait, just kidding, if it's not transformed then `disj.transformation_block` will be `None`. If it points to something, then the disjunct is transformed, and it's actually a bug in the transformation if it doesn't have the `bigm_src` dictionary.

pyomo/gdp/plugins/bigm.py
emma58: Do you think that is really safe? It's useful if you're not going to change anything on the model, but if for some reason you rewrote some things to make an equivalent model, what's lower and what's upper could change. I just worry because that's not really a mathematical distinction so much as a how-the-constraint-expressions-are-represented distinction.

pyomo/gdp/plugins/bigm.py
jsiirola: So hull adds `bigm_src`?

jsiirola (approved):

pyomo/gdp/plugins/bigm.py
jsiirola: Minor nit: I think it is easier / faster to just do:
```python
for cons in transBlock.bigm_src:
    m_values[cons] = self.get_M_value(cons)
```

pyomo/gdp/plugins/bigm.py
jsiirola: Yeah - good point.  Plus, if the user wants to calculate the tuple, it is not difficult to do that using `get_all_M_values_by_constraint()`

PR #2530, opened 2022-09-09
bugfix for gurobi work limit

tharwan:
## Fixes # 

If the WorkLimit is set in gurobi and the optimisation is stoped because the limit is reached pyomo does not detect the feasible solution.

## Summary/Motivation:

Gurobi 9.5 introduced a deterministic [work limit](https://www.gurobi.com/documentation/9.5/refman/worklimit.html) as alternative to a time limit. Pyomo does not detect this correctly if the optimisation is stopped because of WORK_LIMIT.

Work limit would be very useful for us and we would love to see a quick bugfix release including this fix.

- handle a stop due to work limit the same as time limit

### Legal Acknowledgement

### DIFF
--- a/pyomo/solvers/plugins/solvers/GUROBI_RUN.py
+++ b/pyomo/solvers/plugins/solvers/GUROBI_RUN.py
@@ -167,6 +167,11 @@ def gurobi_run(model_file, warmstart_file, soln_file, mipgap, options, suffixes)
         message = 'Optimization terminated because the time expended exceeded the value specified in the TimeLimit parameter.'
         term_cond = 'maxTimeLimit'
         solution_status = 'stoppedByLimit'
+    elif (solver_status == GRB.WORK_LIMIT):
+        status = 'aborted'
+        message = 'Optimization terminated because the work expended exceeded the value specified in the WorkLimit parameter.'
+        term_cond = 'maxTimeLimit'
+        solution_status = 'stoppedByLimit'
     elif (solver_status == GRB.SOLUTION_LIMIT):
         status = 'aborted'
         message = 'Optimization terminated because the number of solutions found reached the value specified in the SolutionLimit parameter.'
--- a/pyomo/solvers/plugins/solvers/GUROBI_RUN.py
+++ b/pyomo/solvers/plugins/solvers/GUROBI_RUN.py
@@ -167,7 +167,7 @@ def gurobi_run(model_file, warmstart_file, soln_file, mipgap, options, suffixes)
         message = 'Optimization terminated because the time expended exceeded the value specified in the TimeLimit parameter.'
         term_cond = 'maxTimeLimit'
         solution_status = 'stoppedByLimit'
-    elif (solver_status == GRB.WORK_LIMIT):
+    elif hasattr(GRB, "WORK_LIMIT") and (solver_status == GRB.WORK_LIMIT):
         status = 'aborted'
         message = 'Optimization terminated because the work expended exceeded the value specified in the WorkLimit parameter.'
         term_cond = 'maxTimeLimit'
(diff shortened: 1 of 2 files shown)


pyomo/solvers/plugins/solvers/GUROBI_RUN.py
```diff
@@ -167,6 +167,11 @@ def gurobi_run(model_file, warmstart_file, soln_file, mipgap, options, suffixes)
         message = 'Optimization terminated because the time expended exceeded the value specified in the TimeLimit parameter.'
         term_cond = 'maxTimeLimit'
         solution_status = 'stoppedByLimit'
+    elif hasattr(GRB, "WORK_LIMIT") and (solver_status == GRB.WORK_LIMIT):
+        status = 'aborted'
+        message = 'Optimization terminated because the work expended exceeded the value specified in the WorkLimit parameter.'
+        term_cond = 'maxTimeLimit'
```
michaelbynum, line 173: Would `'maxIterations'` or `'maxEvaluations'` be more appropriate than `'maxTimeLimit'`?

michaelbynum (changes requested):
@tharwan, thanks for this contribution. Can you please add a test for this functionality?

tharwan:
> @tharwan, thanks for this contribution. Can you please add a test for this functionality?

I can try. Any hints where to start?

pyomo/solvers/plugins/solvers/GUROBI_RUN.py
tharwan: I thought about this but in the docs it is an alternative to a time limit. 🤷‍♂️

michaelbynum: Good enough for me.

tharwan:
@michaelbynum I really need some help where to even find the tests that are responsible for this part.

michaelbynum:
I would recommend putting them in pyomo/solvers/tests/checks. I don't see any gurobi-specific tests in there right now, so I would create a new file (like we have for gurobi persistent).

tharwan:
I don't think this is very helpful to be honest, but here you go.

tharwan:
@michaelbynum any suggestions why the tests failed 3.7? Can we rerun the pipeline?

pyomo/solvers/tests/checks/test_gurobi.py
```diff
@@ -0,0 +1,43 @@
+import unittest
```
mrmundt, line 1: Pyomo-specific suggestion: we have an extended implementation of `unittest`, so we prefer:
```
import pyomo.common.unittest as unittest
```

It still has all of the standard `unittest` libraries - just adds some extra functionality for some of our testing choices.

pyomo/solvers/tests/checks/test_gurobi.py
```diff
@@ -0,0 +1,43 @@
...
+try:
+    from pyomo.solvers.plugins.solvers.GUROBI_RUN import gurobi_run
+    from gurobipy import GRB
+    gurobipy_available = True
+except:
+    gurobipy_available = False
+
+class GurobiTest(unittest.TestCase):
+
+    @unittest.skipIf(not gurobipy_available,
```
mrmundt, line 13: Suggestion: move this first `skipIf` to right before the class declaration. That should fix the failure on the `slim` build.

tharwan:
this is quite frustrating.

mrmundt:
@tharwan  - I totally understand. I have some time in a couple hours to chase down some of your bugs.

mrmundt:
@tharwan - Just pending this very last test (fingers crossed)!

mrmundt (approved):

PR #2535, opened 2022-09-14
Improve handling of restricted words on Blocks

jsiirola:
## Summary/Motivation:
This PR improves handling of reserved words on Blocks:
- prevent deleting or replacing components that are declared as reserved words
- add a (private) context manager to simplify declaring "reserved" components
- improve preservation of declaration order (remove `sorted()` that is unnecessary in Python 3.7+)
- update and expand testing

- (see above)

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/block.py
+++ b/pyomo/core/base/block.py
@@ -1062,7 +1062,7 @@ class _BlockData(ActiveComponentData):
         if not val.valid_model_component():
             raise RuntimeError(
                 "Cannot add '%s' as a component to a block" % str(type(val)))
-        if name in self._Block_reserved_words:# and hasattr(self, name):
+        if name in self._Block_reserved_words:
             raise ValueError("Attempting to declare a block component using "
                              "the name of a reserved attribute:\n\t%s"
                              % (name,))
--- a/pyomo/core/base/block.py
+++ b/pyomo/core/base/block.py
@@ -1046,6 +1046,8 @@ class _BlockData(ActiveComponentData):
 
     @contextmanager
     def _declare_reserved_components(self):
+        # Temporarily mask the class reserved words like with a local
+        # instance attribute
         self._Block_reserved_words = ()
         yield
         del self._Block_reserved_words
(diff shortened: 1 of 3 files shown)


pyomo/core/tests/unit/test_block.py
```diff
@@ -776,7 +778,7 @@ def __init__(self, *args, **kwds):
 
         b.clear()
         b.transfer_attributes_from(c)
-        self.assertEqual(list(b.component_map()), ['y','x','z'])
+        self.assertEqual(list(b.component_map()), ['y','z','x'])
```
emma58, line 781: Was this a bug before? Why did it change?

pyomo/core/base/block.py
```diff
@@ -1028,6 +1044,12 @@ def find_component(self, label_or_component):
             cuid = ComponentUID(label_or_component)
         return cuid.find_component_on(self)
 
+    @contextmanager
+    def _declare_reserved_components(self):
+        self._Block_reserved_words = ()
+        yield
+        del self._Block_reserved_words
```
emma58, line 1051: Why doesn't this restore `_Block_reserved_words` to what it was before?

emma58 (approved):
Just asking questions to make sure I understand... But I think this looks good! (And I'm very excited about it!)

pyomo/core/tests/unit/test_block.py
jsiirola: We used to sort the incoming dict to ensure determinism.  Now that we only support Python 3.7+, we can remove the sort and preserve decl order.  So not a bug per se, but this is a "better" approach (the user can actually control the decl order if they know what they are doing).

pyomo/core/base/block.py
jsiirola: It doesn't need to.  This is masking the class attribute with an instance attribute.  "Restoring" the class attribute just boils down to deleting the local instance attribute.

emma58: Ohhh, that makes sense.

PR #2539, opened 2022-09-19
Resolve mapping gradient/hessian for external function with string args

### LINKED ISSUE #2528: ExternalFunction evaluate_fgh doesn't work correctly with string arguments
## Summary

When you have an ExternalFunction that expects string arguments, the evaluate_fgh method includes the string arguments when counting how big the gradient and hessian arrays are. This means that you get the gradient and hessian but they are larger than they should be and padded with random junk from memory at the end.

Run the example below (requires IDAES).  If you have an example function that uses string args that will work too. 

```
# example.py
import pyomo.environ as pyo
from idaes.models.properties.general_helmholtz.helmholtz_functions import (
    add_helmholtz_external_functions,
    HelmholtzParameterBlock,
    HelmholtzThermoExpressions,
    AmountBasis,
)

m = pyo.ConcreteModel()
m.param_block = HelmholtzParameterBlock(
    pure_component="r134a", amount_basis=AmountBasis.MASS
)

te = HelmholtzThermoExpressions(m, m.param_block)

add_helmholtz_external_functions(m, "tau_func")

y, g, h = m.tau_func.evaluate_fgh(args=("r134a", 529.47, 100))
print(f"T = {374.18/y}")
print(g)
print(h)
```

### Error Message

There is no error message. 

Pyomo version: Pyomo 6.4.2 (devel {typing-overload}) (CPython 3.10.4 on Darwin 21.6.0)

jsiirola:
## Summary/Motivation:
The `AmplExternalFunction` interface did not correctly map the gradient / Hessian for external functions with string arguments.  This PR resolves that mapping, inserting 0's in the corresponding locations.

Question for reviewers (and @eslickj): is it correct to insert `0`s, or should this put `nan`s in locations corresponding to string arguments?

- correct mapping of gradient / Hessian information for external functions with string arguments
- update test to exercise this behavior.

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/ampl_function_demo/tests/test_ampl_function_demo.py
+++ b/pyomo/contrib/ampl_function_demo/tests/test_ampl_function_demo.py
@@ -14,6 +14,7 @@ import pyomo.environ as pyo
 import pyomo.common.unittest as unittest
 from pyomo.common.fileutils import find_library
 from pyomo.opt import check_available_solvers
+from pyomo.core.base.external import nan
 
 flib = find_library("asl_external_demo")
 is_pypy = platform.python_implementation().lower().startswith('pypy')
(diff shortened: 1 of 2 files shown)


eslickj:
@jsiirola the external functions just return the first and second derivatives for the floating point arguments and nothing for the strings (like the string args don't exist).  It doesn't really matter to me how Pyomo handles it internally as long as I know what to expect.  I don't have first hand knowledge of this but I've been told that the simpy interface for differentiation works with external functions using ```evaluate_fgh```.  As long as that works (if it is supposed to) and I know what to expect, I think it's fine.

I may have a slight preference for nan, so if you try to use that derivative you know you are messing up.  Not sure what trouble that could cause though. If 0 makes thing run more smoothly that good with me.

mrmundt (approved):

jsiirola:
I implemented this as `0`s, because that is the "simple default thing to do".  But the more I think about it, the more I believe that `nan` is the correct answer.  The output *could* change if the string argument changed - so 0 is not really correct - but since you also can't finite difference a string, I think `nan` is more mathematically correct.

mrmundt:
@jsiirola  - Do you want to switch it to `nan`s or keep it as `0`? That's basically the only thing holding up this PR.

jsiirola:
I am leaning to changing to `nan`, but was hoping that @emma58, @michaelbynum, @blnicho and/or you would weigh in.

mrmundt:
I would also be inclined to choose `nan` out of the two.

emma58:
I agree!

michaelbynum:
Agreed

PR #2550, opened 2022-09-28
Implementing tighter logical constraints in bigm and hull transformations of nested GDPs

emma58:
## Summary/Motivation:

In a nested GDP, the tightest formulation is to set the sum of the inner disjunction's binary indicator variables to the outer disjunct's binary indicator variable. This PR both makes this change in the bigm and hull transformations, as well as finally taking advantage of the preprocessing we do in order to not recurse into already-transformed components.

- Changing form of algebraic constraints for nested Disjunctions (as well as the "re-aggregation" constraints in hull linking original variables to disaggregated variables)
- No longer moving transformation blocks for inner Disjuncts to their parent Disjunct's transformation block: Instead we declare it at the "root" disjunct to begin with.
- Changing hull to transform from the root to the leaves of the "GDP tree" rather than vice versa (what bigm does). This is because it is more efficient if we know the parent Disjunct's disaggregated variables when we transform inner Disjunctions--else we'll have to transform constraints we created later.
- NOTE: The above change changes the order in which components are transformed in hull. This changed both the baseline tests in GDP and the FME tests that rely on the hull transformation.
- (In general this PR is revealing of some problems with the GDP tests given how many had to change--I'll get to that in another PR)

### Legal Acknowledgement


emma58:
The Jenkins failures are I think Gurobi license related, not related to the changes... So this is ready for review @jsiirola.

pyomo/gdp/plugins/bigm.py
```diff
@@ -152,14 +151,15 @@ def __init__(self):
             Set:         False,
             SetOf:       False,
             RangeSet:    False,
-            Disjunction: self._warn_for_active_disjunction,
+            Disjunction: False,
```
jsiirola, line 154: Do we no longer need to warn for active disjunctions?  I suppose not, as the preprocessing would have identified all Disjunctions on the active block/disjunct tree.  A comment to this effect might be warranted.

pyomo/gdp/util.py
```diff
@@ -124,7 +162,7 @@ def _visit_vertex(self, u, leaf_to_root):
 
     def _topological_sort(self):
         # this is reverse of the list we should return (but happens to be what
-        # we want for hull and bigm)
+        # we want for bigm)
```
jsiirola, line 165: Should the really be called `_topological_sort`?  It is confusing that `topological_sort` returns the reverse of this generator.  What about renaming it to `_reverse_topological_iterator`

jsiirola (approved):
Some minor questions about naming / documentation - but nothing that should prevent moving forward with this.

jsiirola:
One note: merging this will mean that you cannot expect to solve individual relaxed disjuncts after either the BigM or Hull transformation is applied to the entire model, correct?

pyomo/gdp/plugins/bigm.py
```diff
@@ -291,102 +291,59 @@ def _filter_inactive(targets):
                 logger.warning(warning_msg)
 
     def _add_transformation_block(self, instance):
+        if instance in self._transformation_blocks:
+            return self._transformation_blocks[instance]
+
```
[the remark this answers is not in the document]
emma58: Good point, that's a very dated argument name. Maybe `to_block` would be better? Or we could do `parent_block`.

pyomo/gdp/plugins/bigm.py
emma58: Yep, exactly, it's unreachable code. In bigm because of going from leaf to root. In hull because we actually handle them as we encounter them. I'll add a comment.

pyomo/gdp/util.py
emma58: Hehehe, good point. I'm all for not naming a method as the opposite of what it does... :P

emma58:
@jsiirola, you could solve `disjunct.transformation_block()` (if you added an objective) for not-nested models. But you are right that for nested models, the nested stuff is not a descendant of that block. But I don't think that it was before either?

Do you think that's something we should support? I think we'd have to transform from root to leaf in bigm too, if we were going to support that. Because we would need to be putting inner-Disjunct transformation blocks on the transformation block of their parents.

PR #2557, opened 2022-10-05
Support MOSEK 10 affine conic constraints

### LINKED ISSUE #2442: Support for MOSEK v10
## Summary

Update the Pyomo interface to support the _affine conic constraints_ and the new _domain_ types introduced in MOSEK v10.

### Rationale

[MOSEK 10 is a major version update](https://docs.mosek.com/10.0/releasenotes/changes.html) with many changes. The changes that concern this request are:

- _Affine conic constraints_ are introduced in the Optimizer API. Users can construct conic constraints like `Fx + g \in K`, where `F` is a coefficient matrix, `g` is a vector of constants, and `K` is a conic **_domain_**. It is therefore no longer necessary to introduce auxiliary variables to include expressions in a nonlinear cone. 
- The old _cone_ types will be deprecated in favor of the new affine conic constraints and the domains therein.
- The new _domains_ include the primal/dual _geometric mean conic domains_ and the _vectorized PSD cone domain_, in addition to the cones that were previously supported in the cones. The power conic domains can also handle long left-hand side terms.

### Description

I am working on passing the conic constraints from Pyomo kernel interface to MOSEK as affine conic constraints (instead of the old cone types) if MOSEK v10 is detected.

But, I would also like to suggest the following:

- Conic constraints in the Kernel interface when constructed `as_domains` should be passed to MOSEK as an ACC, without introducing any auxiliary variables. In fact, I believe it would be more intuitive if the `as_domain` option was the only way to construct conic constraints for MOSEK v10 onwards.
- Support for geometric mean cone and the vectorized PSD cones. (Support for PSD cones will let users to construct linear matrix inequalities in Pyomo)
- The conic constraints for power cones should start supporting long left-hand side terms (i.e. passing a tuple r, instead of two scalars r1 and r2).

If agreed upon, I can work on implementing these suggestions. I would welcome suggestions, especially on how to go about keeping backward compatibility (if at all) in the first suggestion mentioned above.

Utkarsh-Detha:
## Motivation:

This PR adds support for the newly introduced affine conic constraints in MOSEK 10, and also provides new conic domains (corresponding to the new domains in MOSEK 10).

- Primal/Dual geometric mean cone and the vectorized PSD cone are introduced. This means that Pyomo-MOSEK can now be used to solve SDPs.
- If MOSEK 10 is detected, the conic constraints will be passed down to MOSEK as _affine conic constraints_ (as opposed to the old cones).
- New examples (in `examples/kernel/mosek`) maximum_volume_cuboid.py and semidefinite.py  are provided to illustrate the new conic constraints.

NOTE: Ideally, when conic constraints are constructed using the `as_domain()` method, they can be passed down to MOSEK 10 as-is (no need to introduce auxiliary variables and then set equality constraints with expressions). But, I was not sure if this would be a good idea (as pyomo-MOSEK stands now) because it could break backward compatibility and I didn't wanna make changes of this type in the `conic.py` file. Any feedback/guidance on this would be greatly appreciated.

### Legal Acknowledgement


blnicho:
@Utkarsh-Detha we're working on renewing our MOSEK testing license and upgrading to 10.

Utkarsh-Detha:
@blnicho thanks for the update! Let me know if/when any assistance is needed from my side.

jsiirola (approved):

blnicho (approved):

Issue #2562, opened 2022-10-10
Could not write legal LP file


Betristor:
## Summary
Pyomo could not solve the linear model and report bug as follows:
```
  File "pyomo/repn/plugins/cpxlp.pyx", line 158, in pyomo.repn.plugins.cpxlp.ProblemWriter_cpxlp.__call__
  File "pyomo/repn/plugins/cpxlp.pyx", line 159, in pyomo.repn.plugins.cpxlp.ProblemWriter_cpxlp.__call__
  File "pyomo/repn/plugins/cpxlp.pyx", line 160, in pyomo.repn.plugins.cpxlp.ProblemWriter_cpxlp.__call__
  File "pyomo/repn/plugins/cpxlp.pyx", line 532, in pyomo.repn.plugins.cpxlp.ProblemWriter_cpxlp._print_model_LP
RuntimeError: Cannot write legal LP file.  Objective 'objective' has nonlinear terms that are not quadratic.
```

Main problem may be related to cpxlp.pyx file.

### Error Message

```console
Traceback (most recent call last):
  File "/Users/xxx/Documents/GitHub/Adaptive-clustering-for-ESOM/src/main.py", line 90, in <module>
    m = solve(m, config)
  File "/Users/xxx/Documents/GitHub/Adaptive-clustering-for-ESOM/src/solve_model.py", line 13, in solve
    opt.solve(m)
  File "/Users/xxx/miniforge3/envs/eso/lib/python3.8/site-packages/pyomo/opt/base/solvers.py", line 570, in solve
    self._presolve(*args, **kwds)
  File "/Users/xxx/miniforge3/envs/eso/lib/python3.8/site-packages/pyomo/solvers/plugins/solvers/GUROBI.py", line 234, in _presolve
    ILMLicensedSystemCallSolver._presolve(self, *args, **kwds)
  File "/Users/xxx/miniforge3/envs/eso/lib/python3.8/site-packages/pyomo/opt/solver/shellcmd.py", line 209, in _presolve
    OptSolver._presolve(self, *args, **kwds)
  File "/Users/xxx/miniforge3/envs/eso/lib/python3.8/site-packages/pyomo/opt/base/solvers.py", line 667, in _presolve
    self._convert_problem(args,
  File "/Users/xxx/miniforge3/envs/eso/lib/python3.8/site-packages/pyomo/opt/base/solvers.py", line 718, in _convert_problem
    return convert_problem(args,
[...]

PR #2567, opened 2022-10-16
Expression.set_value: raise exception for non-numeric expressions

### LINKED ISSUE #378: Using relational Expression in Objective
The following model is sent to a solver without any warning or error:

    from pyomo.environ import *
    m = ConcreteModel()
    m.v = Var()
    m.e = Expression(expr=m.v**2 + 2 >= 10)
    m.obj = Objective(expr=m.e)

While the user should be able to construct this model, the writers should be checking `is_relational` on the Expression object before the model is sent to a solver.

jsiirola:
## Summary/Motivation:
The root cause of #2562 was that the user was setting the value of an `Expression` component (in this case, the `Objective`) to a relational expression and not a numeric expression.  Now that Pyomo is supporting multiple expression "systems", we should be more careful about what we accept for `Expression`/ `Objective` values.

This PR raises an exception when setting an `Expression`/ `Objective` to anything other than a valid numeric expression.

- `ExpressionData.set_value()` raise exception for non-numeric expressions 
- Update test to exercise exception

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/expression.py
+++ b/pyomo/core/base/expression.py
@@ -191,7 +191,7 @@ class _GeneralExpressionDataImpl(_ExpressionData):
             raise ValueError(
                 f"Cannot assign {expr.__class__.__name__} to "
                 f"'{self.name}': {self.__class__.__name__} components only "
-                "allows numeric expression types.")
+                "allow numeric expression types.")
         # In-place operators will leave self as an argument.  We need to
         # replace that with the current expression in order to avoid
         # loops in the expression tree.
--- a/pyomo/core/base/expression.py
+++ b/pyomo/core/base/expression.py
@@ -187,6 +187,11 @@ class _GeneralExpressionDataImpl(_ExpressionData):
             self._expr = None
             return
         expr = as_numeric(expr)
+        if not expr.is_numeric_type():
+            raise ValueError(
+                f"Cannot assign {expr.__class__.__name__} to "
+                f"'{self.name}': {self.__class__.__name__} components only "
+                "allows numeric expression types.")
         # In-place operators will leave self as an argument.  We need to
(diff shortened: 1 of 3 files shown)


pyomo/core/base/expression.py
```diff
@@ -187,6 +187,11 @@ def set_value(self, expr):
             self._expr = None
             return
         expr = as_numeric(expr)
+        if not expr.is_numeric_type():
+            raise ValueError(
+                f"Cannot assign {expr.__class__.__name__} to "
+                f"'{self.name}': {self.__class__.__name__} components only "
+                "allows numeric expression types.")
```
emma58, line 194: Typo: allows -> allow

pyomo/core/tests/unit/test_pickle.py
```diff
@@ -227,7 +227,7 @@ def test_pickle_concrete_model_constant_objective(self):
     def test_pickle_concrete_model_objective(self):
         model = ConcreteModel()
         model.x = Var()
-        model.A = Objective(expr=model.x <= 0)
+        model.A = Objective(expr=model.x**2)
```
emma58, line 230: Was this a typo? We actually tested you could set the objective to a relational expression?!

emma58 (approved):
There's a typo in the error message, but otherwise this looks good. :)

pyomo/core/tests/unit/test_pickle.py
jsiirola: Alas, yes.  That test has been there for a long, long time.  I'm pretty sure it was a copy-paste-and-forget-to-update error when the test suite was originally thrown together.

blnicho (approved):
This also fixes #378

PR #2569, opened 2022-10-20
Be more targeted when retrying failed downloads

jsiirola:
## Summary/Motivation:
This provides a more target implementation for retrying downloads that fail with a '403: Forbidden' error.  It also corrects a bug in the previous implementation that did not pass the SSL context when retrying the download.

- Only retry the download when a HTTP:403 error is encountered
- Use the SSL context when retrying the download

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/download.py
+++ b/pyomo/common/download.py
@@ -25,6 +25,7 @@ import pyomo.common
 from pyomo.common.dependencies import attempt_import
 
 request = attempt_import('urllib.request')[0]
+urllib_error = attempt_import('urllib.error')[0]
 ssl = attempt_import('ssl')[0]
 zipfile = attempt_import('zipfile')[0]
 gzip = attempt_import('gzip')[0]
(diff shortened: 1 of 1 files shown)


pyomo/common/download.py
```diff
@@ -293,22 +294,27 @@ def destination(self):
...
-            if self.insecure:
-                ctx.check_hostname = False
-                ctx.verify_mode = ssl.CERT_NONE
             fetch = request.urlopen(url, context=ctx)
-        except:
+        except urllib_error.HTTPError as e:
+            if e.code != 403: # Forbidden
+                raise
+            fetch = None
+        if fetch is None:
```
mrmundt, line 309: Can you explain the logic behind removing this from the `try-except` block? Wouldn't it be less messy to just keep it all within the `except`?

jsiirola: If we keep it in the try-except, then the exception will be raised with the less-than helpful "while processing this exception, another exception occurred."  By moving it out, the second exception is "detangled" from the first.

mrmundt: ... Fine, but I want it noted that I don't like it from a readability perspective.

PR #2570, opened 2022-10-20
Adding constraint programming package to contrib

emma58:
## Summary/Motivation:

Adds `contrib.cp`, which has new components to accommodate the scheduling-optimization-specific functionality in IBM's CP Optimizer solver along with an interface to said solver (through their Python `docplex` package). 

- Adds scheduling-specific components: `IntervalVar`s that have start times, end times, lengths, and a Boolean representing whether or not they are scheduled. Also `Pulse` and `Step` for tracking resources as functions of `IntervalVars`.
- Adds solver interface to CP Optimizer
- Adds "indirection" to Pyomo Vars, Params, and BooleanVars, so that it is now a valid expression to use an integer-valued Var as an index of any of these.
- Adds an initializer for things like variable bounds
- Adds a not-equal expression node

### Legal Acknowledgement


pyomo/core/expr/visitor.py
```diff
@@ -1258,6 +1258,9 @@ def evaluate_expression(exp, exception=True, constant=False):
     finally:
         if clear_active:
            evaluate_expression.visitor_active = False
+    if ans.__class__ not in native_types and ans.is_numeric_type() is True:
```
emma58, line 1261: @jsiirola, the failing tests are because of this line. But I'm not sure what to do about it?

jsiirola (commented):
Initial comments...

pyomo/contrib/cp/repn/docplex_writer.py
```diff
@@ -0,0 +1,1094 @@
...
+        ans = []
+        for o in objects:
+            if attr[1] == 'start_time':
+                ans.append(cp.start_of(o))
+            elif attr[1] == 'end_time':
+                ans.append(cp.end_of(o))
+            elif attr[1] == 'length':
+                ans.append(cp.length_of(o))
+            elif attr[1] == 'is_present':
+                ans.append(cp.presence_of(o))
```
jsiirola, line 229: Can this make use of a `dict` instead of an if tree?  Something like
```python
getattr_dispatcher = {
    'start_time': cp.start_of,
    'end_time': cp.end_of,
    'length': cp.length_of,
    'is_present': cp.presence_of,
}
```
then
```python
ans = list(map(getattr_dispatcher[attr[1]], obj[1][0]))
```

pyomo/contrib/cp/repn/docplex_writer.py
```diff
@@ -0,0 +1,1094 @@
...
+        return arg[1]
+    elif arg[0] is _ELEMENT_CONSTRAINT:
+        # docplex doesn't bother to check if 'element' expressions are integer-
+        # or boolean-valued: they just complain if you use them in a boolean
+        # context. So if we are about to use one that way, we set it equivalent
+        # to True so that it will be boolean-valued according to docplex's
+        # idiosyncracies.
+        return arg[1] == True
+    elif arg[0] is _BEFORE:
+        # We're using a start-before-start or it's ilk in a boolean-valued
```
jsiirola, line 531: typo: `its`

pyomo/contrib/cp/repn/docplex_writer.py
```diff
@@ -0,0 +1,1094 @@
...
+
+    def __exit__(self, t, v, traceback):
+        pass
+
+    def available(self, exception_flag=True):
+        return Executable('cpoptimizer').available() and docplex_available
+
+    def license_is_valid(self):
+        # TODO: not sure how to check the CPLEX license?
+        True
```
jsiirola, line 1002: Shouldn't this `return self.available()`?

pyomo/contrib/cp/repn/docplex_writer.py
```diff
@@ -0,0 +1,1094 @@
...
+                obj_expr = visitor.walk_expression((obj.expr, obj, 0))
+                if obj.sense is minimize:
+                    cpx_model.add(cp.minimize(obj_expr[1]))
+                else:
+                    cpx_model.add(cp.maximize(obj_expr[1]))
+
+        # No objective is fine too, this is CP afterall...
+
+        # Write algebraic constraints
+        for block in component_map[Constraint]:
```
jsiirola, line 890: I think we should move away from the `component_map` approach -- at least for Var/Constraint-like things: if we get all the constraints from a single `component_data_objects(Constraint)` call, then the function will remove duplicates (e.g., due to References).  If we have multiple calls to `component_data_objects`, then deduplication will only happen within each call and not between them.

pyomo/contrib/cp/scheduling_expr/precedence_expressions.py
```diff
@@ -0,0 +1,93 @@
...
+#  Under the terms of Contract DE-NA0003525 with National Technology and
+#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain
+#  rights in this software.
+#  This software is distributed under the 3-clause BSD License.
+#  ___________________________________________________________________________
+
+from pyomo.core.expr.base import ExpressionBase
+from pyomo.core.expr.boolean_value import BooleanValue
+
+class BeforeExpression(ExpressionBase, BooleanValue):
```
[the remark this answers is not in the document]
emma58: I like the base class idea. And yes, I think it should inherit from `BooleanExpression`.

pyomo/contrib/cp/scheduling_expr/step_function_expressions.py
```diff
@@ -0,0 +1,282 @@
...
+            return CumulativeFunction(_self._args_, nargs=len(_self._args_))
+        else:
+            # we have to clone the list of _args_
+            return CumulativeFunction(_self.args + [NegatedStepFunction(a) for a
+                                                    in _other.args])
+    else:
+        raise TypeError("Cannot subtract object of class %s from object of "
+                        "class %s" % (_other.__class__, _self.__class__))
+
+class StepFunction(PyomoObject):
```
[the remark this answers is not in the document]
emma58: I was hiding it from the expression system so that it would be treated as a leaf node, but I can do that explicitly in the writer. I think it should inherit from `ExpressionBase`

pyomo/contrib/cp/scheduling_expr/step_function_expressions.py
```diff
@@ -0,0 +1,282 @@
...
+    Args:
+        interval_var (IntervalVar): the interval variable over which the
+            pulse function is non-zero
+        height (int): The value of the pulse function during the time
+            interval_var is scheduled
+    """
+
+    __slots__ = ('_args_')
+    def __init__(self, interval_var, height):
+        self._args_ = [self, interval_var, height]
```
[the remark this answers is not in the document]
emma58: Because I was being waaay too much of a cheater and having these pretend to be `CumulativeFunction`s when I was adding and subtracting. I fixed that...

pyomo/contrib/cp/repn/docplex_writer.py
```diff
@@ -0,0 +1,1094 @@
...
+            raise ValueError(
+                "Variable indirection '%s' permits an index '%s' "
+                "that is not a valid key. In CP Optimizer, this is a "
+                "structural infeasibility." % (node, idx))
+            # NOTE: If we thought it was the right thing to do in the future, we
+            # could fill in with a bogus variable and add a constraint
+            # disallowing it from being selected
+    try:
+        return (_ELEMENT_CONSTRAINT, cp.element(elements, expr))
+    except:
```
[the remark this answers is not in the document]
emma58: Yep, but for reasons I don't understand it's an `AssertionError`...

pyomo/contrib/cp/repn/docplex_writer.py
emma58: Yes, but is there a not-terribly-slow way to catch the `KeyError` and give a useful error message, if I wrap the `ans = ` line in a try-except?

emma58: I guess I know `attr[1]` but not the specific object, right? Maybe that's enough for now...

jsiirola: The dispatcher is independent of the object (and in the code above, the gispatcher is looked up *once* when preparing the arguments to `map()`).  You can always make that more explicit:
```python
try:
    fcn = getattr_dispatcher[attr[1]]
except KeyError:
   # something useful

ans = list(map(fcn, obj[1][0]))
```

There is a slight (25%) performance hit for wrapping something like a dict lookup in a `try-except` ... if you never expect to hit the exception, you could use a defaultdict (but then the message you would get back may not have the context - in particular, not the value of `attr[1]`)

pyomo/contrib/cp/repn/docplex_writer.py
jsiirola: Good point.  It looks like the community edition is limited to search spaces of 2^1000.  I couldn't find anything in the docplex api to directly query the version (CE vs Academic/Full).  We could just try solving a toy model (and cache the result):
```python
if CPOptimizerSolver._unrestricted_license is None:
    # Note: 140*log_2(140) == 998.1 fits in CE, 
    #     141*log_2(141) == 1006.7 does not
    x = cp.integer_var_list(141, 1, 141, "X")
    m = cp.CpoModel()
    m.add(cp.all_diff(x))
    try:
        m.solve()
        CPOptimizerSolver._unrestricted_license = True
    except docplex.cp.solver.solver.CpoSolverException:
        CPOptimizerSolver._unrestricted_license = False
return CPOptimizerSolver._unrestricted_license
```

emma58:
@jsiirola, if the tests pass, this is ready for you to look at again.

pyomo/contrib/cp/interval_var.py
```diff
@@ -0,0 +1,202 @@
...
+            self.start_time = IntervalVarStartTime()
+            self.end_time = IntervalVarEndTime()
+            self.length = IntervalVarLength()
+
+    @property
+    def optional(self):
+        # We only store this information in one place, but it's kind of annoying
+        # to have to check if the BooleanVar is fixed, so this way you can ask
+        # the IntervalVar directly.
+        return not self.is_present.fixed
```
jsiirola, line 107: Is optional really `False` if `is_present.fix(False)`?

pyomo/contrib/cp/interval_var.py
```diff
@@ -0,0 +1,202 @@
...
+        name (str, optional): Name for this component.
+        doc (str, optional): Text describing this component.
+    """
+
+    _ComponentDataClass = IntervalVarData
+
+    def __new__(cls, *args, **kwds):
+        if cls != IntervalVar:
+            return super(IntervalVar, cls).__new__(cls)
+        if args == ():
```
jsiirola, line 143: This should check `if not args or (args[0] is UnindexedComponent_set and len(args)==1):`

pyomo/contrib/cp/repn/docplex_writer.py
```diff
@@ -0,0 +1,1107 @@
...
+    StepAtStart: _handle_step_at_start_node,
+    StepAtEnd: _handle_step_at_end_node,
+    Pulse: _handle_pulse_node,
+}
+
+def _handle_negated_step_function_node(visitor, node):
+    return _step_function_handles[node.args[0].__class__](visitor, node.args[0])
+
+_step_function_handles[
+    NegatedStepFunction] = _handle_negated_step_function_node,
```
jsiirola, line 489: This can all be declared at once (and after `_handle_negated_step_function_node` and `_handle_cumulative_function`)

pyomo/contrib/cp/repn/docplex_writer.py
```diff
@@ -0,0 +1,1107 @@
...
+        _GENERAL: lambda x: x,
+        _ELEMENT_CONSTRAINT: lambda x: x,
+    }
+
+def _handle_before_expression_node(visitor, node, time1, time2, delay):
+    t1 = (_GENERAL, _time_point_dispatchers[time1[0]](time1[1]))
+    t2 = (_GENERAL, _time_point_dispatchers[time2[0]](time2[1]))
+    lhs = _handle_sum_node(visitor, None, t1, delay)
+    if (time1[0] in {_GENERAL, _ELEMENT_CONSTRAINT} or
+        time2[0] in {_GENERAL, _ELEMENT_CONSTRAINT}):
```
jsiirola, line 706: It is slightly more efficient to only create this set once and cache it at the module scope

pyomo/contrib/cp/repn/docplex_writer.py
```diff
@@ -0,0 +1,1107 @@
...
+        return True, None
+
+    def exitNode(self, node, data):
+        return self._operator_handles[node.__class__](self, node, *data)
+
+    finalizeResult = None
+
+
+def collect_valid_components(model, active=True, sort=None, valid=set(),
+                             targets=set()):
```
jsiirola, line 824: I worry a bit about the efficiency of this method.  This checks the ctype of each data object and not just at the container level, plus creating explicit lists of every component data in the model...

pyomo/contrib/cp/repn/docplex_writer.py
```diff
@@ -0,0 +1,1107 @@
...
+            m = cp.CpoModel()
+            m.add(cp.all_diff(x))
+            try:
+                m.solve()
+                CPOptimizerSolver._unrestricted_license = True
+            except docplex.cp.solver.solver.CpoSolverException:
+                CPOptimizerSolver._unrestricted_license = False
+        return CPOptimizerSolver._unrestricted_license
+
+    def solve(self, model, **kwds):
```
jsiirola, line 1017: The `solve()` is completely uncovered by tests?

pyomo/contrib/cp/interval_var.py
emma58: Eeeek, no, good catch. It had better be optional in that case.

jsiirola (approved):

PR #2576, opened 2022-10-25
Pickle transformed GDP models

### LINKED ISSUE #2417: Transformed GDP models don't pickle with non-dill pickles
## Summary

The GDP transformations create weakrefs, meaning that transformed models are currently unpickleable.

emma58:
## Summary/Motivation:

The `bigm` and `hull` transformations create Blocks for each `Disjunct` they transform, each with a weakref back to the original component. This PR makes those Blocks into a new component that defines an `__autoslot_mappers__` dict so that the transformed model can be pickled.

- Defines `_TransformedDisjunct` class, which is just a Block with a weakref on it
- Adds `__autoslot_mappers__` dicts to Disjuncts and Disjunctions
- Tests pickling GDP models transformed with bigm and hull

### Legal Acknowledgement

### DIFF
--- a/pyomo/gdp/transformed_disjunct.py
+++ b/pyomo/gdp/transformed_disjunct.py
@@ -11,7 +11,9 @@
 
 from pyomo.common.autoslots import AutoSlots
 from pyomo.core.base.block import _BlockData, Block
-from pyomo.core.base.global_set import UnindexedComponent_index
+from pyomo.core.base.global_set import (
+    UnindexedComponent_index, UnindexedComponent_set
+)
 
 class _TransformedDisjunctData(_BlockData):
     __slots__ = ('_src_disjunct',)
@@ -32,7 +34,7 @@ class _TransformedDisjunct(Block):
     def __new__(cls, *args, **kwds):
         if cls != _TransformedDisjunct:
             return super(_TransformedDisjunct, cls).__new__(cls)
-        if args == ():
+        if not args or (args[0] is UnindexedComponent_set and len(args) == 1):
             return _ScalarTransformedDisjunct.__new__(
                 _ScalarTransformedDisjunct)
         else:
--- a/pyomo/gdp/transformed_disjunct.py
+++ b/pyomo/gdp/transformed_disjunct.py
@@ -31,27 +31,6 @@ class _TransformedDisjunctData(_BlockData):
 class _TransformedDisjunct(Block):
     _ComponentDataClass = _TransformedDisjunctData
 
-    def __new__(cls, *args, **kwds):
-        if cls != _TransformedDisjunct:
-            return super(_TransformedDisjunct, cls).__new__(cls)
-        if not args or (args[0] is UnindexedComponent_set and len(args) == 1):
-            return _ScalarTransformedDisjunct.__new__(
-                _ScalarTransformedDisjunct)
-        else:
-            return _IndexedTransformedDisjunct.__new__(
-                _IndexedTransformedDisjunct)
-
     def __init__(self, *args, **kwargs):
         kwargs.setdefault('ctype', Block)
         Block.__init__(self, *args, **kwargs)
-
-class _ScalarTransformedDisjunct(_TransformedDisjunctData,
-                                 _TransformedDisjunct):
-    def __init__(self, *args, **kwds):
-        _TransformedDisjunctData.__init__(self, self)
-        _TransformedDisjunct.__init__(self, *args, **kwds)
-        self._data[None] = self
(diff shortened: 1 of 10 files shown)


pyomo/gdp/transformed_disjunct.py
```diff
@@ -0,0 +1,55 @@
...
+from pyomo.core.base.block import _BlockData, Block
+from pyomo.core.base.global_set import UnindexedComponent_index
+
+class _TransformedDisjunctData(_BlockData):
+    __slots__ = ('_src_disjunct',)
+    __autoslot_mappers__ = {'_src_disjunct': AutoSlots.weakref_mapper}
+
+    @property
+    def src_disjunct(self):
+        return self._src_disjunct
```
jsiirola, line 22: If we are changing the API, should this property resolve the weakref?  That is,
```python
    @property
    def src_disjunct(self):
        return None if self._src_disjunct is None else self._src_disjunct()
```

pyomo/gdp/transformed_disjunct.py
```diff
@@ -0,0 +1,55 @@
...
+        # pointer to the Disjunct whose transformation block this is.
+        self._src_disjunct = None
+
+class _TransformedDisjunct(Block):
+    _ComponentDataClass = _TransformedDisjunctData
+
+    def __new__(cls, *args, **kwds):
+        if cls != _TransformedDisjunct:
+            return super(_TransformedDisjunct, cls).__new__(cls)
+        if args == ():
```
jsiirola, line 35: This test is usually:
```python
if not args or (args[0] is UnindexedComponent_set and len(args) == 1):
```

jsiirola (commented):
A couple quick questions.  Also, can we add a test somewhere to help with patch coverage?

pyomo/gdp/transformed_disjunct.py
emma58: It could, but it would be inconsistent with the mirror `transformation_block` property on `Disjunct`, which is why I chose this. I don't have strong feelings though.

jsiirola: I am coming around to the opinion that the use of weakrefs in Pyomo is an implementational detail that should not "bleed" into the user experience.  So maybe I would argue for hiding the weakref behind a property in both cases?  Doesn't necessarily have to be this PR, though.

emma58:
@jsiirola, so, I think I was being a bit silly: Probably the patch coverage is low because we never actually make ScalarTransformedDisjuncts... Should we just not define it? Or I can make one in a test, but I don't really know what it's for.

jsiirola:
There really isn't a use case for the scalar, is there?  It would have to be a model with a single 1-term disjunction?

If that's the case, then I think the right thing is to remove the scalar class entirely.  If there is only indexed, then you can also delete the `__new__` stuff, too!

pyomo/gdp/transformed_disjunct.py
emma58: OK, I think I agree with you. I'm going to slip it into this PR...

pyomo/gdp/transformed_disjunct.py
```diff
@@ -0,0 +1,36 @@
...
+        _BlockData.__init__(self, component)
+        # pointer to the Disjunct whose transformation block this is.
+        self._src_disjunct = None
+
+class _TransformedDisjunct(Block):
+    _ComponentDataClass = _TransformedDisjunctData
+
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault('ctype', Block)
+        Block.__init__(self, *args, **kwargs)
```
jsiirola, line 36: FWIW, setting the `ctype` is redundant (it would get set to Block by the base class), so this entire method could be omitted.

pyomo/gdp/transformed_disjunct.py
```diff
@@ -0,0 +1,36 @@
...
+    @property
+    def src_disjunct(self):
+        return None if self._src_disjunct is None else self._src_disjunct()
+
+    def __init__(self, component):
+        _BlockData.__init__(self, component)
+        # pointer to the Disjunct whose transformation block this is.
+        self._src_disjunct = None
+
+class _TransformedDisjunct(Block):
```
jsiirola, line 31: Should this not inherit from `IndexedBlock`?

jsiirola (approved):
A couple (minor) questions, but this looks good!

pyomo/gdp/transformed_disjunct.py
emma58: Ah, right, good point!

PR #2592, opened 2022-11-01
Implementing multiple-bigM transformation in GDP

emma58:
## Summary/Motivation:

The 2015 paper of Trespalacios and Grossmann, "Improved Big-M reformulation for generalized disjunctive programs," proposes a variation of the bigm transformation that relaxes disjunctive constraints not just based on whether or not their disjunct is selected, but also based on which other disjunct in the same disjunction *is* selected. This allows for more specific M values, so is at least as tight as traditional bigm, and is likely often much tighter for problems where many of the disjunctive constraints have the same variables (and similar sub-expressions on those variables). The starring example of the moment being GDP formulations for decision trees.

- Implements the multiple-bigm transformation in pyomo.gdp. Specifically, if M values are not specified, the transformation will calculate them using the approach given in equation (2) of Trespalacios and Grossmann. This involves solving a great many continuous optimization problems (on the order number of disjunctive constraints times number of disjuncts in their disjunctions), so the transformation includes a method `get_all_M_values` that exports the M values used in the transformation. So as long as the model remains unchanged, the user can only calculate Ms once.
- The only major embellishment to the Trespalacios and Grossmann paper is the addition of an option to handle "bound constraints" specially. That is, instead of doing the rote transformation for disjunctive constraints that enforce new upper and lower bounds on single variables, we can optionally write a single transformed constraint bounding that variable above or below with the sum over the disjuncts in the disjunction of the products of the disjunct's indicator_var and its local bound. This option is currently called `tighten_bound_constraints` which is (I think) a terrible name, as I'm not sure it's true. I'm taking suggestions! :)
- Note this transformation does support nested GDPs, though in a fairly brain-dead way: We still only calculate M values locally to the Disjuncts ( since doing more would make the subproblems into MI(N)LPs), and put transformed constraints above the root Disjunct in the tree. This is essentially the same way we handle nested GDP in bigm.
- Last, note that this moves some general utilities from bigm to `util.py` so they can be used in this transformation also. I'm actually working up the courage to write a base class for the GDP-to-MI(N)LP transformations, so this is intended to be a temporary solution.

### Legal Acknowledgement


pyomo/gdp/tests/test_mbigm.py
```diff
@@ -0,0 +1,684 @@
...
+        # we don't transform d1
+        self.assertIsNone(m.d4.transformation_block)
+        # and everything else is the same
+        self.check_linear_func_constraints(m, mbm)
+        self.check_all_untightened_bounds_constraints(m, mbm)
+
+    @unittest.skipUnless(gurobi_available, "Gurobi is not available")
+    def test_var_bounds_substituted_for_missing_bound_constraints(self):
+        m = self.make_model()
+        # Add a new thing with conttraints that don't give both bounds on x1 and
```
bernalde, line 524: Typo: constraints

pyomo/gdp/tests/test_mbigm.py
```diff
@@ -0,0 +1,684 @@
...
+        # We'll just put another disjunction on one of our disjuncts to test
+        # this. There is little that is special here--we transform from leaf to
+        # root, so the child Disjuncts do not interfere in the calculation of
+        # the M values for their parents. However, their transformed constraints
+        # are placed above the parent, so note that this means that for certain
+        # models, it would be beneficial to flatten the model before using this
+        # transformation, if the children *do* provide insight into the bounds
+        # of the parent. (One could use them and in the subproblems and then
+        # move them up, but that means the M-value subproblems would be
+        # MI(N)LPs, which is a downside.)
```
bernalde, line 593: Shouldn't the constraints from the leaf disjuncts be included in their parent disjunction? That would mean they only interfere within that disjunct (contrary to leaving them above the parent). In most cases, moving the leaf disjuncts' reformulated constraints to the root of the model will lead to unnecessary binary variables, but I can imagine an example where it might be invalid for the other parent disjunctions, e.g., [1 <= x <= 3, [1 <= y <= 5] v [6 <= y <= 10]] v [5 <= x <= 10, 15 <= y <= 20]. Moving the inner disjunction constraints to the root and computing bounds based on it for the second parent disjunct leads to infeasibility.

doc/OnlineDocs/modeling_extensions/gdp/solving.rst
```diff
@@ -87,6 +86,23 @@ To apply the BM reformulation within a python script, use:
...
+By default, the MBM transformation will solve continuous subproblems in order to calculate M values.
+This process can be time-consuming, so the transformation also provides a method to export the M values used as a dictionary and allows for M values to be provided through the `bigM` argument.
+
+For example, to apply the transformation and store the M values, use:
+
+.. code::
+
+    mbigm = TransformationFactory('gdp.mbigm')
+    mbigm.apply_to(model)
+    M_values = mbigm.get_all_M_values()
```
bernalde, line 102: Maybe add a line here showing how these M_values can be fed to another (could be a copy of model) model?

bernalde (commented):
This is a great PR @emma58 ! It relates directly to a recent conversation with @carldlaird and Bashar about linear trees.
This method is valid for convex GDPs, right? I didn't see any nonlinear case, but I assume it should work out of the box.
Finally, although explicitly mentioned by Trespalacios and Grossmann, some work proposes a similar technique to the special bounded case you are using for the bound-disjunctions, shown in https://doi.org/10.1137/130915303 section 6.2. I think it might be worth adding it to the references.

pyomo/gdp/tests/test_mbigm.py
emma58: Ah, oops, that's a very good point. Because this transformation assumes the exactly-one constraint will be enforced and it isn't necessarily going to be for nested disjunctions. I think you're right that the safest thing to do is just to put transformed constraints on the parent Disjunct. That probably means that flattening the model will often lead to a better relaxation for this transformation...

emma58: Actually, on second thought, I really don't think this transformation makes sense for nested GDPs. Because if I did what I said above, the calculated M values would depend on the order the Disjuncts are transformed in--that is, whether or not you use the inner disjunctions' information. I can construct examples where this is weaker than flattening the GDP first, so I don't know why you would do it.

bernalde: So you are proposing to flatten the disjunction first?

emma58:
@bernalde, thank you so much for your review--that was a very good catch on the nested GDP front! I've made the changes you suggested, though I'm punting on if this is even worth it for nested GDP at the moment and changed to not supporting them at all. And to answer your question, yes: it should work for nonlinear (I think even non-convex, if you solve for valid M's globally) GDPs. I set the default solver as Gurobi, so that would need adjusting. I didn't test with anything nonlinear because calculating the M's is expensive even when you're solving LPs, but the logic should be the same.

pyomo/gdp/tests/test_mbigm.py
emma58: I'm really proposing to punt. I think flattening makes more sense than the leaf-to-root version. So unless we can think of a use case for the latter, I'm thinking I don't implement it.

jsiirola (approved):

pyomo/gdp/plugins/multiple_bigm.py
```diff
@@ -0,0 +1,724 @@
...
+                boolean.associate_binary_var(new_var)
+
+        # add references to all local variables on block (including the
+        # indicator_var). We won't have to do this when the writers can find
+        # Vars not in the active subtree.
+        varRefBlock = disjunct._transformation_block().localVarReferences
+        for v in disjunct.component_objects(Var, descend_into=Block,
+                                            active=None):
+            varRefBlock.add_component(unique_component_name(
+                varRefBlock, v.getname(fully_qualified=True)), Reference(v))
```
jsiirola, line 354: Per #2605, shouldn’t this be `fully_qualified=False`?

emma58: Yes...

pyomo/gdp/plugins/multiple_bigm.py
```diff
@@ -0,0 +1,724 @@
...
+        and Disjuncts to Big-M values valid for relaxing the constraint if
+        the Disjunct is chosen.
+
+        Note: Unlike in the bigm transformation, we require the keys in this
+        mapping specify the components the M value applies to exactly in order
+        to avoid ambiguity.
+        """
+    ))
+    CONFIG.declare('reduce_bound_constraints', ConfigValue(
+        default=False,
```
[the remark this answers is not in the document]
emma58: I put it has False because I thought in some sense the behavior is "unexpected" when it's True--that is it makes a structure different than rote multiple bigm would. But there's no mathematical or performance reason not to let it default to True. It's actually a nicer formulation, and it saves a lot of work if you are calculating M values, especially.

PR #2605, opened 2022-11-10
BigM tranformation performance improvements

emma58:
## Summary/Motivation:

This is a first pass at making the bigm transformation less slow. In a certain large model (that takes about 14 seconds to build), this reduces the bigm transformation time from 87 seconds to 66 seconds. We can do better, especially if we didn't have to mess around with variables so much in `_transform_block_components`, but it helps, and this is probably already a big enough PR...

- Adds an option to fbbt's `compute_bounds_on_expr` function to ignore that variables are fixed, if you want the bounds calculated based on fixed Vars' bounds rather than their values. (This saves us having to iterate through every variable in the model to unfix and refix it on either side of calculating M values.)
- Creates only one indexed constraint per Disjunct for holding transformed constraints. Regenerating all of the index sets in the original model is slow and redundant.
- Uses only local names when naming transformed components. The clarity can come from mapping to the originals, and fully-qualified names are slower.
- Changes the behavior of the `algebraic_constraint` property for Disjunctions to match the changes in Disjunct (hiding weakrefs from the user)
- Overhaul of tests for bigm so that they do not rely on the structure of the transformed components or their names--making much better use of the transformation's methods for mapping original and transformed components. (This creates a few more differences in from `hull` that I plan to handle more elegantly later, when I make reciprocal changes to the hull transformation).

### Legal Acknowledgement


pyomo/gdp/plugins/bigm.py
```diff
@@ -499,28 +501,24 @@ def _transform_constraint(self, obj, disjunct, bigMargs, arg_list,
...
+        
+        # We will make indexes from ({obj.local_name} x obj.index_set() x ['lb',
+        # 'ub']), but don't bother construct that set here, as taking Cartesian
+        # products is kind of expensive (and redundant since we have the
+        # original model)
+        newConstraint = transBlock.transformedConstraints
+        # Since we are both combining components from multiple blocks and using
+        # local names, we need to make sure that the first index for
+        # transformedConstraints is guaranteed to be unique. We just grab the
+        # currentl length of the list here since that will be monotonically
```
jsiirola, line 513: Typo: "current"

jsiirola (approved):
Looks good.  One minor typo.

PR #2619, opened 2022-11-21
Adding support for local suffixes in scaling transformation

### LINKED ISSUE #898: Scaling transformations on Block structured models
I have been working with the Pyomo scaling transformation in preparation for implementing this within IDAES, and I have run into some unexpected behaviour when working with block-structured models. It appears that only scaling factors declared at the top level of the model are considered by the scaling transformation, and that it does not descend into the block tree to look for additional scaling factors.

I have tested a few cases to try to work out what was happening.

Case 1:

```
import pyomo.environ as pe

###
# create the original unscaled model
###
model = pe.ConcreteModel()

model.b = pe.Block()
model.b.a = pe.Var(initialize=10, bounds=(0,20))

###
# set the scaling parameters
###
model.b.scaling_factor = pe.Suffix(direction=pe.Suffix.EXPORT)
model.b.scaling_factor[model.b.a] = 10

###
# build and solve the scaled model
###
scaled_model = pe.TransformationFactory('core.scale_model').create_using(model)
```

Returns:
`ValueError: ScaleModel transformation called with scaling_method='user', but cannot find the suffix 'scaling_factor' on the model`

This is possibly expected behaviour, but without the `scaling_factor` suffix present at the top level, the transformation returns an error. I do find it strange that the error is a `ValueError` - I would generally expect an `AttributeError` in the case where the attribute was missing completely.

Case 2:

```
import pyomo.environ as pe

###
# create the original unscaled model
###
model = pe.ConcreteModel()

model.b = pe.Block()
model.b.a = pe.Var(initialize=10, bounds=(0,20))

###
# set the scaling parameters
###
model.scaling_factor = pe.Suffix(direction=pe.Suffix.EXPORT)
model.b.scaling_factor = pe.Suffix(direction=pe.Suffix.EXPORT)
model.b.scaling_factor[model.b.a] = 10

###
# build and solve the scaled model
###
scaled_model = pe.TransformationFactory('core.scale_model').create_using(model)

# print the scaled model
scaled_model.b.scaled_a.pprint()
```

Returns:
```
scaled_a : Size=1, Index=None
    None :   0.0 :  10.0 :  20.0 : False : False :  Reals
```

The model has been transformed, and the variable `a` replaced by `scaled_a`, but the value and bounds have not been scaled. It appears that the scaling factor(s) defined in child blocks are not being picked up by the transformation.

Case 3:
```
import pyomo.environ as pe

###
# create the original unscaled model
###
model = pe.ConcreteModel()

model.b = pe.Block()
model.b.a = pe.Var(initialize=10, bounds=(0,20))

###
# set the scaling parameters
###
model.scaling_factor = pe.Suffix(direction=pe.Suffix.EXPORT)
model.scaling_factor[model.b.a] = 10

###
# build and solve the scaled model
###
scaled_model = pe.TransformationFactory('core.scale_model').create_using(model)

# print the scaled model
scaled_model.b.scaled_a.pprint()
```

Returns:
```
scaled_a : Size=1, Index=None
    None :   0.0 : 100.0 : 200.0 : False : False :  Reals
```

This works now, and the variable value and bounds have been scaled.

It would be good if the scaling transformation would descend into the block structure to look for scaling factors, as it makes more sense to declare these at the same level as the component is declared. This also simplifies modular construction of the scaling factors, as each block can be responsible for its own scaling factors, without having to propagate these up to the top level of the model.

andrewlee94:
## Fixes #898.

Related to #2641 

## Summary/Motivation:
The scaling transformation currently only looks for a scaling suffix on the model to which the transform is being applied. However, when dealing with hierarchical models with sub-Blocks, each sub-Block may contain a scaling suffix for its own components (this occurs frequently in IDAES where sub-Blocks are generated procedurally from pre-defined classes).

This PR adds additional checks such that if a suffix for a component is not found on the main instance being transformed, it will then look to see if there is a suffix declared on the `parent_block` of the component as well.

- Add `elif` statements to check for scaling suffix on `component.parent_block()`
- Adds a test for a hierarchical model

### Legal Acknowledgement


Robbybp:
@andrewlee94 Thanks for starting these docs!

doc/OnlineDocs/model_transformations/scaling.rst
```diff
@@ -0,0 +1,45 @@
...
+
+In-Place Scaling
+****************
+
+The ``ScaleModel.apply_to(model)`` method can be used to apply scaling directly to an existing model. When using this method, all the variables, constraints and objectives within the target model are replaced with new scaled components and the appropriate scaling factors applied. The model can then be sent to a solver as usual, however the results will be in terms of the scaled components and must be un-scaled by the user.
+
+Creating a New Scaled Model
+***************************
+
+Alternatively, the ``ScaleModel.create_using(model)`` method can be used to create a new, scaled version of the model which can be solved. In this case, a clone of the original model is generated with the variables, constraints and objectives replaced by scaled equivalents. Users can the send the scaled model to a solver after which the ``ScaleModel.propagate_solution`` method can be used to map the scaled solution back onto the original model for further analysis.
```
blnicho, line 42: Typo: in the last sentence "Users can **then** send the..."

doc/OnlineDocs/model_transformations/scaling.rst
```diff
@@ -0,0 +1,45 @@
+Model Scaling Transformation
+============================
+
+Good scaling of models can greatly improve the numerical properties of a problem and thus increase reliability and convergence. The ``ScaleModel`` transformation allows users to separate scaling of a model from the declaration of the model variables and constraints which allows for models to be written in more natural forms and to be scaled and rescaled as required without having to rewrite the model code.
```
blnicho, line 4: Instead of using the `ScaleModel` class name should we use the `core.scale_model` name provided to the `TransformationFactory`?

blnicho (approved):
Two minor comments related to the documentation but otherwise I think this looks fine.

pyomo/core/plugins/transform/scaling.py
```diff
@@ -87,6 +87,10 @@ def _get_float_scaling_factor(self, instance, component_data):
             scaling_factor = instance.scaling_factor[component_data]
         elif component_data.parent_component() in instance.scaling_factor:
             scaling_factor = instance.scaling_factor[component_data.parent_component()]
+        elif component_data in component_data.parent_block().scaling_factor:
+            scaling_factor = component_data.parent_block().scaling_factor[component_data]
+        elif component_data.parent_component() in component_data.parent_block().scaling_factor:
+            scaling_factor = component_data.parent_block().scaling_factor[component_data.parent_component()]
```
jsiirola, line 93: This makes the assumption that the `parent_block` *has* a `scaling_factor` Suffix.  If it is not present, this will raise an AtributeError.  I think that a better solution may be to follow the pattern in GDP's logic for hunting fors BigM suffixes (start at the parent component and walk up each block in the hierarchy looking for the first Suffix that "matches".

jsiirola (changes requested):
Overall, this is very nice.  There is a problem with the implementation, though (the implementation makes assumptions about the existence of Suffxes that is not likely to be universally true).

pyomo/core/plugins/transform/scaling.py
```diff
@@ -81,16 +81,61 @@ def _create_using(self, original_model, **kwds):
...
+                elif component_data.parent_component() in suffix:
+                    val = suffix[component_data.parent_component()]
+                # If we don't have a value yet, look for a default value whist we are here
+                # Only need this if we haven't found anything specific yet
+                elif val is None and None in suffix:
+                    val = suffix[None]
+
+            except AttributeError:
+                # It's OK if there is no suffix
+                pass
```
jsiirola, line 125: It is slightly more efficient to avoid raising an exception (and not have a try-except):
```python
suffix = getattr(parent, suffix_name, None)
if suffix is None:
    continue
```

pyomo/core/plugins/transform/scaling.py
```diff
@@ -81,16 +81,61 @@ def _create_using(self, original_model, **kwds):
...
+        Args:
+            component_data: component data object to find suffix value for.
+            suffix_name: name of Suffix to search for.
+            top_level: Block at which to stop searching for Suffixes (default=None).
+
+        Returns:
+            Value for suffix associated with component data if found, else None.
+        """
+        # Prototype for Suffix finder
+        val = None  # default return of None if no corresponding suffix found
```
jsiirola, line 103: This should probably default to `NOTSET`: A user may assign a value of None to the Suffix...

pyomo/core/plugins/transform/scaling.py
```diff
@@ -81,16 +81,61 @@ def _create_using(self, original_model, **kwds):
...
+                suffix = getattr(parent, suffix_name)
+                # If a value exists for component_data, this supersedes any value we already have
+                if component_data in suffix:
+                    val = suffix[component_data]
+                elif component_data.parent_component() in suffix:
+                    val = suffix[component_data.parent_component()]
+                # If we don't have a value yet, look for a default value whist we are here
+                # Only need this if we haven't found anything specific yet
+                elif val is None and None in suffix:
+                    val = suffix[None]
```
jsiirola, line 121: This logic doesn't match the description above: among other things, `None` at a higher level will override a specific suffix at a lower level.

pyomo/core/plugins/transform/scaling.py
```diff
@@ -81,16 +81,61 @@ def _create_using(self, original_model, **kwds):
...
+            Value for suffix associated with component data if found, else None.
+        """
+        # Prototype for Suffix finder
+        val = None  # default return of None if no corresponding suffix found
+
+        # Walk parent tree and search for suffixes
+        parent = component_data.parent_block()
+        # TODO: Stop at instance, or run out the top of the stack?
+        # For now, run out the top
+        while parent is not top_level:
```
jsiirola, line 109: Note that this will NOT look for suffixes at the top_level if top_level is not None...

jsiirola (changes requested):
This is a great start, however there are several logic errors in `_suffix_finder()`.

jsiirola (approved):
After talking through this with @andrewlee94, I went ahead and implemented the necessary changes (and added tests for the cases that were failing in the original implementation)

PR #2625, opened 2022-11-25
Xpress fix nonconvex miqcp

djunglas:
## Fixes 

This pull request fixes handling of models with non-convex quadratic constraints.

## Summary/Motivation:

The Xpress optimizer is able to handle non-convex quadratic constraints using its non-linear solver. The API for this is a bit different from the API for MIP and LP, though. So some special care is required to handle these problems correctly.

- Always use Xpress's `nlpoptimize()` for models with quadratic constraints. If all constraints are convex then this will implicitly forward to a plain MIP/continuous solve.
- Shuffle around some code so that supporting 9.0 will be easier.

### Legal Acknowledgement

### DIFF
--- a/pyomo/solvers/plugins/solvers/xpress_direct.py
+++ b/pyomo/solvers/plugins/solvers/xpress_direct.py
@@ -400,6 +400,7 @@ class XpressDirect(DirectSolver):
             # The problem was non-linear
             status = xprob_attrs.xslp_nlpstatus
             have_soln, check_soln = False, False
+            optimal = False # *globally* optimal?
             if status == xp.nlp_unstarted:
                 results.solver.status = SolverStatus.unknown
                 results.solver.termination_message = "Non-convex model solve was not start"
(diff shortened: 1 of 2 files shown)


blnicho:
@bknueven would you mind reviewing this PR?

blnicho:
@djunglas could you merge main into your branch? Should resolve the test failures.

pyomo/solvers/plugins/solvers/xpress_direct.py
```diff
@@ -223,8 +223,268 @@ def _apply_solver(self):
...
+                pass
+            try:
+                results.problem.lower_bound = xprob_attrs.mipbestobjval
+            except (XpressDirect.XpressException, AttributeError):
+                pass
+            
+        # TODO: Pyomo only queries solution if the LP status is optimal.
+        #       This seems wrong at first glance
+        return (xprob_attrs.lpstatus == xp.lp_optimal and mip_sols > 0,
+                mip_sols > 0)
```
bknueven, line 307: I agree with the `TODO` (and it was almost certainly my mistake originally). I think the `_get_*_results` functions should just return a single flag, "is a solution available or not". This will simplify things in several places.

bknueven (commented):
Looks good. The new functionality should "just work" with XpressPersistent as well.

I only have a minor comment -- we should fix the existing inconsistencies on when a solution is loaded.

Issue #2626, opened 2022-11-25
Fixed Vars unpickle as stale


emma58:
## Summary

I'm not sure if this is a bug, but it seems unexpected? Anyway, if you pickle a model that has a fixed variable (not stale), when you unpickle it, it comes back as stale.

### Steps to reproduce the issue

```
from pyomo.environ import *
import pickle

m = ConcreteModel()
m.x = Var(domain=Binary)
m.x.fix(1)

unpickle = pickle.loads(pickle.dumps(m))

m.x.pprint()
unpickle.x.pprint()
```

```
x : Size=1, Index=None
    Key  : Lower : Value : Upper : Fixed : Stale : Domain
    None :     0 :     1 :     1 :  True : False : Binary
x : Size=1, Index=None
    Key  : Lower : Value : Upper : Fixed : Stale : Domain
    None :     0 :     1 :     1 :  True :  True : Binary
```

### Error Message

It seems like these models should be identical, even up to stale-ness, right?

### Information on your system

Pyomo version: main
Python version: 3.8
Operating system: linux
How Pyomo was installed (PyPI, conda, source): source 
Solver (if applicable):

PR #2627, opened 2022-11-25
Deprecating logical_to_linear, adding logical_to_disjunctive

emma58:
## Summary/Motivation:

The `core.logical_to_linear` transformation is very slow because it uses sympy to convert logical expressions to CNF. Since we don't really need them in CNF in order to write a MIP (which is the goal of the transformation), this PR adds a transformation to `contrib.cp` called `logical_to_disjunctive` that transforms logical constraints into disjunctive programs in the sense that 1) logical constraints on only Boolean variables are converted to algebraic constraints on binary variables, but 2) logical constraints on mixtures of Boolean and integer variables (such as `exactly`, `atmost`, and `atleast`) are transformed into disjunctions over binaries and integers. (Note to future us: We're using the term "disjunctive program" rather than GDP intentionally here, as the promise of the transformation is to no longer have any logical expressions in the model: only algebraic constraints and disjunctions of algebraic constraints.)

This reduces the `gdp.bigm` transformation time for a certain model that builds in 15 seconds from 129 seconds (current main branch) to 75 seconds (this branch).

NOTE: This is currently failing tests because of #2626 (because there are logical constraints in the pickle tests and `logical_to_disjunctive` fixes some variables.)

- Adds `contrib.logical_to_disjunctive` transformation to `contrib.cp`
- Deprecates `core.logical_to_linear`, mainly because it will not be easily extensible as we add to the set of supported nodes in the logical expression system. For example, `exactly`, `atmost`, and `atleast` are handled with ad hoc bigm-ing right now--we would have to keep adding special handling as we include `alldiff`, etc.
- Transitions all the GDP transformations to use `contrib.logical_to_disjunctive`
- Removes support for transforming BooleanVars declared on Disjuncts that are not used in expressions in the active tree being transformed.

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/cp/transform/logical_to_disjunctive_walker.py
+++ b/pyomo/contrib/cp/transform/logical_to_disjunctive_walker.py
@@ -12,14 +12,18 @@
 import collections
 
 from pyomo.common.collections import ComponentMap
+from pyomo.common.errors import MouseTrap
 from pyomo.core.expr.visitor import StreamBasedExpressionVisitor
+from pyomo.core.expr.numeric_expr import NumericExpression
+from pyomo.core.expr.relational_expr import RelationalExpression
 import pyomo.core.expr.current as EXPR
 import pyomo.core.expr.logical_expr as LE
-from pyomo.core.base import Binary, VarList, ConstraintList
+from pyomo.core.base import (
+    Binary, Constraint, ConstraintList, NonNegativeIntegers, VarList)
 import pyomo.core.base.boolean_var as BV
 from pyomo.core.base.expression import ScalarExpression, _GeneralExpressionData
 from pyomo.core.base.var import ScalarVar, _GeneralVarData
-from pyomo.gdp.disjunct import AutoLinkedBooleanVar
+from pyomo.gdp.disjunct import AutoLinkedBooleanVar, Disjunct, Disjunction
 
 def _dispatch_boolean_var(visitor, node):
     if node not in visitor.boolean_to_binary_map:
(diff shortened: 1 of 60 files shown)


pyomo/contrib/cp/transform/logical_to_disjunctive_walker.py
```diff
@@ -0,0 +1,235 @@
...
+    def beforeChild(self, node, child, child_idx):
+        if child.__class__ in EXPR.native_types:
+            return False, child
+
+        if isinstance(child, NumericExpression):
+            # Just pass it through, we'll figure it out later
+            return False, child
+        if isinstance(child, RelationalExpression):
+            # Eventually we'll handle these. Right now we set a MouseTrap
+            return _before_relational_expr(self, child)
```
jsiirola, line 222: I think you should use `child.is_numeric_type()` and `child.is_expression_type(ExpressionType.RELATIONAL)` instead of `isinstance()` here (it is faster, and allows for duck-typing).

jsiirola (approved):
This looks good!  One minor performance-related comment, but doesn't need to hold up this PR

emma58:
Oh, whoops, that's only kind of true... There are GDPopt tests that are unrelated to the pickle problems.

emma58:
Marking this as WIP for now--I didn't realize there are tests in core that call bigm and hull, so I need to fix those.

jsiirola (approved):
I think this looks great.  One question: I do agree that we should switch GDP from `logical_to_linear` to using `logical_to_disjunctive`; however, do we want/need to deprecate `logical_to_linear`?  While the implementation has several limitations, it does provide a somewhat different capability (the CNF form).

Issue #443, opened 2018-04-25
component_data_objects(active=False) does not yield correct objects


qtothec:
# Test case
```
from __future__ import division

from pyomo.environ import *
from pyomo.gdp import *

model = ConcreteModel()
model.lt = Var(initialize=0)
model.x1 = Var(initialize=0, bounds=(0, 15))
model.x2 = Var(initialize=0, bounds=(0, 15))
model.x3 = Var(initialize=0, bounds=(0, 15))
model.y1 = Var(initialize=6, bounds=(6, 10))
model.y2 = Var(initialize=7, bounds=(7, 10))
model.y3 = Var(initialize=3, bounds=(3, 10))

model.c1 = Constraint(expr=model.lt >= model.x1 + 6)
model.c2 = Constraint(expr=model.lt >= model.x2 + 5)
model.c3 = Constraint(expr=model.lt >= model.x3 + 4)


def disjunct11(disj):
    disj.indicator_var.value = 1
    disj.c = Constraint(expr=model.x1 + 6 <= model.x2)


def disjunct12(disj):
    disj.indicator_var.value = 0
    disj.c = Constraint(expr=model.x2 + 5 <= model.x1)


def disjunct13(disj):
    disj.indicator_var.value = 0
    disj.c = Constraint(expr=model.y1 - 6 >= model.y2)


def disjunct14(disj):
    disj.indicator_var.value = 0
    disj.c = Constraint(expr=model.y2 - 7 >= model.y1)


def disjunct21(disj):
    disj.indicator_var.value = 1
    disj.c = Constraint(expr=model.x1 + 6 <= model.x3)


def disjunct22(disj):
    disj.indicator_var.value = 0
    disj.c = Constraint(expr=model.x3 + 4 <= model.x1)


def disjunct23(disj):
    disj.indicator_var.value = 0
    disj.c = Constraint(expr=model.y1 - 6 >= model.y3)


def disjunct24(disj):
    disj.indicator_var.value = 0
    disj.c = Constraint(expr=model.y3 - 3 >= model.y1)


model.d11 = Disjunct(rule=disjunct11)
model.d12 = Disjunct(rule=disjunct12)
model.d13 = Disjunct(rule=disjunct13)
model.d14 = Disjunct(rule=disjunct14)
model.d21 = Disjunct(rule=disjunct21)
model.d22 = Disjunct(rule=disjunct22)
model.d23 = Disjunct(rule=disjunct23)
[...]

PR #2643, opened 2022-12-01
Add `active` filter to flattener

### LINKED ISSUE #2529: Pyomo DAE flatten ignores whether blocks are active
## Summary

In July, I ran into unexpected behavior when the PETSc interface in IDAES uses `pyomo.dae.flatten` to create the block with variables/equations to be solved. It appears that the flattening process ignores whether the block it's aggregating variables from is active or not, and so variables and constraints from deactivated blocks end up being used in the problem because they are not individually deactivated.

@jsiirola suggested the following fix, which has worked for my applications:

```
diff --git a/pyomo/dae/flatten.py b/pyomo/dae/flatten.py
index 922ffa9..13f56ee 100644
--- a/pyomo/dae/flatten.py
+++ b/pyomo/dae/flatten.py
@@ -305,7 +305,10 @@ def generate_sliced_components(b, index_stack, slice_, sets, ctype, index_map):
                 if type(descend_data) is IndexedComponent_slice:
                     try:
                         # Attempt to find a data object matching this slice
-                        descend_data = next(iter(descend_data))
+                        _data = next(iter(descend_data))
+                        while not _data.active:
+                            _data = next(iter(descend_data))
+                        descend_data = _data
                     except StopIteration:
                         # For this particular idx (and given indices), no
                         # block data object exists to descend into.
@@ -313,6 +316,8 @@ def generate_sliced_components(b, index_stack, slice_, sets, ctype, index_map):
                         continue
             else:
                 descend_data = sub
+                if not descend_data.active:
+                    continue

             # Recursively generate sliced components from this data object
             for st, v in generate_sliced_components(
```

I didn't open an issue then, because I assumed our conversation meant it would be fixed without it, but it probably got forgotten about as soon as a more important issue came to the fore.

Edit: Also, it would be good to make including inactive components an option. I can imagine contexts (like scaling methods) where including them would be the desired behavior.

Robbybp:
## Summary/Motivation:
If a block has been deactivated, that usually means you want to ignore it when searching the block hierarchy.

- Add `active` flag to `flatten_dae_components` and `flatten_components_along_sets` that functions similarly to the `active` flag in `component_objects`
- Pass this `active` flag to component objects so slices with in active parent components are not generated
- ~~If a Component object is active, do not check individual data objects of a slice for activity. The "activity" of a slice is undefined and we do not attempt to infer it.~~ Consider a slice to match the active flag if *any* of its data objects match the flag.

### Legal Acknowledgement


dallan-keylogic:
This works as expected on my examples. Here's a test I was using if it makes things easier on your side.

```python3
    def test_block_deactivation(self):
        m = self._model1_1d_sets()
        # Deactivating b1 should get rid of both variables directly on it
        # as well as those on the subblock b_s
        m.b.b1.deactivate()
        sets = ComponentSet((m.time,))
        sets_list, comps_list = flatten_components_along_sets(m, sets, Var, active=True)

        S = m.space
        SS = m.space * m.space
        SC = m.space * m.comp
        SSC = m.space * m.space * m.comp

        assert len(sets_list) == 3
        for sets, comps in zip(sets_list, comps_list):
            if len(sets) == 1 and sets[0] is UnindexedComponent_set:
                ref_data = {
                    self._hashRef(Reference(m.v0)),
                }
                assert len(comps) == len(ref_data)
                for comp in comps:
                    self.assertIn(self._hashRef(comp), ref_data)
            elif len(sets) == 1 and sets[0] is m.time:
                ref_data = {
                    # Components indexed only by time;
                    self._hashRef(Reference(m.v1)),
                }
                # Components indexed by time and some other set(s)
                ref_data.update(self._hashRef(Reference(m.v2[:, x])) for x in S)
                ref_data.update(self._hashRef(Reference(m.v3[:, x, j])) for x, j in SC)
                ref_data.update(self._hashRef(Reference(m.b.b2[:, x].v0)) for x in S)
                ref_data.update(self._hashRef(Reference(m.b.b2[:, x].v1[j])) for x, j in SC)
                assert len(comps) == len(ref_data)
                for comp in comps:
                    self.assertIn(self._hashRef(comp), ref_data)
            elif len(sets) == 2 and sets[0] is m.time and sets[1] is m.time:
                ref_data = {
                    self._hashRef(Reference(m.v_tt)),
                }
                ref_data.update(self._hashRef(Reference(m.v_tst[:, x, :])) for x in S)
                ref_data.update(self._hashRef(Reference(m.b.b2[:, x].v2[:, j])) for x, j in SC)
                assert len(comps) == len(ref_data)
                for comp in comps:
                    self.assertIn(self._hashRef(comp), ref_data)
            else:
                raise RuntimeError()
```

Two remaining questions, one substantial and one trivial: what should default behavior be and is `active=None` any better than `active=False`? I think it's more intuitive for components on deactivated blocks to be ignored by default than for them to be included, at least as far as using `flatten_dae_components` for PETSc is concerned.

Robbybp:
Thanks for the code, this will make it easier for me to write the tests. Regarding default behavior, I think `active=None` should be used to maintain backwards compatibility. This is also the default behavior in `component_objects` and `component_data_objects`, which I think it is nice to be consistent with.

Robbybp:
In the current implementation, the behavior of the `flatten_*` methods differs slightly from `component_objects` when the `active` flag is passed as using `active=False` will not prevent the method from returning variables. Instead it will return only the variables whose parent block are all inactive. I believe this is a bug in `component_objects` as this is inconsistent with its behavior when `active=True` (see my comment on #443), so will keep the current behavior for flattening methods unless I'm told otherwise.

Robbybp:
Just made some updates that actually do try to determine the activity of a slice. I did this because I ran into problems descending into BlockData that didn't match the active arg when choosing the index to descend into for a slice. I now descend into the first index I find that matches the active flag. In doing so, a slice is returned if *any* data object it defines matches the active flag. I also updated the handling for ActiveComponents to be consistent with this.

I'm not super thrilled about having to potentially check every data object defined by a slice to determine whether we match the active flag. This could be bypassed by only checking the flag on the Component (rather than data object), which is just handled by `component_objects`, but we really need to care about the data object when choosing BlockData to descend into. But it's probably better to be correct than fast. And of course all of this is bypassed if `active=None`.

A script summarizing some of the behavior:
```python
import pyomo.environ as pyo 
from pyomo.dae.flatten import flatten_components_along_sets
from pyomo.core.base.set import UnindexedComponent_set

m = pyo.ConcreteModel()
m.time = pyo.Set(initialize=[0, 1, 2, 3]) 
m.b = pyo.Block(m.time)
for t in m.time:
    m.b[t].v = pyo.Var()
def c_rule(m, t): 
    return m.b[t].v == 2
m.c = pyo.Constraint(m.time, rule=c_rule)

def display(sets_list, comps_list):
    for sets, comps in zip(sets_list, comps_list):
        print("  %s" % [s.name for s in sets])
        if len(sets) == 1 and sets[0] is UnindexedComponent_set:
            for comp in comps:
                print("  %s" % pyo.ComponentUID(comp))
        else:
            for comp in comps:
                print("  %s" % pyo.ComponentUID(comp.referent))

sets = (m.time,)
print("active=True")
display(*flatten_components_along_sets(
    m, sets, pyo.Var, active=True
))
print()

print("active=False")
display(*flatten_components_along_sets(m, sets, pyo.Var, active=False))
print()

m.deactivate()
m.b.deactivate()
print("m, m.b deactivated; active=False")
display(*flatten_components_along_sets(m, sets, pyo.Var, active=False))
print()
m.activate()
m.b.activate()

m.b[0].deactivate()
print("m.b[:] partially deactivated; active=True")
display(*flatten_components_along_sets(m, sets, pyo.Var, active=True))
print()

m.b[:].deactivate()
print("m.b[:] fully deactivated; active=True")
display(*flatten_components_along_sets(m, sets, pyo.Var, active=True))
print()
m.b[:].activate()

m.deactivate()
m.c.deactivate()
print("m, m.c deactivated; active=False")
display(*flatten_components_along_sets(m, sets, pyo.Constraint, active=False))
print()

m.c[:].activate()
print("m.c[:] activated, active=False")
display(*flatten_components_along_sets(m, sets, pyo.Constraint, active=False))
print()
```

pyomo/dae/flatten.py
```diff
@@ -234,47 +239,97 @@ def slice_component_along_sets(
...
+            c, sets, context_slice=context_slice, normalize=False
+        ):
+            # If we have to check activity, check data objects defined by
+            # slice. If any match, we yield the slice. This is done for
+            # compatibility with the behavior when slicing blocks, where
+            # we can only descend into a block that matches our active flag.
+            if not check_active or any(
+                data.active == c_active for data in new_slice.duplicate()
+            ):
+                yield sliced_sets, new_slice
```
dallan-keylogic, line 329: Calling `new_slice.duplicate()` causes problems for scalar components:

```
  File "C:\Users\[REDACTED]\Work\soec-control\soec_dynamic_flowsheet_helper.py", line 324, in <module>
    results = petsc.petsc_dae_by_time_element(
  File "c:\users\[REDACTED]\work\idaes-pse\idaes\core\solvers\petsc.py", line 489, in petsc_dae_by_time_element
    regular_cons, time_cons = flatten_dae_components(m, time, pyo.Constraint, active=True)
  File "c:\users\[REDACTED]\work\pyomo\pyomo\dae\flatten.py", line 556, in flatten_dae_components
    sets_list, comps_list = flatten_components_along_sets(
  File "c:\users\[REDACTED]\work\pyomo\pyomo\dae\flatten.py", line 483, in flatten_components_along_sets
    for index_sets, slice_ in generate_sliced_components(
  File "c:\users\[REDACTED]\work\pyomo\pyomo\dae\flatten.py", line 400, in generate_sliced_components
    for st, v in generate_sliced_components(
  File "c:\users\[REDACTED]\work\pyomo\pyomo\dae\flatten.py", line 400, in generate_sliced_components
    for st, v in generate_sliced_components(
  File "c:\users\[REDACTED]\work\pyomo\pyomo\dae\flatten.py", line 327, in generate_sliced_components
    data.active == c_active for data in new_slice.duplicate()
AttributeError: 'ScalarConstraint' object has no attribute 'duplicate'
python-BaseException
Backend TkAgg is interactive backend. Turning interactive mode on.
```

Robbybp: Thanks! I'll add a test that covers this.

dallan-keylogic: Thanks for the fix. With the new update, it works for my application.

pyomo/dae/flatten.py
```diff
@@ -323,8 +323,19 @@ def generate_sliced_components(
...
+            # is empty.
+            if (
+                # Yield if (a) we're not checking activity
+                not check_active
+                # or (b) we have not sliced and data object activity matches
+                or (not sliced_sets and new_slice.active == c_active)
+                # or (c) we did slice and *any* data object activity matches
+                or any(
+                    data.active == c_active for data in new_slice.duplicate()
+                )
```
Robbybp, line 338: TODO: This needs to make sure we have sliced something, i.e. need to check `or (sliced_sets and any(...))`

Robbybp: Need to cover this with a test as well

pyomo/dae/flatten.py
```diff
@@ -234,47 +239,108 @@ def slice_component_along_sets(
...
+    # Define this class so we don't have to call issubclass again later.
+    check_active = issubclass(ctype, ActiveComponent) and (active != None)
+
+    # If active=False and ctype is not an ActiveComponent (e.g. it is Var)
+    # we will not generate any components. To prevent this, only pass the
+    # active argument if we are looking for active components.
+    c_active = active if check_active else None
+
+    # Looks for components indexed by specified sets immediately in our block.
+    for c in b.component_objects(ctype, descend_into=False, active=c_active):
```
Robbybp, line 312: @jsiirola This potentially gets slightly simpler if one of the `active=False` fixes we talked about gets merged. If we remove the `active=True` default for variables, then we can just pass `component_objects` the caller's active argument rather than our computed `c_active`. However, we still need to check if we are dealing with an ActiveComponent so we know whether to check the slice for an active data object below.

pyomo/dae/flatten.py
```diff
@@ -234,47 +239,108 @@ def slice_component_along_sets(
...
+                # Yield if (a) we're not checking activity
+                not check_active
+                # or (b) we have not sliced and data object activity matches
+                or (not sliced_sets and new_slice.active == c_active)
+                # or (c) we did slice and *any* data object activity matches
+                or (sliced_sets and any(
+                    data.active == c_active for data in new_slice.duplicate()
+                ))
+            ):
+                yield sliced_sets, new_slice
```
Robbybp, line 340: This is where we check a slice for an active data object. This is for consistency with behavior when slicing blocks in the hierarchy.

mrmundt (approved):
This looks great - I found one tiny little typo.

doc/OnlineDocs/advanced_topics/flattener/index.rst
```diff
@@ -0,0 +1,44 @@
...
+``component_objects`` or ``component_data_objects`` to access all objects
+of a specific ``Component`` or ``ComponentData`` type.
+The generated objects may be thought of as a "flattened" representation
+of the model, as they may be accessed without any knowledge of the model's
+block structure.
+These methods are very useful, but it is still challenging to use them
+to access specific components.
+Specifically, we often want to access "all components indexed by some set,"
+or "all component data at a particular index of this set."
+In addition, we often want to generate even the components in a block that
```
jsiirola, line 26: typo: remove "even"

doc/OnlineDocs/advanced_topics/flattener/index.rst
```diff
@@ -0,0 +1,44 @@
...
+utilities to generate all components indexed, explicitly or implicitly, by
+user-provided sets.
+
+**When we say "flatten a model," we mean "generate all components in the model,
+preserving all user-specified indexing sets."**
+
+Data structures
+---------------
+The components returned are either ``ComponentData`` objects, for components
+not indexed by any of the provided sets, and references-to-slices, for
```
jsiirola, line 39: "and" -> "or"

pyomo/dae/flatten.py
```diff
@@ -234,47 +239,108 @@ def slice_component_along_sets(
...
+        Sets that will be sliced
+
+    ctype: Subclass of Component
+        Type of components to generate
+
+    index_map: ComponentMap
+        Map from (some of) the specified sets to a "representative index"
+        to use when descending into subblocks. While this map does not need
+        to contain every set in the sliced sets, it must not contain any
+        sets that will not be sliced.
```
jsiirola, line 279: Missing documentation for `active` (mildly ironic, given the PR title... ;) )

jsiirola (approved):
This looks good.  Just a couple typos to resolve.

blnicho (approved):

pyomo/dae/flatten.py
Robbybp: I was probably avoiding actually having to write a succinct description of how this flag works. Thanks for adding @blnicho

Issue #2613, opened 2022-11-20
Warmstart broken for MOSEK>=10


jsiirola:
## Summary

It appears that warmstart is partially broken on MOSEK 10.  The example below was pulled from `pyomo/solvers/tests/models/MIQP_simple.py`.  While it worked in MOSEK 8 / 9, it is failing in MOSEK 10.

As the assertion is on the C side, this error is untrappable and kills the entire Python interpreter.

@Utkarsh-Detha, can you provide more insight on what may be happening?

### Steps to reproduce the issue

```python
from pyomo.environ import *

model = ConcreteModel()

model.a = Param(initialize=1.0)
model.x = Var(within=NonNegativeReals, initialize=1)
model.y = Var(within=Binary, initialize=1)

model.obj = Objective(expr=model.x**2 + 3.0*model.y**2)
model.c1 = Constraint(expr=model.a <= model.y)
model.c2 = Constraint(expr=(2.0, model.x/model.a - model.y, 10))

solver = SolverFactory('mosek_direct').solve(
    model, warmstart=True, load_solutions=False)
```

### Error Message

```
MOSEK fatal error stoptask
Version : 10.0.28
File    : src/mosek/mosekqo.c
Line    : 5104
Msg     : Assertion failed
MOSEK fatal error stopenv
Version : 10.0.28
Platform: Linux/64-X86
File    : src/mosek/mosekqo.c
Line    : 5104
Msg     : Assertion failed
Segmentation fault (core dumped)
```

### Information on your system

Pyomo version: 6.4.3.dev0
Python version: 3.7
Operating system: Linux
How Pyomo was installed (PyPI, conda, source): source
Solver (if applicable): MOSEK 10.0.27 and 10.0.28

Utkarsh-Detha:
I am looking into this and will get back as soon as I have an answer/fix. Thank you for bringing this up.

Utkarsh-Detha:
@jsiirola the newest release (the second point in the [release notes](https://docs.mosek.com/latest/releasenotes/bugfixes.html)) addresses this issue. The recent pull-request does _not_ fix this issue (as it was not caused by the interface), but does address something related to warm-start that I noticed while looking at it in the past few days.

PR #2647, opened 2022-12-02
Enabled warm-start flag for MOSEK10 and moved var initialization to _add_vars

Utkarsh-Detha:
## Summary/Motivation:
Issue #2613 was caused by a rare assert failure within MOSEK 10. That has since been fixed (please upgrade to **MOSEK 10.0.30**). 

But, while looking at the warmstart function in mosek_direct.py, I noticed that a parameter required to turn on initial solution utilization (only for v10 onwards) was missing. 

- Enabed `mosek.iparam.mio_construct_sol` in warm_start function in mosek_direct.py
- Moved the code that passes variable initialization values from the `_warm_start` function to the `_add_vars` function.

### Legal Acknowledgement


jsiirola (approved):
Looks good!

pyomo/solvers/plugins/solvers/mosek_direct.py
```diff
@@ -361,6 +362,12 @@ def _add_vars(self, var_seq):
         self._pyomo_var_to_solver_var_map.update(zip(var_seq, var_ids))
         self._solver_var_to_pyomo_var_map.update(zip(var_ids, var_seq))
         self._referenced_variables.update(zip(var_seq, [0]*len(var_seq)))
+        # Initialize variables
+        for m_v, p_v in zip(var_ids, var_seq):
+            if p_v.value is not None:
+                for sol_type in mosek.soltype.values:
+                    self._solver_model.putxxslice(
+                        sol_type, m_v, m_v + 1, [(p_v.value)])
```
jsiirola, line 370: Question: by moving this to `_add_vars()`, does this prevent the variable values from being updated between subsequent solves in the case of the persistent solver?

Utkarsh-Detha: The same `task` object instance will be used over subsequent solves (when using mosek_persistent). So, the variable values from the previous solve will already be in that particular task. Pyomo would not have to pass them again. We only need to tell mosek if it should use those values or ignore them.

blnicho: @Utkarsh-Detha, reminder that this PR is waiting on a response from you to the above comment from @jsiirola.

Utkarsh-Detha: @jsiirola then I suppose we just leave it as is. I will undo the changes concerning this.

(Sorry about the slow response, I have been on holiday)

jsiirola (approved):
This looks good!

mrmundt (approved):

PR #2659, opened 2022-12-06
Fix PyROS `UncertaintySet.bouding_problem` Method

shermanjasonaf:
## Summary/Motivation:
The method `pyros.UncertaintySet.bounding_model` may require access to the subordinate global optimizer passed to the PyROS solver, depending on the type of the uncertainty set passed to the solver. In particular, this requirement holds for the PyROS `IntersectionSet`.

## Changes proposed in this PR:
- Add `config` argument to `UncertaintySet.bounding_problem`.
- Add test for two-stage model instances with an uncertainty set of type `IntersectionSet`.

### Legal Acknowledgement

By contributing to this software project, I have read the [contribution guide](https://pyomo.readthedocs.io/en/stable/contribution_guide.html) and agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #2660, opened 2022-12-07
Add more judicious enforcement of PyROS Solver time limit

shermanjasonaf:
- Adjust subordinate subsolver time limit option prior to every PyROS subproblem (master feasibility, master, DR polishing, separation), based on the `time_limit` passed to PyROS and the time elapsed. This time limit adjustment method will be supported for `'gams'`, `'ipopt'`, and `'baron'`.
- Perform time elapsed vs time limit check after every master feasibility, master, DR polishing, and separation subproblem. Currently, this check is performed only after master and separation problems.

## Todo (after #2659 merged)
- [x] Add unit tests, and/or resolve issues with current test set due to changes. This should resolve failing test cases.
- [x] Modify `pyros.py` such that subsolvers are copied/cloned on argument parsing, so that state of solver(s) passed by user remain unaffected.
- [x] Update version number, changelog

### Legal Acknowledgement


blnicho (approved):

pyomo/contrib/pyros/pyros.py
```diff
@@ -79,13 +80,13 @@ class SolverResolvable(object):
...
         if obj is a string, return the Solver object for that solver name
-        if obj is a Solver object, return the Solver
+        if obj is a Solver object, return a copy of the Solver
         if obj is a list, and each element of list is solver resolvable, return list of solvers
         '''
         if isinstance(obj, str):
             return SolverFactory(obj.lower())
         elif callable(getattr(obj, "solve", None)):
-            return obj
+            return deepcopy(obj)
```
jsiirola, line 89: I am a little nervous about trusting deepcopy and solver objects.  The Pyomo side should be OK, but if the solver has python bindings, then who knows what might happen.  What is the motivation for making a copy of the solver?

pyomo/contrib/pyros/util.py
```diff
@@ -66,13 +68,54 @@ def get_main_elapsed_time(timing_data_obj):
...
+        #     For this reason, extra 30s is added to time remaining
+        #     for subsolver time limit
+        if options_key is not None:
+            time_remaining = (
+                config.time_limit - get_main_elapsed_time(timing_data_obj)
+            )
+
+            # ensure positive value assigned to avoid application error
+            solver.options[options_key] = max(30, 30 + time_remaining)
+
```
jsiirola, line 109: You need an else clause to at least emit a warning that the user option (`time_limit`) is being ignored.

jsiirola (changes requested):
Overall, this looks reasonable.  In addition to the question and change below, can you also add a test that exercises the time_limit option (and improves the code coverage for processing timeouts)?  That might be tricky (you may need to create a custom "mock solver" that does nothing but wait until the timeout and returns a prearranged state).

pyomo/contrib/pyros/pyros.py
shermanjasonaf: To preserve the original subsolver time limit setting prescribed by the user (since this setting may now be changed by PyROS in advance of solving a subproblem).

jsiirola: I see.  I think it might be safer to create a new solver using the factory.  Maybe something like:
```python
new_solver = SolverFactory(obj.name)
new_solver.options.set_value(obj.options)
return new_solver
```

shermanjasonaf: Understood. How should I handle solvers which do not have a `name` attribute (such as GAMS solver types)?

jsiirola: Eww.  You are right.  This is a mess: the 'name' is not necessarily correlated with the registration in the SolverFactory, and the old servers are not at all consistent with how they store / manage ephemeral options.  Maybe the better answer is to just use the solver the user gave us and remember / restore the timeout on exit (either normal or error / exception)?

pyomo/contrib/pyros/util.py
shermanjasonaf: Do you mean to say that the subsolver time limit will be ignored? In this case, the warning will be emitted in advance of every subproblem to be solved by the optimizer

jsiirola: Sure - but the user should know that: they specified a time limit, but the time limit is not going to be honored.

shermanjasonaf: Understood

pyomo/contrib/pyros/pyros.py
shermanjasonaf: See subsequent commits

jsiirola (approved):
This looks good.  I have two comments that are "nice-to-haves", but are not strictly required to merge.

pyomo/contrib/pyros/util.py
```diff
@@ -66,13 +68,155 @@ def get_main_elapsed_time(timing_data_obj):
...
+            if isinstance(solver.options["add_options"], list):
+                solver.options["add_options"].append(reslim_str)
+            else:
+                solver.options["add_options"] = [reslim_str]
+        else:
+            # determine name of option to adjust
+            if isinstance(solver, type(SolverFactory("baron"))):
+                options_key = "MaxTime"
+            elif isinstance(solver, type(SolverFactory("ipopt"))):
+                options_key = "max_cpu_time"
```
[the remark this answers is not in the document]
shermanjasonaf: Noted.

jsiirola (approved):

PR #2666, opened 2022-12-14
Warn when infeasibility tools will not log output

jsiirola:
## Summary/Motivation:
One of our developers was recently burned by using `log_infeasible_constraints()` and assuming that because they didn't see any output that the model was feasible.  In fact the model was infeasible; but, because the default Pyomo logging level is WARNING, the log messages were suppressed.

This PR is a quick fix to emit a warning when the infeasibility tools are called with a logger that will not output INFO level messages.

- Log a warning when the logger passed to `pyomo.util.infeasible.log_*()` will not emit INFO level messages

### Legal Acknowledgement

### DIFF
--- a/pyomo/util/infeasible.py
+++ b/pyomo/util/infeasible.py
@@ -41,7 +41,7 @@ def log_infeasible_constraints(
         logger.warning(
             'log_infeasible_constraints() called with a logger whose '
             'effective level is higher than logging.INFO: no output '
-            'will be logged reguardless of constraint feasibility'
+            'will be logged regardless of constraint feasibility'
         )
 
     # Iterate through all active constraints on the model
@@ -137,7 +137,7 @@ def log_infeasible_bounds(m, tol=1E-6, logger=logger):
         logger.warning(
             'log_infeasible_bounds() called with a logger whose '
             'effective level is higher than logging.INFO: no output '
-            'will be logged reguardless of bound feasibility'
+            'will be logged regardless of bound feasibility'
         )
 
     for var in m.component_data_objects(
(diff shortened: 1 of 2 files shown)


pyomo/util/infeasible.py
```diff
@@ -37,6 +37,13 @@ def log_infeasible_constraints(
         log_variables (bool): If true, prints the constraint variable names and values
 
     """
+    if logger.getEffectiveLevel() > logging.INFO:
+        logger.warning(
+            'log_infeasible_constraints() called with a logger whose '
+            'effective level is higher than logging.INFO: no output '
+            'will be logged reguardless of constraint feasibility'
```
blnicho, line 44: Repeated Typo: regardless

pyomo/util/tests/test_infeasible.py
```diff
@@ -51,6 +51,16 @@ def build_model(self):
     def test_log_infeasible_constraints(self):
         """Test for logging of infeasible constraints."""
         m = self.build_model()
+
+        with LoggingIntercept(None, 'pyomo.util.infeasible') as LOG:
+            log_infeasible_constraints(m)
+        self.assertEqual(
+            'log_infeasible_constraints() called with a logger whose '
+            'effective level is higher than logging.INFO: no output '
+            'will be logged reguardless of constraint feasibility',
```
blnicho, line 60: Repeated typo: regardless

pyomo/util/infeasible.py
blnicho: Should this warning include a note on how to change the logging level?
"This warning can be avoided by changing the Pyomo logging level: `<CODE TO CHANGE LOGGING LEVEL>`"

jsiirola: I don't think so.  There are too many ways to change the logging level (adding handlers, calling a standard setup, setting the logging level) to put something specific here.  If we start to see a bunch of questions about it, then we can add a reference to the online docs (in the [Common Warnings/Errors](https://pyomo.readthedocs.io/en/latest/errors.html#warnings) list) -- where we can have a longer and more detailed discussion.

blnicho (approved):

PR #2672, opened 2022-12-19
Convert subprocess timeout adjustment parameters to module attributes

### LINKED ISSUE #2671: Timeout on very short Optimization runs
Very short optimizations result in a timeout error in Pyomo. Actually, the optimization problem is sufficiently solved and usable. However, the Callback from CPLEX to pyomo takes too long, so pyomo throws an timeout error. The problem lies in this line of code in shellcmd.py. 
image

Here the max(1, 0.01 * self._timelimit) is taken, so the timeout for a 2 minute optimization is 1.19 seconds. This is sometimes not enough for the callback. So you could adjust this line to max(120, 0.01 * self._timelimit) and the problem would be solved... :) Thanks in advance

BenediktCosmo:
## Summary/Motivation:
The parameters that control how the subprocess adjusts the timeout to allow for subprocess termination are currently embedded constants that do not allow users to adjust them for their specific environment.  This PR converts them to module attributes that can be changed from user code before invoking the solver.

- convert timeout-related embedded constants to module attributes 

### Legal Acknowledgement

### DIFF
--- a/pyomo/opt/solver/shellcmd.py
+++ b/pyomo/opt/solver/shellcmd.py
@@ -32,6 +32,16 @@ from pyomo.opt.results import SolverStatus, SolverResults
 
 logger = logging.getLogger('pyomo.opt')
 
+# The minimum absolute time (in seconds) to add to the solver timeout
+# when setting the timeout for the solver subprocess.  This provides
+# time for a solver that timed out to clean up / report a solution
+# before we forcibly kill the subprocess.
+SUBPROCESS_TIMEOUT_ABS_ADJUST = 1
+# The additional time (relative to the user-specified timeout) to add to
+# the solver timeout when setting the timeout for the solver subprocess.
+# This provides time for a solver that timed out to clean up / report a
+# solution before we forcibly kill the subprocess.
+SUBPROCESS_TIMEOUT_REL_ADJUST = 0.01
 
 class SystemCallSolver(OptSolver):
     """ A generic command line solver """
@@ -312,7 +322,8 @@ class SystemCallSolver(OptSolver):
 
         timeout = self._timelimit
         if timeout is not None:
-            timeout += max(120, 0.01*self._timelimit)
+            timeout += max(SUBPROCESS_TIMEOUT_ABS_ADJUST,
+                           SUBPROCESS_TIMEOUT_REL_ADJUST*self._timelimit)
 
         ostreams = [StringIO()]
         if self._tee:
--- a/pyomo/opt/solver/shellcmd.py
+++ b/pyomo/opt/solver/shellcmd.py
@@ -312,7 +312,7 @@ class SystemCallSolver(OptSolver):
 
         timeout = self._timelimit
         if timeout is not None:
-            timeout += max(1, 0.01*self._timelimit)
+            timeout += max(120, 0.01*self._timelimit)
 
         ostreams = [StringIO()]
         if self._tee:


jsiirola (changes requested):
Thank you for the issue and PR!

I am afraid that I don't completely understand what problem this PR is attempting to solve.  The changed line of code is *adding* to the timeout specified by the user to account for the time to startup / shutdown the underlying process.  Unfortunately, the 2 minute minimum seems just as arbitrary (perhaps more so) than the previous 1 second buffer.  I feel that we will get more complaints that Pyomo can take more than 2 minutes longer than the timeout specified by the user.

Before we make any changes, I would like to see a more detailed explanation and complete (reproducible) example demonstrating the problem and quantifying what the minimal change needs to be.

BenediktCosmo:
The problem occurs not every time and depends on CPLEXs answer time. Most of the time it works, but sometimes it comes to an timeout through pyomo. But I will try to explain it in more detail. 
The timeout is set to max(1, 0.01 * self._timelimit). Thus if you solve your optimization problem with a timelimit of several ours, for example 10 ours, self._timelimit will become 10 * 60 * 60 = 36000 and max(1, 0.01 * self._timelimit) = 360 seconds. These 360 seconds will be added to the timeout of the underlying process in pyomo. But it is only the maximum time to wait. If the solver returns an answer more earlier, pyomo will shutdown the underlying process as well earlier. Thus for optimization with a long solving time there is no problem. 
In my case I solve an optimization problem with a timelimit of 2 minutes, Thus max(1, 0.01 * self._timelimit) = 1.2 seconds. Thus after CPLEX has solved the problem within 2 minutes, pyomo will wait for only 1.2 seconds until pyomo expects a return of CPLEX. But sometimes CPLEX returns a little bit later and than a timeout Error occurs. 
May be you are right, a general minimum of 120 seconds is a litte bit to long and a minimum waiting time of 10 seconds is sufficient. But it is not really taking into account, because it is only the maximum time, for that pyomo will wait for an answer.

emma58:
@BenediktCosmo, thanks for the explanation. Could you add to your issue report in #2671 both information about your system and the Pyomo code that shows this timeout (even intermittently)? While you are likely right that 1 second can be too short, this change proposes changing an arbitrary number to a different arbitrary number. It would be better if we could base the new timeout limit on an understanding of what is going on (and perhaps also use this information to add a test).

BenediktCosmo:
@emma58 
https://github.com/Pyomo/pyomo/issues/2671#issuecomment-1375564408
I add the error message and my system information to the Issue #2671 
I hope that will help to solve this issue and change the minimal timeout. It is currently not possible to publish the code because I use it in an business context and have to protect trade secrets. If minimal code is required, I would have to implement it first.

jsiirola:
@BenediktCosmo,Thanks!  It looks like you are experiencing the problem when running within a virtual machine?  

I understand you can't share your model, but can you perform a search over the minimum timeout (`max(i, 0.01*self._timelimit)` for i in {1, 1.5, 2, 2.5, 3, ...}) and report back the number failures for each value?  That will really help identify what an appropriate minimal adjustment of the timeout should be.

There is a concern with making this value "too long", as some solvers do not handle timeouts internally, so the only timeout is the one enforced by subprocess.  Users tend to complain when they set a timeout and then the solve function call takes (significantly) longer than the timeout.

BenediktCosmo:
Yes it is right. I am working within a virtual machine in the Azure Cloud. The error occurs regardless of the scaling of the machine.
I'm very sorry, but I don't have the time at the moment to do an analysis with different timeout values. A good alternative would certainly be to parameterize this timeout value, i.e. max(i, 0.01 * self._timelimit), so that the pyomo user can specify i within a parameter. You could keep 1 second as the default value, but the user would be able modifying this value.

jsiirola:
I was just thinking about the same solution.  For the moment, what to you think about adding a module-level `SUBPROCESS_MIN_TIMEOUT_ADJUST` attribute (or something similarly descriptive) to `shellcmd.py` -- with the intent to eventually migrate it into a proper Pyomo configuration option when we get a chance to rework that data structure.

Do you want to make that change to your PR?   (Otherwise, if we can post to it, I am happy to push that up & review).

BenediktCosmo:
That sounds good. If you would like, I would be happy if you push up this attribute. However, I still don't quite understand how the user can change this value. Is it like a environment variable then? But please don't hesitate to push your idea into PR.

jsiirola:
By making it a module-level attribute, users could change it with, e.g.:
```python
import pyomo.opt.solver.shellcmd
pyomo.opt.solver.shellcmd.SUBPROCESS_MIN_TIMEOUT_ADJUST = 10
```

It's not as nice as a system-level configuration (because the user needs to know a pretty deep / obscure module path), but it is "do-able".  I'll go ahead and push the change in a little bit.

jsiirola (approved):

BenediktCosmo:
Ah, perfect. I think that's a good solution for now. Thank you for your commitment. This PR will help me a lot! 
But one last Question. When will the Pyomo package be updated?

blnicho (approved):

PR #2680, opened 2023-01-03
Manage Gurobi environments in GurobiDirect

### LINKED ISSUE #2408: Manage Gurobi environments explicitly in GurobiDirect
## Summary

It would be useful to add some capability to manage Gurobi environments (`gurobipy.Env` objects) in the `GurobiDirect` solver.

I'm happy to submit a PR for the changes described below, but could use some guidance as to the best option that fits with pyomo's style & API.

### Rationale

Some Gurobi customers using pyomo to build their models have run into issues with locking up shared resources (compute servers, floating licenses, etc). When using gurobipy directly the solution is to use context managers to explicitly free those resources once they are done with. This is tricky to achieve in pyomo since Gurobi environments are not used explicitly (the global 'default' environment is used).

Additionally, some parameters cannot be set without explicitly creating Gurobi environments e.g. ISV keys (see #2401), or [MemLimit](https://www.gurobi.com/documentation/9.5/refman/memlimit.html) added in a recent release.

### Description

Several possible solutions:

- Explicitly create a `gurobipy.Env` in `__init__` or `__enter__` in `GurobiDirect`
- Provide a `.close()` method in `GurobiDirect` to free all gurobi resources
- Allow users to create and manage an environment themselves using `gurobipy` and pass it as a keyword argument to `SolverFactory`

Some technical points to fix:

- `GurobiDirect.available()` checks for a license, but if it encounters an error (e.g. the license is temporarily in use) then it tracks this outcome globally and the solver is completely locked. `available()` might be better off just checking if `gurobipy` can be imported and leave the license check until a model needs to be created.
- `_solver_model` should be explicitly freed if overwritten

### Additional information

I've written some [test cases](https://github.com/simonbowly/pyomo/blob/gurobi-environments/pyomo/solvers/tests/checks/test_gurobi_environments.py) to try to make the above points clear.

simonbowly:
## Summary/Motivation:

There are currently several limitations in the GurobiDirect interface:

1. Some Gurobi parameters cannot be used with the current approach, as they require explicitly creating a gurobipy `Env` object. This includes connection parameters for compute servers, token servers, and instant cloud (can be worked around via a license file, but this isn't always an ideal approach) and special parameters such as MemLimit.
2. There is no clean way to close Gurobi models and environments, which leaves license tokens in-use and compute server connections open longer than a user need them.
3. A user cannot retry acquiring a Gurobi license token (important in shared license environments) since the GurobiDirect class caches errors in global state.

Introduces a constructor flag `manage_env` for GurobiDirect (defaults to False), and two public methods `.close()` and `.close_global()`. If users set `manage_env=True`:

- GurobiDirect explicitly creates a Gurobi environment bound to the solver instance. This enables Gurobi resources to be properly freed by the solver object:
```python
with SolverFactory('gurobi', solver_io='python', manage_env=True) as opt:
    opt.solve(model)
# All Gurobi models and environments are freed
```
- Calling `.close()` achieves the same result as the context manager:
```python
opt = SolverFactory('gurobi', solver_io='python', manage_env=True)
try:
    opt.solve(model)
finally:
    opt.close()
# All Gurobi models and environments are freed
```
- Internally, solver options are passed to the Env constructor (instead of the Model, as is currently done) to allow environment-level connection parameters to be used:
```python
options = {
    "CSManager": "<url>",
    "CSAPIAccessID": "<access-id>",
    "CSAPISecret": "<api-key>",
}
with SolverFactory('gurobi', solver_io='python', manage_env=True, options=options) as opt:
    opt.solve(model)  # Solved on compute server
# Compute server connection terminated
```

If `manage_env=False` (the default) is set, then users will get the old behaviour, which uses the Gurobi default/global environment. There are some minor changes:

- Calling `.close()`, or exiting the context properly disposes of all models created by the solver
```python
with SolverFactory('gurobi', solver_io='python') as opt:
    opt.solve(model)
# Gurobi models created by `opt` are freed; the default/global Gurobi environment is still active
```
- Calling `.close_global()` disposes of models created by the solver, and disposes the Gurobi default environment. This will free all Gurobi resources *assuming* the user did not create any other models (e.g. via another GurobiDirect object with `manage_env=False`):
```python
opt = SolverFactory('gurobi', solver_io='python')
try:
    opt.solve(model)
finally:
    opt.close_global()
# Gurobi models created by `opt` are freed, the default/global Gurobi environment is closed
```

Finally, the `available()` call no longer stores errors globally and repeats them back if users retry the check. So users can do the following to queue requests if they are using a shared license (regardless of whether manage_env is set to True or False):

```python
with SolverFactory('gurobi', solver_io='python') as opt:
    while not available(exception_flag=False):
        time.sleep(1)
    opt.solve(model)
```

### Legal Acknowledgement

6. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.


simonbowly:
Thanks @mrmundt! Are the current jenkins test failures relevant? They seem to be only codecov related.

Just one note I should add re: the tests. Some of the tests I added are really integration tests that rely on Gurobi single-use license behaviour, so they are skipped in CI. I can look at adding a few more unit tests if the changes seem reasonable.

jsiirola:
@simonbowly: yes: the Jenkins failure is legitimate:
```
    def test_set_environment_parameters(self):
        # Solver options should handle parameters which must be set before the
        # environment is started (i.e. connection params, memory limits). This
        # can only work with a managed env.
    
        with SolverFactory(
            "gurobi_direct",
            manage_env=True,
            options={"ComputeServer": "/url/to/server"},
        ) as opt:
            # Check that the error comes from an attempted connection, not from setting
            # the parameter after the environment is started.
            with self.assertRaisesRegex(ApplicationError, "Could not resolve host"):
>               opt.solve(self.model)
E               AssertionError: "Could not resolve host" does not match "Could not create Model for <class 'pyomo.solvers.plugins.solvers.gurobi_direct.GurobiDirect'> solver plugin - gurobi message=Could not create Model - gurobi message=Unrecognized response (status 404, POST http:///url/to/server/api/v1/cluster/jobs, 'Cannot resolve hostname 'url' in explicit-proxy re')
E               
E               Set parameter TokenServer to value "*****"
E               Set parameter ComputeServer to value "/url/to/server"
E               "

pyomo/solvers/tests/checks/test_gurobi_direct.py:73: AssertionError
```

simonbowly:
Thanks @jsiirola. I pushed a fix to that test and rebased.

pyomo/solvers/tests/checks/test_gurobi_direct.py
```diff
@@ -0,0 +1,257 @@
...
+            with SolverFactory("gurobi_direct") as opt:
+                opt.available()
+                opt.available()
+                PatchModel.assert_called_once_with()
+            patch_dispose.assert_not_called()
+
+
+@unittest.skipIf(not gurobipy_available, "gurobipy is not available")
+@unittest.skipIf(not single_use_license(), reason="test needs a single use license")
+class GurobiSingleUseTests(GurobiBase):
```
michaelbynum, line 171: I don't think these tests will ever get run with our current CI setup.

michaelbynum (commented):
Sorry for the delay. For the most part, this looks great. I've mentioned one potential issue below. I'll check on the failing test.

michaelbynum:
The Jenkins tests did not run due to an issue downloading GSL. I'll try to get the tests restarted.

pyomo/solvers/tests/checks/test_gurobi_direct.py
simonbowly: True. I've now added additional tests in `GurobiEnvironmentTests` which cover essentially the same behaviour. I'd like to leave `GurobiSingleUseTests` in as they're useful for local development/testing.

simonbowly:
@michaelbynum sorry to leave this hanging for so long, thanks for merging changes in the meantime!

I've added additional tests in [7c16c62](https://github.com/Pyomo/pyomo/pull/2680/commits/7c16c629555b353245f893317bc114cac403119c) which cover similar behaviour to the single-use tests but via patching gurobipy functions.

One last thing I'm not sure on - where is the appropriate place to document the new option? There aren't solver-specific pages in the docs, is the docstring of `GurobiDirect` a suitable place?

mrmundt (approved):

simonbowly:
I've (hopefully) fixed the failed tests on windows with 325d2ca. Can you please re-run?

simonbowly:
I took one more pass to hopefully sort out that failing conda test 🤞

michaelbynum:
> @michaelbynum sorry to leave this hanging for so long, thanks for merging changes in the meantime!
> 
> I've added additional tests in [7c16c62](https://github.com/Pyomo/pyomo/pull/2680/commits/7c16c629555b353245f893317bc114cac403119c) which cover similar behaviour to the single-use tests but via patching gurobipy functions.
> 
> One last thing I'm not sure on - where is the appropriate place to document the new option? There aren't solver-specific pages in the docs, is the docstring of `GurobiDirect` a suitable place?

I think the docstring of `GurobiDirect` is a good place.

michaelbynum:
@simonbowly - One more question. Would it be better to allow users to pass in an instance of `gurobipy.Env` to the constructor rather than passing a flag to create one under the hood? Would that be more flexible at all? I'm happy either way, but I wanted to throw the idea out there.

simonbowly:
> Would it be better to allow users to pass in an instance of gurobipy.Env to the constructor rather than passing a flag to create one under the hood? Would that be more flexible at all?

I did think about that initially. The distinction would be:

```python
import gurobipy as gp

with gp.Env(params={...}) as env, pyo.SolverFactory("gurobi_direct", env=env) as opt:
    opt.solve(model)
```

vs.

```python
with pyo.SolverFactory("gurobi_direct", options={...}, manage_env=True) as opt:
    opt.solve(model)
```

where `params` in the first case and `options` in the second case are identical parameter dictionaries. I went with the second approach, reasoning that:

1. Interacting directly with the underlying solver library didn't seem in keeping with pyomo's design, and the dual context is a little more unwieldy.
2. The first approach doesn't gain anything: the `Env` object accepts the same parameter dictionary and provides the same closable context.

Unless there is a use case for creating one `Env` and sharing it across solvers? The canonical thing to do there though seems to be to re-use the solver for multiple models.

simonbowly:
Added documentation with 317b8c5. That's everything I wanted to add, please review when time allows :-)

mrmundt (approved):

simonbowly:
@blnicho I see this didn't make it into the release, did you want me to make any changes?

blnicho:
@simonbowly we don't have any change requests at this time. We ran out of time to get a second review on this and needed to move forward with the release to meet a deadline for a downstream project.

michaelbynum (approved):

michaelbynum:
There was a pypy-3.9 test failure. I restarted that job. If it passes, we will merge.

Sorry for the delay, @simonbowly. It is my fault.

simonbowly:
No problem at all! I really appreciate the time you all spent discussing & reviewing this PR, thanks again!

andyvk85:
Very great job folks! Thank you!

PR #2686, opened 2023-01-10
GDPOpt: Switching a LBB test to use Gurobi as MINLP solver

emma58:
## Summary/Motivation:

BARON 23.1.5 is returning a wrong answer during `gdpopt.lbb`, causing it to converge to the wrong solution. This switches the test to use Gurobi for now because it's quadratic anyway...

### Legal Acknowledgement

By contributing to this software project, I have read the [contribution guide](https://pyomo.readthedocs.io/en/stable/contribution_guide.html) and agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #2687, opened 2023-01-10
Adds GDP-to-MIP transformation base class

emma58:
## Summary/Motivation:

This PR adds a base class to transformations from GDP to MIP, in the first known attempt at removing duplicated code in `pyomo.gdp` this decade. It moves the bigm, hull, and multiple bigm transformations onto this base class, and adds a bigm mix-in class for common functionality between the two bigm variants. It does not yet touch between steps or cutting planes in an attempt to get @jsiirola to review it--I mean in order to keep it from getting humongous. Overall, though, it has several effects that will help me sleep better at night:

1. Multiple bigm no longer instantiates the bigm transformation in order to call private methods on it while tricking it about its state. (The mix-in class solves this issue.)
2. Not that I checked, but this should improve performance in `hull` because it propagates the improvements from #2605 to the base class.
3. Did I mention this removes duplicated code?

NOTE: This contains the change in #2686

- Adds `GDP_to_MIP_Transformation` and `_BigM_MixIn` classes, and has `hull`, `bigm`, and `mbigm` inherit from the appropriate ones.
- This implicitly updates `hull` with the changes from #2605. Most radically, this undoes the neurotic naming of transformed constraints, and instead maintains a list of transformed constraints on each transformed Disjunct. This is what causes all the changes to hull tests that previously relied on how transformed constraints were named, and it is also responsible for the changes in the job shop baselines. Last, it is also what broke the two tests in FME that always break when hull's transformed constraint structure changes.

### Legal Acknowledgement

5. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.


pyomo/gdp/plugins/bigm_mixin.py
```diff
@@ -0,0 +1,228 @@
...
+    # on whatever components you want, we won't pick them up.
+    suffix_list = []
+
+    # go searching above block in the tree, stop when we hit stopping_block
+    # (This is so that we can search on each Disjunct once, but get any
+    # information between a constraint and its Disjunct while transforming
+    # the constraint).
+    while block is not None:
+        bigm = block.component('BigM')
+        if type(bigm) is Suffix:
```
bernalde, line 53: Is this the right moment to unify the notation between `M`, `m`, `bigm`, `bigM`, `big_M`, ... ?

pyomo/gdp/plugins/hull.py
```diff
@@ -924,25 +758,27 @@ def _transform_constraint(self, obj, disjunct, var_substitute_map,
...
 
                 if obj.is_indexed():
-                    newConstraint.add((i, 'ub'), newConsExpr)
+                    newConstraint.add((name, i, 'ub'), newConsExpr)
                     # map (have to account for fact we might have created list
                     # above
                     transformed = constraintMap['transformedConstraints'].get(c)
                     if transformed is not None:
-                        transformed.append(newConstraint[i,'ub'])
+                        transformed.append(newConstraint[name, i, 'ub'])
```
bernalde, line 766: Codecov is crying that this line is not covered by any test, but I think it's alright.

pyomo/gdp/tests/common_tests.py
```diff
@@ -1438,28 +1438,6 @@ def check_deactivated_disjunct_leaves_nested_disjunct_active(self,
     self.assertTrue(m.d1.d4.active)
     self.assertFalse(m.d1.d4.indicator_var.fixed)
 
-def check_mappings_between_disjunctions_and_xors(self, transformation):
```
bernalde, line 1441: Is this test irrelevant now? Should we not test the mappings now that there is a unified interface for the transformations?

pyomo/gdp/tests/test_hull.py
```diff
@@ -108,16 +108,16 @@ def check_furman_et_al_denominator(self, expr, ind_var):
...
         disjBlock = m._pyomo_gdp_hull_reformulation.relaxedDisjuncts
 
         # the only constraint on the first block is the non-linear one
-        disj1c = disjBlock[0].component("d[0].c")
-        self.assertIsInstance(disj1c, Constraint)
+        disj1c = hull.get_transformed_constraints(m.d[0].c)
         # we only have an upper bound
         self.assertEqual(len(disj1c), 1)
-        cons = disj1c['ub']
+        cons = disj1c[0]
```
bernalde, line 120: So the order of the constraints is `0`: `ub`, `1`: `lb`? Would this break the older code if we were looking for the transformed upper and lower bounds constraints using the `ub`/`lb` keys?

pyomo/gdp/tests/test_hull.py
```diff
@@ -1197,7 +1232,27 @@ def test_deactivated_disjunct_leaves_nested_disjuncts_active(self):
...
+            (m.disjunct[1].innerdisjunction[0],
+             transBlock.innerdisjunction_xor[0]),
+            (m.simpledisjunct.innerdisjunction,
+             transBlock.innerdisjunction_xor_4)
+        ]
+
+        # check disjunction mappings
+        for disjunction, xor in disjunctionPairs:
+            self.assertIs(disjunction.algebraic_constraint, xor)
+            self.assertIs(transform.get_src_disjunction(xor), disjunction)
```
bernalde, line 1255: This addresses my previous comment. Awesome!

pyomo/gdp/plugins/bigm_mixin.py
```diff
@@ -0,0 +1,228 @@
...
+                raise
+    if len(M) != 2:
+        constraint_name = constraint.name
+        if disjunct is not None:
+            constraint_name += " relative to Disjunct %s" % disjunct.name
+        raise GDP_Error("Big-M %s for constraint %s is not of "
+                        "length two. "
+                        "Expected either a single value or "
+                        "tuple or list of length two for M."
+                        % (str(M), constraint.name))
```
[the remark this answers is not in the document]
emma58: I'd rather keep the error message terse, but I agree it would be a good idea to provide a link to documentation about providing M's for two-sided inequalities.

pyomo/gdp/plugins/bigm_mixin.py
emma58: Are you referring to the name of the Suffix? Or just the variable naming within the code?

pyomo/gdp/tests/common_tests.py
emma58: Yeah, I think this is the one you answered below. It's just way simpler. I don't want the test to rely on how I name things.

pyomo/gdp/plugins/bigm_mixin.py
bernalde: I refer to the variable naming within the code to try to make it more uniform across it. Some functions are for `m` others for `bigM`

pyomo/gdp/tests/test_hull.py
bernalde: I think that the performance improvement justifies the breaking of older code. As long as documentation reports this and if from the error it is easy to know when the actual syntax is supposed to be I fully support these changes.

pyomo/gdp/plugins/bigm_mixin.py
bernalde: Just to emphasize what I was meaning, the next two functions in this file alone are called `_warn_for_unused_bigM_args` `_get_bigm_arg_list` and `_process_M_value`

emma58: Yeah, I see your point. I've used `M` in the public API, so we should stick with that mostly. I like the idea of not using `bigm` since that overload the name of the transformation. I switched the ones that are specifically referring to args and Suffixes to `bigM` and tried to leave everything else as `M`.

jsiirola (commented):
Overall, this looks really good.  One question about naming / structure (it is a bit convoluted that `apply_to()` calls `_apply_to()` calls `_apply_to_impl`.  I think we should reconsider this structure (or at leas the naming) to improve the clarity of what is going on.

pyomo/contrib/gdpopt/tests/test_gdpopt.py
```diff
@@ -1597,6 +1597,9 @@ def test_no_default_algorithm(self):
             opt.solve(m, algorithm='LBB', tee=True, mip_solver=mip_solver,
                       nlp_solver=nlp_solver, minlp_solver='gurobi')
         self.assertIn('using LBB algorithm', buf.getvalue())
+        ## DEBUG
+        opt.solve(m, algorithm='LBB', tee=True, mip_solver=mip_solver,
+                  nlp_solver=nlp_solver, minlp_solver='gurobi')
```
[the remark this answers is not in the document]
emma58: Whoops, yes!

pyomo/gdp/plugins/bigm.py
```diff
@@ -133,136 +130,38 @@ class BigM_Transformation(Transformation):
...
-        self._generate_debug_messages = is_debug_set(logger)
         self.used_args = ComponentMap() # If everything was sure to go well,
                                         # this could be a dictionary. But if
                                         # someone messes up and gives us a Var
                                         # as a key in bigMargs, I need the error
                                         # not to be when I try to put it into
                                         # this map!
         try:
-            self._apply_to_impl(instance, **kwds)
+            super()._apply_to(instance, **kwds)
```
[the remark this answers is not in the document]
emma58: OK, I think I made this better. Now the base class doesn't implement `_apply_to` or `_apply_to_impl` it just has a helper method for processing arguments and another for restoring state (that the subclasses can call in the `finally` block).

emma58:
@jsiirola, this is ready for you to look at again. (There's still a little bit of method overriding weirdness in the `add_transformation_block` logic, but that one (I think) makes more sense because the base class implementation is a vanilla implementation of the same thing. The ones that call it embellish it.)

jsiirola (approved):

PR #2691, opened 2023-01-11
Rework handling of 'dimensionless' units in Pyomo

### LINKED ISSUE #1570: Setting var .value with units does not behave as expected
Pyomo allows a user to set the .value property on a var where the rhs contains an expression with units. This either needs to throw an error or do the conversion, but not silently allow this with no conversion. This needs to be investigated - I believe that this will just ignore the units now (which is problematic).

Need to decide if we will do implicit conversion or throw an error (and require v.set_unitted_value(...)) instead. @andrewlee94 favors the explicit use of something set_unitted_value.

jsiirola:
## Summary/Motivation:
@andrewlee94 reported strangeness with units where the following did not raise an error:
```python
m = ConcreteModel()
m.v = Var(units=units.dimensionless)
m.v.set_value(42*units.s)
```

This was due to `get_units()` returning `None` instead of `dimensionless`.  This caused Pyomo to not be able to distinguish between `Var()`, and `Var(units=units.dimensionless)`.  This PR changes the current behavior so that `get_units()` will *always* return a `_PyomoUnit`.

In addition, this PR reworks the `PintUnitExtractionVisitor` to make it slightly more performant.

While much of #1570 has already been fixed on `main`, this PR fixes the last known issues.

- `dimentionless` is returned from `get_units()` instead of `None`
- efficiency improvements in the `PintUnitExtractionVisitor`
- cleaned up some long lines and string formatting

### Legal Acknowledgement


pyomo/core/base/units_container.py
```diff
@@ -847,39 +855,64 @@ def _get_unit_sqrt(self, node, child_units):
         'floor': _get_unit_for_single_child
     }
 
+    def initializeWalker(self, expr):
+        # Refresh the cached dimensionless (in case the underlying pint
+        # registry was either changed or had not been set when the
+        # PyomoUnitsContainer was originally created.
```
mrmundt, line 861: Very small typo - no close parenthesis.

mrmundt (approved):
I don't know this code well enough to have a good idea of whether this is an efficient/good solution, but for what I do know, it looks fine to me (minus the smallest of typos).

blnicho (approved):

PR #2740, opened 2023-02-14
Add true deprecation warning to `pyomo` command

### LINKED ISSUE #160: Deprecate the old 'pyomo' command syntax
In earlier releases, we supported commands like
`  pyomo --solver=glpk foo.py bar.dat`
And we moved to a nested pyomo command, which uses the following syntax:
`  pyomo solve --solver=glpk foo.py bar.dat`

However, to enable graceful transition for users, we added a hack to transition to the old syntax if it looks like the user is not specifying a valid subcommand.  However, this leads to confusing error messages, and I think it's time to deprecate that hack and start throwing warnings that this hack will be removed in an upcoming release.

mrmundt:
## Summary/Motivation:
This PR adds a true deprecation warning to the `pyomo` command.

- Print deprecation warning if `pyomo` is called without a subcommand

### Legal Acknowledgement


blnicho (approved):

Robbybp (approved):
This just changes how a warning is displayed, and even has a test

jsiirola (approved):

PR #2760, opened 2023-03-02
Refactor CyIpopt interface to subclass `cyipopt.Problem`

Robbybp:
## Summary/Motivation:
In https://github.com/mechmotum/cyipopt/pull/182, I am working on an interface to let us call `GetIpoptCurrentIterate` and `GetIpoptCurrentViolations` during an intermediate callback. The non-hack way to do this without augmenting CyIpopt's callback signature is to add methods on `cyipopt.Problem`. This means that our callback in Pyomo will need to know about the `cyipopt.Problem` object, which can be achieved by having our "interface object" subclass `cyipopt.Problem` rather than passing our interface object to `cyipopt.Problem` when it is constructed.

This is a very small change in terms of lines of code, as `cyipopt.Problem` will automatically look for methods on `self`, so we don't need a new wrapper class, if it is not given a `problem_obj` (our CyIpoptProblemInterface) during construction.

Accidentally built this on top of #2759...

## Changes
- Update CyIpoptProblemInterface to subclass `cyipopt.Problem`
- Call `super().__init__` to initialize the `cyipopt.Problem` (if we forget this on any derived class, we segfault!)
- Use the CyIpoptProblemInterface instead of creating a new `cyipopt.Problem` during solve methods

### Legal Acknowledgement


pyomo/contrib/pynumero/algorithms/solvers/cyipopt_solver.py
```diff
@@ -148,7 +148,7 @@ def _cyipopt_importer():
     'Internal_Error': TerminationCondition.internalSolverError,
 }
 
-class CyIpoptProblemInterface(object, metaclass=abc.ABCMeta):
+class CyIpoptProblemInterface(cyipopt.Problem, metaclass=abc.ABCMeta):
```
Robbybp, line 151: This fails if cyipopt is not available. We could get around this with
```
cyipopt_Problem = cyipopt.Problem if cyipopt_available else object
```
at the top of the file, but then we get a confusing error message when we try to call `CyIpoptProblemInterface.solve`. Maybe do this but raise an error in `__init_subclass__` if we're not an instance of `cyipopt.Problem`?

jsiirola: What about using `cyipopt_Problem` as you suggest, then raising a sensible exception in `__init__` if `not cyipopt_available`?

Robbybp: Not bad, we just have to remember to call the right combination of `super().__init__`s in subclasses. But I could move the `cyipopt.Problem.__init__` call into `CyIpoptProblemInterface.__init__`, so subclass developers only have to remember to call one `super().__init__`.

Robbybp:
Test failures (on windows and linux/conda) are `'cyipopt' : Unexpected module found in 5 slowest-loading TPL modules`, so I added cyipopt to the list of potentially slow-loading modules in `test_tpl_import_time`. Now, locally, I get failures like this:
```
    self.assertLess(tpl_time / total, 0.65)
AssertionError: 0.7089268883802343 not less than 0.65
```
I'm a little hesitant to adjust this hand-coded tolerance when I don't know why import time has suddenly increased. Maybe
```python
cyipopt_Problem = cyipopt.Problem if cyipopt_available else object
```
is taking a ton of time during import?

pyomo/environ/tests/test_environ.py
```diff
@@ -151,6 +152,8 @@ def test_tpl_import_time(self):
         ref.add('pyutilib')
         if numpy_available:
             ref.add('numpy')
+        if cyipopt_available:
+            ref.add('cyipopt')
```
jsiirola, line 156: We don't want to add cyipopt to this list: this is indicating that cyipopt is being unconditionally imported by `pyomo.environ`.  We should re-think the imports so that we don't have to update this list.

Robbybp: Is this because `PyomoCyIpoptSolver` is registered as a plugin so it can be accessed via `pyo.SolverFactory("cyipopt")`?

jsiirola: Yes: `pyomo.environ` imports `pyomo.contrib.pynumero.plugins` which imports `pyomo.contrib.pynumero.algorithms.solers.cyipopt_solver`

Robbybp: I tried using a deferred import (via `attempt_import) for `pyomo.contrib.pynumero.algorithms.solvers.cyipopt_solver` in `pynumero.plugins`, but that didn't change anything, nor did moving `from .algorithms.solvers.cyipopt_solver import PyomoCyIpoptSolver` to within the `load` function of `pynumero.plugins`.

Do you have any ideas?

Robbybp:
> * split the ProblemInterface classes into a separate module from the solver interfaces (maybe `.pynumero.interfaces.cyipopt_problem`?)
> * Then, the solver interface can `attempt_import` the problem interface module, e.g.:
>   ```python
>    cyipopt_problem, _ = attempt_import('pyomo.contrib.pynumero.interfaces.cyipopt_problem')
>   ```
> * then the `PyomoCyIpoptSolver` can just `problem = cyipopt_problem.CyIpoptNLP(...`

Thanks for the suggestion, this works for me locally.

pyomo/contrib/pynumero/algorithms/solvers/cyipopt_solver.py
```diff
@@ -9,61 +9,58 @@
...
+relocated_module_attribute(
+    "CyIpoptProblemInterface",
+    "pyomo.contrib.pynumero.interfaces.cyipopt_interface.CyIpoptProblemInterface",
+    "6.5.1.dev0",
+)
+relocated_module_attribute(
+    "CyIpoptNLP",
+    "pyomo.contrib.pynumero.interfaces.cyipopt_interface.CyIpoptNLP",
+    "6.5.1.dev0",
+)
```
Robbybp, line 62: I commonly import `cyipopt_available` from this module, so I added these relocation warnings so these imports will still be available. This seems not to increase import time, judging by test_environ.py still passing locally.

Robbybp:
```
        solver = pyo.SolverFactory('cyipopt')
        solver.config.options = {'hessian_approximation':'limited-memory',
                                 'nlp_scaling_method': 'user-scaling',
                                 'output_file': '_cyipopt-external-greybox-scaling.log',
                                 'file_print_level':10,
                                 'max_iter': 0}
        status = solver.solve(m, tee=False)
    
        with open('_cyipopt-external-greybox-scaling.log', 'r') as fd:
            solver_trace = fd.read()
>       os.remove('_cyipopt-external-greybox-scaling.log')
E       PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: '_cyipopt-external-greybox-scaling.log'
```
@jsiirola Actually, these are the options sent to Ipopt, so this file is created by Ipopt and probably not closed until the `IpoptProblem` (or whatever the name is) C struct is freed, which is whenever `CyIpoptNLP.__dealloc__` is called, which I assume is when it gets garbage collected. This may be sufficient reason to implement a workaround in CyIpopt to avoid needing to subclass Problem to access the `get_current_iterate` functionality.

I just pushed a commit with `gc.collect()` before these calls to `os.remove` to test this hypothesis.

jsiirola:
@Robbybp, I think we can resolve the Windows test failures by calling `cyipopt.Problem.close()` before attempting to delete the log file that cyipopt generated.  The catch is that the `Problem` is private from the point of view of the tests (i.e., `_problem`).  We could just access the private attribute in the test (i.e., `solver._problem.close()`), or we could make a more public API for releasing the cyipopt object (possibilities include promoting `close()`, or making a context manager that calls `close()` on `__exit__()`).

Thoughts?

Robbybp:
@jsiirola Just pushed a commit that calls `close()` before attempting to remove the file. If this works, I'll work on a public API.

Robbybp:
I added `Problem.close()` to the end of `PyomoCyIpoptSolver.solve`, but there is still the question of what to do about `CyIpoptSolver`, which accepts the `CyIpoptProblemInterface` as an argument. Since the user provides the problem interface, calling its `close` method, as in the following, is a public API. But it's still non-obvious that you have to do this (and even worse that it's only required on Windows). We could put `problem_interface.close()` into the `CyIpoptSolver.solve` method, but this might have reprocussions for performing multiple solves with the same solver/interface objects...
```python
        solver = CyIpoptSolver(cyipopt_problem, options=options)
        x, info = solver.solve(tee=False)

        with open('_cyipopt-pyomo-ext-scaling.log', 'r') as fd: 
            solver_trace = fd.read()
        cyipopt_problem.close()
        os.remove('_cyipopt-pyomo-ext-scaling.log')
```

Robbybp:
Highlights:
- Updated `CyIpoptProblemInterface` to subclass `cyipopt.Problem`
- Added a `CyIpoptProblemInterface.__init__` method that calls `cyipopt.Problem.__init__`
- Split `cyipopt_solver.py` into two files so `PyomoCyIpoptSolver` can be imported (and registered) even if `cyipopt` is not available
- Add calls to `CyIpoptNLP.close` in a few tests to force deallocation Ipopt data structures so that Ipopt's log file can be removed on windows

@carldlaird can you take a look at this?

pyomo/contrib/pynumero/algorithms/solvers/pyomo_ext_cyipopt.py
```diff
@@ -286,6 +286,10 @@ def __init__(
 
         # currently, this interface does not do anything with Hessians
 
+        # Call CyIpoptProblemInterface.__init__, which calls
+        # cyipopt.Problem.__init__
+        super(PyomoExternalCyIpoptProblem, self).__init__()
+
```
Robbybp, line 292: Note that this `__init__` call, which wasn't previously necessary, is now necessary in subclasses of `CyIpoptProblemInterface` to make sure that the `cyipopt.Problem` get initialized so we don't segfault when we try to solve it later.

In this way, this PR could break some user code if they are subclassing `CyIpoptProblemInterface`.

pyomo/contrib/pynumero/algorithms/solvers/tests/test_pyomo_ext_cyipopt.py
```diff
@@ -179,6 +179,7 @@ def test_pyomo_external_model_scaling(self):
 
         with open('_cyipopt-pyomo-ext-scaling.log', 'r') as fd:
             solver_trace = fd.read()
+        cyipopt_problem.close()
         os.remove('_cyipopt-pyomo-ext-scaling.log')
```
Robbybp, line 183: This call to `cyipopt.Problem.close` is now necessary on windows if we try to remove a file too early (I think before the garbage collector has deallocated the `cyipopt.Problem`, but my testing was inconclusive).

This is another way this PR could break a user's code.

pyomo/contrib/pynumero/algorithms/solvers/cyipopt_solver.py
```diff
@@ -563,7 +292,7 @@ class PyomoCyIpoptSolver(object):
         "intermediate_callback",
         ConfigValue(
             default=None,
-            description="Set the function that will be called each iteration.",
+            description="Set the function that will be called each" " iteration.",
```
Robbybp, line 295: Is this part of black's string normalization?

Robbybp: This appears to be what happens when black combines a two-line string onto a single line. I think combining the strings would be "altering your code," so black doesn't do it.

blnicho: Yeah, we saw this happen in most of the PRs applying black and I think Miranda went through and manually removed the extra quotation marks.

jsiirola (approved):

pyomo/contrib/pynumero/algorithms/solvers/pyomo_ext_cyipopt.py
Robbybp: I am thinking of adding an `__initialized` flag to `CyIpoptProblemInterface.__init__` so we can raise a warning/error if `solve` is called when `cyipopt.Problem` hasn't been initialized.

Robbybp:
I've added a `_problem_initialized` flag that we set in `CyIpoptProblemInterface.__init__`, then check in `CyIpoptProblemInterface.solve`. In this way, if somebody overrides `__init__` but doesn't call `super().__init__`, they will get an error when we attempt to call `solve` in e.g. `PyomoCyIpoptSolver`. This should prevent segfaults in this case, unless the user has also overridden `solve`, in which case they are on their own.

blnicho (approved):

PR #2761, opened 2023-03-02
[CODECOV] Rework Upload of Coverage Reports

mrmundt:
## Summary/Motivation:

We are getting constant failures for codecov uploads. This attempts to fix that.

- Turn off codecov uploads on branches
- Add token for upload on PR/main test suite

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/test_branches.yml
+++ b/.github/workflows/test_branches.yml
@@ -807,6 +807,7 @@ jobs:
       uses: codecov/codecov-action@v3
       with:
         files: coverage.xml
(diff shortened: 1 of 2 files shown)


jsiirola:
It looks like the codecov-action may not actually be able to read the secret: either because the secret is not available to PRs from forks, or because there is an issue with the current version of the codecov uploader (see codecov/codecov-action#598).  I think it still makes sense to merge this, as it
1. reduces attempts at coverage uploads (skipping them on branches)
2. should use the token when building / uploading main branch builds

Issue #2771, opened 2023-03-18
Load solution with cyipopt


ZedongPeng:
### Steps to reproduce the issue

```console
$ command1 [options]
$ command2 [options]
...
```

```
# example.py
from pyomo.environ import *
from pyomo.solvers.tests.models.QCP_simple import QCP_simple

qcp = QCP_simple()
qcp.generate_model()
model = qcp.model
model.pprint()
opt = SolverFactory('cyipopt')
results = opt.solve(model,tee=True,load_solutions=False)
print(results)

model.solutions.load_from(results)
```

### Error Message

```console
$ # Output message here, including entire stack trace, if available
```

### Information on your system

Pyomo version:
Python version:
Operating system:
How Pyomo was installed (PyPI, conda, source):
Solver (if applicable):

Robbybp:
A naive fix is to just add the `_cuid` flag that is missing to the `Solution` object at some point, but I'm not sure if this is the "right" fix, as I'm not sure why this doesn't cause an error with `SolverFactory("ipopt")` as well.

ZedongPeng:
Any progress on this? @Robbybp

Robbybp:
Nope. `opt` isn't a part of the codebase I know well, and the missing `_cuid` flag seems confusing. I just took a brief look at this because it involved CyIpopt.

ZedongPeng:
Any thoughts on this? @jsiirola

adowling2:
@jsiirola @blnicho Can we look at this next week?

PR #2773, opened 2023-03-19
Add default `_cuid` flag to `Solution`

Robbybp:
## Summary/Motivation:
`ModelSolutions.add_solution` appeared to be relying on `Solution` objects having a `_cuid` attribute, which they only have if they were created via `ModelSolutions.store_to`. Adding a default `_cuid` flag to `Solution` seems to fix the immediate problem, although it still does not add a public API to specify whether a `Solution`'s keys are CUIDs (I think that's what this flag is specifying...)

### Legal Acknowledgement

### DIFF
--- a/pyomo/opt/results/solution.py
+++ b/pyomo/opt/results/solution.py
@@ -64,6 +64,7 @@ class Solution(MapContainer):
         self.declare('constraint', value={})
 
         self._option = default_print_options
+        self._cuid = False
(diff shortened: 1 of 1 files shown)


Robbybp:
While this PR fixes the MWE #2771, it leads to many other errors that I don't understand. Not clear what the right way to fix this.

PR #2789, opened 2023-03-24
Test indexed block in Incidence Analysis

Robbybp:
## Summary/Motivation:
`IncidenceGraphInterface` does not support being given indexed blocks as the "model" (only `PyomoNLP` or `_BlockData`). This PR updates an error message to indicate that `_BlockData` is supported (rather than `Block`, which may imply that `IndexedBlock` is supported). It also adds a test to make sure we get the correct error message when an `IndexedBlock` is provided and that `IncidenceGraphInterface` can be constructed from a non-`Block` `_BlockData` (i.e. a data object of an `IndexedBlock`).

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/incidence_analysis/interface.py
+++ b/pyomo/contrib/incidence_analysis/interface.py
@@ -344,7 +344,8 @@ class IncidenceGraphInterface(object):
         else:
             raise TypeError(
                 "Unsupported type for incidence matrix. Expected "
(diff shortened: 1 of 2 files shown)


mrmundt (approved):

mrmundt:
FYI, @Robbybp , Jenkins is acting up - ideally it'll be back to normal in a bit, but just be aware.

PR #2797, opened 2023-04-07
contrib.piecewise: Fixing a bug with caused by degenerate simplices

emma58:
## Summary/Motivation:

When we construct a `PiecewiseLinearFunction` from a set of points, a lot of times users will give us a grid. This means that the Delaunay triangulation is quite degenerate, so we get a lot of degenerate simplices back. This PR adds a check after calling the triangulation to filter out all simplices that are not full-dimensional. 

- Filter the simplices from `scipy.spatial.Delaunay` and keep only full-dimensional ones
- Log when points were dropped in the triangulation due to numerical issues
- Raise a `DeveloperError` when the linear system solve to find the hyperplane above a simplex is unexpectedly singular: this is a sign that we didn't correctly filter out degenerate simplices (in particular, it's possible we'll need a tolerance rather than keeping anything with a determinant that's not exactly 0.)

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/piecewise/piecewise_linear_function.py
+++ b/pyomo/contrib/piecewise/piecewise_linear_function.py
@@ -288,7 +288,7 @@ class PiecewiseLinearFunction(Block):
             raise
 
         # Get the points for the triangulation because they might not all be
-        # there if any where coplanar.
+        # there if any were coplanar.
         obj._points = [pt for pt in map(tuple, triangulation.points)]
         obj._simplices = []
         for simplex in triangulation.simplices:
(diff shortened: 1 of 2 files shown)


pyomo/contrib/piecewise/piecewise_linear_function.py
```diff
@@ -30,6 +33,8 @@
 # be zero.
 ZERO_TOLERANCE = 1e-8
 
+logger = logging.getLogger('contrib.piecewise')
```
mrmundt, line 36: We love logging. Yes, we do.

mrmundt (approved):

jsiirola (approved):

PR #2804, opened 2023-04-12
New add_column() method for appsi solvers

darrylmelander:
## Summary/Motivation:
This PR adds a new add_column() method to the solvers in the pyomo.contrib.appsi.solvers package. Like the add_column() method in the persistent solvers in the main pyomo.solvers package, this function allows you to add a new variable, and add it to the objective and to existing constraints, all in a single function call. This gives persistent solvers an opportunity to add the new variable more efficiently than is possible when modifying existing constraints one by one.

- Add a new `add_column()` method to appsi solvers.

### Legal Acknowledgement


darrylmelander:
@michaelbynum, This is a PR you will probably want to review.

One thing I'd like to note is that I had to comment out some lines in the test for this PR (see [these commented out lines](https://github.com/darrylmelander/pyomo/blob/appsi-add-column/pyomo/contrib/appsi/solvers/tests/test_persistent_solvers.py#L1225)). Those lines were intended to verify that the pyomo model components were modified as expected, but I could never get the comparisons to consistently work as expected, even though the string representation of the expressions showed they were equivalent. If there's a preferred way to make those comparisons, great, I'd love to have them added to the test.

pyomo/contrib/appsi/solvers/gurobi.py
```diff
@@ -857,6 +857,43 @@ def _set_objective(self, obj):
         self._solver_model.setObjective(gurobi_expr + value(repn_constant), sense=sense)
         self._needs_updated = True
 
+    def _add_column(
```
mrmundt, line 860: Any particular reason that `gurobi` and `highs` only have protected instantiations of this method?

darrylmelander: Yes, they are the two appsi solvers that inherit from `PersistentBase`. They get their public implementation from that base class.

mrmundt (approved):

jsiirola:
@michaelbynum, can you take a look at this?

michaelbynum (approved):
A couple minor comments below.

michaelbynum (changes requested):
we are pondering the above questions a bit

michaelbynum:
I think we should move this method into a "helper function" that takes in a model and a solver and modifies both. This way, we maintain that solver interfaces do not modify models as a "side effect".

blnicho:
@darrylmelander @michaelbynum this PR has been open for a while, any updates or plans for finishing this?

bknueven:
> @darrylmelander @michaelbynum this PR has been open for a while, any updates or plans for finishing this?

If @darrylmelander is unavailable I could pick this up. It would still be a useful thing for APPSI (or any of our "in memory") solver interfaces to have.

bknueven:
> I think we should move this method into a "helper function" that takes in a model and a solver and modifies both. This way, we maintain that solver interfaces do not modify models as a "side effect".

@michaelbynum would the function signature we currently use for persistent solvers be better?
https://github.com/Pyomo/pyomo/blob/b5825e9e146552690d942903386102105251464c/pyomo/solvers/plugins/solvers/persistent_solver.py#L197-L214

mrmundt:
Opened issue #3930 to keep track of this feature request. Due to the age, conflicts, and desire to change implementation/location of implementation, we are going to close this PR. We do believe this feature is worthwhile - it just needs some attention and a fresh look.

PR #2809, opened 2023-04-17
Adding (reversible) `gdp.transform_current_disjunctive_logic` transformation

emma58:
## Summary/Motivation:

The current `gdp.fix_disjuncts` transformation guarantees that it transforms a GDP all the way to a MI(N)LP. Because this means it needs to call another transformation to transform any LogicalConstraints, this makes it difficult to add a "reverse" transformation to undo the reclassification of the Disjuncts (which is a useful thing for initializing nonlinear GDP models for solving with GDPopt, etc.) Instead of trying to deprecate the going-to-MINLP guarantee, this PR adds a new transformation, `gdp.transform_current_disjunctive_state` that doesn't even promise to transform all the GDP components: It just goes through all of the Disjunctions, and if any have enough indicator_vars fixed to determine which Disjunct(s) will be selected, it reclassifies the Disjuncts and activates/deactivates accordingly. It currently does nothing with LogicalConstraints, though in the future it would be useful if it took them into account in the logic to decide what Disjuncts will be selected.

- Adds a `ReversibleTransformation` base class that lets `apply_to` have a return value, which is assumed to be a token containing the right information so it can be passed into a `reverse` argument in order to undo the transformation. 
- Adds `gdp.transform_current_disjunctive_logic` transformation

### Legal Acknowledgement

### DIFF
--- a/pyomo/gdp/plugins/transform_current_disjunctive_logic.py
+++ b/pyomo/gdp/plugins/transform_current_disjunctive_logic.py
@@ -10,28 +10,37 @@
 #  ___________________________________________________________________________
 
 from pyomo.common.collections import ComponentSet
+from pyomo.common.config import ConfigDict, ConfigValue
 from pyomo.common.errors import InfeasibleConstraintException
-from pyomo.core.base import Transformation, TransformationFactory
+from pyomo.core.base import (
+    Block,
+    SortComponents,
+    ReversibleTransformation,
+    TransformationFactory,
+)
 from pyomo.gdp import Disjunct, Disjunction
+
+from pytest import set_trace
 
 
 @TransformationFactory.register(
-    'gdp.transform_fixed_disjunctive_logic',
-    doc="""Convert fixed disjunctive logic (fixed Disjunct indicator_vars) to
-    MI(N)LP form in cases where enough of the logic is fixed to determine 
-    the logical solution.""",
+    'gdp.transform_current_disjunctive_logic',
+    doc="""Convert current disjunctive logic (values of Disjunct
+    indicator_vars) to MI(N)LP form in cases where enough of the 
+    logic is fixed to determine which Disjunct(s) will be selected.
+    """,
 )
-class TransformFixedDisjunctiveLogic(Transformation):
-    """This transformation finds fixed disjunctive logic (fixed indicator_vars)
-    in the given GDP, and, for any Disjunction, when the fixed logic fully
-    determines what Disjunct will be selected it reclassifies all the Disjuncts
+class TransformCurrentDisjunctiveLogic(ReversibleTransformation):
+    """This transformation finds disjunctive logic (indicator_vars values)
+    in the given GDP, and, for any Disjunction, when the logic fully
+    determines what Disjunct will be selected, it reclassifies all the Disjuncts
     of the Disjunction as Blocks, and activates or deactivates them according
-    to whether they are fixed (or implied to be fixed) as 'True' or 'False'
-    (respectively).
+    to whether their indicator_vars are set (or implied to be set) as
+    'True' or 'False' (respectively).
 
     Note that this transformation does not necessarily return a MI(N)LP since
     it will not transform Disjunctions that are not fully determined by the
-    fixed logic.
+    current logic.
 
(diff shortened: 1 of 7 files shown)


pyomo/gdp/plugins/transform_current_disjunctive_logic.py
```diff
@@ -0,0 +1,213 @@
...
+                    no_val, false_val, reverse_token['_disjuncts']
+                )
+                reverse_token['_disjunctions'].add(disjunction)
+                disjunction.deactivate()
+        # It's only an 'at-least' not an 'exactly'...
+        elif len(false_val) == num_disjuncts:
+            raise InfeasibleConstraintException(
+                "Atleast-one constraint for Disjunction "
+                "'%s' is violated. That is, all the Disjunct "
+                "indicator_vars are 'False'." % disjunction.name
```
bernalde, line 176: Should this exception be copied above in case all disjunct are False?

bernalde (approved):
Cool PR. I see how this would simplify things for GDPOpt. I see no issues with it.

pyomo/gdp/plugins/transform_current_disjunctive_logic.py
emma58: Ah, good point, I can check for this one before I even need to know about exactly-one or at-least. That will probably make a cleaner error message.

mrmundt (approved):

bernalde (approved):

pyomo/gdp/plugins/transform_current_disjunctive_state.py
```diff
@@ -0,0 +1,211 @@
...
+            reverse_dict['_disjunctions'].add(disjunction)
+            disjunction.deactivate()
+
+    def _reclassify_disjuncts(
+        self, true_disjunctions, false_disjunctions, reverse_dict
+    ):
+        for disj in true_disjunctions:
+            reverse_dict[disj] = (disj.indicator_var.fixed, disj.indicator_var.value)
+            disj.indicator_var.fix(True)
+            disj.parent_block().reclassify_component_type(disj, Block)
```
jsiirola, line 196: I worry as this reclassifies *all* Disjuncts in an indexed Disjunct, whether nor not all the disjuncts state is actually determined by the `indicator_var` (e.g., only transforming part of a model)

jsiirola (commented):
Overall, this looks reasonable, but there is a bug that needs to be addressed before merging (supporting transforming partial models)

pyomo/gdp/plugins/transform_current_disjunctive_state.py
```diff
@@ -0,0 +1,297 @@
...
+        false_disjuncts,
+        reverse_dict,
+        disjunct_set,
+        disjunct_containers,
+    ):
+        for disj in true_disjuncts:
+            reverse_dict[disj] = (disj.indicator_var.fixed, disj.indicator_var.value)
+            self._update_transformed_disjuncts(disj, disjunct_set, disjunct_containers)
+            parent_block = disj.parent_block()
+            parent_block.reclassify_component_type(disj, Block)
```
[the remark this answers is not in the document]
emma58: I think this is actually okay--we're comparing the GDP hierarchy with the component hierarchy as we do this transformation, so if there are components that we didn't catch as we transform GDP stuff, we throw an error at the end since we may have inadvertently reclassified them. So your example above throws this error:
```
  File "/home/esjohn/src/pyomo/thing.py", line 10, in <module>
    TransformationFactory('gdp.transform_current_disjunctive_state').apply_to(m)
  File "/home/esjohn/src/pyomo/pyomo/core/base/transformation.py", line 77, in apply_to
    reverse_token = self._apply_to(model, **kwds)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/esjohn/src/pyomo/pyomo/gdp/plugins/transform_current_disjunctive_state.py", line 146, in _apply_to
    raise GDP_Error(
pyomo.gdp.disjunct.GDP_Error: Found active Disjuncts on the model that were not included in any Disjunctions:
d[3], d[2], e[2]
Please deactivate them or include them in a Disjunction.
```

PR #2772, opened 2023-03-19
[WIP] Use `AMPLRepnVisitor` to generate incidence graph in Incidence Analysis

Robbybp:
## Fixes
Potentially unexpected behavior in Incidence Analysis caused (partially) by https://github.com/Pyomo/pyomo/pull/2722. See added tests in test_interface.py.

## Summary/Motivation:
Behavior of Incidence Analysis when terms can cancel out of constraint expressions is not well defined. Currently, `0*x` expressions are removed during expression construction, so they don't appear in the incidence graph. Other "linear cancellations", such as `x - x`, are not removed. https://github.com/Pyomo/pyomo/pull/2722 will remove the `0*x -> 0` optimization, in part because the result is undefined (or our current behavior is incorrect) in odd cases like `0*float('nan')`. This PR uses the `AMPLRepnVisitor` introduced in the new NL writer to more robustly construct the "correct" incidence graph.

## Changes proposed in this PR:
- Use `AMPLRepnVisitor` instead of `identify_variables` to decide which variables to include in an incidence graph. This involves a couple of tricks/hacks to get it to work correctly (according to the current tests). See comments in the code below.
- A small patch to `TemporarySubsystemManager` allowing us to specify variables to temporarily unfix. This is necessary to preserve the current behavior of the `include_fixed` argument to `IncidenceGraphInterface` while updating to the NL visitor.

## Performance
The following is a rough profile of running Dulmage-Mendelsohn on a 35-node dynamic pipeline model I was debugging recently. Here `NFE` is the number of (backward difference) finite elements in both the time domain and the length domain of each pipeline.

Using `main`:
```
NFE = 2
N. var: 5101
N. con: 5101
N. edges: 14530
N. over/under-constrained: 3269
Identifier                ncalls   cumtime   percall      %
-----------------------------------------------------------
root                           1     1.215     1.215  100.0
     ------------------------------------------------------
     construct_graph           1     0.135     0.135   11.1
     create_instance           1     0.959     0.959   78.9
     dulmage_mendelsohn        1     0.121     0.121    9.9
     other                   n/a     0.000       n/a    0.0
     ======================================================
===========================================================

NFE = 4
N. var: 12899
N. con: 12899
N. edges: 38134
N. over/under-constrained: 9613
Identifier                ncalls   cumtime   percall      %
-----------------------------------------------------------
root                           1     2.468     2.468  100.0
     ------------------------------------------------------
     construct_graph           1     0.241     0.241    9.8
[...]

PR #2816, opened 2023-04-27
[WIP] Visitor for incidence graph generation

Robbybp:
## Fixes
Variables multiplied by zero appear in incidence graphs in Incidence Analysis.

## Summary/Motivation:
Expected behavior:
- Zero coefficients and other "linear cancellations" cause variables to not appear in incidence graphs
- NaN behaves as in IEEE-754 and absorbs everything to become a NaN-valued constant
- `None` means "uninitialized". It absorbs any other non-NaN, non-zero constant but never cancels out a variable. It is appropriately absorbed by zero or NaN.

I am happy to take other suggestions for expected behavior. This supersedes #2772. 

## To-dos

- [ ] Finish implementing visitor, clean up code, test thoroughly
- [x] Change name from `AMPLRepnVisitor`
- [x] Edit: Improve API for some `get_incident_variables(expr)` function
- [ ] Documented intended behavior for incidence graph generation

### Legal Acknowledgement


Robbybp:
This PR now includes tests that we can identify the expected "incident variables" from a variety of expressions. It also updates incident variable identification to use a variation of `AMPLRepn` that stores the nonlinear part of the expression as simply a set of variable IDs.

Next priorities:
- [ ] "Regression test" on some of my examples
- [ ] Test that the "linear only" functionality works as expected
- [ ] Document expected behavior for generating incidence graphs

There is also still a good bit of unused code in the `incidence_analysis/visitor.py` file, and some that I don't fully understand.
- [ ] Remove redundant/unused code from `visitor.py`

Robbybp:
This branch passes tests for my three implicit function benchmark problems, which rely on block triangularization. In addition, here is a summary of running my Incidence Analysis examples with this branch:

| Example  | Runs w/o error? | N. edges (main) | N. edges (PR) | Correct? |
| -------- | --------------- | --------------- | ------------- | -------- |
| DoF  | Y  |  3,933 | 3,921 | Y |
| Stuct. sing. | Y | 6,943 | 6,922 | Y |
| Num. sing. | Y | 7,792 | 7,126 | Y |
| Init. cond. | Y | 14,560 | 14,536 | Y* |
| Discretization | Y | 3,916 | 3,900 | Y** |
| Initialization | Y | 900 | 897 | Y** |

*Leads to one fewer variable in the underconstrained subsystem
**Leads to a different number of diagonal blocks in the block triangularization

The "Correct" column means: "Does the incidence graph allow us to draw the same conclusions about the problem with the model?"

Robbybp:
A test failure I'm confused about:
```python
import pyomo.environ as pyo
from pyomo.contrib.incidence_analysis.visitor import get_incident_variables

m = pyo.ConcreteModel()
m.x = pyo.Var([1, 2, 3])
m.p = pyo.Param(mutable=True, initialize=None)
expr = - m.p * m.x[2] + 2 * m.x[2]
variables = get_incident_variables(expr)
assert len(variables) == 1
assert variables[0] is m.x[2]
```
This fails in the following `not in native_types` branch in `_before_linear`:
```python
for arg in child.args:
        c, v = arg.args
        if c.__class__ not in native_types:
            c = c()
```
This probably needs a branch or try/except to catch `NPV_Expression` and properly convert expressions of uninitialized mutable parameters to `None`.

Robbybp:
> - [ ] Test that the "linear only" functionality works as expected

This is some new functionality I would like to add that is made possible by switching to a new walker. The functionality is currently in the PR, but it is not tested. This could potentially be removed and saved for another PR.

Robbybp:
I've again encountered the question of how to handle NaN, particularly when it is a coefficient of a variable. The current behavior is the following:
```python
import pyomo.environ as pyo
from pyomo.contrib.incidence_analysis.visitor import get_incident_variables
m = pyo.ConcreteModel()
m.x = pyo.Var([1, 2])
expr = float('nan') * m.x[1] + m.x[2]
variables = get_incident_variables(expr)
assert len(variables) == 1
assert variables[0] is m.x[2]
```
I.e. if a variable has a coefficient of NaN, it is ignored. However, by "proper treatment" of NaN, the entire expression here would reduce to a single constant of NaN, and therefore would have **no** incident variables. I'm wondering:
- Is there is a legitimate use case for keeping variables multiplied by NaN in the expression system when doing incidence graph analysis?
- Is there a legitimate use case for keeping any expression involving NaN when doing incidence graph analysis?

Robbybp:
An implementation using `generate_standard_repn` might look like:
```python
def get_incident_variables(
    expr, include_fixed=False, linear_only=False, filter_zeros=True
):
    if include_fixed:
        to_unfix = [] 
        var_set = set()
        for var in identify_variables(expr, include_fixed=True):
            if id(var) not in var_set:
                var_set.add(id(var))
                to_unfix.append(var)
        # We will temporarily unfix variables to generate the incident
        # variables
        context = TemporarySubsystemManager(to_unfix=to_unfix)
    else:
        context = nullcontext()

    with context:
        repn = generate_standard_repn(expr, compute_values=False)

    linear_vars = [] 
    for var, coef in zip(repn.linear_vars, repn.linear_coefs):
        try: 
            coef_value = pyo.value(coef)
        except:
            coef_value = None 
        if coef_value is None or (coef_value != 0 and not math.isnan(coef_value)):
            linear_vars.append(var)

    if linear_only:
        return linear_vars
    else:
        variables = tuple(linear_vars) + repn.quadratic_vars + repn.nonlinear_vars

        unique_variables = [] 
        seen_var_ids = set()
        for var in variables:
            v_id = id(var)
            if v_id not in seen_var_ids:
                seen_var_ids.add(v_id)
                unique_variables.append(var)
        return unique_variables
```
This implementation fails tests, apparently because it is not doing "linear cancellations" to the extent my tests currently expect. E.g.:
```python
    def test_var_pow_one(self):
        m = pyo.ConcreteModel()
        m.x = pyo.Var([1, 2, 3], initialize=2)
        m.x[1].fix(1)
        # x**1 appears to get simplified during expression generation, so
        # using a fixed var here is necessary to cover the x**1 simplification
        # in the visitor.
        expr = m.x[2] ** m.x[1]
        variables = get_incident_variables(expr)
        self.assertEqual(len(variables), 1)
        self.assertIs(variables[0], m.x[2])
    
        expr = m.x[2] ** m.x[1] - m.x[2]
        variables = get_incident_variables(expr)
>       self.assertEqual(len(variables), 0)
E       AssertionError: 1 != 0
```
This also fails several tests in `test_interface.py` and `test_dulmage_mendelsohn.py`, which I haven't looked into. These tests should not rely on cancellation.

Robbybp:
I'm closing this PR in favor of #2834. This PR makes some optimizations that are not present in #2834:
- Cancellations due to "fixed" coefficients (instead of just constant coefficients)
- Simplifies expressions like `x**y` where `y` is fixed to one.

While I would like to have these optimizations, my concern with this PR is testing. There are many lines of arithmetic in the visitor that are covered but not tested properly as, if I remove them, my tests still pass. This is because, for most arithmetic in the visitor, the only way the incidence graph can change is if the resulting numeric value induces a cancellation. This means that, for proper testing, every line of arithmetic needs tests where the result of the line do and don't result in a cancellation. This is more work than I have time to put into this PR, and stretches my understanding of the `incidence_analysis/visitor.py` file, much of which is inherited from the nl writer (v2).

jsiirola:
I think this makes sense.  PR #2823 has a new `LinearRepnVisitor` that has these optimizations (linear cancellation and handling of `**1`).  We can revisit this after that PR is merged (and hopefully when I have a chance to make a proper "numerical / structural" split for things like `identify_variables` and `polynomial_degree`).

PR #2826, opened 2023-05-09
contrib.piecewise: Adding tolerance when filtering degenerate simplices

emma58:
## Summary/Motivation:

When constructing a `PiecewiseLinearFunction` from a set of user-given points, we use scipy to calculate a triangulation of those points. This can create degenerate simplices that need to be filtered out before we calculate the linear functions. Previously, when checking if simplices were full-dimensional, we were checking for a non-zero determinant with no numerical tolerance whatsoever. Unsurprisingly, that doesn't work. This switches to using numpy's `matrix_rank` function which has a default implementation to account for numerical error in the SVD calculation. (We could alternatively add our own tolerance on the determinant, but I think that the numpy implementation is more clever than we would be, though we may someday need to extend and allow for a user to pass their own tolerance through to the numpy function.)

- Switches from filtering simplices using determinant to using `matrix_rank`
- Adds some documentation of the simplex filtering so that I might believe my past self more quickly in the future
- Adds a kind of ugly test (more of an integration test) that was hitting numerical issues in the determinant calculation

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/piecewise/piecewise_linear_function.py
+++ b/pyomo/contrib/piecewise/piecewise_linear_function.py
@@ -295,7 +295,7 @@ class PiecewiseLinearFunction(Block):
         # there if any were coplanar.
         obj._points = [pt for pt in map(tuple, triangulation.points)]
         obj._simplices = []
-        for idx, simplex in enumerate(triangulation.simplices):
+        for simplex in triangulation.simplices:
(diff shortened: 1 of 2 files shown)


mrmundt (approved):
Thanks for all the comments in here. It's a really good explanation.

PR #2828, opened 2023-05-12
fix appsi gurobi available()

michaelbynum:
## Summary/Motivation:
This PR updates the `available()` method in the appsi-Gurobi interface so that license availability can be checked multiple times. 

### Legal Acknowledgement


michaelbynum:
This has some undesired side effects...

jsiirola:
Why does this remove the code to distinguish between demo and full licenses?  More generally, I think I would prefer to leave the default interface to rely on caching and instead add either an option or a new method to forget the previously cached result?

michaelbynum:
I don't think caching is a good idea when the result can change.

michaelbynum:
I guess we could cache demo vs full, but recheck availability each call. Demo vs full should not change.

pyomo/contrib/appsi/solvers/gurobi.py
```diff
@@ -257,42 +257,46 @@ def __init__(self, only_child_vars=True):
...
             with capture_output(capture_fd=True):
                 m = gurobipy.Model()
-                m.dispose()
-        except ImportError:
-            # Triggered if this is the first time the deferred import of
-            # gurobipy is resolved. _import_gurobipy will have already
-            # set _available appropriately.
-            return
+            if self._solver_model is None:
+                self._solver_model = m
```
jsiirola, line 275: This looks like the solver interface will create (and perpetually hold on to) a Gurobi model -- I suspect to grab the license and keep others from "stealing" it from us.  Can you document somewhere (preferably obvious for users) the strategy that we are using for managing Gurobi licenses (and in particular what a user should do to release their license)?

michaelbynum: Done!

mrmundt (approved):

jsiirola (approved):

Issue #101, opened 2017-01-15
Rework doc and examples hierarchy 


blnicho:
The pyomo examples directory needs to be cleaned up for consistency and for inclusion in testing.

franktoffel:
Also, pyomo.org web and PDF docs are referred to the 4.1 version. Not sure if there is backwards compatibility but seems confusing.

mrmundt:
Potentially problematic examples:
- `examples/pyomo/p-median/solver1.py` - two unknown imports
- `examples/pyomo/p-median/solver2.py` - two unknown imports
- `examples/pyomo/quadratic/example3.py` - use of `xrange`; might need to be replaced with `range`?
- `example/pyomo/draft/api.py` - I don't even know what's going on in here

PR #2838, opened 2023-05-23
Add some more instances to the GDP constrained layout example

sadavis1:
## Summary/Motivation:
(remaking PR to use branch properly)
Adds some additional example instances from Sawaya (2006) to this GDP constrained layout example, and adds some drawings.

 -  Add additional problem instances
 -  Add matplotlib drawing code
 -  Add euclidean distance option

### Legal Acknowledgement


examples/gdp/constrained_layout/cons_layout_model.py
```diff
@@ -63,16 +150,16 @@ def y_max(m, rect):
             for circ in m.circles
         )
 
-    m.ordered_rect_pairs = Set(
-        initialize=m.rectangles * m.rectangles, filter=lambda _, r1, r2: r1 != r2
-    )
     m.rect_pairs = Set(
-        initialize=[(r1, r2) for r1, r2 in m.ordered_rect_pairs if r1 < r2]
+        initialize=m.rectangles * m.rectangles, filter=lambda _, r1, r2: r1 < r2
```
emma58, line 154: Thank you, this is waaay prettier!

examples/gdp/constrained_layout/cons_layout_model.py
```diff
@@ -91,20 +178,40 @@ def y_bounds(m, rect):
...
+    @m.Constraint(m.rect_pairs, doc="x-distance between rectangles")
+    def dist_x_defn_2(m, r1, r2):
+        return m.dist_x[(r1, r2)] >= m.rect_x[r1] - m.rect_x[r2]
 
-    @m.Constraint(m.ordered_rect_pairs, doc="x-distance between rectangles")
-    def dist_x_defn(m, r1, r2):
-        return m.dist_x[tuple(sorted([r1, r2]))] >= m.rect_x[r2] - m.rect_x[r1]
+    @m.Constraint(m.rect_pairs, doc="y-distance between rectangles")
+    def dist_y_defn_1(m, r1, r2):
+        return m.dist_y[(r1, r2)] >= m.rect_y[r2] - m.rect_y[r1]
```
emma58, line 210: You don't need the parentheses around `r1, r2` here (and the same below).

examples/gdp/constrained_layout/cons_layout_model.py
```diff
@@ -148,3 +255,92 @@ def rectangle_in_circle(m, r):
...
+        ax.text(
+            m.circle_x[c],
+            m.circle_y[c] + m.circle_r[c] + 1.5,
+            f"Circle {c}",
+            horizontalalignment="center",
+        )
+
+    for r in m.rectangles:
+        print(
+            f"drawing rectangle {r}: x={m.rect_x[r]()}, y={m.rect_y[r]()} (center), L={m.rect_length[r]}, H={m.rect_height[r]}"
```
emma58, line 300: Instead of using `__call__` to get the value of the Vars, you should use Pyomo's `value` function. So for example, replace `m.rect_y[r]()` with `value(m.rect_y[r])`. It will get you better error checking because the implementation of `__call__` will happily return `None` if the Var doesn't have a value, and in this case, getting an error would be better.

examples/gdp/constrained_layout/cons_layout_model.py
```diff
@@ -148,3 +255,92 @@ def rectangle_in_circle(m, r):
...
+            fontsize=8,
+        )
+
+    plt.show()
+
+
+# Solve some example problems and draw some pictures
+if __name__ == "__main__":
+    from pyomo.environ import SolverFactory
+    from pyomo.core.base import TransformationFactory
```
emma58, line 329: You can actually import this from `environ` too

examples/gdp/constrained_layout/cons_layout_model.py
```diff
@@ -148,3 +255,92 @@ def rectangle_in_circle(m, r):
...
+    for d in ["l1", "l2"]:
+        for key in constrained_layout_model_examples.keys():
+            print(f"Solving example problem: {key}")
+            model = build_constrained_layout_model(
+                constrained_layout_model_examples[key], metric=d
+            )
+            transformer.apply_to(model)
+            solver.solve(model)
+            print(f"Found objective function value: {model.min_dist_cost()}")
+            draw_model(model, title=(key if d == "l1" else f"{key} ({d} distance)"))
```
emma58, line 345: Minor nit, but why not put which norm in the title all the time?

emma58 (commented):
A few small nits, but this looks good!

examples/gdp/constrained_layout/cons_layout_model.py
sadavis1: Hm. I think I wrote it like this because the l1 norm was the "default" problem specified in the paper and the euclidean distance was a variant invented by me. I can change it though, it should only make things clearer.

sadavis1:
I've made some changes to resolve these issues, please let me know if there are any others.

emma58 (approved):

emma58:
@blnicho, @mrmundt this is only failing coverage, which is kind of its own can of worms since it's in `examples` and we don't have tests to begin with... Are y'all okay with merging anyway?

blnicho (approved):

blnicho:
@emma58 I'm fine with merging this. Consistent testing of the code in the `examples` directory is a bigger issue (https://github.com/Pyomo/pyomo/issues/101)

Issue #875, opened 2019-02-21
bug in core/base/sos.py


DLWoodruff:
# taken from core/base/sos.py docstring and modified slightly
```
import pyomo.environ as pyo
model = pyo.ConcreteModel()
model.A = pyo.Set(initialize = range(2))
model.B = pyo.Set(model.A, initialize={1: "hello", 2: "goodbye"})
model.X = pyo.Var(model.B)
model.C1 = pyo.SOSConstraint(model.A, var=model.X, set=model.B, sos=1)
```
According to the docstring in core/base/sos.py:

    This constraint actually creates one SOS-1 constraint for each
    element of model.A (e.g., if |A| == N, there are N constraints).
    In each constraint, model.X is indexed by the elements of
    model.B[a], where 'a' is the current index of model.A.

But, this example cannot run at all because a Var cannot be indexed by an 
indexed Set.
```
#So maybe this is sort of what was meant :
import pyomo.environ as pyo
model = pyo.ConcreteModel()
model.A = pyo.Set(initialize = range(2))
model.B = pyo.Set(initialize=["hello", "goodbye"])
model.C = model.A * model.B
model.X = pyo.Var(model.C)
model.C1 = pyo.SOSConstraint(model.A, var=model.X, set=model.C, sos=1)
```
This example results in:
```
ValueError: Unexpected keyword options found while constructing 'IndexedSOSConstraint':
	set
```

tkorvola:
More bugs in the SOSConstraint docstring:
- The keyword for SET is `index`; `set` is not recognized.
- N does not default to 1 but must be specified.
- There is an alternative way to call the constructor, which is not documented at all: instead of `var`, `index` and `weights` one specifies a `rule`, which is called in the usual manner.

mrmundt:
This theoretically was addressed in the linked PR, so closing this issue. Please open a new one if there are still concerns.

Issue #1407, opened 2020-04-22
Allow SOS constraints in GAMS plugin


edurauh:
GAMS can handle SOS1 and SOS2 constraints, but the GAMS plugin does not allow it
Can we please add this functionality?

PR #2875, opened 2023-06-16
Improved documentation for SOS

pmlpm1986:
## Summary/Motivation:

The documentation for declaring special ordered sets was lacking in some regards.

This was previously mentioned in: https://github.com/Pyomo/pyomo/issues/875

I had previously used SOS through pyomo but always struggled to know if I was declaring them correctly or if there was another way to do it.

I therefore decided to explore the various possibilities and improve the documentation.

- Updated the SOSConstraint docstring
- Created an entry for special ordered sets in the website/online documentation
- Created a test file to validate all the examples mentioned in the docstring and in the website/online documentation

## Other comments

1) I did not try to use a rule to declare the constraint, as mentioned in https://github.com/Pyomo/pyomo/issues/875;
2) There appear to be very few type checks on the inputs.

### Legal Acknowledgement

## EDIT

I also included information about how to use the rule parameter.


doc/OnlineDocs/advanced_topics/sos_constraints.rst
```diff
@@ -0,0 +1,160 @@
+Special Ordered Sets (SOS)
+==========================
+
+Pyomo allows users to declare special ordered sets (SOS) within their problems.
+These are sets of variables among which only a certain number of variables can
+be non-zero and those that are must be adjacent according to a given order.
```
mrmundt, line 6: Just a nitpick.
```suggestion
be non-zero, and those that are must be adjacent according to a given order. 
```

pyomo/core/base/sos.py
```diff
@@ -106,48 +106,125 @@ def set_items(self, variables, weights):
...
-
-      model = AbstractModel()
-      model.A = Set()
-      model.X = Var(model.A)
-
-      model.C2 = SOSConstraint(var=model.X, sos=2)
-
-    This produces exactly one SOS-2 constraint using all the variables
-    in model.X.
+    Implements constraints for indexed special ordered sets (SOS).
```
mrmundt, line 109: This is amazing. Thanks for this re-document.

pmlpm1986:
I will revisit these points soon. I will also add some documentation about the use of the rules in the constructor.

doc/OnlineDocs/advanced_topics/sos_constraints.rst
```diff
@@ -0,0 +1,160 @@
...
+    ...     sos=n,
+    ...     index=model.mysosvarindexset,
+    ...     weights=model.mysosweights,
+    ...     )
+
+Compatible solvers
+------------------
+
+Not all LP/MILP solvers are compatible with SOS declarations and Pyomo might
+not be ready to interact with all those that are. The following is a list of
```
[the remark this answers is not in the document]
pmlpm1986: Some time ago I couldn't use SOS with cbc and I am led to believe that it is currently not possible with/through GAMS: https://github.com/Pyomo/pyomo/issues/1407

pyomo/core/tests/unit/test_sos_v2.py
```diff
@@ -0,0 +1,613 @@
...
+solver_name = "scip"
+
+solver_available = bool(check_available_solvers(solver_name))
+
+# *****************************************************************************
+# *****************************************************************************
+
+
+class TestSOS(unittest.TestCase):
+    @unittest.skipIf(not solver_available, "The solver is not available.")
```
mrmundt, line 18: Another recommendation: Move this `skipIf` statement to above the class so it only has to be checked once/can be easily changed.

pyomo/core/tests/unit/test_sos_v2.py
```diff
@@ -0,0 +1,613 @@
...
+            assert math.isclose(pyo.value(problem.OBJ), exp_res, abs_tol=abs_tol)
+
+        # *********************************************************************
+
+
+# *****************************************************************************
+# *****************************************************************************
+
+
+def problem_nonindexed_sosn_different_vars(n: int = 1):
```
mrmundt, line 193: I don't care either way, but these methods can live inside the TestSOS class above.

mrmundt (changes requested):
One last little change requested: move the `skipIf`. The other comment is a suggestion about which I don't feel strongly.

pyomo/core/base/sos.py
```diff
@@ -106,48 +106,180 @@ def set_items(self, variables, weights):
...
+        SOS, or as a dict, for indexed SOS(s).
+    weights : pyo.Param or dict, optional
+        A data structure with the weights for each member of the SOS(s). These
+        can be provided as pyo.Param or as a dict. If not provided, these will
+        be determined automatically using their position in the structure.
+    rule : optional
+        A method returning a 2-tuple with lists of variables and the respective
+        weights in the same order or, alternatively, pyo.Constraint.Skip if the
+        constraint should be not be included in the model/instance. This parame
+        ter cannot be used in combination with var, index or weights.
```
jsiirola, line 131: line break: "parame" / "ter"

pyomo/core/base/sos.py
```diff
@@ -106,48 +106,180 @@ def set_items(self, variables, weights):
...
+        either indexed, if the SOS is indexed; or non-indexed, otherwise.
+        Alternatively, the indexes can be provided as a list, for a non-indexed
+        SOS, or as a dict, for indexed SOS(s).
+    weights : pyo.Param or dict, optional
+        A data structure with the weights for each member of the SOS(s). These
+        can be provided as pyo.Param or as a dict. If not provided, these will
+        be determined automatically using their position in the structure.
+    rule : optional
+        A method returning a 2-tuple with lists of variables and the respective
+        weights in the same order or, alternatively, pyo.Constraint.Skip if the
```
jsiirola, line 129: This can also return a list of variables, with the weights taken from the position in the list.

pyomo/core/base/sos.py
```diff
@@ -106,48 +106,180 @@ def set_items(self, variables, weights):
...
+    index : pyo.Set, list (non-indexed SOS) or dict (indexed SOS), optional
+        A data structure with the indexes for the variables that are to be
+        members of the SOS(s). The indexes can be provided as a pyomo Set:
+        either indexed, if the SOS is indexed; or non-indexed, otherwise.
+        Alternatively, the indexes can be provided as a list, for a non-indexed
+        SOS, or as a dict, for indexed SOS(s).
+    weights : pyo.Param or dict, optional
+        A data structure with the weights for each member of the SOS(s). These
+        can be provided as pyo.Param or as a dict. If not provided, these will
+        be determined automatically using their position in the structure.
```
jsiirola, line 126: This is a bit vague.  I would suggest "If not provided, variable weights will be determined by the `var` index set"

pyomo/core/base/sos.py
```diff
@@ -106,48 +106,180 @@ def set_items(self, variables, weights):
...
-    This produces exactly one SOS-2 constraint using all the variables
-    in model.X.
+    Implements constraints for special ordered sets (SOS).
+
+    Parameters
+    ----------
+    sos : int
+        The type of SOS.
+    var : pyo.Var
+        The set of variables from which the SOS(s) will be created.
```
jsiirola, line 116: This is not technically correct: `var` needs to be an IndexedVar (not a set of variables).  That said, we could probably revise the component to also admit dictionaries of vardata objects (it will actually work now, but only if `index` is provided).

pyomo/core/base/sos.py
```diff
@@ -106,48 +106,180 @@ def set_items(self, variables, weights):
...
+    rule : optional
+        A method returning a 2-tuple with lists of variables and the respective
+        weights in the same order or, alternatively, pyo.Constraint.Skip if the
+        constraint should be not be included in the model/instance. This parame
+        ter cannot be used in combination with var, index or weights.
+
+    Examples
+    -------
+
+    1 - An SOS of type **N** made up of all members of a pyomo Var component.
```
jsiirola, line 136: SOSConstraint should be added to the list of autoclass documentation in `doc/OnlineDocs/library_reference/aml/index.rst`.

Note that that will cause doctest to test this code -- which will generate an error because `model` and `pyo` are not defined.

pyomo/core/tests/unit/test_sos_v2.py
```diff
@@ -0,0 +1,613 @@
...
+
+
+def problem_nonindexed_sosn(case: int == 0, n: int = 1, use_rule: bool = False):
+    # concrete model
+
+    model = pyo.ConcreteModel()
+    model.x = pyo.Var([1], domain=pyo.NonNegativeReals, bounds=(0, 40))
+
+    model.A = pyo.Set(initialize=[1, 2, 4, 6])
+    model.y = pyo.Var(model.A, domain=pyo.NonNegativeReals)
```
jsiirola: You can just set `bounds=(0,2)` here and avoid the `*_ub`  constraints below.

pyomo/core/tests/unit/test_sos_v2.py
```diff
@@ -0,0 +1,613 @@
...
+
+    model.OBJ = pyo.Objective(
+        expr=(
+            1 * model.x[1]
```
jsiirola: Can we negate the coefficient here?  I believe that as written, the final objective value (the only thing you are testing) is the same with / without the SOS1 constraint.

pyomo/core/tests/unit/test_sos_v2.py
```diff
@@ -0,0 +1,613 @@
...
+            try:
+                model = problem_nonindexed_sosn(case=case, n=sos, use_rule=use_rule)
+            except NotImplementedError:
+                continue
+
+            opt.solve(model, tee=show_output)
+
+            assert len(model.mysos) != 0
+
+            assert math.isclose(pyo.value(model.OBJ), exp_res, abs_tol=abs_tol)
```
jsiirola, line 77: We generally prefer using unittest assertions and not `assert`.

pyomo/core/tests/unit/test_sos_v2.py
pmlpm1986: Done.

pyomo/core/tests/unit/test_sos_v2.py
pmlpm1986: Done.

mrmundt (approved):
Thanks! Once tests pass, I am good with this.

pyomo/core/tests/unit/test_sos_v2.py
```diff
@@ -0,0 +1,613 @@
...
+    def test_nonindexedsos(self, show_output: bool = False):
+        "Test non-indexed SOS using a single pyomo Var component."
+
+        opt = pyo.SolverFactory(solver_name)
+
+        # *********************************************************************
+
+        # non-indexed SOS constraints.....
+
+        test_vectors = [
```
jsiirola, line 28: This is a stylistic issue, but I would greatly prefer if each case were it's own `test_` function and then the variants were derived classes.  That is, something like:
```python
class BaseTests(object):
    abs_tol = 1e-3

    def setupClass(self):
        self.model = pyo.ConcreteModel()
        self.model.x = pyo.Var([1], domain=pyo.NonNegativeReals, bounds=(0, 40))
        #...

    def test_scalar(self):
        def rule_mysos(m):
            return ([m.y[a] for a in m.A], [i + 1 for i, _ in enumerate(m.A)])

        if self.use_rule:
            self.model.mysos = pyo.SOSConstraint(rule=rule_mysos, sos=self.sos)
        else:
            # index is not provided, nor are the weights
            self.model.mysos = pyo.SOSConstraint(var=self.model.y, sos=self.sos)

        opt.solve(model, tee=show_output)
        self.assertEqual(len(self.model.mysos), 1)
        self.assertAlmostEqual(pyo.value(model.OBJ), exp_res, self.abs_tol)
        # Can you also check the variable values??

   def test_scalar_with_index(self):
        #...

class TestSOS1_rule(BaseTests, unittest.TestCase):
    sos = 1
    use_rule = True

class TestSOS1_norule(BaseTests, unittest.TestCase):
    sos = 1
    use_rule = False

class TestSOS2_rule(BaseTests, unittest.TestCase):
    sos = 2
    use_rule = True

class TestSOS2_norule(BaseTests, unittest.TestCase):
    sos = 2
    use_rule = False
```

The reason for this is that it is more clear what happened when there is a test failure (in the current implementation, the test fails, but there is no information in the failure as to which case you were actually running)

jsiirola (changes requested):
This is really helpful.  A couple comments and I think it will be good to go.

doc/OnlineDocs/advanced_topics/sos_constraints.rst
```diff
@@ -0,0 +1,300 @@
...
+   # the variables under consideration
+   model.x = pyo.Var(model.A)
+   # the set indexing the sos constraints
+   model.B = pyo.Set()
+   # the sets containing the variable indexes for each constraint
+   model.mysosvarindexset = pyo.Set(model.B)
+   # the set that indexes the variables used in the sos constraints
+   model.C = pyo.Set(within=model.A)
+   # the weights for each variable used in the sos constraints
+   model.mysosweights = pyo.Param(model.C)
```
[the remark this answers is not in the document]
pmlpm1986: I totally agree. I only started using the rule after I began enhancing this documentation. It's much easier.

pyomo/core/base/sos.py
pmlpm1986: Here the 'set of variables' refers to the collection of variables in the mathematical model that will be part of the SOS. I propose to use 'group' to distinguish from the concept of sets (i.e., as in set theory).

jsiirola: "Group" sounds good to me.

doc/OnlineDocs/advanced_topics/sos_constraints.rst
```diff
@@ -0,0 +1,300 @@
...
+A single SOS of type **N** involving all members of a pyomo Var component can 
+be declared in one line:
+
+.. currentmodule:: pyomo.environ
+
+.. doctest::
+   :hide:
+
+   >>> import pyomo.environ as pyo
+   >>> model = pyo.AbstractModel()
```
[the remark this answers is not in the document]
pmlpm1986: I would argue that using `ConcreteModel`s makes it harder to understand what the code is doing, since I have to provide the data in the declaration too. While trying to understand how the SOSConstraint class works, I ultimately relied on `ConcreteModel`s, namely to check which set was actually being used for what, but to provide a high-level explanation to others `AbstractModel`s feel adequate.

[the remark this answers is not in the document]
jsiirola: That works for me.  Please make (at least the first) declaration of the `model` not hidden so that it is clear to users that you are demonstrating with an `AbstractModel`

pyomo/core/base/sos.py
pmlpm1986: I thought the docstring was just a string and wouldn't be tested. Is there a workaround?

jsiirola: We use doctests to test code snippets in all the documentation that is rendered by Sphinx - including both the code examples in the RST files, and the docstrings from rendered source.  This is important to help ensure that the documentation stays current (and correct).  

Looking at the test failures, it appears to mostly be due to repeat definitions (the python environment is preserved across code sections in a single docstring).  You should be able to silence almost all of it by starting each example with a fresh model (e.g., `model = pyo.AbstractModel()`)

jsiirola (changes requested):
One (very minor) correction and we can merge this!

I would still like to rework the tests to avoid the `test_vectors` list and the `case` tree, but I think that what is here is acceptable and that revision of the tests can be deferred to a subsequent PR.

jsiirola (approved):

PR #2895, opened 2023-07-04
Remove old, unused PyNumero code

### LINKED ISSUE #2894: PyNumero relies on deprecated APIs
## Summary

PyNumero relies on 2 functions from `scipy.sparse.sputils` that have been deprecated (moved to a private module).

In addition, PyNumero relies on the NumPy `matrix` subclass, which "is not the recommended way to represent matrices or deal with linear algebra (see https://docs.scipy.org/doc/numpy/user/numpy-for-matlab-users.html)."  This interface should be updated to use regular `ndarray`.

See `_sputils` imports in 
- pyomo/contrib/pynumero/sparse/block_matrix.py
- pyomo/contrib/pynumero/sparse/utils.py

See the warning emitted by `pytest pyomo/contrib/pynumero`:
```
pyomo/contrib/pynumero/interfaces/tests/test_external_pyomo_model.py: 50 warnings
  pyomo/contrib/pynumero/interfaces/tests/test_external_pyomo_model.py:694: PendingDeprecationWarning:

  the matrix subclass is not the recommended way to represent matrices or deal with linear algebra (see https://docs.scipy.org/doc/numpy/user/numpy-for-matlab-users.html). Please adjust your code to use regular ndarray.
```

### Error Message

(see above)

Pyomo version: 6.6.2.dev0
Python version: all (testing on 3.11)
Operating system: all (testing on Linux)
Solver (if applicable): n/a

### Additional information

PR #2897 silenced the `sputils` deprecation messages (so that the import tests would pass)

michaelbynum:
- Remove old, unused PyNumero code
- Remove the use of deprecated SciPy APIs

### Legal Acknowledgement


pyomo/contrib/pynumero/sparse/utils.py
```diff
@@ -1,53 +0,0 @@
...
-#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain
-#  rights in this software.
-#  This software is distributed under the 3-clause BSD License.
-#  ___________________________________________________________________________
-from scipy.sparse.sputils import isscalarlike
-from scipy.sparse import tril, triu, isspmatrix
-
-import numpy as np
-
-
```
jsiirola, line 16: Are we going to regret removing these utilities?

pyomo/contrib/pynumero/sparse/block_matrix.py
```diff
@@ -393,9 +392,8 @@ def tocoo(self, copy=True):
 
         # create pointers for COO matrix (row, col, data)
         data = np.empty(nonzeros, dtype=dtype)
-        idx_dtype = get_index_dtype(maxval=max(shape))
-        row = -np.ones(nonzeros, dtype=idx_dtype)
-        col = -np.ones(nonzeros, dtype=idx_dtype)
+        row = -np.ones(nonzeros, dtype=int)
+        col = -np.ones(nonzeros, dtype=int)
```
jsiirola, line 396: Do we know when NumPy moved to just using `int` as the index dtype? Do we need to update the minimum numpy version?  Doesn't that place an upper limit of ~4B nonzeros?  (Is that a problem?)

jsiirola (commented):
@michaelbynum: the PR is probably fine, but I had a bunch of questions as I was reading through it

pyomo/contrib/pynumero/sparse/utils.py
michaelbynum: I highly doubt it. I don't even know where these were originally used. If we ever need them again, it should be easy to get them from the git history.

pyomo/contrib/pynumero/sparse/block_matrix.py
michaelbynum: I'll have to check on the numpy minimum version. 

A dtype of `int` is really `int64`:
```
In [1]: import numpy as np

In [2]: a = np.zeros(3, dtype=int)

In [3]: a.dtype
Out[3]: dtype('int64')

In [4]: a[0] = 9.1e18

In [5]: a
Out[5]: array([9100000000000000000,                   0,                   0])
```

michaelbynum: Also, these are the row and column indices, so it is possible to have matrices of dimensions 9e18 by 9e18...

michaelbynum: Looks like this works all the way back to numpy 1.13 which is the current minimum version.

jsiirola (approved):

pyomo/contrib/pynumero/sparse/block_matrix.py
jsiirola: Great!  Thanks!

PR #2901, opened 2023-07-07
Add `PyNumeroEvaluationError`

### LINKED ISSUE #2899: Handle evaluation errors in PyNumero
## Summary

Currently, PyNumero's ASL interface handles evaluation errors by asserting the return value of the interface's `eval_g` function [here](https://github.com/Pyomo/pyomo/blob/main/pyomo/contrib/pynumero/asl.py#L426C1-L426C47). To handle this error, e.g. in a solver interface, one would have to catch the error, something like:
```python
try:
    nlp.evaluate_constraints()
except AssertionError:
    # Follow whatever convention the solver uses for eval errors
    return False
```
This is dangerous as it catches any assertion error (e.g. the check for consistent dimensions). I propose we have something like a `PyNumeroEvaluationError` that we raise in this case that can safely be caught and handled by solver interfaces.

Presumably, we could also provide some information about which constraint caused the evaluation error, which would be useful for debugging. Ipopt's AMPL interface does this, although from a quick glance at the ASL evaluation functions it was not obvious to me how this is done.

### Rationale

This is one step towards a CyIpopt interface that handles function evaluation errors in the same way that Ipopt (AMPL interface) does. The other steps are to [handle evaluation errors in CyIpopt](https://github.com/mechmotum/cyipopt/issues/210) and to integrate both of these in our CyIpopt interface.

Robbybp:
## Fixes #2899

Partially.

## Summary/Motivation:
To handle evaluation errors in PyNumero solver interfaces currently, we would have to catch `AssertionErrors`, which seems dangerous.

This is WIP for now as I haven't written any tests.

- Add `PyNumeroEvaluationError` in a new `pyomo.contrib.pynumero.exceptions` module

This error is currently raised in the `AmplInterface` functions, at the same location the old asserts happened. We could imagine a different API where the low-level evaluation functions return some `EvaluationResult` object that contains error information, then it is the `NLP`'s responsibility to raise the exception (currently the `AmplInterface` evaluation methods don't return anything). This would make it easier to return useful debugging information to the NLP, which has access to the Pyomo variables and constraints needed to display useful information. I've punted on trying to do anything like this for now, but don't think the current design precludes doing this in the future.

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/pynumero/asl.py
+++ b/pyomo/contrib/pynumero/asl.py
@@ -11,6 +11,7 @@
 
 from pyomo.common.fileutils import find_library
 from pyomo.common.dependencies import numpy as np
+from pyomo.contrib.pynumero.exceptions import PyNumeroEvaluationError
 import ctypes
 import logging
 import os
@@ -364,7 +365,8 @@ class AmplInterface(object):
         res = self.ASLib.EXTERNAL_AmplInterface_eval_f(
             self._obj, x, self._nx, ctypes.byref(sol)
         )
-        assert res, "Error in AMPL evaluation"
+        if not res:
+            raise PyNumeroEvaluationError("Error in AMPL evaluation")
         return sol.value
 
     def eval_deriv_f(self, x, df):
@@ -373,7 +375,8 @@ class AmplInterface(object):
             x.dtype == np.double
         ), "Error: array type. Function eval_deriv_f expects an array of type double"
         res = self.ASLib.EXTERNAL_AmplInterface_eval_deriv_f(self._obj, x, df, len(x))
-        assert res, "Error in AMPL evaluation"
+        if not res:
+            raise PyNumeroEvaluationError("Error in AMPL evaluation")
 
     def struct_jac_g(self, irow, jcol):
         irow_p = irow.astype(np.intc, casting='safe', copy=False)
@@ -409,7 +412,8 @@ class AmplInterface(object):
         res = self.ASLib.EXTERNAL_AmplInterface_eval_jac_g(
             self._obj, xeval, self._nx, jac_eval, self._nnz_jac_g
         )
-        assert res, "Error in AMPL evaluation"
+        if not res:
+            raise PyNumeroEvaluationError("Error in AMPL evaluation")
 
     def eval_g(self, x, g):
         assert x.size == self._nx, "Error: Dimension mismatch."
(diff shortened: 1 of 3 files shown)


michaelbynum (approved):
I think your proposed (long-term) plan where the low-level functions return error information and the NLP classes raise the exception is a good one.

Robbybp:
I think this is ready (pending any other reviews). I'm not adding a return object from the low-level evaluation functions at this point, but the tests I've added are high-level enough that they should not need to change if/when we do.

pyomo/contrib/pynumero/asl.py
```diff
@@ -364,7 +365,8 @@ def eval_f(self, x):
         res = self.ASLib.EXTERNAL_AmplInterface_eval_f(
             self._obj, x, self._nx, ctypes.byref(sol)
         )
-        assert res, "Error in AMPL evaluation"
+        if not res:
+            raise PyNumeroEvaluationError("Error in AMPL evaluation")
```
mrmundt, line 369: Likely not your responsibility, but is there a way to make this error more explicit? It seems vague in its current state.

mrmundt (approved):
I think this looks good and I like the wrapping. I have one suggestion that may or may not be in your wheelhouse, so feel free to ignore.

pyomo/contrib/pynumero/asl.py
Robbybp: We could say e.g. "AMPL evaluation error in objective function". Or were you thinking of more explicit in some other way?

mrmundt: I was thinking, is there a way to make the error more helpful to the user? Like, is this an error that they can fix (e.g., is it an error in their model)? If so, are there any hints we can give them about _how_ to fix it?

Robbybp: Nope! We have no idea what happened inside ASL. Even if we could narrow down exactly what constraint (or objective) the error occurred in (which I have a vague proposal for in #2899), we would have no idea what expression caused ASL to fail. The best we could do is try to evaluate the Pyomo constraints (which we don't have access to at this layer of the interface) and hope that Pyomo raises an error about the same expression that ASL didn't like.

mrmundt: _Lovely_... in which case, IGNORE this comment.

michaelbynum (approved):

PR #2913, opened 2023-07-18
GAMS: fix pickling models, update tests

jsiirola:
## Summary/Motivation:
This resolves some long-standing issues in the GAMS interface that were not (until recently) being caught by the test harness

## Changes proposed in this PR:
- Implement custom picklable `GAMSSymbolMap`
- Update list of expected GAMS test failures

### Legal Acknowledgement

By contributing to this software project, I have read the [contribution guide](https://pyomo.readthedocs.io/en/stable/contribution_guide.html) and agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #2916, opened 2023-07-19
Resolve kernel pickle/clone errors

jsiirola:
## Summary/Motivation:
This resolves errors cloning Kernel models that used the `noclone` class.  `noclone` appears to not have been safe to pickle / deepcopy for quite a while, but the error was only highlighted by a mostly unrelated test that started failing after updating GAMS.  This PR removes the `noclone` class and replaces it with a deprecated function that mimics the old behavior (but is correctly picklable / clonable).

As part of resolving this, it became apparent that we needed to change the default deepcopy "scope" rules for Components not attached to blocks.  The previous behavior defaulted to `memo['__block_scope__'][None]` (usually "_don't clone_").  This changes the behavior to default to cloning components not attached to blocks.  Global components (like `GlobalSet` instances) will still not be cloned, as they override the deepcopy / pickle semantics to ensure their global nature.

- Deprecate `pyomo.core.kernel.expression.noclone`
- Switch default handling of cloning components not attached to blocks to "always clone"
- Rework kernel's block scoping scheme to use the same memo entry as the core AML's
- Restore the tests marked "expected failure" by #2913

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/component.py
+++ b/pyomo/core/base/component.py
@@ -117,13 +117,17 @@ class _ComponentBase(PyomoObject):
             _scope = memo['__block_scope__']
             _new = None
             tmp = self.parent_block()
+            # "Floating" components should be in scope by default (we
+            # will handle 'global' components like GlobalSets in the
+            # components)
+            _in_scope = tmp is None
             # Note: normally we would need to check that tmp does not
             # end up being None.  However, since clone() inserts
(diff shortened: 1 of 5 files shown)


mrmundt (approved):
I love it when we can remove hacks the same day they go in.

PR #2930, opened 2023-07-31
Improve NaN handling in LP, NL writers

### LINKED ISSUE #2906: NameError in nl writer
## Summary

NL writer gives a confusing NameError `name 'InvalidNumber' is not defined` when writing an nl file with a variable fixed to None. Seems to be a missing import of `InvalidNumber` or something.

```python
import pyomo.environ as pyo
m = pyo.ConcreteModel()
m.x = pyo.Var([1, 2, 3, 4])
m.x[1].fix()
m.x[2] = 1
m.x[3] = 2
m.x[4] = 3
m.con = pyo.Constraint(expr= m.x[1] - ((m.x[2]*m.x[3])*m.x[4]) == 0)
m.write("file.nl")
```
This gives:
```
NameError: name 'InvalidNumber' is not defined
```

Pyomo version: recent main
Python version: 3.8.10

jsiirola:
## Summary/Motivation:
#2906 pointed out that the NaN handling in the NL writer was missing an import.  While this PR does add the missing import, it also added tests to check that the NaN handling was functioning correctly (it wasn't).  This resolves certain consistency issues when processing NaN values in the LP and NL writers.  It also ports certain approaches developed in the LP writer that were more efficient than the ones in the NL writer.  This results in a modest performance improvement in the NL writer.

- standardize the handling of both expression evaluation and fixed variable evaluation in the LP and NL writers
- port use of `defaultdict` from LP writer to NL writer
- improve testing of `nan` values in expressions

### Legal Acknowledgement

### DIFF
--- a/pyomo/repn/linear.py
+++ b/pyomo/repn/linear.py
@@ -42,20 +42,18 @@ from pyomo.core.base.objective import ScalarObjective, _GeneralObjectiveData
 import pyomo.core.kernel as kernel
 from pyomo.repn.util import (
     ExprType,
+    InvalidNumber,
     apply_node_operation,
     complex_number_error,
-    InvalidNumber,
+    nan,
+    sum_like_expression_types,
 )
 
 logger = logging.getLogger(__name__)
 
-nan = float("nan")
-
 _CONSTANT = ExprType.CONSTANT
 _LINEAR = ExprType.LINEAR
 _GENERAL = ExprType.GENERAL
-
-_SumLikeExpression = {SumExpression, LinearExpression, NPV_SumExpression}
 
 
 def _merge_dict(dest_dict, mult, src_dict):
@@ -879,7 +877,7 @@ class LinearRepnVisitor(StreamBasedExpressionVisitor):
     def enterNode(self, node):
         # SumExpression are potentially large nary operators.  Directly
         # populate the result
-        if node.__class__ in _SumLikeExpression:
+        if node.__class__ in sum_like_expression_types:
             return node.args, self.Result()
         else:
             return node.args, []
--- a/pyomo/repn/linear.py
+++ b/pyomo/repn/linear.py
@@ -576,6 +576,12 @@ def _before_native(visitor, child):
     return False, (_CONSTANT, child)
 
 
+def _before_invalid(visitor, child):
+    return False, (
+        _CONSTANT, InvalidNumber(child, "'{child}' is not a valid numeric type")
+    )
+
+
 def _before_complex(visitor, child):
     return False, (_CONSTANT, complex_number_error(child, visitor, child))
 
@@ -734,6 +740,8 @@ def _register_new_before_child_dispatcher(visitor, child):
             dispatcher[child_type] = _before_complex
         else:
             dispatcher[child_type] = _before_native
+    elif child_type in native_types:
+        dispatcher[child_type] = _before_invalid
     elif not child.is_expression_type():
         if child.is_potentially_variable():
             dispatcher[child_type] = _before_var
--- a/pyomo/repn/linear.py
+++ b/pyomo/repn/linear.py
@@ -768,18 +768,8 @@ _before_child_dispatcher = collections.defaultdict(
 # complex number types
 _complex_types = set((complex,))
 
-# Register an initial set of known expression types with the "before
-# child" expression handler lookup table.
-for _type in native_numeric_types:
-    _before_child_dispatcher[_type] = _before_native
 # We do not support writing complex numbers out
 _before_child_dispatcher[complex] = _before_complex
-# general operators
-for _type in _exit_node_handlers:
-    _before_child_dispatcher[_type] = _before_general_expression
-# override for named subexpressions
-for _type in _named_subexpression_types:
-    _before_child_dispatcher[_type] = _before_named_expression
 # Special handling for expr_if and external functions: will be handled
 # as terminal nodes from the point of view of the visitor
 _before_child_dispatcher[Expr_ifExpression] = _before_expr_if
--- a/pyomo/repn/linear.py
+++ b/pyomo/repn/linear.py
@@ -436,7 +436,7 @@ def _handle_expr_if_const(visitor, node, arg1, arg2, arg3):
     _type, _test = arg1
     assert _type is _CONSTANT
     if _test:
-        if _test != _test:
+        if _test != _test or test.__class__ is InvalidNumber:
             # nan
             return _handle_expr_if_nonlinear(visitor, node, arg1, arg2, arg3)
         return arg2
@@ -680,17 +680,26 @@ def _before_named_expression(visitor, child):
 
 
 def _before_expr_if(visitor, child):
+    # We want a _before_expr_if to catch constant test expressions so
+    # that we ONLY evaluate the appropriate branch.
     test, t, f = child.args
     if is_fixed(test):
         try:
-            test = test()
+            test = visitor._eval_expr(test)
         except:
             return True, None
-        subexpr = LinearRepnVisitor(
+        if test.__class__ is InvalidNumber:
+            return True, None
+        subexpr = visitor.__class__(
             visitor.subexpression_cache, visitor.var_map, visitor.var_order
         ).walk_expression(t if test else f)
         if subexpr.nonlinear is not None:
             return False, (_GENERAL, subexpr)
+        # This test is not needed for LINEAR, but is for the derived
+        # QUADRATIC.  As this is the ONLY _before_* specialization
+        # needed for Quadratic, we will handle it here.
+        elif hasattr(subexpr, 'quadratic') and subexpr.quadratic:
+            return False, (ExprType._QUADRATIC, subexpr)
         elif subexpr.linear:
             return False, (_LINEAR, subexpr)
         else:
--- a/pyomo/repn/linear.py
+++ b/pyomo/repn/linear.py
@@ -821,7 +821,10 @@ class LinearRepnVisitor(StreamBasedExpressionVisitor):
 
     def _eval_fixed(self, obj):
         ans = obj()
-        if ans is None or ans != ans:
+        if ans.__class__ not in native_numeric_types:
+            if ans.__class__ is not InvalidNumber:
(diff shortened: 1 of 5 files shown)


pyomo/repn/plugins/nl_writer.py
```diff
@@ -2328,23 +2305,46 @@ def _before_linear(visitor, child):
...
+            else:
+                linear[_id] = arg1
         elif arg.__class__ in native_types:
             const += arg
         else:
-            const += visitor._eval_expr(arg)
+            try:
+                const += visitor._eval_expr(arg)
+            except (ValueError, ArithmeticError):
+                return True, None
```
Robbybp, line 2346: What is the implication of catching these errors and moving on here (or generally in this PR)? What happens when `True, None` is returned here? In the example in #2906, the nl file is produced with `nan` as a constant in it. Surprisingly, sending this nl file to Ipopt solves with `Optimal solution found`. If I edit the model to have an objective, e.g.
```python
import pyomo.environ as pyo 
m = pyo.ConcreteModel()
m.x = pyo.Var([1, 2, 3, 4]) 
m.x[1].fix()
m.x[2] = 1 
m.x[3] = 2 
m.x[4] = 3 
m.con = pyo.Constraint(expr= m.x[1] - ((m.x[2]*m.x[3])*m.x[4]) == 0)
m.obj = pyo.Objective(expr=m.x[1]**2 + m.x[2]**2 + m.x[3]**2 + m.x[4]**2)
m.write("file.nl")

solver = pyo.SolverFactory("ipopt")
solver.solve(m, tee=True)
```
then Ipopt errors with its "Invalid number" exception, which is what I would expect from NaN in the nl file. But I'm not sure whether this is better for the user than an error trying to evaluate or parse an expression. Should this behavior be controlled by `pyomo.repn.util.HALT_ON_EVALUATION_ERROR`?

Robbybp: This is a somewhat rambling review comment. I'm mostly wondering if invalid numbers were always intended to propagate to the nl file.

Robbybp: I tried this with the `nl_v1` writer, and the exact error we get seems to depend on how `None` is encountered. It could be an error like:
```
ValueError: No value for uninitialized NumericValue object x[1]
```
or an error like:
```
TypeError: unsupported operand type(s) for *: 'int' and 'NoneType'
```

Robbybp (commented):
Some minor comments/questions, then the following question on design: Is converting `None` to `NaN` and putting it in the nl file better than raising an error when a value of `None` is encountered?

pyomo/repn/linear.py
```diff
@@ -857,13 +809,27 @@ def __init__(self, subexpression_cache, var_map, var_order):
         self.var_order = var_order
         self._eval_expr_visitor = _EvaluationVisitor(True)
 
+    def _eval_fixed(self, obj):
+        ans = obj()
+        if ans.__class__ not in native_numeric_types:
+            if ans.__class__ is not InvalidNumber:
+                ans = InvalidNumber(nan if ans is None else ans)
+        elif ans != ans:
+            ans = InvalidNumber(nan)
```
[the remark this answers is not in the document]
jsiirola: Yes.  This is a standard way to check for NaN in things like C/C++, and is better than twice faster than `isnan()`

pyomo/repn/plugins/nl_writer.py
```diff
@@ -2328,23 +2305,46 @@ def _before_linear(visitor, child):
     linear = {}
     for arg in child.args:
         if arg.__class__ is MonomialTermExpression:
-            c, v = arg.args
-            if c.__class__ not in native_types:
-                c = visitor._eval_expr(c)
-            if v.fixed:
-                const += c * v.value
```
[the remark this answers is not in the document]
jsiirola: Yes - among several places.

pyomo/repn/plugins/nl_writer.py
```diff
@@ -2328,23 +2305,46 @@ def _before_linear(visitor, child):
...
+            if _id not in var_map:
+                if arg2.fixed:
+                    const += arg1 * visitor._eval_fixed(arg2)
+                    continue
+                var_map[_id] = arg2
+                linear[_id] = arg1
+            elif _id in linear:
+                linear[_id] += arg1
+            else:
+                linear[_id] = arg1
```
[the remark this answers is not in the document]
jsiirola: Yes - this is safe because `_eval_fixed` maps fixed None / NaN values to `InvalidNumber()` objects, which will support all the standard mathematical operations (like `+` and `*`)

pyomo/repn/plugins/nl_writer.py
jsiirola: Some answers:
- returning `(True, None)` from a `_before_*` function tells the walker that it needs to descend into that expression.  So the idea is that if evaluating an expression led to an error / exception, then let this walker descend into it to try and locate / sort out exactly what the problem is.  When the first return value is `True`, the second return value is ignored.
- The goal here is to never emit an invalid number to the NL file, as those generally lead to parse errors - not Ipopt evaluation errors - and parse errors are really hard for a user to track down.  Now, invalid numbers (that are not representable in an NL file) generate errors within Pyomo at write time.
- Resolving the ValueError/TypeError inconsistency is one of the big goals of this refactor.

pyomo/repn/linear.py
```diff
@@ -718,17 +680,26 @@ def _before_named_expression(visitor, child):
...
-        subexpr = LinearRepnVisitor(
+        if test.__class__ is InvalidNumber:
+            return True, None
+        subexpr = visitor.__class__(
             visitor.subexpression_cache, visitor.var_map, visitor.var_order
         ).walk_expression(t if test else f)
         if subexpr.nonlinear is not None:
             return False, (_GENERAL, subexpr)
+        # This test is not needed for LINEAR, but is for the derived
+        # QUADRATIC.  As this is the ONLY _before_* specialization
```
mrmundt, line 699: Long-term question: is there potential for other "before" specializations? If so, maybe it makes more sense to move this.

pyomo/repn/linear.py
```diff
@@ -857,12 +818,64 @@ def __init__(self, subexpression_cache, var_map, var_order):
         self.var_order = var_order
         self._eval_expr_visitor = _EvaluationVisitor(True)
 
+    def _eval_fixed(self, obj):
+        ans = obj.value
+        if ans.__class__ not in native_numeric_types:
```
mrmundt, line 823: This is completely duplicate code to that on L#845-878 (with `obj` instead of `expr`). Worth pulling out to its own function.

pyomo/repn/plugins/nl_writer.py
```diff
@@ -2279,42 +2259,29 @@ def _before_monomial(visitor, child):
...
-        return False, (_CONSTANT, 0)
+        if arg2.fixed:
+            arg2 = visitor._eval_fixed(arg2)
+            if arg2 != arg2:
+                deprecation_warning(
+                    f"Encountered {arg1}*{arg2} in expression tree.  "
+                    "Mapping the NaN result to 0 for compatibility "
+                    "with the nl_v1 writer.  In the future, this NaN "
+                    "will be preserved/emitted to comply with IEEE-754.",
+                    version='6.4.3',
```
mrmundt, line 2277: Just a question: did we deprecate this same functionality multiple times? This and the deprecation warning in L#2311 are almost exactly the same but have different versions.

pyomo/repn/plugins/nl_writer.py
```diff
@@ -2279,42 +2259,29 @@ def _before_monomial(visitor, child):
...
-                version='6.4.3',
-            )
-            _prod = 0
-        return False, (_CONSTANT, _prod)
-
-    # Trap multiplication by 0.
+    # Trap multiplication by 0 and nan.
     if not arg1:
-        return False, (_CONSTANT, 0)
+        if arg2.fixed:
```
mrmundt, line 2269: This is completely duplicate code to that on L#2306-3126. Worth pulling out to its own function.

pyomo/repn/linear.py
mrmundt: Oh, and it's duplicated in `nl_writer.py`. So even more relevant suggestion.

mrmundt (changes requested):
There are two relevant refactor opportunities here.

pyomo/repn/plugins/nl_writer.py
jsiirola: Good catch.  I copied logic from the lp_writer but forgot to update the deprecation version.  Fixed.

pyomo/repn/linear.py
jsiirola: I caught that and hope to further refactor the LP / NL writers to remove redundant code / logic between them (but that will be a separate PR after the next release)

Robbybp (changes requested):
I still get `Ipopt::TNLP::INVALID_TNLP` due to `None` in the nl file when I run the following code, adapted from #2906.
```python
import pyomo.environ as pyo
m = pyo.ConcreteModel()
m.x = pyo.Var([1, 2, 3, 4])
m.x[1].fix()
m.x[2] = 1
m.x[3] = 2
m.x[4] = 3
m.con = pyo.Constraint(expr= m.x[1] - ((m.x[2]*m.x[3])*m.x[4]) == 0)
solver = pyo.SolverFactory("ipopt")
solver.solve(m, tee=True)
```
Is this intended as we never need to evaluate the constant term `m.x[1]`? My guess is that this is not intentional.

pyomo/repn/plugins/nl_writer.py
Robbybp: > * returning `(True, None)` from a `_before_*` function tells the walker that it needs to descend into that expression.  So the idea is that if evaluating an expression led to an error / exception, then let this walker descend into it to try and locate / sort out exactly what the problem is.

How does the walker know to raise the error? And where does it get raised now?

mrmundt (approved):

jsiirola:
@Robbybp: your model is not fixed by this PR, but it is by #2953 (so the combination of this PR and that PR will result in a reasonable exception/error message)

PR #2972, opened 2023-08-25
Fixes #2945: Add Maximize() and Minimize() instead of Objective(sense=...)

### LINKED ISSUE #2945: Add Maximize() and Minimize() instead of Objective(sense=...)
## Summary

Instead of writing:
```python
@model.Objective(sense=pye.maximize)
def objective(m):
     return ...
```

suggest to write:
```python
@model.Maximize()
def objective(m):
     return ...
```

Similarly for `Minimize()`.

### Rationale

 * Less typing, fewer things to remember.
 * Clear intention (no need to remember whether maximize or minimize is the default sense).

### Alternatives

I could create my own wrapper:

```python
def Maximize(*args, **kwargs):
   return Objective(*args, **kwargs, sense=pye.maximise)
```

However, this would not handle the nice decorator syntax. It would also only benefit myself, while the target audience of this change is new users.

### Additional information

I'm happy to submit a PR. This should be a very simple wrapper around `Objective' and a trivial documentation update. The traditional way using `Objective` will keep working and does not need to be deprecated.

MLopez-Ibanez:
## Fixes # .

Fixes #2945 Add Maximize() and Minimize() instead of Objective(sense=...)

## Summary/Motivation:

Instead of writing:
```python
@model.Objective(sense=pye.maximize)
def objective(m):
     return ...
```

suggest to write:
```python
@model.Maximize()
def objective(m):
     return ...
```

Similarly for `Minimize()`.

- Add `Maximize()`and `Minimize()`

### Legal Acknowledgement


MLopez-Ibanez:
All checks passed? That doesn't sound right...

mrmundt:
> All checks passed? That doesn't sound right...

They did not run because this is marked a draft.

jsiirola:
I just gave this a quick glance, and have some comments before you spend too much time on this:
- Please to not change existing tests (unless you are changing / breaking functionality
- If you are going to register `Maximize` / `Minimize` with the `ModelComponentFactory`, then I think they need to be proper classes and not just functions.  Consumers of the factory are allowed to assume that the Factory is a registry of classes.

MLopez-Ibanez:
> I just gave this a quick glance, and have some comments before you spend too much time on this:
> 
>     * Please to not change existing tests (unless you are changing / breaking functionality

OK, I will duplicate the tests.

>     * If you are going to register `Maximize` / `Minimize` with the `ModelComponentFactory`, then I think they need to be proper classes and not just functions.  Consumers of the factory are allowed to assume that the Factory is a registry of classes.

Is it better to make them proper classes inheriting from Objective or use a different mechanism and keep them as thin functions?

MLopez-Ibanez:
Converting to draft to avoid triggering the CI until this is ready...

MLopez-Ibanez:
> * If you are going to register `Maximize` / `Minimize` with the `ModelComponentFactory`, then I think they need to be proper classes and not just functions.  Consumers of the factory are allowed to assume that the Factory is a registry of classes.

Making `Maximize` / `Minimize` into classes turns out to be quite difficult due to the way that `Objective` plays tricks with `__new__`.

PR #2973, opened 2023-08-25
Fix bugs in gdp.bound_pretransformation

emma58:
## Summary/Motivation:

This fixes three bugs in `gdp.bound_pretransformation`: The first has to do with the fact that, since `0*some_var` stays in expressions, it's possible to encounter trivial constraints (with no variables) that this transformation should ignore. The second is more nefarious: We were not correctly handling negative coefficients in bounds constraints, so we were wrongly flipping the bounds. The third is even worse: We were accidentally assuming that all Disjunctions in nested GDP were only children, as well as assuming that all nested indicator variables are local. This rewrites the transformation to write a constraint per Disjunction in that Disjunction's scope, with the tightest bounds from the hierarchy.

- Rewrites transformation to actually work for nested GDPs
- Fixes bugs where we flip bounds when we encounter bounds constraints with negative coefficients
- Catches trivial constraints and ignores them
- Adds some tests

### Legal Acknowledgement


emma58:
There's still a bug, so marking this as a WIP for now...

pyomo/gdp/plugins/bound_pretransformation.py
```diff
@@ -230,77 +233,99 @@ def _update_bounds_dict(self, v_bounds, lower, upper, disjunct, gdp_forest):
...
+                deactivate_lower = set()
+                deactivate_upper = set()
+                for disj in disjunction.disjuncts:
+                    (lb, ub) = self._get_tightest_ancestral_bounds(
+                        v_bounds, disj, gdp_forest
+                    )
+                    if lb is None:
+                        all_lbs = False
+                    if ub is None:
+                        all_ubs = False
```
jsiirola, line 272: Is it worth breaking out of this loop early if both `all_lbs` and `all_ubs` are False?  Something like
```python
if lb is None:
    all_lbs = False
    if not all_ubs:
        break
if ub is None:
    all_ubs = False
    if not all_lbs:
        break
```

jsiirola (approved):
A couple comments, but nothing that must prevent merging

pyomo/gdp/plugins/bound_pretransformation.py
```diff
@@ -164,54 +166,55 @@ def _update_bounds_from_constraints(
...
-            # (which we know if the tightest of the ancestral values because
-            # we've already come down the tree once.)
-            lb = parent_lb
-        if ub is None:
-            ub = parent_ub
-        return (lb, ub)
+                    if disjunct in v_bounds['to_deactivate']:
+                        v_bounds['to_deactivate'][disjunct].add(constraint)
+                    else:
+                        v_bounds['to_deactivate'][disjunct] = set([constraint])
```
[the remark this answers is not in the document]
emma58: Oh, whoops, good catch. I thought it didn't matter because it was just the order things get deactivated, but that's not true because I have to add constraints in that process sometimes when I'm deactivating the <= side but not the >= side or vice versa.

pyomo/gdp/plugins/bound_pretransformation.py
emma58: Oh, it probably is. In the absence of performance tests, I'm going to do it for now.

emma58:
This should be ready to merge, but I can't convince that Windows test to not time out, and I think the Jenkins failure is unrelated? @mrmundt @jsiirola

PR #2979, opened 2023-08-30
Prevent multiple applications of the scaling transform

jsiirola:
## Summary/Motivation:
@andrewlee94 noted that the scaling transform caused problems because the `scaling_factor` Suffix was not deactivated after applying the transformation, which caused the resulting model to be doubly-scaled when written out to Ipopt.  This updates the transformation to deactivate the `scaling_factor` Suffixes on the transformed model at the conclusion of the transformation.

In addition, this PR refactors the logic for finding the suffixes into a general-purpose `pyomo.core.base.suffix.SuffixFinder` class (so that it can easily be reused elsewhere, e.g. in GDP).  It also removes the require ment that the scaling factor suffix values are forcibly converted to floats so that clients can use other data structures (in particular, mutable Params).

- Refactor the scaling transformation to promote the suffix finder as a general utility
- Deactivate `scaling_factor` suffixes after the transformation
- Remove the restriction that all scaling factors are casted to `float` before using them to update constraints.

### Legal Acknowledgement

### DIFF
--- a/examples/doc/samples/case_studies/network_flow/README.txt
+++ b/examples/doc/samples/case_studies/network_flow/README.txt
@@ -109,7 +109,7 @@ def supplyDemandRule(nn, model):
 model.supplyDemand = Constraint(model.places, rule=supplyDemandRule)
 }}}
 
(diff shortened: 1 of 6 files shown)


andrewlee94 (approved):
Looks good.

mrmundt (approved):

PR #2984, opened 2023-08-31
Latex printer

codykarcher:
## Fixes # 752
https://github.com/Pyomo/pyomo/issues/752

## Summary/Motivation:
Latex is good, Pyomo is good, together are great

Adds a util function that prints out the latex of various pyomo objects.  Docs and unit tests have also been included

### Legal Acknowledgement


pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,619 @@
...
+    expr, indices = templatize_rule(expr.parent_block(), expr._rule, expr.index_set())
+    return (expr, indices)
+
+
+def templatize_passthrough(con):
+    return (con, [])
+
+
+def handle_negation_node(node, arg1):
+    childPrecedence = []
```
mrmundt, line 59: This is a general quibble for the whole file: it's PEP8-standard to use snake case (`child_precedence`) for variables: https://peps.python.org/pep-0008/#function-and-variable-names

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,619 @@
...
+def handle_monomialTermExpression_node(node, arg1, arg2):
+    if arg1 == '1':
+        return arg2
+    elif arg1 == '-1':
+        return '-' + arg2
+    else:
+        return arg1 + arg2
+
+
+def handle_named_expression_node(node, arg1):
```
mrmundt, line 222: Is this necessary?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,619 @@
...
+    ## Raises not implemented error
+    # raise NotImplementedError('Expr_if objects not supported by the Latex Printer')
+
+    ## Puts cases in a bracketed matrix
+    # pstr = ''
+    # pstr += '\\begin{Bmatrix} '
+    # pstr += arg2 + ' , & ' + arg1 + '\\\\ '
+    # pstr += arg3 + ' , & \\text{otherwise}' + '\\\\ '
+    # pstr += '\\end{Bmatrix}'
+    # return pstr
```
mrmundt, line 242: Remove?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,619 @@
...
+        }
+
+    def exitNode(self, node, data):
+        return self._operator_handles[node.__class__](node, *data)
+
+
+def latex_printer(
+    pyomoElement, filename=None, useAlignEnvironment=False, splitContinuousSets=False
+):
+    '''
```
mrmundt, line 332: This docstring will need to follow a standard supported by `sphinx.ext.napoleon`: https://www.sphinx-doc.org/en/master/usage/extensions/napoleon.html

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,619 @@
...
+from pyomo.core.base.objective import ScalarObjective, _GeneralObjectiveData
+import pyomo.core.kernel as kernel
+from pyomo.core.expr.template_expr import (
+    GetItemExpression,
+    GetAttrExpression,
+    TemplateSumExpression,
+    IndexTemplate,
+)
+from pyomo.core.expr.template_expr import Numeric_GetItemExpression
+from pyomo.core.expr.template_expr import templatize_constraint, resolve_template
```
mrmundt, line 31: These (plus L#34) should all be in the same import statement

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,619 @@
+import pyomo.environ as pe
```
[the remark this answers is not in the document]
mrmundt: Following up on this, though, since you have a finite set of objects being imported, it's probably better for maintainability long-term to just import said objects instead of all of `pyomo.environ`.

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,619 @@
...
+
+    splitContinuous: Default behavior has all sum indices be over "i \in I" or similar.  Setting this flag to
+        True makes the sums go from: \sum_{i=1}^{5} if the set I is continuous and has 5 elements
+
+    '''
+
+    # Various setup things
+    isSingle = False
+    if not isinstance(pyomoElement, _BlockData):
+        if isinstance(pyomoElement, pe.Objective):
```
mrmundt, line 352: For maintainability, why not do something like this:
```python
objectives = []
constraints = []
expressions = []

if isinstance(pyomo_element, pyo.Objective):
    objectives.append(pyomo_element)
    templatize_fcn = templatize_constraint

... (etc.)
```

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,619 @@
...
+        else:
+            pstr += '\n'
+
+    # Iterate over the constraints
+    if len(constraints) > 0:
+        # only print this if printing a full formulation
+        if not isSingle:
+            pstr += ' ' * tbSpc + '& \\text{subject to} \n'
+
+        # first constraint needs different alignment because of the 'subject to'
```
mrmundt, line 435: "Subject to"? This comment confuses me

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,619 @@
...
+        objective and constraint.  Each objective and constraint will be labeled with its name in the pyomo model.
+        This flag is only relevant for Models and Blocks.
+
+    splitContinuous: Default behavior has all sum indices be over "i \in I" or similar.  Setting this flag to
+        True makes the sums go from: \sum_{i=1}^{5} if the set I is continuous and has 5 elements
+
+    '''
+
+    # Various setup things
+    isSingle = False
```
mrmundt, line 350: I know that `isSingle` is a bool but I don't actually know why it's necessary / what it's doing. What does `Single` mean in this context?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,619 @@
...
+            con_template, indices = templatize_fcn(con)
+
+            # Walk the constraint
+            conLine = (
+                ' ' * tbSpc + algn + ' %s ' % (visitor.walk_expression(con_template))
+            )
+
+            # Multiple constraints are generated using a set
+            if len(indices) > 0:
+                nm = indices[0]._set
```
mrmundt, line 458: What does `nm` stand for?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,619 @@
...
+                gp = indices[0]._group
```
mrmundt, line 459: I assume `gp` stands for `group` but is there an issue with using `group`?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,619 @@
...
+
+                ixTag = '__INDEX_PLACEHOLDER_8675309_GROUP_%s_%s__' % (gp, nm)
```
mrmundt, line 461: Recommend: `ixTag` -> `idxTag`

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,619 @@
...
+                stTag = '__SET_PLACEHOLDER_8675309_GROUP_%s_%s__' % (gp, nm)
```
mrmundt, line 462: Recommend: `stTag` -> `setTag`

pyomo/util/tests/test_latex_printer.py
```diff
@@ -0,0 +1,349 @@
...
+#  This software is distributed under the 3-clause BSD License.
+#  ___________________________________________________________________________
+
+import pyomo.common.unittest as unittest
+from pyomo.util.latex_printer import latex_printer
+import pyomo.environ as pe
+
+
+def generate_model():
+    import pyomo.environ as pe
```
mrmundt, line 18: No need to re-import

pyomo/util/tests/test_latex_printer.py
```diff
@@ -0,0 +1,349 @@
...
+
+    m.objective_2 = pe.Objective(rule=ruleMaker)
+
+    m.objective_3 = pe.Objective(expr=m.x + m.y + m.z, sense=-1)
+
+    return m
+
+
+def generate_simple_model():
+    import pyomo.environ as pe
```
mrmundt, line 72: No need to re-import

pyomo/util/tests/test_latex_printer.py
```diff
@@ -0,0 +1,349 @@
...
+    def ruleMaker(m):
+        return sum(m.p[k] for k in m.K) == 1
+
+    m.constraint_8 = pe.Constraint(rule=ruleMaker)
+
+    return m
+
+
+def generate_simple_model_2():
+    import pyomo.environ as pe
```
mrmundt, line 103: No need to re-import

pyomo/util/tests/test_latex_printer.py
```diff
@@ -0,0 +1,349 @@
...
+        <= m.y_sub1_sub2_sub3
+    )
+    m.constraint_2 = pe.Constraint(
+        expr=(m.x_dot + m.x_bar) ** -(m.x_star + m.x_hat) <= m.y_sub1_sub2_sub3
+    )
+    m.constraint_3 = pe.Constraint(
+        expr=-(m.x_dot + m.x_bar) + -(m.x_star + m.x_hat) <= m.y_sub1_sub2_sub3
+    )
+
+    return m
```
mrmundt, line 124: These can be inside the test class, FYI.

pyomo/util/tests/test_latex_printer.py
```diff
@@ -0,0 +1,349 @@
...
+
+class TestLatexPrinter(unittest.TestCase):
+    def test_latexPrinter_objective(self):
+        m = generate_model()
+        pstr = latex_printer(m.objective_1)
+        bstr = ''
+        bstr += '\\begin{equation} \n'
+        bstr += '    & \\text{minimize} \n'
+        bstr += '    & & x + y + z \n'
+        bstr += '\end{equation} \n'
```
mrmundt, line 135: You can do all of these as a multi-line string:
```python
bstr = """
\\begin{equation} \n
    & \\text{minimize} \n
....

"""
```

pyomo/util/tests/test_latex_printer.py
```diff
@@ -0,0 +1,349 @@
...
+        bstr += '    & \\text{minimize} \n'
+        bstr += '    & & x + y \label{obj:basicFormulation_objective_1} \\\\ \n'
+        bstr += '    & \\text{subject to} \n'
+        bstr += '    & & x^{2} + y^{2} \leq 1 \label{con:basicFormulation_constraint_1} \\\\ \n'
+        bstr += '    &&& 0 \leq x \label{con:basicFormulation_constraint_2} \\\\ \n'
+        bstr += '    &&&  \left( x + y \\right) \sum_{i = 1}^{5} v_{i} + u_{i,j}^{2} \leq 0 , \quad j \\in I \label{con:basicFormulation_constraint_7} \\\\ \n'
+        bstr += '    &&& \sum_{k \in K} p_{k} = 1 \label{con:basicFormulation_constraint_8} \n'
+        bstr += '\end{align} \n'
+        self.assertEqual(pstr, bstr)
+
```
mrmundt, line 328: Learn some more Pyomo funness! You can write a test to ensure that it outputs to a file using the `TempfileManager`.

mrmundt (changes requested):
I warned you in advance of all the comments... ;)

doc/OnlineDocs/model_debugging/latex_printing.rst
```diff
@@ -0,0 +1,134 @@
...
+--------
+
+A Model
++++++++
+
+.. doctest::
+
+    >>> # Note: this model is not mathematically sensible
+
+    >>> import pyomo.environ as pe
```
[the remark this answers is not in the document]
codykarcher: Fixed in coming commit

pyomo/util/latex_printer.py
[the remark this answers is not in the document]
codykarcher: Fixed in coming commit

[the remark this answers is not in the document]
codykarcher: I am rejecting the direct import of objects to keep explicit namespace clarity.  If someone feels strong enough to overrule me, I could be convinced.

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,619 @@
...
+
+from pyomo.core.base.block import _BlockData
+
+from pyomo.repn.util import ExprType
+
+_CONSTANT = ExprType.CONSTANT
+_MONOMIAL = ExprType.MONOMIAL
+_GENERAL = ExprType.GENERAL
+
+# see:  https://github.com/Pyomo/pyomo/blob/main/pyomo/repn/plugins/nl_writer.py
```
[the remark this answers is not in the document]
codykarcher: I didn't want to delete this, but I did.

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,619 @@
...
+                childPrecedence.append(-1)
+            else:
+                childPrecedence.append(a.PRECEDENCE)
+        else:
+            childPrecedence.append(-1)
+
+    if hasattr(node, 'PRECEDENCE'):
+        precedence = node.PRECEDENCE
+    else:
+        # Should never hit this
```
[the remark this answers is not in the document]
codykarcher: Raising an error in coming commit

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,619 @@
...
```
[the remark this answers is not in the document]
codykarcher: Raising an error in coming commit

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,619 @@
...
+        arg1 = ' \\left( ' + arg1 + ' \\right) '
+
+    if childPrecedence[1] > precedence:
+        arg2 = ' \\left( ' + arg2 + ' \\right) '
+
+    return ''.join([arg1, arg2])
+
+
+def handle_division_node(node, arg1, arg2):
+    # return '/'.join([arg1,arg2])
```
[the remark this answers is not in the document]
codykarcher: This could be a valid alternative, but is not correct given the precedence issues.  Deleting in coming commit

pyomo/util/latex_printer.py
codykarcher: Yes, need to handle this type

pyomo/util/latex_printer.py
codykarcher: No I want to keep this because of the ambiguity of how to handle this.  Someone in the future may wish to look at this

pyomo/util/latex_printer.py
codykarcher: No

codykarcher: Seriously though, I use camel case a lot because the _ character causes additional stress on my hands while typing and so I strongly prefer not to use it as much as possible. I do sometimes use it when I need a clear separation in my variable names, but otherwise, I strongly prefer not to do this

pyomo/util/latex_printer.py
codykarcher: Added a comment to clarify this function is needed

pyomo/util/latex_printer.py
codykarcher: Fixed

pyomo/util/latex_printer.py
codykarcher: Fixed

pyomo/util/latex_printer.py
codykarcher: Corrected to avoid the double if, but avoiding the use of the slow `append` function

pyomo/util/latex_printer.py
codykarcher: added a comment that clarifies this:
```python
        # first constraint needs different alignment because of the 'subject to':
        # & minimize   & & [Objective]
        # & subject to & & [Constraint 1]
        # &            & & [Constraint 2]
        # &            & & [Constraint N]

        # The double '& &' renders better for some reason
```

pyomo/util/latex_printer.py
codykarcher: Added a comment.  `isSingle` flags true for a constraint, expression, or objective.  It flags false if it is a model or block.  This helps correct behavior for a single constraint, expression, or objective.

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,619 @@
...
+        precedence = node.PRECEDENCE
+    else:
+        # Should never hit this
+        precedence = -1
+
+    if childPrecedence[0] > precedence:
+        arg1 = ' \\left( ' + arg1 + ' \\right) '
+
+    if childPrecedence[1] > precedence:
+        arg2 = ' \\left( ' + arg2 + ' \\right) '
```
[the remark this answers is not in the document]
codykarcher: Fixed

pyomo/util/latex_printer.py
codykarcher: it stands for set number.  Open to suggestions

codykarcher: Deleted the intermediate 'nm' and 'gp' variables

pyomo/util/latex_printer.py
codykarcher: Deleted the intermediate 'nm' and 'gp' variables

pyomo/util/latex_printer.py
codykarcher: Fixed

pyomo/util/latex_printer.py
codykarcher: fixed

pyomo/util/tests/test_latex_printer.py
codykarcher: Fixed

pyomo/util/tests/test_latex_printer.py
codykarcher: I prefer to have the generate model function be copy-paste-able and believe this import is of minimal overhead

pyomo/util/tests/test_latex_printer.py
codykarcher: I prefer to have the generate model function be copy-paste-able and believe this import is of minimal overhead

pyomo/util/tests/test_latex_printer.py
codykarcher: I prefer to have the generate model function be copy-paste-able and believe this import is of minimal overhead

pyomo/util/tests/test_latex_printer.py
codykarcher: I prefer them here, but would be willing to move if others feel strongly.

pyomo/util/latex_printer.py
emma58: You might include a comment that the way to handle this in the future might be to define a function of the inputs using `cases` and then substitute that function in place of the exprIf wherever it appears.

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,619 @@
...
+            Numeric_GetItemExpression: handle_numericGIE_node,
+            TemplateSumExpression: handle_templateSumExpression_node,
+        }
+
+    def exitNode(self, node, data):
+        return self._operator_handles[node.__class__](node, *data)
+
+
+def latex_printer(
+    pyomoElement, filename=None, useAlignEnvironment=False, splitContinuousSets=False
```
emma58, line 330: I think at least for user-facing variables, using snake case would be nice for the sake of consistency. (And again, I'd advocate for `pyomo_component`)

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,619 @@
...
+        else:
+            tbSpc = 4
+
+    # Iterate over the objectives and print
+    for obj in objectives:
+        obj_template, obj_indices = templatize_fcn(obj)
+        if obj.sense == 1:
+            pstr += ' ' * tbSpc + '& \\text{%s} \n' % ('minimize')
+        else:
+            pstr += ' ' * tbSpc + '& \\text{%s} \n' % ('maximize')
```
emma58, line 419: I'd prefer using `\min` and `\max` in the latex here.

pyomo/util/tests/test_latex_printer.py
codykarcher: Fixed

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,619 @@
...
+        if not isSingle:
+            pstr += '\\\\ \n'
+        else:
+            pstr += '\n'
+
+    # Iterate over the constraints
+    if len(constraints) > 0:
+        # only print this if printing a full formulation
+        if not isSingle:
+            pstr += ' ' * tbSpc + '& \\text{subject to} \n'
```
emma58, line 433: I'd prefer `\text{s.t.}` in the latex here

pyomo/util/latex_printer.py
emma58: Maybe rename this variable to be `singleEquation` or something like that then?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,619 @@
...
+        'N',
+        'P',
+        'Q',
+        'R',
+        'S',
+        'T',
+        'U',
+        'V',
+        'W',
+    ]
```
emma58, line 503: What happens if you run out of names in this list?

emma58 (commented):
@codykarcher, thanks so much for doing this (and at the speed of light too!).

Some design questions around how we name variables, and a lot of small comments. I haven't actually played with it yet, so there's probably more coming. :)

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,871 @@
...
+            vr
+            for vr in pyomoElement.component_objects(
+                pyo.Var, descend_into=True, active=True
+            )
+        ]
+        if xOnlyMode == 1:
+            newVariableList = ['x' for i in range(0, len(variableList))]
+            for i in range(0, len(variableList)):
+                newVariableList[i] += '_' + str(i + 1)
+            overwriteDict = dict(zip([v.name for v in variableList], newVariableList))
```
emma58, line 590: Wondering if this dictionary should really be a `ComponentMap` and that the keys should by pyomo components (rather than their names). Also, do you think we could support naming Sets and Params out of it as well? Sets are the hard ones because there's what you call the set, and then what you use for indices. But perhaps to start we could have people specify a list of index names and if you run out you can just yell at them. Thinking of cases like:
```
@m.Constraint(m.TIME)
def something(m, t):
    return sum(m.some_var[t_prime] for t_prime in m.TIME if t_prime < t) == 1
```
You'd want `overwriteDict[m.TIME] = ('\mathcal T', ['t', "t'"])` or something like that...

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,871 @@
...
+                idxTag = '__INDEX_PLACEHOLDER_8675309_GROUP_%s_%s__' % (
+                    indices[0]._group,
+                    indices[0]._set,
+                )
+                setTag = '__SET_PLACEHOLDER_8675309_GROUP_%s_%s__' % (
+                    indices[0]._group,
+                    indices[0]._set,
+                )
+
+                conLine += ', \\quad %s \\in %s ' % (idxTag, setTag)
```
emma58, line 708: Could we add a `\forall` before idxTag?

emma58 (commented):
A few more thoughts after actually playing around with it this morning...

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,871 @@
...
+        else:
+            childPrecedence.append(-1)
+
+    if hasattr(node, 'PRECEDENCE'):
+        precedence = node.PRECEDENCE
+    else:
+        # Should never hit this
+        raise RuntimeError(
+            'This error should never be thrown, node does not have a precedence.  Report to developers'
+        )
```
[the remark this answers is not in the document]
codykarcher: fixed

fleimgruber:
@codykarcher  As an early [subscriber](https://github.com/Pyomo/pyomo/issues/752#issuecomment-466622938) to #752 I was happy to see this attempt upstream! In the meantime we rolled our own internal version of this and use it on and off. So of course I was curious and tried this PR with one of our test cases (never mind the actually infeasible and silly model) and saw this error (below test case), hope it helps:

```python
import pyomo.environ as pyo
from pyomo.util.latex_printer import latex_printer


def create_model() -> pyo.ConcreteModel:
    model = pyo.ConcreteModel("Example Model")

    model.years = pyo.Set(initialize=[2030, 2040])
    model.strings = pyo.Set(initialize=["a", "b"])

    model.p = pyo.Param(model.years, initialize={2030: 1, 2040: 2}, mutable=False)
    model.pp = pyo.Param(model.strings, model.years, initialize={("a", 2030): 1, ("b", 2040): 2})

    model.x_etc = pyo.Var(model.years, domain=pyo.NonNegativeReals)

    def obj(m):
        return sum(
            m.p[j_st] * m.pp["a", j_st] * m.x_etc[j_st] for j_st in m.years if j_st == m.p[2030] and j_st != 2030
        ) - (m.p[2040] if m.p[2030] not in [1] else 0)

    model.OBJ = pyo.Objective(rule=obj)

    return model


latex_printer(create_model())
```

Error:

```
  File "/home/user/dev/pyomo_latex_printer/latex_printer_upstream.py", line 35, in <module>
    latex_printer(model)
  File "/home/user/dev/pyomo/pyomo/util/latex_printer.py", line 671, in latex_printer
    obj_template, obj_indices = templatize_fcn(obj)
                                ^^^^^^^^^^^^^^^^^^^
  File "/home/user/dev/pyomo/pyomo/core/expr/template_expr.py", line 1178, in templatize_constraint
    expr, indices = templatize_rule(con.parent_block(), con.rule, con.index_set())
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/user/dev/pyomo/pyomo/core/expr/template_expr.py", line 1166, in templatize_rule
    "templatizing the rule '%s':\n\t%s" % (rule.name, internal_error[1])
                                           ^^^^^^^^^
AttributeError: 'ScalarCallInitializer' object has no attribute 'name'
```

mrmundt:
@fleimgruber - that is insidious. That is an error that occurs while trying to raise a separate error. Thanks for being an early tester!

mrmundt:
@fleimgruber - I ran your code (with a silly workaround to get past the initial error), and this is what prints out:

```
PyomoException: Cannot convert non-constant Pyomo expression (_1  ==  1) to bool.
This error is usually caused by using a Var, unit, or mutable Param in a
Boolean context such as an "if" statement, or when checking container
membership or equality. For example,
    >>> m.x = Var()
    >>> if m.x >= 1:
    ...     pass
and
    >>> m.y = Var()
    >>> if m.y in [m.x, m.y]:
    ...     pass
would both cause this exception.

TemplateExpressionError                   Traceback (most recent call last)
Cell In[1], line 26
     21     model.OBJ = pyo.Objective(rule=obj)
     23     return model
---> 26 latex_printer(create_model())

File ~/Documents/idaes/cody-pyomo/pyomo/util/latex_printer.py:671, in latex_printer(pyomo_component, filename, use_align_environment, split_continuous_sets, use_smart_variables, x_only_mode, use_short_descriptors, use_forall, overwrite_dict)
    669 # Iterate over the objectives and print
    670 for obj in objectives:
--> 671     obj_template, obj_indices = templatize_fcn(obj)
    672     if obj.sense == 1:
    673         pstr += ' ' * tbSpc + '& %s \n' % (descriptorDict['minimize'])

File ~/Documents/idaes/cody-pyomo/pyomo/core/expr/template_expr.py:1178, in templatize_constraint(con)
   1177 def templatize_constraint(con):
-> 1178     expr, indices = templatize_rule(con.parent_block(), con.rule, con.index_set())
   1179     if expr.__class__ is tuple:
   1180         expr = tuple_to_relational_expr(expr)

File ~/Documents/idaes/cody-pyomo/pyomo/core/expr/template_expr.py:1168, in templatize_rule(block, rule, index_set)
   1163         if internal_error is not None:
   1164             logger.error(
   1165                 "The following exception was raised when "
   1166                 "templatizing the rule '%s':\n\t%s" % (rule, internal_error[1])
   1167             )
-> 1168         raise TemplateExpressionError(
   1169             None,
   1170             "Explicit iteration (for loops) over Sets is not supported "
   1171             "by template expressions.  Encountered loop over %s"
   1172             % (context.cache[-1][0]._set,),
   1173         )
   1174 return None, indices

TemplateExpressionError: Explicit iteration (for loops) over Sets is not supported by template expressions.  Encountered loop over years

```
This is an issue in how the `latex_printer` iterates, so thanks for the info! It exposes a real problem on our end. We appreciate it.

codykarcher:
@jsiirola I think this bug needs a quick look from you to triage if its in my code or your code

fleimgruber:
@mrmundt happy to hear it helped!

codykarcher:
@fleimgruber Would you mind posting the expected latex output of your test case?  That way I can verify my fixes.

codykarcher:
Enough issues have come up, and enough design decisions have been raised that I've converted this to a draft pending a team discussion

fleimgruber:
@codykarcher I am on the road at the moment so this is a quick one. I am taking some time off and will be back in two weeks. Please note that our solution works directly on the Python AST using https://docs.python.org/3/library/ast.html. This is the LaTeX that is generated for the objective function:

```tex
\operatorname{sum}\left(\text{p}_{j\_st} \cdot \text{pp}_{\text{a}, j\_st} \cdot \bm{x\_etc}_{j\_st}\text{ for }j\_st\text{ in }years\text{ if }j\_st = \text{p}_{2030}\text{ and }j\_st \neq 2030\right) - \left(\text{p}_{2040}\text{ if }\text{p}_{2030} \in 1\text{ else }0\right)
```

which renders to:

image

doc/OnlineDocs/model_debugging/latex_printing.rst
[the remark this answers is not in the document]
mrmundt: Poke poke on this one again!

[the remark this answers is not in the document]
codykarcher: Missed this, fixed now

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+
+    # is Single implies Objective, constraint, or expression
+    # these objects require a slight modification of behavior
+    # isSingle==False means a model or block
+
+    if latex_component_map is None:
+        latex_component_map = ComponentMap()
+        existing_components = ComponentSet([])
+    else:
+        existing_components = ComponentSet(list(latex_component_map.keys()))
```
emma58, line 715: Can't this just be `existing_components = latex_component_map.keys()`?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+            for obj in pyomo_component.component_data_objects(
+                pyo.Objective, descend_into=True, active=True
+            )
+        ]
+        constraints = [
+            con
+            for con in pyomo_component.component_objects(
+                pyo.Constraint, descend_into=True, active=True
+            )
+        ]
```
emma58, line 835: (I think) you probably want to add `sort=SortComponents.deterministic` to these `component_data_objects` calls

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+                pyo.Param, descend_into=True, active=True
+            )
+        ]
+
+        setList = [
+            st
+            for st in pyomo_component.component_objects(
+                pyo.Set, descend_into=True, active=True
+            )
+        ]
```
emma58, line 894: You probably need to sort these as well, for determinism?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+
+    descriptorDict = {}
+    if use_short_descriptors:
+        descriptorDict['minimize'] = '\\min'
+        descriptorDict['maximize'] = '\\max'
+        descriptorDict['subject to'] = '\\text{s.t.}'
+        descriptorDict['with bounds'] = '\\text{w.b.}'
```
emma58, line 901: I don't think we need 'w.b' or 'with bounds'. These are constraints and can just be in the same block of latex as the other constraints.

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+    # In the case where just a single expression is passed, add this to the constraint list for printing
+    constraints = constraints + expressions
+
+    # Declare a visitor/walker
+    visitor = _LatexVisitor()
+
+    variableMap = ComponentMap()
+    vrIdx = 0
+    for i in range(0, len(variableList)):
+        vr = variableList[i]
```
emma58, line 917: It doesn't look like you use `i`? So I think you just need `for vr in variableList:`

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+        else:
+            raise DeveloperError(
+                'Variable is not a variable.  Should not happen.  Contact developers'
+            )
+    visitor.variableMap = variableMap
+
+    parameterMap = ComponentMap()
+    pmIdx = 0
+    for i in range(0, len(parameterList)):
+        vr = parameterList[i]
```
emma58, line 935: Same here that I don't think you use `i`.

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+
+    # Iterate over the objectives and print
+    for obj in objectives:
+        try:
+            obj_template, obj_indices = templatize_fcn(obj)
+        except:
+            if throw_templatization_error:
+                raise RuntimeError(
+                    "An objective has been constructed that cannot be templatized"
+                )
```
emma58, line 979: It would be helpful to include the name of the offending objective in this error as well.

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+            else:
+                obj_template = obj
+
+        if obj.sense == 1:
```
emma58, line 983: You can make this easier to read with `if obj.sense == pe.minimize`

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+
+        # first constraint needs different alignment because of the 'subject to':
+        # & minimize   & & [Objective]
+        # & subject to & & [Constraint 1]
+        # &            & & [Constraint 2]
+        # &            & & [Constraint N]
+
+        # The double '& &' renders better for some reason
+
+        for i in range(0, len(constraints)):
```
emma58, line 1013: I think you can do `for i, con in enumerate(constraints):`

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+            # grab the constraint and templatize
+            con = constraints[i]
+            try:
+                con_template, indices = templatize_fcn(con)
+                con_template_list = [con_template]
+            except:
+                if throw_templatization_error:
+                    raise RuntimeError(
+                        "A constraint has been constructed that cannot be templatized"
+                    )
```
emma58, line 1036: Again, would you include the name of the Constraint here?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+                        }
+                    )
+                else:
+                    varBoundData += varBoundData_indexedVar
+            else:
+                raise DeveloperError(
+                    'Variable is not a variable.  Should not happen.  Contact developers'
+                )
+
+        # print(varBoundData)
```
emma58, line 1130: Looks like forgotten debugging?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+
+        # print the accumulated data to the string
+        bstr = ''
+        appendBoundString = False
+        useThreeAlgn = False
+        for i in range(0, len(varBoundData)):
+            vbd = varBoundData[i]
```
emma58, line 1137: You can use `enumerate` here too

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+
+            if split_continuous_sets:
+                for ky, vl in setInfo.items():
+                    st = vl['setObject']
+                    stData = st.data()
+                    stCont = True
+                    for ii in range(0, len(stData)):
+                        if ii + stData[0] != stData[ii]:
+                            stCont = False
+                            break
```
emma58, line 1255: @jsiirola, isn't there a fancy (less expensive) way to do this check?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+                            % (vl['indices'][i], ky),
+                            alphabetStringGenerator(indexCounter, True),
+                        )
+                        indexCounter += 1
+
+        latexLines[jj] = ln
+
+    pstr = '\n'.join(latexLines)
+
+    vrIdx = 0
```
emma58, line 1339: Doesn't look like you use this?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+    new_variableMap = ComponentMap()
+    for i in range(0, len(variableList)):
+        vr = variableList[i]
```
emma58, line 1342: Could use `enumerate`

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+                sdString = str(sd)
+                if sdString[0] == '(':
+                    sdString = sdString[1:]
+                if sdString[-1] == ')':
+                    sdString = sdString[0:-1]
+                new_variableMap[vr[sd]] = vr[sd].name
+        else:
+            raise DeveloperError(
+                'Variable is not a variable.  Should not happen.  Contact developers'
+            )
```
emma58, line 1359: Is this code reachable? You would have already hit this the first time you enumerated through variableList, right?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+
+    pmIdx = 0
```
emma58, line 1361: I don't think you use this either?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+    new_parameterMap = ComponentMap()
+    for i in range(0, len(parameterList)):
```
emma58, line 1363: Could use `enumerate`

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+                sdString = str(sd)
+                if sdString[0] == '(':
+                    sdString = sdString[1:]
+                if sdString[-1] == ')':
+                    sdString = sdString[0:-1]
+                new_parameterMap[pm[sd]] = str(pm[sd])  # .name
+        else:
+            raise DeveloperError(
+                'Parameter is not a parameter.  Should not happen.  Contact developers'
+            )
```
emma58, line 1381: Same question here--not sure this is reachable

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+
+    for ky, vl in new_variableMap.items():
+        if ky not in ComponentSet(latex_component_map.keys()):
```
emma58, line 1384: I don't think you need to make a ComponentSet here--can just do `if ky not in latex_component_map.keys()`

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+            latex_component_map[ky] = vl
+    for ky, vl in new_parameterMap.items():
+        if ky not in ComponentSet(latex_component_map.keys()):
```
emma58, line 1387: I don't think you need to make a ComponentSet here--can just do `if ky not in latex_component_map.keys()`

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+            latex_component_map[ky] = vl
+
+    rep_dict = {}
+    for ky in ComponentSet(list(reversed(list(latex_component_map.keys())))):
```
emma58, line 1391: `reversed` already returns an iterator, so you don't need the outer `list`. Also, why does order matter here?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+                'Formulation is infeasible due to bounds on variable %s' % (vr.name)
+            )
+        elif upperBoundValue == 0:
+            upperBound = ' = 0 '
+        elif upperBoundValue < 1:
+            upperBound = ' \\leq ' + str(upperBoundValue)
+        else:
+            upperBound = ' \\leq 1 '
+        # else:
+        #     upperBound = ' \\leq 1 '
```
emma58, line 615: Do you need these comments?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+                'Formulation is infeasible due to bounds on variable %s' % (vr.name)
+            )
+        elif lowerBoundValue == 1:
+            lowerBound = ' = 1 '
+        elif lowerBoundValue > 0:
+            lowerBound = str(lowerBoundValue) + ' \\leq '
+        else:
+            lowerBound = ' 0 \\leq '
+        # else:
+        #     lowerBound = ' 0 \\leq '
```
emma58, line 601: Do you need these comments?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+        #     upperBound = ' < 0 '
+
+    elif domainName in ['NonNegativeReals', 'NonNegativeIntegers']:
+        # if lowerBoundValue is not None:
+        if lowerBoundValue > 0:
+            lowerBound = str(lowerBoundValue) + ' \\leq '
+        else:
+            lowerBound = ' 0 \\leq '
+        # else:
+        #     lowerBound = ' 0 \\leq '
```
emma58, line 570: Do you still need these comments?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+        else:
+            lowerBound = ''
+
+        # if upperBoundValue is not None:
+        if upperBoundValue >= 0:
+            upperBound = ' < 0 '
+        else:
+            upperBound = ' \\leq ' + str(upperBoundValue)
+        # else:
+        #     upperBound = ' < 0 '
```
emma58, line 561: Do you still need these comments?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+        else:
+            lowerBound = ''
+
+        # if upperBoundValue is not None:
+        if upperBoundValue >= 0:
+            upperBound = ' \\leq 0 '
+        else:
+            upperBound = ' \\leq ' + str(upperBoundValue)
+        # else:
+        #     upperBound = ' \\leq 0 '
```
emma58, line 542: Looks like you don't need these comments?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+        else:
+            lowerBound = ' 0 < '
+        # else:
+        #     lowerBound = ' 0 < '
+
+        if upperBoundValue is not None:
+            if upperBoundValue <= 0:
+                raise ValueError(
+                    'Formulation is infeasible due to bounds on variable %s' % (vr.name)
+                )
```
emma58, line 517: There actually is an `InfeasibleConstraintException` in `pyomo.common.errors` that might be the right thing here (and in place of the ValueErrors below also).

emma58 (commented):
@codykarcher, a lot more comments, but most of them are minor. My only design-related thought is that I think we could still reduce the number of options since the global formatting for latex documents (like margins and fontsize) is very easy to change after the fact. That will reduce some complexity in your option parsing especially.

doc/OnlineDocs/model_debugging/latex_printing.rst
```diff
@@ -0,0 +1,152 @@
...
+
+    Prints a pyomo element (Block, Model, Objective, Constraint, or Expression) to a LaTeX compatible string
+
+   :param pyomo_component: The Pyomo component to be printed
+   :type pyomo_component: _BlockData or Model or Objective or Constraint or Expression
+   :param latex_component_map: A map keyed by Pyomo component, values become the latex representation in the printer
+   :type latex_component_map: pyomo.common.collections.component_map.ComponentMap
+   :param write_object: The object to print the latex string to.  Can be an open file object, string I/O object, or a string for a filename to write to
+   :type write_object: io.TextIOWrapper or io.StringIO or str
+   :param use_equation_environment: If False, the equation/aligned construction is used to create a single LaTeX equation.  If True, then the align environment is used in LaTeX and each constraint and objective will be given an individual equation number
```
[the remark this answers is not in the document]
codykarcher: I prefer to have the flag flip true for non-default behavior, but will attempt to make the note more clear

doc/OnlineDocs/model_debugging/latex_printing.rst
```diff
@@ -0,0 +1,152 @@
...
+   :type use_equation_environment: bool
+   :param split_continuous_sets: If False, all sums will be done over 'index in set' or similar.  If True, sums will be done over 'i=1' to 'N' or similar if the set is a continuous set
```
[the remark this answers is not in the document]
codykarcher: Yea, I'll mull this and do something better

doc/OnlineDocs/model_debugging/latex_printing.rst
```diff
@@ -0,0 +1,152 @@
...
+   :type split_continuous_sets: bool
+   :param use_short_descriptors: If False, will print full 'minimize' and 'subject to' etc.  If true, uses 'min' and 's.t.' instead
```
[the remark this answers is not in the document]
codykarcher: Fine Emma, you win :P  I'll remove the long form and the flag

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+
+def indexCorrector(ixs, base):
+    for i in range(0, len(ixs)):
+        ix = ixs[i]
+        if i + 1 < len(ixs):
+            if ixs[i + 1] == 0:
+                ixs[i] -= 1
+                ixs[i + 1] = base
+                if ixs[i] == 0:
+                    ixs = indexCorrector(ixs, base)
```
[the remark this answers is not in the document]
codykarcher: Recursion kind of required here, this function required a great deal of work.  Recursion only goes once every like 1000 variables or something, so the model would have to be massive and have no templatized constraints to really worry me.

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+
+
+def alphabetStringGenerator(num, indexMode=False):
+    if indexMode:
+        alphabet = ['.', 'i', 'j', 'k', 'm', 'n', 'p', 'q', 'r']
+
+    else:
+        alphabet = [
+            '.',
+            'a',
```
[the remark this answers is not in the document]
codykarcher: Sure, it provides context but serves no purpose

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+
+def handle_unary_node(visitor, node, arg1):
+    fcn_handle = node.getname()
+    if fcn_handle == 'log10':
+        fcn_handle = 'log_{10}'
+
+    if fcn_handle == 'sqrt':
+        return '\\sqrt { ' + arg1 + ' }'
+    else:
+        return '\\' + fcn_handle + ' \\left( ' + arg1 + ' \\right) '
```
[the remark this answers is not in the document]
codykarcher: I singled out these two because the latex render is different from just the standard default on 237.  Personally, I would bin this in future work.

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+def handle_external_function_node(visitor, node, *args):
+    pstr = ''
+    pstr += 'f('
+    for i in range(0, len(args) - 1):
+        pstr += args[i]
+        if i <= len(args) - 3:
+            pstr += ','
+        else:
+            pstr += ')'
+    return pstr
```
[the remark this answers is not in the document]
codykarcher: This is a bit of a pain, but I can do it

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+
+
+def handle_functionID_node(visitor, node, *args):
+    # seems to just be a placeholder empty wrapper object
+    return ''
```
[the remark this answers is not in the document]
codykarcher: Pyomo just has some strange nodes sometimes depending on how modelers create their models.  This was present in one of the models I tested on and was necessary to not throw a bug

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+    else:
+        visitor.setMap[node._set] = 'SET%d' % (len(visitor.setMap.keys()) + 1)
+
+    return '__I_PLACEHOLDER_8675309_GROUP_%s_%s__' % (
+        node._group,
+        visitor.setMap[node._set],
+    )
+
+
+def handle_numericGIE_node(visitor, node, *args):
```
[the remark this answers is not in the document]
codykarcher: Sure

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+
+    pstr = ''
+    pstr += joinedName + '_{'
+    for i in range(1, len(args)):
+        pstr += args[i]
+        if i <= len(args) - 2:
+            pstr += ','
+        else:
+            pstr += '}'
+    return pstr
```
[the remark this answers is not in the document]
codykarcher: Good catch, thanks

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+        'PositiveReals': '\\mathds{R}_{> 0}',
+        'NonPositiveReals': '\\mathds{R}_{\\leq 0}',
+        'NegativeReals': '\\mathds{R}_{< 0}',
+        'NonNegativeReals': '\\mathds{R}_{\\geq 0}',
+        'Integers': '\\mathds{Z}',
+        'PositiveIntegers': '\\mathds{Z}_{> 0}',
+        'NonPositiveIntegers': '\\mathds{Z}_{\\leq 0}',
+        'NegativeIntegers': '\\mathds{Z}_{< 0}',
+        'NonNegativeIntegers': '\\mathds{Z}_{\\geq 0}',
+        'Boolean': '\\left\\{ 0 , 1 \\right \\}',
```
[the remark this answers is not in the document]
codykarcher: Sure

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+        'Binary': '\\left\\{ 0 , 1 \\right \\}',
+        # 'Any': None,
+        # 'AnyWithNone': None,
+        'EmptySet': '\\varnothing',
+        'UnitInterval': '\\mathds{R}',
```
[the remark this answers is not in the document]
codykarcher: I've interpreted this as all real with constraints that bound to [0,1] and printed it as such.  Willing to be convinced I'm wrong

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+        except:
+            raise DeveloperError(
+                'Latex printer encountered an error when processing type %s, contact the developers'
+                % (node.__class__)
+            )
+
+
+def analyze_variable(vr):
+    domainMap = {
+        'Reals': '\\mathds{R}',
```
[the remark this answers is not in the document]
codykarcher: Will try and evaluate

doc/OnlineDocs/model_debugging/latex_printing.rst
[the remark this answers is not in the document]
codykarcher: But I'm putting a back door in the component dict so I can still make it do what I want

pyomo/util/latex_printer.py
codykarcher: I think it's necessary to indicate to the user what are actual pyomo constraints and what are bounds.  Plus the whole printer is now somewhat structured around this decision, so I'm going to leave it in

pyomo/util/latex_printer.py
codykarcher: Will check all of these  and the enumerates.  There is an issue here with python defaulting to a copy of array elements that I have to be careful of

pyomo/util/latex_printer.py
codykarcher: But pyo.minimize is 1...  I don't know, sure I guess

pyomo/util/latex_printer.py
codykarcher: It matters because x_1 also replaces x_10 so I have to go backwards

doc/OnlineDocs/model_debugging/latex_printing.rst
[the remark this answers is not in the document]
codykarcher: I also had the explanation backwards

doc/OnlineDocs/model_debugging/latex_printing.rst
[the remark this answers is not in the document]
codykarcher: went with `explicit_set_summation`  Maybe better, maybe not

pyomo/util/latex_printer.py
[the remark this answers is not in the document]
codykarcher: TLDR: be very careful how you use blocks if you want the latex printer to still work like you think it should

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+
+    pstr = ''
+    pstr += joinedName + '['
+    for i in range(1, len(args)):
+        pstr += args[i]
+        if i <= len(args) - 2:
+            pstr += ','
+        else:
+            pstr += ']'
+    return pstr
```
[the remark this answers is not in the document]
codykarcher: I'd have to weave together type checking stuff, which feels like more trouble than it is worth.  I'm going to leave this, though the point is valid

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+        'Integers': '\\mathds{Z}',
+        'PositiveIntegers': '\\mathds{Z}_{> 0}',
+        'NonPositiveIntegers': '\\mathds{Z}_{\\leq 0}',
+        'NegativeIntegers': '\\mathds{Z}_{< 0}',
+        'NonNegativeIntegers': '\\mathds{Z}_{\\geq 0}',
+        'Boolean': '\\left\\{ 0 , 1 \\right \\}',
+        'Binary': '\\left\\{ 0 , 1 \\right \\}',
+        # 'Any': None,
+        # 'AnyWithNone': None,
+        'EmptySet': '\\varnothing',
```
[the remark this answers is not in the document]
codykarcher: I'm staying with \varnothing on the basis that I don't feel like changing my unit tests

pyomo/util/latex_printer.py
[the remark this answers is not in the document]
codykarcher: Again choosing to stay with mathds to avoid changing my tests

pyomo/util/latex_printer.py
codykarcher: I believe this returns a list, and I need a component set to do equality checking

pyomo/util/latex_printer.py
codykarcher: Done

pyomo/util/latex_printer.py
codykarcher: you are right, but I prefer to leave it in.  Only two things are infinite, the universe and the capacity for users to do things I don't expect.  And I'm not sure about the first

pyomo/util/latex_printer.py
codykarcher: Im leaving these because they work at present and I'm unsure about the suggested change.  But I believe I needed the component set to do equality checking at some point in the function

pyomo/util/latex_printer.py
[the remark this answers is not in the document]
codykarcher: After evaluation, I agree with the sentiment, but do not believe this meets the standard of effort vs. reward.  It's a fairly involved change for something of questionable value.  I'll add it to my personal issue list, but will not implement it for the November pyomo release

doc/OnlineDocs/model_debugging/latex_printing.rst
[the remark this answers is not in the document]
emma58: Oh, definitely better! I like it.

pyomo/util/latex_printer.py
emma58: You still don't need the call to `list` then do you?

pyomo/util/latex_printer.py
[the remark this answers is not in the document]
emma58: Sorry to push back, but my concern here is this is borderline on the outputting something inaccurate front. If a bunch of functions are named `f`, I would expect them to be the same function, and they aren't necessarily. I think throwing a `MouseTrap` for all external functions would be fine at this point, if you want to keep it simple. But naming them all `f` is what concerns me a bit.

pyomo/util/latex_printer.py
codykarcher: .keys() returns an iterator, but I need a list for the initialization of ComponentSet

pyomo/util/latex_printer.py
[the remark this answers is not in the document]
codykarcher: So an easy thing is I can basically have a counter and do f_1, f_2, f_3...  That I can implement easily.  But adding to the component map of latex names is not trivial (I have to detect all the black boxes, do substitutions, map back to the desired names).  If we felt this was a significant gain (or if I had infinite bandwidth) then this would be a no brainer, but I'm not sure it is worth the time. I will implement the counter strategy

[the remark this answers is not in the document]
emma58: The counter strategy sounds good to me!

pyomo/util/latex_printer.py
emma58: You can initialize a `ComponentSet` with an iterator.

pyomo/util/tests/test_latex_printer.py
```diff
@@ -0,0 +1,791 @@
...
+    def blackbox(a, b):
+        return sin(a - b)
+
+    m.bb = ExternalFunction(blackbox)
+    m.constraint_6 = pyo.Constraint(expr=m.x + m.bb(m.x, m.y) == 2)
+
+    m.I = pyo.Set(initialize=[1, 2, 3, 4, 5])
+    m.J = pyo.Set(initialize=[1, 2, 3])
+    m.K = pyo.Set(initialize=[1, 3, 5])
```
blnicho, line 69: Have you tested the printer on Sets containing non-numeric entries like strings?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1336 @@
...
+            upperBound = ' = 0 '
+        elif upperBoundValue < 1:
+            upperBound = ' \\leq ' + str(upperBoundValue)
+        else:
+            upperBound = ' \\leq 1 '
+
+    else:
+        raise DeveloperError(
+            'Invalid domain somehow encountered, contact the developers'
+        )
```
jsiirola, line 562: This is actually easy to hit.  Variable domains can be any valid Pyomo Set, so the following is perfectly legitimate:
```python
m = ConcreteModel()
m. v_domain = Set(initialize=[3,4,5,6])
m.x = Var(domain=m.v_domain)
```
The thing to use to infer domains are the Var `is_continuous()`, `is_integer()`, and `is_is_binary()` methods:
```python
>>> m.x.is_integer()
True
>>> m.x.is_binary()
False
>>> m.x.is_continuous()
False
>>> m.x.bounds
(3, 6)
```
For the LP / NL writers, we only require that one of those three return True.  So, for example, you can write a model with a domain `[1,2,3,7,8,9]`, but the solver interfaces will complain (because it is not a bounded contiguous integer domain)

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1336 @@
...
+
+    return varBoundData
+
+
+def multiple_replace(pstr, rep_dict):
+    pattern = re.compile("|".join(rep_dict.keys()), flags=re.DOTALL)
+    return pattern.sub(lambda x: rep_dict[x.group(0)], pstr)
+
+
+def latex_printer(
```
jsiirola, line 580: This is a HUGE function, and we should look at breaking it up into helper functions

blnicho (approved):
I have a few minor comments and suggestions but otherwise I think this is ready to merge.

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1336 @@
...
+
+def handle_param_node(visitor, node):
+    return visitor.parameterMap[node]
+
+
+def handle_str_node(visitor, node):
+    return node.replace('_', '\\_')
+
+
+def handle_npv_structuralGetItemExpression_node(visitor, node, *args):
```
mrmundt, line 343: Does this need `visitor` and `node`? I don't see them being used.

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1336 @@
...
+                    parameterList.append(p_write)
+            else:
+                if p not in ComponentSet(parameterList):
+                    parameterList.append(p)
+
+        # Will grab the sets as the expression is walked
+        setList = []
+
+    else:
+        variableList = [
```
jsiirola, line 739: We should reconsider this mode.  The writers have been moving *away* from pre-gathering variables / parameters / sets and instead just recording the variables as they are encountered when walking expressions.  In part, this is because we are relaxing *where* a Var has to exist (in particular, it doesn't have to exist within the block where is it used).

pyomo/util/latex_printer.py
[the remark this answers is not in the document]
mrmundt: Any consensus on this?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1484 @@
...
+        'NonPositiveIntegers': '\\mathds{Z}_{\\leq 0}',
+        'NegativeIntegers': '\\mathds{Z}_{< 0}',
+        'NonNegativeIntegers': '\\mathds{Z}_{\\geq 0}',
+        'Boolean': '\\left\\{ 0 , 1 \\right \\}',
+        'Binary': '\\left\\{ 0 , 1 \\right \\}',
+        # 'Any': None,
+        # 'AnyWithNone': None,
+        'EmptySet': '\\varnothing',
+        'UnitInterval': '\\mathds{R}',
+        'PercentFraction': '\\mathds{R}',
```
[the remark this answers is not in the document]
mrmundt: Any consensus on this?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1336 @@
...
+        'NonNegativeIntegers': '\\mathds{Z}_{\\geq 0}',
+        'Boolean': '\\left\\{ \\text{True} , \\text{False} \\right \\}',
+        'Binary': '\\left\\{ 0 , 1 \\right \\}',
+        # 'Any': None,
+        # 'AnyWithNone': None,
+        'EmptySet': '\\varnothing',
+        'UnitInterval': '\\mathds{R}',
+        'PercentFraction': '\\mathds{R}',
+        # 'RealInterval' :        None    ,
+        # 'IntegerInterval' :     None    ,
```
mrmundt, line 445: Remove?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1336 @@
...
+    pyomo_component,
+    latex_component_map=None,
+    write_object=None,
+    use_equation_environment=False,
+    explicit_set_summation=False,
+    throw_templatization_error=False,
+):
+    """This function produces a string that can be rendered as LaTeX
+
+    As described, this function produces a string that can be rendered as LaTeX
```
mrmundt, line 590: Don't need this line.

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1336 @@
...
+
+def multiple_replace(pstr, rep_dict):
+    pattern = re.compile("|".join(rep_dict.keys()), flags=re.DOTALL)
+    return pattern.sub(lambda x: rep_dict[x.group(0)], pstr)
+
+
+def latex_printer(
+    pyomo_component,
+    latex_component_map=None,
+    write_object=None,
```
jsiirola, line 583: We generally refer to this as `ostream`.  It would probably be best to maintain consistency.

pyomo/util/latex_printer.py
mrmundt: Poke about this again. I have strong feelings about `isSingle` and how it doesn't make sense.

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1336 @@
...
+        if not isSingle:
+            pstr += ' ' * tbSpc + '& %s \n' % (descriptorDict['subject to'])
+
+        # first constraint needs different alignment because of the 'subject to':
+        # & minimize   & & [Objective]
+        # & subject to & & [Constraint 1]
+        # &            & & [Constraint 2]
+        # &            & & [Constraint N]
+
+        # The double '& &' renders better for some reason
```
mrmundt, line 883: Remove?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1336 @@
...
+            for con_template in con_template_list:
+                # Walk the constraint
+                conLine = (
+                    ' ' * tbSpc
+                    + algn
+                    + ' %s %s'
+                    % (visitor.walk_expression(con_template), trailingAligner)
+                )
+
+                # setMap = visitor.setMap
```
mrmundt, line 922: Remove?

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1336 @@
...
+
+    throw_templatization_error: bool
+        Option to throw an error on templatization failure rather than
+        printing each constraint individually, useful for very large models
+
+
+    Returns
+    -------
+    str
+        A LaTeX string of the pyomo_component
```
jsiirola, line 623: I am not sure I am in favor of this method always returning a big string. Maybe only return the big string if the user didn't provide an `ostream`?

This is a special case of a concern I have with the bulk of the method below where we are building up a single HUGE string.  Because Python strings are immutable, every time you `+=`, you are allocating a new string that is a copy of the old one with the extra characters.  This leads to "O(n^2)" behavior that can cause this to be very slow for large models.  I would rather this be structured to build up (and export) smaller chunks as it goes.

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1336 @@
...
+                    replacement = (
+                        r'sum_{ __I_PLACEHOLDER_8675309_GROUP_\1_%s__ \\in __S_PLACEHOLDER_8675309_GROUP_\1_%s__  }'
+                        % (ky, ky)
+                    )
+                    ln = re.sub(setInfo[ky]['sumSetRegEx'], replacement, ln)
+
+                replacement = repr(defaultSetLatexNames[setInfo[ky]['setObject']])[1:-1]
+                ln = re.sub(setInfo[ky]['setRegEx'], replacement, ln)
+
+            # groupNumbers = re.findall(r'__I_PLACEHOLDER_8675309_GROUP_([0-9*])_SET[0-9]*__',ln)
```
mrmundt, line 1152: Remove?

pyomo/util/tests/test_latex_printer.py
```diff
@@ -0,0 +1,791 @@
...
+    NonNegativeReals,
+    Integers,
+    PositiveIntegers,
+    NonPositiveIntegers,
+    NegativeIntegers,
+    NonNegativeIntegers,
+    Boolean,
+    Binary,
+    Any,
+    # AnyWithNone,
```
mrmundt, line 34: Remove?

pyomo/util/tests/test_latex_printer.py
```diff
@@ -0,0 +1,791 @@
...
+    EmptySet,
+    UnitInterval,
+    PercentFraction,
+    # RealInterval,
+    # IntegerInterval,
```
mrmundt, line 39: Remove?

pyomo/util/tests/test_latex_printer_vartypes.py
```diff
@@ -0,0 +1,3283 @@
...
+    NonNegativeReals,
+    Integers,
+    PositiveIntegers,
+    NonPositiveIntegers,
+    NegativeIntegers,
+    NonNegativeIntegers,
+    Boolean,
+    Binary,
+    Any,
+    # AnyWithNone,
```
mrmundt, line 33: Remove?

pyomo/util/tests/test_latex_printer_vartypes.py
```diff
@@ -0,0 +1,3283 @@
...
+    EmptySet,
+    UnitInterval,
+    PercentFraction,
+    # RealInterval,
+    # IntegerInterval,
```
mrmundt, line 38: Remove?

pyomo/util/latex_printer.py
mrmundt, line 580: This could be broken up into smaller functions and not look so monstrously large.

mrmundt (changes requested):

pyomo/util/tests/test_latex_printer.py
codykarcher: No, but in theory it shouldn't matter what's in the set unless you use the `explicit_set_summation` flag.  There is a known issue with how things like pro.Set()*pyo.Set() end up rendering, but I'm living with that for now

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1336 @@
...
+
+from pyomo.common.errors import InfeasibleConstraintException
+
+from pyomo.common.dependencies import numpy, numpy_available
+
+if numpy_available:
+    import numpy as np
+
+
+def decoder(num, base):
```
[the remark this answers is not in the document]
codykarcher: I don't understand this code, but it worked so going with it

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1336 @@
...
+        else:
+            childPrecedence.append(-1)
+
+    if hasattr(node, 'PRECEDENCE'):
+        precedence = node.PRECEDENCE
+    else:
+        # Should never hit this
+        raise DeveloperError(
+            'This error should never be thrown, node does not have a precedence.  Report to developers'
+        )
```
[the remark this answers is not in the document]
codykarcher: Used developer error for explicitness.

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1336 @@
...
+
+    if childPrecedence[0] > precedence:
+        arg1 = ' \\left( ' + arg1 + ' \\right) '
```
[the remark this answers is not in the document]
codykarcher: possibly, but this was the bane of my existence for like 3 days.  If there is no performance improvement, it's staying as is for now

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1336 @@
...
+
+
+def handle_indexTemplate_node(visitor, node, *args):
+    if node._set in ComponentSet(visitor.setMap.keys()):
+        # already detected set, do nothing
+        pass
+    else:
+        visitor.setMap[node._set] = 'SET%d' % (len(visitor.setMap.keys()) + 1)
+
+    return '__I_PLACEHOLDER_8675309_GROUP_%s_%s__' % (
```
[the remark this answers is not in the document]
codykarcher: Agreed this is less than ideal and was always intended to be a placeholder, but at this point represents a major change.  Will re-evaluate after some users test things out for a bit

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1336 @@
...
+
+    return pstr
+
+
+def handle_param_node(visitor, node):
+    return visitor.parameterMap[node]
+
+
+def handle_str_node(visitor, node):
+    return node.replace('_', '\\_')
```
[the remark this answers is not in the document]
codykarcher: Sure

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1336 @@
...
+        'UnitInterval': '\\mathds{R}',
+        'PercentFraction': '\\mathds{R}',
+        # 'RealInterval' :        None    ,
+        # 'IntegerInterval' :     None    ,
+    }
+
+    domainName = vr.domain.name
+    varBounds = vr.bounds
+    lowerBoundValue = varBounds[0]
+    upperBoundValue = varBounds[1]
```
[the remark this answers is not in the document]
codykarcher: I handle this later on, my code is just more complicated than it needs to be

pyomo/util/latex_printer.py
codykarcher: Changed this to a NotImplementedError for the time being

pyomo/util/latex_printer.py
```diff
@@ -0,0 +1,1336 @@
...
+        arg1 = ' \\left( ' + arg1 + ' \\right) '
+
+    if arg2 is not None:
+        if childPrecedence[1] > precedence:
+            arg2 = ' \\left( ' + arg2 + ' \\right) '
+
+    return arg1, arg2
+
+
+def handle_negation_node(visitor, node, arg1):
```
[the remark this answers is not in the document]
codykarcher: For this and all that follow, the functions need to have the same first two inputs for the scheme to work because some of the functions do use visitor

mrmundt (approved):

PR #2988, opened 2023-09-08
Add the support of GreyBox in MindtPy

### LINKED ISSUE #2767: Use cyipopt with Mindtpy
## Summary

When using the GreyBox model in Pyomo, we get an error message saying cyipopt cannot be used with mindtpy. 

### Description

Error message: 
  File "/home/parallels/anaconda3/envs/pyomo/lib/python3.7/site-packages/pyomo/common/config.py", line 1262, in _cast
    return self._domain(value)
  File "/home/parallels/anaconda3/envs/pyomo/lib/python3.7/site-packages/pyomo/common/config.py", line 249, in __call__
    raise ValueError("value %s not in domain %s" % (value, self._domain))
ValueError: value cyipopt not in domain ['ipopt', 'gams', 'baron']

```
  File "rotary-bed-cost-strategy.py", line 439, in <module>
    results = solver.solve(mod, mip_solver='gurobi_persistent', nlp_solver='cyipopt', tee=True, integer_tolerance=0.5)
  File "/home/parallels/anaconda3/envs/pyomo/lib/python3.7/site-packages/pyomo/contrib/mindtpy/MindtPy.py", line 79, in solve
    config.set_value(kwds)
  File "/home/parallels/anaconda3/envs/pyomo/lib/python3.7/site-packages/pyomo/common/config.py", line 2193, in set_value
    self[key] = value[_decl_map[key]]
  File "/home/parallels/anaconda3/envs/pyomo/lib/python3.7/site-packages/pyomo/common/config.py", line 2010, in __setitem__
    self._data[_key].set_value(val)
  File "/home/parallels/anaconda3/envs/pyomo/lib/python3.7/site-packages/pyomo/common/config.py", line 1652, in set_value
    self._data = self._cast(value)
  File "/home/parallels/anaconda3/envs/pyomo/lib/python3.7/site-packages/pyomo/common/config.py", line 1273, in _cast
    (self.name(True), value, _dom, err))
ValueError: invalid value for configuration 'nlp_solver':
	Failed casting cyipopt
	to <class 'pyomo.common.config.In'>
	Error: value cyipopt not in domain ['ipopt', 'gams', 'baron']
```

ZedongPeng:
## Summary/Motivation:
This Pull Request adds the support of the Greybox in MindtPy.

- Generates OA cuts using the information from the GreyBox model.
- Fixes a bug in [pyomo_grey_box_nlp.py](https://github.com/Pyomo/pyomo/blob/main/pyomo/contrib/pynumero/interfaces/pyomo_grey_box_nlp.py)
### Legal Acknowledgement


pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -1311,6 +1329,12 @@ def solve_feasibility_subproblem(self):
         update_solver_timelimit(
             self.feasibility_nlp_opt, config.nlp_solver, self.timing, config
         )
+        TransformationFactory('contrib.deactivate_trivial_constraints').apply_to(
+            feas_subproblem,
+            tmp=True,
+            ignore_infeasible=False,
+            tolerance=config.constraint_tolerance,
+        )
```
bernalde, line 1337: The use of this transformation is new. Is it motivated by our new tests? Is the transformation well-supported? @emma58

pyomo/contrib/mindtpy/cut_generation.py
```diff
@@ -181,6 +181,53 @@ def add_oa_cuts(
...
+                        for var_index, var in enumerate(
+                            target_model_grey_box.inputs.values()
+                        )
+                    )
+                )
+                - (output - value(output))
+                - (slack_var if config.add_slack else 0)
+                <= 0
+            )
+    # TODO: gurobi_persistent currently does not support greybox model.
```
bernalde, line 217: Should this become an issue?

pyomo/contrib/mindtpy/cut_generation.py
```diff
@@ -181,6 +181,53 @@ def add_oa_cuts(
...
+        and mip_iter > 0
+        and cb_opt is not None
+    ):
+        cb_opt.cbLazy(
+            target_model.MindtPy_utils.cuts.oa_cuts[
+                len(target_model.MindtPy_utils.cuts.oa_cuts)
+            ]
+        )
+
+
```
bernalde, line 230: These cuts assume that the black box is: 1. in the objective, 2. is convex. For the 1st one, can we generalize it? For the second one, can we use an option to decide whether or not we want to generate the augmented penalty cuts? Our examples/tests satisfy both assumptions, but we can get some flexibility for not so much effort.

pyomo/contrib/pynumero/interfaces/pyomo_grey_box_nlp.py
```diff
@@ -426,7 +426,11 @@ def load_state_into_pyomo(self, bound_multipliers=None):
         # not we are minimizing or maximizing - this is done in the ASL interface
         # for ipopt, but does not appear to be done in cyipopt.
         obj_sign = 1.0
-        objs = list(m.component_data_objects(ctype=pyo.Objective, descend_into=True))
+        objs = list(
+            m.component_data_objects(
+                ctype=pyo.Objective, active=True, descend_into=True
```
bernalde, line 431: Leave a comment on why are we adding `active=True` here and at least let's run this through @michaelbynum to see if it is a good change

bernalde (changes requested):
Good work, Zedong! I request some changes and also some revisions from our friends that motivated this development! @adowling2 !

bernalde:
The tests are failing because:
`FAILED pyomo/contrib/mindtpy/tests/test_mindtpy.py::TestMindtPy::test_OA_APPSI_ipopt - RuntimeError: A feasible solution was not found, so no solution can be loaded.Please set opt.config.load_solution=False and check results.termination_condition and results.best_feasible_objective before loading a solution.`
This has to do with the new option passed. Let's solve this first

pyomo/contrib/mindtpy/cut_generation.py
ZedongPeng: `Gurobi` will fail even when we deactivate the `greybox` block. Therefore, we cannot use `Gurobi` as the mip solver.

pyomo/contrib/mindtpy/algorithm_base_class.py
ZedongPeng: After MindtPy rewrite, we repeatedly use the fixed-nlp subproblem. Therefore, we will call `'contrib.deactivate_trivial_constraints'` transformation and revert it after solving. I missed the feasibility subproblem and this might be a bug.

pyomo/contrib/mindtpy/cut_generation.py
ZedongPeng: In this implementation, we don't assume the grey box is in the objective function. It will also work if it is in the constraints. 
The constraints inside the grey box are nonlinear equality constraints. If the constraint is convex, we should use **Equality Relaxation**. If the constraint is nonconvex, we should use both **Equality Relaxation** and **Augmented Penalty**. We can add slack variables using `config.add_slack`, the same as algebraic constraints. 
Therefore, our current implementation works even without the two assumptions.

pyomo/contrib/mindtpy/cut_generation.py
ZedongPeng: I will submit an issue related to this.

ZedongPeng: https://github.com/Pyomo/pyomo/issues

ZedongPeng: @emma58 Could you provide me a `reference` example?

emma58:
```
from pyomo.environ import *

m = ConcreteModel()
m.b = Block()
m.b.x = Var(bounds=(2, 5))
m.y = Var(domain=NonNegativeReals)
m.c = Constraint(expr=m.b.x - m.y <= 4)
m.obj = Objective(expr=m.b.x + m.y)

m.b.deactivate()

m.somewhere_safe_to_keep_things = Block()
m.somewhere_safe_to_keep_things.x = Reference(m.b.x)

opt = SolverFactory('gurobi_persistent')
opt.set_instance(m)

opt.solve(m, tee=True)
```
This is the main idea. You'll want to figure out how to do your own accounting in terms of finding the Vars you need to reference and naming the references, etc.

bernalde: #3000 Issue opened

ZedongPeng: @emma58 . I don't think `Reference` is an elegant way to resolve this issue. I will comment the code of adding `Gurobi` lazy constraints and only support `greybox` in multi-tree implementation this time. 
Btw, since persistent solver will be totally replaced by `appsi_solver`, does appsi_solver support both `CPLEX` and `Gurobi` callbacks as `persistent solver` now? @michaelbynum

ZedongPeng:
This PR is ready for review. @emma58

pyomo/contrib/mindtpy/cut_generation.py
michaelbynum: appsi_gurobi does support callbacks. appsi_cplex will, but does not currently.

bernalde:
I don't think we're ready yet, Zedong.
Some errors from the tests:
```
FAILED pyomo/contrib/mindtpy/tests/test_mindtpy.py::TestMindtPy::test_OA_APPSI_ipopt - RuntimeError: A feasible solution was not found, so no solution can be loaded.Please set opt.config.load_solution=False and check results.termination_condition and results.best_feasible_objective before loading a solution.
FAILED pyomo/environ/tests/test_environ.py::TestPyomoEnviron::test_tpl_import_time - AssertionError: Items in the first set but not the second:
'scipy' : Unexpected module found in 5 slowest-loading TPL modules
==== 2 failed, 12120 passed, 5009 skipped, 29 xfailed in 954.63s (0:15:54) =====
```

We are responsible for the first, and scipy taking so much longer seems to be a new thing. Let's aim to fix the first error and then we can keep trying.

ZedongPeng:
Hi @jsiirola . `scipy` cannot be imported from Pyomo dependency. How can I fix this?

Import code
`from pyomo.common.dependencies.scipy.sparse import coo_matrix`

Output
```
  File "/home/runner/work/pyomo/pyomo/pyomo/contrib/pynumero/interfaces/external_grey_box.py", line 15, in <module>
    from pyomo.common.dependencies.scipy.sparse import coo_matrix
  File "/home/runner/work/pyomo/pyomo/pyomo/common/dependencies.py", line 168, in __getattr__
    return getattr(_mod, attr)
  File "/home/runner/work/pyomo/pyomo/pyomo/common/dependencies.py", line 75, in __getattr__
    raise DeferredImportError(self._moduleunavailable_message())
pyomo.common.errors.DeferredImportError: The scipy.sparse module (an optional Pyomo dependency) failed to import: ModuleNotFoundError: No module named 'scipy'
```

jsiirola:
@ZedongPeng, you want to do something like:
```python
import pyomo.common.dependencies.scipy.sparse as scipy_sparse

# then later...
return scipy_sparse.coo_matrix((data, (row, col)), shape=(1, 5))
```

ZedongPeng:
The error is not inside `MindtPy` now. How can we deal with it? @jsiirola 
```
FAILED pyomo/environ/tests/test_environ.py::TestPyomoEnviron::test_tpl_import_time - AssertionError: Items in the first set but not the second:
'scipy' : Unexpected module found in 5 slowest-loading TPL modules
==== 1 failed, 12121 passed, 5009 skipped, 29 xfailed in 1001.61s (0:16:41) ====
```

jsiirola:
Ahh - the issue is that mindtpy is importing PyNumero (in `pyomo/contrib/mindtpy/algorithm_base_class.py`), which is importing scipy.

You can see this by running `python pyomo/environ/tests/test_environ.py`.  You will see (in a lot of other output):
```
TPLS:
                            __future__:    179  (from pyomo.core.expr.visitor)
[....]
                          scipy.sparse:  12384  (from pyomo.contrib.pynumero.interfaces.external_grey_box)
                           scipy.stats: 134511  (from pyomo.contrib.pynumero.dependencies)
```

You can resolve things by deferring the import of pynumero using the `attempt_import` mechanism.  Replace the pynumero import with:
```python

egb = attempt_import('pyomo.contrib.pynumero.interfaces.external_grey_box')[0]
```
and then replace instances of `ExternalGreyBoxBlock` with `egb.ExternalGreyBoxBlock`.

bernalde (changes requested):
There seem to be some test failing still.
`E   pyomo.common.errors.DeferredImportError: The pyomo.contrib.pynumero.interfaces.external_grey_box module (an optional Pyomo dependency) failed to import: ModuleNotFoundError: No module named 'scipy'
------- generated xml file: /home/runner/work/pyomo/pyomo/TEST-pyomo.xml -------
ERROR pyomo/contrib/mindtpy/tests/test_mindtpy.py
ERROR pyomo/contrib/mindtpy/tests/test_mindtpy_ECP.py
ERROR pyomo/contrib/mindtpy/tests/test_mindtpy_feas_pump.py
ERROR pyomo/contrib/mindtpy/tests/test_mindtpy_grey_box.py
ERROR pyomo/contrib/mindtpy/tests/test_mindtpy_lp_nlp.py
!!!!!!!!!!!!!!!!!!! Interrupted: 5 errors during collection !!!!!!!!!!!!!!!!!!!!
================== 30 skipped, 5 errors in 100.53s (0:01:40) ===================`
These tests fail on the slim verrsion and pypy versions (those without numpy). We are importing numpy instead of attempting that import

ZedongPeng:
We should find a chance to fix all the `import` in `pyomo`.

emma58:
@ZedongPeng, checking the slim tests and the pypy-3.9 ones, it looks like you need to skip some tests when numpy and scipy aren't available.

ZedongPeng:
Do we have to put `import` at the front of the file? @emma58 
If so, we cannot skip the `scipy import error`. And the `scipy import error` is from `pynumero` instead of `MindtPy`.
```
error: cannot format /home/runner/work/pyomo/pyomo/pyomo/contrib/mindtpy/tests/test_mindtpy_grey_box.py: Cannot parse: 20:0: from pyomo.contrib.mindtpy.tests.MINLP_simple import SimpleMINLP as SimpleMINLP
```

emma58:
Sorry, I mislead you, It looks like your test skips were okay before. My guess is that the test failures had to do with someplace that you import something from pynumero that is in turn importing numpy. Remember that when you use the deferred imports, the package gets imported when you use it, so you will probably still need to check if numpy is available and, if not, throw an error from mindtpy, before you import what you need from pynumero. I'll mark a couple places in the review where I think this might be happening.

pyomo/contrib/mindtpy/tests/MINLP_simple_grey_box.py
```diff
@@ -0,0 +1,139 @@
+from pyomo.common.dependencies import numpy as np
+import pyomo.common.dependencies.scipy.sparse as scipy_sparse
+from pyomo.common.dependencies import attempt_import
+
+egb = attempt_import('pyomo.contrib.pynumero.interfaces.external_grey_box')[0]
+
+
+class GreyBoxModel(egb.ExternalGreyBoxModel):
+    """Greybox model to compute the example OF."""
```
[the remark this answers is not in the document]
ZedongPeng: Objective function.

pyomo/contrib/mindtpy/tests/test_mindtpy_grey_box.py
```diff
@@ -0,0 +1,78 @@
...
+#  ___________________________________________________________________________
+
+"""Tests for the MindtPy solver."""
+from pyomo.core.expr.calculus.diff_with_sympy import differentiate_available
+import pyomo.common.unittest as unittest
+from pyomo.environ import SolverFactory, value, maximize
+from pyomo.opt import TerminationCondition
+from pyomo.common.dependencies import numpy_available, scipy_available
+@unittest.skipIf(not numpy_available, 'Required numpy %s is not available')
+@unittest.skipIf(not scipy_available, 'Required scipy %s is not available')
```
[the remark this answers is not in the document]
ZedongPeng: Got it.

pyomo/contrib/mindtpy/tests/MINLP_simple_grey_box.py
```diff
@@ -0,0 +1,139 @@
...
+        y1 = self._input_values[2]
+        y2 = self._input_values[3]
+        y3 = self._input_values[4]
+        # z
+        z = x1**2 + x2**2 + y1 + 1.5 * y2 + 0.5 * y3
+
+        if self.verbose:
+            pass
+            # print("\n Consider inputs [x1,x2,y1,y2,y3] =\n",x1, x2, y1, y2, y3)
+            # print("   z = ",z,"\n")
```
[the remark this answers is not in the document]
ZedongPeng: I changed the default value of verbose to `False`.

bernalde:
Test failures point to import of scipy when it is not available. A conditional attempt import should resolve it.

pyomo/contrib/mindtpy/cut_generation.py
```diff
@@ -181,6 +181,54 @@ def add_oa_cuts(
...
+                        jacobian_matrix[index][var_index] * (var - value(var))
+                        for var_index, var in enumerate(
+                            target_model_grey_box.inputs.values()
+                        )
+                    )
+                )
+                - (output - value(output))
+                - (slack_var if config.add_slack else 0)
+                <= 0
+            )
```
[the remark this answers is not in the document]
ZedongPeng: If we don't call `enumerate` over `values`, any suggestions about how to implement it? Directly `enumerate` over `variable`?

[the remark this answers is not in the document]
emma58: I don't really know--you would know better than me. It's completely possible that this is fine. My concern is that you will end up with the wrong coefficient (`jacobian_matrix[index][var_index]`) paired with the wrong `var - value(var)` term if the `target_model_greyboox.inputs` dictionary is every constructed in a different order that doesn't correspond to the indices in `jacobian_matrix`. I don't know greybox well at all, so maybe it makes a promise that what I just said will never happen, in which case your code is totally fine. I'm just asking if you know if that's the case or not. It's always worth thinking carefully when you are relying on the order things come up when iterating over a dictionary...

pyomo/contrib/mindtpy/tests/MINLP_simple_grey_box.py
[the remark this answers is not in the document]
emma58: Ah. Could you write it out in the docstring?

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -2200,6 +2235,17 @@ def check_config(self):
...
+        # related to https://github.com/Pyomo/pyomo/issues/2363
+        if (
+            'appsi' in config.mip_solver
+            or 'appsi' in config.nlp_solver
+            or (
+                config.mip_regularization_solver is not None
+                and 'appsi' in config.mip_regularization_solver
+            )
+        ):
+            config.load_solutions = False
```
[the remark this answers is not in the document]
ZedongPeng: Yes. This option should not be up to the users. Is there a way to define an option (not open to users) instead of config (open to users) in Pyomo?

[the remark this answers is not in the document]
emma58: It's not really an option at all, just something you calculate based on some of the config options, right? You could probably store it privately on the algorithm class. Just make sure to restore state afterwards if you do keep it on the class.

ZedongPeng:
Now all the tests passed except the Jenkins one. Could you show me the output of Jenkins? @emma58 Thanks.

emma58:
@ZedongPeng, I'm not convinced those failures are actually your fault... I just merged main, and we'll see what happens. It looks like you still have a few comments above to address though?

pyomo/contrib/mindtpy/algorithm_base_class.py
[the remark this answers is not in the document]
ZedongPeng: Is it a good choice to change `load_solutions` to the attribute of the Solver Object?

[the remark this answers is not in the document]
emma58: That sounds fine to me. You should probably make it private as well. The only risk of doing this is that you should be careful to restore the state after the solve.

pyomo/contrib/mindtpy/cut_generation.py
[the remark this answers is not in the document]
ZedongPeng: Currently, I cannot come up with a good way to improve this since the `evaluate_jacobian_outputs` returns a `scipy.sparse._coo.coo_matrix`.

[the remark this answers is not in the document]
emma58: Maybe just put in a comment that you are relying on the order in which items are added to the `target_model_greybox.inputs` dictionary, so that future you (or someone else) will have a hint if this ever doesn't work.

ZedongPeng:
I think this PR is ready to merge if all the tests pass.

emma58 (approved):

emma58:
This is ready for a second review. @jsiirola, @mrmundt, @blnicho, @michaelbynum, any volunteers?

blnicho (approved):

PR #2994, opened 2023-09-15
Handle evaluation errors in CyIpopt solver

### LINKED ISSUE #2899: Handle evaluation errors in PyNumero

Robbybp:
## Fixes #2899
(The part of #2899 that was not fixed by #2901)

## Summary/Motivation:
A roadblock to using CyIpopt more often is that our interface did not handle evaluation errors. As of https://github.com/mechmotum/cyipopt/pull/215, CyIpopt handles evaluation errors by allowing callbacks to raise `cyipopt.CyIpoptEvaluationError`, which cause `false` to be returned to Ipopt. When raised in the constraint or objective evaluation, this triggers the step size-reduction routine that many are familiar with through the AMPL interface. This also opens the possibility of displaying more informative debugging information when an evaluation error occurs, although this will likely require tighter integration with ASL and may not be straightforward.

This PR will not pass tests until (a) CyIpopt publishes a release or (b) we test against CyIpopt/main. I'm opening this PR to publicize that this functionality is in the pipeline, but can close it until I expect tests to pass if that's better.

- Catch `PyNumeroEvaluationError` in `CyIpoptNLP` and raise `CyIpoptEvaluationError`

### Legal Acknowledgement


blnicho:
@Robbybp looks like cyipopt 1.3.0 was released a couple days ago so this PR can now move forward when you have time to come back to it.

Robbybp:
@blnicho Awesome, this is ready for review once tests pass.

pyomo/contrib/pynumero/interfaces/cyipopt_interface.py
```diff
@@ -328,24 +329,44 @@ def scaling_factors(self):
...
     def objective(self, x):
-        self._set_primals_if_necessary(x)
-        return self._nlp.evaluate_objective()
+        try:
+            self._set_primals_if_necessary(x)
+            return self._nlp.evaluate_objective()
+        except PyNumeroEvaluationError:
+            # TODO: halt_on_evaluation_error option. If set, we re-raise the
+            # original exception.
+            raise cyipopt.CyIpoptEvaluationError(
```
jsiirola, line 338: This requires cyipopt 1.3.0.  Should we update the `attempt_import` only allow cyipopt >= 1.3.0, or should we update the importer on line 28 to define a `cyipopt.CyIpoptEvaluationError` exception as a compatibility shim for older versions?  

I would tend to lean toward the latter.

Robbybp: Good point. I agree with defining a dummy evaluation error for compatibility.

This also raises the question of behavior consistency among different CyIpopt versions. Should the default be to re-raise the original exception, with a `halt_on_evaluation_error` flag that a user can set to `False` if they want to use the new feature?

Robbybp:
Note that this PR changes behavior in our CyIpopt interface somewhat significantly (although the new behavior is closer to the default ASL-Ipopt behavior). Should the default be to halt on these evaluation errors for "backward compatibility" (for all the users who currently rely on CyIpopt solves failing due to evaluation errors...)?

pyomo/contrib/pynumero/interfaces/cyipopt_interface.py
jsiirola: This is in contrib, so backwards compatibility requirements are not as strict. I think I would vote for consistency with the ASL IPOPT interface and have `halt_on_evaluation_error` default to False.

Robbybp:
OSX 3.7 test failure:
```
=================================== FAILURES ===================================
____________________ TestFileUtils.test_find_library_system ____________________

self = <pyomo.common.tests.test_fileutils.TestFileUtils testMethod=test_find_library_system>

    @unittest.skipIf(
        sys.version_info[:2] < (3, 8)
        and (
            platform.mac_ver()[0].startswith('10.16')
            or platform.mac_ver()[0].startswith('12.6')
        ),
        "find_library has known bugs in Big Sur/Monterey for Python<3.8",
    )
    def test_find_library_system(self):
        # Find a system library (before we muck with the PATH)
        _args = {'cwd': False, 'include_PATH': False, 'pathlist': []}
        if FileDownloader.get_sysinfo()[0] == 'windows':
            a = find_library('ntdll', **_args)
            b = find_library('ntdll.dll', **_args)
            c = find_library('foo\\bar\\ntdll.dll', **_args)
        else:
            a = find_library('c', **_args)
            b = find_library('libc.so', **_args)
            c = find_library('foo/bar/libc.so', **_args)
>       self.assertIsNotNone(a)
E       AssertionError: unexpectedly None

pyomo/common/tests/test_fileutils.py:252: AssertionError
------ generated xml file: /Users/runner/work/pyomo/pyomo/TEST-pyomo.xml -------
FAILED pyomo/common/tests/test_fileutils.py::TestFileUtils::test_find_library_system - AssertionError: unexpectedly None
==== 1 failed, 11169 passed, 5937 skipped, 11 xfailed in 770.13s (0:12:50) =====
Error: Process completed with exit code 1.
```

Robbybp:
Updates:
- Added a `halt_on_evaluation_error` option to `CyIpoptNLP` and `PyomoCyIpoptSolver`. The default behavior depends on whether the CyIpopt version is at least 1.3.0. This is also how we avoid raising `CyIpoptEvaluationError` if it doesn't exit. If `halt_on_evaluation_error`, we re-raise the original exception.
- Tests are run/skipped depending on CyIpopt version, as appropriate

The OSX/3.7 failure appears unrelated, so I believe this is ready for review (assuming no other test failures pop up).

jsiirola:
OSX errors will be "resolved" by #3008

pyomo/contrib/pynumero/interfaces/cyipopt_interface.py
```diff
@@ -328,24 +346,56 @@ def scaling_factors(self):
...
 
     def objective(self, x):
-        self._set_primals_if_necessary(x)
-        return self._nlp.evaluate_objective()
+        try:
+            self._set_primals_if_necessary(x)
+            return self._nlp.evaluate_objective()
+        except PyNumeroEvaluationError:
+            # TODO: halt_on_evaluation_error option. If set, we re-raise the
+            # original exception.
```
blnicho, line 354: Can this TODO be removed? Doesn't this PR add the `halt_on_evaluation_error` option?

blnicho (approved):
I found one minor thing but otherwise this looks great!

pyomo/contrib/pynumero/interfaces/cyipopt_interface.py
Robbybp: Yes, that is out of date. Thanks for catching

michaelbynum (approved):

PR #2996, opened 2023-09-18
Fix appsi auto-update when unfixing a variable and changing its bound

michaelbynum:
## Summary/Motivation:
This PR fixes a bug in appsi. The bug occurs when both unfixing a variable and changing its bound. Because of the ordering of some if statements, we were not hitting some code that we should. As a result, objectives/constraints that used the fixed variable were not getting updated correctly.

- fix the bug
- add a test

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/appsi/base.py
+++ b/pyomo/contrib/appsi/base.py
@@ -1406,17 +1406,17 @@ class PersistentBase(abc.ABC):
             vars_to_update = list()
             for v in vars_to_check:
                 _v, lb, ub, fixed, domain_interval, value = self._vars[id(v)]
-                if lb is not v._lb:
-                    vars_to_update.append(v)
-                elif ub is not v._ub:
-                    vars_to_update.append(v)
-                elif (fixed is not v.fixed) or (fixed and (value != v.value)):
+                if (fixed is not v.fixed) or (fixed and (value != v.value)):
                     vars_to_update.append(v)
                     if self.update_config.treat_fixed_vars_as_params:
                         for c in self._referenced_variables[id(v)][0]:
                             cons_to_remove_and_add[c] = None
                         if self._referenced_variables[id(v)][2] is not None:
                             need_to_set_objective = True
+                elif lb is not v._lb:
+                    vars_to_update.append(v)
+                elif ub is not v._ub:
+                    vars_to_update.append(v)
                 elif domain_interval != v.domain.get_interval():
                     vars_to_update.append(v)
             self.update_variables(vars_to_update)
--- a/pyomo/contrib/appsi/base.py
+++ b/pyomo/contrib/appsi/base.py
@@ -1406,7 +1406,7 @@ class PersistentBase(abc.ABC):
             vars_to_update = list()
             for v in vars_to_check:
                 _v, lb, ub, fixed, domain_interval, value = self._vars[id(v)]
-                if (fixed is not v.fixed) or (fixed and (value != v.value)):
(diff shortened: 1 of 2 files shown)


mrmundt (approved):

pyomo/contrib/appsi/base.py
```diff
@@ -1406,17 +1406,17 @@ def update(self, timer: HierarchicalTimer = None):
             vars_to_update = list()
             for v in vars_to_check:
                 _v, lb, ub, fixed, domain_interval, value = self._vars[id(v)]
-                if lb is not v._lb:
-                    vars_to_update.append(v)
-                elif ub is not v._ub:
-                    vars_to_update.append(v)
-                elif (fixed is not v.fixed) or (fixed and (value != v.value)):
+                if (fixed is not v.fixed) or (fixed and (value != v.value)):
```
jsiirola, line 1409: Linters would probably prefer this to be `!=` and not `is not` (because you are comparing `bool`).  The issue is that it is possible to get POD values that compare equal, but are not the same object (so `is not` returns True).  In this case, a false positive it fine (it just means APPSI does a little extra unnecessary work), but it might be better to make a distinction for when we care if an object changes or when the value changes.

pyomo/contrib/appsi/base.py
```diff
@@ -1406,17 +1406,17 @@ def update(self, timer: HierarchicalTimer = None):
...
                     vars_to_update.append(v)
                     if self.update_config.treat_fixed_vars_as_params:
                         for c in self._referenced_variables[id(v)][0]:
                             cons_to_remove_and_add[c] = None
                         if self._referenced_variables[id(v)][2] is not None:
                             need_to_set_objective = True
+                elif lb is not v._lb:
+                    vars_to_update.append(v)
+                elif ub is not v._ub:
+                    vars_to_update.append(v)
```
jsiirola, line 1419: This catches if the user re-assigns the lower / upper bound, but what about bounds that include mutable Params?  Does changing the Param value get propagated to updated bounds?

jsiirola (commented):
Some questions (although nothing that necessarily should block this PR)

pyomo/contrib/appsi/base.py
michaelbynum: great point. I'll update this.

pyomo/contrib/appsi/base.py
michaelbynum: that should get caught on the next two lines.

jsiirola: I don't think that's the case: that only catches if the *domain* changes, and not if the value of the bounds change:
```python
>>> m = ConcreteModel()
>>> m.x = Var()
>>> m.p = Param(mutable=True, initialize=1)
>>> m.x.bounds = -m.p, m.p
>>> m.x.bounds
(-1.0, 1.0)
>>> m.x.domain.get_interval()
(None, None, 0)
>>> m.p = 5
>>> m.x.bounds
(-5.0, 5.0)
>>> m.x.domain.get_interval()
(None, None, 0)
```

michaelbynum: oh, my mistake. You are right. The case of bounds with mutable parameters is actually caught with the `update_params` method and is handled differently for each solver. [Here](https://github.com/Pyomo/pyomo/blob/c471729e7b9fd6236e88e1913760f1fa4f2be121/pyomo/contrib/appsi/solvers/gurobi.py#L430) is how it is handled with Gurobi.

PR #2998, opened 2023-09-18
Add `Preformatted` class for logging preformatted messages

jsiirola:
## Summary/Motivation:
MindtPy and PyROS both want to emit solver iteration logs and do NOT want those messages to be reformatted or line-wrapped.  Their current solution is to hijack the entire Pyomo logging system to suppress the automatic formatting that the default log handler implements.  This PR simplifies that by defining a `Preformatted` class that users can wrap a preformatted log message in before sending it to the logger.  These messages will then be directly emitted (without being processed by the default formatter).  As `Preformatted` is castable to `str`, it is compatible with other user-defined log handlers / formatters (without requiring any specialized handling).

- define a `Preformatted` wrapper class for logging preformatted messages
- Update the default Pyomo formatter to ignore / do not reformat Preformatted messages.
- Add tests

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/log.py
+++ b/pyomo/common/log.py
@@ -107,11 +107,14 @@ class WrappingFormatter(logging.Formatter):
         super(WrappingFormatter, self).__init__(**kwds)
 
     def format(self, record):
+        msg = record.getMessage()
+        if record.msg.__class__ is not str and isinstance(record.msg, Preformatted):
+            return msg
+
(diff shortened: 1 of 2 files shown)


mrmundt (approved):

shermanjasonaf (approved):
Looks good to me.

PR #3023, opened 2023-10-26
fix scip results processing

michaelbynum:
### Summary

With SCIP 8.0.3, the results processing is currently broken:

```
import ampl_module_scip
import os
os.environ['PATH'] = os.pathsep.join([ampl_module_scip.bin_dir, os.environ['PATH']])
import pyomo.environ as pyo


m = pyo.ConcreteModel()
m.x = pyo.Var(domain=pyo.Integers)
m.objective = pyo.Objective( expr= m.x)
m.con = pyo.Constraint(expr= m.x >= 1)

opt = pyo.SolverFactory('scip')
results = opt.solve(m, tee=True)
```

Outputs:

```
WARNING: No dual information available when presolving was performed.
WARNING: Unexpected SCIP solver message: SCIP 8.0.3\x3a optimal solution;
objective 1; 0 simplex iterations; 0 branching nodes
  File "/Users/.../pyomo_with_scip.py", line 13, in <module>
    results = opt.solve(m, tee=True)
              ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/.../pyomo/pyomo/opt/base/solvers.py", line 661, in solve
    _model.solutions.load_from(
  File "/Users/.../pyomo/pyomo/core/base/PyomoModel.py", line 230, in load_from
    raise ValueError(
ValueError: Cannot load a SolverResults object with bad status: unknown
```

To correct this, I changed the bit of code that parses the solver message to be more flexible by checking if specific strings are in the message instead of checking if they are equal. In this case, Pyomo is looking for "optimal solution found", but the message is just "optimal solution". 

The other problem is the SCIP log is parsed to get the primal bound. However, the log parsing does not work properly if SCIP solves the problem in the preprocessing phase. To fix this, I added a try/except and simply leave the primal bound empty if the log is not parsed correctly.

### Legal Acknowledgement


pyomo/solvers/plugins/solvers/SCIPAMPL.py
```diff
@@ -288,115 +288,118 @@ def _postsolve(self):
...
-            if results.problem.sense == ProblemSense.minimize:
-                results.problem.lower_bound = results.solver.primal_bound
-            else:
-                results.problem.upper_bound = results.solver.primal_bound
+            try:
+                if results.problem.sense == ProblemSense.minimize:
+                    results.problem.lower_bound = results.solver.primal_bound
+                else:
+                    results.problem.upper_bound = results.solver.primal_bound
+            except AttributeError:
```
blnicho, line 381: @michaelbynum would you mind adding a short comment here explaining why this is needed or when you would end up here? Is this from a difference in solver versions?

michaelbynum: Done. Sorry - I was in a hurry yesterday.

michaelbynum: Sorry, I only added comments to the PR. Now, I've added comments to the code.

mrmundt:
@michaelbynum - Jenkins failures are not related. Our build environment was the problem. I have fixed it and flagged this for re-test.

blnicho (approved):

PR #3026, opened 2023-10-28
Fix exception due to interaction among Gurobi, Pint, Dask, and Threading

jsiirola:
## Summary/Motivation:
This PR fixes a bizarre interaction among GurobiPy, Pint, and Dask: When the python interpreter is exiting, gurobipy can interrogate objects looking at their `__class__` attribute.  If it hits the `_DeferredUnitsSingleton`, then it can trigger `pint` to be imported.  If `dask` is also installed, `pint` will import it, which in turn sets up threading.  Unfortunately, setting up threading registers an `atexit` callback, generating a `RuntimeError` (because the interpreter is shutting down).  This PR adds special handling to prevent importing `pint` in this situation, thereby avoiding the RuntimeError.

- Do not have `__class__` lookup on `_DeferredUnitsSingleton` trigger `pint` import

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/units_container.py
+++ b/pyomo/core/base/units_container.py
@@ -1514,6 +1514,19 @@ class _DeferredUnitsSingleton(PyomoUnitsContainer):
         # present, at which point this instance __class__ will fall back
         # to PyomoUnitsContainer (where this method is not declared, OR
         # pint is not available and an ImportError will be raised.
+        #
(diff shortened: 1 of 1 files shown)


mrmundt (approved):
This is the _weirdest_ interaction.

michaelbynum:
Wow

PR #3027, opened 2023-10-31
contrib.fbbt, pyomo.gdp: Adding new walker for compute_bounds_on_expr

emma58:
## Summary/Motivation:

The majority of the time in the `gdp.bigm` transformation is spent calculating M values, mostly in the function (from fbbt) `compute_bounds_on_expr`. This PR adds a new walker to contrib.fbbt specifically for computing bounds on expressions based on Var bounds. It is very similar to the leaf-to-root walker, but not identical as it is designed for only a single pass (unlike in actual fbbt), and it caches only the leaf node bounds. Moving the bigm transformations onto it helps performance in calculating M values significantly.

- Adds `ExpressionBoundsVisitor` to contrib.fbbt
- Has `compute_bounds_on_expr` wrap `ExpressionBoundsVisitor` instead of `_FBBTVisitorLeafToRoot`
- (unrelatedly) rewrites `_FBBTVisitorLeafToRoot`, moving it onto the `StreamBasedExpressionVisitor` base class
- Rewrites the implementation of `mul` in `interval.py`: This is actually quite significant performance-wise.
- Rewrites bigm calculations in `gdp.bigm` and `gdp.mbigm` to store an instance of `ExpressionBoundsVisitor` for all of `apply_to` to take advantage of caching, rather than relying on `compute_bounds_on_expr`.

### Legal Acknowledgement


pyomo/contrib/fbbt/fbbt.py
```diff
@@ -1084,7 +970,77 @@ def _check_and_reset_bounds(var, lb, ub):
...
+
+def _before_NPV(visitor, child):
+    if child in visitor.bnds_dict:
+        return False, None
+    val = value(child)
+    visitor.bnds_dict[child] = (val, val)
+    return False, None
+
+
+def _before_other(visitor, child):
```
michaelbynum, line 1011: What is this handler for?

pyomo/contrib/fbbt/fbbt.py
```diff
@@ -1084,7 +970,77 @@ def _check_and_reset_bounds(var, lb, ub):
...
+
+def _before_var(visitor, child):
+    if child in visitor.bnds_dict:
+        return False, None
+    elif child.is_fixed() and not visitor.ignore_fixed:
+        lb = value(child.value)
+        ub = lb
+    else:
+        lb = value(child.lb)
+        ub = value(child.ub)
```
michaelbynum, line 989: Same as above. I think the call to `value` can go away now.

pyomo/contrib/fbbt/expression_bounds_walker.py
```diff
@@ -0,0 +1,282 @@
...
+        ignore fixed status and use the bounds. Set to 'True' if you do not
+        anticipate the fixed status of Variables to change for the duration that
+        the computed bounds should be valid.
+    """
+
+    def __init__(
+        self,
+        leaf_bounds=None,
+        feasibility_tol=1e-8,
+        use_fixed_var_values_as_bounds=False,
```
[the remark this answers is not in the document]
emma58: I did False because it seems safer to me... Because then the bounds are right regardless of if variables get unfixed later. Basically this way we get positive confirmation from a user that they know things are potentially invalid if they unfix Vars.

pyomo/contrib/fbbt/expression_bounds_walker.py
```diff
@@ -0,0 +1,282 @@
...
+            val = child.value
+            if val is None:
+                raise ValueError(
+                    "Var '%s' is fixed to None. This value cannot be used to "
+                    "calculate bounds." % child.name
+                )
+            leaf_bounds[child] = (child.value, child.value)
+        else:
+            lb = value(child.lb)
+            ub = value(child.ub)
```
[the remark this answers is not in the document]
emma58: Whoops, good point

pyomo/contrib/fbbt/expression_bounds_walker.py
```diff
@@ -0,0 +1,282 @@
...
+    @staticmethod
+    def _before_complex(visitor, child):
+        raise ValueError(
+            f"Cannot compute bounds on expressions containing "
+            f"complex numbers. Encountered when processing {child}"
+        )
+
+    @staticmethod
+    def _before_npv(visitor, child):
+        return False, (value(child), value(child))
```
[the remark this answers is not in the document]
emma58: Because it was so fun the first time... :P This is a good point--I'll change it.

pyomo/contrib/fbbt/fbbt.py
emma58: The previous walker basically had a default that if it didn't understand what was going on, it still returned bounds of `(-inf, inf)`. So this maintains that behavior. I'm not sure that it's a good idea long-term though?

pyomo/contrib/fbbt/fbbt.py
```diff
@@ -1084,7 +970,77 @@ def _check_and_reset_bounds(var, lb, ub):
...
+
+def _register_new_before_child_handler(visitor, child):
+    handlers = _before_child_handlers
+    child_type = child.__class__
+    if child.is_variable_type():
+        handlers[child_type] = _before_var
+    elif not child.is_potentially_variable():
+        handlers[child_type] = _before_NPV
+    else:
+        handlers[child_type] = _before_other
```
jsiirola, line 1031: I feel like if we hit this, then we should warn about an unknown / unhandled expression type?

jsiirola (approved):
I like this, but I do wonder that if we are going to start passing the visitor as an argument to all of the handlers, then why are they not just methods on the visitor class?  That would look more "normal".

michaelbynum (approved):

pyomo/contrib/fbbt/fbbt.py
emma58: I think we should, but that's actually going to require a more significant rewrite because right now this is catching both things we don't understand and things that we do understand, just don't need a beforeChild handler for (like SumExpression). So I think it would make the most sense to move this onto the new BeforeChildDispatcher base class and then finally get pedantic about when we're confused and when we just don't care. But I'd rather that be a separate PR.

PR #3034, opened 2023-11-13
Fix MindtPy bugs

ZedongPeng:
## Fixes # .
1. [fix gurobi single tree termination check bug](https://github.com/Pyomo/pyomo/commit/905503b13907a610a1e7f9d8620ee88127551ec5)
2. [fix Gurobi single tree cycle handling](https://github.com/Pyomo/pyomo/commit/3b01104f586f9a33d1386ff89c5c3e57b3ae7fc7)
3. [improve copy_var_list_values function](https://github.com/Pyomo/pyomo/commit/a66f955500686fef09d8605571ef6f5c238101f8)
4. [fix bug in feasibility pump method](https://github.com/Pyomo/pyomo/commit/cefd4a66a06711e90005d20cd470efca762754a6)
5. [add special handle for infeasible relaxed NLP](https://github.com/Pyomo/pyomo/commit/a767f281acd08629aea99f6c616e59b11372d2f5) 
6. [update the log format of infeasible fixed NLP subproblems](https://github.com/Pyomo/pyomo/commit/f766c0a9d6f66da8fea8b11f277bb719ef94f48f)
7. [create new copy_var_list_values function](https://github.com/Pyomo/pyomo/commit/a6e92c53e63b9febad79e8a19a5230c8c308328a)

## Summary/Motivation:
This PR fixes several bugs in MindtPy according to the benchmark results.

### Legal Acknowledgement

9. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.


blnicho:
This is waiting for #2988 to be merged first

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -840,6 +848,29 @@ def init_rNLP(self, add_oa_cuts=True):
...
+
+        # Sometimes, the NLP solver might be trapped in a infeasible solution if the objective function is nonlinear and partition_obj_nonlinear_terms is True. If this happens, we will use the original objective function instead.
+        if (
+            subprob_terminate_cond == tc.infeasible
+            and config.partition_obj_nonlinear_terms
+            and self.rnlp.MindtPy_utils.objective_list[0].expr.polynomial_degree()
+            not in self.mip_objective_polynomial_degree
+        ):
+            config.logger.info(
+                'Initial relaxed NLP problem is infeasible. This might be related to partition_obj_nonlinear_terms. Try to solve it again without partitioning nonlinear objective function.'
```
emma58, line 860: Suggestion: change 'try' to 'trying'

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -1626,7 +1683,11 @@ def solve_main(self):
                         "No-good cuts are added and GOA algorithm doesn't converge within the time limit. "
                         'No integer solution is found, so the CPLEX solver will report an error status. '
                     )
-            return None, None
+            # Value error will be raised if the MIP problem is unbounded and appsi solver is used when loading solutions. Although the problem is unbounded, a valid result is provided and we do not return None to let the algorithm continue.
+            if 'main_mip_results' in dir():
```
emma58, line 1687: `dir()` returns a list, so this `if` statement is linear time, which isn't ideal. Is there a more efficient way to check that you have MIP results?

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -805,6 +810,9 @@ def MindtPy_initialization(self):
             self.integer_list.append(self.curr_int_sol)
             fixed_nlp, fixed_nlp_result = self.solve_subproblem()
             self.handle_nlp_subproblem_tc(fixed_nlp, fixed_nlp_result)
+            self.int_sol_2_cuts_ind[self.curr_int_sol] = list(
+                range(1, len(self.mip.MindtPy_utils.cuts.oa_cuts) + 1)
+            )
```
emma58, line 815: Why not just store the number of OA cuts for each of these solutions? It seems inefficient to store a list of indices when you can always construct a generator from one number, but maybe I'm wrong?

pyomo/contrib/mindtpy/util.py
```diff
@@ -963,3 +970,44 @@ def generate_norm_constraint(fp_nlp_model, mip_model, config):
...
+        elif ignore_integrality and v_to.is_integer():
+            v_to.set_value(value(v_from, exception=False), skip_validation=True)
+        elif v_to.is_integer() and (
+            math.fabs(var_val - rounded_val) <= config.integer_tolerance
+        ):
+            v_to.set_value(rounded_val)
+        elif abs(var_val) <= config.zero_tolerance and 0 in v_to.domain:
+            v_to.set_value(0)
+        else:
+            raise ValueError("copy_var_list_values failed.")
```
emma58, line 1013: It looks like this code is copied three times in this PR: Here, in the single-tree implementation and above for the solution pools. Could this part be a utility function to reduce the redundant code?

emma58, line 1013: It will be helpful to add which Vars you were copying from and to in this error message.

emma58 (commented):
A few comments and questions. I also think it would be good to add tests for the bugs you are addressing in this PR.

pyomo/contrib/mindtpy/algorithm_base_class.py
ZedongPeng: `int_sol_2_cuts_ind` is a `dict`, with `self.curr_int_sol` list as the `key` and `index list` as the `value`. 
Do you want me to improve the definition of the key or the value? I think improving the value is doable.

pyomo/contrib/mindtpy/algorithm_base_class.py
ZedongPeng: Is it better to use `locals()`, which returns a dict?

emma58: If you're going to use `in` then a dict would be much better, yes.

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -1100,8 +1100,9 @@ def solve_subproblem(self):
                 tolerance=config.constraint_tolerance,
             )
         except InfeasibleConstraintException as e:
+            config.logger.error(e, exc_info=True)
             config.logger.error(
-                str(e) + '\nInfeasibility detected in deactivate_trivial_constraints.'
+                'Infeasibility detected in deactivate_trivial_constraints.'
```
emma58, line 1105: In this pattern, it is still good to include the text from the original error. Here you are hiding it from the user completely and replacing it with your own.

pyomo/contrib/mindtpy/algorithm_base_class.py
ZedongPeng: That's a good point. However, one integer is not enough. We need two integers denoting the starting index and ending index respectively. I will try to implement it.

mrmundt:
@ZedongPeng - the `gurobi` failures are not your fault. #3053 is going to temporarily fix them; Michael is going to (more) permanently fix them soon.

pyomo/contrib/mindtpy/extended_cutting_plane.py
```diff
@@ -140,7 +140,7 @@ def all_nonlinear_constraint_satisfied(self):
                     lower_slack = nlc.lslack()
                 except (ValueError, OverflowError) as e:
                     # Set lower_slack (upper_slack below) less than -config.ecp_tolerance in this case.
-                    config.logger.error(e)
+                    config.logger.error(e, exc_info=True)
                     lower_slack = -10 * config.ecp_tolerance
```
ZedongPeng, line 144: yes

ZedongPeng:
Most of the uncovered lines are `logging.errors`. I added the unit tests for new `set_var_value` function.

ZedongPeng:
I think this PR is ready for a second review.

emma58 (commented):
A few more comments, and it would still be best if you could increase coverage: You can turn up the logging level and test some of those error messages.

pyomo/contrib/mindtpy/util.py
```diff
@@ -963,3 +946,99 @@ def generate_norm_constraint(fp_nlp_model, mip_model, config):
...
+
+def set_var_value(v_to, var_val, integer_tolerance, zero_tolerance, ignore_integrality):
+    """This function copies variable value from one to another.
+    Rounds to Binary/Integer if necessary.
+    Sets to zero for NonNegativeReals if necessary.
+
+    NOTE: PEP 2180 changes the var behavior so that domain /
+    bounds violations no longer generate exceptions (and
+    instead log warnings).  This means that the following will
+    always succeed and the ValueError should never be raised.
```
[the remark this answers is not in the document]
ZedongPeng: I think both docstring and plain comments will work.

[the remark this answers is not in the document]
emma58: I'm just doesn't feel like documentation to me--it seems more like a comment. It's also not clear what "the following" is referring to anymore since there's no longer a try-except pattern.

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -1626,7 +1684,11 @@ def solve_main(self):
                         "No-good cuts are added and GOA algorithm doesn't converge within the time limit. "
                         'No integer solution is found, so the CPLEX solver will report an error status. '
                     )
-            return None, None
+            # Value error will be raised if the MIP problem is unbounded and appsi solver is used when loading solutions. Although the problem is unbounded, a valid result is provided and we do not return None to let the algorithm continue.
+            if 'main_mip_results' in locals():
+                return self.mip, main_mip_results
```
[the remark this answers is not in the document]
ZedongPeng: I tried. However, it's not easy to find a problem to generate a model with an unbounded MIP problem, with the initialization strategy. Tbh, I forgot which instance resulted in this situation in the benchmark tests.

pyomo/repn/tests/ampl/test_nlv2.py
```diff
@@ -1054,6 +1054,40 @@ def test_log_timing(self):
...
+      [+   #.##] Constraint c2
+      [+   #.##] Categorized model variables: 14 nnz
+      [+   #.##] Set row / column ordering: 6 var [3, 1, 2 R/B/Z], 3 con [2, 1 L/NL]
+      [+   #.##] Generated row/col labels & comments
+      [+   #.##] Wrote NL stream
+      [    #.##] Generated NL representation
+""",
+            re.sub(r'\d\.\d\d\]', '#.##]', LOG.getvalue()),
+        )
+
```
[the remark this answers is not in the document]
ZedongPeng: This came from merging the `main` branch into the current `benchmark` branch. It is a little bit wired. Is it good to manually remove this?

pyomo/contrib/mindtpy/algorithm_base_class.py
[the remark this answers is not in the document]
emma58: You could always make the whole problem unbounded--then you would be sure to hit this (as long as you didn't discover it was unbounded during initialization I guess). I think there are a couple small examples that might be useful to build off of in `test_gdpopt.py`. Another option would be to just directly call the `solve_main` method and skip initialization.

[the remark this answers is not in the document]
ZedongPeng: This code will not be triggered for all unbounded problems. It only happens when the problem is unbounded and the heuristic method in the solver also finds a feasible solution, **as follows**. That's why the problem is unbounded and still calls the `load_solutions` function. I can manually reproduce this situation by changing the code of MindtPy.  However, I didn't find an instance that could result in this situation without any change. It might not be worth to introduce some redundant code just to trigger this error. I will try to increase the code coverage in other parts.
```
Gurobi Optimizer version 10.0.2 build v10.0.2rc0 (mac64[x86])

CPU model: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
Thread count: 6 physical cores, 12 logical processors, using up to 12 threads

Optimize a model with 4 rows, 6 columns and 8 nonzeros
Model fingerprint: 0xb497cb3e
Variable types: 5 continuous, 1 integer (0 binary)
Coefficient statistics:
  Matrix range     [1e+00, 4e+00]
  Objective range  [1e+00, 1e+00]
  Bounds range     [0e+00, 0e+00]
  RHS range        [3e+00, 2e+01]
Found heuristic solution: objective -3.60103e+09
Presolve removed 3 rows and 3 columns
Presolve time: 0.00s

Explored 0 nodes (0 simplex iterations) in 0.00 seconds (0.00 work units)
Thread count was 1 (of 12 available processors)

Solution count 1: -3.60103e+09 
No other solutions better than 0

Model is unbounded
Best objective -3.601030018124e+09, best bound -, gap -
```

bernalde:
All test failures correspond to python >= 3.10 and the GDPOpt LBB test when importing Z3. This is unrelated to this PR but might be good to look into @emma58

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -797,14 +805,18 @@ def MindtPy_initialization(self):
...
-                config.logger.error(e)
+                config.logger.error(e, exc_info=True)
                 raise ValueError(
                     'The initial integer combination is not provided or not complete. '
                     'Please provide the complete integer combination or use other initialization strategy.'
                 )
             self.integer_list.append(self.curr_int_sol)
             fixed_nlp, fixed_nlp_result = self.solve_subproblem()
             self.handle_nlp_subproblem_tc(fixed_nlp, fixed_nlp_result)
+            self.int_sol_2_cuts_ind[self.curr_int_sol] = [
```
bernalde, line 816: The name of this function is unclear

pyomo/contrib/mindtpy/single_tree.py
```diff
@@ -309,9 +269,9 @@ def add_lazy_affine_cuts(self, mindtpy_solver, config, opt):
...
                     mc_eqn = mc(constr.body)
                 except MCPP_Error as e:
+                    config.logger.error(e, exc_info=True)
                     config.logger.debug(
-                        'Skipping constraint %s due to MCPP error %s'
-                        % (constr.name, str(e))
+                        'Skipping constraint %s due to MCPP error' % (constr.name)
                     )
                     continue  # skip to the next constraint
                 # TODO: check if the value of ccSlope and cvSlope is not Nan or inf. If so, we skip this.
```
bernalde, line 277: Should we address this now in this PR or leave it for later?

pyomo/contrib/mindtpy/single_tree.py
```diff
@@ -718,6 +678,7 @@ def __call__(self):
         # Lazy constraints separated when processing a MIP start will be discarded after that MIP start has been processed.
         # This means that the callback may have to separate the same constraint again for the next MIP start or for a solution that is found later in the solution process.
         # https://www.ibm.com/docs/en/icos/22.1.1?topic=SSSA5P_22.1.1/ilog.odms.cplex.help/refpythoncplex/html/cplex.callbacks.LazyConstraintCallback-class.htm
+        # For the MINLP3_simple example, all the solutions are obtained from mip_start (solution source). Therefore, it will not go to a branch and bound process.Cause an error output.
```
bernalde, line 681: This comment is to mention some reasons why we trigger tests?

pyomo/contrib/mindtpy/outer_approximation.py
```diff
@@ -96,7 +96,10 @@ def check_config(self):
     def initialize_mip_problem(self):
         '''Deactivate the nonlinear constraints to create the MIP problem.'''
```
bernalde, line 97: We use """ """ for function documentation

bernalde (approved):
Very small cosmetic changes in this request. Maybe address to restart tests as previous failures were unrelated to this PR. Use the chance to increase coverage if easy to do.

pyomo/contrib/mindtpy/single_tree.py
ZedongPeng: To be more precise, the algorithm will terminate during the MIP start process as soon as the bounds converge. In such scenarios, we utilize `self.abort()` to terminate CPLEX. As a consequence, the `results.solver.status` will be set to `error`, accompanied by an error message. Our initial choice to use `self.abort()` was driven by the need for immediate termination upon meeting the convergence criteria. However, I am currently uncertain about the extent of the difference it would make if we allowed CPLEX to terminate on its own.

ZedongPeng: Since the error will be captured by the `try except` command. We don't need to worry about it and it is still the most efficient way to terminate CPLEX. I add a TODO here. We can decide based on the benchmark results in the future.

pyomo/contrib/mindtpy/single_tree.py
ZedongPeng: This has already been checked in the following code. We can remove the TODO.

ZedongPeng:
Could you help me to check why `test_ECP_add_slack` failed only in `linux/pypy-39`? @emma58 @jsiirola
I cannot find the detailed log and the test succeeds in all other environments and my local environment.

mrmundt:
@ZedongPeng - there are issues in general with pypy right now. #3087 is going to fix them.

ZedongPeng:
Hi @mrmundt,

I've noticed that the `pypy-3.9` test continues to fail even after merging #3087. Interestingly, when I ran the tests locally, they all passed successfully. Could you advise me on how to investigate the reasons behind the test exceeding the `timelimit`?

Thank you for your assistance.

bernalde:
The same error seems to be happening although merging master. It always times out in the test_feaspump code from MindtPy, suggesting we are the ones introducing issues here. We need to replicate the testing environment locally to check what's going on. Can we replicate it (in the scenario we can get our hands on a Linux machine)? Somehow, installing promo and some packages via PyPy in Python 3.9 might suggest it is the way forward. Some indication might be useful.

ZedongPeng:
The `osx/3.9` test encountered a failure, which appears to be due to an internet connectivity issue. The specific error message was: `Error: nodename nor servname provided, or not known (api.github.com:443)`. 
Based on the overall progress, I believe this PR is now ready for merging. Could you please give it a final review, @emma58?

pyomo/contrib/mindtpy/global_outer_approximation.py
```diff
@@ -108,4 +108,5 @@ def deactivate_no_good_cuts_when_fixing_bound(self, no_good_cuts):
             if self.config.use_tabu_list:
                 self.integer_list = self.integer_list[:valid_no_good_cuts_num]
         except KeyError as e:
-            self.config.logger.error(str(e) + '\nDeactivating no-good cuts failed.')
+            self.config.logger.error(e, exc_info=True)
+            self.config.logger.error('Deactivating no-good cuts failed.')
```
emma58, line 112: How can this fail? Wouldn't this be a `DeveloperError` (if you need an error here at all) since mindtpy can keep track of what cuts it has added?

emma58 (commented):
A few questions, but overall this looks good. Thanks for the added tests!

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -102,12 +102,14 @@ def __init__(self, **kwds):
...
         # We store bounds, timing info, iteration count, incumbent, and the
-        # expression of the original (possibly nonlinear) objective function.
+        # Expression of the original (possibly nonlinear) objective function.
         self.results = SolverResults()
         self.timing = Bunch()
         self.curr_int_sol = []
         self.should_terminate = False
         self.integer_list = []
+        # Dictionary {integer solution (list): [cuts begin index, cuts end index] (list)}
+        self.integer_solution_to_cuts_index = dict()
```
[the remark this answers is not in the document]
ZedongPeng: Thanks for pointing it out. The type of integer solution is `tuple` instead of `list`.

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -1241,7 +1290,7 @@ def handle_subproblem_infeasible(self, fixed_nlp, cb_opt=None):
         #         elif var.has_lb() and abs(value(var) - var.lb) < config.absolute_bound_tolerance:
         #             fixed_nlp.ipopt_zU_out[var] = -1
 
-        config.logger.info('Solving feasibility problem')
+        # config.logger.info('Solving feasibility problem')
```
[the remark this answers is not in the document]
ZedongPeng: Nope.

pyomo/contrib/mindtpy/global_outer_approximation.py
ZedongPeng: Thanks for pointing this out. I checked the log of the benchmark tests and found this error came out when `timelimit` was reached. Let me try to fix it before merging this PR. Thanks.

ZedongPeng: The `KeyError` happens when no feasible solution has been found within the `timelimit`. I added one more condition for the `fix_dual_bound` function in the **OA, GOA, LP/NLP B&B and GLP/NLP B&B** methods. 
I also removed `fix_dual_bound` for the **ECP** method since we will never add no-good cuts in the **ECP** method.
I would appreciate it if @bernalde can give it a double-check.

ZedongPeng:
I have already run the black format locally. Are there any new requirements for the black format?
```
~/Github/pyomo/pyomo/contrib/mindtpy benchmark *1 !2 ❯ black -S -C ./
All done! ✨ 🍰 ✨
40 files left unchanged.
```

mrmundt:
@ZedongPeng - check your black version. They released an update this month that made some changes.

emma58 (approved):

pyomo/contrib/mindtpy/util.py
```diff
@@ -965,3 +929,101 @@ def generate_norm_constraint(fp_nlp_model, mip_model, config):
...
+    skip_stale=False,
+    skip_fixed=True,
+    ignore_integrality=False,
+):
+    """Copy variable values from one list to another.
+    Rounds to Binary/Integer if necessary
+    Sets to zero for NonNegativeReals if necessary
+
+    from_list : list
+        The variables that provides the values to copy from.
```
blnicho, line 947:
```suggestion
        The variables that provide the values to copy from.
```

pyomo/contrib/mindtpy/util.py
```diff
@@ -965,3 +929,101 @@ def generate_norm_constraint(fp_nlp_model, mip_model, config):
...
+        var.set_value(var.ub)
+    elif ignore_integrality and var.is_integer():
+        var.set_value(var_val, skip_validation=True)
+    elif var.is_integer() and (math.fabs(var_val - rounded_val) <= integer_tolerance):
+        var.set_value(rounded_val)
+    elif abs(var_val) <= zero_tolerance and 0 in var.domain:
+        var.set_value(0)
+    else:
+        raise ValueError(
+            "copy_var_list_values failed with variable {}, value = {} and rounded value = {}"
```
blnicho, line 1027:
```suggestion
            "set_var_valid_value failed with variable {}, value = {} and rounded value = {}"
```

blnicho (approved):
@ZedongPeng I have a couple minor edits but otherwise I think this looks good.

PR #3037, opened 2023-11-14
NLv2: add linear presolve and general problem scaling support

jsiirola:
## Summary/Motivation:
This PR updates the NLv2 writer to provide optional model scaling and a basic linear presolve.  

While some solvers (notably Ipopt) support user-specified problem scaling through suffixes, that is not universally supported by all solvers.  This PR adds scaling to the NL write step, looking at the `scaling_factor` Suffix and writing out the model in scaled space.

This also implements a basic linear presolve / variable aggregator.  This presolve will:
- fix variables whose lower / upper bounds are equal
- fix variables appearing as the sole unfixed variable in a linear equality expression
- identify linear equality expressions with exactly 2 unfixed variables and replace one variable with the other (variable aggregation without fill-in).

Because both of these features require providing additional information the the solution reader in order to correctly import the solver result, these features are available ONLY through the "new" `.write()` API and not through the legacy `__call__()` API.

- Support outputting a scaled model based on the `scaling_factor` suffix
- Add a basic linear presolver
- Redesign how suffixes are processed (by deferring when the suffix values are "compiled" [gathered])

### Legal Acknowledgement


pyomo/repn/plugins/nl_writer.py
```diff
@@ -264,6 +327,15 @@ def __call__(self, model, filename, solver_capability, io_options):
...
 
         config = self.config(io_options)
+
+        # There is no (convenient) way to pass the scaling factors or
+        # information about presolved variables back to the solver
+        # through the old "call" interface (since solvers that used that
+        # interface predated scaling / presolve).  We will play it safe
+        # and disable scaling / presolve when called through this API
+        config.scale_model = False
+        config.linear_presolve = False
```
michaelbynum, line 337: Should this issue a warning if they were `True`?

pyomo/repn/plugins/nl_writer.py
```diff
@@ -1479,6 +1656,130 @@ def _count_subexpression_occurrences(self):
...
+                con_id, info = two_var.popitem()
+                expr_info, lb = info
+                _id, coef = expr_info.linear.popitem()
+                id2, coef2 = expr_info.linear.popitem()
+                # For no particularly good reason, we will solve for
+                # (and substitute out) the variable with the smaller
+                # magnitude)
+                if abs(coef2) < abs(coef):
+                    _id, id2 = id2, _id
+                    coef, coef2 = coef2, coef
```
michaelbynum, line 1709: I would recommend substituting out the variable with the coefficient closest to 1 for numerical stability (i.e., avoid dividing by large or small numbers).

pyomo/repn/plugins/nl_writer.py
```diff
@@ -1479,6 +1656,130 @@ def _count_subexpression_occurrences(self):
...
+                substitutions_by_linear_var[x].add(_id)
+                eliminated_vars[_id] = expr_info
+                logger.debug(
+                    "NL presolve: substituting %s := %s*%s + %s",
+                    var_map[_id],
+                    a,
+                    var_map[x],
+                    b,
+                )
+                # Tighten variable bounds
```
michaelbynum, line 1724: What about the variable domains? What if the variable being substituted out is binary or integer? Apologies if it's in here and I missed it.

pyomo/repn/plugins/nl_writer.py
```diff
@@ -1576,6 +1877,16 @@ def __str__(self):
     def __repr__(self):
         return str(self)
 
+    def __eq__(self, other):
```
michaelbynum, line 1880: Are we okay making this class not hashable (I'm fine with it; I just wanted to make sure)?

pyomo/repn/plugins/nl_writer.py
```diff
@@ -252,6 +304,17 @@ class NLWriter(object):
         variables'.""",
         ),
     )
+    CONFIG.declare(
+        'linear_presolve',
+        ConfigValue(
+            default=False,
```
michaelbynum, line 310: Can we change this default to `True` at a later date?

michaelbynum (commented):
I am very excited for this functionality. I left a few (hopefully useful) comments/questions, but this looks great.

pyomo/repn/plugins/nl_writer.py
michaelbynum: I realize that "closest to 1" is not clear. I think I've measured this using log10 in the past. In other words, we want `log10(coef)` to be close to 0.

pyomo/repn/plugins/nl_writer.py
jsiirola: I needed it comparable for testing, but that is a good point on hashability.  I added a reasonable implementation of `__hash__`.

pyomo/repn/plugins/nl_writer.py
michaelbynum: Ahh, good point. Maybe we can deprecate the `__call__` after the solver refactor is complete.

jsiirola: That is my hope.

michaelbynum (approved):

Robbybp:
@jsiirola Is this available through e.g. `pyo.SolverFactory("ipopt").solve(m, linear_presolve=True)`? I tried this and did not see a reduction in the number of variables reported by Ipopt. I assume this uses the `__call__` interface, so I should not expect the `linear_presolve` to be available? If this is the case, how can I go about testing this functionality?

jsiirola:
@Robbybp, alas, no - this is not available through any existing solver.  The problem is that the writer needs to send additional information to the solver so that the solver knows how to populate the presolved variables after the solve.  Unfortunately, the old `__call__` / solver interfaces to not make it easy to add more information to get passed from the writer to the solver.

@michaelbynum and @mrmundt are working on redesigning the standard solver interfaces (based on APPSI) to - among many other things - be able to use the presolver.  I believe that they have a draft of a new ipopt interface on a branch.  Alternatively, if you are just interested in seeing how ipopt behaves (and not getting the answer back), you can use the new writer to write the nl file and run ipopt manually.

PR #3041, opened 2023-11-19
Fix two bugs in incidence analysis

### LINKED ISSUE #3040: Dulmage-Mendelsohn decomposition does not automatically put matching on the diagonal
MWE: Run the following script in the `examples` subdirectory of this repo.
```python
from run_distill import instance
from pyomo.contrib.incidence_analysis import IncidenceGraphInterface
from pyomo.contrib.incidence_analysis.interface import get_structural_incidence_matrix
import matplotlib.pyplot as plt 

igraph = IncidenceGraphInterface(instance, include_inequality=False, linear_only=True)
var_dmp, con_dmp = igraph.dulmage_mendelsohn()
vorder = sum(var_dmp, []) 
corder = sum(con_dmp, []) 

# My intention is for this to have a non-zero diagonal
imat = get_structural_incidence_matrix(vorder, corder, linear_only=True)
fig, ax = plt.subplots()
ax.spy(imat, markersize=1)
plt.show()
```
This displays the following image:
dm_mwe
Note the kink in the "diagonal" entries around (3500, 3500). I would like to be able to recover the maximum matching from the Dulmage-Mendelsohn decomposition with
```python
matching = list(zip(
    con_dmp.underconstrained + con_dmp.square + con_dmp.overconstrained,
    var_dmp.underconstrained + var_dmp.square + var_dmp.underconstrained,
))
```
This is half the reason that I broke the decomposition down into four subsets instead of only three. This should be a simple change in `pyomo.contrib.incidence_analysis.common.dulmage_mendelsohn`.

Robbybp:
## Summary/Motivation:
- The subsets returned by the Dulmage-Mendelsohn decomposition should be ordered so that zipping the constraint and variable subsets allows us to recover a maximum matching.
- `get_incident_variables(expr, linear_only=True)` should correctly filter out variables that have coefficients of zero

- Change the order of nodes returned in the "square subset" of the second bipartite set in `pyomo.contrib.common.dulmage_mendelsohn` 
- Fix bug in behavior with `linear_only=True`

Re-running the motivating example in #3040,
```python
from run_distill import instance
from pyomo.contrib.incidence_analysis import IncidenceGraphInterface
from pyomo.contrib.incidence_analysis.interface import get_structural_incidence_matrix
import matplotlib.pyplot as plt 

igraph = IncidenceGraphInterface(instance, include_inequality=False, linear_only=True)
var_dmp, con_dmp = igraph.dulmage_mendelsohn()
vorder = sum(var_dmp, []) 
corder = sum(con_dmp, []) 

vmatch = var_dmp.underconstrained + var_dmp.square + var_dmp.overconstrained
cmatch = con_dmp.underconstrained + con_dmp.square + con_dmp.overconstrained

matching = zip(vmatch, cmatch)
# Assert that zipping the subsets gives us a valid matching
assert len(vmatch) == len(ComponentSet(vmatch))
assert len(cmatch) == len(ComponentSet(cmatch))
for var, con in matching:
    assert var in ComponentSet(igraph.get_adjacent_to(con))

# My intention is for this to have a non-zero diagonal
imat = get_structural_incidence_matrix(vorder, corder, linear_only=True)
fig, ax = plt.subplots()
ax.spy(imat, markersize=1)
plt.show()
```
we get:
distill_imat

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/incidence_analysis/common/dulmage_mendelsohn.py
+++ b/pyomo/contrib/incidence_analysis/common/dulmage_mendelsohn.py
@@ -90,7 +90,8 @@ def dulmage_mendelsohn(bg, top_nodes=None, matching=None):
     _filter.update(b_unmatched)
     _filter.update(b_matched_with_reachable)
     t_other = [t for t in top_nodes if t not in _filter]
-    b_other = [b for b in bot_nodes if b not in _filter]
+    b_other = [matching[t] for t in t_other]
+    #b_other = [b for b in bot_nodes if b not in _filter]
 
     return (
         (t_unmatched, t_reachable, t_matched_with_reachable, t_other),
--- a/pyomo/contrib/incidence_analysis/common/dulmage_mendelsohn.py
+++ b/pyomo/contrib/incidence_analysis/common/dulmage_mendelsohn.py
@@ -91,7 +91,6 @@ def dulmage_mendelsohn(bg, top_nodes=None, matching=None):
     _filter.update(b_matched_with_reachable)
     t_other = [t for t in top_nodes if t not in _filter]
     b_other = [matching[t] for t in t_other]
-    #b_other = [b for b in bot_nodes if b not in _filter]
 
     return (
         (t_unmatched, t_reachable, t_matched_with_reachable, t_other),
--- a/pyomo/contrib/incidence_analysis/dulmage_mendelsohn.py
+++ b/pyomo/contrib/incidence_analysis/dulmage_mendelsohn.py
@@ -70,6 +70,21 @@ def dulmage_mendelsohn(matrix_or_graph, top_nodes=None, matching=None):
     - **overconstrained** - The columns matched with *possibly* unmatched
       rows (unmatched and overconstrained rows)
 
+    While the Dulmage-Mendelsohn decomposition does not specify an order within
+    any of these subsets, the order returned by this function preserves the
+    maximum matching that is used to compute the decomposition. That is, zipping
+    "corresponding" row and column subsets yields pairs in this maximum matching.
+    For example:
+
+    >>> row_dmpartition, col_dmpartition = dulmage_mendelsohn(matrix)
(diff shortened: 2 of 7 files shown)


Robbybp:
Doctests are failing with the following error:
```
Document: contributed_packages/incidence/dulmage_mendelsohn
-----------------------------------------------------------
**********************************************************************
File "../../pyomo/contrib/incidence_analysis/dulmage_mendelsohn.py", line ?, in default
Failed example:
    row_dmpartition, col_dmpartition = dulmage_mendelsohn(matrix)
Exception raised:
      File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/doctest.py", line 1336, in __run
        exec(compile(example.source, filename, "single",
      File "<doctest default[0]>", line 1, in <module>
        row_dmpartition, col_dmpartition = dulmage_mendelsohn(matrix)
    NameError: name 'dulmage_mendelsohn' is not defined
**********************************************************************
```
The culprit code is a few lines that I added as an example, not intended for doctest. These lines don't get picked up when I run `make doctest` locally, so I'm a little bit confused why they're getting picked up here. I will add a `:skipif: True` annotation to hopefully get tests passing, although this does not seem like a good solution.

Maybe the philosophy that I'm violating here is "any code in the docs should be tested." If that is what we're enforcing, I can add the appropriate additional context to make these examples actually runnable, although I'm still confused why my code without the `.. doctest::` annotation is getting doctested.

jsiirola:
> Maybe the philosophy that I'm violating here is "any code in the docs should be tested." If that is what we're enforcing, I can add the appropriate additional context to make these examples actually runnable, although I'm still confused why my code without the `.. doctest::` annotation is getting doctested.

I would argue for testing (at least executing) all code from documentation, so adding a (hidden) block to import `dulmage_mendelsohn` would be a simple / good thing

blnicho (approved):

jsiirola (approved):

PR #3056, opened 2023-11-29
Remove presolve-eliminated variables from named expressions

### LINKED ISSUE #3055: Incorrect degrees of freedom with linear presolve
## Summary

Testing out `NLWriter().write(model, file, linear_presolve=True)`, I have noticed that sometimes a variable that is eliminated is still written to the nl file. 

```python
import pyomo.environ as pyo 
import pyomo.repn.plugins.nl_writer as nl_writer
import subprocess

# We use one constraint to eliminate one variable,
# but apparently no variables get eliminated.

m = pyo.ConcreteModel()

m.x = pyo.Var([1, 2, 3], initialize=1, bounds=(0, 10))

m.subexpr = pyo.Expression(pyo.Integers)
m.subexpr[1] = m.x[1] + m.x[2]

m.eq = pyo.Constraint(pyo.Integers)
m.eq[1] = m.x[1] == 7
m.eq[2] = m.x[3] == 0.1 * m.subexpr[1] * m.x[2]

m.obj = pyo.Objective(expr=m.x[1]**2 + m.x[2]**2 + m.x[3]**3)

nl_fname = "test.nl"
with open(nl_fname, "w") as f:
    nlinfo = nl_writer.NLWriter().write(m, f, linear_presolve=True)

print(f"N. var = {len(nlinfo.variables)}")
print(f"N. con = {len(nlinfo.constraints)}")
print(f"N. eliminated = {len(nlinfo.eliminated_vars)}")

print(f"Eliminated: {nlinfo.eliminated_vars[0][0].name}")
print("Variables in nl file:")
for var in nlinfo.variables:
    print(f"  {var.name}")

# Not strictly necessary. This is just here to support the argument that
# the eliminated variable actually makes its way to the nl file.
subprocess.run(["ipopt", nl_fname])
```
This gives:
```console
N. var = 3
N. con = 1
N. eliminated = 1
Eliminated: x[1]
Variables in nl file:
  x[2]
  x[1]
  x[3]
```
Followed by the Ipopt log, which includes:
```console
Total number of variables............................:        3
                     variables with only lower bounds:        0
                variables with lower and upper bounds:        3
                     variables with only upper bounds:        0
Total number of equality constraints.................:        1
Total number of inequality constraints...............:        0
        inequality constraints with only lower bounds:        0
   inequality constraints with lower and upper bounds:        0
        inequality constraints with only upper bounds:        0
```
I would expect the model to have two variables and one constraint after presolve, where instead it (supposedly) has three variables and one constraint.

Pyomo version: 6.7.0.dev0 (current main)
Python version: 3.9.6
Solver (if applicable): Ipopt 3.14.11 (should not be relevant)

### Hypothesis

This appears to be happening because the call to `NLWriter._categorize_vars` for named expressions is adds `expr_info.linear` to the `linear_by_comp` map, approximately here:
https://github.com/Pyomo/pyomo/blob/91729ba44cdbaa888eedc7072251015d1d9e8da2/pyomo/repn/plugins/nl_writer.py#L1637-L1650
Then, when running `_categorize_vars` for the constraint in which the subexpression participates, we add `linear_by_comp[subexpr_id]` to `nonlinear_vars` for the constraint. The core issue seems to be that we don't check whether any of the linear variables have been eliminated (i.e., are present in `linear_by_comp`) when recording the linear portion of components. However, I am not sure why this does not cause a problem for constraints (perhaps because they don't appear in other components' expressions - but maybe the wrong variable could still show up since the elimination is not applied to `expo_info.linear`?). 

I did not understand the substitution code well enough to propose a fix. In particular, it was not clear to me where the substitution is happening. I think I see it for the nonlinear case, but not for the linear case.

jsiirola:
## Summary/Motivation:
#3055 pointed out that any variables eliminated by the linear presolve were not being removed from the linear portion of named expressions, allowing them to be emitted to the solver.  This resolves the issue and adds a test for the example from the issue.

- Ensure that named expressions with linear components are included in the `comp_by_linear_var` map, thereby ensuring that they are updated with the results of variable elimination.
- Add the model from #3055 as a test.

### Legal Acknowledgement

### DIFF
--- a/pyomo/repn/plugins/nl_writer.py
+++ b/pyomo/repn/plugins/nl_writer.py
@@ -1685,6 +1685,15 @@ class _NLWriter_impl(object):
         if not self.config.linear_presolve:
             return eliminated_cons, eliminated_vars
 
+        # We need to record all named expressions with linear components
+        # so that any eliminated variables are removed from them.
+        for expr, info, _ in self.subexpression_cache.values():
+            if not info.linear:
+                continue
+            expr_id = id(expr)
+            for _id in info.linear:
+                comp_by_linear_var[_id].append((expr_id, info))
+
         fixed_vars = [
             _id for _id, (lb, ub) in var_bounds.items() if lb == ub and lb is not None
         ]
--- a/pyomo/repn/tests/ampl/test_nlv2.py
+++ b/pyomo/repn/tests/ampl/test_nlv2.py
@@ -1480,7 +1480,6 @@ G0 1
         # Note: bounds on x[1] are:
         #   min(22/3, 82/17, 23/4, -39/-6) == 4.823529411764706
         #   max(2/3, 62/17, 3/4, -19/-6) == 3.6470588235294117
-        print(OUT.getvalue())
         self.assertEqual(
             *nl_diff(
                 """g3 1 1 0	# problem unknown
@@ -1743,7 +1742,7 @@ G0 3
         self.assertEqual(LOG.getvalue(), "")
 
         nl2 = OUT.getvalue()
-        print(nl2)
+
         self.assertEqual(
             *nl_diff(
                 """g3 1 1 0	# problem unknown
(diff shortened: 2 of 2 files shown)


Robbybp (approved):
Thanks for the quick fix. This resolves the issues I was seeing.

If my understanding is correct, replacement in the linear portion of constraints (and now, Expressions) happens in the `_linear_presolve` function, while replacement in the nonlinear portion happens in `_categorize_vars`. Can you confirm?

PR #3073, opened 2023-12-21
Various bug fixes in gdp.mbigm transformation

emma58:
## Summary/Motivation:

Fixes three bugs in the multiple bigm transformation:

1. There was an indentation error that was revealed by having (active) empty constraint containers on Disjuncts. Now we only deactivate constraints that exist, wheee.
2. Not *all* Suffixes are bad Suffixes: Ceasing to complain when we ignore Suffixes, but adding a note to the documentation that we will ignore BigM Suffixes, since that might not be intuitive. (We're also safe that this will never change because the keys will always be different for bigm and mbigm: mbigm needs to know the Disjunct the M is relative to as well.)
3. Disjuncts can be infeasible--it's okay: When the solve to calculate M comes back infeasible, this means the Disjunct is. We log that fact at the DEBUG level, and properly deactivate the Disjunct. I'm setting the M values to 0 for now for the sake of being explicit.

- Fixes to the above
- Tests!

### Legal Acknowledgement

4. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.


pyomo/gdp/plugins/multiple_bigm.py
```diff
@@ -636,8 +627,20 @@ def _calculate_missing_M_values(
...
                         scratch.obj.expr = constraint.body - constraint.lower
                         scratch.obj.sense = minimize
-                        results = self._config.solver.solve(other_disjunct)
+                        results = self._config.solver.solve(
+                            other_disjunct, load_solutions=False
+                        )
                         if (
+                            results.solver.termination_condition
+                            is TerminationCondition.infeasible
+                        ):
```
jsiirola, line 636: Are we sure that we want to check against "infeasible"?  For the "old" (current) solver interfaces, I am not sure if this is always a guarantee of infeasibility.  (In particular, I don't think that Ipopt will ever _guarantee_ infeasibility)

pyomo/gdp/plugins/multiple_bigm.py
```diff
@@ -647,14 +650,28 @@ def _calculate_missing_M_values(
...
                         scratch.obj.expr = constraint.body - constraint.upper
                         scratch.obj.sense = maximize
-                        results = self._config.solver.solve(other_disjunct)
+                        results = self._config.solver.solve(
+                            other_disjunct, load_solutions=False
+                        )
                         if (
+                            results.solver.termination_condition
+                            is TerminationCondition.infeasible
+                        ):
```
jsiirola, line 667: Same question on `infeasible` here...

pyomo/gdp/plugins/multiple_bigm.py
```diff
@@ -201,7 +200,6 @@ class MultipleBigMTransformation(GDP_to_MIP_Transformation, _BigM_MixIn):
 
     def __init__(self):
         super().__init__(logger)
-        self.handlers[Suffix] = self._warn_for_active_suffix
```
jsiirola, line 204: Are we sure we want to silently ignore Suffixes?

jsiirola (commented):
A couple general questions

pyomo/gdp/plugins/multiple_bigm.py
emma58: I changed this to ignore `LocalVars`, warn about ignoring `BigM`, and yell for everything else.

pyomo/gdp/plugins/multiple_bigm.py
emma58: I changed this so that we maintain a list of solvers we trust that 'infeasible' means 'proven infeasible' and we do this for those. For others (looking at you, ipopt), we just revert to the old behavior and scream about an unsuccessful solve.

emma58:
OK, @jsiirola, this is ready for another look.

jsiirola (approved):

PR #3075, opened 2024-01-05
Convert implicit sets created by `IndexedComponent`s to "anonymous" sets 

### LINKED ISSUE #45: del_component issue when deleting constraints

### LINKED ISSUE #2068: [PEP] add "implicit" sets to the Block as "anonymous" components
## Summary

Historically, Set objects implicitly created by a component were added to the same block as the component with hard-coded names derived from the component name.  This PEP proposes we remove that behavior and allow for implicit Sets to be attached to the Block "anonymously".

### Discussion

It is important that Sets created implicitly during the process of constructing a Component (and attaching it to a Block) are also attached to the same Block.  This ensures that the Sets are properly constructed (especially in the case of AbstractModels), and ensures that `model.clone()` correctly duplicates the Sets.  However, the use of a hard-coded name causes problems when deleting and re-adding components (as the implicit sets are not removed from the Block - see #45).  Further, because this is done when the Component is added to the Block, we cannot disable access to attributes like `Var.domain` for unconstructed Components (because Block._add_implicit_sets() looks for '`domain`' to attach it to the Block).  Finally this convention obscures the model that the user wrote: both by filling it with numerous "auto" Sets (e.g., `x_index`, `y_index_0`, `z_domain`), and by reporting the name of the "auto" set as the Component index (instead of something more informative like the set expression).

This PEP proposes that we still "add" implicit Sets to the Block, but not using an explicit name.  We would add the component using a new `Block.add_anonymous_component()` method.  This would set the Component's `parent_block()` to the Block (so that `block.clone()` works correctly) and would categorize it in the Block's component data structures according to the `ctype` (so, e.g., it would be returned by `component_objects()`), however it would not add the component to the Block's `__dict__`.

We would also transfer the responsibility for constructing implicit components from the Block to the Component that created the implicit components.  Deferring it to the "originator's" construction time would allow us to deactivate interfaces like `domain` on abstract components.

jsiirola:
## Summary/Motivation:
Pyomo creates Set objects on the fly when constructing IndexedComponents indexed by anything other than a single Pyomo Set object.  These objects get added to the Block using a fixed auto-generated names.  This causes problems when deleting an indexed component and then trying to create it again, because the sets implicitly created by Pyomo are still on the Block and lead to errors about name conflicts.

This PR resolves this by adding support for "anonymous" (unnamed) Set objects.  With this PR, he Set objects implicitly created during component declaration (including `index_set` and `domain`) will not be explicitly attached to the Block.  However, they will have their `parent_block()` point tot he "owning" block, thereby ensuring that the anonymous sets obey the correct scoping rules when deepcopied / cloned.  While this may seem like a departure to the convention that all modeling components are explicitly attached to Blocks, there is precedent in the case of Sets for global Set objects (e.g., `Reals`, `Integers`, etc).

This touches a lot of files due to changes in test baselines.  If you want to see the diffs for everything except the test changes, see https://github.com/Pyomo/pyomo/compare/d5b53d5..0325346

- rework how component Set-like arguments are processed and anonymous sets are tracked
- update string representation for Set objects with a `parent_block()` but no `_name` (use the Set data as the name)
- remove `_implicit_subsets` attribute (previously used for the implicit arguments to SetProduct) and replaced it with the general purpose `_anonymous_sets` (used by all anonymous sets)
- some standardization in `Component.construct()` implementations
- resolved some issues when using slices when set flattening is disabled
- remove the dependence on `cplex_direct` in Benders tests

### Legal Acknowledgement


pyomo/core/base/constraint.py
```diff
@@ -717,8 +717,6 @@ class Constraint(ActiveIndexedComponent):
             A dictionary from the index set to component data objects
         _index
             The set of valid indices
-        _implicit_subsets
```
mrmundt, line 720: Do you need to add something to the doc class about `_anonymous_sets`?

pyomo/dae/flatten.py
```diff
@@ -200,8 +200,28 @@ def slice_component_along_sets(component, sets, context_slice=None, normalize=No
...
                     # Note that c_slice is not necessarily a slice.
                     # We enter this loop even if no sets need slicing.
-                    temp_slice = c_slice.duplicate()
-                    next(iter(temp_slice))
+                    try:
+                        next(iter(c_slice.duplicate()))
+                    except IndexError:
+                        if normalize_index.flatten:
+                            raise
+                        # There is an edge case where when we are not
```
mrmundt, line 208: I would recommend adding a test for this edge case, if possible, considering that we explicitly check for it.

pyomo/core/base/constraint.py
jsiirola: The catch is that `_anonymous_sets` is not *always* an attribute,.  I made the decision to try and save some time/memory and only add the attribute if it was needed.  I wasn't sure if I wanted to document an attribute that was not necessarily always there.

That may / may not be a good idea, though.  I keep wanting to try and figure out how to slotize the container classes, and this potentially further complicates that.  Maybe we should move it back to being a mandatory attribute.

pyomo/core/base/set.py
```diff
@@ -2934,14 +2960,12 @@ def __init__(self, *args, **kwds):
             pass
 
     def __str__(self):
-        if self.parent_block() is not None:
+        # Named, components should return their name e.g., Reals
```
blnicho, line 2963: Grammar: unnecessary comma after "Named"

pyomo/core/base/set.py
```diff
@@ -3303,7 +3344,7 @@ def __len__(self):
         )
 
     def __str__(self):
-        if self.parent_block() is not None:
+        if self._name is not None:
             return self.name
```
blnicho, line 3348: This is checking for `self._name` and returning `self.name`. Is that correct? What's the difference between `_name` and `name`?

pyomo/core/base/set.py
```diff
@@ -4202,7 +4244,7 @@ def domain(self):
         return Any
 
     def __str__(self):
-        if self.parent_block() is not None:
+        if self._name is not None:
             return self.name
```
blnicho, line 4248: Same question about `_name` and `name`.

pyomo/core/base/set.py
```diff
@@ -4247,7 +4290,7 @@ def domain(self):
         return EmptySet
 
     def __str__(self):
-        if self.parent_block() is not None:
+        if self._name is not None:
             return self.name
```
blnicho, line 4294: Same question about `_name` and `name`.

examples/pyomobook/scripts-ch/warehouse_script.txt
```diff
@@ -1,36 +1,10 @@
...
-    x_index_1 : Size=1, Index=None, Ordered=Insertion
-        Key  : Dimen : Domain : Size : Members
-        None :     1 :    Any :    4 : {'NYC', 'LA', 'Chicago', 'Houston'}
-    y_index : Size=1, Index=None, Ordered=Insertion
-        Key  : Dimen : Domain : Size : Members
-        None :     1 :    Any :    3 : {'Harlingen', 'Memphis', 'Ashland'}
-
 2 Var Declarations
-    x : Size=12, Index=x_index
+    x : Size=12, Index={Harlingen, Memphis, Ashland}*{NYC, LA, Chicago, Houston}
```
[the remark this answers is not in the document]
jsiirola: I've been wanting to to that for a long time (and not just here, but in expressions, too), but that would require a significant redesign of the way we convert things to strings.  I could hack something in here, but wonder if we should do it as a separate PR.

(The challenge is this is really part of converting expressions to strings, and strings are built from the leaf nodes up to the root node.  So, you may not know that the result is "too long" until you get up the expression tree , but then need to go back down the tree and truncate things closer to the leaves)

pyomo/core/base/constraint.py
```diff
@@ -772,6 +770,10 @@ def construct(self, data=None):
         if is_debug_set(logger):
             logger.debug("Constructing constraint %s" % (self.name))
 
+        if self._anonymous_sets is not None:
```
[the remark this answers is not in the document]
jsiirola: In this case, no.  I partially misspoke in my reply to Miranda (I missed which part of the documentation we were talking about).  `_anonymous_sets` will exist on all `IndexedComponent` classes - I ended up taking the conditional definition of the attribute out (although it can be None [and not a `ComponentSet`]).

pyomo/core/base/set.py
jsiirola: This is intentional: `self._name` is `None` if it is an anonymous set.  If it is not None, then `_name` contains the local (block-scope) name of the Set.  `.name` is a property that returns the fully-qualified name of the Set (including all parent block names).

pyomo/core/base/expression.py
```diff
@@ -394,6 +394,10 @@ def construct(self, data=None):
                 % (self.name, str(data))
             )
 
+        if self._anonymous_sets is not None:
+            for _set in self._anonymous_sets:
+                _set.construct()
```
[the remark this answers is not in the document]
jsiirola: This was a concern for me, too.  At the time, I was worried about making that change (that is, requiring all components to also call construct() on their base class - in part because a lot of what I would like to move to the base class is a "prefilter" (e.g., the oft-repeated debug logging, the logic to skip-if-constructed, and now this).  But the more I think about it, the more I am inclined to propose that we implement `construct()` in the base class and have it call a `_construct_impl()` that is overridden on the derived classes.  This would reduce a lot of repeated code and shouldn't impact performance significantly.

This would be a somewhat large change.  Would you object to doing it as a separate PR?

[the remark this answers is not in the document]
blnicho: No objections to making that a separate PR. I'll open an issue for it.

blnicho (approved):

pyomo/core/base/expression.py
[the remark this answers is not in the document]
blnicho: Tracking this in issue #3136

pyomo/core/tests/unit/test_block.py
```diff
@@ -2668,9 +2665,9 @@ def test_pprint(self):
     c : Size=1, Index=None, Active=True
         0 Declarations: 
 
-6 Declarations: a1_IDX a3_IDX c a_index a b
+5 Declarations: a1_IDX a3_IDX c a b
 """
-        print(buf.getvalue())
+        self.maxDiff = None
```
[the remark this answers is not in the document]
blnicho: I never knew this was an option. I think adding it to the base class would make a lot of sense. This seems like behavior we would always want when tests fail.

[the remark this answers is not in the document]
mrmundt: I thoroughly agree.

pyomo/dae/flatten.py
mrmundt: Poke again on this. Feel free to say "No"

pyomo/mpec/complementarity.py
```diff
@@ -357,13 +357,18 @@ def construct(self, data=None):
...
         if self._constructed:
             return
-        timer = ConstructionTimer(self)
         self._constructed = True
 
+        timer = ConstructionTimer(self)
+        if is_debug_set(logger):
+            logger.debug("Constructing complementarity list %s", self.name)
+
+        if self._anonymous_sets is not None:
```
mrmundt, line 368: After seeing these three lines of code _everywhere_, I wondering if it wouldn't make sense to have some sort of helper function instead to call? If this ever changes, you've got dozens of places to change it.

blnicho: I had the same comment and opened #3136 to track moving this to the base class.

pyomo/dae/flatten.py
jsiirola: I think I am going to go with "No".  I am pretty sure that the case I was describing is already being covered by tests - the comment was mostly a note to our future selves why this was implemented the way it was.

mrmundt (approved):

emma58 (approved):
This is really exciting! :)

PR #3082, opened 2024-01-10
GDP => MINLP Transformation

michaelbynum:
## Summary/Motivation:
This PR adds a GDP transformation to convert a GDP to a MINLP by converting f(x) <= 0 on a disjunct to f(x)*y <= 0.

### Legal Acknowledgement


bernalde (commented):
@michaelbynum you are reading our minds! We were just discussing about this transformation with @emma58 @jsiirola and @ZedongPeng 
The code looks great, but since we have the chance of creating a new reformulation, should we use it to give it a good name? GDP to MINLP is not specific enough (as BM and Hull might do that as well). How about binary multiplication? Short and descriptive!

michaelbynum:
Great suggestion! I like it!

pyomo/gdp/plugins/binary_multiplication.py
```diff
@@ -0,0 +1,175 @@
+from .gdp_to_mip_transformation import GDP_to_MIP_Transformation
+from pyomo.common.config import ConfigDict, ConfigValue
+from pyomo.core.base import TransformationFactory
+from pyomo.core.util import target_list
+from pyomo.gdp import Disjunction
+from pyomo.gdp.plugins.gdp_to_mip_transformation import GDP_to_MIP_Transformation
+from pyomo.core.util import target_list
```
emma58, line 7: This looks like a duplicate import

pyomo/gdp/plugins/binary_multiplication.py
```diff
@@ -0,0 +1,175 @@
+from .gdp_to_mip_transformation import GDP_to_MIP_Transformation
+from pyomo.common.config import ConfigDict, ConfigValue
+from pyomo.core.base import TransformationFactory
+from pyomo.core.util import target_list
+from pyomo.gdp import Disjunction
+from pyomo.gdp.plugins.gdp_to_mip_transformation import GDP_to_MIP_Transformation
```
emma58, line 6: This looks like a duplicate import

pyomo/gdp/plugins/binary_multiplication.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+logger = logging.getLogger('pyomo.gdp.binary_multiplication')
+
+
+@TransformationFactory.register(
+    'gdp.binary_multiplication',
+    doc="Reformulate the GDP as an MINLP by multiplying f(x) <= 0 by y to get f(x) * y <= 0.",
+)
+class GDPToMINLPTransformation(GDP_to_MIP_Transformation):
```
emma58, line 19: Maybe change the class name to something having to do with binary multiplication as well?

pyomo/gdp/plugins/binary_multiplication.py
```diff
@@ -0,0 +1,175 @@
...
+from weakref import ref as weakref_ref
+import logging
+
+
+logger = logging.getLogger('pyomo.gdp.binary_multiplication')
+
+
+@TransformationFactory.register(
+    'gdp.binary_multiplication',
+    doc="Reformulate the GDP as an MINLP by multiplying f(x) <= 0 by y to get f(x) * y <= 0.",
```
emma58, line 17: Maybe specify in this doc string that `y` is the binary corresponding to the Boolean indicator var of the Disjunct containing `f(x) <= 0`.

pyomo/gdp/plugins/binary_multiplication.py
```diff
@@ -0,0 +1,175 @@
...
+from pyomo.core.base import TransformationFactory
+from pyomo.core.util import target_list
+from pyomo.gdp import Disjunction
+from pyomo.gdp.plugins.gdp_to_mip_transformation import GDP_to_MIP_Transformation
+from pyomo.core.util import target_list
+from weakref import ref as weakref_ref
+import logging
+
+
+logger = logging.getLogger('pyomo.gdp.binary_multiplication')
```
emma58, line 12: I'm starting to like the convention of setting this as `logging.getLogger(__name__)`.

pyomo/gdp/plugins/binary_multiplication.py
```diff
@@ -0,0 +1,175 @@
...
+    CONFIG = ConfigDict("gdp.binary_multiplication")
+    CONFIG.declare(
+        'targets',
+        ConfigValue(
+            default=None,
+            domain=target_list,
+            description="target or list of targets that will be relaxed",
+            doc="""
+
+        This specifies the list of components to relax. If None (default), the
```
emma58, line 29: This is my typo propagating, probably, but 'transform' would be a clearer verb than 'relax'

pyomo/gdp/plugins/binary_multiplication.py
```diff
@@ -0,0 +1,175 @@
...
+
+        # add or (or xor) constraint
+        or_expr = 0
+        for disjunct in obj.disjuncts:
+            or_expr += disjunct.binary_indicator_var
+            self._transform_disjunct(disjunct, transBlock)
+
+        rhs = 1 if parent_disjunct is None else parent_disjunct.binary_indicator_var
+        if obj.xor:
+            xorConstraint[index] = or_expr == rhs
```
emma58, line 86: Unfortunately (and this is my fault too), this isn't actually correct... It assumes that nested indicator variables are local, which is not necessarily true. The safest thing for now is to always set the rhs to 1 and to put the exactly-one constraint on the parent block (and transform from leaf to root, which I think you're already doing).

pyomo/gdp/tests/test_binary_multiplication.py
```diff
@@ -0,0 +1,316 @@
...
+        ct.check_improperly_deactivated_disjuncts(self, 'binary_multiplication')
+
+    def test_do_not_transform_userDeactivated_IndexedDisjunction(self):
+        ct.check_do_not_transform_userDeactivated_indexedDisjunction(
+            self, 'binary_multiplication'
+        )
+
+    # helper method to check the M values in all of the transformed
+    # constraints (m, M) is the tuple for M.  This also relies on the
+    # disjuncts being transformed in the same order every time.
```
emma58, line 173: This looks like a copy-paste thing, but not relevant to this transformation

pyomo/gdp/tests/test_binary_multiplication.py
```diff
@@ -0,0 +1,316 @@
...
+        repn = generate_standard_repn(c[0].body)
+        self.assertIsNone(repn.nonlinear_expr)
+        self.assertEqual(len(repn.quadratic_coefs), 1)
+        self.assertEqual(len(repn.linear_coefs), 1)
+        ind_var = model.d[0].indicator_var
+        ct.check_quadratic_coef(self, repn, model.a, ind_var, 1)
+        ct.check_linear_coef(self, repn, ind_var, -model.d[0].c.lower)
+        self.assertEqual(repn.constant, 0)
+        self.assertEqual(c[0].lower, 0)
+        self.assertIsNone(c[0].upper)
```
emma58, line 195: You don't have to change this if you don't have time, but I've been trying to move to using `assertExpressionsEqual` instead of this mess with the standard repn because it makes for much more readable tests.

emma58 (commented):
Thanks so much for doing this, @michaelbynum! A few comments (most of which are my mistakes propagating, sorry...), but this generally looks good.

pyomo/gdp/plugins/binary_multiplication.py
michaelbynum: Your point with putting the exactly-one constraint on the parent block (instead of the root disjunct's parent block) is that the exactly-one constraint needs to be transformed again if it is for a nested disjunction?

michaelbynum: Maybe I'm misunderstanding what the "root disjunct" is...

michaelbynum: Is putting the exactly-one constraint on the parent disjunct equivalent?

pyomo/gdp/tests/test_binary_multiplication.py
michaelbynum: It was easy to change and is indeed far more readable.

pyomo/gdp/plugins/binary_multiplication.py
emma58: Correct. It should be on the parent disjunct so that it does get transformed again.

pyomo/gdp/plugins/binary_multiplication.py
```diff
@@ -0,0 +1,173 @@
...
+from weakref import ref as weakref_ref
+import logging
+
+
+logger = logging.getLogger(__name__)
+
+
+@TransformationFactory.register(
+    'gdp.binary_multiplication',
+    doc="Reformulate the GDP as an MINLP by multiplying f(x) <= 0 by y to get f(x) * y <= 0 where y is the binary corresponding to the Boolean indicator var of the Disjunct containing f(x) <= 0.",
```
jsiirola, line 15: Can we break this into multiple strings to keep within Black's preferred 88-character lines?

pyomo/gdp/plugins/binary_multiplication.py
```diff
@@ -0,0 +1,173 @@
...
+        # local names, we need to make sure that the first index for
+        # transformedConstraints is guaranteed to be unique. We just grab the
+        # current length of the list here since that will be monotonically
+        # increasing and hence unique. We'll append it to the
+        # slightly-more-human-readable constraint name for something familiar
+        # but unique. (Note that we really could do this outside of the loop
+        # over the constraint indices, but I don't think it matters a lot.)
+        unique = len(newConstraint)
+        name = c.local_name + "_%s" % unique
+
```
jsiirola, line 147: Can we make this a little simpler by starting with:
```python
constraintMap['transformedConstraints'][c] = transformed = []
```
Then everything else can unconditionally append to `transformed`?

jsiirola (approved):
To really minor things, but nothing to prevent merging

pyomo/gdp/plugins/binary_multiplication.py
```diff
@@ -0,0 +1,173 @@
+from .gdp_to_mip_transformation import GDP_to_MIP_Transformation
```
mrmundt, line 1: I know it's in contrib, but can you please add the copyright statement at the top?

pyomo/gdp/plugins/binary_multiplication.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+logger = logging.getLogger(__name__)
+
+
+@TransformationFactory.register(
+    'gdp.binary_multiplication',
+    doc="Reformulate the GDP as an MINLP by multiplying f(x) <= 0 by y to get f(x) * y <= 0 where y is the binary corresponding to the Boolean indicator var of the Disjunct containing f(x) <= 0.",
+)
+class GDPBinaryMultiplicationTransformation(GDP_to_MIP_Transformation):
```
mrmundt, line 17: Keeping in mind that I know very little about stuff in the `TransformationFactory`, is it standard for all methods to be private?

mrmundt (changes requested):

pyomo/gdp/plugins/binary_multiplication.py
emma58: Yeah, the only things that are going to get called in the outside world are `apply_to` and `create_using` and sometimes some methods that provide APIs for mapping between original and transformed components and vice versa. So at least with our current designs it's quite typical.

mrmundt: _TIL_.

emma58:
OK, what I just pushed should address @jsiirola's and @mrmundt's comments. If tests pass, we should merge!

mrmundt (approved):

PR #3090, opened 2024-01-14
fix an error in the documentation for LinearExpression

DLWoodruff:
## Summary/Motivation:
There was a minor error in documentation.

- The documentation said "LinearModel" but now, correctly, says "LinearExpression"
-

### Legal Acknowledgement


jsiirola (dismissed):
This fix certainly makes the documentation less wrong, but in all likelihood, the entire section should probably be removed:
- We generate `LinearExpressions` automatically (when we can) through the normal operator oveloading, and (I believe) at nearly the same efficiency as the if they were specified explicitly
- The warning that is being corrected is no longer true: the LinearExpression arguments are single list of MonomialTermExpressions and native_numeric_types: the lists in this example are torn apart and converted to MonomialTermExpressions.
- the "3-argument" interface for creating LinearExpressions is no longer the preferred syntax

DLWoodruff:
Based on the comment made by @jsiirola I edited to be even a little less wrong; however, maybe a few words (even cryptic words) about when it is no longer needed would be a great improvement. Deleting the section might not be a great idea because the components appears in some models and folks might wonder what it does and why the don't really need it anymore.

jsiirola (approved):

PR #3109, opened 2024-01-26
Remove unnecessary `__future__` imports

mrmundt:
## Summary/Motivation:
There were some leftover imports that are unnecessary since we no longer support Python2.

- Remove `from __future__ import *`
- Update comments, as necessary

### Legal Acknowledgement


pyomo/core/expr/numvalue.py
```diff
@@ -264,15 +264,6 @@ def polynomial_degree(obj):
...
 #
-# Note:
-#   For now, all constants are coerced to floats.  This avoids integer
-#   division in Python 2.x.  (At least some of the time.)
-#
-#   When we eliminate support for Python 2.x, we will not need this
-#   coercion.  The main difference in the following code is that we will
-#   need to index KnownConstants by both the class type and value, since
-#   INT, FLOAT and LONG values sometimes hash the same.
-#
```
blnicho, line 275: I think we should keep this comment unless we can confirm with @jsiirola that it isn't needed. Line 305 below still seems to be casting things to floats and this comment claims that this isn't needed.

blnicho (changes requested):
There is one comment I think we should leave

pyomo/core/expr/numvalue.py
mrmundt: I will say, I tried to remove the casting to float and it caused failures. So I can undo the comment, but I will also note that the comment is misleading and should be edited sooner rather than later.

blnicho (approved):

examples/dae/stochpdegas_automatic.py
```diff
@@ -1,7 +1,7 @@
 # stochastic pde model for natural gas network
 # victor m. zavala / 2013
 
-# from __future__ import division
+#
```
jsiirola, line 4: Can we delete this?  It looks like a stray "#"

examples/performance/dae/stochpdegas1_automatic.py
```diff
@@ -1,7 +1,7 @@
 # stochastic pde model for natural gas network
 # victor m. zavala / 2013
 
-# from __future__ import division
+#
```
jsiirola, line 4: Can we delete this?

jsiirola (changes requested):

PR #3113, opened 2024-01-30
avoid duplicate entries in config _decl_order

michaelbynum:
## Summary/Motivation:
This fixes a bug that causes `config.display()` to print multiple times.

```

In [1]: from pyomo.common.config import ConfigDict, ConfigValue

In [2]: class Conf(ConfigDict):
   ...:     def __init__(
   ...:         self,
   ...:         description=None,
   ...:         doc=None,
   ...:         implicit=False,
   ...:         implicit_domain=None,
   ...:         visibility=0,
   ...:     ):
   ...:         super().__init__(
   ...:             description=description,
   ...:             doc=doc,
   ...:             implicit=implicit,
   ...:             implicit_domain=implicit_domain,
   ...:             visibility=visibility,
   ...:         )
   ...:         self.foo = self.declare('foo', ConfigValue(default=1))
   ...:

In [3]: c1 = Conf()

In [4]: c1.display()
foo: 1

In [5]: c2 = c1()

In [6]: c2.display()
foo: 1
foo: 1
```

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/config.py
+++ b/pyomo/common/config.py
@@ -1692,9 +1692,12 @@ class ConfigBase(object):
                 if preserve_implicit or k in self._declared:
                     v = self._data[k]
                     ans._data[k] = _tmp = v(preserve_implicit=preserve_implicit)
-                    ans._decl_order.append(k)
-                    if k in self._declared:
-                        ans._declared.add(k)
+                    if k not in ans._declared:
+                        # the above if statement is needed just in case the options
+                        # were declared in the constructor above
+                        ans._decl_order.append(k)
+                        if k in self._declared:
+                            ans._declared.add(k)
                     _tmp._parent = ans
                     _tmp._name = v._name
 


pyomo/common/config.py
```diff
@@ -1692,9 +1692,12 @@ def __call__(
                 if preserve_implicit or k in self._declared:
                     v = self._data[k]
                     ans._data[k] = _tmp = v(preserve_implicit=preserve_implicit)
-                    ans._decl_order.append(k)
-                    if k in self._declared:
-                        ans._declared.add(k)
+                    if k not in ans._declared:
```
mrmundt, line 1695: Could you add a very small test for this? Example:
```
with unittest.mock.patch('sys.stdout', new = StringIO()) as captured: 
    expected_output = '...whatever...'
    config.display() 
    self.assertEqual(captured.getvalue(), expected_output) 
```

mrmundt (changes requested):

jsiirola:
I think this is fine to resolve the observed behavior, however, I think the correct solution is to completely remove the `_decl_order` list: it was only necessary for Python < 3.7 (i.e., before the `dict` iteration order was deterministic).

PR #3116, opened 2024-01-31
Prevent duplicate entries in ConfigDict declaration order

jsiirola:
## Summary/Motivation:
#3113 pointed out that ConfigDict instances that declared attributes in the `__init__` could end up with duplicate entries in the `_decl_order` list, resulting in, e.g., duplicate entries in the `display()` output.  `_decl_order` was only added to resolve non-determinism in `dict` for Python before 3.7.  As we no longer support those versions of Python, the entire `_decl_order` logic can be removed (and implicitly resolves the problem reported in #3113).

- Remove `_decl_order` from `ConfigDict`
- Add a test for the `display()` behavior observed in #3113.

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/tests/test_config.py
+++ b/pyomo/common/tests/test_config.py
@@ -3027,6 +3027,46 @@ option_2: int, default=5
             self.assertEqual(add_docstring_list("", ExampleClass.CONFIG), ref)
         self.assertIn('add_docstring_list is deprecated', LOG.getvalue())
 
+    def test_declaration_in_init(self):
+        class CustomConfig(ConfigDict):
+            def __init__(
+                    self,
+                    description=None,
+                    doc=None,
+                    implicit=False,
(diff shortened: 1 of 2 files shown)


pyomo/common/tests/test_config.py
```diff
@@ -3027,6 +3027,40 @@ def fcn(self):
             self.assertEqual(add_docstring_list("", ExampleClass.CONFIG), ref)
         self.assertIn('add_docstring_list is deprecated', LOG.getvalue())
 
+    def test_declaration_in_init(self):
```
mrmundt, line 3030: YAAAAY TEST!!!

michaelbynum (approved):
Do we still need `_declared`? Is that used for implicit things?

jsiirola:
Yes.  `_declared` is how we track declared vs implicit keys.

PR #3135, opened 2024-02-13
NLv2: only raise exception for empty models in the legacy API

### LINKED ISSUE #3131: NL Writer Errors when presolve removes all variables
## Summary

It would be extremely convenient if we could just return the solution somehow.

```
>>> import pyomo.environ as pe
>>> from pyomo.repn.plugins.nl_writer import NLWriter
>>> import io
>>> m = pe.ConcreteModel()
>>> m.x = pe.Var()
>>> m.y = pe.Var()
>>> m.obj = pe.Objective(expr=m.y)
>>> m.c1 = pe.Constraint(expr=m.y == m.x + 2)
>>> m.c2 = pe.Constraint(expr=m.y == -m.x + 1)
>>> writer = NLWriter()
>>> out = io.StringIO()
>>> writer.write(m, out)
  File "/.../pyomo/repn/plugins/nl_writer.py", line 397, in write
    return impl.write(model)
           ^^^^^^^^^^^^^^^^^
  File "/.../pyomo/repn/plugins/nl_writer.py", line 865, in write
    raise ValueError(
ValueError: No variables appear in the Pyomo model constraints or objective. This is not supported by the NL file interface
```

jsiirola:
## Summary/Motivation:
This updates the NLv2 writer to only raise an exception for empty models (no variables in either objectives or constraints) when it is called using the legacy "call" interface.  The new recommended interface will not raise an exception and instead rely on the caller (the solver interface) to inspect the returned list of variables and do the correct thing when that list is empty.

- Move the exception for an empty model from the core writer method and into the logacy `__call__` interface.

### Legal Acknowledgement

### DIFF
--- a/pyomo/repn/plugins/nl_writer.py
+++ b/pyomo/repn/plugins/nl_writer.py
@@ -346,6 +346,17 @@ class NLWriter(object):
             row_fname
         ) as ROWFILE, _open(col_fname) as COLFILE:
             info = self.write(model, FILE, ROWFILE, COLFILE, config=config)
+        if not info.variables:
+            # This exception is included for compatibility with the
+            # original NL writer v1.
+            os.remove(filename)
+            os.remove(row_filename)
+            os.remove(col_filename)
+            raise ValueError(
+                "No variables appear in the Pyomo model constraints or"
+                " objective. This is not supported by the NL file interface"
+            )
+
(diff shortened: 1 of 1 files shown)


michaelbynum (approved):
Thanks, @jsiirola!

Issue #3130, opened 2024-02-10
Setting opt.config.load_solution = False still loads the model when running opt.solve


lukasbiton:
## Summary

When I feed pyomo an unfeasible model, I get the following RuntimeError:

`A feasible solution was not found, so no solution can be loaded.Please set opt.config.load_solution=False and check results.termination_condition and results.best_feasible_objective before loading a solution.`

Following the instructions and setting `opt.config.load_solution = False` does not stop the RuntimeError from being raised, so I still can't check the termination condition. I would expect that when choosing this setting, no error would be raised and I would be able to look at my results object.

Looking through the stack trace, the final error happens here: https://github.com/Pyomo/pyomo/blob/main/pyomo/contrib/appsi/solvers/highs.py#L671
This block of code seems to be indented incorrectly, resulting in unexpected behavior given the provided error message. More specifically, the else statement that raises the RuntimeError is aligned with `if has_feasible_solution:` on line 661 and not with `if config.load_solution:` on line 660. Aligning the indent to line 660 induces the expected behavior given the error message provided.

### Steps to reproduce the issue

I am using pyomo with the HiGHS solver. I installed the HiGHS solver following the instructions on their website: https://ergo-code.github.io/HiGHS/dev/interfaces/cpp/.
I then locally installed highspy to make sure pyomo could find the solver.

In order to reproduce my output, you need to install pyomo and highspy. Then, run the `test.py` script presented below with the following command:

```console
$ pyhton -m test
```

Here is my example script: (stolen with a small added constraint from https://pyomo.readthedocs.io/en/stable/pyomo_overview/simple_examples.html)

```
# test.py

import pyomo.environ as pyo

model = pyo.ConcreteModel()

model.x = pyo.Var([1,2], domain = pyo.NonNegativeReals)

model.OBJ = pyo.Objective(expr = 2 * model.x[1] + 3 * model.x[2])

model.Constraint1 = pyo.Constraint(expr = 3 * model.x[1] + 4 * model.x[2] >= 1)
# New constraint makes the problem unfeasible
model.Constraint2 = pyo.Constraint(expr = model.x[1] + model.x[2] == -10)

opt = pyo.SolverFactory("appsi_highs")
opt.config.load_solution = False
results = opt.solve(model)
```

### Error Message

Here is the entire stack trace:

```console
$ python -m test
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
[...]

PR #3141, opened 2024-02-16
error msg more explicit wrt different interfaces

lukasbiton:
## Summary/Motivation:

As demonstrated in [Issue 3130](https://github.com/Pyomo/pyomo/issues/3130), the error message raised when a problem is unfeasible doesn't take into account the two different interfaces that can be used to create models: via contrib.appsi or via environ.SolverFactory.

For each of the appsi solvers, the error message raised now details the syntax more explicitly. The original issue was written using the HiGHS solver, but the following error message was added for Cbc, Cplex, Gurobi, Highs, Ipopt, and Wntr.

'A feasible solution was not found, so no solution can be loaded. '
'If using the appsi.solvers.Highs interface, you can '
'set opt.config.load_solution=False. If using the environ.SolverFactory '
'interface, you can set opt.solve(model, load_solutions = False). '
'Then you can check results.termination_condition and '
'results.best_feasible_objective before loading a solution.'

### Legal Acknowledgement


michaelbynum (approved):
Thanks for this! Looks great!

blnicho (approved):

PR #3140, opened 2024-02-16
Extend Path and Type Checking Validators of `common.config`

shermanjasonaf:
## Summary/Motivation:
[This comment](https://github.com/Pyomo/pyomo/pull/3126#discussion_r1488706754) of #3126 suggests incorporating the use of [`os.fsdecode`](https://docs.python.org/3/library/os.html#os.fsdecode) into the `common.config.Path` validator to provide support for generic [path-like](https://docs.python.org/3/glossary.html#term-path-like-object) objects.

[This comment](https://github.com/Pyomo/pyomo/pull/3126#discussion_r1488716106) of #3126 suggests adding a new domain validator for type checking.

## Changes proposed in this PR:
- Extend range of support of `common.config.Path` to generic path-like objects.
- Add `IsInstance` domain validator for arbitrary type checking.

### Legal Acknowledgement

By contributing to this software project, I have read the [contribution guide](https://pyomo.readthedocs.io/en/stable/contribution_guide.html) and agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #3126, opened 2024-02-07
Update PyROS Solver Argument Resolution and Validation Routines

shermanjasonaf:
## Changes proposed in this PR:
- Use domain interfaces of `common.config` more extensively for argument-wise validation.
- Provide more informative warning/error messages in event of invalid or ambiguous inputs.
- Add more rigorous validation checks and appropriate tests.
- Use interfaces of #2754 in lieu of that of #2707 to generate the docstring of `PyROS.solve()`. Remove interface of #2707.

### Legal Acknowledgement

By contributing to this software project, I have read the [contribution guide](https://pyomo.readthedocs.io/en/stable/contribution_guide.html) and agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #3144, opened 2024-02-19
Update `Path`, `PathList`, and `IsInstance` Domain Validators

shermanjasonaf:
## Summary/Motivation:

This PR is a follow-up to #3140, itself motivated by discussion in #3126.

- Add `IsInstance` to the `config` library reference in the online documentation.
- Make the usage of fully qualified valid type names in `IsInstance.domain_name()` and any error messages optional (as sometimes, the fully qualified names can be very long).
- Implement `Path.domain_name()`.
- Make documentation of `Path` more consistent with updates of #3140.
- Adapt behavior and documentation of `PathList` to updated implementation of `Path`.

### Legal Acknowledgement


mrmundt (approved):

pyomo/common/config.py
```diff
@@ -562,10 +589,15 @@ class PathList(Path):
...
-            return [super(PathList, self).__call__(data)]
+        try:
+            pathlist = [super(PathList, self).__call__(data)]
+        except TypeError as err:
+            is_not_path_like = "expected str, bytes or os.PathLike" in str(err)
+            if is_not_path_like and hasattr(data, "__iter__"):
+                pathlist = [super(PathList, self).__call__(i) for i in data]
+            else:
+                raise
+        return pathlist
```
jsiirola, line 600: What is the motivation for this change (especially because I don't see [an obvious] new test that would pass here and fail before)?  I am concerned about putting in a try-except in the validator that is expected to fail in normal cases.

shermanjasonaf: In lieu of the changes, I am thinking about changing `PathList.__call__` to something like:

```python
def __call__(self, data):
    is_path_like = isinstance(data, (str, bytes)) or hasattr(data, "__fspath__")
    if is_path_like:
        return [super(PathList, self).__call__(data)]
    else:
        return [super(PathList, self).__call__(i) for i in data]
```

Thoughts?

jsiirola: (we were both concurrently  thinking along the exact same lines)

jsiirola (approved):

PR #3150, opened 2024-02-19
Add support for tuples in `ComponentMap`; add `DefaultComponentMap`

jsiirola:
## Summary/Motivation:
`ComponentMap` / `ComponentSet` historically used the `id()` of the key for all entries.  This prevented the reliable use of tuples as keys in `ComponentMap` / items in `ComponentSet`, as two tuples that contained the same data would potentially have separate ids.  This PR changes the hashing scheme to:
- only replace unhashable elements with their `id()`
- recursively process tuples using the same logic

This also adds a new class `DefaultComponentMap` that provides the `defaultdict` API for `ComponetMap` objects.

While I was looking at things in `collections`, this PR removes support for Python<3.8 in `OrderdSet` and standardizes some of the logic across ComponentMap, ComponentSet, and OrderedSet.

- change `ComponentMap`/`ComponentSet` hashing cheme to 
   - only use `id()` for unhashable types
   - recursively process tuples
- add the `DefaultComponentMap` class
- update the storage mechanism for OrderedSet to rely on `dict` (now that we only support Python 3.8+)
- add a test for using tuples in `ComponentMap` / `DefaultComponentMap`

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/collections/component_map.py
+++ b/pyomo/common/collections/component_map.py
@@ -179,3 +179,32 @@ class ComponentMap(AutoSlots.Mixin, collections.abc.MutableMapping):
         else:
             self[key] = default
         return default
+
+
+class DefaultComponentMap(ComponentMap):
+    """A :py:class:`defaultdict` admitting Pyomo Components as keys
+
+    This class is a replacement for defaultdict that allows Pyomo
+    modeling components to be used as entry keys. The base
+    implementation builds on :py:class:`ComponentMap`.
+
+    """
+
+    __slots__ = ('default_factory',)
+
+    def __init__(self, default_factory=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.default_factory = default_factory
+
+    def __missing__(self, key):
+        if self.default_factory is None:
+            raise KeyError(key)
+        self[key] = ans = self.default_factory()
+        return ans
+
+    def __getitem__(self, obj):
+        _key = _hasher[obj.__class__](obj)
+        if _key in self._dict:
+            return self._dict[_key][1]
+        else:
+            return self.__missing__(obj)
--- a/pyomo/common/collections/component_map.py
+++ b/pyomo/common/collections/component_map.py
@@ -80,7 +80,7 @@ class ComponentMap(AutoSlots.Mixin, collections.abc.MutableMapping):
     __autoslot_mappers__ = {'_dict': _rehash_keys}
 
     def __init__(self, *args, **kwds):
-        # maps id(obj) -> (obj,val)
+        # maps id_hash(obj) -> (obj,val)
         self._dict = {}
         # handle the dict-style initialization scenarios
         self.update(*args, **kwds)
--- a/pyomo/common/collections/component_map.py
+++ b/pyomo/common/collections/component_map.py
@@ -9,21 +9,49 @@
 #  This software is distributed under the 3-clause BSD License.
 #  ___________________________________________________________________________
 
-from collections.abc import MutableMapping as collections_MutableMapping
+import collections
 from collections.abc import Mapping as collections_Mapping
 from pyomo.common.autoslots import AutoSlots
 
 
-def _rebuild_ids(encode, val):
+def _rehash_keys(encode, val):
     if encode:
-        return val
+        return list(val.values())
     else:
         # object id() may have changed after unpickling,
         # so we rebuild the dictionary keys
-        return {id(obj): (obj, v) for obj, v in val.values()}
+        return {_hasher[obj.__class__](obj): (obj, v) for obj, v in val}
 
 
-class ComponentMap(AutoSlots.Mixin, collections_MutableMapping):
+class _Hasher(collections.defaultdict):
+    def __init__(self, *args, **kwargs):
+        super().__init__(lambda: self._missing_impl, *args, **kwargs)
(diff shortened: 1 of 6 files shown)


pyomo/common/tests/test_component_map.py
```diff
@@ -0,0 +1,50 @@
+#  ___________________________________________________________________________
+#
+#  Pyomo: Python Optimization Modeling Objects
+#  Copyright (c) 2008-2022
```
mrmundt, line 4: ⚠️ Wrong year!

pyomo/common/collections/component_set.py
```diff
@@ -125,6 +132,7 @@ def clear(self):
     def remove(self, val):
         """Remove an element. If not a member, raise a KeyError."""
         try:
-            del self._data[id(val)]
+            del self._data[_hasher[val.__class__](val)]
         except KeyError:
-            raise KeyError("Component with id '%s': %s" % (id(val), str(val)))
+            _id = _hasher[val.__class__](val)
+            raise KeyError("Component with id '%s': %s" % (_id, val))
```
mrmundt, line 138: Do you need to cast `val` to a string here?

mrmundt (changes requested):

pyomo/common/collections/component_set.py
jsiirola: No.  `%s` will do it if needed.

pyomo/common/collections/component_map.py
```diff
@@ -68,18 +96,20 @@ def __str__(self):
 
     def __getitem__(self, obj):
         try:
-            return self._dict[id(obj)][1]
+            return self._dict[_hasher[obj.__class__](obj)][1]
         except KeyError:
-            raise KeyError("Component with id '%s': %s" % (id(obj), str(obj)))
+            _id = _hasher[obj.__class__](obj)
+            raise KeyError("Component with id '%s': %s" % (_id, obj))
```
emma58, line 102: This error message isn't so precise anymore, right, because `_id` might not be an id, if `obj` was hashable to begin with?

pyomo/common/collections/component_map.py
```diff
@@ -68,18 +96,20 @@ def __str__(self):
...
+        self._dict[_hasher[obj.__class__](obj)] = (obj, val)
 
     def __delitem__(self, obj):
         try:
-            del self._dict[id(obj)]
+            del self._dict[_hasher[obj.__class__](obj)]
         except KeyError:
-            raise KeyError("Component with id '%s': %s" % (id(obj), str(obj)))
+            _id = _hasher[obj.__class__](obj)
+            raise KeyError("Component with id '%s': %s" % (_id, obj))
```
emma58, line 112: Same question here

pyomo/common/collections/orderedset.py
```diff
@@ -9,42 +9,30 @@
...
+from collections.abc import MutableSet
+from pyomo.common.autoslots import AutoSlots
 
 
-class OrderedSet(MutableSet):
+class OrderedSet(AutoSlots.Mixin, MutableSet):
     __slots__ = ('_dict',)
 
     def __init__(self, iterable=None):
         # TODO: Starting in Python 3.7, dict is ordered (and is faster
```
mrmundt, line 21: I know this comment is old but it's (1) confusing - what is actually "TODO" about this? - and (2) we now only support 3.8+, so does some sort of action need to be taken?

pyomo/common/collections/component_map.py
```diff
@@ -149,3 +179,32 @@ def setdefault(self, key, default=None):
         else:
             self[key] = default
         return default
+
+
+class DefaultComponentMap(ComponentMap):
```
emma58, line 184: You're my favorite Pyomo developer of the week! :D

pyomo/common/collections/component_set.py
emma58, line 138: Same question as above here: `_id` is not necessarily an id, right?

pyomo/common/collections/orderedset.py
```diff
@@ -9,42 +9,30 @@
...
 #  ___________________________________________________________________________
 
-from collections.abc import MutableSet
 from collections import OrderedDict
+from collections.abc import MutableSet
+from pyomo.common.autoslots import AutoSlots
 
 
-class OrderedSet(MutableSet):
+class OrderedSet(AutoSlots.Mixin, MutableSet):
```
emma58, line 17: Aren't ComponentSets themselves ordered, now the dicts are ordered? This class seems like it's becoming a moot point, unless I'm misunderstanding something. I guess it's doesn't have the overhead of checking for hashability?

emma58 (approved):
This makes me *very* happy indeed! A few questions where I think the KeyError messages might need adjusted, and one general question (that doesn't really affect this PR) about whether there's any point to `OrderedSet` anymore.

pyomo/common/collections/orderedset.py
jsiirola: They are, and yes: the difference between `ComponentSet` and `OrderedSet` is very small (really just the overhead for hashability) - that said, the hasher is not trivial, so there is a performance difference.  The other reason to keep them both is backwards compatibility (and there is a nice symmetry to have an `OrderedSet` to go with `OrderedDict`)

mrmundt (approved):

PR #3152, opened 2024-02-20
Generalize the simple_constraint_rule decorator

### LINKED ISSUE #3092: simple_constraint_rule and numpy boolean
Hi, Pyomo development team! First of all, thank you so much for creating such an amazing package and popularizing OR!

## Summary

My suggestion is concerning an improvement of the wrapper/decorator `simple_constraint_rule` for constraints returning `None` or trivial booleans. I believe it would be helpful to extend it to Numpy booleans.

### Rationale

Numpy booleans might occur in some situations where arguments of the constraint function are Numpy float64 and the current wrapper `simple_constraint_rule` does not capture them.

### Description

I spent some time trying to figure out by myself a reliable workaround but I couldn't. So if anyone is interested in exploring this possibility in more detail, I believe it would be a nice new feature.

### Additional information
One of the problems I faced when trying to implement a wrapper myself was concerning functions with a single argument (the model). It seems that a tuple `(None,)` was also being passed when trying to instantiate the constraints.

jsiirola:
## Summary/Motivation:
This PR generalizes the `simple_constraint_rule` decorator so that it has a better chance of correctly handling `numpy.bool_` return types.  Note that this is not perfect, as it will not trigger automatic type registration - so if this is the first time a numpy type has been "seen" by Pyomo, it will still generate the normal exception.

- Generate the constraint result map based on the current value of `native_logical_types`
- Support passing the map_types set as an argument (for cases like numpy.bool_ where instances hash the same as other types)

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/constraint.py
+++ b/pyomo/core/base/constraint.py
@@ -90,7 +90,7 @@ def simple_constraint_rule(rule):
     for l_type in native_logical_types:
         result_map[l_type(True)] = Constraint.Feasible
         result_map[l_type(False)] = Constraint.Infeasible
-    # Note: some logical types has the same as bool (e.g., np.bool_), so
+    # Note: some logical types hash the same as bool (e.g., np.bool_), so
     # we will pass the set of all logical types in addition to the
     # result_map
     return rule_wrapper(rule, result_map, map_types=map_types)
@@ -116,7 +116,7 @@ def simple_constraintlist_rule(rule):
     for l_type in native_logical_types:
         result_map[l_type(True)] = Constraint.Feasible
         result_map[l_type(False)] = Constraint.Infeasible
-    # Note: some logical types has the same as bool (e.g., np.bool_), so
+    # Note: some logical types hash the same as bool (e.g., np.bool_), so
     # we will pass the set of all logical types in addition to the
     # result_map
     return rule_wrapper(rule, result_map, map_types=map_types)
--- a/pyomo/core/base/constraint.py
+++ b/pyomo/core/base/constraint.py
@@ -27,6 +27,7 @@ from pyomo.core.expr.numvalue import (
     as_numeric,
     is_fixed,
     native_numeric_types,
+    native_logical_types,
     native_types,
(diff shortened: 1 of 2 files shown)


pyomo/core/base/constraint.py
```diff
@@ -84,14 +85,14 @@ def C_rule(model, i, j):
...
-            None: Constraint.Skip,
-            True: Constraint.Feasible,
-            False: Constraint.Infeasible,
-        },
-    )
+    result_map = {None: Constraint.Skip}
+    for l_type in native_logical_types:
+        result_map[l_type(True)] = Constraint.Feasible
+        result_map[l_type(False)] = Constraint.Infeasible
+    # Note: some logical types has the same as bool (e.g., np.bool_), so
```
mrmundt, line 92: I think you meant this?
```suggestion
    # Note: some logical types act the same as bool (e.g., np.bool_), so
```

pyomo/core/base/constraint.py
```diff
@@ -109,14 +110,14 @@ def C_rule(model, i, j):
...
-            None: ConstraintList.End,
-            True: Constraint.Feasible,
-            False: Constraint.Infeasible,
-        },
-    )
+    result_map = {None: ConstraintList.End}
+    for l_type in native_logical_types:
+        result_map[l_type(True)] = Constraint.Feasible
+        result_map[l_type(False)] = Constraint.Infeasible
+    # Note: some logical types has the same as bool (e.g., np.bool_), so
```
mrmundt, line 117: I think you meant this?
```suggestion
    # Note: some logical types act the same as bool (e.g., np.bool_), so
```

mrmundt (commented):
There are no updated / added tests - is there a way to test to ensure the generalizability?

pyomo/core/base/constraint.py
jsiirola: Nope - I actually meant `hash`

mrmundt (approved):

PR #3161, opened 2024-02-22
fix division by zero error in linear presolve

michaelbynum:
## Summary/Motivation:
```
   ...: from pyomo.repn.plugins.nl_writer import NLWriter
   ...: import io
   ...: m = pe.ConcreteModel()
   ...: m.x = pe.Var()
   ...: m.y = pe.Var()
   ...: m.z = pe.Var()
   ...: m.obj = pe.Objective(expr=m.x**2 + m.y**2 + m.z**2)
   ...: m.c1 = pe.Constraint(expr=m.x == m.y + m.z + 1.5)
   ...: m.c2 = pe.Constraint(expr=m.z == -m.y)
   ...: writer = NLWriter()
   ...: out = io.StringIO()
   ...: writer.write(m, out)
ValueError                                Traceback (most recent call last)
Cell In [1], line 13
     11 writer = NLWriter()
     12 out = io.StringIO()
---> 13 writer.write(m, out)

File /.../pyomo/repn/plugins/nl_writer.py:409, in NLWriter.write(self, model, ostream, rowstream, colstream, **options)
    405 # Pause the GC, as the walker that generates the compiled NL
    406 # representation generates (and disposes of) a large number of
    407 # small objects.
    408 with _NLWriter_impl(ostream, rowstream, colstream, config) as impl:
--> 409     return impl.write(model)

File /.../pyomo/repn/plugins/nl_writer.py:775, in _NLWriter_impl.write(self, model)
    769 # This may fetch more bounds than needed, but only in the cases
    770 # where variables were completely eliminated while walking the
    771 # expressions, or when users provide superfluous variables in
    772 # the column ordering.
    773 var_bounds = {_id: v.bounds for _id, v in var_map.items()}
--> 775 eliminated_cons, eliminated_vars = self._linear_presolve(
    776     comp_by_linear_var, lcon_by_linear_nnz, var_bounds
    777 )
    778 del comp_by_linear_var
    779 del lcon_by_linear_nnz

File /.../pyomo/repn/plugins/nl_writer.py:1771, in _NLWriter_impl._linear_presolve(self, comp_by_linear_var, lcon_by_linear_nnz, var_bounds)
   1766         coef, coef2 = coef2, coef
   1767 else:
   1768     # In an attempt to improve numerical stability, we will
   1769     # solve for (and substitute out) the variable with the
   1770     # coefficient closer to +/-1)
-> 1771     log_coef = _log10(abs(coef))
   1772     log_coef2 = _log10(abs(coef2))
   1773     if abs(log_coef2) < abs(log_coef) or (
   1774         log_coef2 == -log_coef and log_coef2 < log_coef
   1775     ):

ValueError: math domain error
```

- Check for division by zero in NL writer's linear presolve

### Legal Acknowledgement

### DIFF
--- a/pyomo/repn/plugins/nl_writer.py
+++ b/pyomo/repn/plugins/nl_writer.py
@@ -1760,7 +1760,7 @@ class _NLWriter_impl(object):
                 id2_isdiscrete = var_map[id2].domain.isdiscrete()
                 if var_map[_id].domain.isdiscrete() ^ id2_isdiscrete:
                     # if only one variable is discrete, then we need to
-                    # substiitute out the other
+                    # substitute out the other
                     if id2_isdiscrete:
                         _id, id2 = id2, _id
                         coef, coef2 = coef2, coef
--- a/pyomo/repn/plugins/nl_writer.py
+++ b/pyomo/repn/plugins/nl_writer.py
@@ -1833,7 +1833,7 @@ class _NLWriter_impl(object):
                             if abs(expr_info.const) > TOL:
                                 # constraint is trivially infeasible
                                 raise InfeasibleConstraintException(
-                                    "model contains a trivially infeasible constrint "
+                                    "model contains a trivially infeasible constraint "
                                     f"{expr_info.const} == {coef}*{var_map[x]}"
                                 )
                             # constraint is trivially feasible
--- a/pyomo/repn/plugins/nl_writer.py
+++ b/pyomo/repn/plugins/nl_writer.py
@@ -1829,15 +1829,6 @@ class _NLWriter_impl(object):
                     if expr_info.linear[x] == 0:
                         nnz -= 1
                         coef = expr_info.linear.pop(x)
-                        if not nnz:
-                            if abs(expr_info.const) > TOL:
-                                # constraint is trivially infeasible
-                                raise InfeasibleConstraintException(
-                                    "model contains a trivially infeasible constraint "
-                                    f"{expr_info.const} == {coef}*{var_map[x]}"
-                                )
-                            # constraint is trivially feasible
-                            eliminated_cons.add(con_id)
                 elif a:
                     expr_info.linear[x] = c * a
                     # replacing _id with x... NNZ is not changing,
@@ -1847,6 +1838,15 @@ class _NLWriter_impl(object):
                     continue
                 _old = lcon_by_linear_nnz[old_nnz]
                 if con_id in _old:
+                    if not nnz:
+                        if abs(expr_info.const) > TOL:
+                            # constraint is trivially infeasible
+                            raise InfeasibleConstraintException(
+                                "model contains a trivially infeasible constraint "
+                                f"{expr_info.const} == {coef}*{var_map[x]}"
+                            )
+                        # constraint is trivially feasible
+                        eliminated_cons.add(con_id)
                     lcon_by_linear_nnz[nnz][con_id] = _old.pop(con_id)
             # If variables were replaced by the variable that
             # we are currently eliminating, then we need to update
--- a/pyomo/repn/plugins/nl_writer.py
+++ b/pyomo/repn/plugins/nl_writer.py
@@ -1820,10 +1820,24 @@ class _NLWriter_impl(object):
                 # appropriately (that expr_info is persisting in the
                 # eliminated_vars dict - and we will use that to
                 # update other linear expressions later.)
+                old_nnz = len(expr_info.linear)
                 c = expr_info.linear.pop(_id, 0)
+                nnz = old_nnz - 1
                 expr_info.const += c * b
                 if x in expr_info.linear:
                     expr_info.linear[x] += c * a
+                    if expr_info.linear[x] == 0:
+                        nnz -= 1
+                        coef = expr_info.linear.pop(x)
+                        if not nnz:
+                            if abs(expr_info.const) > TOL:
+                                # constraint is trivially infeasible
+                                raise InfeasibleConstraintException(
+                                    "model contains a trivially infeasible constrint "
+                                    f"{expr_info.const} == {coef}*{var_map[x]}"
+                                )
+                            # constraint is trivially feasible
+                            eliminated_cons.add(con_id)
                 elif a:
                     expr_info.linear[x] = c * a
                     # replacing _id with x... NNZ is not changing,
@@ -1831,9 +1845,7 @@ class _NLWriter_impl(object):
                     # this constraint
                     comp_by_linear_var[x].append((con_id, expr_info))
                     continue
-                # NNZ has been reduced by 1
-                nnz = len(expr_info.linear)
-                _old = lcon_by_linear_nnz[nnz + 1]
+                _old = lcon_by_linear_nnz[old_nnz]
                 if con_id in _old:
                     lcon_by_linear_nnz[nnz][con_id] = _old.pop(con_id)
             # If variables were replaced by the variable that
--- a/pyomo/contrib/solver/ipopt.py
+++ b/pyomo/contrib/solver/ipopt.py
@@ -17,7 +17,11 @@ from typing import Mapping, Optional, Sequence
 
 from pyomo.common import Executable
 from pyomo.common.config import ConfigValue, document_kwargs_from_configdict, ConfigDict
-from pyomo.common.errors import PyomoException, DeveloperError
+from pyomo.common.errors import (
+    PyomoException,
+    DeveloperError,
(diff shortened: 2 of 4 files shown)


pyomo/repn/plugins/nl_writer.py
```diff
@@ -1790,21 +1796,27 @@ def _linear_presolve(self, comp_by_linear_var, lcon_by_linear_nnz, var_bounds):
...
-                    x_lb = lb
-                if x_ub is None or (ub is not None and ub < x_ub):
-                    x_ub = ub
-                var_bounds[x] = x_lb, x_ub
-                if x_lb == x_ub and x_lb is not None:
-                    fixed_vars.append(x)
+                if a == 0:
+                    if (lb is not None and lb > b) or (ub is not None and ub < b):
+                        raise InfeasibleConstraintException(
+                            f'Model is infeasible: {lb} <= {var_map[_id]} <= {ub} and {var_map[_id]} == {a}*{var_map[x]} + {b} cannot both be satisfied'
```
jsiirola, line 1803: (black): Please split this string to keep lines under 88 characters

pyomo/repn/plugins/nl_writer.py
```diff
@@ -1790,21 +1796,27 @@ def _linear_presolve(self, comp_by_linear_var, lcon_by_linear_nnz, var_bounds):
...
+                        )
```
jsiirola, line 1804: If the model is feasible, shouldn't this end up fixing `var_map[_id]`?  Maybe the logic for trapping a == 0 should be moved earlier (to new line 1784?)?

michaelbynum: Ahh. I see. I missed that logic. Good catch. I'll update.

pyomo/repn/plugins/nl_writer.py
michaelbynum: will do

pyomo/repn/plugins/nl_writer.py
```diff
@@ -1779,7 +1785,8 @@ def _linear_presolve(self, comp_by_linear_var, lcon_by_linear_nnz, var_bounds):
                 a = -coef2 / coef
                 x = id2
                 b = expr_info.const = (lb - expr_info.const) / coef
-                expr_info.linear[x] = a
+                if a != 0:
+                    expr_info.linear[x] = a
```
michaelbynum, line 1789: @jsiirola, is this what you meant by "fixing `var_map[_id]`" below? I thought I knew what you meant, but know I'm not so sure. When `a` is `0`, we are replacing variable `_id` with `b`, which, I think, is the same as fixing it. Previously, we were still adding a coefficient of `0` to `expr_info.linear` for `x`, which was unnecessary. I fixed that. Is that what you meant?

pyomo/repn/plugins/nl_writer.py
```diff
@@ -1779,7 +1785,8 @@ def _linear_presolve(self, comp_by_linear_var, lcon_by_linear_nnz, var_bounds):
                 a = -coef2 / coef
```
jsiirola, line 1785: Will you still get a divide-by-0 error if both coefficients are 0?  Maybe something like:
```
m.c1 = pe.Constraint(expr=m.x + m.w == m.y + m.z + 1.5)
m.c2 = pe.Constraint(expr=m.z == -m.y)
m.c3 = pe.Constraint(expr=m.w == -m.x)
```
Maybe the right answer is to filter out 0's when we are performing the substitution?

pyomo/repn/plugins/nl_writer.py
jsiirola: I was thinking about not having to guard all the bounds manipulation code in the `if a != 0` block.  But, thinking more about it, I don't think that makes much sense: you would need the if `a == 0` block here, and then would still need the guard here (because we still want to finish this block out (to log the debug message and to add the `con_id` to the `eliminated_cons`.

pyomo/repn/plugins/nl_writer.py
michaelbynum: If both coeffiicents are zero, the constraint should simplify to  trivially satisfied (if `b` is `0`) or trivially infeasible (if `b` is not `0`). Let me test for that real quick.

michaelbynum: Okay, take a look at this version.

pyomo/repn/plugins/nl_writer.py
michaelbynum: Done.

pyomo/repn/plugins/nl_writer.py
jsiirola: I think this can work, but I wonder if there is a simpler implementation: (I think) the only way to get 0 coefficients is through previous substitutions.  If that is the case, then, you could revert all the edits to this file with (starting around @ L1853):
```diff
                 # appropriately (that expr_info is persisting in the
                 # eliminated_vars dict - and we will use that to
                 # update other linear expressions later.)
+                old_nnz = len(expr_info.linear)
                 c = expr_info.linear.pop(_id, 0)
+                nnz = old_nnz - 1
                 expr_info.const += c * b
                 if x in expr_info.linear:
                     expr_info.linear[x] += c * a
+                    if expr_info.linear[x] == 0:
+                        nnz -= 1
+                        coef = expr_info.linear.pop(x)
+                        if not nnz:
+                            if abs(expr_info.const) > TOL:
+                                # constraint is trivially infeasible
+                                raise InfeasibleConstraintException(
+                                    "model contains a trivially infeasible constrint "
+                                    f"{expr_info.const} == {coef}*{var_map[x]}"
+                                )
+                            # constraint is trivially feasible
+                            eliminated_cons.add(con_id)
                 elif a:
                     expr_info.linear[x] = c * a
                     # replacing _id with x... NNZ is not changing,
                     # but we need to remember that x is now part of
                     # this constraint
                     comp_by_linear_var[x].append((con_id, expr_info))
                     continue
-                # NNZ has been reduced by 1
-                nnz = len(expr_info.linear)
-                _old = lcon_by_linear_nnz[nnz + 1]
+                _old = lcon_by_linear_nnz[old_nnz]
                 if con_id in _old:
                     lcon_by_linear_nnz[nnz][con_id] = _old.pop(con_id)

```

michaelbynum: That worked! I'll push shortly.

michaelbynum: Okay, third time is the charm! Check out this version.

P.s. git is the best.

mrmundt: I'm not part of this conversation, but I had to +1 this. Git is the best.

jsiirola: Ick.  I think I lied to you.  The block beginning `if nnz == 0:` should be moved farther down:
```diff
@@ -1829,15 +1829,6 @@ class _NLWriter_impl(object):
                     if expr_info.linear[x] == 0:
                         nnz -= 1
                         coef = expr_info.linear.pop(x)
-                        if not nnz:
-                            if abs(expr_info.const) > TOL:
-                                # constraint is trivially infeasible
-                                raise InfeasibleConstraintException(
-                                    "model contains a trivially infeasible constraint "
-                                    f"{expr_info.const} == {coef}*{var_map[x]}"
-                                )
-                            # constraint is trivially feasible
-                            eliminated_cons.add(con_id)
                 elif a:
                     expr_info.linear[x] = c * a
                     # replacing _id with x... NNZ is not changing,
@@ -1847,6 +1838,15 @@ class _NLWriter_impl(object):
                     continue
                 _old = lcon_by_linear_nnz[old_nnz]
                 if con_id in _old:
+                    if not nnz:
+                        if abs(expr_info.const) > TOL:
+                            # constraint is trivially infeasible
+                            raise InfeasibleConstraintException(
+                                "model contains a trivially infeasible constrint "
+                                f"{expr_info.const} == {coef}*{var_map[x]}"
+                            )
+                        # constraint is trivially feasible
+                        eliminated_cons.add(con_id)
                     lcon_by_linear_nnz[nnz][con_id] = _old.pop(con_id)
             # If variables were replaced by the variable that
             # we are currently eliminating, then we need to update
```
The issue is in the original location, it will check the linear NNZ for all constraints (including inequalities and nonlinear constraints).  In the new location, it will only check for the linear equality constraints (which is what we want).

michaelbynum: Done.

jsiirola (approved):

PR #3169, opened 2024-02-28
Simplify Custom PyROS Domain Validators

shermanjasonaf:
## Summary/Motivation:

Following #3159 and [this comment](https://github.com/Pyomo/pyomo/pull/3126#discussion_r1496462769) in #3126, this PR simplifies a few custom PyROS domain validators which do not require statefulness.

Simplify custom PyROS domain validators.

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/pyros/tests/test_config.py
+++ b/pyomo/contrib/pyros/tests/test_config.py
@@ -12,9 +12,9 @@ from pyomo.core.base.param import Param, _ParamData
 from pyomo.contrib.pyros.config import (
     InputDataStandardizer,
     mutable_param_validator,
-    LoggerType,
+    logger_domain,
     SolverNotResolvable,
-    PositiveIntOrMinusOne,
+    positive_int_or_minus_one,
     pyros_config,
     SolverIterable,
     SolverResolvable,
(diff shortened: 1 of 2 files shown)


mrmundt (approved):

pyomo/contrib/pyros/tests/test_config.py
```diff
@@ -557,16 +557,22 @@ def test_positive_int_or_minus_one(self):
         """
         Test positive int or -1 validator works as expected.
         """
-        standardizer_func = PositiveIntOrMinusOne()
+        standardizer_func = positive_int_or_minus_one
         self.assertIs(
```
jsiirola, line 561: This should really be `assertEqual` and not `assertIs`

jsiirola (commented):
One super-minor change (that was not technically part of this PR, but should be fixed).

pyomo/contrib/pyros/tests/test_config.py
shermanjasonaf: I originally used `assertIs` in lieu of `assertEqual` to ensure `positive_int_or_minus_one` casts a numeric type object to an `int`, instead of just returning the object; `assertEqual` would admit a comparison of `1` with `1.00`, for example. Are you suggesting that `assertEqual` is surely sufficient for this check? If so, I will proceed with the change you have suggested. Otherwise, the check in line 569, which uses `assertEqual` and should otherwise be similar to the check here, may also need to be changed.

jsiirola: The problem is that there is no guarantee that `int(1) is int(1)`.  I think that if you want to check both the type and the value, I would recommend:
```python
ans = standardizer_func(1.0)
self.assertEqual(ans, 1)
self.assertIs(type(ans), int)
```

jsiirola (approved):

PR #3173, opened 2024-03-03
add type hints to components

michaelbynum:
- Add some type hints to components

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/block.py
+++ b/pyomo/core/base/block.py
@@ -2265,8 +2265,11 @@ class IndexedBlock(Block):
     def __init__(self, *args, **kwds):
         Block.__init__(self, *args, **kwds)
 
+    @overload
     def __getitem__(self, index) -> _BlockData:
-        return super().__getitem__(index)
+        ...
+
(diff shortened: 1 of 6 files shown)


jsiirola (commented):
I have a minor concern about this: by overriding `__getitem__`, you are adding the overhead of an additional function call to *every* element access.  Can you run the performance suite and document the impact on model creation time (if any)?

jsiirola:
It looks like there might be a workaround to avoid the function call:
```python
    @overload
    def __getitem__(self, index) -> _BlockData:
        ...

    __getitem__ = IndexedComponent.__getitem__  # type: ignore
```

michaelbynum:
> It looks like there might be a workaround to avoid the function call:
> 
> ```python
>     @overload
>     def __getitem__(self, index) -> _BlockData:
>         ...
> 
>     __getitem__ = IndexedComponent.__getitem__  # type: ignore
> ```

Genius!!!

mrmundt (approved):

Issue #2574, opened 2022-10-24
Appsi fbbt KeyError when Var bounded by Param


emma58:
## Summary

There's a KeyError when using appsi fbbt on a model where there are Params in the Var bounds

### Steps to reproduce the issue

```
from pyomo.environ import *
from pyomo.contrib import appsi

m = ConcreteModel()
m.p = Param(initialize=4)
m.x = Var(bounds=(0, m.p))
m.c = Constraint(expr=m.x >= 2)

m.pprint()

it = appsi.fbbt.IntervalTightener()
it.perform_fbbt(m)

m.pprint()
```

### Error Message

```
Traceback (most recent call last):
  File "maybe.py", line 62, in <module>
    it.perform_fbbt(m)
  File "/home/esjohn/src/pyomo/pyomo/contrib/appsi/fbbt.py", line 219, in perform_fbbt
    self.set_instance(model, symbolic_solver_labels=symbolic_solver_labels)
  File "/home/esjohn/src/pyomo/pyomo/contrib/appsi/fbbt.py", line 94, in set_instance
    self.add_block(model)
  File "/home/esjohn/src/pyomo/pyomo/contrib/appsi/base.py", line 914, in add_block
    self.add_variables(list(dict((id(var), var) for var in block.component_data_objects(Var, descend_into=True)).values()))
  File "/home/esjohn/src/pyomo/pyomo/contrib/appsi/base.py", line 787, in add_variables
    self._add_variables(variables)
  File "/home/esjohn/src/pyomo/pyomo/contrib/appsi/fbbt.py", line 107, in _add_variables
    cmodel.process_pyomo_vars(self._pyomo_expr_types, variables, self._var_map, self._param_map,
KeyError: 140146219916720
```

Note that replacing `m.p` with `value(m.p)` works as expected.

### Information on your system

Pyomo version: main
Python version: 3.8
Operating system: linux
How Pyomo was installed (PyPI, conda, source): source
Solver (if applicable):

jsiirola:
Note: as part of resolving this issue, we should revisit the use of APPSI FBBT in GDPopt (as disabled in #2575)

bknueven:
I just ran into this issue on current `main`. The same bug affects `appsi_cbc`.

bknueven:
And the same thing affects constraint bounds:

```python
from pyomo.environ import *                                                                                

m = ConcreteModel()
m.x = Var()
m.clb = Param(initialize=0)
[...]

PR #3181, opened 2024-03-07
BUGFIX: Solver Refactor - Allow no objective

mrmundt:
## Summary/Motivation:
There was a bug reported in https://github.com/Pyomo/pyomo/issues/2574#issuecomment-1983781290 that points out that you can't use `ipopt_v2` to solve something without an objective, which you technically can do in `v1`. This fixes that bug.

- Allow no objective 
- Populate the number of objectives / constraints / params attributes on `problem`

### Legal Acknowledgement


blnicho (approved):

pyomo/contrib/solver/base.py
```diff
@@ -418,9 +418,14 @@ def _map_results(self, model, results):
...
         legacy_soln.status = legacy_solution_status_map[results.solution_status]
         legacy_results.solver.termination_message = str(results.termination_condition)
-        obj = get_objective(model)
-        if len(list(obj)) > 0:
+        legacy_results.problem.number_of_constraints = model.nconstraints()
+        legacy_results.problem.number_of_variables = model.nvariables()
+        if model.nobjectives() == 0:
+            legacy_results.problem.number_of_objectives = 0
+        else:
+            obj = get_objective(model)
```
michaelbynum, line 426: This function does one of three things:
1. returns `None` if there is not an active objective
2. raises an error if there are multiple active objectives
3. returns the objective if there is exactly one objective

As a result, you will not be able to do `len(obj)` below.

michaelbynum (changes requested):
Thanks for fixing this @mrmundt. I left one comment below.

pyomo/contrib/solver/base.py
mrmundt: So I should just do `model.nobjectives()` is how I interpret that. Accurate?

michaelbynum: Yes. Then you just need to grab the objective to get the sense.

mrmundt: Updated!

michaelbynum (approved):

pyomo/contrib/solver/base.py
michaelbynum: yes

jsiirola: So this patch will raise an `AttributeError` for multiobjective NL files (which is in theory supported by the ASL / Pyomo API), in line 428 `None.sense`

mrmundt: What is the right way to get `sense` for multiple objectives?

jsiirola: A single `sense` is meaningless for multiple objectives (it should be a vector).  But the old interface never really handled MO problems well.  That said, my concern is that this is probably not the right place for a MO problem to end up with an error.

jsiirola (approved):

Issue #3179, opened 2024-03-06
Unavoidable `InfeasibleConstraintException` in NL writer for variable fixed outside of bounds


Robbybp:
## Summary
If a variable is fixed to a value outside of its bounds, the NL writer throws an `InfeasibleConstraintException` in `cache_fixed_var`. This may be reasonable default behavior, but I would argue there are cases where you would want to ignore this.

My particular use case is an initialization method where I solve a sequence of small subproblems, sometimes with solvers that do not support bounds (e.g. `calculate_variable_from_constraint` or `pyo.SolverFactory("scipy.fsolve")`). The solutions may violate bounds, and I may fix some of the resulting variable values before a subsequent solve. If this subsequent solve uses the NL writer and any of the fixed variables violate their bounds, I get the above error.

The workaround is to temporarily remove bounds on fixed variables before each sub-problem solve. This is not difficult, but I thought this might be a common enough issue to warrant an option in the NL writer.

### Steps to reproduce the issue

A basic example of my workflow that gives me the error is:

```python
import pyomo.environ as pyo                                                                                                                                                                                                                                                     
from pyomo.util.calc_var_value import calculate_variable_from_constraint                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                
m = pyo.ConcreteModel()                                                                                                                                                                                                                                                         
                                                                                                                                                                                                                                                                                
m.x = pyo.Var([1,2], initialize=1.0, bounds=(0, None))                                                                                                                                                                                                                          
m.u = pyo.Var([1,2], initialize=1.0, bounds=(0, None))                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                
m.eq = pyo.Constraint(pyo.PositiveIntegers)                                                                                                                                                                                                                                     
m.eq[1] = m.u[1] == 5                                                                                                                                                                                                                                                           
m.eq[2] = m.u[1] + 2 * m.u[2] == -1 + m.x[2]                                                                                                                                                                                                                                    
m.eq[3] = m.x[1] * m.x[2] == m.u[1] + m.u[2] + 7                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                
m.obj = pyo.Objective(expr=m.x[1]**2 + m.x[2]**2)                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                
calculate_variable_from_constraint(m.u[1], m.eq[1])                                                                                                                                                                                                                             
[...]

PR #3186, opened 2024-03-07
Improve performance of `solve_strongly_connected_components` for models with named expressions

Robbybp:
## Summary/Motivation:
Working with a small IDAES model, I noticed that `solve_strongly_connected_components` was much slower than I would have expected. This is for a model with about 900 variables and constraints:
```console
Identifier                                    ncalls   cumtime   percall      %                                                                    
-------------------------------------------------------------------------------                                                                    
root                                               1    40.800    40.800  100.0                                                                    
     --------------------------------------------------------------------------                                                                    
     full model post-solve                         1     0.463     0.463    1.1                                                                    
     solve-scc                                     1    40.337    40.337   98.9                                                                    
                          -----------------------------------------------------                                                                    
                          calc-var-from-con      530    20.910     0.039   51.8                                                                    
                          generate-scc           547    11.948     0.022   29.6                                                                    
                          igraph                   1     7.371     7.371   18.3                                                                    
                          scc-subsolver           16     0.100     0.006    0.2                                                                    
                          other                  n/a     0.008       n/a    0.0                                                                    
                          =====================================================                                                                    
     other                                       n/a     0.000       n/a    0.0                                                                    
     ==========================================================================                                                                    
===============================================================================                                                                    
```
`solve_strongly_connected_components` takes 40 s, half of which is in `calculate_variable_from_constraint`. This prompted me to review a lot of the infrastructure that `solve_strongly_connected_components` relies on and look for ways to speed things up. Most of the performance gains I will demonstrate come from not repeating work for the same named expressions, although reusing incidence graphs instead of walking expressions helps as well.

- Add an option for `solve_strongly_connected_components` to avoid using `calculate_variable_from_constraint`. This can be beneficial for models with large, nested named expressions, which the NL writer exploits but `calculate_variable_from_constraint` does not.
- Use `IncidenceMethod.ampl_repn` when constructing incidence graphs. Again, this exploits named expressions, where `IncidenceMethod.standard_repn` does not.
- Reuse the already existing incidence graph for the block triangularization in `generate_strongly_connected_components`
- Update `_ExternalFunctionVisitor` and `add_local_external_functions` to exploit named expressions
- Add an option for `TemporarySubsystemManager` to temporarily remove bounds on fixed variables. This was the motivation for the suggestion in https://github.com/Pyomo/pyomo/issues/3179

The timing breakdown after these improvements looks like:
```console
Identifier                                                           ncalls   cumtime   percall      %                     
------------------------------------------------------------------------------------------------------                     
root                                                                      1     4.159     4.159  100.0                     
     -------------------------------------------------------------------------------------------------                     
     full model post-solve                                                1     0.445     0.445   10.7                     
     solve-scc                                                            1     3.713     3.713   89.3                     
                          ----------------------------------------------------------------------------                     
                          generate-scc                                  547     2.431     0.004   65.5                     
                                       ---------------------------------------------------------------                     
                                       block-triang                       1     0.033     0.033    1.4                     
                                       generate-block                   547     2.397     0.004   98.6                     
                                                     -------------------------------------------------                     
                                                     block              546     0.015     0.000    0.6                     
                                                     external-fcns      546     0.341     0.001   14.2                     
                                                     identify-vars      546     1.999     0.004   83.4                     
                                                     reference         1092     0.036     0.000    1.5                     
                                                     other              n/a     0.007       n/a    0.3                     
                                                     =================================================                     
                                       other                            n/a     0.002       n/a    0.1                     
                                       ===============================================================                     
                          igraph                                          1     0.347     0.347    9.3                     
                          scc-subsolver                                 546     0.922     0.002   24.8                     
                          other                                         n/a     0.012       n/a    0.3                     
                          ============================================================================                     
     other                                                              n/a     0.000       n/a    0.0                     
     =================================================================================================                     
======================================================================================================                     
```
We are now bottlenecked by `identify_variables`, of all things.

### Legal Acknowledgement


pyomo/util/subsystems.py
```diff
@@ -17,18 +17,35 @@
...
+    def beforeChild(self, parent, child, index):
+        if (
+            not self._descend_into_named_expressions
+            and isinstance(child, NumericValue)
+            and child.is_named_expression_type()
+        ):
+            self.named_expressions.append(child)
+            return False, None
+        else:
+            return True, None
```
jsiirola, line 47: The following will likely be faster:
```python
if child.__class__ in native_types:
    return False, None
elif not self._descend_into_named_expressions and child.is_named_expression_type():
    self.named_expressions.append(child)
    return False, None
return True, None
```

pyomo/util/subsystems.py
```diff
@@ -38,26 +55,45 @@ def exitNode(self, node, data):
...
-        pass
-
-    def acceptChildResult(self, node, data, child_result, child_idx):
-        if child_result.__class__ in native_types:
-            return False, None
-        return child_result.is_expression_type(), None
+    # def acceptChildResult(self, node, data, child_result, child_idx):
+    #    if child_result.__class__ in native_types:
+    #        return False, None
+    #    return child_result.is_expression_type(), None
```
jsiirola, line 64: Should these be deleted?

pyomo/util/subsystems.py
```diff
@@ -75,7 +111,12 @@ def add_local_external_functions(block):
     return fcn_comp_map
 
 
-def create_subsystem_block(constraints, variables=None, include_fixed=False):
+from pyomo.common.timing import HierarchicalTimer
```
jsiirola, line 114: Can this import be moved to the top of the file?

jsiirola (approved):
Looks good.  A couple questions / suggestions, but nothing that would prevent merging

pyomo/util/subsystems.py
Robbybp: I included these for some supposed performance benefit in the initial implementation of `_ExternalFunctionVisitor`, but I actually notice a small performance improvement (about 0.1 s in the external function visitor) on my motivating test problem when I remove them. I will remove these.

pyomo/util/subsystems.py
Robbybp: I will decide what (if any) timing calls I want to keep in these helper functions, then move (or remove) this import.

mrmundt:
Real failure again, @Robbybp 

```
self = <pyomo.util.tests.test_subsystems.TestSubsystemBlock testMethod=test_local_external_functions_with_named_expressions>

    @unittest.skipUnless(find_GSL(), "Could not find the AMPL GSL library")
    def test_local_external_functions_with_named_expressions(self):
        m = self._make_model_with_external_functions(named_expressions=True)
>       variables = list(pyo.component_data_objects(pyo.Var))

/pyomo/pyomo/util/tests/test_subsystems.py:344: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

name = 'component_data_objects'

    def __getattr__(name):
        info = _relocated.get(name, None)
        if info is not None:
            target_obj = _import_object(name, *info)
            f_globals[name] = target_obj
            return target_obj
        elif _mod_getattr is not None:
            return _mod_getattr(name)
>       raise AttributeError(
            "module '%s' has no attribute '%s'" % (f_globals['__name__'], name)
        )
E       AttributeError: module 'pyomo.environ' has no attribute 'component_data_objects'

/pyomo/pyomo/common/deprecation.py:439: AttributeError
```

jsiirola (approved):

Robbybp:
> Attention: Patch coverage is 80.00000% with 9 lines in your changes are missing coverage. Please review.

My latest commits should address this.

PR #3199, opened 2024-03-18
Fix: mosek_direct updated to use putqconk instead of putqcon

Utkarsh-Detha:
## Fixes
This fix concerns QCQP models when solved using mosek. In MOSEK's Optimizer API, the putqcon method resets the Q matrix entries for all constraints to zero, while putqconk does so only for k-th constraint. The _add_constraints method in mosek_direct would call putqcon, but this would lead to loss of Q info with every subsequent call to the _add_constraints (if new Q info was given). This is now fixed, because the Q matrix in each constraint is updated in its own call to putqconk.

### Legal Acknowledgement

### DIFF
--- a/pyomo/solvers/plugins/solvers/mosek_direct.py
+++ b/pyomo/solvers/plugins/solvers/mosek_direct.py
@@ -492,13 +492,10 @@ class MOSEKDirect(DirectSolver):
             ptrb = (0,) + ptre[:-1]
             asubs = tuple(itertools.chain.from_iterable(l_ids))
             avals = tuple(itertools.chain.from_iterable(l_coefs))
-            qcsubi = tuple(itertools.chain.from_iterable(q_is))
-            qcsubj = tuple(itertools.chain.from_iterable(q_js))
-            qcval = tuple(itertools.chain.from_iterable(q_vals))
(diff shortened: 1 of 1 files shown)


Utkarsh-Detha:
Could someone help me figure out how to get this merged? Maybe @jsiirola? Thank you.

mrmundt (approved):

jsiirola (approved):

jsiirola:
This looks fine - One question: is there a significant performance difference between `putqcon` and `putqconk`?  If so, is it worth tracking if this is the first time `add_constraints()` was called for a model (allowing the interface to use `putqcon`  the first time and `putqconk` for all subsequent calls)?

Utkarsh-Detha:
Good question/suggestion, @jsiirola! My expectation was that for large problems `putqcon` should be faster than using a for-loop with `putqconk`. However, surprisingly enough, passing an n-by-n (dense) Q matrix in each constraint in an n-constraint problem is faster when using a for-loop with `putqconk`! 

We are looking at as to why this is the case in Optimizer API. In the meantime, as far as mosek_direct is concerned, I think the incoming changes already implement the fastest way to do things!

PR #3165, opened 2024-02-26
Added MAiNGO appsi-interface

MAiNGO-github:
## Fixes:
None

## Summary/Motivation:
Added an interface to our global optimization solver MAiNGO

## Changes proposed in this PR:
Add appsi-interface for MAiNGO

### Legal Acknowledgement

By contributing to this software project, I have read the [contribution guide](https://pyomo.readthedocs.io/en/stable/contribution_guide.html) and agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #2880, opened 2023-06-20
[WIP] Add COPT support for Pyomo

wujianjack:
## Fixes
None, new interface.

## Summary/Motivation:
This PR adds COPT support to Pyomo, including APPSI, IIS, the NL IO (`coptampl`), the Python IO (`copt_direct` and `copt_persistent`). The first two features are not tested so far, will do as soon as possible.

The `copt_direct` and `copt_persistent` depends on `coptpy`, which can be installed with `pip install coptpy`.

For larger instances, a valid COPT license is required, you can apply one from https://www.shanshu.ai/copt.

## Changes proposed in this PR:
- COPT support to APPSI
- COPT support to IIS
- COPT support to NL(`coptampl`) and Python IO (`copt_direct` and `copt_persistent`)

### Legal Acknowledgement

By contributing to this software project, I have read the [contribution guide](https://pyomo.readthedocs.io/en/stable/contribution_guide.html) and agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #3200, opened 2024-03-18
Add SCIP Persistent Support

Opt-Mucca:
## Fixes # .

None.

## Summary/Motivation:

Adds support for SCIP persistent solving. Persistent solving uses the python interface to SCIP directly as opposed to the current standard that involves read and writing a temporary model file.

The new dependency that it introduces is `PySCIPOpt`

- SCIPPersistent
- SCIPDirect
- Some new tests for the introduced features

### Legal Acknowledgement

### DIFF
--- a/pyomo/solvers/plugins/solvers/scip_direct.py
+++ b/pyomo/solvers/plugins/solvers/scip_direct.py
@@ -1,0 +1,838 @@
+#  ___________________________________________________________________________
+#
+#  Pyomo: Python Optimization Modeling Objects
+#  Copyright (c) 2008-2024
+#  National Technology and Engineering Solutions of Sandia, LLC
+#  Under the terms of Contract DE-NA0003525 with National Technology and
+#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain
+#  rights in this software.
+#  This software is distributed under the 3-clause BSD License.
+#  ___________________________________________________________________________
+
+import logging
+import re
+import sys
+
+from pyomo.common.collections import ComponentSet, ComponentMap, Bunch
+from pyomo.common.tempfiles import TempfileManager
+from pyomo.core import Var
+from pyomo.core.expr.numeric_expr import (
+    SumExpression,
+    ProductExpression,
+    UnaryFunctionExpression,
+    PowExpression,
+    DivisionExpression,
+)
+from pyomo.core.expr.numvalue import is_fixed
+from pyomo.core.expr.numvalue import value
+from pyomo.core.staleflag import StaleFlagManager
+from pyomo.repn import generate_standard_repn
+from pyomo.solvers.plugins.solvers.direct_solver import DirectSolver
+from pyomo.solvers.plugins.solvers.direct_or_persistent_solver import (
+    DirectOrPersistentSolver,
+)
+from pyomo.core.kernel.objective import minimize, maximize
+from pyomo.opt.results.results_ import SolverResults
+from pyomo.opt.results.solution import Solution, SolutionStatus
+from pyomo.opt.results.solver import TerminationCondition, SolverStatus
+from pyomo.opt.base import SolverFactory
+from pyomo.core.base.suffix import Suffix
+
+
+logger = logging.getLogger("pyomo.solvers")
+
+
+class DegreeError(ValueError):
+    pass
+
+
+def _is_numeric(x):
+    try:
+        float(x)
+    except ValueError:
+        return False
+    return True
+
+
+@SolverFactory.register("scip_direct", doc="Direct python interface to SCIP")
+class SCIPDirect(DirectSolver):
+
+    def __init__(self, **kwds):
+        kwds["type"] = "scipdirect"
+        DirectSolver.__init__(self, **kwds)
+        self._init()
+        self._solver_model = None
+
+    def _init(self):
+        try:
+            import pyscipopt
+
+            self._scip = pyscipopt
+            self._python_api_exists = True
+            self._version = str(self._scip.Model().version())
+            self._version_major = self._version.split(".")[0]
+        except ImportError:
+            self._python_api_exists = False
+        except Exception as e:
+            print("Import of pyscipopt failed - SCIP message=" + str(e) + "\n")
+            self._python_api_exists = False
+
+        # Note: Undefined capabilities default to None
+        self._max_constraint_degree = None
+        self._max_obj_degree = 1
+        self._capabilities.linear = True
+        self._capabilities.quadratic_objective = False
+        self._capabilities.quadratic_constraint = True
+        self._capabilities.integer = True
+        self._capabilities.sos1 = True
+        self._capabilities.sos2 = True
+
+        # Dictionary used exclusively for SCIP, as we want the constraint expressions
+        self._pyomo_var_to_solver_var_expr_map = ComponentMap()
+        self._pyomo_con_to_solver_con_expr_map = dict()
+
+    def _apply_solver(self):
+        StaleFlagManager.mark_all_as_stale()
+
+        # Supress solver output if requested
+        if self._tee:
+            self._solver_model.hideOutput(quiet=False)
+        else:
+            self._solver_model.hideOutput(quiet=True)
+
+        # Redirect solver output to a logfile if requested
+        if self._keepfiles:
+            # Only save log file when the user wants to keep it.
+            self._solver_model.setLogfile(self._log_file)
+            print("Solver log file: " + self._log_file)
+
+        # Set user specified parameters
+        for key, option in self.options.items():
+            try:
+                key_type = type(self._solver_model.getParam(key))
+            except KeyError:
+                raise ValueError(f"Key {key} is an invalid parameter for SCIP")
+
+            if key_type == str:
+                self._solver_model.setParam(key, option)
+            else:
+                if not _is_numeric(option):
+                    raise ValueError(
+                        f"Value {option} for parameter {key} is not a string and can't be converted to float"
+                    )
+                self._solver_model.setParam(key, float(option))
+
+        self._solver_model.optimize()
+
+        # TODO: Check if this is even needed, or if it is sufficient to close the open file
+        # if self._keepfiles:
+        #     self._solver_model.setLogfile(None)
+
+        # FIXME: can we get a return code indicating if SCIP had a significant failure?
+        return Bunch(rc=None, log=None)
+
+    def _get_expr_from_pyomo_repn(self, repn, max_degree=None):
+        referenced_vars = ComponentSet()
+
+        new_expr = repn.constant
+
+        if len(repn.linear_vars) > 0:
+            referenced_vars.update(repn.linear_vars)
+            new_expr += sum(
+                repn.linear_coefs[i] * self._pyomo_var_to_solver_var_expr_map[var]
+                for i, var in enumerate(repn.linear_vars)
+            )
+
+        for i, v in enumerate(repn.quadratic_vars):
+            x, y = v
+            new_expr += (
+                repn.quadratic_coefs[i]
+                * self._pyomo_var_to_solver_var_expr_map[x]
+                * self._pyomo_var_to_solver_var_expr_map[y]
+            )
+            referenced_vars.add(x)
+            referenced_vars.add(y)
+
+        # TODO: Introduce handling on non-linear expressions
+        if repn.nonlinear_expr is not None:
+
+            def get_nl_expr_recursively(pyomo_expr):
+                if not hasattr(pyomo_expr, "args"):
+                    if not isinstance(pyomo_expr, Var):
+                        return float(pyomo_expr)
+                    else:
+                        referenced_vars.add(pyomo_expr)
+                        return self._pyomo_var_to_solver_var_expr_map[pyomo_expr]
+                scip_expr_list = [0 for i in range(pyomo_expr.nargs())]
+                for i in range(pyomo_expr.nargs()):
+                    scip_expr_list[i] = get_nl_expr_recursively(pyomo_expr.args[i])
+                if isinstance(pyomo_expr, PowExpression):
+                    if len(scip_expr_list) != 2:
+                        raise ValueError(
+                            f"PowExpression has {len(scip_expr_list)} many terms instead of two!"
+                        )
+                    return scip_expr_list[0] ** (scip_expr_list[1])
+                elif isinstance(pyomo_expr, ProductExpression):
+                    return self._scip.quickprod(scip_expr_list)
+                elif isinstance(pyomo_expr, SumExpression):
+                    return self._scip.quicksum(scip_expr_list)
+                elif isinstance(pyomo_expr, DivisionExpression):
+                    if len(scip_expr_list) != 2:
+                        raise ValueError(
+                            f"DivisonExpression has {len(scip_expr_list)} many terms instead of two!"
+                        )
+                    return scip_expr_list[0] / scip_expr_list[1]
+                elif isinstance(pyomo_expr, UnaryFunctionExpression):
+                    if len(scip_expr_list) != 1:
+                        raise ValueError(
+                            f"UnaryExpression has {len(scip_expr_list)} many terms instead of one!"
+                        )
+                    if pyomo_expr.name == "sin":
+                        return self._scip.sin(scip_expr_list[0])
+                    elif pyomo_expr.name == "cos":
+                        return self._scip.cos(scip_expr_list[0])
+                    elif pyomo_expr.name == "exp":
+                        return self._scip.exp(scip_expr_list[0])
+                    elif pyomo_expr.name == "log":
+                        return self._scip.log(scip_expr_list[0])
+                    else:
+                        raise NotImplementedError(
+                            f"PySCIPOpt through Pyomo does not support the unary function {pyomo_expr.name}"
+                        )
+                else:
+                    raise NotImplementedError(
+                        f"PySCIPOpt through Pyomo does not yet support expression type {type(pyomo_expr)}"
+                    )
+
+            new_expr += get_nl_expr_recursively(repn.nonlinear_expr)
+
+        return new_expr, referenced_vars
+
+    def _get_expr_from_pyomo_expr(self, expr, max_degree=None):
+        if max_degree is None or max_degree >= 2:
+            repn = generate_standard_repn(expr, quadratic=True)
+        else:
+            repn = generate_standard_repn(expr, quadratic=False)
+
+        scip_expr, referenced_vars = self._get_expr_from_pyomo_repn(repn, max_degree)
+
+        return scip_expr, referenced_vars
+
+    def _scip_lb_ub_from_var(self, var):
+        if var.is_fixed():
+            val = var.value
+            return val, val
+        if var.has_lb():
+            lb = value(var.lb)
+        else:
+            lb = -self._solver_model.infinity()
+        if var.has_ub():
+            ub = value(var.ub)
+        else:
+            ub = self._solver_model.infinity()
+        return lb, ub
+
+    def _add_var(self, var):
+        varname = self._symbol_map.getSymbol(var, self._labeler)
+        vtype = self._scip_vtype_from_var(var)
+        lb, ub = self._scip_lb_ub_from_var(var)
+
+        scip_var = self._solver_model.addVar(lb=lb, ub=ub, vtype=vtype, name=varname)
+
+        self._pyomo_var_to_solver_var_expr_map[var] = scip_var
+        self._pyomo_var_to_solver_var_map[var] = scip_var.name
+        self._solver_var_to_pyomo_var_map[varname] = var
+        self._referenced_variables[var] = 0
+
+    def close(self):
+        """Frees SCIP resources used by this solver instance."""
+
+        if self._solver_model is not None:
+            self._solver_model.freeProb()
+            self._solver_model = None
+
+    def __exit__(self, t, v, traceback):
+        super().__exit__(t, v, traceback)
+        self.close()
+
+    def _set_instance(self, model, kwds={}):
+        DirectOrPersistentSolver._set_instance(self, model, kwds)
+        try:
+            self._solver_model = self._scip.Model()
+        except Exception:
+            e = sys.exc_info()[1]
+            msg = (
+                "Unable to create SCIP model. "
+                "Have you installed PySCIPOpt correctly?\n\n\t"
+                + "Error message: {0}".format(e)
+            )
+            raise Exception(msg)
+
+        self._add_block(model)
+
+        for var, n_ref in self._referenced_variables.items():
+            if n_ref != 0:
+                if var.fixed:
+                    if not self._output_fixed_variable_bounds:
+                        raise ValueError(
+                            "Encountered a fixed variable (%s) inside "
+                            "an active objective or constraint "
+                            "expression on model %s, which is usually "
+                            "indicative of a preprocessing error. Use "
+                            "the IO-option 'output_fixed_variable_bounds=True' "
+                            "to suppress this error and fix the variable "
+                            "by overwriting its bounds in the SCIP instance."
+                            % (var.name, self._pyomo_model.name)
+                        )
+
+    def _add_block(self, block):
+        DirectOrPersistentSolver._add_block(self, block)
+
+    def _add_constraint(self, con):
+        if not con.active:
+            return None
+
+        if is_fixed(con.body) and self._skip_trivial_constraints:
+            return None
+
+        conname = self._symbol_map.getSymbol(con, self._labeler)
+
+        if con._linear_canonical_form:
+            scip_expr, referenced_vars = self._get_expr_from_pyomo_repn(
+                con.canonical_form(), self._max_constraint_degree
+            )
+        else:
+            scip_expr, referenced_vars = self._get_expr_from_pyomo_expr(
+                con.body, self._max_constraint_degree
+            )
+
+        if con.has_lb():
+            if not is_fixed(con.lower):
+                raise ValueError(
+                    "Lower bound of constraint {0} is not constant.".format(con)
+                )
+        if con.has_ub():
+            if not is_fixed(con.upper):
+                raise ValueError(
+                    "Upper bound of constraint {0} is not constant.".format(con)
+                )
+
+        if con.equality:
+            scip_cons = self._solver_model.addCons(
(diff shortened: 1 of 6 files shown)


Opt-Mucca:
Two open questions left for this pull request:
- Is the current standard usage of persistent solving through the APPSI? If so then I should consider adding that.
- I avoided tinkering with the GitHub action workflow scripts. I'm not sure if it is appropriate to add a PySCIPOpt dependency there.

For info of how I tested the code:
- `pip install pyscipopt`
- python pyomo/solvers/tests/checks/test_SCIP{Persistent/Direct}.py

mrmundt:
@Opt-Mucca - answering your questions:

1. The current standard is APPSI, yes; however, the future standard is going to be using the persistent interface within `pyomo.contrib.solver` (the preview of the reworked solver interface system)
2. If `PySCIPOpt` is needed in order to run tests, then yes, absolutely, it's appropriate. Happy to help with that if you'd like.

Opt-Mucca:
@mrmundt Thanks for the fast reply!

1. After having a glance at `pyomo.contrib.solver`, it seems to overlap pretty heavily with the direct interfaces in `pyomo.solvers.plugins`. So that should save on work. I will probably avoid writing the APPSI functionality if there's soon to be a newer interface, but can do if there's demand.
2. Help would be appreciated! I assumed the only changes needed were those that are present in the other PRs #3165 and #2880 (replacing `coptpy` / `maingopy` with `pyscipopt`). If that's right then I can add those lines.

Opt-Mucca:
I have now added:
- Some automatic documentation for SCIPPersistent similar to other solver interfaces
- The `pip` instructions to the `.github/workflows`. (I avoided `conda` because I'm not familiar with it. I also only added PySCIPOpt via `pip` when using CPython)

pyomo/solvers/plugins/solvers/scip_direct.py
```diff
@@ -0,0 +1,838 @@
...
+        DirectOrPersistentSolver._set_instance(self, model, kwds)
+        try:
+            self._solver_model = self._scip.Model()
+        except Exception:
+            e = sys.exc_info()[1]
+            msg = (
+                "Unable to create SCIP model. "
+                "Have you installed PySCIPOpt correctly?\n\n\t"
+                + "Error message: {0}".format(e)
+            )
```
mrmundt, line 268: It may be preferable to check `self.available` first. If it's not available, then you can raise an error that it couldn't be found. If it's any _other_ error, you can say, "Hey, it's available, but something isn't working correctly. Here is your error message."

pyomo/solvers/plugins/solvers/scip_direct.py
```diff
@@ -0,0 +1,838 @@
...
+        else:
+            scip_expr, referenced_vars = self._get_expr_from_pyomo_expr(
+                con.body, self._max_constraint_degree
+            )
+
+        if con.has_lb():
+            if not is_fixed(con.lower):
+                raise ValueError(
+                    "Lower bound of constraint {0} is not constant.".format(con)
```
mrmundt, line 312: A general note: This file alone switches between f-strings, `str.format`, and conversion specifiers. It's kind of weird to see all used in a single file.

pyomo/solvers/plugins/solvers/scip_direct.py
```diff
@@ -0,0 +1,838 @@
...
+                extract_slacks = True
+                flag = True
+            if re.match(suffix, "rc"):
+                extract_reduced_costs = True
+                flag = True
+            if not flag:
+                raise RuntimeError(
+                    "***The scip_direct solver plugin cannot extract solution suffix="
+                    + suffix
+                )
```
mrmundt, line 460: Design question: is it better to raise an error or to log (or print loudly) that other requests are being ignored?

(Note: I don't have a strong preference either way.)

pyomo/solvers/plugins/solvers/scip_direct.py
```diff
@@ -0,0 +1,838 @@
...
+            )
+            self.results.solver.termination_condition = TerminationCondition.error
+            soln.status = SolutionStatus.error
+        else:
+            self.results.solver.status = SolverStatus.error
+            self.results.solver.termination_message = (
+                "Unhandled SCIP status (" + str(status) + ")"
+            )
+            self.results.solver.termination_condition = TerminationCondition.error
+            soln.status = SolutionStatus.error
```
mrmundt, line 599: I recommend putting all of this in its own helper function. It'll clean up `_postsolve`.

pyomo/solvers/plugins/solvers/scip_direct.py
```diff
@@ -0,0 +1,838 @@
...
+        self.results.problem.name = scip.getProbName()
+
+        if scip.getObjectiveSense() == "minimize":
+            self.results.problem.sense = minimize
+        elif scip.getObjectiveSense() == "maximize":
+            self.results.problem.sense = maximize
+        else:
+            raise RuntimeError(
+                f"Unrecognized SCIP objective sense: {scip.getObjectiveSense()}"
+            )
```
mrmundt, line 610: Do you need this check twice? It also exists in `_set_objective`.

pyomo/solvers/plugins/solvers/scip_persistent.py
```diff
@@ -0,0 +1,190 @@
...
+        vtype = self._scip_vtype_from_var(var)
+        lb, ub = self._scip_lb_ub_from_var(var)
+
+        self._solver_model.chgVarLb(scip_var, lb)
+        self._solver_model.chgVarUb(scip_var, ub)
+        self._solver_model.chgVarType(scip_var, vtype)
+
+    def write(self, filename, filetype=""):
+        """
+        Write the model to a file (e.g., and lp file).
```
mrmundt, line 101:
```suggestion
        Write the model to a file (e.g., an lp file).
```

mrmundt (changes requested):
Thanks for submitting this PR! Some comments from a quick pass.

mrmundt:
@Opt-Mucca - Please make sure you're on the most recent version of `black` and run (in the root): `black . -S -C  --exclude examples/pyomobook/python-ch/BadIndent.py`

.github/workflows/test_branches.yml
```diff
@@ -268,6 +268,12 @@ jobs:
                 || echo "WARNING: Gurobi is not available"
             python -m pip install --cache-dir cache/pip xpress \
                 || echo "WARNING: Xpress Community Edition is not available"
+            if [[ ${{matrix.python}} == pypy* ]]; then
+                echo "skipping SCIP for pypy"
+            else
+                python -m pip install --cache-dir cache/pip pyscipopt==5.0.0 \
```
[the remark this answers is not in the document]
Opt-Mucca: Removed the specific version. I added this as it will fail for any `pyscipopt<=4.3.0`.
Added the appropriate conda line.

.github/workflows/test_pr_and_main.yml
```diff
@@ -298,6 +298,12 @@ jobs:
                 || echo "WARNING: Gurobi is not available"
             python -m pip install --cache-dir cache/pip xpress \
                 || echo "WARNING: Xpress Community Edition is not available"
+            if [[ ${{matrix.python}} == pypy* ]]; then
+                echo "skipping SCIP for pypy"
+            else
+                python -m pip install --cache-dir cache/pip pyscipopt==5.0.0 \
```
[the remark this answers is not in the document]
Opt-Mucca: Same as above

pyomo/solvers/plugins/solvers/scip_direct.py
Opt-Mucca: I added a call to `self.available()` as suggested. It can't hurt to be there. The call itself raises an appropriate warning by default if the python binding can't be found.

pyomo/solvers/plugins/solvers/scip_direct.py
Opt-Mucca: This is my bad on copying code from the other persistent interfaces, and probably someone else's bad for copying code in turn (I assume that's why there's three standards). 
I am guilty of all the f-strings, and prefer that format, so have changed everything in scip_x.py to match.

pyomo/solvers/plugins/solvers/scip_direct.py
Opt-Mucca: I would lean slightly towards the print loudly option, but can see the potential danger for users. As a larger preference though, I would like to stick to similar behavior with the other interfaces.

pyomo/solvers/plugins/solvers/scip_direct.py
Opt-Mucca: Done as suggested

pyomo/solvers/plugins/solvers/scip_direct.py
```diff
@@ -0,0 +1,838 @@
...
+
+                if extract_duals:
+                    vals = [scip.getDualSolVal(con) for con in scip_cons]
+                    for val, name in zip(vals, con_names):
+                        soln_constraints[name]["Dual"] = val
+
+                if extract_slacks:
+                    vals = [scip.getSlack(con, scip_sol) for con in scip_cons]
+                    for val, name in zip(vals, con_names):
+                        soln_constraints[name]["Slack"] = val
```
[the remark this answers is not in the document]
Opt-Mucca: I decided against this suggestion. I'm not a fan of helper functions with a huge amount of arguments, and I think this code isn't as overly repetitive as the status handling.

pyomo/solvers/plugins/solvers/scip_persistent.py
```diff
@@ -0,0 +1,190 @@
...
+
+    def set_scip_param(self, param, val):
+        """
+        Set a SCIP parameter.
+
+        Parameters
+        ----------
+        param: str
+            The SCIP parameter to set. Options include any SCIP parameter.
+            Please see the SCIP documentation for options.
```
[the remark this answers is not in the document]
Opt-Mucca: Nice suggestion! Added a link

pyomo/solvers/plugins/solvers/scip_persistent.py
```diff
@@ -0,0 +1,190 @@
...
+                )
+            self._solver_model.addConsCoeff(scip_con, scip_var, coefficients[i])
+            con = self._solver_con_to_pyomo_con_map[scip_con.name]
+            self._vars_referenced_by_con[con].add(var)
+
+        sense = self._solver_model.getObjectiveSense()
+        self._solver_model.setObjective(obj_coef * scip_var, sense=sense, clear=False)
+
+    def reset(self):
+        """ This function is necessary to call before making any changes to the
```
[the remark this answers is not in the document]
Opt-Mucca: Thank you! I'm hoping that it's clear enough, and that this won't cause a headache for many users.

Opt-Mucca:
@mrmundt Black should be run and all the issues responded to.

pyomo/solvers/plugins/solvers/scip_direct.py
mrmundt: Copy-pasta catches us all.  🤷‍♀️

pyomo/solvers/plugins/solvers/scip_direct.py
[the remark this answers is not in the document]
mrmundt: Makes sense to me

mrmundt:
@Opt-Mucca - Spelling is the new failure. (You are suffering the same issue as all the core devs... The auto-linter and spell checker.)

Opt-Mucca:
@mrmundt TIL there are auto-linters that double as spell checkers. (I'll fix the spelling errors that come up each time the test fails. If it's just the two errors then I'm properly amazed.)

mrmundt:
@Opt-Mucca - If you could have heard the griping in our weekly dev call when I introduced the spell checker...

Opt-Mucca:
@mrmundt I can imagine (would probably be guilty of griping even if makes sense)

For the currently failing runs: I've added a fix that should work, but I will ask some colleagues tomorrow. (I'm not the greatest the warm start code). The fixes raise some warning where a loaded solution has variables of incorrect values (this may stem from non specified variables having a default value of 0)

The current reason of the pipeline failure: SCIP was trying to load an invalid solution before checking for feasibility (I think this is a longstanding python error with the interface and one of the functions)

The added solution: I simply added a check on the feasibility of the solution before adding it to the solution storage.

I will check with colleagues, but I'm not sure on how SCIP handles partial solutions. I am not familiar or don't know if we have any functionality like Gurobi that tries to complete the solution. This also holds for trying to repair a given infeasible solution.

Opt-Mucca:
Support for partial solution loading is now added. The failed tests are now passing locally too.

Opt-Mucca:
@mrmundt I am completely unsure on the protocol here. I've never done a merge request where I was not a member of the repository. 

The failed test now passes locally, but I am unable to trigger any pipelines here.

mrmundt:
@Opt-Mucca - Please make sure to read our [Contribution Guide](https://pyomo.readthedocs.io/en/stable/contribution_guide.html) as it'll answer most questions about our process.

As for triggering tests here, because you are a first-time contributor, one of our core dev team members needs to approve them (which I have been doing for this PR).

Opt-Mucca:
@mrmundt This will take awhile.
- I fixed the current issue (I was accessing the transformed problem after solving and the pyomo map was failing + SOS and nonlinear constraints can't access slack information). Such a fix required some new wrapped functions in PySCIPOpt, so will have to add them and release a new version.
- There are still errors in `test_writers`. I don't see this being easy. Most all (aside from an error with trivial constraints) stem from the reduced costs + slack being incorrect. This information is always weird to handle with SCIP for non-LP problems, so will need to think on it.

Opt-Mucca:
@mrmundt After talking to some colleagues we decided to not support reduced cost and dual values. They are simply too difficult to access for SCIP in a context that people would understand without disabling presolve (The loss in performance just doesn't outweigh the people that want to access this information).

There are currently two issues, both of which I would appreciate some advice or help on:
- I cannot calculate the slack correctly. After some tricks with ranged constraints I've ensured that SCIP is outputting the correct values, but the baseline comparisons seem to arbitrarily take either `rhs - activity` or `activity - lhs`. Internally we take the minimum of the two, but that doesn't seem to be the cases in the tests. Defaulting to one side in the case of ranged constraints also seems to fail. 
- I am not sure how to handle trivial constraints. I have set the flag `_skip_trivial_constraints=True`, but the corresponding tests are now failing. Adding a dummy constraint key to the internal dictionary mappings also doesn't seem to help.

blnicho:
We discussed this during the developer call and would advise that you not support the `slack` suffix. This is old functionality that we don't intend to support in the new solver interfaces and the same information can be calculated using a simple utility function. We're supportive if you want to skip trivial constraints, however this will likely require some updates to baseline tests.

Opt-Mucca:
I have now removed support for the `slack` suffix. 

I am still a bit unsure on what to do with the trivial constraints. The Python interface for SCIP just simply doesn't allow dummy constraints to be added. Some variables have to feature somewhere in the constraint. 
I think changing the tests to skip solvers with the `_skip_trivial_constraints=True` flag is a good idea. The alternative would be to pretend there is a constraint on the SCIP interface side, which would be extremely messy. Is anyone willing to change the test scope on the Pyomo dev team side?

mrmundt:
@Opt-Mucca - We have been discussing redesigning the solver test suite for a fair bit of time and have added it as an action item to #1030 . We will address this as a corner case for that.

Opt-Mucca:
@mrmundt Can you please run the pipeline now? I put `scip_persistent` as an expected fail for the `LP_trivial_constraint` test case. Everything is now passing locally.

Opt-Mucca:
Hmmmmmmmmmmmmmm. I didn't have scipy installed when I tested locally so it skipped that test. The error is that `np.float` is a non-allowed type when using `PySCIPOpt`. The user must give Python `float` or `int` types. Should probably consider expanding on that. For now there's two options, both of which are pretty fast:
- I just check if the user is giving non-standard Python values, e.g. `numpy.float64`, and convert them to Python (Will do this unless you advise against). Downside is this adds a tiny bit of overhead.
- Skip tests where `numpy.float64` values are passed as LHS and RHS values.

mrmundt:
@Opt-Mucca - Our general thought process is "protect users from themselves." I would recommend converting them but also log / alert the user so they know the conversion is happening.

Opt-Mucca:
@mrmundt I've added a warning now for each time a constraint is added and the RHS / LHS needs to be changed.
I had to use `type(x) == y`, because `isinstance(x, float)` also passes for `np.float` (didn't know that). The failed tests are now passing.

Opt-Mucca:
@mrmundt I'm not sure if the failing tests are because of my changes. Can I get a second opinion on that?

Opt-Mucca:
The latest change is pretty minor, but would mean that it's not backwards compatible to `pyscipopt` versions from years ago (I think compatibility would now be the last 2-3 releases). There was some basic missing functionality for retrieving the original constraints after starting / ending the solving process, and the Pyomo implementation was then giving some incorrect statistics.

pyomo/solvers/plugins/solvers/scip_direct.py
```diff
@@ -0,0 +1,769 @@
...
+
+        if obj.active is False:
+            raise ValueError("Cannot add inactive objective to solver.")
+
+        if obj.sense == minimize:
+            sense = "minimize"
+        elif obj.sense == maximize:
+            sense = "maximize"
+        else:
+            raise ValueError(f"Objective sense is not recognized: {obj.sense}")
```
mrmundt, line 444: I would instead do:
```
if obj.sense not in ['maximize', 'minimize']:
    raise ValueError(f"Objective sense is not recognized: {obj.sense}")
...
self._solver_model.setObjective(scip_expr, sense=obj.sense)
```

pyomo/solvers/plugins/solvers/scip_direct.py
```diff
@@ -0,0 +1,769 @@
...
+
+            try:
+                soln.gap = (
+                    self.results.problem.upper_bound - self.results.problem.lower_bound
+                )
+            except TypeError:
+                soln.gap = None
+
+        self.results.problem.number_of_constraints = scip.getNConss(transformed=False)
+        # self.results.problem.number_of_nonzeros = None
```
mrmundt, line 645: Remove?

pyomo/solvers/tests/checks/test_SCIPDirect.py
```diff
@@ -0,0 +1,335 @@
...
+    value,
+)
+from pyomo.opt import SolverFactory, TerminationCondition, SolutionStatus
+
+try:
+    import pyscipopt
+
+    scip_available = True
+except ImportError:
+    scip_available = False
```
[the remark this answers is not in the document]
mrmundt: I will ping @jsiirola again about this thought.

mrmundt (changes requested):

mrmundt (approved):
Sorry, I meant to click "Approve" - whoops. I left a few comments but none of them are so substantial that they should hold up this PR. They are mostly stylistic.

pyomo/solvers/plugins/solvers/scip_direct.py
Opt-Mucca: Removed

mrmundt:
FYI, the doc build/doc test failures are NOT RELATED. I have #3479 open to resolve it.

pyomo/solvers/plugins/solvers/scip_direct.py
Opt-Mucca: This does not work because you're comparing to string objects. `maximize` and `minimzie` in the code above are imports from pyomo.

I think I could change it to `obj.sense.value` or `obj.sense.name` (can't remember), but I wasn't 100% confident in covering all cases. I've left it as is.

Opt-Mucca:
I added a minor change: I no longer check what type the `LHS` of a constraint is. I added it originally because `pyomo` supports `numpy` floats and ints being passed, and I observed some crashes in SCIP's model creation. I looked into this a month or two ago, and concluded that it only happens from `np.version() < 2` and for `LHS` values. So I've just removed the check for the RHS. This should save many users getting spammed with warnings. Associated issues if anyone wants to help: https://github.com/scipopt/PySCIPOpt/issues/937

mrmundt:
@Opt-Mucca - can you do me one last favor please? I updated all of the existing files recently to reflect that our copyright is 2008 - 2025 now (see PR #3515 ) - could you please update your new files here to reflect that as well?

pyomo/solvers/tests/checks/test_SCIPPersistent.py
```diff
@@ -0,0 +1,318 @@
...
+    log,
+)
+from pyomo.opt import SolverFactory
+
+try:
+    import pyscipopt
+
+    scip_available = True
+except ImportError:
+    scip_available = False
```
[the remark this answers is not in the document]
jsiirola: In general, yes - but as this is in a testing file, it is not a big deal.

jsiirola (changes requested):
Apologies for how long it took to get to this.  Overall, this is not in too bad of shape, but there are a couple things that would be very good to rework:
- avoid the use of StandardRepn, and instead use / write a walker based on `StreamBasedExpressionVisitor`.
- rework this so we only need to walk each expression once (this is a performance issue, and is particularly important for persistent interfaces where performance is one of the big motivators).

pyomo/solvers/plugins/solvers/scip_direct.py
```diff
@@ -0,0 +1,762 @@
...
+
+        for i, v in enumerate(repn.quadratic_vars):
+            x, y = v
+            new_expr += (
+                repn.quadratic_coefs[i]
+                * self._pyomo_var_to_solver_var_expr_map[x]
+                * self._pyomo_var_to_solver_var_expr_map[y]
+            )
+            referenced_vars.add(x)
+            referenced_vars.add(y)
```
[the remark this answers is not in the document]
Opt-Mucca: Hey @jsiirola thanks for the review! It's now my turn to apologise for taking so long though...... Have started a new job and simply been waiting on a burst of motivation some evening to look at this.

I used the representation compiler because it was the first thing that I found and seemed to fit what I wanted. I do not have an overview whatsoever of the various Pyomo data structures, so that's on me for not doing enough research. I will try and look into `StreamBasedExpressionVisitor` and make the relevant changes (No complaints from me if someone else on the team wants to do the changes themself). Is there some particular date for when it would be good if this PR was finished?

pyomo/solvers/plugins/solvers/scip_direct.py
```diff
@@ -0,0 +1,762 @@
...
+        return Bunch(rc=None, log=None)
+
+    def _get_expr_from_pyomo_repn(self, repn, max_degree=None):
+        referenced_vars = ComponentSet()
+
+        degree = repn.polynomial_degree()
+        if (max_degree is not None) and (degree > max_degree):
+            raise DegreeError(
+                "While SCIP supports general non-linear constraints, the objective must be linear. "
+                "Please reformulate the objective by introducing a new variable. "
```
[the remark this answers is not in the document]
Opt-Mucca: I am of the opinion that the modelling language should not be doing something that the user didn't specifically ask for.  For example, creating some "shadow" variable that is likely never revealed to the user and would cause a mismatch in number of variables in the model and the number of variables in the users list / dict of "standard" variables. We have had many debates about this in `PySCIPOpt`.
Completely fair to have the stronger opinion of "Make it as easy as possible for the user"

pyomo/solvers/plugins/solvers/scip_direct.py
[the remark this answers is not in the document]
jsiirola: @Opt-Mucca: No worries!  The internals of Pyomo are still rather poorly documented (not enough nights and weekends), so there is no reason to expect you to have stumbled across the "right" set of utilities.  If you are looking for a potentially useful inspiration, you might look at the `Pyomo2SympyVisitor` walker for converting Pyomo expressions to sympy expressions (in `pyomo/core/expr/sympy_tools.py`)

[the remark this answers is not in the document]
jsiirola: Oh - and you asked about timeline...  We are currently planning for the next release on 6 August, so ideally, we would aim for all PRs to be in final review about a week before that.

pyomo/solvers/plugins/solvers/scip_direct.py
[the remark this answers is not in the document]
jsiirola: In general, I completely agree.  However, we have historically made exceptions for trivial changes that 
- can only be implemented in a single way
- enable a solver to be run on a model that a rational user would expect the solver to run on (that is, we want users to be able to easily switch between solvers without having to recode their model)

Examples of changes that we do make are:
- Adding a trivial (fixed) variable to LP files so that we can add constant terms to objectives (several solvers do not accept a constant as part of the objective expression in LP format)
- Converting the optimization sense (max to min) for solvers that only accept one sense.

In my opinion, this would fall into the same category...  But I also think that this is on the edge, and tossing a warning for the user would be completely warranted.  (I can also be convinced that this reformulation is a "bridge too far" and the error is the right path forward).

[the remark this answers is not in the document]
Opt-Mucca: You've now rather convinced me..... I still don't like the choice, but I do believe that if there's already some exceptions then this would be an additional sensible one,

michaelbynum:
@Opt-Mucca, we are refactoring the solver interfaces in pyomo/contrib/solver (https://pyomo.readthedocs.io/en/stable/explanation/experimental/solvers.html). We would really like this functionality there instead of in pyomo/solvers/plugins/solvers (which will eventually become deprecated). I plan to build off your branch by moving the files into pyomo/contrib/solver and reworking them to match the new interfaces. Please let us know if you have any objections/concerns.

Opt-Mucca:
@michaelbynum No complaints at all. Sorry in advance if the code is a bit sloppy (mainly the SCIP naming and constant checks for `numpy.float` as a LHS). Feel free to ping me if you have questions about it.
Does that mean that there's no pressure to complete this PR before mid August? Edit: Or complete this PR at all?

michaelbynum:
@Opt-Mucca, Thanks. We will continue to work off of this PR, so it will eventually get merged. However, I doubt it will happen before mid-August.

michaelbynum:
I'm closing this one as it has been superseded by #3744. Note that #3744 builds off of this branch. It just moves the new interfaces to contrib and makes the "automatic" part of the persistent solver work.

PR #3201, opened 2024-03-18
Add "mixed" standard form representation

jsiirola:
## Summary/Motivation:
This adds a new option to the "Standard Form" compiler that returns a mix of inequality and equality constraints.  This representation is particularly useful for solver interfaces.  This PR also resolves an error when collapsing unused variables out of the returned matrices.  The new implementation is also slightly more performant.

- add a "mixed" representation to the standard form compiler
- correct indexing error when removing unused columns from the `A` matrix and `c` vector 

### Legal Acknowledgement

### DIFF
--- a/pyomo/repn/plugins/standard_form.py
+++ b/pyomo/repn/plugins/standard_form.py
@@ -140,6 +140,15 @@ class LinearStandardFormCompiler(object):
         ),
     )
     CONFIG.declare(
+        'mixed_form',
+        ConfigValue(
+            default=False,
+            domain=bool,
+            description='Return A in mixed form (the comparison operator is a '
+            'mix of <=, ==, and >=)',
+        ),
+    )
+    CONFIG.declare(
         'show_section_timing',
         ConfigValue(
             default=False,
@@ -332,6 +341,9 @@ class _LinearStandardFormCompiler_impl(object):
         # Tabulate constraints
         #
         slack_form = self.config.slack_form
+        mixed_form = self.config.mixed_form
+        if slack_form and mixed_form:
+            raise ValueError("cannot specify both slack_form and mixed_form")
         rows = []
         rhs = []
         con_data = []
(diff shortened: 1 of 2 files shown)


pyomo/repn/plugins/standard_form.py
```diff
@@ -372,7 +384,30 @@ def write(self, model):
                     f"model contains a trivially infeasible constraint, '{con.name}'"
                 )
 
-            if slack_form:
+            if mixed_form:
```
mrmundt, line 387: Is it more likely that a user will want `mixed_form` over `slack_form`?

pyomo/repn/tests/test_standard_form.py
```diff
@@ -222,6 +241,28 @@ def test_alternative_forms(self):
...
+        self.assertEqual(list(map(str, repn.x)), ['x', 'y[0]', 'y[1]', 'y[3]'])
+        self.assertEqual(
+            list(v.bounds for v in repn.x), [(None, None), (0, 10), (-5, 10), (-5, -2)]
+        )
+        ref = np.array(
+            [[1, 0, 2, 0], [0, 0, 1, 4], [0, 1, 6, 0], [0, 1, 6, 0], [1, 1, 0, 0]]
+        )
+        self.assertTrue(np.all(repn.A == ref))
+        print(repn)
+        print(repn.b)
```
mrmundt, line 260: Debugging?

pyomo/repn/tests/test_standard_form.py
```diff
@@ -222,6 +241,28 @@ def test_alternative_forms(self):
...
+        self.assertTrue(np.all(repn.b == np.array([3, 5, 6, -3, 8])))
+        self.assertTrue(np.all(repn.c == np.array([[-1, 0, -5, 0], [1, 0, 0, 15]])))
+        # Note that the solution is a mix of inequality and equality constraints
+        # self._verify_solution(soln, repn, False)
```
mrmundt, line 264: Debugging?

pyomo/repn/plugins/standard_form.py
jsiirola: *Users* will probably do the regular form first, and then maybe the slack form.  However, I am working on some (more efficient) solver interfaces (a rewrite of `gurobi_direct`)  that will use the `mixed_form`.

pyomo/repn/tests/test_standard_form.py
jsiirola: Not debugging: a breadcrumb as to why that test is not an exact copy-and-paste of the cases above.  I have expanded the comment.

pyomo/repn/tests/test_standard_form.py
```diff
@@ -222,6 +241,28 @@ def test_alternative_forms(self):
...
+        self.assertEqual(list(map(str, repn.x)), ['x', 'y[0]', 'y[1]', 'y[3]'])
+        self.assertEqual(
+            list(v.bounds for v in repn.x), [(None, None), (0, 10), (-5, 10), (-5, -2)]
+        )
+        ref = np.array(
+            [[1, 0, 2, 0], [0, 0, 1, 4], [0, 1, 6, 0], [0, 1, 6, 0], [1, 1, 0, 0]]
+        )
+        self.assertTrue(np.all(repn.A == ref))
+        self.assertTrue(np.all(repn.b == np.array([3, 5, 6, -3, 8])))
+        self.assertTrue(np.all(repn.c == np.array([[-1, 0, -5, 0], [1, 0, 0, 15]])))
```
emma58, line 260: EEEEEK, I didn't notice this before--should we really let a "standard form" have more than one objective? This caused me a moment of terror on the the thought that I would definitely assume `c` if a vector of the same dimension as the number of columns in `A`... The matrix `c` makes sense to me from a writer perspective. From a user/transformation perspective, we're always going to need to be checking that this isn't the case... But I guess we have to do that anyway, so why not here?

emma58 (approved):
This looks good--I'm excited about it! :) My only big question is that I had a little argument with myself about multiple objectives... I feel like you're probably right, but the idea of more than one objective in "standard form" seems a little terrifying.

PR #3206, opened 2024-03-22
AMPL solver duplicate funcadd fix

eslickj:
## Summary/Motivation:

IDAES recently made a change where they set AMPLFUNC for external functions in IDAES on import.  I'm not exactly sure of the details of why this was done, but it causes external functions to be imported twice.  This is a minor change to only add lines from PYOMO_AMPLFUNC to AMPLFUNC if they are not already in AMPLFUNC.  

- Check the AMPLFUNC environment variable contents before adding PYOMO_AMPL contents to avoid duplicate lines.  

### Legal Acknowledgement


Robbybp:
This was done to work around this issue in ASL: https://github.com/ampl/asl/issues/13. Thanks for this patch.

pyomo/solvers/plugins/solvers/ASL.py
```diff
@@ -158,13 +158,15 @@ def create_command_line(self, executable, problem_files):
...
         #
-        if 'PYOMO_AMPLFUNC' in env:
-            if 'AMPLFUNC' in env:
-                env['AMPLFUNC'] += "\n" + env['PYOMO_AMPLFUNC']
+        if "PYOMO_AMPLFUNC" in env:
+            if "AMPLFUNC" in env:
+                for line in env["PYOMO_AMPLFUNC"].split("\n"):
+                    if line not in env["AMPLFUNC"]:
+                        env["AMPLFUNC"] += "\n" + line
             else:
```
jsiirola, line 166: There is an edge case where this would match a substring in AMPLFUNC.  Not terribly likely, as the `.so` / `.dll` extension would end up appearing in the middle of the string, but still possible.  I think a better solution would be:
```python
exitsing = set(env['AMPLFUNC'].split("\n")
for line in env["PYOMO_AMPLFUNC"].split("\n"):
    if line not in existing:
        env["AMPLFUNC"] += "\n" + line
```

pyomo/solvers/plugins/solvers/IPOPT.py
```diff
@@ -119,30 +119,32 @@ def create_command_line(self, executable, problem_files):
...
         # libraries
         #
-        if 'PYOMO_AMPLFUNC' in env:
-            if 'AMPLFUNC' in env:
-                env['AMPLFUNC'] += "\n" + env['PYOMO_AMPLFUNC']
+        if "PYOMO_AMPLFUNC" in env:
+            if "AMPLFUNC" in env:
+                for line in env["PYOMO_AMPLFUNC"].split("\n"):
+                    if line not in env["AMPLFUNC"]:
+                        env["AMPLFUNC"] += "\n" + line
```
jsiirola, line 126: Same comment as above

jsiirola (commented):
It is unfortunate that Black was allowed to muck with all the string endings (we generally use `black -C -S` for the codebase).  One minor implementation comment and we can merge.

pyomo/solvers/plugins/solvers/IPOPT.py
eslickj: Thanks for looking at this.  I'll make the changes.

Robbybp (changes requested):
Can you make this change in `pyomo.contrib.pynumero.interfaces.pyomo_nlp` as well? The following patch should work:
```patch
diff --git a/pyomo/contrib/pynumero/interfaces/pyomo_nlp.py b/pyomo/contrib/pynumero/interfaces/pyomo_nlp.py
index 51edd0931..ce148f50e 100644
--- a/pyomo/contrib/pynumero/interfaces/pyomo_nlp.py
+++ b/pyomo/contrib/pynumero/interfaces/pyomo_nlp.py
@@ -92,15 +92,13 @@ class PyomoNLP(AslNLP):
             # The NL writer advertises the external function libraries
             # through the PYOMO_AMPLFUNC environment variable; merge it
             # with any preexisting AMPLFUNC definitions
-            amplfunc = "\n".join(
-                filter(
-                    None,
-                    (
-                        os.environ.get('AMPLFUNC', None),
-                        os.environ.get('PYOMO_AMPLFUNC', None),
-                    ),
-                )
-            )
+            amplfunc_lines = os.environ.get("AMPLFUNC", "").split("\n")
+            existing = set(amplfunc_lines)
+            for line in os.environ.get("PYOMO_AMPLFUNC", "").split("\n"):
+                # Skip (a) empty lines and (b) lines we already have
+                if line != "" and line not in existing:
+                    amplfunc_lines.append(line)
+            amplfunc = "\n".join(amplfunc_lines)
             with CtypesEnviron(AMPLFUNC=amplfunc):
                 super(PyomoNLP, self).__init__(nl_file)
```
I would just push it to this branch, but don't think I have permissions.

eslickj:
@Robbybp, yes, I'll do it.

Robbybp (approved):
Looks good, thanks for the patch

jsiirola (approved):
This looks good, and I won't hold up merging, but can we add a regression test to verify the expected behavior?  Maybe in the future we can abstract this to a helper function so we don't have the same code in 3 locations.

eslickj:
> This looks good, and I won't hold up merging, but can we add a regression test to verify the expected behavior? Maybe in the future we can abstract this to a helper function so we don't have the same code in 3 locations.

@jsiirola, I can do both of these things in the next couple days.  I'm not in a really big hurry for this to be merged, unless someone else is.

blnicho (approved):

jsiirola (approved):

PR #3212, opened 2024-03-26
Require variables and constraints to be specified separately in `IncidenceGraphInterface.remove_nodes`

Robbybp:
## Fixes
If invalid components (those not contained in the incidence graph, including random objects like lists and tuples), were provided to `remove_nodes`, we would silently do nothing instead of raising an error.

## Summary/Motivation:
The above behavior was present because we weren't calling `self._validate_input` in `remove_nodes`. Updating to call `self._validate_input` would still silently continue if a bad component was provided, and would give a non-helpful error message if a random object was provided. 

- Update `_validate_input` to give a helpful error message if something that is not in the cached var/con `ComponentMap`s is provided
- Call `_validate_input` in `remove_nodes`
- Require that variables and constraints are specified separately. This allows us to raise helpful error messages in `_validate_input`. The alternative would probably be to have some more complicated, custom error-checking code in `remove_nodes` directly. I'm considering this breaking change because (a) it makes the method consistent with all the other methods on `IncidenceGraphInterface` and (b) I don't think many people are using this method.

### Legal Acknowledgement


pyomo/contrib/incidence_analysis/interface.py
```diff
@@ -866,35 +884,48 @@ def remove_nodes(self, nodes, constraints=None):
...
+        variables: list
+            VarData objects whose nodes will be removed from the incidence graph
         constraints: list
-            VarData or ConData objects whose columns or rows will be
-            removed from the incidence matrix.
+            ConData objects whose nodes will be removed from the incidence graph
+
+        .. note::
+
+           **Breaking change in Pyomo vTBD**
```
mrmundt, line 894: Do you want to add in a deprecation warning? It would be "louder" to the users.

Robbybp: I've updated `remove_nodes` to allow the old API by manually sorting variables/constraints and raising a deprecation warning if either is provided in the wrong list.

pyomo/contrib/incidence_analysis/interface.py
```diff
@@ -866,35 +884,72 @@ def remove_nodes(self, nodes, constraints=None):
...
+        cons_to_validate = []
+        depr_msg = (
+            "In IncidenceGraphInterface.remove_nodes, passing variables and"
+            " constraints in the same list is deprecated. Please separate your"
+            " variables and constraints and pass them in the order variables,"
+            " constraints."
+        )
+        for var in variables:
+            if var in self._con_index_map:
+                deprecation_warning(depr_msg, version="TBD")
```
jsiirola, line 925: Please make the version `6.7.2.dev0` instead of `TBD` (this makes it easier to search for when we are cutting the release)

pyomo/contrib/incidence_analysis/interface.py
```diff
@@ -866,35 +884,72 @@ def remove_nodes(self, nodes, constraints=None):
...
-        cons_to_include = [c for c in self.constraints if c not in to_exclude]
+
+        vars_to_validate = []
+        cons_to_validate = []
+        depr_msg = (
+            "In IncidenceGraphInterface.remove_nodes, passing variables and"
+            " constraints in the same list is deprecated. Please separate your"
+            " variables and constraints and pass them in the order variables,"
+            " constraints."
+        )
```
jsiirola, line 922: The logic below will emit this message for each variable in the list.  While the `deprecation_warning` logic should suppress the duplicate messages, it would probably be better to not emit them in the first place.  Would it be simpler to just run through the loops below and then
```python
for var in variables:
    if var in self._var_index_map:
        vars_to_validate.append(var)
    elif var in sef._con_index_map:
        cons_to_validate.append(var)
    else:
        unknown_comp.append(var)
if cons_to_validate:
    deprecation_warning(depr_msg, version="6.7.2.dev0")
nVars = len(vars_to_validate)
for var in variables:
    if var in self._var_index_map:
        vars_to_validate.append(var)
    elif var in sef._con_index_map:
        cons_to_validate.append(var)
    else:
        unknown_comp.append(var)
if len(vars_to_validate) != nVars:
    deprecation_warning(depr_msg, version="6.7.2.dev0")
if unknown_comp:
    raise ValueError(...)
```

jsiirola (approved):

PR #3214, opened 2024-03-28
BUG FIXES: Solver Refactor for IDAES Integration

mrmundt:
## Summary/Motivation:

Working with @andrewlee94 on adding initial support for v2 solvers in IDAES, we discovered some bugs. This PR addresses those bugs.

- Remove usage of `model.nobjectives` (etc.) due to them not existing on `ScalarBlocks`
- Allow forward compatibility of `solver_options` and `writer_config` in `LegacySolverWrapper`
- Yell loudly if users try to use both `options` and `solver_options` in the same command
- Add some docs (not a lot, but some) about how both types of option passing are supported

### Legal Acknowledgement


pyomo/contrib/solver/base.py
```diff
@@ -348,9 +348,20 @@ class LegacySolverWrapper:
...
+        # and new options. That is silly. So we will yell at them.
+        if 'options' in kwargs and 'solver_options' in kwargs:
+            raise ApplicationError(
+                "Both 'options' and 'solver_options' were requested. "
+                "Please use one or the other, not both."
+            )
+        elif 'options' in kwargs:
+            self.options = kwargs.pop('options')
+        elif 'solver_options' in kwargs:
+            self.solver_options = kwargs.pop('solver_options')
```
blnicho, line 364: Is there a reason we need both the `self.options` and `self.solver_options` attributes? Can't we simplify it down to using the same attribute for both keyword options at this point?

pyomo/contrib/solver/base.py
```diff
@@ -393,8 +406,29 @@ def _map_config(
...
+        if hasattr(self, 'solver_options'):
+            self.config.solver_options.set_value(self.solver_options)
+        if (options is not NOTSET) and (solver_options is not NOTSET):
+            # There is no reason for a user to be trying to mix both old
+            # and new options. That is silly. So we will yell at them.
+            # Example that would raise an error:
+            # solver.solve(model, options={'foo' : 'bar'}, solver_options={'foo' : 'not_bar'})
+            raise ApplicationError(
+                "Both 'options' and 'solver_options' were declared "
+                "in the 'solve' call. Please use one or the other, "
```
blnicho, line 420: Isn't `_map_config` also called from set_options? Is referencing "the solve call" here too specific?

pyomo/contrib/solver/base.py
```diff
@@ -464,7 +502,15 @@ def _solution_handler(
...
-        model.solutions.add_symbol_map(symbol_map)
+        try:
+            model.solutions.add_symbol_map(symbol_map)
+        except AttributeError:
+            # Something wacky happens in IDAES due to the usage of ScalarBlock
+            # instead of PyomoModel. This is an attempt to fix that.
+            from pyomo.core.base.PyomoModel import ModelSolutions
+
+            setattr(model, 'solutions', ModelSolutions(model))
+            model.solutions.add_symbol_map(symbol_map)
```
blnicho, line 513: This seems weird to include here. Is this pointing to an edge case that was missed in #110 that should be fixed directly in the `Block` or `_BlockData` class?

Robbybp:
> - Allow forward compatibility of solver_options and writer_config in LegacySolverWrapper
> - Yell loudly if users try to use both options and solver_options in the same command
> - Add some docs (not a lot, but some) about how both types of option passing are supported

Thanks! I too was confused by this when trying the new interface a few days ago.

pyomo/contrib/solver/base.py
mrmundt: Absolutely, yes, it should be. It was something that Andrew ran into while trying to get this working for IDAES, since they inherit from `ScalarBlock`

pyomo/contrib/solver/base.py
mrmundt: It was so a user could do both of these things:
```
solver.options['blah'] = whatever
solver.solver_options['foo'] = 'bar'
```
I don't really care but it seemed logical?

blnicho (approved):

pyomo/contrib/solver/base.py
andrewlee94: That said, this is part of what caused the troubles I ran into. With both of these being present, it was difficult to know which one to assign values to and how each would be used.

pyomo/contrib/solver/base.py
```diff
@@ -435,9 +466,13 @@ def _map_results(self, model, results):
...
-        legacy_results.problem.number_of_constraints = model.nconstraints()
-        legacy_results.problem.number_of_variables = model.nvariables()
-        number_of_objectives = model.nobjectives()
+        legacy_results.problem.number_of_constraints = len(
+            list(model.component_map(ctype=Constraint))
+        )
+        legacy_results.problem.number_of_variables = len(
+            list(model.component_map(ctype=Var))
+        )
+        number_of_objectives = len(list(model.component_map(ctype=Objective)))
```
jsiirola, line 475: These are almost certainly *not* the correct numbers (you are returning the number of component containers on the root block, and not the number of ComponentData objects on the entire tree).  Worse, tabulating "less incorrect" information is going to be expensive.  Do we really need to provide this in the results, or for the LegacySolverWrapper can we just set them to `NaN`?

pyomo/contrib/solver/base.py
```diff
@@ -393,8 +406,26 @@ def _map_config(
             self.config.time_limit = timelimit
         if report_timing is not NOTSET:
             self.config.report_timing = report_timing
-        if options is not NOTSET:
+        if hasattr(self, 'options'):
+            self.config.solver_options.set_value(self.options)
```
jsiirola, line 410: Looking for a "possibly defined" attribute seems weird.  I think I would prefer always defining a reasonable attribute (either None, or an empty dict or an empty`Bunch`), and if the container is not empty then pass it to the ConfigDict.

pyomo/contrib/solver/base.py
jsiirola: This is really wacky - and probably *not* something that should be added to Block (in the new world order, there should never be a reason to attach a symbolmap to the model)

pyomo/contrib/solver/base.py
mrmundt: I don't mind setting them to `NaN` but that's up to you all. Is this something that is promised in the book? Also, it is printed out when you run `model.pprint()` which is why I put this in here to begin with.

andrewlee94: Just speaking personally, I generally would not look to the solver results object for number of variables, constraints and objectives. I would either look at my original model (for the full model) or the solver logs (for what the solver saw).

That said, this was before we had a linear presolve, which can change the number of variables and constraints.

mrmundt: So I am... not sure how to handle this now. Suggestion?

jsiirola (approved):

PR #3224, opened 2024-04-03
Redefine objective sense as a proper `IntEnum`

### LINKED ISSUE #993: Conflicting definitions of minimize
There are multiple definitions of "minimize" that are used in Pyomo. Crucially, we run into this issue:

```
>>> from pyomo.environ import minimize
>>> from pyomo.opt import ProblemSense
>>> ProblemSense.minimize == minimize
False
```

The `minimize` from `pyomo.environ` is being imported from `pyomo.core.kernel.objective`.

This can obviously cause some issues when checking objective sense.

jsiirola:
## Summary/Motivation:
There have been some longstanding inconsistencies regarding how we manage the definition of "objective sense" in Pyomo.  The results object uses an Enum with 3 values (minimize, maximize, and unknown).  In the modeling side, we make heavy use of two named constants (`minimize` and `maximize`).  These constants are just `int`s (1 and -1, respectively), and those constants are defined in `pyomo.core.kernel.objective` (causing one of the few inter-linkages between kernel and the AML).

This PR defines a new Enum (`pyomo.common.enums.ObjectiveSense`), and redefines the `minimize` and `maximize` constants to be the members of that Enum.  To preserve consistency / backwards compatibility, `ObjectiveSense` is an `IntEnum`, so the old constants / old usage does not need to change.  

`ProblemSense` is now an extension of `ObjectiveSense` that also admits `unknown`.  Unfortunately, due to the way that the old solvers are structured, it is not (easy) to remove the use of `unknown` when processing results from the old solver interfaces.  The compromise here is the development of `ExtendedEnumType`, which allows us to define extended (or derived) Enums that combine the enums from another class with additional locally-defined members - so the `minimize` and `maximize` members in `ProblemSense` are actually members of `ObjectiveSense`.

- Create `ExtendedEnumType` metaclass to support extended/derived Enums
- Add `ObjectiveSense` enum (with `minimize` and `maximize` members)
- Redefine `ProblemSense` to be an extension of `ObjectiveSense`
- Remove references to `ProblemSense` from almost all of Pyomo
- Fix an error in processing the lower / upper bound in the SCIPAMPL interface

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/tests/test_enums.py
+++ b/pyomo/common/tests/test_enums.py
@@ -57,9 +57,7 @@ class TestExtendedEnumType(unittest.TestCase):
         self.assertIs(ProblemSense('minimize'), ObjectiveSense.minimize)
         self.assertIs(ProblemSense('maximize'), ObjectiveSense.maximize)
 
-        with self.assertRaisesRegex(
-                ValueError, "'foo' is not a valid ProblemSense"
-        ):
+        with self.assertRaisesRegex(ValueError, "'foo' is not a valid ProblemSense"):
             ProblemSense('foo')
 
     def test_contains(self):
@@ -71,11 +69,11 @@ class TestExtendedEnumType(unittest.TestCase):
         self.assertIn(ProblemSense.minimize, ObjectiveSense)
         self.assertIn(ProblemSense.maximize, ObjectiveSense)
 
+
 class TestObjectiveSense(unittest.TestCase):
     def test_members(self):
         self.assertEqual(
-            list(ObjectiveSense),
-            [ObjectiveSense.minimize, ObjectiveSense.maximize],
+            list(ObjectiveSense), [ObjectiveSense.minimize, ObjectiveSense.maximize]
         )
 
     def test_hasattr(self):
@@ -89,9 +87,7 @@ class TestObjectiveSense(unittest.TestCase):
         self.assertIs(ObjectiveSense('minimize'), ObjectiveSense.minimize)
         self.assertIs(ObjectiveSense('maximize'), ObjectiveSense.maximize)
 
-        with self.assertRaisesRegex(
-                ValueError, "'foo' is not a valid ObjectiveSense"
-        ):
+        with self.assertRaisesRegex(ValueError, "'foo' is not a valid ObjectiveSense"):
             ObjectiveSense('foo')
 
     def test_str(self):
--- a/pyomo/common/tests/test_enums.py
+++ b/pyomo/common/tests/test_enums.py
@@ -1,0 +1,99 @@
+#  ___________________________________________________________________________
+#
+#  Pyomo: Python Optimization Modeling Objects
+#  Copyright (c) 2008-2024
+#  National Technology and Engineering Solutions of Sandia, LLC
+#  Under the terms of Contract DE-NA0003525 with National Technology and
+#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain
+#  rights in this software.
+#  This software is distributed under the 3-clause BSD License.
+#  ___________________________________________________________________________
+
+import enum
+
+import pyomo.common.unittest as unittest
+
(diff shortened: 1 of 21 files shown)


pyomo/solvers/tests/mip/test_scip_solve_from_instance.baseline
```diff
@@ -1,7 +1,7 @@
 {
     "Problem": [
         {
-            "Lower bound": -Infinity,
+            "Lower bound": 1.0,
```
mrmundt, line 4: How is this related? If not, no biggie, just curious.

pyomo/common/tests/test_enums.py
```diff
@@ -0,0 +1,95 @@
...
+        self.assertTrue(hasattr(ProblemSense, 'maximize'))
+
+    def test_call(self):
+        self.assertIs(ObjectiveSense(1), ObjectiveSense.minimize)
+        self.assertIs(ObjectiveSense(-1), ObjectiveSense.maximize)
+
+        self.assertIs(ObjectiveSense('minimize'), ObjectiveSense.minimize)
+        self.assertIs(ObjectiveSense('maximize'), ObjectiveSense.maximize)
+
+        with self.assertRaisesRegex(ValueError, "'foo' is not a valid ObjectiveSense"):
```
mrmundt, line 90: It may be worth adding another check here for if someone tries to use something not in the set of [-1, 0, 1].

pyomo/solvers/tests/mip/test_scip_solve_from_instance.baseline
jsiirola: It's not - apart from as I was removing use of `ProblemSense` I ran across this bug in the SCIP interface where we always reported the incumbent objective value as the upper bound (regardless of the optimization sense).  I fixed that (instead of preserving the bug and fixing it in a separate PR) - and then promptly realized that we never returned the opposing (dual) bound, and that we should.  So I fixed that, too.  That latter fix caused the change in the baseline.

mrmundt (approved):

pyomo/solvers/plugins/solvers/SCIPAMPL.py
```diff
@@ -374,9 +369,11 @@ def _postsolve(self):
             if len(results.solution) > 0:
                 results.solution(0).status = SolutionStatus.optimal
             try:
-                if results.problem.sense == ProblemSense.minimize:
+                if results.solver.primal_bound < results.solver.dual_bound:
                     results.problem.lower_bound = results.solver.primal_bound
+                    results.problem.upper_bound = results.solver.dual_bound
                 else:
+                    results.problem.lower_bound = results.solver.dual_bound
                     results.problem.upper_bound = results.solver.primal_bound
```
emma58, line 377: Can this go wrong if your bounds cross by a little bit?

emma58 (approved):
I like this! I have one question about the SCIP change below--just wondering if you could fall victim to numerical issues.

pyomo/solvers/plugins/solvers/SCIPAMPL.py
jsiirola: Yes - you probably can.  This will be easily fixable in the new solver interfaces (where the results parser can have easier access to the original model and won't have to guess about the sense), but there isn't an obvious better solution here (unless the scip output log prints something about the sense - but I really don't want to put more dependence on the parsing of output logs).

pyomo/opt/results/problem.py
```diff
@@ -12,19 +12,17 @@
 import enum
 from pyomo.opt.results.container import MapContainer
 
+from pyomo.common.deprecation import deprecated, deprecation_warning
```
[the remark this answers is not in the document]
jsiirola: Nope.  Removed.

blnicho (approved):

PR #3241, opened 2024-04-18
contrib.preprocessing: Fixing bug where variable aggregator did not intersect domains

emma58:
## Summary/Motivation:

The `contrib.aggregate_vars` transformation correctly intersected bounds of variables that are linked by equality constraints, but it did not handle domains, meaning that in a model with two binaries linked by an equality transformation, the transformed model was actually a relaxation since integrality for both of the variables was relaxed. This PR fixes this bug by setting the domain of the variable that will be substituted in to the intersection of the domains of all the variables linked by equality.

- Fixes bug
- Adds a couple tests

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/preprocessing/plugins/var_aggregator.py
+++ b/pyomo/contrib/preprocessing/plugins/var_aggregator.py
@@ -13,7 +13,8 @@
 
 
 from pyomo.common.collections import ComponentMap, ComponentSet
-from pyomo.core.base import Block, Constraint, VarList, Objective, TransformationFactory
+from pyomo.core.base import (Block, Constraint, VarList, Objective, Reals,
+                             TransformationFactory)
 from pyomo.core.expr import ExpressionReplacementVisitor
 from pyomo.core.expr.numvalue import value
 from pyomo.core.plugins.transform.hierarchy import IsomorphicTransformation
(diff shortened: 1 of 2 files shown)


pyomo/contrib/preprocessing/plugins/var_aggregator.py
```diff
@@ -248,6 +255,12 @@ def _apply_to(self, model, detect_fixed_vars=True):
             # the variables in its equality set.
             z_agg.setlb(max_if_not_None(v.lb for v in eq_set if v.has_lb()))
             z_agg.setub(min_if_not_None(v.ub for v in eq_set if v.has_ub()))
+            # Set the domain of the aggregate variable to the intersection of
+            # the domains of the variables in its equality set
+            domain = Reals
+            for v in eq_set:
+                domain = domain & v.domain
```
jsiirola, line 262: This isn't particularly efficient, but it is clean and correct.

jsiirola (approved):
This is a clean way to ensure the correct behavior.  However, it is not particularly efficient, as it will generally lead to an expression of `Reals & v1.domain & v2.domain`.  The first `Reals` is unnecessary, and there are no checks that `v1.domain != v2.domain`.  That said, this particular transformation is not at all about efficiency, so it is probably fine that it trades efficiency for clarity.

pyomo/contrib/preprocessing/plugins/var_aggregator.py
emma58: OK, this makes sense. I think I'll merge as-is because if we ever want more efficiency, we will want to do a rewrite anyway, I think.

PR #3250, opened 2024-04-26
Pyomo.DoE adds checking sentences to the type of objects

jialuw96:
## Fixes # .

No issue

## Summary/Motivation:

In Pyomo.DoE, check if an object is `Var` or `Param` before proceeding to fix it or set bounds to it. 

- Check if an object is `Var` before setting upper and lower bounds to it. If it is not a type of `Var`, no need to fix it.  
- Check if an object is `Var` before fixing it to a given value. If it is one type of `Param`, check if it is mutable before changing it value. 

### Legal Acknowledgement


jialuw96:
@adowling2 This PR is ready for your review. Thank you!

pyomo/contrib/doe/doe.py
```diff
@@ -598,7 +605,29 @@ def block_build(b, s):
...
+                        pcb.var.Var,
+                        pcb.var.IndexedVar,
+                        pcb.var.ScalarVar,
+                        pcb.var.SimpleVar,
+                    ),
+                ):
+                    var.fix(self.scenario_data.scenario[s][par])
+                # if it is a param, give it a new value
+                elif isinstance(
+                    var, (pcb.param.ScalrParam, pcb.param.IndexedParam, pcb.param.Param)
```
mrmundt, line 622: I believe this is a typo - should be `ScalarParam`, yes?

pyomo/contrib/doe/doe.py
```diff
@@ -598,7 +605,30 @@ def block_build(b, s):
...
+                    var,
+                    (pcb.param.ScalarParam, pcb.param.IndexedParam, pcb.param.Param),
+                ):
+                    # check if param is mutable
+                    if var.mutable:
+                        var = self.scenario_data.scenario[s][par]
+                    else:
+                        raise TypeError(
+                            "If defined as a Param, parameters should be mutable by mutable=True"
+                        )
```
mrmundt, line 631: It would be good to add a test to ensure the behavior is correct.

mrmundt (changes requested):

jialuw96:
The unit test is added in the following file: 
- `contrib/doe/examples/reactor_compute_FIM.py`

We added the unit test where in the kinetics example, the design variables are defined as `Param` instead of `Var`, to test if the checking sentences can identify the type of the objects and fix them to another value.

mrmundt (approved):

pyomo/contrib/doe/doe.py
```diff
@@ -494,8 +496,11 @@ def _direct_kaug(self):
         for par in self.param.keys():
             cuid = pyo.ComponentUID(par)
             var = cuid.find_component_on(mod)
-            var.setlb(self.param[par])
-            var.setub(self.param[par])
+
+            # only set up bounds if they are variables
+            if isinstance(var, (pcb.var.Var, pcb.var.IndexedVar, pcb.var.ScalarVar)):
+                var.setlb(self.param[par])
+                var.setub(self.param[par])
```
blnicho, line 503: Why are you setting the bounds to be equal instead of just fixing the variable?

pyomo/contrib/doe/doe.py
```diff
@@ -38,6 +38,8 @@
 from pyomo.contrib.sensitivity_toolbox.sens import get_dsdp
 from pyomo.contrib.doe.scenario import ScenarioGenerator, FiniteDifferenceStep
 from pyomo.contrib.doe.result import FisherResults, GridSearchResult
+
+import pyomo.core.base as pcb
```
blnicho, line 42:
```suggestion
from pyomo.environ import Var, Param
```

pyomo/contrib/doe/doe.py
```diff
@@ -494,8 +496,11 @@ def _direct_kaug(self):
         for par in self.param.keys():
             cuid = pyo.ComponentUID(par)
             var = cuid.find_component_on(mod)
-            var.setlb(self.param[par])
-            var.setub(self.param[par])
+
+            # only set up bounds if they are variables
+            if isinstance(var, (pcb.var.Var, pcb.var.IndexedVar, pcb.var.ScalarVar)):
```
blnicho, line 501:
```suggestion
            if var.ctype is Var:
```

pyomo/contrib/doe/doe.py
```diff
@@ -632,7 +637,24 @@ def block_build(b, s):
             for par in self.param:
                 cuid = pyo.ComponentUID(par)
                 var = cuid.find_component_on(b)
-                var.fix(self.scenario_data.scenario[s][par])
+
+                # if it is a variable, fix it to a new value
+                if isinstance(
+                    var, (pcb.var.Var, pcb.var.IndexedVar, pcb.var.ScalarVar)
+                ):
```
blnicho, line 644:
```suggestion
                if var.ctype is Var:
```

pyomo/contrib/doe/doe.py
```diff
@@ -632,7 +637,24 @@ def block_build(b, s):
...
+                    var.fix(self.scenario_data.scenario[s][par])
+                # if it is a param, give it a new value
+                elif isinstance(
+                    var,
+                    (pcb.param.ScalarParam, pcb.param.IndexedParam, pcb.param.Param),
+                ):
```
blnicho, line 650:
```suggestion
                elif var.ctype is Param:
```

pyomo/contrib/doe/doe.py
```diff
@@ -1190,18 +1212,36 @@ def _fix_design(self, m, design_val, fix_opt=True, optimize_option=None):
...
+
+            if isinstance(
+                var,
+                (
+                    pcb.var.Var,
+                    pcb.var.IndexedVar,
+                    pcb.var.ScalarVar,
+                    pcb.var._GeneralVarData,
+                ),
+            ):
```
blnicho, line 1224:
```suggestion
            if var.ctype is Var:
```

pyomo/contrib/doe/doe.py
```diff
@@ -1190,18 +1212,36 @@ def _fix_design(self, m, design_val, fix_opt=True, optimize_option=None):
...
-                    if optimize_option[name]:
+                    if optimize_option is None:
                         var.unfix()
+                    else:
+                        if optimize_option[name]:
+                            var.unfix()
+
+            elif isinstance(
+                var, (pcb.param.ScalarParam, pcb.param.IndexedParam, pcb.param.Param)
+            ):
```
blnicho, line 1236:
```suggestion
            elif var.ctype is Param:
```

blnicho (changes requested):

pyomo/contrib/doe/doe.py
jialuw96: I change these sentences to fixing the variable.

pyomo/contrib/doe/doe.py
```diff
@@ -1190,18 +1200,20 @@ def _fix_design(self, m, design_val, fix_opt=True, optimize_option=None):
...
-                # Otherwise check optimize_option
-                if optimize_option is None:
-                    # If optimize_option is None, unfix all design variables
-                    var.unfix()
+
+            if var.ctype is Var:
+                if fix_opt:
+                    var.fix(design_val[name])
                 else:
-                    # Otherwise, unfix only the design variables listed in optimize_option with value True
```
blnicho, line 1202: Why did you remove the comments explaining the logic here?

pyomo/contrib/doe/examples/reactor_compute_FIM.py
```diff
@@ -106,6 +110,24 @@ def main():
...
+        discretize_model=disc_for_measure,  # discretize model function
+    )
+
+    result2 = doe_object2.compute_FIM(
+        mode="direct_kaug",  # calculation mode
+        scale_nominal_param_value=True,  # scale nominal parameter value
+        formula="central",  # formula for finite difference
+    )
+
+    result2.result_analysis()
```
blnicho, line 129: Are you expecting the results to be the same as the previous case? If so I think you should have the same assert statements checking the trace and determinant.

pyomo/contrib/doe/doe.py
```diff
@@ -632,7 +636,13 @@ def block_build(b, s):
             for par in self.param:
                 cuid = pyo.ComponentUID(par)
                 var = cuid.find_component_on(b)
```
blnicho, line 638: There is another copy of this code starting on line 612. Does it need to be updated also?

pyomo/contrib/doe/doe.py
```diff
@@ -1190,18 +1200,20 @@ def _fix_design(self, m, design_val, fix_opt=True, optimize_option=None):
...
-                    # Otherwise, unfix only the design variables listed in optimize_option with value True
-                    if optimize_option[name]:
+                    if optimize_option is None:
                         var.unfix()
+                    else:
+                        if optimize_option[name]:
+                            var.unfix()
+
+            elif var.ctype is Param:
+                var = design_val[name]
```
blnicho, line 1215: What if this is called with `fix_opt=False`? Are there cases where the user expects var to be a degree of freedom and they would be surprised by this code silently leaving it as fixed? If this is expected behavior then I think it needs to be clear in the documentation what the tradeoff is when using a design variable declared as a Pyomo Param vs a Pyomo Var.

blnicho (changes requested):
@jialuw96 could you explain the motivation behind this PR? It looks like the Pyomo.DoE API was meant to support design variables specified as either Pyomo mutable Param or Var components but there were a few bugs with the Param case. Is that right?

The docstrings for the `DesignOfExperiments` and `DesignVariables` classes should also be updated. Right now they don't mention that both Param and Var components are accepted: 
`design_vars – A DesignVariables which contains the Pyomo variable names and their corresponding indices and bounds for experiment degrees of freedom`

pyomo/contrib/doe/doe.py
jialuw96: This is added back

pyomo/contrib/doe/doe.py
jialuw96: Yes, they are updated now

pyomo/contrib/doe/doe.py
jialuw96: if with `fix_opt=False`, var will be unfixed, so it is fine if the user wants to leave it as DOF (they can use fix_opt=False to do that). I add a statement in the doc to note that, if a design variable is defined as `Param`, `stochastic_program` function will not work because it cannot be optimized.

pyomo/contrib/doe/examples/reactor_compute_FIM.py
jialuw96: I modified this test to use both `Param` components for parameters and variables, change one parameter value, use assertions to check if the generated model uses the updated model parameter value, and check the trace and determinant.

jialuw96:
> @jialuw96 could you explain the motivation behind this PR? It looks like the Pyomo.DoE API was meant to support design variables specified as either Pyomo mutable Param or Var components but there were a few bugs with the Param case. Is that right?

The current Pyomo.DoE API can only support design variables and to-be-estimated parameters as `Var` components; This PR allows the user to specify them as either `Var` or `Param` components. 

> 
> The docstrings for the `DesignOfExperiments` and `DesignVariables` classes should also be updated. Right now they don't mention that both Param and Var components are accepted: `design_vars – A DesignVariables which contains the Pyomo variable names and their corresponding indices and bounds for experiment degrees of freedom`

This is updated. Thank you!

adowling2:
Decision: @adowling2 will work on resolving conflicts and extensive testing after #3267 is merged. This current PR is NOT critical for the Pyomo.DoE workshop next week.

blnicho:
These changes are being superseded by #3317

PR #3260, opened 2024-05-14
contrib.solver.highs

michaelbynum:
- Port the HiGHS interface from appsi to contrib.solver
- Move the `treat_fixed_vars_as_params` option from a config option to a private attribute that can only be set at construction time. The reason for this is that changing the value of this parameter between solves can result in incorrect results.
- Make it easier to change the "active" config object so the correct options get used in other parts of the class.

### Legal Acknowledgement


mrmundt:
@michaelbynum - I am going to temporarily close this while I poke at it to keep computational cycles free. Don't panic!

PR #3268, opened 2024-05-21
Adding ParameterizedLinearRepn and corresponding walker

emma58:
## Summary/Motivation:

With eventual bilevel programming applications in mind, this PR adds a `ParameterizedLinearRepn` object that has the same structure as `LinearRepn` but allows for an argument `wrt` to specify a list of Vars that will be treated as data for the purposes of creating the repn.

- Adds `ParameterizedLinearRepn` component that inherits from `LinearRepn` but has to re-implement everything that assumes that constants and coefficients are numbers rather than expressions.
- Adds `ParameterizedLinearRepnVisitor` that inherits from `LinearRepnVisitor` but adds a "pseudo-constant" expression type and new exitNode handlers for psuedo-constants (i.e., Pyomo expressions that we are treating as constant *mathematically*)
- Adds handling for nan in `assertExpressionsEqual`

### Legal Acknowledgement


pyomo/core/expr/compare.py
```diff
@@ -230,10 +230,14 @@ def assertExpressionsEqual(test, a, b, include_named_exprs=True, places=None):
...
-                test.assertEqual(_a, _b)
+            # If _a is nan, check _b is nan
+            if _a != _a:
+                test.assertTrue(_b != _b)
             else:
-                test.assertAlmostEqual(_a, _b, places=places)
+                if places is None:
+                    test.assertEqual(_a, _b)
+                else:
+                    test.assertAlmostEqual(_a, _b, places=places)
```
jsiirola, line 240: Instead of this, should we just have `places` default to the default value in `assertAlmostEqual`?

pyomo/repn/parameterized_linear.py
```diff
@@ -0,0 +1,400 @@
...
+    LinearRepn,
+    LinearRepnVisitor,
+)
+from pyomo.repn.util import ExprType
+from . import linear
+
+
+class ParameterizedExprType(enum.IntEnum, metaclass=ExtendedEnumType):
+    __base_enum__ = ExprType
+    PSEUDO_CONSTANT = 50
```
jsiirola, line 45: Shouldn't this be `5` (something between 0 and 10)?

pyomo/repn/parameterized_linear.py
```diff
@@ -0,0 +1,400 @@
...
+)
+from pyomo.repn.linear import (
+    ExitNodeDispatcher,
+    _initialize_exit_node_dispatcher,
+    LinearBeforeChildDispatcher,
+    LinearRepn,
+    LinearRepnVisitor,
+)
+from pyomo.repn.util import ExprType
+from . import linear
```
jsiirola, line 40: This is a strange mix of relative and absolute imports.  Maybe `import pyomo.repn.linear as linear`?

pyomo/repn/parameterized_linear.py
```diff
@@ -0,0 +1,400 @@
...
+            ans *= self.multiplier
+        return ans
+
+    def append(self, other):
+        """Append a child result from acceptChildResult
+
+        Notes
+        -----
+        This method assumes that the operator was "+". It is implemented
+        so that we can directly use a LinearRepn() as a `data` object in
```
jsiirola, line 126: `LinearRepn` -> `ParameterizedLinearRepn`

pyomo/repn/parameterized_linear.py
```diff
@@ -0,0 +1,400 @@
...
+        the expression walker (thereby allowing us to use the default
+        implementation of acceptChildResult [which calls
+        `data.append()`] and avoid the function call for a custom
+        callback).
+
+        """
+        _type, other = other
+        if _type in (_CONSTANT, _PSEUDO_CONSTANT):
```
jsiirola, line 134: This is slower than necessary (it creates a tuple, does a linear scan, and throws it away each time).  Maybe
```
if _type is _CONSTANT or _type is _PSEUDO_CONSTANT:
```

pyomo/repn/parameterized_linear.py
```diff
@@ -0,0 +1,400 @@
...
+)
+
+
+#
+# PRODUCT handlers
+#
+
+
+def _handle_product_constant_constant(visitor, node, arg1, arg2):
+    return _CONSTANT, arg1[1] * arg2[1]
```
jsiirola, line 238: Why did you need to override this?

jsiirola (commented):
Overall this looks good!  Just a few questions...

pyomo/core/expr/compare.py
emma58: There are a few tests in `contrib.piecewise` where I had to reduce it because of numerical stuff. So that's why there's an argument for it.

pyomo/repn/parameterized_linear.py
emma58: Sure!

pyomo/repn/parameterized_linear.py
emma58: Thanks, good catch

pyomo/repn/parameterized_linear.py
emma58: Because I didn't include the deprecation path for `0 * nan`. I'll add a comment because it can get un-overridden when that deprecation path goes away.

jsiirola (approved):

jsiirola (approved):

PR #3275, opened 2024-05-28
Handle uninitialized variable in `propagate_solution` of scaling transformation

### LINKED ISSUE #2817: Propagate State from scaling tools throws an error if any variable is uninitialized
## Summary

`propagate_state` from Pyomo's scaling tools cannot handle if any variables in the model are uninitialized. 

In the original IDAES model this was encountered in (the `ControlVolume1D`), we have unpaired material and energy difference terms along either the x=0 or x=1 boundary (depending on the discretization direction).  Presently, we use the fact those terms are uninitialized in initialization methods to determine whether the user has provided an initial guess or not, so just initializing them at zero isn't an immediate solution.

Minimum example:

```
import pyomo.environ as pyo

m = pyo.ConcreteModel()
m.x = pyo.Var()
m.y = pyo.Var()
m.z = pyo.Var()

m.x.value = 1
m.y.value = 1

@m.Constraint()
def eqn1(b):
    return 10 * b.y == b.x

m.scaling_factor[m.x] = 0.1
m.scaling_factor[m.eqn1] = 0.1

pyomo_scaling = pyo.TransformationFactory('core.scale_model')

m_scaled = pyomo_scaling.create_using(m, rename=False)

pyomo_scaling.propagate_solution(m_scaled, m)
```

### Error Message
```console
ERROR: evaluating object as numeric value: z
        (object: <class 'pyomo.core.base.var.ScalarVar'>)
    No value for uninitialized NumericValue object z
ValueError                                Traceback (most recent call last)
Cell In[3], line 26
     21 m_scaled = pyomo_scaling.create_using(m, rename=False)
     23 # solver_obj = pyo.SolverFactory('ipopt')
     24 # solver_obj.solve(m_scaled, tee=True)
---> 26 pyomo_scaling.propagate_solution(m_scaled, m)

File ~\.conda\envs\idaes-newer\lib\site-packages\pyomo\core\plugins\transform\scaling.py:360, in ScaleModel.propagate_solution(self, scaled_model, original_model)
    356 original_v = original_model.find_component(original_v_path)
    358 for k in scaled_v:
    359     original_v[k].set_value(
--> 360         value(scaled_v[k]) / component_scaling_factor_map[scaled_v[k]],
    361         skip_validation=True,
    362     )
    363     if check_reduced_costs and scaled_v[k] in scaled_model.rc:
    364         original_model.rc[original_v[k]] = scaled_model.rc[scaled_v[k]] * component_scaling_factor_map[
    365             scaled_v[k]] / objective_scaling_factor

File ~\.conda\envs\idaes-newer\lib\site-packages\pyomo\core\expr\numvalue.py:140, in value(obj, exception)
    138     tmp = obj(exception=True)
    139     if tmp is None:
--> 140         raise ValueError(
    141             "No value for uninitialized NumericValue object %s"
    142             % (obj.name,))
    143     return tmp
    144 except TemplateExpressionError:
    145     # Template expressions work by catching this error type. So
    146     # we should defer this error handling and not log an error
    147     # message.

ValueError: No value for uninitialized NumericValue object z
```

Pyomo version 6.5.1a0

### LINKED ISSUE #3273: Unscaling model after solving using propagate_solution(scaled_model, original_model) throws error
## Summary

I am trying to scale, solve, and unscale a pyomo model. The code itself worked well until I tried to scale but it didn't get a solution probably due to scaling issues. Now, after scaling, the model solves but I get an error when I try to scale it back. Unfortunately, I couldn't find any application in the doc on how to actually use the ```propagate_solution''' method.

I will try to provide a few lines of code to summarize what I'm doing without making it extensively long. The last line is what gives me the error

```console


```

```
# example.py
from pyomo.environ import *
from pyomo.dae import *

model = ConcreteModel()

model.t = ContinuousSet(bounds=(0, 5))
model.x = ContinuousSet(bounds=(0, 3))
model.y = Var(model.t, model.x)
model.z = Var(model.t, model.x)
model.dydt = DerivativeVar(model.y, wrt=model.t)
model.dzdt = DerivativeVar(model.z, wrt=model.t)
model.dydx = DerivativeVar(model.y, wrt=model.x)
model.dzdx = DerivativeVar(model.z, wrt=model.x)

2 differential equations here in y and z
2 algebraic equations here in y and z

Discretize both y and z

# create the scaling factors
model.scaling_factor = Suffix(direction=Suffix.EXPORT)
model.scaling_factor[model.diffeq1] = 10 # scale diffeq1
model.scaling_factor[model.z] = 0.1 # scale z
# transform the model
scaled_model = TransformationFactory('core.scale_model').create_using(model)

solver = SolverFactory('ipopt')
results = solver.solve(scaled_model, tee=True)

model = TransformationFactory('core.scale_model').propagate_solution(scaled_model, model)
```

### Error Message

```console
$ # ERROR: evaluating object as numeric value: scaled_dydt[0,0]
        (object: <class 'pyomo.core.base.var._GeneralVarData'>)
    No value for uninitialized NumericValue object scaled_dydt[0,0]
  File "/Users/$USER/filename", line 291, in <module>
    model = TransformationFactory('core.scale_model').propagate_solution(scaled_model, model)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/$USER/anaconda3/envs/lib/python3.12/site-packages/pyomo/core/plugins/transform/scaling.py", line 326, in propagate_solution
    value(scaled_v[k]) / component_scaling_factor_map[scaled_v[k]],
    ^^^^^^^^^^^^^^^^^^
  File "/Users/$USER/anaconda3/envs/watertap-dev/lib/python3.12/site-packages/pyomo/common/numeric_types.py", line 318, in value
    raise ValueError(
ValueError: No value for uninitialized NumericValue object scaled_dydt[0,0]
```

Pyomo version: 6.7.1
Python version: 3.12
Operating system: Mac OS Sonoma
How Pyomo was installed (PyPI, conda, source): conda
Solver (if applicable): ipopt

Robbybp:
- If a variable has a value of `None` in the scaled model, we don't attempt to set its value in the original model

### Legal Acknowledgement


pyomo/core/plugins/transform/scaling.py
```diff
@@ -313,10 +313,13 @@ def propagate_solution(self, scaled_model, original_model):
             original_v = original_model.find_component(original_v_path)
 
             for k in scaled_v:
-                original_v[k].set_value(
-                    value(scaled_v[k]) / component_scaling_factor_map[scaled_v[k]],
-                    skip_validation=True,
-                )
+                if scaled_v[k].value is not None:
```
jsiirola, line 316: Is the correct check here to test `None`, or to not transfer the values of any `stale` variables?  I am wondering if the correct answer is to only propagate variable values when `.stale==False`

Robbybp: I guess I can see three options:
1. Propagate values in the model, unless they are None (what this PR currently does)
2. Propagate values in the model, regardless of what they are
3. Propagate non-stale values in the model

My instinct is that relying on the stale flag can lead to non-intuitive behavior, e.g. if a user sets a value after a solve (or solves via `solve_strongly_connected_components`), only the last values updated will be propagated. If this is what we decide on, the docstring should be updated to indicate that only the most recent solution is propagated.

Robbybp:
This should fix #2817 as well.

pyomo/core/plugins/transform/scaling.py
Robbybp: Based on @dallan-keylogic's comment on #2817, I've updated this PR to propagate values from the scaled model even if they are None, but to log a warning if None overrides a non-None value.

pyomo/core/tests/transform/test_scaling.py
```diff
@@ -690,6 +690,23 @@ def test_get_float_scaling_factor_intermediate_level(self):
...
+        m.scaling_factor[m.x[2]] = 10.0
+        scaled_model = pyo.TransformationFactory("core.scale_model").create_using(m)
+        scaled_model.scaled_x[1] = 20.0
+        scaled_model.scaled_x[2] = None
+        pyo.TransformationFactory("core.scale_model").propagate_solution(
+            scaled_model, m
+        )
+        self.assertAlmostEqual(m.x[1].value, 2.0, delta=1e-8)
+        # Note that value of x[2] in original model *has* been overriddeen to None.
+        # In this case, a warning has been raised.
```
blnicho, line 707: Could you add a test for the warning?

Robbybp: Added.

blnicho (approved):

pyomo/core/plugins/transform/scaling.py
jsiirola: I think I am OK with this behavior, but if we are always going to propagate the entire model state, I wonder if a better name for this method is `propagate_state()` (as we are propagating the current state of the model and not necessarily the last thing that came back from the solver).

pyomo/core/tests/transform/test_scaling.py
```diff
@@ -690,6 +692,27 @@ def test_get_float_scaling_factor_intermediate_level(self):
...
+        scaled_model = pyo.TransformationFactory("core.scale_model").create_using(m)
+        scaled_model.scaled_x[1] = 20.0
+        scaled_model.scaled_x[2] = None
+
+        OUTPUT = io.StringIO()
+        with LoggingIntercept(OUTPUT, "pyomo.core.plugins.transform.scaling"):
+            pyo.TransformationFactory("core.scale_model").propagate_solution(
+                scaled_model, m
+            )
+        self.assertIn("replacing value of variable", OUTPUT.getvalue())
```
jsiirola, line 710: It would be nice to test against more of the logged message (maybe the entire string, as that would also test that the message was only logged once)

jsiirola (approved):
I have a minor comment that would improve testing, but that shouldn't necessarily prevent merging.

pyomo/core/plugins/transform/scaling.py
Robbybp: I agree that `propagate_state`, or something like `propagate_variable_values`, would be a better name, although I'm not convinced it's worth changing.

PR #3281, opened 2024-06-02
Sort indices while removing constraints to fix #3127

### LINKED ISSUE #3127: HiGHS solver does not update the model correctly after variable fix/unfix
## Summary

HiGHS solver does not update the model correctly after a sequence of variable fix/unfix. This only occurs when the treat_fixed_vars_as_params flag is True (the default option).

```python
import pyomo.environ as pyo
from pyomo.contrib import appsi

m = pyo.ConcreteModel()
m.x = pyo.Var(domain = pyo.Binary)
m.y = pyo.Var(domain = pyo.Binary)
m.fx = pyo.Var(domain = pyo.NonNegativeReals)
m.fy = pyo.Var(domain = pyo.NonNegativeReals)
m.c1 = pyo.Constraint(expr = m.fx <= m.x)
m.c2 = pyo.Constraint(expr = m.fy <= m.y)
m.c3 = pyo.Constraint(expr = m.x + m.y <= 1)

m.obj = pyo.Objective(
    expr = m.fx * 0.5 + m.fy * 0.4, 
    sense = pyo.maximize,
)

opt = appsi.solvers.Highs()
opt.config.load_solution = False
# opt.update_config.treat_fixed_vars_as_params = False

# solution 1 has m.x == 1 and m.y == 0
r = opt.solve(m)
print(r)
assert r.best_feasible_objective is not None

# solution 2 has m.x == 0 and m.y == 1
m.y.fix(1)
r = opt.solve(m)
print(r)
assert r.best_feasible_objective is not None

# solution 3 should be equal solution 1 but get infeasible
m.y.unfix()
m.x.fix(1)
r = opt.solve(m)
print(r)
assert r.best_feasible_objective is None
```

gives

```console
termination_condition: TerminationCondition.optimal
best_feasible_objective: 0.5
best_objective_bound: 0.5
termination_condition: TerminationCondition.optimal
best_feasible_objective: 0.4
best_objective_bound: 0.4
termination_condition: TerminationCondition.infeasible
best_feasible_objective: None
best_objective_bound: inf
```

The expected behavior is that in the 3rd solve the solution obtained is the same as the first solve. This behavior is only achieved if the treat_fixed_vars_as_params flag is set to False.

### Error Message

The highs log gives a hint : ERROR:   Set supplied to Highs::deleteRows is not ordered

Pyomo version: 6.7.0
Python version:  3.9.16
Operating system: Win 11
Solver (if applicable): highspy 1.5.3

kutlay:
## Summary/Motivation:
Issue 3127 showed that HiGHs solution was different when the same variable was fixed and unfixed. A solution should not change after fixing and unfixing the same variable. The issue showed the hint that HiGHs solver was expecting the deleted row indices to be sorted. This fix sorts the indices before sending it to the solver.

Unfortunately, HiGHs code doesn't explain why the indices supplied to deleteRows need to be sorted: https://github.com/ERGO-Code/HiGHS/blob/13363c9f1252b015cf6527132eb9cf8f4b5bf020/src/lp_data/Highs.cpp#L2871-L2883

However, the error message indicates that they need to be sorted, so I can't think of any problems sorting the indices in pyomo other than slight degradation in performance because of sorting.

- Sort indices before sending to the solver to be removed
- Test the fix / unfix functionality

### Legal Acknowledgement


pyomo/contrib/appsi/solvers/tests/test_highs_persistent.py
```diff
@@ -80,6 +80,43 @@ def test_mutable_params_with_remove_vars(self):
...
+            expr = m.fx * 0.5 + m.fy * 0.4, 
+            sense = pe.maximize,
+        )
+
+        opt = Highs()
+
+        # solution 1 has m.x == 1 and m.y == 0
+        r = opt.solve(m)
+        solution_1_objective = r.best_feasible_objective
+        assert r.best_feasible_objective is not None
```
jsiirola, line 105: As general practice, Pyomo makes use of unittest-style assertions in tests (so here, `self.assertIsNotNone(r.best_feasible_objective)`, or better yet, test the actual result returned by HiGHS (`self.assertAlmostEqual(r.best_feasible_objective, 0.5, places=5)`)

pyomo/contrib/appsi/solvers/tests/test_highs_persistent.py
```diff
@@ -80,6 +80,43 @@ def test_mutable_params_with_remove_vars(self):
...
+        m.c3 = pe.Constraint(expr = m.x + m.y <= 1)
+
+        m.obj = pe.Objective(
+            expr = m.fx * 0.5 + m.fy * 0.4, 
+            sense = pe.maximize,
+        )
+
+        opt = Highs()
+
+        # solution 1 has m.x == 1 and m.y == 0
```
jsiirola, line 102: It would be good to test the actual result returned by HiGHS here and after the next two solves.  Something like
```python
self.assertAlmostEqual(m.fx.value, 1, places=5)
self.assertAlmostEqual(m.fy.value, 0, places=5)
```

jsiirola (changes requested):
This looks pretty reasonable.  Thank you!

Two really minor updates to the test and I think this can be merged.

kutlay:
@jsiirola Thank you for the help! Changes are done and it should be good to go now.

jsiirola (approved):

mrmundt:
@kutlay  - can you please install the latest version of [`black`](https://pypi.org/project/black/) and run `black -S -C` on your code? Thank you!

kutlay:
@mrmundt sorry for the delay, the black reformatting is done. You can kick off the workflow again.

mrmundt:
@kutlay - FYI, the pypy job will fail, but it's not your fault.

blnicho (approved):

blnicho:
We're waiting for Jenkins to run on this before merging.

PR #3293, opened 2024-06-17
Constraint: only store the original expression (not lower/body/upper)

jsiirola:
## Summary/Motivation:
This PR changes the internal data store for Constraint to only store the original (relational) expression and not break it apart into (`lower`, `body`, and `upper`).  Those attributes are preserved as properties (and the expression rearrangement happens dynamically when they are called).  Most clients who need the expression "standardized" should use the new `normalize_constraint()` method (which is more efficient, as it avoids duplicate work and unnecessary calls to `as_numeric()`).

While this would appear to be strictly cosmetic, this change is the first part of the templatized LP writer.

- remove `_lower`, `_body`, and `_upper` ConstraintData attributes
- enforce the use of `.args` as the only public API for getting at an expression's arguments
- update persistent interfaces to look at the `expr` (and not `body` ) to detect changed expressions
- update FBBT to use the same relational expression handling logic as `compute_bounds_on_expression`

### Legal Acknowledgement


pyomo/core/base/constraint.py
```diff
@@ -317,15 +337,22 @@ def expr(self):
...
         # Clear any previously-cached normalized constraint
-        self._lower = self._upper = self._body = self._expr = None
-
+        self._expr = None
         if expr.__class__ in _known_relational_expressions:
+            if getattr(expr, 'strict', False) in _strict_relational_exprs:
+                raise ValueError(
+                    "Constraint '%s' encountered a strict "
+                    "inequality expression ('>' or '< '). All"
+                    " constraints must be formulated using "
```
mrmundt, line 351: This doesn't matter and I won't bother you to change it but the inconsistent space at end / beginning of new lines irked me here.

pyomo/core/base/constraint.py
```diff
@@ -64,6 +64,7 @@
     InequalityExpression,
     RangedExpression,
 }
+_strict_relational_exprs = {True, (False, True), (True, False), (True, True)}
```
mrmundt, line 67: Do you really need to declare this as a global var if you're only using it once?

jsiirola: Yes: It is used once, but for every ConstraintData... so for a 200M constraint problem, that would create and throw away the set 200M times 

[This is one of the weird(er) performance tricks: if you are looking up into a group of more than ~2 things, it is faster to use a set than, say, a list - but only if you are not creating & throwing the set away each time]

pyomo/core/base/constraint.py
jsiirola: Fixed.

pyomo/core/base/constraint.py
```diff
@@ -167,126 +168,145 @@ def __init__(self, expr=None, component=None):
...
 
     def __call__(self, exception=True):
         """Compute the value of the body of this constraint."""
-        return value(self.body, exception=exception)
+        body = self.normalize_constraint()[1]
+        if body.__class__ not in native_numeric_types:
+            body = value(self.body, exception=exception)
+        return body
+
+    def normalize_constraint(self):
```
emma58, line 182: I'm not sure I like the name of this method... 'Normalize' makes me think of means of 1 and stuff... Could it be `tupelize_constraint` or something?

pyomo/core/base/constraint.py
```diff
@@ -422,120 +449,6 @@ def set_value(self, expr):
                     "\n   (0, model.price[item], 50)" % (self.name, str(expr))
                 )
                 raise ValueError(msg)
-        #
-        # Normalize the incoming expressions, if we can
-        #
```
emma58, line 427: This is a very beautiful block of deleted code! :D

pyomo/core/kernel/constraint.py
```diff
@@ -177,6 +177,9 @@ class _MutableBoundsConstraintMixin(object):
     # Define some of the IConstraint abstract methods
     #
 
+    def normalize_constraint(self):
+        return self.lower, self.body, self.upper
```
emma58, line 181: If you agree with my naming quibbles, this should change too.

pyomo/core/tests/unit/test_con.py
```diff
@@ -134,9 +168,11 @@ def rule(model):
         model.c = Constraint(rule=rule)
 
         self.assertEqual(model.c.equality, False)
-        self.assertEqual(model.c.lower, None)
+        self.assertEqual(model.c.lower, float('-inf'))
```
emma58, line 171: Is this change backwards compatible? Or were we testing for the existence of a bug?

emma58 (commented):
This mostly looks good (nice sneaky FBBT additions!), but I have quibbles with the naming of `normalize_constraint`.

pyomo/core/base/constraint.py
jsiirola: I'm not too keen on `tuplize_constraint` (it doesn't feel very descriptive); but I agree that `normalize_constraint` is probably worse.  What about `standardize_constraint` or `to_range_constraint` (the latter is a bit misleading as it does not return a `RangedExpression`)?

emma58: Hmmm. I actually kind of like `to_range_constraint`... Maybe `to_tuple` (also not super descriptive, but since we accept Constraint expressions in that form, it kinda makes sense)? I don't love `standardize_constraint` because of potential standard form connotations, but I do like it a lot better than `normalize_constraint`!

jsiirola: `to_bounded_expression()`?

emma58: Oooo, I think I like that one!

pyomo/core/tests/unit/test_con.py
jsiirola: We were testing for the existence of a bug:  By (current) convention, `lower` returns the *expression* for the lower bound and `lb` returns the *value* (or `None` if not specified or `-inf`).  The problem was that in the old world order, some bounds processing was done "early" in `set_value` -- including some of the conversion from +/-inf to `None`.  We now defer that processing until the user asks for the value.  So, in this test, since the user provided the "expression" `float('-inf')` for the lower bound, the "correct" thing to do is return that expression from `lower`.

pyomo/core/base/constraint.py
jsiirola: Implemented!

pyomo/contrib/fbbt/tests/test_fbbt.py
```diff
@@ -1335,3 +1335,31 @@ def test_named_expr(self):
 class TestFBBT(FbbtTestBase, unittest.TestCase):
     def setUp(self) -> None:
         self.tightener = fbbt
+
+    def test_ranged_expression(self):
+        # The python version of FBBT is slightly more flexible than
+        # APPSI's cmodel (it allows - and correctly handles -
+        # RangedExpressions with variable lower / upper bounds.  If we
```
mrmundt, line 1342: I think this is missing a close parenthesis?

mrmundt (approved):
I had one minor nitpick about a missed close-parenthesis but otherwise, I'm happy with it!

emma58 (approved):

pyomo/core/tests/unit/test_con.py
```diff
@@ -84,21 +84,55 @@ def rule(model):
...
-            return (model.x, float('inf'))
-
-        model.c = Constraint(rule=rule)
-        self.assertRaises(ValueError, model.create_instance)
-
-        model = self.create_model(abstract=True)
-
-        def rule(model):
-            return (float('inf'), model.x)
+        model = self.create_model(abstract=True).create_instance()
```
blnicho, line 87: Why not change this to just create a `ConcreteModel`? Are there behaviors we expect with `AbstractModels` that are no longer being tested because of this change?

pyomo/core/tests/unit/test_con.py
```diff
@@ -276,21 +324,23 @@ def rule(model):
...
 
     def test_expr_construct_inf_equality(self):
-        model = self.create_model(abstract=True)
-
-        def rule(model):
-            return model.x == float('inf')
-
-        model.c = Constraint(rule=rule)
-        self.assertRaises(ValueError, model.create_instance)
+        model = self.create_model(abstract=True).create_instance()
```
blnicho, line 327: Same question about just using `ConcreteModel`.

pyomo/core/base/constraint.py
```diff
@@ -167,126 +168,172 @@ def __init__(self, expr=None, component=None):
...
+        This method "standardizes" the expression into a 3-tuple of
+        expressions: (`lower_bound`, `body`, `upper_bound`).  Upon
+        conversion, `lower_bound` and `upper_bound` are guaranteed to be
+        `None`, numeric constants, or fixed (not necessarily constant)
+        expressions.
+
+        Note
+        ----
+        As this method operates on the *current state* of the
+        expression, the any required expression manipulations (and by
```
blnicho, line 194:
```suggestion
        expression, any required expression manipulations (and by
```

pyomo/core/base/constraint.py
```diff
@@ -167,126 +168,172 @@ def __init__(self, expr=None, component=None):
...
+        extension, the result) can change after fixing / unfixing
+        :py:class:`Var` objects.
+
+        Raises
+        ------
+
+        ValueError: Raised if the expression cannot be mapped to this
+            form (i.e., :py:class:`RangedExpression` constraints with
+            variable lower of upper bounds.
```
blnicho, line 203:
```suggestion
            variable lower or upper bounds.
```

blnicho (commented):
I suspect that this section in `pyomo.contrib.sensitivity_toolbox` can also be simplified with this change: https://github.com/Pyomo/pyomo/blob/1ada528693d8879f17978fe47263f8bb2b467b0a/pyomo/contrib/sensitivity_toolbox/sens.py#L693

pyomo/core/tests/unit/test_con.py
jsiirola: I think we *could* just switch this over to a ConcreteModel.  I left it as it was mostly to minimize changes in old code / tests.  The biggest reason for the change here was to remove the test for an exception on model creation, as this PR defers the bounds sanity checking until later (model compilation step and not model construction step).

pyomo/core/tests/unit/test_con.py
jsiirola: Same as above: we certainly could change these tests, but I didn't to minimize (unnecessary) changes to old code.

jsiirola:
@blnicho: I went ahead and updated the logic in `sens.py`, but did not rework the use of abstract models in the old Constraint tests

blnicho:
@jsiirola I don't see the changes to `sens.py`. Did you forget to push?

blnicho (approved):

PR #3295, opened 2024-06-19
PyomoNLP scaling factors on sub-blocks

bknueven:
## Summary/Motivation:
Scaling factors in the `PyomoNLP` interface will not be applied if the `scaling_factor` Suffix is on a sub-block.

- Change `PyomoNLP` scaling determination methods to use `SuffixFinder` instead (23d316e, a7791e1)
- Change `PyomoGreyBoxNLP` and `PyomoNLPWithGreyBoxes` to use `SuffixFinder` (8215ef8, c7b1bca)
- Maintain backwards compatibility ([9ca3928](https://github.com/Pyomo/pyomo/pull/3295/commits/9ca392825ad90696d06440845f74ff33392c4f4a))
- Update the tests so they fail with nested `scaling_factors` (0b4d16b) 

As-is this PR would change the behavior of these methods in that they would always return scaling factors. I cannot seem to exactly replicate the prior behavior which would allow all existing tests to pass without modification.

### Legal Acknowledgement


pyomo/contrib/pynumero/algorithms/solvers/tests/test_cyipopt_interfaces.py
```diff
@@ -156,9 +156,9 @@ def test_model1_CyIpoptNLP_scaling(self):
 
         cynlp = CyIpoptNLP(PyomoNLP(m))
         obj_scaling, x_scaling, g_scaling = cynlp.scaling_factors()
-        self.assertTrue(obj_scaling is None)
-        self.assertTrue(x_scaling is None)
-        self.assertTrue(g_scaling is None)
+        self.assertTrue(obj_scaling == 1.0)
+        self.assertTrue((x_scaling == 1.0).all())
+        self.assertTrue((g_scaling == 1.0).all())
```
Robbybp, line 161: Not sure how others feel about this, but I would prefer to maintain backwards compatibility here. I care less about the `CyIpoptNLP.scaling_factors` than maintaining consistency of `PyomoNLP.get_primals_scaling`. Somebody could very well be using `PyomoNLP.get_primals_scaling() is None` to check whether scaling is used.

pyomo/contrib/pynumero/interfaces/pyomo_nlp.py
```diff
@@ -298,34 +299,24 @@ def get_inequality_constraint_indices(self, constraints):
...
-            for i, v in enumerate(self.get_pyomo_variables()):
-                if v in scaling_suffix:
-                    primals_scaling[i] = scaling_suffix[v]
-            return primals_scaling
-        return None
+        scaling_suffix_finder = SuffixFinder('scaling_factor', 1.0)
+        primals_scaling = np.ones(self.n_primals())
+        for i, v in enumerate(self.get_pyomo_variables()):
+            primals_scaling[i] = scaling_suffix_finder.find(v)
+        return primals_scaling
```
Robbybp, line 311: Returning ones instead of `None` here when no scaling factors have been defined is the backward incompatibility I'm concerned about. Maybe something like this would be better?
```python
scaling_suffix_finder = SuffixFinder("scaling_factor")
scaling_factors = [scaling_suffix_finder.find(v) for v in self.get_pyomo_variables()]
has_scaling_factors = any(sf is not None for sf in scaling_factors)
if has_scaling_factors:
    return [sf if sf is not None else 1.0 for sf in scaling_factors]
else:
    return None
```
It's not exactly the same as having any non-empty `scaling_factor` suffix used to be enough to avoid returning `None`. Now we need to actually contain one of the primal variables. Maybe we could make the `SuffixFinder._get_suffix_list` method public, and return `None` if all these suffixes are empty?

Robbybp (commented):
Thanks for the PR, this is a good change. I have a minor concern about backward incompatibility, see below. Also, can you add a small unit test to `TestPyomoNLP` in `test_nlp.py` that exercises this functionality?

pyomo/contrib/pynumero/interfaces/pyomo_nlp.py
bknueven: I implemented something equivalent originally (see https://github.com/Pyomo/pyomo/pull/3295/commits/23d316e857b35e81d990114a918f215e5e3a556e) but this will break the existing "hacks" in the implicit function solver: https://github.com/Pyomo/pyomo/blob/ae354aeb7d62b62b049143be84ac6a7437bbe44d/pyomo/contrib/pynumero/algorithms/solvers/implicit_functions.py#L406-L410

To support both cases this function could an optional argument `always_return_numeric=False` or something similar to determine which behavior is desired. Maybe you have a better idea about how to de-conflict?

Robbybp: Yes, if we do this, I'll have to explicitly set scaling factors to 1 in the implicit function solver, which wouldn't be the worst thing ever. I think explicitly checking `SuffixFinder._get_suffix_list` for at least one non-empty suffix would get around this as well.

bknueven: In 9ca3928 I have opted just to replicate the original check. This seems the safest and doesn't involve accessing private methods or modifying other parts of Pyomo.

bknueven:
> Thanks for the PR, this is a good change. I have a minor concern about backward incompatibility, see below. Also, can you add a small unit test to `TestPyomoNLP` in `test_nlp.py` that exercises this functionality?

Thanks for the review. I believe I have address all comments except for potential modifications to `SuffixFinder` to support a `context` argument. I believe the same potential issue would also affect the NL writer.

pyomo/contrib/pynumero/interfaces/pyomo_grey_box_nlp.py
```diff
@@ -227,12 +228,12 @@ def __init__(self, pyomo_model):
...
-            need_scaling = True
-            for i, v in enumerate(self._pyomo_model_var_datas):
-                if v in scaling_suffix:
-                    self._primals_scaling[i] = scaling_suffix[v]
+        scaling_suffix_finder = SuffixFinder('scaling_factor')
+        for i, v in enumerate(self._pyomo_model_var_datas):
+            v_scaling = scaling_suffix_finder.find(v)
+            if v_scaling is not None:
+                need_scaling = True
+                self._primals_scaling[i] = v_scaling
```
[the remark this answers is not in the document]
Robbybp: @blnicho @jsiirola Can I get a second opinion here? The question is: what should happen when a PyomoNLP is constructed from a subblock that has relevant scaling factors on the parent block:
```python
import pyomo.environ as pyo
from pyomo.contrib.pynumero.interfaces.pyomo_nlp import PyomoNLP

m = pyo.ConcreteModel()
m.b = pyo.Block()
m.b.x = pyo.Var([1, 2], initialize={1: 100, 2: 20})

# Components so we don't have an empty NLP
m.b.eq = pyo.Constraint(expr=m.b.x[1]*m.b.x[2] == 2000)
m.b.obj = pyo.Objective(expr=m.b.x[1]**2 + m.b.x[2]**2)

m.scaling_factor[m.b.x[1]] = 1e-2
m.scaling_factor[m.b.x[2]] = 1e-1

nlp = PyomoNLP(m.b)
scaling = nlp.get_primals_scaling()
print(scaling)
```
Current behavior:
```console
None
```
Proposed:
```console
[0.01, 0.1 ]
```

Robbybp (commented):
Thanks for making this backwards compatible. That part looks good. Just want to get a second opinion about suffixes on parent components.

pyomo/contrib/pynumero/interfaces/pyomo_grey_box_nlp.py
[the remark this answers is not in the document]
jsiirola: I think that I would advocate the old behavior (although I can see value in both approaches!).

My rationale is the following: if we define a "model" as all "active components reachable through active blocks contained within the reference block," then we should exclude Suffixes outside of the subtree, as Suffix is an active component.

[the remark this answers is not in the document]
Robbybp: I'm inclined to agree.

blnicho:
@bknueven the PR adding the `context` option to `SuffixFinder` has been merged. I think you were planning to use that in this PR based on the discussion during the dev call. Leave a comment when this is ready to re-review.

bknueven:
@Robbybp @blnicho this is ready for re-review after integrating the new `context` option for the `SuffixFinder`.

pyomo/contrib/pynumero/interfaces/pyomo_nlp.py
```diff
@@ -298,34 +299,51 @@ def get_inequality_constraint_indices(self, constraints):
...
+        # maintain backwards compatibility
         scaling_suffix = self._pyomo_model.component('scaling_factor')
         if scaling_suffix and scaling_suffix.ctype is pyo.Suffix:
-            if obj in scaling_suffix:
-                return scaling_suffix[obj]
-            return 1.0
-        return None
+            return 1.0 if val is None else val
+        else:
+            return val
```
jsiirola, line 308: This is fine, but could be significantly simpler:
```python
scaling_finder = SuffixFinder('scaling_factor', default=1.0, context=self._pyomo_model)
val = suffix_finder.find(self.get_pyomo_objective())
if not scaling_finder.all_suffixes:
    return None
return val
```

pyomo/contrib/pynumero/interfaces/pyomo_nlp.py
```diff
@@ -298,34 +299,51 @@ def get_inequality_constraint_indices(self, constraints):
...
         scaling_suffix = self._pyomo_model.component('scaling_factor')
         if scaling_suffix and scaling_suffix.ctype is pyo.Suffix:
-            primals_scaling = np.ones(self.n_primals())
-            for i, v in enumerate(self.get_pyomo_variables()):
-                if v in scaling_suffix:
-                    primals_scaling[i] = scaling_suffix[v]
             return primals_scaling
-        return None
+        else:
+            return ret
```
jsiirola, line 327: This is fine, but could be significantly simpler:
```python
scaling_finder = SuffixFinder('scaling_factor', default=1.0, context=self._pyomo_model)
primals_scaling = numpy.fromiter(
    (scaling_finder.find(v) for v in self.get_pyomo_variables()), 
    count=self.n_primals(),
)
if not scaling_finder.all_suffixes:
    return None
return primals_scaling
```

pyomo/contrib/pynumero/interfaces/pyomo_nlp.py
```diff
@@ -298,34 +299,51 @@ def get_inequality_constraint_indices(self, constraints):
...
         scaling_suffix = self._pyomo_model.component('scaling_factor')
         if scaling_suffix and scaling_suffix.ctype is pyo.Suffix:
-            constraints_scaling = np.ones(self.n_constraints())
-            for i, c in enumerate(self.get_pyomo_constraints()):
-                if c in scaling_suffix:
-                    constraints_scaling[i] = scaling_suffix[c]
             return constraints_scaling
-        return None
+        else:
+            return ret
```
jsiirola, line 346: This is fine, but could be significantly simpler:
```python
scaling_finder = SuffixFinder('scaling_factor', default=1.0, context=self._pyomo_model)
constraints_scaling = numpy.fromiter(
    (scaling_finder.find(v) for v in self.get_pyomo_constraints()), 
    count=self.n_constriaints(),
)
if not scaling_finder.all_suffixes:
    return None
return constraints_scaling
```

pyomo/contrib/pynumero/interfaces/pyomo_nlp.py
```diff
@@ -607,12 +625,18 @@ def __init__(self, pyomo_model):
...
+            if v_scaling is not None:
+                need_scaling = True
+                self._primals_scaling[i] = v_scaling
+        # maintain backwards compatibility
+        scaling_suffix = self._pyomo_model.component('scaling_factor')
         if scaling_suffix and scaling_suffix.ctype is pyo.Suffix:
             need_scaling = True
-            for i, v in enumerate(self.get_pyomo_variables()):
-                if v in scaling_suffix:
-                    self._primals_scaling[i] = scaling_suffix[v]
```
jsiirola, line 615: This is fine, but could be significantly simpler:
```python
scaling_finder = SuffixFinder('scaling_factor', default=1.0, context=self._pyomo_model)
self._primals_scaling = numpy.fromiter(
    (scaling_finder.find(v) for v in self.get_pyomo_variables()), 
    count=self.n_primals(),
)
need_scaling = bool(scaling_finder.all_suffixes)
```

jsiirola (approved):
This is fine, but there is a cleaner / more efficient way to make use of the SuffixFinder

bknueven:
> This is fine, but there is a cleaner / more efficient way to make use of the SuffixFinder

Thanks for showing me that; I implemented your suggestion.

Robbybp (commented):
Can you add a test that makes sure we don't pull a scaling factor from a parent block?

bknueven:
> Can you add a test that makes sure we don't pull a scaling factor from a parent block?

Sure, I just copy+pasted your example above and made it a test in 7e79e19.

Robbybp (approved):
Thanks, looks good.

PR #3298, opened 2024-06-24
Ignore errors on ASL solver version check

alma-walmsley:
## Summary/Motivation:
Trying to run the `couenne` solver executable through Pyomo throws an error. Pyomo tries to check the version of the executable specified to verify that the solver can be run. In couenne's case, running `couenne -v` returns `couenne (Microsoft cl 15.00.21022.08 für 80x86), ASL(20100130)`, except the `ü` runs into a problem with default encoding / utf-8 and cannot be decoded. This throws an error in the subprocess call which is not caught and so the solver cannot be run. Note that this problem is not necessarily couenne specific, but rather due to `subprocess` expecting the data in a text format where special characters like `ü` could not be recognised.

To reproduce, for example:

```python
from pyomo.environ import *

model = ConcreteModel()

model.x = Var(initialize=0, bounds=(0, 10))
model.obj = Objective(expr=model.x, sense=maximize)  # maximize x -> x=10

# solve the model
solver = SolverFactory("couenne")
solver.solve(model)  # throws an error: 'utf-8' codec can't decode byte 0x81 in position 38: invalid start byte
```

- Add `errors='ignore'` to the ASL solver version check subprocess call. This will ignore any errors while trying to decode the data received from the `-v` command (ie. errors arising from special characters, unexpected bytes, etc). After all, Pyomo does not care about the value of the solver version; the version check appears to only verify the solver exists before attempting to solve - See `opt/base/solvers.py` and `solvers/plugins/solvers/ASL.py` for more details. You can also read about the `errors` argument to `subprocess` here: https://docs.python.org/3.10/library/io.html#io.TextIOWrapper

### Legal Acknowledgement


jsiirola (approved):
This looks good.  It would be nice if we had a way of testing this (i.e., testing against solvers compiled in any `LOCALE` other than `en_US` - as this is not the first time we have hit issues running in languages other than English), but I don't think that should prevent merging.

We should also consider changing the `universal_newlines=True` to `test=True` in that `subprocess.run()` call (`text` is the documented argument, and `universal_newlines` is now being provided for backwards compatibility).

mrmundt (approved):

PR #3302, opened 2024-06-25
Improve `Set` initialization performance

jsiirola:
## Summary/Motivation:
When working on some very large models (100M+ variables), we identified some performance issues in how we were initializing `Set` objects.  This PR resolves those performance bottlenecks:
```python
import time
from pyomo.environ import *
start = time.time()
m = ConcreteModel()
m.I = Set(initialize=range(100000000))
print(time.time() - start)
```
is reduced from 91.3s to 13.2s, and a real model with a more complex set initializer (for 200M elements) is reduced from 305s to 69s.

One side effect is to remove the warning message for adding duplicate items to a Set (which brings the Pyomo `Set` behavior closer in line with the Python `set` behavior).

- Rework `Set` initialization to make initializing large data sets more efficient
- Remove the warning for adding duplicate elements to a Set

### Legal Acknowledgement


pyomo/core/base/set.py
```diff
@@ -1685,15 +1748,15 @@ def discard(self, val):
...
     def pop(self):
         try:
             ans = self.last()
         except IndexError:
-            # Map the index error to a KeyError for consistency with
-            # set().pop()
-            raise KeyError('pop from an empty set')
+            # Map the exception for iterating over an empty dict to a
+            # KeyError for consistency with set().pop()
+            raise KeyError('pop from an empty set') from None
```
emma58, line 1759: Why are you suppressing the context?

emma58 (commented):
I think this looks good. A few questions mostly out of curiosity!

emma58:
(Also I am *very* excited about not warning for duplicate entries! :))

pyomo/core/base/set.py
```diff
@@ -1655,27 +1715,30 @@ class OrderedSetData(_OrderedSetMixin, FiniteSetData):
...
-    def _add_impl(self, value):
-        self._values[value] = len(self._values)
-        self._ordered_values.append(value)
+    def _update_impl(self, values):
+        for val in values:
+            # Note that we reset _ordered_values within the loop because
+            # of an old example where the initializer rule makes
+            # reference to values previously inserted into the Set
+            # (which triggered the creation of the _ordered_values)
+            self._ordered_values = None
```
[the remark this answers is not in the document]
jsiirola: The offending example is:
```python
def U_init(model, z):
    if z == 6:
        return Set.End
    if z == 1:
        return 1
    else:
        return model.U[z - 1] * z

model.U = Set(ordered=True, initialize=U_init)
```
The issue is that when `z == 2` the `__getitem__` causes the `_ordered_values` list to be created with a single entry.  If we don't clear `_ordered_values`, then it will stay a 1-element list, and when `z == 3`, looking for `model.U[2]` looks up `_ordered_values[1]`; running off the end of the list.

pyomo/core/base/set.py
```diff
@@ -1457,17 +1383,145 @@ def clear(self):
...
+                    "The value=%s violates the validation rule of Set %s"
+                    % (value, self.name)
+                )
+            yield value
+
+    def _cb_normalized_dimen_verifier(self, dimen, val_iter):
+        # It is important that the iterator is an actual iterator
+        val_iter = iter(val_iter)
+        for value in val_iter:
+            if value.__class__ is tuple:
```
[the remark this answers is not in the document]
jsiirola: Mostly because isinstance is slower, and because what we are trying to detect are multi-dimensional indices, which we know (because of normalize_index) is a tuple.

pyomo/core/base/set.py
jsiirola: So that `Set` behaves like `set`.

pyomo/core/base/set.py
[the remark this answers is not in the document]
emma58: Ewwww, this is horrifying! But okay, makes sense.

emma58 (approved):

mrmundt (approved):

pyomo/core/base/set.py
```diff
@@ -1457,17 +1383,145 @@ def clear(self):
...
+        #  (like lists) are not hashable and can raise exceptions.
+        try:
+            self._update_impl(val_iter)
+        except Set._SetEndException:
+            pass
 
     def pop(self):
         return self._values.pop()
 
+    def _cb_domain_verifier(self, domain, val_iter):
```
blnicho, line 1441: What does the `cb` in these verifier/validator method names mean?

pyomo/core/base/set.py
```diff
@@ -1457,17 +1383,145 @@ def clear(self):
...
+            if not flag:
+                raise ValueError(
+                    "The value=%s violates the validation rule of Set %s"
+                    % (value, self.name)
+                )
+            yield value
+
+    def _cb_normalized_dimen_verifier(self, dimen, val_iter):
+        # It is important that the iterator is an actual iterator
+        val_iter = iter(val_iter)
```
blnicho, line 1476: Is this needed? Seems like `val_iter` is guaranteed to be an iterator from line 1399 unless you expect this method to be called from elsewhere.

blnicho (approved):
I have a couple minor questions but otherwise this looks great!

pyomo/core/base/set.py
jsiirola: "callback"

pyomo/core/base/set.py
jsiirola: Great catch!  Removed this, plus updated some other bugs / inconsistencies / documentation around flattening & initialization.

blnicho (approved):

PR #3304, opened 2024-06-26
Adds support for the validation of IndexedSets using the index

### LINKED ISSUE #2655: No access to the index when validating indexed set members
## Summary

I would like to validate my indexed sets using the 'validate' keyword. The problem is that the index is not passed as an argument, only the model and the prospective member of the set. It would arguably be good for sets to have the same validation options as parameters (the component).

### Rationale

There should be a way to validate tentative members of indexed sets using the respective index. As I understand it, this is currently not possible. As far as I know, only the potential member is passed as an argument, not the index:

```
import pyomo.environ as pyo
model = pyo.ConcreteModel()
model.A = pyo.Set()
def validate_B(m, member):
    return True
model.B_indexed = pyo.Set(model.A, validate=validate_B)
```

### Description

The 'validate' option works well with parameter components, since the index is passed as an argument too. The same behaviour should be replicated with the set component:

```
import pyomo.environ as pyo
model = pyo.ConcreteModel()
model.A = pyo.Set()
def validate_B(m, member, index):
    return member > index
model.B_indexed = pyo.Set(model.A, validate=validate_B)
```

### Additional information

My current solution relies on validating the data outside of pyomo. Alternatively, I could validate every member of the set every time the validation rule was called but that would be very inefficient.

My setup:
Python 3.8.10
Pyomo 6.4.2

pmlpm1986:
## Summary/Motivation:

Set members could not be validated using the index in indexed sets. This PR adds that option to allow for more complex validation rules. 

- Allows validation of indexed sets to access the member and the respective index just like in Parameter objects
- Supports one- and multi-dimensional set members
- Adds tests to retain the functionality
- Revised the tutorial to illustrate the functionality.

### Legal Acknowledgement


pyomo/core/base/set.py
mrmundt, line 1: It looks like `black` might have gone a bit crazy on this file. Can I please ask you to revert it and add back the section you actually changed (lines 1412 through 1424)?

pyomo/core/base/sets.py
```diff
@@ -27,7 +27,7 @@
 from pyomo.common.deprecation import deprecation_warning
```
mrmundt, line 27: It looks like `black` might have gone a bit crazy on this file. Can you please revert it?

pyomo/core/tests/unit/test_set.py
```diff
@@ -334,12 +334,12 @@ def test_setdefault(self):
     def test_indices(self):
```
mrmundt, line 334: It looks like `black` might have gone a bit crazy on this file. Can I please ask you to revert it and add back the lines you actually changed?

pyomo/core/tests/unit/test_sets.py
```diff
@@ -210,8 +210,8 @@ def test_addInvalid(self):
         # is, the default within is None
```
mrmundt, line 210: It looks like `black` might have gone a bit crazy on this file. Can I please ask you to revert it and add back the lines you actually changed?

pmlpm1986:
I reverted the changes introduced by black. It should work as intended now.

pyomo/core/base/set.py
```diff
@@ -1409,7 +1410,19 @@ def add(self, *values):
 
             if self._validate is not None:
                 try:
-                    flag = self._validate(_block, _value)
+                    # differentiate between indexed and non-indexed sets
+                    if self._index is not None:
+                        # indexed set: the value and the index are given
+                        pass
```
mrmundt, line 1416: I do not think this `pass` is necessary.

pyomo/core/tests/unit/test_set.py
```diff
@@ -4319,15 +4319,15 @@ def _lt_3(model, i):
 
         m = ConcreteModel()
 
-        def _validate(model, i, j):
+        def _validate_I(model, i, j):
```
mrmundt, line 4322: I envision that future developers looking at this code will wonder, "What is `I`?"

pyomo/core/tests/unit/test_set.py
```diff
@@ -4347,7 +4347,10 @@ def _validate(model, i, j):
 
         # Note: one of these indices will trigger the exception in the
         # validot when it is called for the index.
-        m.J = Set([(0, 0), (2, 2)], validate=_validate)
+        def _validate_J(model, i, j, index):
```
mrmundt, line 4350: I envision that future developers looking at this code will wonder, "What is `J`?"

mrmundt (changes requested):

pyomo/core/tests/unit/test_set.py
pmlpm1986: In my mind, it is just an indexed set that we validate through this new rule.

pyomo/core/tests/unit/test_set.py
pmlpm1986: In my mind, it is a simple set that we validate through the normal rule (no index).

pmlpm1986: That is, a non-indexed set.

pyomo/core/base/set.py
pmlpm1986: True. Not sure how it landed there. Will fix it soon.

jsiirola:
This is pretty close.  I really only have two concerns with it:
- I would prefer that this leverage the `Initializer()` logic if at all possible: that logic standardizes how we process arguments to Component constructors so that (hopefully) all Components / arguments behave the same.
- (and this is the bigger issue): this PR breaks backwards compatibility.  That is, older models (including the tests in `test_sets`) no longer work with this change.  Although The Book is ambiguous as to the signature that `IndexedSet.validate` should take, the fact that there are examples / tests that exercise the old convention makes me uneasy about breaking it.  I think we should update this PR to either
   1. Work (silently) with either syntax
   2. Expect the new syntax but accept the old syntax (issuing a deprecation warning) 

I am happy to work on a patch for this, but it will take me a little while to get to it.

pmlpm1986:
> This is pretty close. I really only have two concerns with it:
> 
>     * I would prefer that this leverage the `Initializer()` logic if at all possible: that logic standardizes how we process arguments to Component constructors so that (hopefully) all Components / arguments behave the same.
> 
>     * (and this is the bigger issue): this PR breaks backwards compatibility.  That is, older models (including the tests in `test_sets`) no longer work with this change.  Although The Book is ambiguous as to the signature that `IndexedSet.validate` should take, the fact that there are examples / tests that exercise the old convention makes me uneasy about breaking it.  I think we should update this PR to either
>       
>       1. Work (silently) with either syntax
>       2. Expect the new syntax but accept the old syntax (issuing a deprecation warning)
> 
> 
> I am happy to work on a patch for this, but it will take me a little while to get to it.

As you say, the book is vague about this and it led me to think Sets behaved exactly like Parameters. Once I realised that they did not, I raised this issue. Only recently did I find the time to come up with a solution. Therefore, the `Initializer()` approach makes sense but I never looked into it.

On the subject of backwards compatibility, the adjustments needed are minimal and by design: all that is needed is to add the index to the arguments (it is always the last argument) and ignore it, which is what I did in some of the tests. Issuing a deprecation warning seems reasonable.

jsiirola:
This PR has been superseded by #3338 (that PR includes all commits here plus the rework of the Initializer and restoration of backwards compatibility [with deprecation warnings]).  I am closing this PR to ease some of the burden on the CI system.

PR #3310, opened 2024-07-08
Conic geomean fix

sadavis1:
## Summary/Motivation: 
pyomo.kernel.conic.primal_geomean.as_domain() does not work without a 2-character fix (dies in Mosek code). dual_geomean.as_domain() seems to have the same problem.

- Fix _build_linking_constraints() call in previously mentioned methods

### Legal Acknowledgement


pyomo/core/kernel/conic.py
```diff
@@ -934,7 +934,7 @@ def as_domain(cls, r, x):
         b = block()
         b.r = variable_tuple([variable(lb=0) for i in range(len(r))])
         b.x = variable()
-        b.c = _build_linking_constraints(list(r) + [x], list(b.r) + [x])
+        b.c = _build_linking_constraints(list(r) + [x], list(b.r) + [b.x])
```
mrmundt, line 937: This seems like a perfect case to add a small test to ensure correct behavior in different solvers.

sadavis1: I added a test for the constraint structure. The behavior isn't really specific to mosek, it just happened to fail that way because mosek doesn't like constraints that specify the same variable twice.

pyomo/core/tests/unit/kernel/test_conic.py
```diff
@@ -784,6 +786,40 @@ def test_as_domain(self):
         x[1].value = None
 
 
+# these mosek 10 constraints are really anemic and can't be evaluated, pprinted,
```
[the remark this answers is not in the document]
sadavis1: I guess my idea was that they lacked the methods a healthy conic constraint should have... I'll be clear instead of typing things that make no sense.

mrmundt (approved):

jsiirola (approved):

PR #3315, opened 2024-07-10
contrib.cp: Adding beforeChild handling for bools in logical expressions 

emma58:
## Summary/Motivation:

This fixes a bug in the logical-to-disjunctive walker where we would hit our own carefulness in terms of not mixing algebraic and logical expressions. Specifically, the following code
```
from pyomo.environ import *
from pyomo.contrib.cp.transform.logical_to_disjunctive_walker import (
    LogicalToDisjunctiveVisitor,
)

m = ConcreteModel()
m.a = BooleanVar()
e = m.a.equivalent_to(True)

visitor = LogicalToDisjunctiveVisitor()
m.cons = visitor.constraints
m.z = visitor.z_vars

visitor.walk_expression(e)
```
resulted in this error in the walker:
```
  File "/home/esjohn/src/pyomo/pyomo/contrib/cp/tests/thing.py", line 14, in <module>
    visitor.walk_expression(e)
  File "/home/esjohn/src/pyomo/pyomo/core/expr/visitor.py", line 276, in walk_expression
    result = self._process_node(root, RECURSION_LIMIT)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/esjohn/src/pyomo/pyomo/core/expr/visitor.py", line 489, in _process_node_bx
    return self.exitNode(node, data)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/esjohn/src/pyomo/pyomo/contrib/cp/transform/logical_to_disjunctive_walker.py", line 262, in exitNode
    return _operator_dispatcher[node.__class__](self, node, *data)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/esjohn/src/pyomo/pyomo/contrib/cp/transform/logical_to_disjunctive_walker.py", line 95, in _dispatch_equivalence
    _dispatch_or(visitor, node, 1 - a, b),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/esjohn/src/pyomo/pyomo/contrib/cp/transform/logical_to_disjunctive_walker.py", line 113, in _dispatch_or
    visitor.constraints.add((1 - z) + sum(args) >= 1)
                                      ^^^^^^^^^
TypeError: unsupported operand type(s) for +: 'LinearExpression' and 'bool'
```

- Checks for `bool` types specifically in `beforeChild` in the logical-to-disjunctive walker, and casts them to integer so they will be binary.
- Adds a test

### Legal Acknowledgement


mrmundt (approved):

jsiirola (approved):

pyomo/contrib/cp/transform/logical_to_disjunctive_walker.py
```diff
@@ -243,6 +243,13 @@ def initializeWalker(self, expr):
...
     def beforeChild(self, node, child, child_idx):
+        if child.__class__ is bool:
+            # If we encounter a bool, we are going to need to treat it as
+            # binary explicitly because we are finally pedantic enough in the
+            # expression system to not allow some of the mixing we will need
+            # (like summing a LinearExpression with a bool)
+            return False, int(child)
+
         if child.__class__ in EXPR.native_types:
             return False, child
```
jsiirola, line 254: This is fine, however two questions:
1. Would the following be more efficient:
```python
if child.__class__ in EXPR.native_types:
    if child.__class__ is bool:
        # If we encounter a bool, we are going to need to treat it as
        # binary explicitly because we are finally pedantic enough in the
        # expression system to not allow some of the mixing we will need
        # (like summing a LinearExpression with a bool)
        return False, int(child)
    return False, child
```
2. are there concerns with float constants here?  For Param, we error if the value is fractional, but not here.  If the same concern applies, then maybe we should just do:
```python
if child.__class__ in EXPR.native_types:
    val = int(child)
    if val != child:
        raise ValueError("...")
    return False, val
```
(which covers both the fractional values and the bool case)

emma58: Oh, good catch! But I think it's actually the reverse: I shouldn't check in Param or here because it won't be required when we probably someday allow relational expressions to be in logical expressions here. And we are checking for integrality in places where we need it (e.g., `atmost`, `atleast`, and `exactly`)

PR #3324, opened 2024-07-17
Add `ParameterizedQuadraticRepn` and corresponding expression walker

shermanjasonaf:
## Summary/Motivation:

With upcoming changes to the preprocessing subroutine of the contributed PyROS solver, and, in particular, the preprocessor's coefficient matching step, which requires analyzing expression trees in terms of selected variables only, this PR adds a `ParameterizedQuadraticRepn` object that has the same structure as the existing `QuadraticRepn` (of #2823), and an accompanying expression walker that extends the behavior of the `QuadraticRepnVisitor` and `ParameterizedLinearRepnVisitor` (of #3268) to allow for collection of quadratic/bilinear terms with potentially variable coefficient expressions. As with the `ParameterizedLinearRepnVisitor`, variables to be treated as data are specified through the `wrt` argument to the `ParameterizedQuadraticRepnVisitor` constructor.

- Add `ParameterizedQuadraticRepn` class that inherits from `QuadraticRepn` but re-implements all methods/attributes that assume that constants and coefficients are numbers rather than expressions.
- Add `ParameterizedQuadraticRepnVisitor` that inherits from `ParameterizedLinearRepnVisitor` and features:
  - redefined `exitNode` handlers for products of linear expressions and products of quadratic/nonlinear expressions
  - a redefined `finalizeResult` method
  - expansion of nonlinear products by default

### Legal Acknowledgement


pyomo/repn/parameterized_quadratic.py
```diff
@@ -0,0 +1,439 @@
...
+            if vid in dest_dict:
+                dest_dict[vid] += mult * coef
+            else:
+                dest_dict[vid] = mult * coef
+    else:
+        for vid, coef in src_dict.items():
+            if vid in dest_dict:
+                dest_dict[vid] += coef
+            else:
+                dest_dict[vid] = coef
```
jsiirola, line 62: Despite the comment, this method is functionally identical to https://github.com/Pyomo/pyomo/blob/b5825e9e146552690d942903386102105251464c/pyomo/repn/parameterized_linear.py#L47-L59

We should probably be import / reuse it here (even through it starts with an underscore).

jsiirola (approved):
This has a huge amount of repeated code (from a mix of `parameterized_linear` and `quadratic`.  In the future we should refactor the walkers so that we can make better use of mixins or more general methods to clean this up.

blnicho (approved):

PR #3338, opened 2024-08-10
Support validate / filter for IndexedSet components using the index

### LINKED ISSUE #2655: No access to the index when validating indexed set members

jsiirola:
## Summary/Motivation:
`IndexedSet` components supported the `validate` and `filter` arguments, but those callbacks didn't provide the index.  This prevented index-specific behavior, which is inconsistent to other indexed components.  This PR builds on (and supersedes) #3304 to provide the index to the callbacks for both `validate` and `filter`.  This implementation extends the standard `Initializer()` logic to support parameterized callbacks.

#### More history

We have historically supported filters on multidimensional scalar sets that look like this:
```python
model.I = RangeSet(5)
def lower_triangular(m, i, j):
    return i<= j
model.II = Set(initialize=model.I*model.I, filter=lower_triangular)
```

This starts to break down if the Set itself is indexed:
```python
model.I = RangeSet(5)
def lower_triangular(m, i, j, k):
    return i<= j and i <= k and j <= k
model.II = Set(model.I, initialize=model.I*model.I, filter=lower_triangular)
```

And can become ambiguous if the IndexedSet index or the individual SetData objects are jagged sets.

To remove the ambiguity, this PR moves to a new paradigm where the "extra data" is *not* expanded when passed to the function:
```python
model.I = RangeSet(5)
def lower_triangular(m, val, k):
    i, j = val
    return i<= j and i <= k and j <= k
model.II = Set(model.I, initialize=model.I*model.I, filter=lower_triangular)
```

- Add "Parameterized" initializer classes to the Initializer system
- Convert IndexedSet to using Initializers to process the `validate` and `filter` arguments
- Deprecate support of "expanded" value arguments to the `validate` and `filter` arguments when used on IndexedSet objects (TBD: discussion as to if we should also deprecate that API for scalar Set objects)

### Legal Acknowledgement


pyomo/core/base/initializer.py
```diff
@@ -69,9 +71,51 @@ def Initializer(
...
+        function calls (provided *before* the index value).
+
     """
     if arg is arg_not_specified:
         return None
+    if additional_args:
+        if arg.__class__ in function_types:
+            if allow_generators or inspect.isgeneratorfunction(arg):
+                raise ValueError(
+                    "Generator functions are not allowed when prassing additional args"
```
mrmundt, line 86:
```suggestion
                    "Generator functions are not allowed when passing additional args"
```

pyomo/core/base/set.py
```diff
@@ -1455,22 +1458,83 @@ def _cb_check_set_end(self, val_iter):
                 return
             yield value
 
-    def _cb_validate(self, validate, block, val_iter):
+    def _cb_validate_filter(self, mode, val_iter):
+        failFalse = mode == 'validate'
```
mrmundt, line 1462: Can this be changed to snake case (`fail_false`)? It looks weird here.

pyomo/core/base/set.py
```diff
@@ -1455,22 +1458,83 @@ def _cb_check_set_end(self, val_iter):
...
+                        orig_fcn = fcn._fcn
+                        fcn._fcn = lambda m, v, *i: orig_fcn(m, *v, *i)
+                    yield value
+                    continue
+            except TypeError:
+                pass
+            except Exception as e:
+                exc = e
+            if flag is not None:
+                if failFalse:
```
mrmundt, line 1527: Can this be changed to snake case (`fail_false`)? It looks weird here.

mrmundt (changes requested):

mrmundt (approved):

pyomo/core/base/set.py
```diff
@@ -2200,6 +2264,26 @@ def __init__(self, *args, **kwds):
...
+            if self._filter.__class__ is ParameterizedIndexedCallInitializer:
+                # TBD [JDS: 8/2024]: should we deprecate the "expanded
+                # tuple" version of the filter callback for scalar sets?
+                # It is widely used and we can (reasonably reliably) map to
+                # the expected behavior...
+                orig_fcn = self._filter._fcn
+                self._filter = ParameterizedScalarCallInitializer(
+                    lambda m, v: orig_fcn(m, *v), True
+                )
+
```
jsiirola, line 2286: Note that this code (silently) preserves the old API for scalar Set objects.  However, that old API is
- difficult to reconcile with the "newer" API for IndexedSets
- widely used in Pyomo models in the wild
I think I'd like to put in a deprecation warning here, but that would generate a LOT of noise (basically, it would generate deprecation warnings for *every* model that uses `filter=` or `validate=`).

Thoughts (@blnicho, @mrmundt, @emma58, @michaelbynum, @whart222)?

mrmundt: I don't mind disturbing users in the wild, but I want the deprecation warnings to be helpful / tell them how to change over their models. What is the new, preferred method for doing this?  Either we detail that directly in the deprecation message _or_ we make a new page on RTD that gives more detailed instructions.

jsiirola: We had a long discussion at the Dev call today and decided that we should start issuing deprecation warnings for ScalarSets now.

pyomo/core/base/set.py
```diff
@@ -1455,22 +1458,83 @@ def _cb_check_set_end(self, val_iter):
...
+            # Now try *values and index
+            try:
+                flag = fcn(block, idx, *value)
+                if flag:
+                    deprecation_warning(
+                        f"{self.__class__.__name__} {self.name}: '{mode}=' "
+                        "callback signature matched (block, *value, *index).  "
+                        "Please update the callback to match the signature "
+                        "(block, value, *index).",
+                        version='6.7.4.dev0',
```
blnicho, line 1515: Same question here about the ordering of `idx` and `value`. The deprecation warning matches the argument ordering I was expecting but the call to `fcn` does not. Is there magic happening in the initializer function call logic to swap argument ordering?

pyomo/core/base/initializer.py
```diff
@@ -316,6 +349,18 @@ def __call__(self, parent, idx):
...
+class ParameterizedIndexedCallInitializer(IndexedCallInitializer):
+    """IndexedCallCallInitializer that accepts additional arguments"""
+
+    __slots__ = ()
+
+    def __call__(self, parent, idx, *args):
+        if idx.__class__ is tuple:
+            return self._fcn(parent, *args, *idx)
+        else:
+            return self._fcn(parent, *args, idx)
```
blnicho, line 361: I think I'm answering my own question by noticing that the argument ordering is indeed swapped here.

blnicho (approved):

pyomo/core/base/initializer.py
blnicho: @jsiirola there isn't really anything for you to resolve in my comments/questions, I just wanted to document my initial confusion. I am curious why the argument ordering and subsequent reordering was necessary. I'm guessing it had to do with having more control over the format of the index being passed in.

jsiirola: Yeah - this turned into a bit of a mess (and yes, it is confusing that they were reversed).  The calls you are seeing here are all internal to the Initializer methods, so the user is never aware of them.  In this case, because the "Parameterized" initializer was new, the old `fcn` API was just `(block, index)`.  Parameterized was adding "value", so it made sense for the extra arg to come at the end.  This was further confused by the historical behavior where we would expand the value arg and pass it to the user-provided callback (leading things like the messy deprecation paths around `set.py` lines 1482-1520), and leaving things like this seemed to require less code gymnastics to implement the deprecation paths and make things compatible with the rest of the Initializer system.

Issue #3344, opened 2024-08-13
Scaling transformation does not transform named Expressions


dallan-keylogic:
## Summary

When a model is transformed using the model scaling transformation, `Expressions` are created on the scaled model, but variable scaling factors are not added to them like they are for `Constraints`, resulting in them returning erroneous values. I've encountered this issue twice now and spent significant amounts of time debugging before I realized the scaling transform was the issue.

### Steps to reproduce the issue

```
# expression_not_transformed.py
import pyomo.environ as pyo

m = pyo.ConcreteModel()

m.x = pyo.Var(initialize=10)
m.y = pyo.Var(initialize=0.1)

m.sum = pyo.Expression(expr=m.x+m.y)

m.scaling_factor = pyo.Suffix(direction=pyo.Suffix.EXPORT)
m.scaling_factor[m.x] = 0.1
m.scaling_factor[m.y] = 10

m_scaled = pyo.TransformationFactory('core.scale_model').create_using(m, rename=False)

print(f"Unscaled value: {pyo.value(m.sum)}")
print(f"Scaled value: {pyo.value(m_scaled.sum)}")
```
Unscaled value: 10.1
Scaled value: 2.0
```

### Information on your system

Pyomo version: 6.7.4.dev0
Python version: 3.10.14
Operating system: Windows 10
How Pyomo was installed: Sources
Solver (if applicable): n/a


### Additional information

The behavior I would expect would be to transform the `Expression` like a `Constraint`, using a scaling suffix if provided and otherwise just using a value of 1. In that case, both values in the example script should return `10.1`. If Pyomo does not want to support scaling of `Expressions`, then it shouldn't even generate them on the scaled model---returning some `Exception` for `pyo.value(m_scaled.sum)`.

jsiirola:
I am pretty sure it is not mathematically valid to have a scaling factor for an `Expression`.  That said, I see a value in scaling the variables in the Expression objects.

dallan-keylogic:
How is it not valid to have scaling factors for an `Expression`? It would be pointless if you're just going to totally substitute the expression into whatever constraints you have (because it would just cancel out with its inverse).

The reason we have scaling factors for `Expressions` in IDAES at the moment, though, is because:
1) We don't want to assume whether a property has been created as a `Var` or `Expression` at a unit model level, leaving that to the property package to decide
[...]

PR #3346, opened 2024-08-13
[WIP] Scale named expressions in scaling transformation

Robbybp:
## Fixes part of #3344
Corrects scaled variables in named expressions, but does not use scaling factors assigned to named expressions.

## Summary/Motivation:
Expressions in scaled model were incorrect as they would use scaled variables, but wouldn't have the factor necessary to un-apply the scaling in the expression itself. This normally wouldn't lead to incorrect solutions, as constraints removed named expressions when replacing variables with their scaled versions.

- Don't remove named expressions when scaling
- Don't descend into named expressions when scaling a parent expression
- Scale named expression separately

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/plugins/transform/scaling.py
+++ b/pyomo/core/plugins/transform/scaling.py
@@ -10,7 +10,7 @@
 #  ___________________________________________________________________________
 
 from pyomo.common.collections import ComponentMap
-from pyomo.core.base import Block, Var, Constraint, Objective, Suffix, value
+from pyomo.core.base import Block, Var, Constraint, Objective, Suffix, value, Expression
 from pyomo.core.plugins.transform.hierarchy import Transformation
 from pyomo.core.base import TransformationFactory
 from pyomo.core.base.suffix import SuffixFinder
@@ -174,7 +174,7 @@ class ScaleModel(Transformation):
 
         already_scaled = set()
         for component in model.component_objects(
-            ctype=(Constraint, Objective), descend_into=True
+            ctype=(Constraint, Objective, Expression), descend_into=True
         ):
             if component.is_reference():
                 # Skip any references - these should get picked up when
@@ -187,13 +187,13 @@ class ScaleModel(Transformation):
                     continue
                 already_scaled.add(id(c))
                 # perform the constraint/objective scaling and variable sub
-                scaling_factor = component_scaling_factor_map[c]
                 if c.ctype is Constraint:
+                    scaling_factor = component_scaling_factor_map[c]
                     body = scaling_factor * replace_expressions(
                         expr=c.body,
                         substitution_map=variable_substitution_dict,
-                        descend_into_named_expressions=True,
-                        remove_named_expressions=True,
+                        descend_into_named_expressions=False,
+                        remove_named_expressions=False,
                     )
 
                     # scale the rhs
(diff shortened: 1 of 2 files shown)


mrmundt (approved):
Software-quality-wise, this is lovely. Needs another review from someone for correctness of functionality.

jsiirola (changes requested):
I think this needs a slightly more convoluted implementation to cover cases where `Expression` objects are not on the (sub)block that is being transformed.  The expression walker needs to record all `Expression` objects that it encounters to make sure that they are appropriately transformed.  Maybe we should `remove_named_expressiosn` if the `Expression` is out-of-scope, and transform them if they are in-scope?

Consider:
```python
m = pyo.ConcreteModel()
m.inner_expr = pyo.Expression()

m.sub = Block()
m.sub.x = pyo.Var(initialize=10)
m.sub.y = pyo.Var(initialize=0.1)
m.sub.obj = Objective(expr=m.sub.x + m.sub.y + m.inner_expr)
m.sub.scaling_factor = pyo.Suffix(direction=pyo.Suffix.EXPORT)
m.sub.scaling_factor[m.sub.x] = 0.1
m.sub.scaling_factor[m.sub.y] = 10

m.inner_expr = 3 + m.sub.x**2 + m.sub.y

m_scaled = pyo.TransformationFactory('core.scale_model').create_using(
    m.sub, rename=False
)
```

Robbybp:
Good point. We could:
1. Record the named expressions we encounter in the constraint expressions and scale these, as well as all named expressions that live in the block
2. Do as you suggest and scale in-place all named expressions that live in the block, but replace those that live outside.

I think I lean towards 2, but I don't see a huge problem with 1 (if the variables have been scaled, then they should be replaced with their scaled equivalent anywhere they appear). Regardless, this implementation will actually require some thought.

jsiirola:
There is another case to worry about: what about a `Constraint` that lives outside the transformation scope that references the same `Expression` as a Constraint` within the transformation scope?  If we transform the `Expression`, then that will partially scale the out-of-scope `Constraint`.

Robbybp:
Marking as WIP until we decide how to deal with the additional complexities John points out.

Robbybp:
Maybe the simplest option is to just remove named Expressions from constraint bodies and also remove them from the block that was scaled. I don't see a way to protect against named expressions being used elsewhere, unless we explicitly duplicate all the named expressions used by the scaled block.

Robbybp:
Maybe an option to opt-in to scaling Expressions in-place, for users who know that the expressions aren't being used elsewhere (which is 90% of use cases, where we just want to scale an entire self-contained model).

jsiirola:
These issues become moot if we only allow out-of-place transformations.  If we start by cloning the (sub)block that we are transforming, that will break any external references to the Expression objects.  Then it will be safe to transform all the local Expression objects in the block hierarchy and substitute out any other Expression objects encountered in Constraint / Objective expressions.

Robbybp:
My current preference is to:
- Transform Expressions within the block scope to use scaled variables.
- If there are Expressions in the constraints that live outside the block scope, throw a warning, then replace them explicitly with their (scaled) `expr`.

However, this will break any constraints outside our scope that use expressions that were scaled. The reason I'm suggesting this is that we already break constraints outside our scope that use variables that get scaled. Another option is to clone expressions used by the constraint that are outside the block scope (and then replacing the original expressions with the cloned, scaled versions). This would attempt to preserve the common-subexpression efficiency within the scaled block while not breaking other uses of these expressions. And yes, this all only matters if we are doing the transformation in-place.

This also begs the question of what to do with variables that participate in a constraint, but are not within the block scope. These currently will not be scaled, even if they have a scaling factor on the block.

What this is revealing to me is that in-place expression transformations are not well-defined on a non-self-contained model. We could just throw an error if we detect a non-self-contained model (variables or expressions in a constraint that don't live in the block scope).

Robbybp:
Closing this as I'm not actively working on it. Will re-open when I get back to it.

PR #3363, opened 2024-08-25
Allow initialization of indexed sets via a function returning a dict

mfripp:
## Summary/Motivation:

It is possible to construct `Expression` (and possibly `Constraint`) objects by passing a rule function that constructs a dictionary containing all the elements needed to initialize the `Expression`. Pyomo notices that this function only accepts one argument, marks it as constant==True and then later `IndexedComponent._construct_from_rule_using_setitem` calls the function, accepts the dict, and uses that to build the object. However, if you try something similar with indexed sets, it does not initialize correctly from the dictionary and doesn't give an error message. This pull request fixes the `Set()` definition to allow initialization via a function that produces a dictionary or dict-like object.

The examples below use this shared setup:
```
from pyomo.environ import ConcreteModel, Set
model = ConcreteModel()
model.A = Set(initialize=[1, 2])
model.B = Set(model.A, initialize={1: [2, 3], 2: [3, 4]})
model.B.pprint()
B : Size=2, Index=A, Ordered=Insertion
    Key : Dimen : Domain : Size : Members
      1 :     1 :    Any :    2 : {2, 3}
      2 :     1 :    Any :    2 : {3, 4}
```

Here is the current behavior. Pyomo unexpectedly uses the keys of the dict as the return value for any index in `m.C`:

```
model.C = Set(model.A, initialize=lambda m: {x: [x+1, x+2] for x in m.A})
model.C.pprint()
C : Size=2, Index=A, Ordered=Insertion
    Key : Dimen : Domain : Size : Members
      1 :     1 :    Any :    2 : {1, 2}
      2 :     1 :    Any :    2 : {1, 2}
```

Here is the new behavior after this change. Now `m.C` is the same as `m.B`, which is the behavior you would expect if the rule function returns the same dictionary as was used to initialize `m.B`.

```
model.C = Set(model.A, initialize=lambda m: {x: [x+1, x+2] for x in m.A})
model.C.pprint()
C : Size=2, Index=A, Ordered=Insertion
    Key : Dimen : Domain : Size : Members
      1 :     1 :    Any :    2 : {2, 3}
      2 :     1 :    Any :    2 : {3, 4}
```

The new initialization method allows more efficient and more readable code when initializing indexed sets. For example, my project often produces indexed sets from simple sets. The most readable approach to this is along these lines:

```
import pyomo.environ as pyo
model = ConcreteModel()

model.Nodes = pyo.Set(initialize=['A', 'B', 'C'])
model.Arcs = pyo.Set(initialize=[('A', 'B'), ('A', 'C'), ('C', 'B')])

def NodesOut_init(m, node):
    return [j for (i, j) in m.Arcs if i == node]
model.NodesOut = pyo.Set(model.Nodes, initialize=NodesOut_init)
```

The [Pyomo documentation](https://pyomo.readthedocs.io/en/stable/pyomo_modeling_components/Sets.html) shows another approach with similar performance:

```
def NodesOut_init(m, node):
    for i, j in m.Arcs:
        if i == node:
            yield j
model.NodesOut = pyo.Set(model.Nodes, initialize=NodesOut_init)
model.NodesOut.pprint()
```

Both of the above become very slow if there are thousands of Nodes and millions of Arcs, since they scan all Arcs for every Node. The Pyomo documentation recommends this to improve performance:

```
model.NodesOut = pyo.Set(model.Nodes, within=model.Nodes)

def Populate_NodesOut(model):
    # loop over the arcs and record the end points
    for i, j in model.Arcs:
        model.NodesOut[i].add(j)

model.Populate_NodesOut = pyo.BuildAction(rule=Populate_NodesOut)
```

This requires users to understand `BuildActions` and to know that elements can be added to Sets after they are constructed, which is a different pattern from other components.

The new functionality in this pull request enables a simpler way to efficiently construct indexed sets:

```
def NodesOut_init(m):
    # Create a dict to show NodesOut list for every node
    d = {i: [] for i in m.Nodes}
    # loop over the arcs and record the end points
    for i, j in model.Arcs:
        d[i].append[j]
    return d

model.NodesOut = pyo.Set(model.Nodes, initialize=NodesOut_init)
```

In the future, this could enable efficient construction of indexed sets via other methods that expect a component to be initialized from a function, e.g., 

```
@model.Set(model.Nodes)
def NodesOut(m):
    d = {i: [] for i in m.Nodes}
    for i, j in model.Arcs:
        d[i].append[j]
    return d
```

(Currently this last example doesn't work because the `model.Set()` decorator passes the function to `Set()` via `rule` instead of `initialize`.)

- Revise `Set.construct()` to call the rule function once and then use its result to initialize the indexed set if `self.is_indexed() and type(self._init_values._init) is ScalarCallInitializer`. 
  - The second condition only occurs when Set() is called with a rule that accepts a single argument (the block). In all cases where a set is indexed (the first condition) but the rule function doesn't accept indexes (the second condition), it makes sense to try initializing with the object returned by the function. 
  - This could perhaps be broadened a bit by checking `self._init_values.constant()` and then checking whether `self._init_values._init` uses a function as its initializer. But there doesn't seem to be a clean way to do that, and in all these cases it looks like `self._init_values._init` is of type `ScalarCallInitializer`. The submitted test will catch if this changes.
- Add `pyomo.core.tests.unit.test_sets.TestSetErrors.test_setargs6` to verify that the code above produces the second behavior, not the first.
- Update online documentation to include this new behavior.

### Legal Acknowledgement

### DIFF
--- a/doc/OnlineDocs/pyomo_modeling_components/Sets.rst
+++ b/doc/OnlineDocs/pyomo_modeling_components/Sets.rst
@@ -24,6 +24,8 @@ and postpones creation of its members:
 .. doctest::
    :hide:
 
+   >>> # testcode ran in a different scope, so re-run here
+   >>> model.A = pyo.Set()
    >>> # Add some data to the Set
    >>> model.A.update([1,2,3])
 
@@ -465,20 +467,20 @@ needed for the indexed set:
 .. testcode::
 
     def NodesIn_init(m, node):
-        # Create a dict to show NodesOut list for every node
+        # Create a dict to show NodesIn list for every node
         d = {i: [] for i in m.Nodes}
         # loop over the arcs and record the end points
         for i, j in model.Arcs:
-            d[i].append[j]
-        return d
-    model.NodesOut = pyo.Set(model.Nodes, initialize=NodesOut_init)
-
-    def NodesIn_init(m, node):
-        d = {i: [] for i in m.Nodes}
-        for i, j in model.Arcs:
-            d[j].append[i]
+            d[j].append(i)
         return d
     model.NodesIn = pyo.Set(model.Nodes, initialize=NodesIn_init)
+
+    def NodesOut_init(m, node):
+        d = {i: [] for i in m.Nodes}
+        for i, j in model.Arcs:
+            d[i].append(j)
+        return d
+    model.NodesOut = pyo.Set(model.Nodes, initialize=NodesOut_init)
 
 This can also be done efficiently, and perhaps more clearly, using a
 :class:`BuildAction` (for more information, see :ref:`BuildAction`):
@@ -492,8 +494,8 @@ This can also be done efficiently, and perhaps more clearly, using a
 
 .. testcode::
 
-   model.NodesOut = pyo.Set(model.Nodes, within=model.Nodes)
    model.NodesIn = pyo.Set(model.Nodes, within=model.Nodes)
+   model.NodesOut = pyo.Set(model.Nodes, within=model.Nodes)
 
    def Populate_In_and_Out(model):
        # loop over the arcs and record the end points
--- a/doc/OnlineDocs/pyomo_modeling_components/Sets.rst
+++ b/doc/OnlineDocs/pyomo_modeling_components/Sets.rst
@@ -466,7 +466,7 @@ needed for the indexed set:
 
 .. testcode::
 
-    def NodesIn_init(m, node):
+    def NodesIn_init(m):
         # Create a dict to show NodesIn list for every node
         d = {i: [] for i in m.Nodes}
         # loop over the arcs and record the end points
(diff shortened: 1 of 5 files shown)


mfripp:
On my system, this is failing the style and unit tests, but that seems to be due to problems in the repository before I started. This pull request adds one more successful test and does not increase the number of failed tests.

mrmundt:
@mfripp - Please install the latest version of `black` and run `black -S -C` on the offending file. Thanks!

mfripp:
Thanks @mrmundt , I've done that. Looks like I accidentally used black 22.10.0 from elsewhere on my system instead of the newer one (24.8.0) I had just installed. The newer one seems happier with the Pyomo codebase overall and reformatted a few lines in this pull request, which I've now resubmitted.

mfripp:
@mrmundt, it looks like this pull request passed all tests except "Jenkins CI / test (pull request)". When I click for details on that, I get a message that "pyomo-jenkins.sandia.gov took too long to respond". Is there some way I can run this test locally or something I can fix to get this unblocked?

blnicho:
@mfripp our Jenkins infrastructure is currently down so you can ignore the failure for now. We'll rerun the tests when we get Jenkins running again.

mfripp:
The PR finished all the tests successfully after @jsiirola tagged it as `AT:RETEST`, but then I merged changes from Pyomo:main and it got stuck on `Jenkins CI / inspection` again.

mfripp:
Thanks! Hope it’s able to go through.

PR #3364, opened 2024-09-08
Greybox Objectives

michaelbynum:
## Summary/Motivation:
This PR adds support for objectives in PyNumero grey box models.

### Legal Acknowledgement


pyomo/contrib/pynumero/interfaces/pyomo_grey_box_nlp.py
```diff
@@ -364,11 +364,22 @@ def get_constraints_scaling(self):
...
-        return self._pyomo_nlp.evaluate_grad_objective(out=out)
+        ret = np.zeros(self.n_primals(), dtype=float)
+        for nlp in self._nlps:
+            ret += nlp.evaluate_grad_objective()
+
+        if out is not None:
+            ret.copyto(out)
+            return out
+
+        return ret
```
Robbybp, line 382: Do we want to sum multiple objectives? At first thought, it seems like we should only allow one objective to be implemented. If each grey box intends to implement an objective term, maybe the methods on `ExternalGreyBoxModel` should be called e.g. `evaluate_objective_term`.

Robbybp (commented):
I haven't reviewed changes in `_evaluate_hessian_if_necessary_and_cache` yet, but this looks fine.

Just curious, what is the motivation for doing this instead of just using outputs and putting them in the objective? Is there some motivating use case where this works better?

codykarcher:
> I haven't reviewed changes in `_evaluate_hessian_if_necessary_and_cache` yet, but this looks fine.
> 
> Just curious, what is the motivation for doing this instead of just using outputs and putting them in the objective? Is there some motivating use case where this works better?

This is me.  For applications I work with, it's common for people to just come with a black box and say "minimize".  Clearly that approach is not ideal, but allowing this capability will enable us to reach out to people who are used to doing optimization that way and help us convert them to the light side of the force...

pyomo/contrib/pynumero/interfaces/pyomo_grey_box_nlp.py
codykarcher: I agree with Robby in principle that I don't always want to sum my objectives.  But am happy to accept this as a first pass

codykarcher (approved):
Looks good, will test functionality over the next few weeks

codykarcher:
> > For applications I work with, it's common for people to just come with a black box and say "minimize"
> 
> Okay, so this is basically just for convenience.
> 
> > Clearly that approach is not ideal
> 
> I don't think it's clear at all. I'm usually a fan of eliminating auxiliary variables where possible, so this approach appeals to me. I was just curious if there was an example where it actually mattered for performance or reliability.

It is not ideal in the context of it hides structure I could potentially implement in my optimization problem.  IE, what if my black box is hiding a convex optimization problem?  But yea, it is a matter of convince, not performance.

mrmundt:
@michaelbynum , @Robbybp - We are hoping to cut a patch release on Monday (the 14th). Any chance this is going to make it into that?

mrmundt:
@michaelbynum , @Robbybp - Pinging on this PR again! Any update?

mrmundt (changes requested):

pyomo/contrib/pynumero/interfaces/pyomo_grey_box_nlp.py
michaelbynum: I modified things to raise an exception if there are multiple objectives.

pyomo/contrib/pynumero/examples/external_grey_box/external_with_objective.py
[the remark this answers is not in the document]
michaelbynum: Done.

pyomo/contrib/pynumero/interfaces/external_grey_box.py
```diff
@@ -315,6 +323,29 @@ def evaluate_jacobian_outputs(self):
...
+    def evaluate_grad_objective(self, out=None):
+        """
+        Compute the gradient of the objective from the
+        values set in input_values
+        """
+        raise NotImplementedError(
+            'evaluate_grad_objective called but not '
+            'implemented in the derived class.'
+        )
+
```
Robbybp, line 348: Remind me: Why don't we include default methods for `evaluate_hessian*` that raise `NotImplementedError`? Is the presence of such a method used to determine Hessian support at some point?

pyomo/contrib/pynumero/interfaces/pyomo_grey_box_nlp.py
```diff
@@ -100,8 +117,15 @@ def __init__(self, pyomo_model):
                     fixed_vars.extend(v for v in data.inputs.values() if v.fixed)
                     fixed_vars.extend(v for v in data.outputs.values() if v.fixed)
                     greybox_nlp = _ExternalGreyBoxAsNLP(data)
+                    if greybox_nlp._ex_model.has_objective():
```
Robbybp, line 120: Should this be `data.get_external_model().has_objective()`?

pyomo/contrib/pynumero/interfaces/pyomo_grey_box_nlp.py
Robbybp: Can you add a comment here noting that we've already checked to make sure there's only one objective?

pyomo/contrib/pynumero/interfaces/pyomo_grey_box_nlp.py
```diff
@@ -364,11 +388,22 @@ def get_constraints_scaling(self):
     # overloaded from NLP
     def evaluate_objective(self):
         # objective is owned by the pyomo model
-        return self._pyomo_nlp.evaluate_objective()
+        obj = 0
+        for nlp in self._nlps:
+            obj += nlp.evaluate_objective()
+        return obj
```
Robbybp, line 394: Can you add a comment noting that we've already checked to make sure there's only one objective?

Robbybp (approved):
Some minor comments. Otherwise, this looks good to me.

pyomo/contrib/pynumero/interfaces/external_grey_box.py
michaelbynum: Looks like it: https://github.com/Pyomo/pyomo/blob/e75a1be9b4ee5e6e3b45beb2f27e21bb154ba5ef/pyomo/contrib/pynumero/interfaces/pyomo_grey_box_nlp.py#L578. If we want to change that, I can create an issue.

pyomo/contrib/pynumero/interfaces/pyomo_grey_box_nlp.py
michaelbynum: absolutely.

michaelbynum: done

pyomo/contrib/pynumero/interfaces/external_grey_box.py
Robbybp: I prefer `try`/`except NotImplementedError` to using `hasattr`, but I'm not sure it's important enough to change.

pyomo/contrib/pynumero/examples/external_grey_box/external_with_objective.py
```diff
@@ -0,0 +1,162 @@
+#  ___________________________________________________________________________
+#
+#  Pyomo: Python Optimization Modeling Objects
+#  Copyright (c) 2008-2024
```
mrmundt, line 4: Don't hate me... Can you please change this to 2025?

mrmundt (approved):

PR #3368, opened 2024-09-16
Resolve issue in filter/validate deprecation path

jsiirola:
## Summary/Motivation:
@emma58 pointed out a bug in the Set filter/validate deprecation path where Scalar sets with dim=2 would silently not call filter correctly (as the deprecation path required the filter function to raise an exception or return false when called with the new syntax. This PR resolves the error and adds a test for this scenario. 

In addition, @mfripp pointed out on SO that we were being rather rude deprecating an old syntax in the same release that we added support for the new API. This PR silences the deprecation warning for the (common) case where filter / validate was being used on a Scalar Set. 

- fix filter/validate deprecation path for ScalarSets
- add a test for the edge case that motivated this PR
- Silence (for now) the deprecation warning for scalar sets. 

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/tests/unit/test_set.py
+++ b/pyomo/core/tests/unit/test_set.py
@@ -4181,6 +4181,19 @@ class TestSet(unittest.TestCase):
         self.assertIs(type(m.I[3]), InsertionOrderSetData)
         self.assertEqual(m.I.data(), {1: (4, 2, 5), 2: (4, 2, 5), 3: (4, 2, 5)})
 
+        # Explicit (constant dict) construction
+        m = ConcreteModel()
+        m.I = Set([1, 2], initialize={1: (4, 2, 5), 2: (7, 6)})
+        self.assertEqual(len(m.I), 2)
+        self.assertEqual(list(m.I[1]), [4, 2, 5])
+        self.assertEqual(list(m.I[2]), [7, 6])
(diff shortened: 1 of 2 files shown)


pyomo/core/tests/unit/test_set.py
```diff
@@ -4374,12 +4399,15 @@ def _validate(model, i, j):
...
-                r"callback to match the signature \(block, value\)",
-            )
+            # Note that we are not emitting a deprecation warning (yet)
+            # for scalar sets
+            # self.assertEqual(output.getvalue(), "")
+            #     output.getvalue().replace('\n', ' '),
+            #     r"DEPRECATED: OrderedScalarSet I2: 'validate=' callback "
+            #     r"signature matched \(block, \*value\).  Please update the "
+            #     r"callback to match the signature \(block, value\)",
+            # )
```
mrmundt, line 4409: Is this still supposed to be here?

jsiirola: Sorry - this was leaving a breadcrumb (with no explanation).  Eventually we will (probably?) want to turn the deprecation warning back on, in which case this test will have to revert back.

emma58 (approved):
Thank you for fixing this! :)

blnicho (approved):

PR #3384, opened 2024-10-16
Move away from dependence on gurobi.sh

### LINKED ISSUE #3285: Gurobi interactive shell scripts are deprecated
## Summary

The Gurobi interactive shell (`gurobi.sh` / `gurobi.bat`) is being deprecated. In the next major release these scripts will emit a deprecation warning, and in a future release they will be removed entirely from the Gurobi installers. I'm opening this issue to discuss if any changes are needed to avoid breaking pyomo for users of the file-based interface.

## Description

There is some nice work already done so that the file-based interface would invoke the current python executable if it has gurobipy installed:
- https://github.com/Pyomo/pyomo/pull/2262
- https://github.com/Pyomo/pyomo/issues/2132
- https://github.com/Pyomo/pyomo/issues/1776

There would still be a few changes to consider:
1. Update installation instructions to clarify that a gurobipy installation is needed and the full Gurobi package does not need to be installed
2. Consider changing the default behaviour to use a gurobipy installation in the current interpreter, or the direct interface
3. I suppose PyPy users rely more on the Gurobi shell since gurobipy doesn't support PyPy. Is another option needed there?

Just wanted to bring this up early in case it creates compatibility issues. I'm happy to work on any necessary changes.

jsiirola:
## Summary/Motivation:
Gurobi has deprecated the use of the `gurobi.sh` script, which is used by the default Pyomo interface to Gurobi.  While we could have just changed the default interface to `gurobi_direct`, this PR does not go down that path because:
- `gurobi_direct` is significantly less performant than the current LP shell interface
- the `gurobi_direct` has a slightly different API
- we want to retire / replace all the solver interfaces in `pyomo.solvers` "soon"

Instead, this PR adds a new solver that is a relatively thin veneer around the `GUROBISHELL` interface that avoids the subprocess call and the need to write results to a "soln" file.  While this is adding yet another interface, it offers a minimally-disruptive path to avoid the deprecation warnings that will appear in Gurobi 12.

- Refactor GUROBI_RUN so that the bulk of the functionality can be executed directly (pulling out the soln file writer to a separate function)
- Add GUROBIFILE as a LP/MPS file-based input, direct interface output solver interface that has maximal overlap with the current GUROBISHELL.
- Remove GHA pin of Gurobi to 10.0.3
- Update tests to "not fail" in Gurobi 11

### Legal Acknowledgement


pyomo/solvers/plugins/solvers/GUROBI.py
```diff
@@ -588,3 +597,229 @@ def _postsolve(self):
...
+    def _get_version(self):
+        return (
+            gurobipy.GRB.VERSION_MAJOR,
+            gurobipy.GRB.VERSION_MINOR,
+            gurobipy.GRB.VERSION_TECHNICAL,
+        )
+
+    def _default_executable(self):
+        # Bogus, but not None (because the test infrastructure disables
+        # solvers where the executable() is None
```
mrmundt, line 645: Very small nitpick
```suggestion
        # solvers where the executable() is None)
```

pyomo/solvers/plugins/solvers/GUROBI.py
```diff
@@ -588,3 +597,229 @@ def _postsolve(self):
...
+        return ""
+
+    def create_command_line(self, executable, problem_files):
+        #
+        # Define log file
+        # The log file in CPLEX contains the solution trace, but the
```
mrmundt, line 651: Is this comment wrong? Why mention CPLEX?

pyomo/solvers/plugins/solvers/GUROBI.py
```diff
@@ -588,3 +597,229 @@ def _postsolve(self):
...
+        #
+        if is_debug_set(logger):
+            logger.debug("Running %s", self._command.cmd)
+
+        problem_filename = self._problem_files[0]
+        warmstart_filename = self._warm_start_file_name
+
+        # translate the options into a normal python dictionary, from a
+        # pyutilib SectionWrapper - the gurobi_run function doesn't know
+        # about pyomo, so the translation is necessary.
```
mrmundt, line 674: This comment also appears to be wrong.

pyomo/solvers/plugins/solvers/GUROBI.py
```diff
@@ -588,3 +597,229 @@ def _postsolve(self):
...
+            self._soln = GUROBI_RUN.gurobi_run(
+                problem_filename, warmstart_filename, None, options_dict, self._suffixes
+            )
+        self._log = ostreams[0].getvalue()
+        self._rc = 0
+        sys.stdout.flush()
+        return Bunch(rc=self._rc, log=self._log)
+
+    def process_soln_file(self, results):
+        # the only suffixes that we extract from CPLEX are
```
mrmundt, line 701: Same question about CPLEX in comment here.

pyomo/solvers/plugins/solvers/GUROBI.py
jsiirola: No - this is still correct.  I updated the comment to clarify it a bit.

mrmundt (approved):

jsiirola:
The test failures in Jenkins are real; however, I need #3394 to resolve the issue.

blnicho (approved):
I have a couple questions but nothing major enough to prevent merging

pyomo/gdp/plugins/multiple_bigm.py
```diff
@@ -127,7 +135,8 @@ class MultipleBigMTransformation(GDP_to_MIP_Transformation, _BigM_MixIn):
     CONFIG.declare(
         'solver',
         ConfigValue(
-            default=SolverFactory('gurobi'),
```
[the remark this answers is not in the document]
blnicho: Oops, I just noticed the last commit message which seems to already be using the deferred config stuff so ignore my original question. Is the `Solver` domain implemented here general enough to go into `pyomo.common` or `pyomo.util`?

pyomo/solvers/plugins/solvers/GUROBI.py
```diff
@@ -588,3 +594,233 @@ def _postsolve(self):
...
+        # Execute the command
+        #
+        if is_debug_set(logger):
+            logger.debug("Running %s", self._command.cmd)
+
+        problem_filename = self._problem_files[0]
+        warmstart_filename = self._warm_start_file_name
+
+        # translate the options into a normal python dictionary, from a
+        # pyutilib SectionWrapper - because the gurobi_run function was
```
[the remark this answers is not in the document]
mrmundt: I am pretty sure I asked this same question!

pyomo/gdp/plugins/multiple_bigm.py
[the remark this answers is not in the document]
jsiirola: It is pretty general, but could use some discussion before promulgating it elsewhere.  In particular, what is the "correct" way to know if you were handed a "solver" object?  Given that we have 3 different base classes floating around (core, APPSI, and contrib/solver), inheritance is probably not the right answer.  This implementation just looks for a `solve()` method and hopes it quacks.

pyomo/solvers/plugins/solvers/GUROBI.py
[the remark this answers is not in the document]
jsiirola: Not quite.  `options` is a `Bunch` (which used to be a PyUtilib class).  The translation is necessary because `repr(Bunch)` doesn't produce meaningful python code.  Comment updated (and code simplified).

PR #3401, opened 2024-11-04
Add Python 3.13 to Testing Infrastructure

mrmundt:
## Summary/Motivation:
We want to be as updated as possible with our Python versions.

NOTE: We will be removing Python 3.8 support after the next release.

- Adds Python 3.13

### Legal Acknowledgement


pyomo/contrib/viewer/qt.py
```diff
@@ -38,17 +38,19 @@
...
+    for module_str in supported:
+        try:
+            qt_package = importlib.import_module(module_str)
+            QtWidgets = importlib.import_module(f"{module_str}.QtWidgets")
+            QtCore = importlib.import_module(f"{module_str}.QtCore")
+            QtGui = importlib.import_module(f"{module_str}.QtGui")
+            available = module_str
+            break
+        except Exception as e:
+            import_errors.append(f"{e}")
```
jsiirola, line 53: Why are we marking "qt" as not available due to pint?  I would rather see us leave this as-is, and just skip the tests that rely on pint if pint is not available.

mrmundt: Changed

PR #3403, opened 2024-11-06
Catch catastrophic solver failure when building MIS

bknueven:
## Summary/Motivation:
The minimal intractable system calculation had a try/except around a solve we expect to fail, but when we check for failure we failed to first check the results object against `None`.

- Insert `results is not None` conditional to short-circuit, thus not calling `pyo.check_optimal_termination(results)` if `results is None`.

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/iis/mis.py
+++ b/pyomo/contrib/iis/mis.py
@@ -287,7 +287,7 @@ def compute_infeasibility_explanation(
     except:
         results = None
 
-    if pyo.check_optimal_termination(results):
+    if (results is not None) and pyo.check_optimal_termination(results):
         msg += "Could not determine Minimal Intractable System\n"
     else:
         deletion_filter = []


mrmundt (approved):

pyomo/contrib/iis/mis.py
```diff
@@ -287,7 +287,7 @@ def _constraint_generator():
     except:
         results = None
 
-    if pyo.check_optimal_termination(results):
+    if (results is not None) and pyo.check_optimal_termination(results):
```
jsiirola, line 290: Question: would it better to not swallow all exceptions from the `solve()` above and instead use `load_solutions=False` in the call?  Then results should never be `None`?

bknueven: Probably. Would that still not raise an exception if the solver returns a non-zero exit code? I think that's why I've implemented these as blanket `try / except` blocks.

jsiirola: OK.  I would just be worried about the bare `except:` -- that could be suppressing any number of errors and there would be no way to know what was going on.

jsiirola (approved):
I am OK merging this as it resolves a serious error blocking MIS users.  However, this highlights fragility in the MIS implementation due to the use of a bare (catch-all) "except:" clause.  

It would be good to revisit this and see if the bare "except:" can be removed.

PR #3439, opened 2024-11-30
Extend valid types for PyROS solver argument `uncertain_params`

shermanjasonaf:
## Summary/Motivation:

Motivated by recent user feedback, this PR adds support for the speciifcation of `Var` and `VarData` objects as (entries of) the argument `uncertain_params` to `contrib.pyros.PyROS.solve()`. Thus, the argument `uncertain_params` can now be a `Param`, `ParamData`, `Var`, or `VarData` instance, or an iterable sequence of such instances. All `VarData` objects derived from `uncertain_params` should be fixed by the user in advance of invoking PyROS.

Note that if `uncertain_params` is of type `Var`/`VarData`, or otherwise has `Var` or `VarData` entries, then there may be an appreciable computational overhead during the PyROS solver preprocessing step for models with large numbers of `Constraint`, `Expression`, and/or `Objective` components in which the `Var`/`VarData` objects appear.

- Allow PyROS solver to support `Var` and `VarData` objects in `uncertain_params` argument

## TODO (after #3434 merged)
- [x] Update version number and changelog
- [x] Add bullet on preprocessing progress logging in [logging output explanation table](https://pyomo.readthedocs.io/en/6.8.2/explanation/solvers/pyros.html#id6)?

### Legal Acknowledgement


blnicho (approved):

pyomo/contrib/pyros/util.py
```diff
@@ -1598,6 +1707,71 @@ def turn_adjustable_var_bounds_to_constraints(model_data):
...
+        Mapping from VarData objects to be replaced
+        to the ParamData objects to be introduced.
+    """
+    # invoke the expression replacement method for each
+    # individual component type to ensure the substitutions
+    # are always performed in the named expressions first
+    for ctype in (Expression, Constraint, Objective):
+        _replace_vars_in_component_exprs(
+            block=block, substitution_map=var_to_param_map, ctype=ctype
+        )
```
jsiirola, line 1772: This doesn't need to block this PR, but this could be made more efficient by passing the tuple `(Expression, Constraint, Objective)` directly to `component_data_objects`.  That would remove the need for the loop here and allow us to only walk the model structure once.  That is, on 1734 above:
```python
for cdata in block.component_data_objects((Expression, Constraint, Objective), active=None, descend_into=True):
```

pyomo/contrib/pyros/util.py
```diff
@@ -1654,6 +1828,47 @@ def setup_working_model(model_data, user_var_partitioning):
     delattr(original_model, temp_util_block_attr_name)
     delattr(working_model.user_model, temp_util_block_attr_name)
 
+    uncertain_param_var_idxs = []
+    for idx, obj in enumerate(working_model.orig_uncertain_params):
+        if isinstance(obj, VarData):
+            obj.fix()
+            uncertain_param_var_idxs.append(idx)
```
jsiirola, line 1835: Do you need to make sure that the Var has a value here?  We allow Vars with lower == upper to be uncertain Params even if they don't have a current (numeric) value, correct?

jsiirola (commented):
One question that should probably be addressed before merging.  The other two are enhancements, but not strictly necessary.

pyomo/contrib/pyros/util.py
jsiirola: Great! Thanks for clarifying

jsiirola (approved):

PR #3440, opened 2024-12-02
Disallow calling `del_component` with ComponentData arguments

### LINKED ISSUE #3435: Calling `del_component` on a ComponentData deletes the parent component
## Summary

Calling `del_component` on a ComponentData (at least for `Constraint` and `Var`) deletes the entire parent component.

This script:
```
from pyomo.environ import *

m = ConcreteModel()
m.x = Var(range(10))
@m.Constraint(range(10))
def c(m, i):
    return m.x[i] <= i

m.pprint()

print("Delete c[5]?")
m.del_component(m.c[5])

m.pprint()

print("Delete x[3]?")
m.del_component(m.x[3])

m.pprint()
```
Produces this output:
```
1 Var Declarations
    x : Size=10, Index={0, 1, 2, 3, 4, 5, 6, 7, 8, 9}
          0 :  None :  None :  None : False :  True :  Reals
          1 :  None :  None :  None : False :  True :  Reals
          2 :  None :  None :  None : False :  True :  Reals
          3 :  None :  None :  None : False :  True :  Reals
          4 :  None :  None :  None : False :  True :  Reals
          5 :  None :  None :  None : False :  True :  Reals
          6 :  None :  None :  None : False :  True :  Reals
          7 :  None :  None :  None : False :  True :  Reals
          8 :  None :  None :  None : False :  True :  Reals
          9 :  None :  None :  None : False :  True :  Reals

1 Constraint Declarations
    c : Size=10, Index={0, 1, 2, 3, 4, 5, 6, 7, 8, 9}, Active=True
        Key : Lower : Body : Upper : Active
          0 :  -Inf : x[0] :   0.0 :   True
          1 :  -Inf : x[1] :   1.0 :   True
          2 :  -Inf : x[2] :   2.0 :   True
          3 :  -Inf : x[3] :   3.0 :   True
          4 :  -Inf : x[4] :   4.0 :   True
          5 :  -Inf : x[5] :   5.0 :   True
          6 :  -Inf : x[6] :   6.0 :   True
          7 :  -Inf : x[7] :   7.0 :   True
          8 :  -Inf : x[8] :   8.0 :   True
          9 :  -Inf : x[9] :   9.0 :   True

2 Declarations: x c
Delete c[5]?
1 Var Declarations
    x : Size=10, Index={0, 1, 2, 3, 4, 5, 6, 7, 8, 9}
          0 :  None :  None :  None : False :  True :  Reals
          1 :  None :  None :  None : False :  True :  Reals
          2 :  None :  None :  None : False :  True :  Reals
          3 :  None :  None :  None : False :  True :  Reals
          4 :  None :  None :  None : False :  True :  Reals
          5 :  None :  None :  None : False :  True :  Reals
          6 :  None :  None :  None : False :  True :  Reals
          7 :  None :  None :  None : False :  True :  Reals
          8 :  None :  None :  None : False :  True :  Reals
          9 :  None :  None :  None : False :  True :  Reals

1 Declarations: x
Delete x[3]?
0 Declarations: 
```

### Error Message

No error--we're turfing the entire IndexedComponent silently.

Pyomo version: main
Python version: 3.11
Operating system: linux
Solver (if applicable): NA

emma58:
## Summary/Motivation:

I think silently deleting the containing component when we call `del_component` on a ComponentData is rude, and the post-processing in Fourier-Motzkin elimination was falling victim to this mistake. This PR adds a `ValueError` to `del_component` if it is given a ComponentData and revises the FME transformation post-processing routine.

- Sneaks in a documentation fix for ComponentData
- Adds error to `del_component` if the arguments is a ComponentData
- Revises FME post-processing routine to not misuse `del_component` and, in fact, to not add and delete as many components in general.

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/fme/fourier_motzkin_elimination.py
+++ b/pyomo/contrib/fme/fourier_motzkin_elimination.py
@@ -730,12 +730,9 @@ class Fourier_Motzkin_Elimination_Transformation(Transformation):
                 continue
             # deactivate the constraint
             projected_constraints[i].deactivate()
-            m.del_component(obj)
-            # make objective to maximize its infeasibility
-            obj = Objective(
-                expr=projected_constraints[i].body - projected_constraints[i].lower
-            )
-            m.add_component(obj_name, obj)
+            # Our constraint looks like: 0 <= a^Tx - b, so make objective to
+            # maximize its infeasibility
+            obj.expr = projected_constraints[i].body - projected_constraints[i].lower
             results = solver_factory.solve(m)
             if results.solver.termination_condition == TerminationCondition.unbounded:
                 obj_val = -float('inf')
@@ -753,13 +750,13 @@ class Fourier_Motzkin_Elimination_Transformation(Transformation):
                 obj_val = value(obj)
             # if we couldn't make it infeasible, it's useless
             if obj_val >= tolerance:
-                m.del_component(projected_constraints[i])
                 del projected_constraints[i]
             else:
                 projected_constraints[i].activate()
 
         # clean up
         m.del_component(obj)
+        del obj
         for obj in active_objs:
             obj.activate()
         # undo relax integrality
--- a/pyomo/core/base/block.py
+++ b/pyomo/core/base/block.py
@@ -34,6 +34,7 @@ from pyomo.common.pyomo_typing import overload
 from pyomo.common.timing import ConstructionTimer
 from pyomo.core.base.component import (
     Component,
+    ComponentData,
     ActiveComponentData,
     ModelComponentFactory,
 )
@@ -1139,6 +1140,14 @@ component, use the block del_component() and add_component() methods.
         #    return
 
         name = obj.local_name
+        if isinstance(obj, ComponentData) and not isinstance(obj, Component):
+            raise ValueError(
+                "Argument '%s' to del_component is a ComponentData object. Please "
+                "use the Python 'del' function to delete members of indexed "
+                "Pyomo objects. The del_component function can only be used to "
+                "delete IndexedComponents and ScalarComponents." % name
+            )
+
         if name in self._Block_reserved_words:
             raise ValueError(
                 "Attempting to delete a reserved block component:\n\t%s" % (obj.name,)
--- a/pyomo/core/base/component.py
+++ b/pyomo/core/base/component.py
@@ -732,11 +732,7 @@ class ComponentData(ComponentBase):
     This is the base class for the component data used
     in Pyomo modeling components.  Subclasses of ComponentData are
     used in indexed components, and this class assumes that indexed
-    components are subclasses of IndexedComponent.  Note that
-    ComponentData instances do not store their index.  This makes
-    some operations significantly more expensive, but these are (a)
-    associated with I/O generation and (b) this cost can be managed
-    with caches.
+    components are subclasses of IndexedComponent.
 
     Constructor arguments:
         owner           The component that owns this data object


pyomo/contrib/fme/fourier_motzkin_elimination.py
```diff
@@ -753,13 +750,13 @@ def post_process_fme_constraints(
...
             # if we couldn't make it infeasible, it's useless
             if obj_val >= tolerance:
-                m.del_component(projected_constraints[i])
                 del projected_constraints[i]
             else:
                 projected_constraints[i].activate()
 
         # clean up
         m.del_component(obj)
+        del obj
```
jsiirola, line 759: This is unnecessary (it implicitly happens when you reassign `obj` in the for loop in the next line)

pyomo/core/base/block.py
```diff
@@ -1139,6 +1140,14 @@ def del_component(self, name_or_object):
...
 
         name = obj.local_name
+        if isinstance(obj, ComponentData) and not isinstance(obj, Component):
+            raise ValueError(
+                "Argument '%s' to del_component is a ComponentData object. Please "
+                "use the Python 'del' function to delete members of indexed "
+                "Pyomo objects. The del_component function can only be used to "
+                "delete IndexedComponents and ScalarComponents." % name
+            )
+
```
jsiirola, line 1150: This doesn't do what you think it does: line 1133 (`obj = self.component(name_or_object)`) maps the incoming object into its component, so you shouldn't be able to trigger this exception.  If we are going to change the behavior of `del_component` to only admit Components (either by name or reference), we should
- change (possibly inline the logic from) line 1133
- document this in the docstring
- possibly add a deprecation path (because we are changing behavior that has been in Pyomo for over a decade)??

jsiirola (changes requested):
I am pretty sure that the change to `del_component` won't actually catch attempts to delete a component data.

pyomo/core/base/block.py
michaelbynum: I would argue that this is a bug and does not need a deprecation path.

michaelbynum: I think `block.component()` only works with strings if the string refers to a component and not a component data, so we can probably just check if `name_or_object` is a string?

mrmundt:
We are probably not going to aim to merge this before the patch release next week (because it's fairly intricate / could break some folks). We don't want to sneak this in without sufficient planning/warning.

blnicho (approved):

pyomo/core/base/block.py
```diff
@@ -1125,16 +1126,34 @@ def del_component(self, name_or_object):
...
+                        "ScalarComponents." % name_or_object.local_name
+                    )
+                if obj.parent_block() is not self:
+                    obj = None
+            except AttributeError:
+                pass
         if obj is None:
+            # Maintaining current behavior, but perhaps this should raise an
+            # exception?
             return
```
jsiirola: This is a minor nit, but this is a lot of code where we might be throwing away AttributeErrors.  I think I would prefer something more targeted.  Maybe:

```suggestion
        if isinstance(name_or_object, str):
            if name_or_object in self._decl:
                obj = self._decl_order[self._decl[name_or_object]][0]
            else:
                # Maintaining current behavior, but perhaps this should raise an
                # exception?
                return
        else:
            try:
                obj = name_or_object.parent_component()
            except AttributeError:
                # Maintaining current behavior, but perhaps this should raise an
                # exception?
                return
            if obj is not name_or_object:
                raise ValueError(
                    f"Argument '{name_or_object.local_name}' to del_component is "
                    "a ComponentData object.  Please use the Python 'del' function to "
                    "delete members of indexed Pyomo components. The del_component "
                    "function can only be used to delete entire Scalar or Indexed Components."
                )
            if obj.parent_block() is not self:
                return
```

jsiirola (approved):
One minor question, but overall this looks good.

emma58:
This is ready to merge when tests pass.

PR #3442, opened 2024-12-05
Add option to skip fixing inputs in sequential decomposition

alma-walmsley:
## Summary/Motivation:

I am working on a problem (within `IDAES`) where pyomo network's sequential decomposition is used to initialize the model. For tear ports, I want to either:
- Fix the variables on that port directly, or
- Remove a degree of freedom by adding a constraint for an expression at that port.

For example, if the port has the variables `[flow_mol, pressure, enth_mol]`, and `temperature` is an expression of `pressure` and `enth_mol`, then I want the ability to unfix `enth_mol` and constrain `temperature` to some value to determine the enthalpy.

In pyomo network's case, when running sequential decomposition, all the variables at the port get fixed if they are currently unfixed. I assume this is to prevent user error from too many degrees of freedom. 

However, in my case, my model gets over-defined, since a constraint is added to remove a degree of freedom when a variable is unfixed, but those variables get fixed anyway when sequential decomposition is run.

- Add an option in sequential decomposition to skip fixing all the variables at the tear ports. Defaults to False. Variables can be fixed manually or through the `set_guesses_for` method.

### Legal Acknowledgement

### DIFF
--- a/pyomo/network/decomposition.py
+++ b/pyomo/network/decomposition.py
@@ -150,6 +150,13 @@ class SequentialDecomposition(FOQUSGraph):
             Keyword options to pass to solve method.
 
             `default={}`
+        
+        skip_fixing_inputs: `bool`
+            Skip fixing all the variables on the tear ports. Require the user
+            to fix the port variables so that 0 degrees of freedom are
+            maintained.
+
+            `default=False`
     """
 
     def __init__(self, **kwds):
@@ -176,6 +183,7 @@ class SequentialDecomposition(FOQUSGraph):
         options["tear_solver"] = "cplex"
         options["tear_solver_io"] = None
         options["tear_solver_options"] = {}
+        options["skip_fixing_inputs"] = False
 
         options.update(kwds)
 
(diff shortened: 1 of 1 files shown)


mrmundt (approved):

alma-walmsley:
Had a review of what I was trying to accomplish here. Constraints can't be applied if the state variable is a "guess" and free to be adjusted by a recycle. They can be applied if the state variable should remain constant (or rather, there is nothing in the recycle loop that can adjust the state variable - flow_mass is a good example: the arc's equality constraint for flow_mol needs to be deactivated, and a constraint for flow_mass at the state block should be added). However, to support guesses from constraints for state variables that should be "recycled," we should first initialize the state block, which actually gives us the correct value for flow_mol, which can be fixed - and all constraints deactivated.

At best we could probably save on initializing one state block, which in the grand scheme of things is not much and probably not worth it.

TL;DR this logic is probably not needed, closing this PR :slightly_smiling_face:

PR #3455, opened 2025-01-16
Testing: exclude Xpress 9.5.1 on Windows/GHA/Python3.{0,1}

jsiirola:
## Summary/Motivation:

The Xpress 9.5.1 release through the `fico-xpress` conda channel hangs when calling `xpress.init()` -- but only on Python 3.10 and 3.11 / Windows GHA.  (Quick) attempts to reproduce this locally were not successful, so it is unclear if this is a GHA issue, Conda issue, an Xpress issue, or an issue with our configuration of the build environment.

@djunglas, is there a good way to report things like this upstream?

This PR gets the Pyomo testing infrastructure "working" again by simply pinning xpress to 9.5.0 under Conda on Windows for Python 3.10 and 3.11.  I attempted to make the version pin "float" (and automatically pick up any future Xpress releases), but that was not successful (see comments in the workflows).  We should monitor upstream and revisit if this is still needed after the next Xpress release.

- Pin xpress to 9.5.0 on Windows under conda for Python 3.10 and 3.11 (only)
- Do some minor clean up of the GHA drivers

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/test_branches.yml
+++ b/.github/workflows/test_branches.yml
@@ -377,6 +377,7 @@ jobs:
             for PKG in 'cplex>=12.10' docplex gurobi "$XPRESS" cyipopt pymumps scip; do
                 echo ""
                 echo "*** Install $PKG ***"
+                echo ""
                 # conda can literally take an hour to determine that a
                 # package is not available.  Perform a quick search to see
                 # if the package is available for this interpreter before
@@ -398,6 +399,7 @@ jobs:
                         conda install -y "$PKG" || _BUILDS=""
                     fi
                 fi
+                echo ""
                 if test -z "$_BUILDS"; then
                     echo "WARNING: $PKG is not available"
                 fi
--- a/.github/workflows/test_branches.yml
+++ b/.github/workflows/test_branches.yml
@@ -71,29 +71,29 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        python: ['3.13']
+        python: [3.13]
         other: [""]
         category: [""]
 
         include:
         - os: ubuntu-latest
-          python: '3.13'
+          python: 3.13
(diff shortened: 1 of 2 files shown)


djunglas:
Hi @jsiirola, this problem is new to us. We will try to reproduce and investigate.
If you find problems that you think are related to Xpress then it is probably best to email me directly.
Another option would be to report them on stackoverflow (with tag `xpress-optimizer`) or in [our community](https://community.fico.com/s/ask-a-question/xpress-optimization).

.github/workflows/test_branches.yml
```diff
@@ -71,29 +71,29 @@ jobs:
       fail-fast: false
```
mrmundt, line 71: There is now _NO_ Python 3.9 testing in this file, and I think that's an issue. Maybe win/pip should have it / cover it?

.github/workflows/test_pr_and_main.yml
```diff
@@ -113,14 +113,14 @@ jobs:
           PACKAGES: cython
 
         - os: windows-latest
-          python: 3.9
+          python: '3.10'
```
mrmundt, line 116: ... Let me guess. `3.9` is fine, `3.11` is fine, but `'3.10'` has to have strings?

mrmundt (changes requested):

.github/workflows/test_pr_and_main.yml
jsiirola: Yup - GHA interprets things that look like numbers as numbers unless you explicitly make them strings.  Without the quotes, 3.10 looks like (and becomes) 3.1.

mrmundt: _of course_

mrmundt (approved):

PR #3464, opened 2025-02-02
Allow construction of CUID from another CUID

Robbybp:
## Fixes
`ComponentUID(ComponentUID("x[1]"))` raises an error.

## Summary/Motivation:
When processing keys corresponding to components, e.g., in `find_component` or `contrib.mpc.get_indexed_cuid`, we sometimes want to convert a key into a CUID so it can be handled more consistently. However, we have to handle separately the case where the key is already a CUID. This PR proposes to allow construction of a CUID from another CUID.

- If the argument to `ComponentUID` is another `ComponentUID`, we just copy the `_cids` attribute onto the new instance.

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/componentuid.py
+++ b/pyomo/core/base/componentuid.py
@@ -12,6 +12,7 @@
 import codecs
 import re
 import ply.lex
+import copy
 
 from pyomo.common.collections import ComponentMap
 from pyomo.common.dependencies import pickle
@@ -86,7 +87,8 @@ class ComponentUID(object):
                 self._cids = tuple(self._parse_cuid_v2(component))
             except (OSError, IOError):
                 self._cids = tuple(self._parse_cuid_v1(component))
-
+        elif isinstance(component, ComponentUID):
+            self._cids = copy.deepcopy(component._cids)
         elif type(component) is IndexedComponent_slice:
             self._cids = tuple(
                 self._generate_cuid_from_slice(component, context=context)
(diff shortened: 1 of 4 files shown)


pyomo/core/base/componentuid.py
```diff
@@ -76,17 +77,21 @@ class ComponentUID(object):
...
-                    "Context is not allowed when initializing a "
-                    "ComponentUID object from a string type"
-                )
+                _context_err(str)
             try:
                 self._cids = tuple(self._parse_cuid_v2(component))
             except (OSError, IOError):
                 self._cids = tuple(self._parse_cuid_v1(component))
-
+        elif isinstance(component, ComponentUID):
```
Robbybp, line 91: In `find_component`, we used `type is ComponentUID` instead of `isinstance`, presumably for performance. Should we do the same here?

pyomo/core/base/componentuid.py
```diff
@@ -76,17 +77,21 @@ class ComponentUID(object):
...
+            if context is not None:
+                _context_err(ComponentUID)
+            self._cids = copy.deepcopy(component._cids)
```
Robbybp, line 94: IIRC, `_cids` shouldn't contain anything mutable, so we may not need the copy here.

Robbybp (commented):
Some implementation questions.

pyomo/core/base/componentuid.py
jsiirola: I believe that is correct, and would be in favor of the change.

jsiirola (commented):
A couple super minor changes that would be nice to include and this should be good to go.

jsiirola (approved):
Approved, pending successful tests.

blnicho (approved):

PR #3476, opened 2025-02-18
Solver Refactor Pt. 2: Restructure for Future Redesign

### LINKED ISSUE #1030: [PEP] Redesign of Pyomo Solvers

mrmundt:
## Summary/Motivation:

This is meant to be a "mid-point" for the ongoing solver refactor. It is currently working in this form, so this PR is intended to merge changes up to now so we can start making more fine-tuned changes that are easier to track.

- Introduce highs
- Restructure the directory tree
- Rework some tests for clarity
- Reduce duplicate code
- Apply static analysis suggestions (`pylint`)

### Legal Acknowledgement


pyomo/contrib/solver/common/base.py
```diff
@@ -29,17 +28,41 @@
...
+    NeedsCompiledExtension = -3
+
+    def __bool__(self):
+        return self._value_ > 0
+
+    def __format__(self, format_spec):
+        return format(self.name, format_spec)
+
+    def __str__(self):
+        return self.name
```
michaelbynum, line 62: When I saw this, I thought, "I'm surprised we need this". Then I looked at the version that was removed from within `SolverBase` and saw that we dropped some useful comments. Why not port those comments? Also, I still find it odd that we need to override `__str__` and `__format__`. Should we use a different type of enum?

michaelbynum: Oh, we need `IntEnum` for `__bool__`...

pyomo/contrib/solver/common/config.py
```diff
@@ -334,24 +323,6 @@ def __init__(
...
-                With the default setting of True, fixed variables will be treated like parameters. 
-                This means that z == x*y will be linear if x or y is fixed and the constraint 
-                can be written to an LP file. If the value of the fixed variable gets changed, we have 
-                to completely reprocess all constraints using that variable. If 
-                treat_fixed_vars_as_params is False, then constraints will be processed as if fixed 
-                variables are not fixed, and the solver will be told the variable is fixed. This means 
-                z == x*y could not be written to an LP file even if x and/or y is fixed. However, 
-                updating the values of fixed variables is much faster this way.""",
-            ),
-        )
```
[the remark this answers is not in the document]
michaelbynum: Oh, I see. We put this in the constructor, which makes way more sense.

pyomo/contrib/solver/common/persistent.py
```diff
@@ -331,7 +337,7 @@ def update_parameters(self):
     def update(self, timer: HierarchicalTimer = None):
         if timer is None:
             timer = HierarchicalTimer()
-        config = self.config.auto_updates
+        config = self._active_config.auto_updates
```
michaelbynum, line 340: Good catch!

pyomo/contrib/solver/common/persistent.py
```diff
@@ -494,3 +498,41 @@ def update(self, timer: HierarchicalTimer = None):
         timer.start('vars')
         self.remove_variables(old_vars)
         timer.stop('vars')
+
+
+class PersistentSolverMixin:
+    """
+    The `solve` method in Gurobi and Highs is exactly the same, so this Mixin
+    minimizes the duplicate code
+    """
```
michaelbynum, line 507: Nice!

pyomo/contrib/solver/common/results.py
```diff
@@ -328,17 +284,53 @@ def display(
...
+    - `LegacySolutionStatus.bestSoFar` -> `TerminationCondition.convergenceCriteriaSatisfied`, `SolutionStatus.feasible`
+""",
+        version='6.7.4.dev0',
+    )
+    if results.termination_condition in [
+        TerminationCondition.maxTimeLimit,
+        TerminationCondition.iterationLimit,
+        TerminationCondition.objectiveLimit,
+        TerminationCondition.minStepLength,
+    ]:
```
michaelbynum, line 314: I'm not sure it really matters, but should these be sets instead of lists?

pyomo/contrib/solver/solvers/sol_reader.py
```diff
@@ -10,28 +10,153 @@
 #  ___________________________________________________________________________
```
michaelbynum, line 12: Should this file go in the common directory?

pyomo/contrib/solver/solvers/gurobi_direct.py
```diff
@@ -96,10 +117,7 @@ def __del__(self):
     def load_vars(self, vars_to_load=None, solution_number=0):
         assert solution_number == 0
         if self._grb_model.SolCount == 0:
-            raise RuntimeError(
-                'Solver does not currently have a valid solution. Please '
-                'check the termination condition.'
-            )
+            raise NoSolutionError()
```
michaelbynum, line 120: I love these changes.

pyomo/contrib/solver/solvers/gurobi_direct.py
```diff
@@ -229,9 +220,43 @@ def version(self):
         )
         return version
 
+
+class GurobiDirect(GurobiSolverMixin, SolverBase):
+    """
+    Interface to Gurobi direct (not persistent)
```
michaelbynum, line 226: I would change this to something like "Interface to Gurobi using gurobipy"

pyomo/contrib/solver/solvers/gurobi_persistent.py
```diff
@@ -306,24 +282,16 @@ def release_license(self):
 
     def __del__(self):
         if not python_is_shutting_down():
-            Gurobi._num_instances -= 1
-            if Gurobi._num_instances == 0:
+            GurobiPersistent._num_instances -= 1
+            if GurobiPersistent._num_instances == 0:
```
michaelbynum, line 286: should `_num_instances` live on the mixin class and be shared by direct and persistent???

pyomo/contrib/solver/solvers/gurobi_persistent.py
```diff
@@ -687,6 +627,8 @@ def _add_constraints(self, cons: List[ConstraintData]):
 
     def _add_sos_constraints(self, cons: List[SOSConstraintData]):
         for con in cons:
+            # NOTE: This seems like an unnecessary local variable because it's
+            # never used or called again. But when removed, it causes KeyErrors.
```
michaelbynum, line 631: Haha! The line `self._symbol_map.removeSymbol(con)` in `_remove_sos_constraints` has to be removed with it.

michaelbynum (commented):
I'm only about halfway done, but here are some minor comments.

pyomo/contrib/solver/common/persistent.py
```diff
@@ -10,21 +10,37 @@
...
+from pyomo.core.staleflag import StaleFlagManager
 from pyomo.common.collections import ComponentMap
 from pyomo.common.timing import HierarchicalTimer
-from pyomo.core.expr.numvalue import NumericConstant
-from pyomo.contrib.solver.util import collect_vars_and_named_exprs, get_objective
+from pyomo.contrib.solver.common.results import Results
+from pyomo.contrib.solver.common.util import collect_vars_and_named_exprs, get_objective
 
 
 class PersistentSolverUtils(abc.ABC):
```
mrmundt, line 28: NOT TODO for this PR:

- What is the intent of this class?
  - To manage automatic checking of state / checking differences in the model and updating them; implements the "Auto" part of APPSI
  - What is a better name? `PersistentSolverStateTracker`? This is the classic "is-a" vs. "has-a". Other ideas: `PersistentSolverChangeDetector`, `PersistentModelStateManager`, `PersistentSolverUpdater`

**Summary**: We don't want the persistent solver interfaces to inherit from this; we want the classes to be able to contain one of these classes and use it. `ModelChangeObserver`

mrmundt: Also probably won't need the `_func_name` and `func_name` standard here when re-worked.

pyomo/contrib/solver/common/base.py
```diff
@@ -182,12 +157,35 @@ class PersistentSolverBase(SolverBase):
     Example usage can be seen in the Gurobi interface.
     """
 
-    @document_kwargs_from_configdict(PersistentSolverConfig())
-    @abc.abstractmethod
+    CONFIG = PersistentSolverConfig()
+
+    def __init__(self, **kwds) -> None:
+        super().__init__(**kwds)
+        self._active_config = self.config
```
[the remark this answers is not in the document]
mrmundt: I am going to punt on this considering the conversation we had today about re-working `PersistentSolverUtils`. I think the answer will become more clear after that.

pyomo/contrib/solver/common/base.py
```diff
@@ -196,21 +194,20 @@ def is_persistent(self):
         """
         return True
 
-    def _load_vars(self, vars_to_load: Optional[Sequence[VarData]] = None) -> NoReturn:
+    def _load_vars(self, vars_to_load: Optional[Sequence[VarData]] = None) -> None:
```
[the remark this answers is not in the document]
mrmundt: In reading the documentation for `NoReturn`, it just didn't seem right: https://docs.python.org/3/library/typing.html#typing.NoReturn - but I am happy to be overturned on this one.

```
typing.Never
typing.NoReturn

    Never and NoReturn represent the [bottom type](https://en.wikipedia.org/wiki/Bottom_type), a type that has no members.

    They can be used to indicate that a function never returns, such as [sys.exit()](https://docs.python.org/3/library/sys.html#sys.exit):

    from typing import Never  # or NoReturn

    def stop() -> Never:
        raise RuntimeError('no way')

    Or to define a function that should never be called, as there are no valid arguments, such as [assert_never()](https://docs.python.org/3/library/typing.html#typing.assert_never):

    from typing import Never  # or NoReturn

    def never_call_me(arg: Never) -> None:
        pass

    def int_or_str(arg: int | str) -> None:
        never_call_me(arg)  # type checker error
        match arg:
            case int():
                print("It's an int")
            case str():
                print("It's a str")
            case _:
                never_call_me(arg)  # OK, arg is of type Never (or NoReturn)

    Never and NoReturn have the same meaning in the type system and static type checkers treat both equivalently.
```

pyomo/contrib/solver/solvers/highs.py
```diff
@@ -0,0 +1,705 @@
...
+        self._sol = highs.getSolution()
+        has_feasible_solution = self._sol.value_valid
+        if status == highspy.HighsModelStatus.kOptimal:
+            results.solution_status = SolutionStatus.optimal
+        elif has_feasible_solution:
+            results.solution_status = SolutionStatus.feasible
+        else:
+            results.solution_status = SolutionStatus.noSolution
+
+        if status == highspy.HighsModelStatus.kNotset:
```
mrmundt, line 578: Note for myself for the future - @michaelbynum made a really great `_tc_map` in `gurobi_direct` that I love. All of these redonkulous "if ... elif... elif..." statements should probably be replaced by maps.

pyomo/contrib/solver/solvers/sol_reader.py
mrmundt: As of right now, no, because it is solver-specific. It can be moved later.

pyomo/contrib/solver/common/base.py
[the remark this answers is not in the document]
michaelbynum: Oh, interesting. Thanks.

pyomo/contrib/solver/solvers/gurobi_direct.py
michaelbynum: Why is this one marked resolved?

pyomo/contrib/solver/solvers/highs.py
michaelbynum: Wasn't me, but I agree!

pyomo/contrib/solver/solvers/gurobi_direct.py
mrmundt: Because I updated the comment!

pyomo/contrib/solver/solvers/gurobi_persistent.py
mrmundt: I have been thinking about this question all weekend, and the answer I keep coming to is, "I'm not sure." Going back to the conversation with John last week, is that too much like passing around data (has-a vs. is-a)? I am fine with it either way (because functionally it works fine to put it in the mixin).

pyomo/contrib/solver/solvers/gurobi_direct.py
michaelbynum: Odd. I must not have refreshed my page or something. Apologies.

pyomo/contrib/solver/solvers/gurobi_persistent.py
michaelbynum: Maybe some of this will go away. I think we need to discuss `acquire_license` and `release_license` more. Maybe these should be independent of the number of instances of solver interfaces. Right now, if a someone has instances of both direct and persistent interfaces, the `__del__` method of one could affect the other.

pyomo/contrib/solver/tests/solvers/test_ipopt.py
```diff
@@ -28,6 +34,248 @@
...
+            loader.get_reduced_costs()
+
+        # Set _nl_info to something completely bogus but is not None
+        class NLInfo:
+            pass
+
+        loader._nl_info = NLInfo()
+        loader._nl_info.eliminated_vars = [1, 2, 3]
+        with self.assertRaises(NotImplementedError):
+            loader.get_reduced_costs()
```
michaelbynum, line 86: Can you leave a note that this test could go away some day if we enable returning duals when presolve is on?

pyomo/contrib/solver/tests/solvers/test_ipopt.py
```diff
@@ -28,6 +34,248 @@
...
+        # the command line
+        opt = ipopt.Ipopt(solver_options={'custom_option': 4})
+        with TempfileManager.new_context() as temp:
+            dname = temp.mkdtemp()
+            if not os.path.exists(dname):
+                os.mkdir(dname)
+            filename = os.path.join(dname, 'myfile')
+            result = opt._write_options_file(filename, opt.config.solver_options)
+            self.assertTrue(result)
+            self.assertTrue(os.path.isfile(filename + '.opt'))
```
michaelbynum, line 183: This test made me realize that we don't have a good way to make sure there are not typos in options except to look at the ipopt output. Anyone have good ideas on how to do that? The only thing I can think of is to parse the log file.

pyomo/contrib/solver/tests/unit/test_base.py
```diff
@@ -136,48 +87,43 @@ def test_class_method_list(self):
...
             self.assertEqual(self.instance.update_parameters(), None)
+        with self.assertRaises(NotImplementedError):
+            self.instance._get_primals()
+        with self.assertRaises(NotImplementedError):
+            self.instance._get_duals()
+        with self.assertRaises(NotImplementedError):
+            self.instance._get_reduced_costs()
+
+    def test_context_manager(self):
+        with base.PersistentSolverBase() as self.instance:
```
michaelbynum, line 125: why `self.instance` and not just `instance`?

pyomo/opt/results/solver.py
```diff
@@ -139,42 +139,64 @@ def to_solver_status(tc):
...
-    if results.solver.status == SolverStatus.ok and (
-        results.solver.termination_condition == TerminationCondition.optimal
-        or results.solver.termination_condition == TerminationCondition.locallyOptimal
-        or results.solver.termination_condition == TerminationCondition.globallyOptimal
-    ):
-        return True
+    from pyomo.contrib.solver.common.results import (
+        TerminationCondition as FutureTerminationCondition,
+        SolutionStatus as FutureSolutionStatus,
+    )
```
michaelbynum, line 155: This has to be here to avoid circular imports?

pyomo/opt/results/solver.py
```diff
@@ -139,42 +139,64 @@ def to_solver_status(tc):
...
+
+    if hasattr(results, 'solution_status'):
+        if results.solution_status == FutureSolutionStatus.optimal and (
+            results.termination_condition
+            == FutureTerminationCondition.convergenceCriteriaSatisfied
+        ):
```
michaelbynum, line 161: I don't think there is a need to check the termination condition anymore. Maybe worth a discussion.

michaelbynum (commented):
After this PR is merged, we need to port some of the recent changes in appsi to here.

michaelbynum:
I have now gone through all of the changed files.

pyomo/contrib/solver/solvers/highs.py
mrmundt: Agreed. I am not incorporating it into this one.

pyomo/contrib/solver/tests/unit/test_base.py
mrmundt: For not a single good reason. I'll fix it.

pyomo/opt/results/solver.py
mrmundt: That's a good point. Let's talk about it in the dev call.

pyomo/contrib/solver/tests/solvers/test_ipopt.py
michaelbynum: All good points.

michaelbynum (approved):

pyomo/contrib/solver/common/base.py
```diff
@@ -520,20 +556,20 @@ def _solution_handler(
         delete_legacy_soln = True
         if load_solutions:
             if hasattr(model, 'dual') and model.dual.import_enabled():
-                for c, val in results.solution_loader.get_duals().items():
-                    model.dual[c] = val
+                for cons, val in results.solution_loader.get_duals().items():
+                    model.dual[cons] = val
```
jsiirola, line 560: Style nit (for the future): I would prefer `con` and not `cons` (as I read `cons` as the plural of `con`)

pyomo/contrib/solver/common/base.py
```diff
@@ -638,11 +674,16 @@ def license_is_valid(self) -> bool:
...
+        Preserves config backwards compatibility; allows new solver interfaces
+        to be used in the pyomo solve call
+        """
         return default_config_block(self, init)[0]
 
     def set_options(self, options):
+        """
+        Method to manually set options; can be called outside of the solve method
+        """
         opts = {k: v for k, v in options.value().items() if v is not None}
```
jsiirola, line 687: This implementation implicitly assumes `options` is a `ConfigDict`.  We should update the documentation to include argument type specification, and then update the implementation to accept `dict` or `ConfigDict` (by dropping the call to `value()`:
```python
        opts = {k: v for k, v in options.items() if v is not None}
```

pyomo/contrib/solver/common/util.py
```diff
@@ -0,0 +1,128 @@
...
+    for obj in block.component_data_objects(
+        Objective, descend_into=True, active=True, sort=True
+    ):
+        if objective is not None:
+            raise ValueError('Multiple active objectives found.')
+        objective = obj
+    return objective
+
+
+class _VarAndNamedExprCollector(ExpressionValueVisitor):
```
jsiirola, line 83: TODO (later): this should be ported to the `StreamBasedExpressionVisitor`

michaelbynum: I have lots of these conversions to do...

PR #3482, opened 2025-02-19
BUGFIX: Validator for `tee` in `contrib.solver`

mrmundt:
## Summary/Motivation:
There is repeatable bug in the interaction between `capture_output` and any of the `*_v2` solvers in `contrib.solver` due to the way the `TextIO_or_Logger` domain validator was working. A minimal reproducible example of the bug:

```
import pyomo.environ as pyo
from pyomo.common.tee import capture_output

m = pyo.ConcreteModel()
m.x = pyo.Var([1,2],initialize=1, bounds=(0, None))
m.eq = pyo.Constraint(expr=m.x[1]*m.x[2]**1.5 == 3)
m.obj = pyo.Objective(expr=m.x[1]**2 + m.x[2]**2)

solver = pyo.SolverFactory("ipopt_v2")
with capture_output():
    solver.solve(m, tee=True)

```

- Slightly change the logic in `TextIO_or_Logger`
- Add tests to ensure correct behavior, both alone and under `capture_output`

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/solver/config.py
+++ b/pyomo/contrib/solver/config.py
@@ -31,21 +31,45 @@ from pyomo.common.timing import HierarchicalTimer
 
 
 def TextIO_or_Logger(val):
-    ans = []
-    if not isinstance(val, Sequence):
+    """
+    Validates and converts input into a list of valid output streams.
+
+    Accepts:
+      - sys.stdout
+      - Instances of io.TextIOBase
+      - logging.Logger (wrapped as LogStream)
+      - Boolean values (`True` → sys.stdout)
+
+    Returns:
+      - A list of validated output streams.
+
+    Raises:
+      - ValueError if an invalid type is provided.
+    """
+    if isinstance(val, (sys.stdout.__class__, io.TextIOBase, logging.Logger)):
         val = [val]
+
+    elif isinstance(val, Sequence) and not isinstance(val, (str, bytes)):
+        val = list(val)
+
+    else:
+        val = [val]
+
+    ans = []
+
     for v in val:
         if v.__class__ in native_logical_types:
             if v:
                 ans.append(sys.stdout)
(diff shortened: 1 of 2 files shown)


pyomo/contrib/solver/config.py
```diff
@@ -31,21 +31,45 @@
...
+      - ValueError if an invalid type is provided.
+    """
+    if isinstance(val, (sys.stdout.__class__, io.TextIOBase, logging.Logger)):
         val = [val]
+
+    elif isinstance(val, Sequence) and not isinstance(val, (str, bytes)):
+        val = list(val)
+
+    else:
+        val = [val]
```
jsiirola, line 56: This seems redundant and messy:
- we are checking types here, and again in the loop
- why are we forcing a copy of Sequences?

Why is this not just
```python
if not isinstance(val, Sequence) or isinstance(val, (str, bytes)):
    val = [val]
```

pyomo/contrib/solver/config.py
```diff
@@ -31,21 +31,45 @@
...
+
+    ans = []
+
     for v in val:
         if v.__class__ in native_logical_types:
             if v:
                 ans.append(sys.stdout)
-        elif isinstance(v, io.TextIOBase):
+        elif isinstance(v, (sys.stdout.__class__, io.TextIOBase)):
```
jsiirola, line 64: Please add a comment why we are testing against `stdout.__class__`.  It was not obvious (at least to me) that we are guarding against people (like us!) defining new file-like classes that do not derive from `TextIOBase` and then assigning them to stdout / stderr -- and we still want to accept those classes.

jsiirola (commented):
I have a couple nits that I would like addressed / documented.  I am not officially "requesting changes" so that the release will not be held up by my re-review.

blnicho (approved):

PR #3483, opened 2025-02-20
Multiple dispatch for relational expression generation

jsiirola:
## Summary/Motivation:

**NOTE:** This PR includes #3551; review / merge that PR first.

This is a continuation of the work moving Pyomo expression generation to use multiple dispatch for more efficient / standardized argument processing.  This migrates the relational expression generation.  Overall, it provides more standardized and efficient argument processing.  Overall performance on the performance test suite is improved (in particular, a 3.5% improvement in create_instance:

image

- Move relational expression generation to use multiple dispatch
- Move some definitions (enums and dispatcher utilities) into expr_common
- Improved solution to managing the circular dependency between relations and numeric expressions (with better code documentation)

### Legal Acknowledgement


jsiirola:
(Do not attempt to squeeze this into the release)

pyomo/core/expr/relational_expr.py
```diff
@@ -363,105 +397,319 @@ def _to_string(self, values, verbose, smap):
...
+        )
+    elif getattr(b, 'is_component_type', no)() and b.is_indexed():
+        msg = (
+            f"Argument for {op_type} is an indexed numeric "
+            "value specified without an index:\n\t{b.name}\nIs this "
+            "value defined over an index that you did not specify?"
+        )
     else:
-        return InequalityExpression((lhs, rhs), _relational_op[etype][2])
+        msg = "Attempting to use a non-numeric type in a " "numeric expression context:"
```
mrmundt, line 440: `black` did something weird here

pyomo/core/expr/relational_expr.py
```diff
@@ -363,105 +397,319 @@ def _to_string(self, values, verbose, smap):
...
+        (ARG_TYPE.INEQUALITY, ARG_TYPE.INEQUALITY): _lt_invalid,
+        (ARG_TYPE.INEQUALITY, ARG_TYPE.INVALID_RELATIONAL): _lt_invalid,
+        (ARG_TYPE.INVALID_RELATIONAL, ARG_TYPE.NATIVE): _lt_invalid,
+        (ARG_TYPE.INVALID_RELATIONAL, ARG_TYPE.PARAM): _lt_invalid,
+        (ARG_TYPE.INVALID_RELATIONAL, ARG_TYPE.OTHER): _lt_invalid,
+        (ARG_TYPE.INVALID_RELATIONAL, ARG_TYPE.INEQUALITY): _lt_invalid,
+        (ARG_TYPE.INVALID_RELATIONAL, ARG_TYPE.INVALID_RELATIONAL): _lt_invalid,
+    },
+    ARG_TYPE,
+)
```
mrmundt, line 715: I don't know if there is a "better" way to do this, but there are three versions of this map that are all "almost" the same, and it makes my little "minimize repeat code" heart sad.

pyomo/core/expr/expr_common.py
```diff
@@ -82,6 +78,112 @@ class ExpressionType(enums.Enum):
...
+    def _mutable_any(a, b):
+        a = _recast_mutable(a)
+        return dispatcher[a.__class__, b.__class__](a, b)
+
+    def _mutable_mutable(a, b):
+        if a is b:
+            a = b = _recast_mutable(a)
+        else:
+            a = _recast_mutable(a)
+            b = _recast_mutable(b)
```
emma58, line 168: Mostly just out of curiosity (because knowing you, you checked): is the `if` really so much faster than the `else` that it's worth checking and only recasting once?

emma58 (approved):
Sorry for the long-delayed review... This is really pretty! A few questions, but otherwise it looks good.

pyomo/core/expr/expr_common.py
jsiirola: The real issue is that if `a is b`, we would be calling `_recast_mutable` on the same object twice.  This is a partially inplace operation (the first step is to switch the object from a mutable sum-like expression into a "normal" sum-like expression).  If you call `_recast_mutable` on the same object twice you will get an exception (an `AttributeError`, because once the object is not longable mutable, it no longer has the `make_immutable()` method).

emma58: Oh, I see. Could you maybe add a comment to that effect, because I think only future-you will remember that!

PR #3488, opened 2025-02-24
mosek_direct.py : Fixed iparam parsing and added support for Task.putparam

Utkarsh-Detha:
## Fixes:
In `_apply_solver`
- Call to `putintparam` was missing in the `isinstance(option, str)` conditional when parsing `iparam`. Fixed.
- Was calling `option.pop('mosek')` instead of `option.pop(0)` within the same conditional. Fixed.

## Summary/Motivation:
A user wanted to be able to pass the value of ALL parameters as strings. MOSEK's optimizer API supports this through the [`Task.putparam`](https://docs.mosek.com/latest/pythonapi/optimizer-task.html#mosek.task.putparam) function, which accepts a parameter name (string) and a value (also string).

- Users may pass string values to any parameter (SPAR, DPAR, IPAR) as long as they use [generic parameter names](https://docs.mosek.com/latest/pythonapi/parameters.html)

### Legal Acknowledgement


mrmundt:
@Utkarsh-Detha - please run `black -S -C` on your branch (with the newest version of `black`)

Utkarsh-Detha:
Thank you for the review, @mrmundt . I ran the formatter as you recommended, with the newest version thereof.

blnicho:
@Utkarsh-Detha could you add a couple tests for these changes?

mrmundt:
@Utkarsh-Detha - Pinging again about this! Could you please add in some tests?

pyomo/solvers/plugins/solvers/mosek_direct.py
```diff
@@ -163,21 +163,25 @@ def _process_stream(msg):
...
+                    self._solver_model.putparam(key, option)
+                else:
+                    if param[0] == 'mosek':
+                        param.pop(0)
+                    param = getattr(mosek, param[0])(param[1])
+                    if 'sparam' in key.split('.'):
+                        self._solver_model.putstrparam(param, option)
+                    elif 'dparam' in key.split('.'):
+                        self._solver_model.putdouparam(param, option)
+                    elif 'iparam' in key.split('.'):
```
jsiirola, line 176: I worry that there is no `else:` clause: users who forget to explicitly include `.sparam`, `.dparam`, or `.iparam` will have those options silently ignored.

pyomo/solvers/plugins/solvers/mosek_direct.py
```diff
@@ -163,21 +163,25 @@ def _process_stream(msg):
...
+                        if isinstance(option, int):
+                            self._solver_model.putintparam(param, option)
+                        elif isinstance(option, str):
```
jsiirola, line 179: Same here - without an `else`, option values that aren't int / str will be silently skipped (this would include, e.g., numpy ints)

jsiirola (changes requested):
Overall this looks reasonable, but could be improved by some defensive checks.  In addition, it would be good to test this functionality thought the unit tests.

pyomo/solvers/plugins/solvers/mosek_direct.py
Utkarsh-Detha: I have included a warning under an else statement in this situation.

pyomo/solvers/plugins/solvers/mosek_direct.py
Utkarsh-Detha: Before the `if` clause, I call `param = getattr(mosek, param[0])(param[1])`. If an invalid key is in the options dict, it should result in an `AttributeError`. So I thought it should be ok even without an else statement.

Utkarsh-Detha:
@jsiirola I am starting to wonder if I should simply remove the option of allowing strings as iparam values. The idea initially was to allow users to pass MOSEK enums as strings, for example `mosek.onoffkey.off`. But I now think that it is more straightforward to just pass the option value directly to MOSEK, which would raise a TypeError if wrong input type is provided. As this would mean removing functionality, I wanted your opinion. (Lmk if I should rather open an Issue to discuss this)

mrmundt (changes requested):

pyomo/solvers/tests/checks/test_MOSEKDirect.py
```diff
@@ -313,6 +313,55 @@ def test_conic_duals(self):
                     results.Solution.constraint['x11']['Dual'][i], check[i], 5
                 )
 
+    def test_solver_parameters(self):
+        import mosek
+
+        with self.assertLogs('pyomo.solvers', level='WARNING') as warnLogs:
+            solver = pmo.SolverFactory('mosek_direct')
```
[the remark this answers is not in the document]
Utkarsh-Detha: Ah, right. I've made the change.

pyomo/solvers/tests/checks/test_MOSEKDirect.py
```diff
@@ -313,6 +313,55 @@ def test_conic_duals(self):
                     results.Solution.constraint['x11']['Dual'][i], check[i], 5
                 )
 
+    def test_solver_parameters(self):
```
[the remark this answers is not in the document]
Utkarsh-Detha: I added tests for both.

mrmundt (approved):

pyomo/solvers/plugins/solvers/mosek_direct.py
```diff
@@ -163,21 +163,27 @@ def _process_stream(msg):
...
-                            option.pop('mosek')
-                        option = getattr(mosek, option[0])(option[1])
-                    else:
-                        self._solver_model.putintparam(param, option)
+                if key == param[0]:
+                    self._solver_model.putparam(key, option)
+                else:
+                    if param[0] == 'mosek':
+                        param.pop(0)
+                    param = getattr(mosek, param[0])(param[1])
```
jsiirola, line 171: We should raise an exception if `len(param) != 2`.

pyomo/solvers/plugins/solvers/mosek_direct.py
```diff
@@ -163,21 +163,27 @@ def _process_stream(msg):
...
+                    if 'sparam' in key.split('.'):
```
jsiirola, line 172: I believe the following is both simpler and faster: 
```suggestion
                    if key.endswith('.sparam':
```
(and the same changes in the next 2 tests)

pyomo/solvers/plugins/solvers/mosek_direct.py
```diff
@@ -163,21 +163,27 @@ def _process_stream(msg):
...
+                        if isinstance(option, int):
+                            self._solver_model.putintparam(param, option)
+                        elif isinstance(option, str):
+                            option = option.split('.')
+                            if option[0] == 'mosek':
+                                option.pop(0)
+                            option = getattr(mosek, option[0])(option[1])
+                            self._solver_model.putintparam(param, option)
+                        else:
+                            logger.warning("Ignoring invalid value for iparam.\n")
```
jsiirola, line 186: Ideally, this message would include the name of the offending param (i.e., `key`), and the (invalid) type (e.g., `type(option).__name__`)

jsiirola (commented):
I think this is really close.  It would be good to add one extra check and clarify the error message.  If you have a chance to address this in the next couple hours, great - otherwise I may try and push to this branch and try to squeeze it into tomorrow's release.

jsiirola:
@Utkarsh-Detha Good question.  I don't have a strong opinion on supporting strings for `iparam`.  However, this  is a "core" solver: if you want to remove it, I think we we would need to put in a deprecation path for several releases before deleting the functionality completely.

pyomo/solvers/plugins/solvers/mosek_direct.py
Utkarsh-Detha: Ah, fair point. I have replaced it with `if 'sparam.' in key:` and removed the call to split (the parameter names are things like `mosek.iparam.log`, so can't use endswith)

pyomo/solvers/plugins/solvers/mosek_direct.py
Utkarsh-Detha: Sorry for flip-flopping like this, but your comment made me think that it is best if I simply raise TypeError in case a value with an incorrect type is passed. The Error message thus raised by MOSEK will contain the necessary information as you have suggested. Furthermore, this makes the way iparams are handled similar to the sparam/dpararm.

Utkarsh-Detha:
@jsiirola I also added/changed some tests to check if the correct exceptions are raised in case wrong types are passed for values of params. If a "str" type is passed to iparam but the name is recognized as a valid MOSEK enum, then an AttributeError is raised, which is also tested for in the tests. Let me know if you don't think this is a reasonable approach. Thanks for your help on this.

pyomo/solvers/plugins/solvers/mosek_direct.py
Utkarsh-Detha: Regarding this, I was still relying on the AttributeError in case a wrong parameter name is passed. Would you rather have it so that a specific assertion error is raised if the length is too long/short?

jsiirola: The AttributeError is probably fine if the parameter name is wrong.  I was worried about a user specifying something like `mosek.iparam.log.level`.  The current code would accept this, ignore `level`, and set `log`.  Maybe that is OK...?

pyomo/solvers/plugins/solvers/mosek_direct.py
jsiirola: Actually, I think the "else"  is still worthwhile.  As written, if a user provides a key `mosek.mark.up`, nothing will happen and there will be no warning/error.  Maybe not a big deal (it took me a bit of digging to actually come up with this...)?

What would you think of something like this:
```python
if key.startswith('mosek.'):
    key = key.split('.', 1)[1]
param = key.split('.', 1)
if len(param) == 1:
    self._solver_model.putparam(key, option)
else:
    param_obj = getattr(mosek, param[0])(param[1])
    if param[0] == 'sparam':
        self._solver_model.putstrparam(param_obj, option)
    elif param[0] == 'dparam':
        self._solver_model.putdouparam(param_obj, option)
    elif param[0] == 'iparam':
        if isinstance(option, str):
            if option.startswith('mosek'.):
                option = option.split('.', 1)[1]
            option = option.split('.', 1)
            option = getattr(mosek, option[0])(option[1])
        self._solver_model.putintparam(param_obj, option)
    else:
        raise ValueError(f"unrecognized option name '{key}'")
```

pyomo/solvers/plugins/solvers/mosek_direct.py
jsiirola: That seems quite reasonable.

pyomo/solvers/plugins/solvers/mosek_direct.py
Utkarsh-Detha: You make a good point, I will make the change. Thank you for the suggestion.

pyomo/solvers/plugins/solvers/mosek_direct.py
Utkarsh-Detha: Added an assertion to check this.

mrmundt (approved):

jsiirola (approved):

PR #3491, opened 2025-02-24
Updates to CodeCov configuration

jsiirola:
## Summary/Motivation:
This resolves some issues with our CodeCov configuration:
- Allow codecov to generates reports even when the CI fails.  This should help ensure that we get coverage information more often (even in the presence of intermittent infrastructure failures)
- prevent uploads with multiple flags (codecov is issuing warnings for reports with multiple flags)
- simplify the job name when reporting from Jenkins
- rename `codecov.yml` -> `.codecov.yml` to clean up the project directory

- (see above)

### Legal Acknowledgement


blnicho (approved):

blnicho:
@jsiirola I brought this up at the dev call last week but I noticed that codecov has some new test analytics features: https://docs.codecov.com/docs/test-analytics

Since this PR is already making changes to the codecov configuration, should we go ahead and enable these test analytics while we're at it?

jsiirola:
I did a quick look at the test analytics feature and it looks like it would require a somewhat significant restructuring of the GHA drivers to have each job always upload both coverage and the test result. Right now we bundle the codecov results into only 5 uploads because a while back CodeCov was having trouble when we uploaded all our job results individually. 

Given the effort, I think I would argue for doing it in a separate PR.

PR #3503, opened 2025-03-05
Add PyROS Uncertain Parameter Reduction

shermanjasonaf:
## Summary/Motivation:

This PR adds to PyROS the careful treatment of uncertain parameters that are constrained to a single value. Such uncertain parameters are now effectively treated as constants in all preprocessing steps and the separation problems. RO problems with uncertain parameters that are constrained to a single value can now be more expeditiously (and, in some cases, accurately) solved.

- Add uncertain parameter reduction
- Update testing

### Legal Acknowledgement


shermanjasonaf:
@jas-yao @natalieisenberg

natalieisenberg:
Out of curiosity, what is an example of a case where an uncertain parameter from a singleton set was not handled correctly prior to this commit?

shermanjasonaf:
> Out of curiosity, what is an example of a case where an uncertain parameter from a singleton set was not handled correctly prior to this commit?

For uncertainty set $Q \subset \mathbb{R}$, consider  the RO problem

$$\min_{(x_1, x_2) \in [1, 2] \times [1, 2]} \\{ x_1 + x_2 \mid q x_1 - x_2 = 0\\,\forall\\, q \in Q  \\}.$$

Prior to this PR, PyROS reports this problem as robust infeasible for any valid uncertainty set $Q$, as the preprocessor's coefficient matching step casts the equality constraint $qx_1 - x_2 = 0$ to the constraints $x_1 = 0$ and $-x_2 = 0$, which render the initial master problem infeasible. However, when $Q$ is the singleton set $\\{ 1 \\}$, for example, the problem is robust feasible, since in this case, the solution $(x_1, x_2) = (1, 1)$ is robust feasible.

natalieisenberg (approved):

jsiirola (approved):
This ended up being significantly more complex than I expected, but the implementation looks good.

blnicho (approved):

PR #3509, opened 2025-03-07
BUGFIX: Add logic for highspy versions

### LINKED ISSUE #3489: highspy >=1.8.0 required but this is not checked
Since keyboard interrupt support was added for HiGHS:
https://github.com/Pyomo/pyomo/blob/03dcc227b199008e445796dceed3dcadc07c49b0/pyomo/contrib/appsi/solvers/highs.py#L250

the minimum version of highspy required is 1.8.0.

This is not enforced or mentioned anywhere, so when this code is called using an earlier version, it fails with:
```
AttributeError: 'Highs' object has no attribute 'HandleKeyboardInterrupt'
```

### Expected behavior:
A check on the highspy version and a message or exception if insufficient, much like the check for the Gurobi solver:
https://github.com/Pyomo/pyomo/blob/03dcc227b199008e445796dceed3dcadc07c49b0/pyomo/contrib/appsi/solvers/gurobi.py#L54-L56

mrmundt:
## Summary/Motivation:
There is a new feature in `highspy` that allows keyboard interruption. It's not available in old versions. This PR checks the version to make sure it meets the minimum requirement.

- Ensure that version is at least 1.8 before setting keyboard interrupt

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/appsi/solvers/highs.py
+++ b/pyomo/contrib/appsi/solvers/highs.py
@@ -273,7 +273,8 @@ class Highs(PersistentBase, PersistentSolver):
                 self._warm_start()
             timer.start('optimize')
             ostreams[-1].write("RUN!\n")
-            self._solver_model.HandleKeyboardInterrupt = True
+            if self.version()[1] >= 8:
+                self._solver_model.HandleKeyboardInterrupt = True
             self._solver_model.run()
             timer.stop('optimize')
 
--- a/pyomo/contrib/appsi/solvers/highs.py
+++ b/pyomo/contrib/appsi/solvers/highs.py
@@ -273,7 +273,7 @@ class Highs(PersistentBase, PersistentSolver):
                 self._warm_start()
             timer.start('optimize')
             ostreams[-1].write("RUN!\n")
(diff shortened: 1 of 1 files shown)


michaelbynum (approved):

pyomo/contrib/appsi/solvers/highs.py
```diff
@@ -273,7 +273,8 @@ def _solve(self, timer: HierarchicalTimer):
                 self._warm_start()
             timer.start('optimize')
             ostreams[-1].write("RUN!\n")
-            self._solver_model.HandleKeyboardInterrupt = True
+            if self.version()[0] >= 1 and self.version()[1] >= 8:
+                self._solver_model.HandleKeyboardInterrupt = True
```
jsiirola, line 277: FWIW, there is a bug in this logic, as version 2.0 will not `HandleKeyboardInterrupt`.  An improved implementation would be:
```
if self.version()[:2] >= (1, 8):
```

michaelbynum: Good catch @jsiirola.

michaelbynum: creating a new PR now

PR #3521, opened 2025-03-12
APPSI: disable sending output to logger by default

michaelbynum:
- In the APPSI solver interfaces, by default, do not send output to the logger.

### Legal Acknowledgement


mrmundt:
We are already doing this correctly in `pyomo.contrib.solvers` (in which there is now a HiGHS interface), so we are not going to incorporate this. We'd prefer people use the new interface.

PR #3531, opened 2025-03-21
add qp support for highs

### LINKED ISSUE #3381: Cannot solve quadratic problem with Highs solver
I'm on Windows 11. I have `HiGHS version 1.7.2 Githash 5ce7a2753. Copyright (c) 2024 HiGHS under MIT licence terms` in a given directory (EDIT: same problem with `AMPL/HiGHS Optimizer [1.7.1] (Windows AMD64), driver(20240724), MP(20240724)`).

This program (standard solver factory doesn't accept `'highs'` argument which is another bug):

```
import os
from pyomo.environ import ConcreteModel, Var, Objective, NonNegativeReals, Constraint
from pyomo.contrib.appsi.solvers import Highs

# Set the environment variable PATH to include the directory of HiGHS executable
os.environ['PATH'] += os.pathsep + r"C:\bin\Highs\bin"

# Define the model
model = ConcreteModel("MyModel")

# Variables (Renamed to x1 and x2)
model.x1 = Var(name='x1', domain=NonNegativeReals)
model.x2 = Var(name='x2', domain=NonNegativeReals)

# Quadratic Objective function
model.obj = Objective(expr=model.x1 + model.x2 , sense=1)  # Minimize

# Constraints
model.con1 = Constraint(expr=model.x1 + 2 * model.x2 >= 1)
model.con2 = Constraint(expr=3 * model.x1 + 4 * model.x2 <= 2)

# Instantiate the HiGHS solver directly via APPSI
solver = Highs()

# Solve the model
results = solver.solve(model)

# Display the results
print("Termination Condition:", results.termination_condition)

# Print the values of the variables
print(f"x1 = {model.x1.value}")
print(f"x2 = {model.x2.value}")

# Display full model details
print()
model.display()

```

works well but changing objective to 

`model.obj = Objective(expr=model.x1*model.x1 + model.x2*model.x2, sense=1)`

results in

```
  File "C:\Micha\Programming\Python\Optimierung\highs_pyomo.py", line 26, in <module>
    results = solver.solve(model)
              ^^^^^^^^^^^^^^^^^^^
  File "C:\Python312\Lib\site-packages\pyomo\contrib\appsi\solvers\highs.py", line 263, in solve
    self.set_instance(model)
  File "C:\Python312\Lib\site-packages\pyomo\contrib\appsi\solvers\highs.py", line 383, in set_instance
    self.add_block(model)
  File "C:\Python312\Lib\site-packages\pyomo\contrib\appsi\base.py", line 1128, in add_block
    self.set_objective(obj)
  File "C:\Python312\Lib\site-packages\pyomo\contrib\appsi\base.py", line 1083, in set_objective
    self._set_objective(obj)
  File "C:\Python312\Lib\site-packages\pyomo\contrib\appsi\solvers\highs.py", line 595, in _set_objective
    raise DegreeError(
pyomo.contrib.appsi.solvers.highs.DegreeError: Highs interface does not support expressions of degree None

```

Solving the corresponding lp file `model.write('model_output.lp', format='lp')` on commandline using highs.exe  works (after some syntactic corrections).

EDIT: Changing the line 591ff in `highs.py`
```
            repn = generate_standard_repn(
                obj.expr, quadratic=False, compute_values=False
            )
```
from `quadratic=False` to `quadratic=True` removes the error and solution is produced.

quantresearch1:
## Summary/Motivation:
HiGHS can solve quadratic programming (QP) models, which contain an objective term x.T @ Q @ x where the Hessian matrix Q is positive semi-definite.

- Allow quadratic terms in the objective by changing quadratic to True in generate_standard_repn

### Legal Acknowledgement


emma58 (changes requested):
@quantresearch1, thank you for the contribution! As written, this does not yet add QP support for Highs, though: By changing the `quadratic` flag to `True` in `generate_standard_repn`, you move the nonlinear term in your objective from `repn.nonlinear_expr` to `repn.quadratic_vars` and `repn.quadratic_coefs` (and hence bypass the error in line 622 of `appsi/solvers/highs.py`) . To properly send quadratic expressions to Highs, you will need to parse this part of the `StandardRepn` object and send the result to Highs.

(Your test appears to get lucky--it is actually sending a constant objective to Highs, but the variables happen to land on the values you are expecting. You can see this by adding an assertion that the `results` object returned by `opt.solve(m)` agrees with your expected objective value: `self.assertEqual(results.best_feasible_objective, 2)`.)

quantresearch1:
@emma58 Thanks for the advice. Looking into the Highs C++ code, I did not find a quadratic equivalent to changeColCost so I think each coefficient update would have to be a call to passHessian which is not ideal in a persistent setting (assuming many updates). I will think about this a bit more and reraise if I can come up with a decent solution

quantresearch1:
I need to look into the highs unit tests failure, they look genuine

quantresearch1:
> I need to look into the highs unit tests failure, they look genuine

I have had a look at the remaining failures and I think it's unrelated to my PR:
-osx
INFO: The following extensions were downloaded:
    [FAIL]  gjh
    [ OK ]  mcpp
- linux
Message: 'Significant delay observed waiting to join reader threads, possible output stream deadlock'
Arguments: ()
Significant delay observed waiting to join reader threads, possible output stream deadlock
------- generated xml file: /home/runner/work/pyomo/pyomo/TEST-pyomo.xml -------
FAILED pyomo/contrib/solver/tests/solvers/test_solvers.py::TestSolvers::test_domain_0_gurobi - RuntimeError: TeeStream: deadlock observed joining reader threads

jsiirola:
> > I need to look into the highs unit tests failure, they look genuine
> 
> I have had a look at the remaining failures and I think it's unrelated to my PR:

I believe that those two test failures will be resolved by #3537.

quantresearch1:
@jsiirola Thanks, I have synced with main and that has helped with a few failures. I am just left with this one failure for python 3.13 on linux. Not clear to me if this branch is the culprit but will try to install linux over the weekend and debug. Let me know if you think it's unrelated to the branch
self = <pyomo.common.tee.capture_output object at 0x7f900cbbd6d0>, et = None
ev = None, tb = None

    def __exit__(self, et, ev, tb):
        # Check that we were nested correctly
        FAIL = []
        if self.tee.STDOUT is not sys.stdout:
            FAIL.append('Captured output does not match sys.stdout.')
        if self.tee.STDERR is not sys.stderr:
            FAIL.append('Captured output does not match sys.stderr.')
        # Exit all context managers.  This includes
        #  - Restore any file descriptors we commandeered
        #  - Close / join the TeeStream
        #  - Close any opened files
        FAIL.extend(self._exit_context_stack(et, ev, tb))
        sys.stdout, sys.stderr = self.old
        self.old = None
        self.tee = None
        self.output_stream = None
        if FAIL:
>           raise RuntimeError("\n".join(FAIL))
E           RuntimeError: TeeStream: deadlock observed joining reader threads

pyomo/common/tee.py:380: RuntimeError
----------------------------- Captured stderr call -----------------------------
Significant delay observed waiting to join reader threads, possible output stream deadlock
TeeStream: deadlock observed joining reader threads
------- generated xml file: /home/runner/work/pyomo/pyomo/TEST-pyomo.xml -------
FAILED pyomo/contrib/solver/tests/solvers/test_solvers.py::TestSolvers::test_equality_4_highs - RuntimeError: TeeStream: deadlock observed joining reader threads

pyomo/contrib/solver/solvers/highs.py
```diff
@@ -96,6 +96,121 @@ def update(self):
...
+
+            # Add the entry
+            q_index.append(row)
+            q_value.append(val)
+
+        while last_col < dim - 1:
+            last_col += 1
+            q_start[last_col] = len(q_value)
+
+        nnz = len(q_value)
```
mrmundt, line 198: I'm not sure what `nnz` means here - can you please name the variable more explicitly?

pyomo/contrib/solver/solvers/highs.py
```diff
@@ -96,6 +96,121 @@ def update(self):
...
+
+    def _build_and_pass_hessian(self):
+        """Build and pass the Hessian to HiGHS in CSC format"""
+        if not self.quad_coef_dict:
+            return
+
+        dim = self.highs.getNumCol()
+
+        # Build CSC format for the lower triangular part
+        q_value = []
```
mrmundt, line 175: Does `q_*` stand for `quadratic_*`?

pyomo/contrib/solver/tests/solvers/test_highs.py
```diff
@@ -0,0 +1,158 @@
...
+#  Copyright (c) 2008-2025
+#  National Technology and Engineering Solutions of Sandia, LLC
+#  Under the terms of Contract DE-NA0003525 with National Technology and
+#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain
+#  rights in this software.
+#  This software is distributed under the 3-clause BSD License.
+#  ___________________________________________________________________________
+
+import pyomo.common.unittest as unittest
+import pyomo.environ as pe
```
mrmundt, line 13: We have standardized to `import pyomo.environ as pyo` - can you please update?

mrmundt (changes requested):

pyomo/contrib/solver/solvers/highs.py
quantresearch1: The hessian is often referred to as a matrix Q (in many books I see a quadratic objective defined as 1/2 x.T @ Q @ x). I am happy to change to whatever you prefer

pyomo/contrib/solver/solvers/highs.py
quantresearch1: I have tried to name the variables in accordance with the highspy function definition:
    def passHessian(
        self,
        dim: int,
        nnz: int,
        format: int,
        q_start: numpy.ndarray[typing.Any, numpy.dtype[numpy.int32]],
        q_index: numpy.ndarray[typing.Any, numpy.dtype[numpy.int32]],
        q_value: numpy.ndarray[typing.Any, numpy.dtype[numpy.float64]],
    ) -> HighsStatus: ...

it's used a lot in scipy, their sparse matrices/arrays have a nnz attributes which returns the number of non-zero entries. Maybe num_nz is clearer (closer to the gurobi model attribute 'NumNZs') ? Happy to change it to whatever you think is clearer

mrmundt: Nah, as I said, jsiirola told me about that standard. I don't like it but it's not _our_ standard.

pyomo/contrib/solver/solvers/highs.py
mrmundt: I think I'd prefer `hessian_value` or `hess_value` then - we do use that pretty widely within Pyomo.

mrmundt (approved):

quantresearch1:
Hello,

I see this was scheduled to be released in April but then removed from the release. Were there any issues or concerns I should look into ?

mrmundt:
Hi @quantresearch1 - we are waiting for one of our other devs to review this. We've have a startlingly small amount of time lately. Thank you for your patience!

quantresearch1:
Understood, thanks for letting me know

pyomo/contrib/solver/solvers/highs.py
```diff
@@ -96,6 +96,121 @@ def update(self):
...
+        """
+        needs_quadratic_update = self._first_update
+
+        self.constant.update()
+        for coef in self.linear_coefs:
+            coef.update()
+
+        for ndx, coef in enumerate(self.quadratic_coefs):
+            current_val = value(coef.expr)
+            if current_val != self.last_quadratic_coef_values[ndx]:
```
michaelbynum: If the set of quadratic variables changes (e.g., the objective changes from `x**2 + y**2` to `x**2 + z**2`) will `ndx` be in `self.last_quadratic_coef_values`?

pyomo/contrib/solver/solvers/highs.py
```diff
@@ -96,6 +96,121 @@ def update(self):
...
+            dim,
+            nnz,
+            highspy.HessianFormat.kTriangular,
+            np.array(hessian_start, dtype=np.int32),
+            np.array(hessian_index, dtype=np.int32),
+            np.array(hessian_value, dtype=np.double),
+        )
+
+        if status != highspy.HighsStatus.kOk:
+            logger.warning(
```
michaelbynum, line 209: I think this should be an exception. Thoughts?

pyomo/contrib/solver/solvers/highs.py
```diff
@@ -96,6 +96,121 @@ def update(self):
...
 
 
+class _MutableQuadraticCoefficient:
+    def __init__(self, expr, row_idx, col_idx):
+        self.expr = expr
+        self.row_idx = row_idx
+        self.col_idx = col_idx
+
+
+class _MutableObjective:
```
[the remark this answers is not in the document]
michaelbynum: I can't think of a better way. I think this is okay.

pyomo/contrib/solver/solvers/highs.py
```diff
@@ -518,17 +637,36 @@ def _set_objective(self, obj):
...
+                    v1_id = id(v1)
+                    v2_id = id(v2)
+                    v1_ndx = self._pyomo_var_to_solver_var_map[v1_id]
+                    v2_ndx = self._pyomo_var_to_solver_var_map[v2_id]
+
+                    coef = repn.quadratic_coefs[ndx]
+
+                    mutable_quadratic_coefficients.append(
+                        _MutableQuadraticCoefficient(
+                            expr=coef, row_idx=v1_ndx, col_idx=v2_ndx
```
michaelbynum: I don't think you should store the variable indices here. If a variable gets removed from the model, these indices will become incorrect. Note how the `_MutableObjectiveCoefficient` stores the variable id and the `_pyomo_var_to_solver_var_map`. That way, it can get the correct index at the time `update` is called.

michaelbynum: I realize this is very convoluted, but I'm not sure how to avoid it.

michaelbynum: It would be great to add a test for this scenario.

pyomo/contrib/solver/solvers/highs.py
michaelbynum: Oh, this should be fine because `_set_objective` should have been called when the objective changed. Can you just make sure there is a test that validates this? There might already be one. I don't remember.

pyomo/contrib/solver/tests/solvers/test_highs.py
```diff
@@ -0,0 +1,158 @@
...
+        self.assertAlmostEqual(m.x1.value, 1, places=5)
+        self.assertAlmostEqual(m.x2.value, 1, places=5)
+        self.assertEqual(results.objective_bound, 1)
+
+        m.p.value = 2.0
+        opt.update_parameters()
+        results = opt.solve(m)
+        self.assertAlmostEqual(m.x1.value, 1, places=5)
+        self.assertAlmostEqual(m.x2.value, 1, places=5)
+        self.assertEqual(results.objective_bound, 2)
```
michaelbynum: I think these tests should be moved to `test_solvers.py`. Let's add a `qp_solvers` group [here](https://github.com/Pyomo/pyomo/blob/56521be56dadcbcd2d45ab95b95d9a0e1953e2ec/pyomo/contrib/solver/tests/solvers/test_solvers.py#L54) and apply these tests to all of the solvers in that group. For now, the group should include `gurobi_persistent`, `ipopt`, and `highs`.

pyomo/contrib/solver/tests/solvers/test_highs.py
```diff
@@ -0,0 +1,158 @@
...
+        # Quadratic Objective function
+        m.obj = pyo.Objective(expr=m.x1 * m.x1 + m.x2 * m.x2, sense=pyo.minimize)
+
+        m.con1 = pyo.Constraint(expr=m.x1 >= 1)
+        m.con2 = pyo.Constraint(expr=m.x2 >= 1)
+
+        results = opt.solve(m)
+        self.assertAlmostEqual(m.x1.value, 1, places=5)
+        self.assertAlmostEqual(m.x2.value, 1, places=5)
+        self.assertEqual(results.objective_bound, 2)
```
michaelbynum: Let's check the incumbent objective instead of the objective bound.

michaelbynum (changes requested):
Overall, this looks great. I have one major concern and a couple of minor concerns.

pyomo/contrib/solver/solvers/highs.py
quantresearch1: Thinking a bit more about it, pretty much every single function in highspy returns a HighsStatus (https://github.com/ERGO-Code/HiGHS/blob/364c83a51e44ba6c27def9c8fc1a49b1daf5ad5c/highs/Highs.h#L962), 
However, if I look at the current usage in this file, we don't capture the status when calling changeRowBounds, changeCoeff, changeColCost, etc... Just for the sake of consistency, there's an argument to remove this check

pyomo/contrib/solver/solvers/highs.py
quantresearch1: Ah yes you're right, let me add one.

pyomo/contrib/solver/tests/solvers/test_solvers.py
```diff
@@ -1128,6 +1129,49 @@ def test_mutable_quadratic_objective(
...
+        m.p.value = 2.0
+        results = opt.solve(m)
+        self.assertAlmostEqual(m.x1.value, 1, places=4)
+        self.assertAlmostEqual(m.x2.value, 1, places=4)
+        self.assertAlmostEqual(results.incumbent_objective, 2, 4)
+
+        del m.x1
+        results = opt.solve(m)
+        self.assertAlmostEqual(m.x2.value, 1, places=4)
+        self.assertAlmostEqual(results.incumbent_objective, 1, 4)
```
quantresearch1: @michaelbynum following your suggestion, I have added a test that deletes one of the variables. GurobiPersistent is failing this test as it finds the incumbent objective to be 2. I have not really played around with deleting variables before, is it the same as fixing that variable to 0 ? Because if I remove x1 from 2* x1^2 + x2^2 -x1*x2 I would expect the objective to be 1 since we have x2 >= 1 as a constraint. Not sure what I am missing

pyomo/contrib/solver/solvers/highs.py
quantresearch1: I have changed the code such that the dictionary index is the id of variable 1 and 2, I think that should address your concern. These indices only get turned to row/col indices when we build the hessian. I am not sold on the Coefficient class having the map itself, it's definitely a model property in my mind.

pyomo/contrib/solver/solvers/highs.py
quantresearch1: I have changed that logic. The test that deletes x1 should cover this now.

pyomo/contrib/solver/tests/solvers/test_highs.py
quantresearch1: Done

pyomo/contrib/appsi/solvers/tests/test_highs_persistent.py
```diff
@@ -18,7 +18,6 @@
 from pyomo.common.log import LoggingIntercept
 from pyomo.common.tee import capture_output
 from pyomo.contrib.appsi.solvers.highs import Highs
-from pyomo.contrib.appsi.base import TerminationCondition
```
mrmundt: This import was necessary - the lack of it is what's causing your test failures currently.

quantresearch1: Thank you, not sure what happened there, let me fix

quantresearch1: I am at work right now and unfortunately they have locked down pushes from both personal AND enterprise accounts... I will fix it once home

quantresearch1: Done

pyomo/contrib/solver/solvers/highs.py
michaelbynum: Good point. I guess I don't feel strongly one way or the other.

pyomo/contrib/solver/tests/solvers/test_solvers.py
```diff
@@ -1129,6 +1130,49 @@ def test_mutable_quadratic_objective(
...
+        m.p.value = 2.0
+        results = opt.solve(m)
+        self.assertAlmostEqual(m.x1.value, 1, places=4)
+        self.assertAlmostEqual(m.x2.value, 1, places=4)
+        self.assertAlmostEqual(results.incumbent_objective, 2, 4)
+
+        del m.x1
+        results = opt.solve(m)
+        self.assertAlmostEqual(m.x2.value, 1, places=4)
+        self.assertAlmostEqual(results.incumbent_objective, 2, 4)
```
michaelbynum: This does not do what you think. Deleting a variable from a model does is not like removing a column from an LP. The convention in Pyomo is that the model is defined by the set of active components (normally constraints and objectives, but also things like disjunctions). If the objective still uses `x1`, then deleting `x1` from the model actually has no effect (except that you cannot easily access `x1` anymore). I'll make a minor modification to this test real quick, and then we can get it merged.

michaelbynum (changes requested):

pyomo/contrib/solver/tests/solvers/test_solvers.py
michaelbynum: Oops. I think I essentially answered this in a new thread. Sorry.

michaelbynum:
@emma58, your review is blocking, but I think it is outdated.

pyomo/contrib/solver/tests/solvers/test_solvers.py
quantresearch1: Thanks for explaining how deleting works and updating the tests for me !

PR #3538, opened 2025-03-25
Type check all `exception` arguments in component `__call__` implementations

### LINKED ISSUE #2922: Not throwing error when missing the * to multiply Param and Var
## Summary

When defining a constraint like the following Pyomo will throw an "Invalid constraint expression" Value Error:
```
def const_1(m, t):
    return 0 == m.x(m.y)
m.const_1 = pyo.Constraint(m.t, rule=const_1)
```
In this case I deliberately forgot to place a `*` in order to multiply the values. The function should be:
```
    return 0 == m.x*(m.y)
```

The expected Value Error is:
```console
$ ValueError: Invalid constraint expression. The constraint expression resolved to a trivial Boolean (False) instead of a Pyomo object. Please modify your rule to return Constraint.Infeasible instead of False.

Error thrown for Constraint 'const_1[0]'
```
**However we found that for some specific cases the error won't throw.**

## Error

The following code is ready to run and replicate:
```
import pyomo.environ as pyo
from pyomo.network import *

m = pyo.ConcreteModel()

l_t = [0,1,2,3,4]
m.t = pyo.Set(initialize=l_t)

m.x = pyo.Param(initialize = 0.5, within=pyo.NonNegativeReals)
m.y = pyo.Var(m.t, initialize = [0,0,0,0,0], within=pyo.NonNegativeReals)

def const_1(m, t):
    return m.y[t] == m.x(m.y[t])
m.const_1 = pyo.Constraint(m.t, rule=const_1)

def obj_fun(m):
    return 0
m.goal = pyo.Objective(rule=obj_fun, sense=pyo.minimize)

instance = m.create_instance()
solver = pyo.SolverFactory('ipopt')
solver.solve(instance, tee=True)
```
In this simplified example we define a Param `m.x` and a Var `m.y` which is time dependant (i.e: `m.y[t]`). It is set as a feasibility problem, therefore the objective function returns the value `0`.
**Running the program throws neither warnings nor errors and the solver is able to find an (wrong) optimal solution**

Then we obtained the values for `instance.y` by running:
```
instance.y.get_values()
```

- Missing the `*` :  `{0: 0.5, 1: 0.5, 2: 0.5, 3: 0.5, 4: 0.5}`
- Placing the `*` :  `{0: 0.0, 1: 0.0, 2: 0.0, 3: 0.0, 4: 0.0}`

### System information

Pyomo version: 6.6.1
Python version: 3.9.13
Operating system: Windows 10
Solver (if applicable): ipopt (Ipopt 3.10.1)

emma58:
## Summary/Motivation:

Params and most other numeric- and Boolean-valued components are callable (to get the value), but they did not previously check that the `exception` argument was in fact a `bool`. This meant it was easy to make a typo missing a `*` character and silently rendering part of the expression constant (with the value of a Param, for example). This adds checks in the implementations of `__call__` that `exception` is a `bool`. I still think we should discuss deprecating call to evaluate entirely, but this is a much less invasive change that will raise an error for many typos that would otherwise create silently wrong models.

- Checks type of `exception` argument in implementations of `__call__`
- Adds a test for the common Param-related typo

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/base/boolean_var.py
+++ b/pyomo/core/base/boolean_var.py
@@ -19,7 +19,8 @@ from pyomo.common.timing import ConstructionTimer
 from pyomo.core.staleflag import StaleFlagManager
 from pyomo.core.expr.boolean_value import BooleanValue
 from pyomo.core.expr import GetItemExpression
-from pyomo.core.expr.numvalue import value, _type_check_exception_arg
+from pyomo.core.expr.expr_common import _type_check_exception_arg
+from pyomo.core.expr.numvalue import value
 from pyomo.core.base.component import ComponentData, ModelComponentFactory
 from pyomo.core.base.global_set import UnindexedComponent_index
 from pyomo.core.base.indexed_component import IndexedComponent, UnindexedComponent_set
--- a/pyomo/core/base/boolean_var.py
+++ b/pyomo/core/base/boolean_var.py
@@ -19,7 +19,7 @@ from pyomo.common.timing import ConstructionTimer
 from pyomo.core.staleflag import StaleFlagManager
 from pyomo.core.expr.boolean_value import BooleanValue
 from pyomo.core.expr import GetItemExpression
-from pyomo.core.expr.numvalue import value
+from pyomo.core.expr.numvalue import value, _type_check_exception_arg
 from pyomo.core.base.component import ComponentData, ModelComponentFactory
 from pyomo.core.base.global_set import UnindexedComponent_index
 from pyomo.core.base.indexed_component import IndexedComponent, UnindexedComponent_set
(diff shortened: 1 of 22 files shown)


blnicho (approved):

jsiirola:
This is great, but should be promulgated to other expression objects:
```
./core/base/boolean_var.py
./core/base/constraint.py
./core/base/expression.py
./core/base/logical_constraint.py
./core/base/matrix_constraint.py
./core/base/objective.py
./core/base/param.py
./core/base/units_container.py
./core/base/var.py
./core/expr/base.py
./core/expr/boolean_value.py
./core/expr/numvalue.py
./core/expr/template_expr.py
./core/kernel/conic.py
./core/kernel/constraint.py
./core/kernel/expression.py
./core/kernel/matrix_constraint.py
./core/kernel/parameter.py
./core/kernel/variable.py
./repn/beta/matrix.py
```

emma58:
@jsiirola this is ready for another review.

Issue #3508, opened 2025-03-07
Invalid doctest in PyROS - unclear what behavior should be


mrmundt:
## Summary

During tracking down a totally different problem, I noticed that there is a doctest from `uncertainty_sets` that is invalid code and doesn't run. We don't fail on warnings, so we hadn't noticed before.

### Steps to reproduce the issue

I noticed this in the doctests when I was looking at something else entirely

```console
D:\a\pyomo\pyomo\doc\OnlineDocs\..\..\pyomo\contrib\pyros\uncertainty_sets.py.rst: 
WARNING: ignoring invalid doctest code: 
'>>> conf_ellipsoid = EllipsoidalSet(\n...     center=np.zeros(4),\n...     shape_matrix=np.diag(range(1, 5)),\n...     scale=None,\n...     gaussian_conf_lvl=0.95,\n... )\n>>> conf_ellipsoid.center\narray([0, 0, 0, 0])\n>>> conf_ellipsoid.shape_matrix\narray([[1, 0, 0, 0]],\n       [0, 2, 0, 0]],\n       [0, 0, 3, 0]],\n       [0, 0, 0. 4]])\n>>> conf_ellipsoid.scale\n...9.4877...\n>>> conf_ellipsoid.gaussian_conf_lvl\n0.95'
...
```

In looking at that code, there is, in fact, a problem. [Here is the code now](https://github.com/Pyomo/pyomo/blob/main/pyomo/contrib/pyros/uncertainty_sets.py#L2424C1-L2443C9):
```

    A 4D 95% confidence ellipsoid:

    >>> conf_ellipsoid = EllipsoidalSet(
    ...     center=np.zeros(4),
    ...     shape_matrix=np.diag(range(1, 5)),
    ...     scale=None,
    ...     gaussian_conf_lvl=0.95,
    ... )
    >>> conf_ellipsoid.center
    array([0, 0, 0, 0])
    >>> conf_ellipsoid.shape_matrix
    array([[1, 0, 0, 0]],
           [0, 2, 0, 0]],
           [0, 0, 3, 0]],
           [0, 0, 0. 4]])
    >>> conf_ellipsoid.scale
    ...9.4877...
    >>> conf_ellipsoid.gaussian_conf_lvl
    0.95
```

I think it should actually be:
```
    A 4D 95% confidence ellipsoid:

    >>> conf_ellipsoid = EllipsoidalSet(
    ...     center=np.zeros(4),
[...]

PR #3558, opened 2025-04-11
Update PyROS Uncertainty Set Validation Methods

jas-yao:
## Summary/Motivation:
This PR provides updates to PyROS uncertainty set validation methods and related tests.
Here, a `validate` method replaces the `is_valid` method (which solves 2N bounding problems to check for set boundedness) in all uncertainty sets, with each set having its own custom `validate` method that efficiently checks set-specific attributes and raises informative exceptions if any issues are found.

- Update `is_bounded` and `is_nonempty` methods in base `UncertaintySet` class
- Provide a `_solve_feasibility` method in base `UncertaintySet` class
- Replace `is_valid` with `validate` method that runs `is_bounded` and `is_nonempty` in the base `UncertaintySet` class
- Override `validate` in subclass uncertainty sets to check set-specific attributes
- Remove attribute setter checks in uncertainty sets that have been moved to the `validate` method
- Update unit tests for `is_bounded`, `is_nonempty`, `_solve_feasibility`, and `validate` methods

## TODO
- [x] Add FBBT for `_compute_parameter_bounds`
- [x] Add `validate_array` to all `validate` methods
- [x] Split `test_validate` tests into separate tests for each validation check
- [x] Delete `_validate` method for `PolyhedralSet`
- [x] Update error messages to clearly indicate issues specific to uncertainty set setup
- [x] Check uncertainty set documentation looks good and test fix for #3508 
- [ ] Update version number and changelog (after #3581 merged)

### Legal Acknowledgement


jas-yao:
@shermanjasonaf

jas-yao:
Thanks for all the feedback and suggestions @jsiirola !
Regarding FBBT, I was able to reproduce the issue with the following short script.

```python
"""Reproducing FBBT error."""
from pyomo.contrib.pyros.uncertainty_sets import PolyhedralSet
from pyomo.contrib.fbbt.fbbt import fbbt

# create a polyhedral uncertainty set that is not bounded
lhs_coefficients_mat = [[1, 2, 3], [4, 5, 6]]
rhs_vec = [1, 3]
pset = PolyhedralSet(lhs_coefficients_mat, rhs_vec)

# create pyomo model and pprint
bounding_model = pset._create_bounding_model()
bounding_model.pprint()

# try to use fbbt and pprint results
fbbt(bounding_model)
bounding_model.pprint()
```

with traceback

```python
  File "<string>", line 17, in __PYTHON_EL_eval
  File "/Users/yletian/Documents/projects/pyomo/pyomo/contrib/pyros/temp_fbbt_error.py", line 15, in <module>
    fbbt(bounding_model)
    ~~~~^^^^^^^^^^^^^^^^
  File "/Users/yletian/Documents/projects/pyomo/pyomo/contrib/fbbt/fbbt.py", line 1543, in fbbt
    _new_var_bounds = _fbbt_block(comp, config)
  File "/Users/yletian/Documents/projects/pyomo/pyomo/contrib/fbbt/fbbt.py", line 1429, in _fbbt_block
    _new_var_bounds = _fbbt_con(c, config)
  File "/Users/yletian/Documents/projects/pyomo/pyomo/contrib/fbbt/fbbt.py", line 1334, in _fbbt_con
    visitorA.walk_expression(con.expr)
    ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/Users/yletian/Documents/projects/pyomo/pyomo/core/expr/visitor.py", line 276, in walk_expression
    result = self._process_node(root, RECURSION_LIMIT)
  File "/Users/yletian/Documents/projects/pyomo/pyomo/core/expr/visitor.py", line 472, in _process_node_bx
    tmp = self.beforeChild(node, child, child_idx)
  File "/Users/yletian/Documents/projects/pyomo/pyomo/contrib/fbbt/fbbt.py", line 1160, in beforeChild
    return _before_child_handlers[child.__class__](self, child)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/Users/yletian/Documents/projects/pyomo/pyomo/contrib/fbbt/fbbt.py", line 1108, in _register_new_before_child_handler
    if child.is_variable_type():
       ^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'numpy.int64' object has no attribute 'is_variable_type'
```

This was fixed with your suggested changes:
```diff
@@ -19,7 +19,7 @@ from pyomo.core.expr.visitor import (
     identify_variables,
     StreamBasedExpressionVisitor,
 )
-from pyomo.core.expr.numvalue import nonpyomo_leaf_types, value
+from pyomo.common.numeric_types import nonpyomo_leaf_types, native_numeric_types, value
 from pyomo.core.expr.numvalue import is_fixed
 import pyomo.contrib.fbbt.interval as interval
 import math
@@ -1105,7 +1105,9 @@ def _before_external_function(visitor, child):
 def _register_new_before_child_handler(visitor, child):
     handlers = _before_child_handlers
     child_type = child.__class__
-    if child.is_variable_type():
+    if child_type in native_numeric_types:
+        handlers[child_type] = _before_constant
+    elif child.is_variable_type():
         handlers[child_type] = _before_var
     elif not child.is_potentially_variable():
         handlers[child_type] = _before_NPV
```

With this fix, I'll take a deeper look into possibly replacing the 2N bounding problems of `is_bounded` with FBBT.

blnicho:
@jas-yao is this ready for review or should we wait for the TODO's in the description to be completed?

jas-yao:
Hi @blnicho, I will need to finish some of the TODOs before this will be ready for review.

shermanjasonaf:
@jas-yao Please delay the version number/changelog update until after #3581 has been merged.

jas-yao:
@shermanjasonaf, @blnicho, @jsiirola I have completed the TODOs and think this PR should be ready for review now.

pyomo/contrib/pyros/tests/test_uncertainty_sets.py
```diff
@@ -2623,6 +3050,88 @@ def test_compute_parameter_bounds(self):
...
+        # when parameter_bounds is not available
+        custom_set.parameter_bounds = None
+        start = time.time()
+        self.assertTrue(custom_set.is_bounded(config=CONFIG), "Set is not bounded")
+        end = time.time()
+        time_without_bounds_provided = end - start
+
+        # check with parameter_bounds should always take less time than solving 2N
+        # optimization problems
+        self.assertLess(
```
jas-yao: I noticed that this test (check that the new `is_bounded` method is faster than the old 2N optimization problems) is failing in some cases. Specifically:

```
AssertionError: 0.0 not less than 0.0 : Boundedness check with provided parameter_bounds took longer than expected.
```

I realize this timing test is not the best due to different runtimes on different systems and was wondering if I should just remove this? Using `assertLessEqual` may also resolve the issue.

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -47,9 +48,11 @@
     POINT_IN_UNCERTAINTY_SET_TOL,
     standardize_component_data,
 )
+from pyomo.contrib.fbbt.fbbt import fbbt
+from pyomo.common.errors import InfeasibleConstraintException
 
 
-valid_num_types = tuple(native_numeric_types)
+valid_num_types = native_numeric_types
```
jsiirola: Instead if "renaming" `native_numeric_types`, it would be better to just reference that set directly.  In particular, it will help developers who are familiar with its use elsewhere to recognize what you are doing here.

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -558,23 +567,38 @@ def is_bounded(self, config):
...
+        param_bounds_arr = self.parameter_bounds
+        if param_bounds_arr:
+            all_bounds_finite = np.all(np.isfinite(param_bounds_arr))
+        else:
+            # initialize uncertain parameter variables
+            param_bounds_arr = np.array(self._fbbt_parameter_bounds(config))
+            if not all(map(lambda x: all(x), param_bounds_arr)):
+                # solve bounding problems if FBBT cannot find bounds
+                param_bounds_arr = np.array(
+                    self._compute_parameter_bounds(solver=config.global_solver)
```
jsiirola: We should probably update `_compute_parameter_bounds` so that any (partially known) bounds (from FBBT) can be skipped (and not re-solved for)?

Also, when looking at `_compute_parameter_bounds`:
```python
        if index is None:
            index = list(range(self.dim))
        bounding_model = self._create_bounding_model()
        objs_to_optimize = (
            (idx, obj)
            for idx, obj in bounding_model.param_var_objectives.items()
            if idx in index
        )
```
 This i *very* inefficient (it is a quadratic search).  I would recommend rewriting it as
```python
        bounding_model = self._create_bounding_model()
        objs_to_optimize = bounding_model.param_var_objectives.items()
        if index is not None:
            set_of_target_indices = set(index)
            objs_to_optimize = filter(lambda idx, obj: idx in set_of_target_indices, objs_to_optimize)
```

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -583,16 +607,65 @@ def is_bounded(self, config):
...
         """
-        return self.is_nonempty(config=config) and self.is_bounded(config=config)
+        check_nonempty = self.is_nonempty(config=config)
+        check_bounded = self.is_bounded(config=config)
+
+        if not check_nonempty:
+            raise ValueError(
+                "Failed nonemptiness check. Nominal point is not in the set. "
+                f"Nominal point:\n {config.nominal_uncertain_param_vals}."
+            )
```
jsiirola: Why do you assign these to new variables?  I think the code would be easier to read if you just:
```suggestion
        if not self.is_nonempty(config=config):
            raise ValueError(
                "Failed nonemptiness check. Nominal point is not in the set. "
                f"Nominal point:\n {config.nominal_uncertain_param_vals}."
            )
```

jsiirola (commented):
Over all, this looks pretty good.  A couple small questions / edits and I think we can be good to go.

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -218,7 +221,7 @@ def validate_arg_type(
         Name of argument to be displayed in exception message.
     arg_val : object
         Value of argument to be checked.
-    valid_types : type or tuple of types
+    valid_types : type, tuple of types, or iterable of types
```
shermanjasonaf: Since a tuple is an iterable and `validate_arg_type` now performs no treatment specific to tuples:

```suggestion
    valid_types : type or iterable of types
```

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -508,6 +514,9 @@ def parameter_bounds(self):
         """
         Bounds for the value of each uncertain parameter constrained
         by the set (i.e. bounds for each set dimension).
+
+        This method should return an empty list if it can't be calculated
+        or a list of length = self.dim if it can.
```
shermanjasonaf: I would replace this sentence with a full description of the (structure of the) returned value. Something like:

```
        Returns
        -------
        : list of tuple
            If the bounds can be calculated, then the list is of
            length `N`, and each entry is a pair of numeric
            (lower, upper) bounds for the corresponding
            (Cartesian) coordinate. Otherwise, the list is empty.
```

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -558,23 +567,38 @@ def is_bounded(self, config):
...
+        bounds, and if not all bound are found,
+        (ii) solving a sequence of maximization and minimization problems
+        (in which the objective for each problem is the value of a
+        single uncertain parameter).
+        If any of the optimization models cannot be solved successfully to
         optimality, then False is returned.
 
-        This method is invoked during the validation step of a PyROS
-        solver call.
+        This method is invoked by validate.
```
shermanjasonaf:
```suggestion
        This method is invoked by ``self.validate()``.
```

pyomo/contrib/pyros/uncertainty_sets.py
shermanjasonaf: @jsiirola I agree that we should probably modify `_compute_parameter_bounds()` such that we can skip the global solution of bounding problems corresponding to finite bounds that were already obtained via FBBT. In so doing, we should take into consideration other methods that invoke `_compute_parameter_bounds()` and rely on the exactness of the bounds returned. One such method is `_is_coordinate_fixed()`, added in #3503.

A possible resolution:

1. Allow the optional `index` argument to `_compute_parameter_bounds()` to be None or either of the following:
    -  A list (with length ``self.dim``) of 2-tuple of bool, such that for each entry of each tuple, a value of True [False] indicates that the corresponding bounding problem should be solved [skipped]. In this case, make `_compute_parameter_bounds()` return a list of 2-tuples such that the entries corresponding to skipped bounds are of value `None`.
    - A list of 2-tuples corresponding to the bounding problems that should be solved. The first entry of each tuple indicates the ordinal position of the uncertain parameter and the second entry is a 0-1 value indicating the bound (lower or upper). In this case, make `_compute_parameter_bounds()` return a dict mapping the tuples to the calculated bounds.
3. Make use of the modified `index` argument to `_compute_parameter_bounds()` in `is_bounded()`, such that only bounds for which non-finite values were reported by the FBBT method are (re-)calculated.
4. Account for changes to the logic of the `index` argument to`_compute_parameter_bounds()` throughout the rest of the PyROS codebase. In particular,  the method `_is_coordinate_fixed()` may need to be modified to account for changes to the structure of the returned bounds.

For clarity, we should probably also rename `_compute_parameter_bounds()` to `_compute_exact_parameter_bounds()` and/or ensure that the introductory one-sentence summary in the docstring explicitly conveys that the bounds returned are exact (that is, modify the sentence to "Compute _exact_ lower and upper bounds...").

I would probably also ensure that the docstring of `_fbbt_parameter_bounds()` explicitly conveys that the bounds returned by `_fbbt_parameter_bounds()` are, or may be, inexact.

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -583,16 +607,65 @@ def is_bounded(self, config):
...
+        Parameters
+        ----------
+        config : ConfigDict
+            PyROS solver configuration.
+
+        Returns
+        -------
+        : bool
+            True if the nominal point is within the set,
+            and False otherwise.
```
shermanjasonaf: Since this method does not necessarily check that the nominal point is located in the set:

```suggestion
            True if the uncertainty set is nonempty,
            and False otherwise.
```

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -583,16 +607,65 @@ def is_bounded(self, config):
...
         """
-        return self.is_bounded(config)
+        # check if nominal point is in set for quick test
+        set_nonempty = False
+        if config.nominal_uncertain_param_vals:
+            if self.point_in_set(config.nominal_uncertain_param_vals):
+                set_nonempty = True
```
shermanjasonaf: Consider this simplification:

```suggestion
        if config.nominal_uncertain_param_vals:
            set_nonempty = self.point_in_set(config.nominal_uncertain_param_vals)
```

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -583,16 +607,65 @@ def is_bounded(self, config):
...
+            PyROS solver configuration.
+
+        Raises
+        ------
+        ValueError
+            If nonemptiness check or boundedness check fail.
         """
-        return self.is_nonempty(config=config) and self.is_bounded(config=config)
+        check_nonempty = self.is_nonempty(config=config)
+        check_bounded = self.is_bounded(config=config)
```
shermanjasonaf: For efficiency, I would invoke `self.is_bounded()` only after the exception for nonemptiness is (not) raised. See also [this comment](https://github.com/Pyomo/pyomo/pull/3558/files#r2159658894).

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -735,6 +808,82 @@ def _compute_parameter_bounds(self, solver, index=None):
...
+        bounding_model = self._create_bounding_model()
+
+        # calculate bounds with FBBT
+        fbbt_exception_str = f"Error computing parameter bounds with FBBT for {self}"
+        try:
+            fbbt(bounding_model)
+        except InfeasibleConstraintException as fbbt_infeasible_con_exception:
+            config.progress_logger.error(
+                f"{fbbt_exception_str}\n" f"{fbbt_infeasible_con_exception}"
+            )
```
shermanjasonaf: 1. Consider adding a single test for this exception (e.g., on `BoxSet(bounds=[[2, 1]])`) in the uncertainty sets testing module.
2. Consider also modifying the error message as follows (to more clearly delineate the exception message):

```suggestion
            config.progress_logger.error(
                "Encountered the following exception "
                f"while computing parameter bounds with FBBT "
                f"for uncertainty set {self}:\n  "
                f"{fbbt_infeasible_con_exception!r}"
            )
```

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -735,6 +808,82 @@ def _compute_parameter_bounds(self, solver, index=None):
...
 
+    def _fbbt_parameter_bounds(self, config):
+        """
+        Obtain parameter bounds of the uncertainty set using FBBT.
+
+        Parameters
+        ----------
+        config : ConfigDict
+            PyROS solver configuration.
+        """
```
shermanjasonaf: For completeness, consider adding a description of the return value to this docstring, to the effect of:

```
        Returns
        -------
        param_bounds : list of tuple
            List, of length `N`, containing
            (lower bound, upper bound) pairs
            for the uncertain parameters.
```

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -735,6 +808,82 @@ def _compute_parameter_bounds(self, solver, index=None):
...
+
+        # solve feasibility problem
+        res = solver.solve(model, load_solutions=False)
+        if not check_optimal_termination(res):
+            raise ValueError(
+                "Could not successfully solve feasibility problem. "
+                f"Solver status summary:\n {res.solver}."
+            )
+
+        return True
```
shermanjasonaf: Since `True` is returned if and only if the feasibility problem is solved successfully, and an exception is raised otherwise, I would consider dropping this return statement, modifying the docstring accordingly, and modifying `UncertaintySet.is_nonempty()` accordingly. (Add line `set_nonempty = True` after the invocation of `_solve_feasibility()` in the else block of `is_nonempty()`.)

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -583,16 +607,65 @@ def is_bounded(self, config):
...
+
+        Parameters
+        ----------
+        config : ConfigDict
+            PyROS solver configuration.
+
+        Raises
+        ------
+        ValueError
+            If nonemptiness check or boundedness check fail.
```
shermanjasonaf:
```suggestion
            If nonemptiness check or boundedness check fails.
```

pyomo/contrib/pyros/uncertainty_sets.py
shermanjasonaf: This message should be modified for general accuracy. There are possible cases in which the nonemptiness check may fail for reasons not related to the nominal point, e.g., the nominal point is not provided, so the point is not checked; or the `is_nonempty()` method has been overriden by the user (in, say, a custom `UncertaintySet` subclass) such that the nominal point is never checked.

I would leave details related to the check on the nominal point to the logging messages of the ``is_nonempty()`` method, and modify the error message here to:

```suggestion
            raise ValueError(f"Nonemptiness check failed for uncertainty set {self}.")
```

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -583,16 +607,65 @@ def is_bounded(self, config):
...
+            raise ValueError(
+                "Failed nonemptiness check. Nominal point is not in the set. "
+                f"Nominal point:\n {config.nominal_uncertain_param_vals}."
+            )
+
+        if not check_bounded:
+            raise ValueError(
+                "Failed boundedness check. Parameter bounds are not finite. "
+                f"Parameter bounds:\n {self.parameter_bounds}."
+            )
```
shermanjasonaf: This message should be modified for general accuracy. There are possible cases in which `is_bounded()` returns False for reasons not related to direct calculations of the parameter bounds, as `is_bounded()` may be overriden by the user (in, say, a custom subclass) to perform an equivalent check that does not involve those calculations. Further, the parameter bounds ultimately checked by `is_bounded()` need not be those contained in `self.parameter_bounds` (e.g., if FBBT or `_compute_parameter_bounds()` was used instead).

I would leave details on the boundedness check to the logging messages of `is_bounded()`, and modify the error message here follows:

```suggestion
            raise ValueError(f"Boundedness check failed for uncertainty set {self}.")
```

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -2519,12 +2837,12 @@ class EllipsoidalSet(UncertaintySet):
...
-    array([[1, 0, 0, 0]],
-           [0, 2, 0, 0]],
-           [0, 0, 3, 0]],
-           [0, 0, 0. 4]])
+    array([[1, 0, 0, 0],
+           [0, 2, 0, 0],
+           [0, 0, 3, 0],
+           [0, 0, 0, 4]])
     >>> conf_ellipsoid.scale
     ...9.4877...
```
shermanjasonaf: The doctest here is failing locally on my Linux system. Modifying the docstring as follows resolves the error and should also fix #3508:

```suggestion
    >>> conf_ellipsoid.scale  # doctest: +ELLIPSIS +NORMALIZE_WHITESPACE
    <BLANKLINE>
    ...9.4877...
```

Similarly, the `IntersectionSet` class doctest fails, and the failure can be resolved by adding just the [`doctest: +ELLIPSIS` directive](https://docs.python.org/3/library/doctest.html#doctest.ELLIPSIS) at the end of the line `>>> intersection.all_sets`.

pyomo/contrib/pyros/tests/test_uncertainty_sets.py
```diff
@@ -359,6 +361,58 @@ def test_add_bounds_on_uncertain_parameters(self):
...
+        with self.assertRaisesRegex(ValueError, exc_str):
+            box_set.validate(config=CONFIG)
+
+    @unittest.skipUnless(baron_available, "BARON is not available")
+    def test_bounded_and_nonempty(self):
+        """
+        Test `is_bounded` and `is_nonempty` for a valid box set.
+        """
+        box_set = BoxSet(bounds=[[1.0, 2.0], [3.0, 4.0]])
+        bounded_and_nonempty_check(self, box_set),
```
shermanjasonaf: In testing methods named `test_bounded_and_nonempty()` throughout this module, lines like this have a redundant trailing comma:

```suggestion
        bounded_and_nonempty_check(self, box_set)
```

pyomo/contrib/pyros/tests/test_uncertainty_sets.py
```diff
@@ -2532,6 +2876,84 @@ def test_add_bounds_on_uncertain_parameters(self):
...
+        """
+        Test `is_bounded` and `is_nonempty` for a valid polyhedral set.
+        """
+        polyhedral_set = PolyhedralSet(
+            lhs_coefficients_mat=[[1.0, 0.0], [-1.0, 1.0], [-1.0, -1.0]],
+            rhs_vec=[2.0, -1.0, -1.0],
+        )
+        bounded_and_nonempty_check(self, polyhedral_set),
+
     def test_is_coordinate_fixed(self):
```
shermanjasonaf, line 2969: The `@unittest.skipUnless(baron_available, "BARON is not available")` decoration of `test_is_coordinate_fixed()` was implicitly removed here. The decoration should be restored, or else `test_is_coordinate_fixed()` will fail, rather than be skipped, when BARON is not available.

pyomo/contrib/pyros/tests/test_uncertainty_sets.py
```diff
@@ -2623,6 +3050,88 @@ def test_compute_parameter_bounds(self):
...
+        )
+
+        # check when feasibility problem fails
+        CONFIG.nominal_uncertain_param_vals = None
+        custom_set.parameter_bounds = [[1, 2], [3, 4]]
+        exc_str = r"Could not successfully solve feasibility problem. .*"
+        with self.assertRaisesRegex(ValueError, exc_str):
+            custom_set.is_nonempty(config=CONFIG)
+
     def test_is_coordinate_fixed(self):
```
shermanjasonaf, line 3136: The `@unittest.skipUnless(baron_available, "BARON is not available")` decoration of `test_is_coordinate_fixed()` was implicitly removed here. The decoration should be restored, or else `test_is_coordinate_fixed()` will fail, rather than be skipped, when BARON is not available.

pyomo/contrib/pyros/tests/test_uncertainty_sets.py
```diff
@@ -1420,6 +1565,48 @@ def test_add_bounds_on_uncertain_parameters(self):
...
+    def test_bounded_and_nonempty(self):
+        """
+        Test `is_bounded` and `is_nonempty` for a valid intersection set.
+        """
+        bset = BoxSet(bounds=[[-1, 1], [-1, 1], [-1, 1]])
+        aset = AxisAlignedEllipsoidalSet([0, 0, 0], [1, 1, 1])
+        intersection_set = IntersectionSet(box_set=bset, axis_aligned_set=aset)
+        bounded_and_nonempty_check(self, intersection_set),
+
     def test_is_coordinate_fixed(self):
```
shermanjasonaf, line 1627: The `@unittest.skipUnless(baron_available, "BARON is not available")` decoration of `test_is_coordinate_fixed()` was implicitly removed here. The decoration should be restored, or else `test_is_coordinate_fixed()` will fail, rather than be skipped, when BARON is not available.

pyomo/contrib/pyros/tests/test_uncertainty_sets.py
```diff
@@ -2623,6 +3050,88 @@ def test_compute_parameter_bounds(self):
...
+        end = time.time()
+        time_without_bounds_provided = end - start
+
+        # check with parameter_bounds should always take less time than solving 2N
+        # optimization problems
+        self.assertLess(
+            time_with_bounds_provided,
+            time_without_bounds_provided,
+            "Boundedness check with provided parameter_bounds took longer than expected.",
+        )
```
shermanjasonaf: In agreement with [this comment](https://github.com/Pyomo/pyomo/pull/3558/files#r2159330412), the success of this time-sensitive check appears to be system-dependent. Is this portion of the test maintainable in the long term?

pyomo/contrib/pyros/tests/test_uncertainty_sets.py
```diff
@@ -1924,6 +2199,76 @@ def test_add_bounds_on_uncertain_parameters(self):
...
+        # check when half lengths are negative
+        a_ellipsoid_set.half_lengths = [1, -1]
+        exc_str = r"Entry -1 of.*'half_lengths' is negative.*"
+        with self.assertRaisesRegex(ValueError, exc_str):
+            a_ellipsoid_set.validate(config=CONFIG)
+
+    @unittest.skipUnless(baron_available, "BARON is not available")
+    def test_bounded_and_nonempty(self):
+        """
+        Test `is_bounded` and `is_nonempty` for a valid axis aligned ellipsoidal set.
```
shermanjasonaf: Wrap this sentence:

```suggestion
        Test `is_bounded` and `is_nonempty` for a valid
        axis-aligned ellipsoidal set.
```

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -583,16 +607,65 @@ def is_bounded(self, config):
...
+            )
+
+        return set_nonempty
 
-    def is_valid(self, config):
+    def validate(self, config):
         """
-        Return True if the uncertainty set is bounded and non-empty,
-        else False.
+        Validate the uncertainty set with a nonemptiness and boundedness check.
```
shermanjasonaf: Consider wrapping this sentence:

```suggestion
        Validate the uncertainty set with a nonemptiness
        and boundedness check.
```

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -558,23 +567,38 @@ def is_bounded(self, config):
...
+        are finite.
+
+        If no parameter bounds are available, the following processes are run
+        to perform the check:
+        (i) feasibility-based bounds tightening is used to obtain parameter
+        bounds, and if not all bound are found,
+        (ii) solving a sequence of maximization and minimization problems
+        (in which the objective for each problem is the value of a
+        single uncertain parameter).
+        If any of the optimization models cannot be solved successfully to
```
shermanjasonaf: Consider (re-)wrapping these sentences (and all other added/modified docstrings in this PR) to a line length of ~72 (including the indentation spaces). I generally wrap docstrings to that length to avoid long lines and since `black` does not wrap docstrings.

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -1640,6 +1800,54 @@ def set_as_constraint(self, uncertain_params=None, block=None):
...
+            valid_type_desc="a valid numeric type",
+            required_shape=None,
+        )
+
+        # check no column is all zeros. otherwise, set is unbounded
+        cols_with_all_zeros = np.nonzero(
+            [np.all(col == 0) for col in lhs_coeffs_arr.T]
+        )[0]
+        if cols_with_all_zeros.size > 0:
+            col_str = ", ".join(str(val) for val in cols_with_all_zeros)
```
shermanjasonaf: This check is not (necessarily) consistent with the method docstring ("...full column rank of the LHS matrix..."). Was your final decision to check that the LHS matrix is full column rank, or that there is no column of zeros? Either the docstring or this check should be modified accordingly. If this check is to be kept as-is, then the first assignment can be simplified to:

```python
        cols_with_all_zeros = np.nonzero(np.all(lhs_coeffs_arr == 0, axis=0))[0]
```

If you intend to check that the matrix is full column rank, then you can adopt the check used in `FactorModelSet.validate()`.

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -1161,6 +1306,34 @@ def set_as_constraint(self, uncertain_params=None, block=None):
...
 
+    def validate(self, config):
+        """
+        Check BoxSet validity.
+
+        Raises
+        ------
+        ValueError
+            If any uncertainty set attributes are not valid.
+            If finiteness or bounds checks fail.
```
shermanjasonaf: For this and other `UncertaintySet` subclass `validate()` method docstrings, the phrase "finiteness ... checks" should be modified for clarity, particularly to new readers viewing this docstring through the Online Docs. (E.g., of what is the "finiteness" being checked?) I would probably change this particular sentence to "If ``self.bounds`` contains invalid (e.g., infinite) values."

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -1419,6 +1578,57 @@ def point_in_set(self, point):
...
 
+    def validate(self, config):
+        """
+        Check CardinalitySet validity.
+
+        Raises
+        ------
+        ValueError
+            If any uncertainty set attributes are not valid.
+            If finiteness, positive deviation, or gamma checks fail.
```
shermanjasonaf: Echoing the comments on the docstring of `BoxSet.validate()`:

```suggestion
            If any uncertainty set attributes are not valid,
            (e.g., numeric values are infinite,
            ``self.positive_deviation`` has negative values,
            or ``self.gamma`` is out of range).
```

I would modify the other subclass-specific `validate()` docstrings similarly.

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -558,23 +567,38 @@ def is_bounded(self, config):
...
+            all_bounds_finite = np.all(np.isfinite(param_bounds_arr))
+        else:
+            # initialize uncertain parameter variables
+            param_bounds_arr = np.array(self._fbbt_parameter_bounds(config))
+            if not all(map(lambda x: all(x), param_bounds_arr)):
+                # solve bounding problems if FBBT cannot find bounds
+                param_bounds_arr = np.array(
+                    self._compute_parameter_bounds(solver=config.global_solver)
+                )
+            all_bounds_finite = np.all(np.isfinite(param_bounds_arr))
```
shermanjasonaf: I propose that this be modified to something like:

```python
        # use parameter bounds if they are available
        param_bounds_arr = np.array(self.parameter_bounds)
        all_bounds_finite = (
            param_bounds_arr.size > 0
            and np.isfinite(param_bounds_arr).all()
        )

        if not all_bounds_finite:
            # use FBBT
            param_bounds_arr = np.array(
                self._fbbt_parameter_bounds(config),
                dtype="float",
            )
            all_bounds_finite = np.isfinite(param_bounds_arr).all()

        if not all_bounds_finite:
            # solve bounding problems
            param_bounds_arr = np.array(
                self._compute_parameter_bounds(solver=config.global_solver)
            )
            all_bounds_finite = np.isfinite(param_bounds_arr).all()

```

In particular, this robustifies the check on the FBBT bounds. The current FBBT bounds finiteness check `all(map(lambda x: all(x), param_bounds_arr))` may return an inaccurate result if `param_bounds_arr` contains an entry of value 0.

pyomo/contrib/pyros/tests/test_uncertainty_sets.py
```diff
@@ -2623,6 +3050,88 @@ def test_compute_parameter_bounds(self):
...
+        CONFIG.global_solver = global_solver
+
+        # using provided parameter_bounds
+        start = time.time()
+        self.assertTrue(custom_set.is_bounded(config=CONFIG), "Set is not bounded")
+        end = time.time()
+        time_with_bounds_provided = end - start
+
+        # when parameter_bounds is not available
+        custom_set.parameter_bounds = None
```
shermanjasonaf: The `parameter_bounds` attribute should be an empty list if no bounds are to be provided:

```suggestion
        custom_set.parameter_bounds = []
```

pyomo/contrib/pyros/tests/test_uncertainty_sets.py
```diff
@@ -2623,6 +3050,88 @@ def test_compute_parameter_bounds(self):
...
+        self.assertTrue(custom_set.is_bounded(config=CONFIG), "Set is not bounded")
+        end = time.time()
+        time_without_bounds_provided = end - start
+
+        # check with parameter_bounds should always take less time than solving 2N
+        # optimization problems
+        self.assertLess(
+            time_with_bounds_provided,
+            time_without_bounds_provided,
+            "Boundedness check with provided parameter_bounds took longer than expected.",
```
shermanjasonaf: Wrap this string to reduce the line length

```suggestion
            "Boundedness check with provided parameter_bounds took longer than "
            "expected.",
```

pyomo/contrib/pyros/uncertainty_sets.py
jas-yao: I ran my doctest with 
```bash
python -m doctest -o=ELLIPSIS
```
which is why they passed. I'm not sure how doctests will normally be run, so it will be good to include these to ensure they pass.

pyomo/contrib/pyros/uncertainty_sets.py
jas-yao: The docstring was modified to clarify the check is for no column of zeros

pyomo/contrib/pyros/tests/test_uncertainty_sets.py
jas-yao: This time-based test has been removed

pyomo/contrib/pyros/tests/test_uncertainty_sets.py
jas-yao: This test is not required for verifying that the `is_bounded` method is working. I have removed it to avoid dealing with the hassle of system-dependent results.

pyomo/contrib/pyros/tests/test_uncertainty_sets.py
jas-yao: This test is not required for verifying that the `is_bounded` method is working. I have removed it to avoid dealing with the hassle of system-dependent results.

jas-yao:
@jsiirola @shermanjasonaf Thank you for all of your feedback! I have addressed most comments, with the remaining being mostly related to rewriting the `_compute_parameter_bounds` method. I will aim to finish updates for these by the end of tomorrow.

pyomo/contrib/pyros/uncertainty_sets.py
jas-yao: This may have an undesirable behavior of returning True, even if `self.parameter_bounds` contains bad bounds (e.g inf, nan) because it will continue to try to find better bounds with `_fbbt_parameter_bounds` and `_compute_parameter_bounds`. I think it might be better to keep the cases of `self.parameter_bounds` exists and doesn't exist separate.

How about this?
```python
        # use parameter bounds if they are available
        param_bounds_arr = self.parameter_bounds
        if param_bounds_arr:
            all_bounds_finite = (
                np.all(np.isfinite(param_bounds_arr))
                )
        else:
            # use FBBT
            param_bounds_arr = np.array(
                self._fbbt_parameter_bounds(config),
                dtype="float",
            )
            all_bounds_finite = np.isfinite(param_bounds_arr).all()

            if not all_bounds_finite:
                # solve bounding problems
                param_bounds_arr = np.array(
                    self._compute_parameter_bounds(solver=config.global_solver)
                )
                all_bounds_finite = np.isfinite(param_bounds_arr).all()
```

pyomo/contrib/pyros/uncertainty_sets.py
jas-yao: I have chosen to go with option 1 in using an index list of 2-tuple of bool for this. Here, I have the check to run/skip optimization problems directly in the for loop using this index list. I have completely removed the inefficiency mentioned by @jsiirola.

I think that doing it this way has not caused any issues with other parts of the PyROS codebase outside of `is_bounded`, since the `_compute_parameter_bounds` return object is the same as before.

jas-yao:
@shermanjasonaf @jsiirola I have addressed all comments. 

I would appreciate further feedback on my implementations for `is_bounded` and `_compute_exact_parameter_bounds`. 
I will also need to complete the final TODO to update the version number and changelog.

pyomo/contrib/pyros/tests/test_uncertainty_sets.py
```diff
@@ -1632,6 +1784,103 @@ def test_add_bounds_on_uncertain_parameters(self):
...
+
+        # construct a valid cardinality set
+        cardinality_set = CardinalitySet(
+            origin=[0.0, 0.0], positive_deviation=[1.0, 1.0], gamma=2
+        )
+
+        # check when values are not finite
+        cardinality_set.origin[0] = np.nan
+        exc_str = (
+            r"Entry 'nan' of the argument `origin` " r"is not a finite numeric value"
```
blnicho:
```suggestion
            r"Entry 'nan' of the argument `origin` is not a finite numeric value"
```

pyomo/contrib/pyros/tests/test_uncertainty_sets.py
```diff
@@ -1924,6 +2217,77 @@ def test_add_bounds_on_uncertain_parameters(self):
...
+
+        # construct a valid axis aligned ellipsoidal set
+        center = [0.0, 0.0]
+        half_lengths = [1.0, 3.0]
+        a_ellipsoid_set = AxisAlignedEllipsoidalSet(center, half_lengths)
+
+        # check when values are not finite
+        a_ellipsoid_set.center[0] = np.nan
+        exc_str = (
+            r"Entry 'nan' of the argument `center` " r"is not a finite numeric value"
```
blnicho:
```suggestion
            r"Entry 'nan' of the argument `center` is not a finite numeric value"
```

pyomo/contrib/pyros/tests/test_uncertainty_sets.py
```diff
@@ -2306,6 +2601,106 @@ def test_add_bounds_on_uncertain_parameters(self):
...
+        # construct a valid ellipsoidal set
+        center = [0.0, 0.0]
+        shape_matrix = [[1.0, 0.0], [0.0, 2.0]]
+        scale = 1
+        ellipsoid_set = EllipsoidalSet(center, shape_matrix, scale)
+
+        # check when values are not finite
+        ellipsoid_set.center[0] = np.nan
+        exc_str = (
+            r"Entry 'nan' of the argument `center` " r"is not a finite numeric value"
```
blnicho:
```suggestion
            r"Entry 'nan' of the argument `center` is not a finite numeric value"
```

pyomo/contrib/pyros/tests/test_uncertainty_sets.py
```diff
@@ -2531,6 +2894,85 @@ def test_add_bounds_on_uncertain_parameters(self):
...
+        # construct a valid polyhedral set
+        polyhedral_set = PolyhedralSet(
+            lhs_coefficients_mat=[[1.0, 0.0], [-1.0, 1.0], [-1.0, -1.0]],
+            rhs_vec=[2.0, -1.0, -1.0],
+        )
+
+        # check when values are not finite
+        polyhedral_set.rhs_vec[0] = np.nan
+        exc_str = (
+            r"Entry 'nan' of the argument `rhs_vec` " r"is not a finite numeric value"
```
blnicho:
```suggestion
            r"Entry 'nan' of the argument `rhs_vec` is not a finite numeric value"
```

jsiirola (approved):
This looks good.  @jas-yao, @shermanjasonaf: do you want to update the version & changelog as part of this PR before we merge it, or just add it to the next PR?

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -583,16 +620,56 @@ def is_bounded(self, config):
...
+        Determine whether the uncertainty set is nonempty.
+
+        Parameters
+        ----------
+        config : ConfigDict
+            PyROS solver configuration.
+
+        Returns
+        -------
+        : bool
```
mrmundt, line 632: The actual varname is missing here.

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -508,6 +511,14 @@ def parameter_bounds(self):
         """
         Bounds for the value of each uncertain parameter constrained
         by the set (i.e. bounds for each set dimension).
+
+        Returns
+        -------
+        : list of tuple
```
mrmundt, line 517: The actual varname is missing here.

mrmundt (approved):
There are two doc issues (missing variables names), but this is otherwise lovely.

pyomo/contrib/pyros/uncertainty_sets.py
jsiirola: These are return values, so the actual var name is not necessary (or really all that relevant).  I believe that this is syntactically correct numpydoc (the `:` just makes it explicit that we are providing the return type and are not specifying a (suggested) name.

shermanjasonaf:
@jas-yao This looks good. You may want to remove the custom implementation of `DiscreteScenarioSet.is_bounded()`, which, though never used in the code (apart from in the discrete uncertainty set tests), unconditionally returns `True`.

shermanjasonaf:
> This looks good. @jas-yao, @shermanjasonaf: do you want to update the version & changelog as part of this PR before we merge it, or just add it to the next PR?

@jsiirola Let's leave the version number and changelog updates for #3649.

PR #3559, opened 2025-04-11
Add time interpolation to mpc data

eslickj:
## Fixes None

Adds option to linearly interpolate from MPC TimeSeriesData.

## Summary/Motivation:

Sometimes you have time series data that doesn't line up perfectly with model time indexes. @blnicho and @jsiirola, this is part of the recent MHE work.

- Add some linear interpolation utility functions and use them to add interpolation capbility to get data and time points not in MPC TimeSeriesData 

### Legal Acknowledgement


pyomo/contrib/mpc/data/tests/test_series_data.py
```diff
@@ -126,6 +126,26 @@ def test_get_data_at_time_with_tolerance(self):
         with self.assertRaisesRegex(RuntimeError, msg):
             new_data = data.get_data_at_time(-0.01, tolerance=1e-3)
```
mrmundt, line 128: You'll need a "if not numpy_available" skip condition here.

Robbybp (commented):
Thanks for the addition. This is definitely functionality we were missing. I think this would be better as it's own `interpolate`, or `get_interpolated_data`, method. Would that work for your application?

There is also the question of whether to support some form of interpolation in `DynamicModelInterface`. (E.g., in `get_data_at_time`, `load_data`, or dedicated methods for interpolation.) I'm fine with (and think I prefer) punting for now  if you don't need this for your application.

pyomo/contrib/mpc/data/interpolation.py
```diff
@@ -0,0 +1,68 @@
...
+#  Under the terms of Contract DE-NA0003525 with National Technology and
+#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain
+#  rights in this software.
+#  This software is distributed under the 3-clause BSD License.
+#  ___________________________________________________________________________
+
+from pyomo.common.dependencies import numpy as np
+
+
+def get_time_index_vec(time_set, time_data):
```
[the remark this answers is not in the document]
Robbybp: Can this be done with `bisect.bisect_right`? E.g.:
```python
import bisect
pos = [bisect.bisect_right(time_set, t) for t in time_data]
```
This would avoid adding a numpy dependency.

pyomo/contrib/mpc/data/tests/test_series_data.py
```diff
@@ -126,6 +126,26 @@ def test_get_data_at_time_with_tolerance(self):
...
+        self.assertEqual(
+            TimeSeriesData({m.var[:, "A"]: [1.5], m.var[:, "B"]: [3]}, [t1]), new_data
+        )
+
+        new_t = [0.05, 0.15]
+        new_data = data.get_data_at_time(new_t, interpolate=True)
+        self.assertEqual(
+            TimeSeriesData({m.var[:, "A"]: [1.5, 2.5], m.var[:, "B"]: [3, 5]}, new_t),
+            new_data,
+        )
```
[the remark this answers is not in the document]
Robbybp: This comment is no longer relevant if we go with a dedicated `interpolate` method.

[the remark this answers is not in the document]
eslickj: @Robbybp, I agree that it makes sense to have a dedicated interpolate method.  I'll do that.  If someone wants to fall back to interpolation then, they can put in a try, except and maybe log a warning or something.

pyomo/contrib/mpc/data/interpolation.py
```diff
@@ -0,0 +1,68 @@
...
+        length as time data.
+
+    Returns
+    -------
+    list
+        If data are Pyomo components, this will return Pyomo expressions
+        interpolation if data are floats, this will return floats.
+    """
+    if indexes is None:
+        indexes = get_surround_time_vec(time_set, time_data)
```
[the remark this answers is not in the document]
eslickj: Yup

pyomo/contrib/mpc/data/series_data.py
```diff
@@ -104,7 +104,7 @@ def get_data_at_time_indices(self, indices):
                 {cuid: values[indices] for cuid, values in self._data.items()}
             )
 
-    def get_data_at_time(self, time=None, tolerance=0.0):
+    def get_data_at_time(self, time=None, tolerance=0.0, interpolate=False):
```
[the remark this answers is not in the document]
eslickj: Split to separate function now.

pyomo/contrib/mpc/data/series_data.py
```diff
@@ -152,6 +155,44 @@ def get_data_at_time(self, time=None, tolerance=0.0):
             indices = indices[0]
         return self.get_data_at_time_indices(indices)
 
+    def get_interpolated_data_at_time(self, time=None):
```
Robbybp, line 158: For concision, can you make this `get_interpolated_data`?

pyomo/contrib/mpc/data/series_data.py
```diff
@@ -152,6 +155,44 @@ def get_data_at_time(self, time=None, tolerance=0.0):
...
+        """
+        if time is None:
+            # If time is not specified, assume we want the entire time
+            # set. Skip all the overhead, don't create a new object, and
+            # return self.
+            return self
+        is_iterable = _is_iterable(time)
+        if not is_iterable:
+            time = [time]
+        idxs = _get_time_index_vec(time, self._time)
```
Robbybp, line 184: What happens if `time` contains points outside our `TimeSeriesData`'s range of time points? I think I'd want an error to be raised in this case. Can you add tests for cases with time points lower and higher than the acceptable range?

pyomo/contrib/mpc/data/series_data.py
```diff
@@ -181,6 +184,9 @@ def get_interpolated_data_at_time(self, time=None):
         is_iterable = _is_iterable(time)
         if not is_iterable:
             time = [time]
+        for t in time:
+            if t > self._time[-1] + tolerance or t < self._time[0] - tolerance:
```
Robbybp, line 188: With `<` and `>` here, this will fail with a tolerance of zero (the default) if the user provides one of the endpoints. Can you update to use `>=` and `<=`?

Robbybp: Unless you think the endpoints should not be valid for interpolation with a tolerance of zero. I'm open to this argument.

eslickj: You're right should have used >= and <=.  I'll fix that.

eslickj: On second look, I realized that the these are a check for whether an exception should be raised.  I think the original way was right.  When the time is equal to the first or last point and exception should not be raised.

Robbybp: You're right! Can you add a test with a time point at each endpoint (and default tolerance) so we don't accidentally break this later?

eslickj: Sure! will do.

Robbybp (approved):
Looks good once tests pass, thanks!

pyomo/contrib/mpc/data/interpolation.py
```diff
@@ -0,0 +1,73 @@
...
+    for i, t in enumerate(time_set):
+        pos[i] = bisect_right(time_data, t)
+        if pos[i] == 0:
+            pos[i] = 1
+        elif pos[i] == len(time_data):
+            pos[i] = len(time_data) - 1
+    return pos
+
+
+def _get_interp_expr_vec(time_set, time_data, data, indexes=None):
```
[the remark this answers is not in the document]
Robbybp: Is `indexes` used? If not, can it be removed?

mrmundt:
@eslickj - Please run `black -S -C .` on your code with the most recent version of `black`

blnicho (approved):

PR #3574, opened 2025-04-23
Pyomo.DoE bugfixes

djlaky:
## Fixes # .

1. Ensured square solve (e.g., fix the experiment inputs) when solving each scenario block.
2. Added `self.tee` to sequential compute FIM method

## Summary/Motivation:
Fixing some bugs found in Pyomo.DoE over the past month.

- Adding fixing/unfixing of experimental design decisions to ensure a square solve for each finite difference scenario
- Added a reference to the internal `tee` value in the compute FIM methods

### Legal Acknowledgement


djlaky:
@smondal13, @adowling2 

Bug we saw this morning that was fixed but not pushed.

adowling2:
@djlaky Is this ready for review by the Pyomo team?

adowling2 (approved):
This looks good to me.

pyomo/contrib/doe/doe.py
```diff
@@ -1119,8 +1119,17 @@ def build_block_scenarios(b, s):
             pyo.ComponentUID(param, context=m.base_model).find_component_on(
                 b
             ).set_value(m.base_model.unknown_parameters[param] * (1 + diff))
+
+            # Fix experiment inputs before solve (enforce square solve)
+            for comp in b.experiment_inputs:
+                comp.fix()
```
blnicho, line 1125: Is there any model checking before this that ensures fixing `b.experiment_inputs` will result in a square model?

blnicho: Also, is there any chance that a user will have already fixed some of these inputs in which case we shouldn't unfix them?

djlaky: > Is there any model checking before this that ensures fixing `b.experiment_inputs` will result in a square model?

There is not. 

Since the blocks are built on cloning the model from 'get_labeled_model()', the model output from this call needs to be square.

Also, is there a way to tell if the model is square using the 'res' object in line 1127? That would be the best case, then the user can get an error if their model is not square and this can be tested.

adowling2: I like the idea of this failing gracefully if the user tries running it with a non-square model. I have less of an opinion on if the check happens here or somewhere else.

adowling2: Maybe we have a custom function that inspects the build model and summarize the number of measurements, equations, parameters, inputs, etc. and estimates the degrees of freedom. I am thinking "If someone contacted me asking for help with ParmEst or Pyomo.DoE, what would I am to know about the model? I'm s there a way we can make the first debugging step to run "model_diagnostics" or similar and report the output.

PR #3575, opened 2025-04-23
Implementing multistart version of theta_est using multiple sampling methods

sscini:
## Summary/Motivation:
Currently, the optimization is only done from a single initial value. This implementation adds the ability to specify multiple initial values using selected sampling techniques: from a random uniform distribution, using Latin Hypercube Sampling, or using Sobol Quasi-Monte Carlo sampling.

- All changes made adding pseudocode in comments
- Added inputs needed for multistart simulation
- Added a function to generate points using the selected method
- Added theta_est_multistart to work for the multistart process

## TODO before converting from draft:
- [x] Receive feedback from collaborators on logical setup
- [x] Convert finalized pseudocode
- [x] Test and debug
- [x] Confirm function with examples

### Legal Acknowledgement


sscini:
@djlaky @adowling2 Please provide early feedback.

adowling2 (commented):
Notes from our in-person discussion/informal code review

sscini:
Extend existing tests for parmest to include multistart, add.

sscini:
Models provided need to include bounds, add exception

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +464,88 @@ def TotalCost_rule(model):
...
+            # Generate random theta values
+            theta_vals_multistart = np.random.uniform(lower_bound, upper_bound, size=len(theta_names))
+
+            # Generate theta values using Latin hypercube sampling or Sobol sampling
+            return theta_vals_multistart
+
+        elif self.method == "latin_hypercube":
+            # Generate theta values using Latin hypercube sampling
+            sampler = scipy.stats.qmc.LatinHypercube(d=len(theta_names), seed=seed)
+            samples = sampler.random(n=self.n_restarts+1)[1:]  # Skip the first sample
```
adowling2, line 507: Why are you skipping the first sample? Please explain in the comments.

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +464,88 @@ def TotalCost_rule(model):
...
+            theta_vals_multistart = np.array([lower_bound + (upper_bound - lower_bound) * theta for theta in samples])
```
adowling2, line 508: Add a comment to explain what you are doing here. I suspect that this function returns an LHS refined on [0, 1] and you are rescaling. But it is best to be explicit.

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +464,88 @@ def TotalCost_rule(model):
...
+
+
+        elif self.method == "sobol":
+            sampler = scipy.stats.qmc.Sobol(d=len(theta_names), seed=seed)
+            samples = sampler.random(n=self.n_restarts+1)[1:]
```
adowling2, line 513: Is this also skipping the first sample? Please explain.

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +464,88 @@ def TotalCost_rule(model):
...
+            theta_vals_multistart = np.array([lower_bound + (upper_bound - lower_bound) * theta for theta in samples])
```
adowling2, line 514: This is a repeated line of code. What do you think about generating the random uniform samples on [0, 1] and then sharing a common line of code to rescale?

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +464,88 @@ def TotalCost_rule(model):
...
+
+        # elif self.method == "prior":
```
adowling2, line 516: What do you mean by prior? I think you mean user supplied. "prior" makes me think about a probability distribution, i.e., the prior in statistical inference.

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +464,88 @@ def TotalCost_rule(model):
...
+
+        else:
+            raise ValueError(
+                "Invalid sampling method. Choose 'random', 'latin_hypercube', 'sobol'." # or 'prior'."
+            )
+        
+        # Make an output dataframe with the theta names and their corresponding values for each restart, 
+        # and nan for the output info values
+        df_multistart = pd.DataFrame(
+            theta_vals_multistart, columns=theta_names
```
adowling2, line 528: How do you want to distinguish column names for the initial versus the result? For each parameter, there will be two columns.

pyomo/contrib/parmest/parmest.py
```diff
@@ -921,6 +1020,116 @@ def theta_est(
...
+        solver="ef_ipopt",
+        return_values=[],
+    ):
+        """
+        Parameter estimation using multistart optimization
+
+        Parameters
+        ----------
+        n_restarts: int, optional
+            Number of restarts for multistart. Default is 1.
```
adowling2, line 1037: You should have n_restarts and method as inputs to this method.

pyomo/contrib/parmest/parmest.py
```diff
@@ -921,6 +1020,116 @@ def theta_est(
             cov_n=cov_n,
         )
 
+    def theta_est_multistart(
+        self,
+        buffer=10,
```
adowling2, line 1025: Need to explain the buffer in the doc string.

pyomo/contrib/parmest/parmest.py
```diff
@@ -921,6 +1020,116 @@ def theta_est(
...
+
+        """
+
+        # check if we are using deprecated parmest
+        if self.pest_deprecated is not None:
+            return print(
+                "Multistart is not supported in the deprecated parmest interface"
+            )
+
+        assert isinstance(self.n_restarts, int)
```
adowling2, line 1064: Replace all of these with more descriptive error messages. Remember that we need tests for each error message.

pyomo/contrib/parmest/parmest.py
```diff
@@ -921,6 +1020,116 @@ def theta_est(
...
+        assert isinstance(self.multistart_sampling_method, str)
+        assert isinstance(solver, str)
+        assert isinstance(return_values, list)
+        
+        if self.n_restarts > 1 and self.multistart_sampling_method is not None:
+            # Generate theta values using the sampling method
+            theta_vals, results_df = self._generate_initial_theta(
```
adowling2, line 1071: If you are already loading the theta_vals into the results_df, why are you returning both? Why not access the data from results_df?

pyomo/contrib/parmest/parmest.py
```diff
@@ -921,6 +1020,116 @@ def theta_est(
...
+                    solver=solver,
+                    return_values=return_values,
+                )
+
+                # Check if the solver terminated successfully
+                if variable_values.solver.termination_condition != pyo.TerminationCondition.optimal:
+                    # If not, set the objective value to NaN
+                    solver_termination = variable_values.solver.termination_condition
+                    solve_time = variable_values.solver.time
+                    thetavals = np.nan
```
adowling2, line 1092: It is helpful to still know the solver time.

pyomo/contrib/parmest/parmest.py
```diff
@@ -921,6 +1020,116 @@ def theta_est(
...
+                results_df.iloc[i, -3] = objectiveval
+                results_df.iloc[i, -2] = variable_values.solver.termination_condition
+                results_df.iloc[i, -1] = variable_values.solver.time
+
+                # Add buffer to save the dataframe dynamically, if save_results is True
+                if save_results and (i + 1) % buffer == 0:
+                    mode = 'w' if i + 1 == buffer else 'a'
+                    header = i + 1 == buffer
+                    results_df.to_csv(
+                        f"multistart_results.csv", mode=mode, header=header, index=False
```
adowling2, line 1120: The file name is a string that the user should provide. You can make the default "multistart_results.csv"

adowling2 (commented):
Here are some more comments for you to consider are you continue to refine this.

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +464,88 @@ def TotalCost_rule(model):
...
+            )
+        
+        # Check the length of theta_names and initial_theta, and make sure bounds are defined
+        if len(theta_names) != len(initial_theta):
+            raise ValueError(
+                "The length of theta_names and initial_theta must be the same."
+            )
+        
+        if self.method == "random":
+            np.random.seed(seed)
```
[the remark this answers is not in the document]
sscini: The default is none for all the functions I use, so if it receives seed = None, it would work as expected. Would skipping it still be best practice?

pyomo/contrib/parmest/parmest.py
sscini: I will add a comment in code to explain as well. The first sample generated using qmc.sobol is always the origin (zero vector). I thought logic applied to all qmc methods, but no only sobol. So to get nonzero points, you need to skip first sample

pyomo/contrib/parmest/parmest.py
sscini: Corrected, thanks.

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +455,130 @@ def TotalCost_rule(model):
         parmest_model = utils.convert_params_to_vars(model, theta_names, fix_vars=False)
 
         return parmest_model
+    
+    # Make new private method, _generate_initial_theta:
+    # This method will be used to generate the initial theta values for multistart
+    # optimization. It will take the theta names and the initial theta values
+    # and return a dictionary of theta names and their corresponding values.
+    def _generate_initial_theta(self, parmest_model, seed=None, n_restarts=None, multistart_sampling_method=None, user_provided=None):
```
djlaky, line 463: All methods (even private) should have a description of each input, what type is expected, and what it's for in a brief yet coherent sentence.

djlaky: Also, it's probably a good idea to make sure each of the inputs are valid, in some sense, when called. For instance, if parmest_model is not a model in a specific form, there may be an error. We didn't require a `model` input for DoE, instead made it `None` and handled the case where this was sent.

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +455,130 @@ def TotalCost_rule(model):
...
+        if n_restarts == 1:
+            # If only one restart, return an empty list
+            return print("No multistart optimization needed. Please use normal theta_est()")
+        
+        # Get the theta names and initial theta values
+        theta_names = self._return_theta_names()
+        initial_theta = [parmest_model.find_component(name)() for name in theta_names]
```
djlaky, line 470: Is it better to use the suffix for this? The suffix value shouldn't change, but the theta value may if the model has been solved for some reason. I don't know if this is a potential issue but I think that grabbing these values from the suffixes would be more dummy-proof.

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +455,130 @@ def TotalCost_rule(model):
...
+        # Check the length of theta_names and initial_theta, and make sure bounds are defined
+        if len(theta_names) != len(initial_theta):
+            raise ValueError(
+                "The length of theta_names and initial_theta must be the same."
+            )
+        
+        if multistart_sampling_method == "random":
+            np.random.seed(seed)
+            # Generate random theta values
+            theta_vals_multistart = np.random.uniform(lower_bound, upper_bound, size=len(theta_names))
```
djlaky, line 490: Is there any merit in allowing the user to pull from a non-uniform distribution? Unsure what extra machinery is required but maybe for a future PR/enhancement.

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +455,130 @@ def TotalCost_rule(model):
...
+                "The lower and upper bounds for the theta values must be defined."
+            )
+        
+        # Check the length of theta_names and initial_theta, and make sure bounds are defined
+        if len(theta_names) != len(initial_theta):
+            raise ValueError(
+                "The length of theta_names and initial_theta must be the same."
+            )
+        
+        if multistart_sampling_method == "random":
```
djlaky, line 487: I don't necessarily like that the first method is called "random". Aren't most of the methods inherently based on random sampling? It should probably be more explicit, like uniform sampling, or random uniform, or something like this.

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +455,130 @@ def TotalCost_rule(model):
...
+
+
+        elif multistart_sampling_method == "sobol":
+            sampler = scipy.stats.qmc.Sobol(d=len(theta_names), seed=seed)
+            # Generate theta values using Sobol sampling
+            # The first value of the Sobol sequence is 0, so we skip it
+            samples = sampler.random(n=n_restarts+1)[1:]
+            theta_vals_multistart = np.array([lower_bound + (upper_bound - lower_bound) * theta for theta in samples])
+
+        elif multistart_sampling_method == "user_provided":
```
djlaky, line 508: I think you probably should have a comment to start each different method that describes what you're doing. It is not entirely obvious what's going on when you haven't yet read the code.

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +455,130 @@ def TotalCost_rule(model):
...
+                # Add user provided dataframe option
+            if user_provided is not None:
+
+                if isinstance(user_provided, np.ndarray):
+                    # Check if the user provided numpy array has the same number of rows as the number of restarts
+                    if user_provided.shape[0] != n_restarts:
```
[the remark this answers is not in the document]
djlaky: This is especially problematic is N = M, as the users would get results back and not know that they have made a mistake.

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +455,130 @@ def TotalCost_rule(model):
...
+
+        # Add the initial theta values to the first row of the dataframe
+        for i in range(1, n_restarts):
+            df_multistart.iloc[i, :] = theta_vals_multistart[i, :]
+        df_multistart.iloc[0, :] = initial_theta
+
+
+        # Add the output info values to the dataframe, starting values as nan
+        for i in range(len(theta_names)):
+            df_multistart[f'converged_{theta_names[i]}'] = np.nan
```
djlaky, line 575: Are all string characters legal for pandas names? I would think so but this line seems (maybe) dangerous? For instance, what if we start getting into block stucturing with multi-index parameters? We should make sure there is a test to ensure the system is robust. I believe I have this on the back burner to make one for Pyomo.DoE as well.

pyomo/contrib/parmest/parmest.py
```diff
@@ -921,6 +1053,136 @@ def theta_est(
...
+        Parameter estimation using multistart optimization
+
+        Parameters
+        ----------
+        n_restarts: int, optional
+            Number of restarts for multistart. Default is 1.
+        theta_sampling_method: string, optional
+            Method used to sample theta values. Options are "random", "latin_hypercube", or "sobol".
+            Default is "random".
+        solver: string, optional
```
djlaky, line 1079: Is there a way to pass custom solver options? I'm not super familiar with this but if this is called separate of the standard call, you need to make sure the user can specify their own solver options.

pyomo/contrib/parmest/parmest.py
```diff
@@ -921,6 +968,125 @@ def theta_est(
...
+            Covariance matrix of the fitted parameters (only for solver='ef_ipopt')
+        """
+
+        # check if we are using deprecated parmest
+        if self.pest_deprecated is not None:
+            return print(
+                "Multistart is not supported in the deprecated parmest interface")
+            )
+
+        assert isinstance(n_restarts, int)
```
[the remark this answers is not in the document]
djlaky: Agree with @adowling2 here, you need to throw an exception so you can test the exception is caught.

pyomo/contrib/parmest/parmest.py
```diff
@@ -921,6 +1053,136 @@ def theta_est(
...
+                    solve_time = variable_values.solver.time
+
+                # Check if the objective value is better than the best objective value  
+                if final_objectiveval < best_objectiveval:
+                    best_objectiveval = objectiveval
+                    best_theta = thetavals
+                    
+                # Store the results in a list or DataFrame
+                # depending on the number of restarts
+                results_df.iloc[i, len(theta_names):len(theta_names) + len(theta_names)] = converged_theta_vals[i, :]
```
djlaky, line 1162: It would be good to have the general format of the pandas frame described in text so lines like this can be checked by reviewers. Currently, it is very hard to check to make sure this is correct without running code ourselves.

adowling2:
@sscini Tag me when this is ready for the next review

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +455,130 @@ def TotalCost_rule(model):
         parmest_model = utils.convert_params_to_vars(model, theta_names, fix_vars=False)
 
         return parmest_model
+    
+    # Make new private method, _generate_initial_theta:
+    # This method will be used to generate the initial theta values for multistart
+    # optimization. It will take the theta names and the initial theta values
+    # and return a dictionary of theta names and their corresponding values.
+    def _generate_initial_theta(self, parmest_model, seed=None, n_restarts=None, multistart_sampling_method=None, user_provided=None):
+        if n_restarts == 1:
```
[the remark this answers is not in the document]
sscini: Alex had initially said just return message, but I will ask him again

sscini:
@adowling2 @djlaky Ready for next round of review, parmest.py file only, example still being developed. Currently functional (a few minor issues) with following output of separate MM example:

Setting theta_1 to 124.2278222925961
Setting theta_2 to 0.7336450144648552
Restart 1/5: Objective Value = 41.49573453661323, Theta = theta_1    212.683743
theta_2      0.064121
dtype: float64
Setting theta_1 to 201.92126417532563
Setting theta_2 to 0.18795708194375038
Restart 2/5: Objective Value = 41.49573453661323, Theta = theta_1    212.683743
theta_2      0.064121
dtype: float64
Setting theta_1 to 438.33825131878257
Setting theta_2 to 0.8983656093478203
Restart 3/5: Objective Value = 41.49573453661323, Theta = theta_1    212.683743
theta_2      0.064121
dtype: float64
Setting theta_1 to 401.7291865311563
Setting theta_2 to 0.01574030425399542
Restart 4/5: Objective Value = 41.49573453661323, Theta = theta_1    212.683743
theta_2      0.064121
dtype: float64
Setting theta_1 to 145.7563922740519
Setting theta_2 to 0.8198425201699138
[/Applications/anaconda3/envs/parmest-dev-mac/lib/python3.13/site-packages/scipy/stats/_qmc.py:993](https://file+.vscode-resource.vscode-cdn.net/Applications/anaconda3/envs/parmest-dev-mac/lib/python3.13/site-packages/scipy/stats/_qmc.py:993): UserWarning: The balance properties of Sobol' points require n to be a power of 2.
  sample = self._random(n, workers=workers)
[/Users/scini/Documents/GitHub/pyomo/pyomo/contrib/parmest/parmest.py:1205](https://file+.vscode-resource.vscode-cdn.net/Users/scini/Documents/GitHub/pyomo/pyomo/contrib/parmest/parmest.py:1205): FutureWarning: Series.__getitem__ treating keys as positions is deprecated. In a future version, integer keys will always be treated as labels (consistent with DataFrame behavior). To access a value by position, use `ser.iloc[pos]`
  results_df.at[i, f'converged_{name}'] = converged_theta[j] if not np.isnan(converged_theta_vals[i, j]) else np.nan
[/Users/scini/Documents/GitHub/pyomo/pyomo/contrib/parmest/parmest.py:1209](https://file+.vscode-resource.vscode-cdn.net/Users/scini/Documents/GitHub/pyomo/pyomo/contrib/parmest/parmest.py:1209): FutureWarning: Setting an item of incompatible dtype is deprecated and will raise an error in a future version of pandas. Value 'optimal' has dtype incompatible with float64, please explicitly cast to a compatible dtype first.
  results_df.at[i, "solver termination"] = solver_termination if 'solver_termination' in locals() else np.nan
[/Users/scini/Documents/GitHub/pyomo/pyomo/contrib/parmest/parmest.py:1205](https://file+.vscode-resource.vscode-cdn.net/Users/scini/Documents/GitHub/pyomo/pyomo/contrib/parmest/parmest.py:1205): FutureWarning: Series.__getitem__ treating keys as positions is deprecated. In a future version, integer keys will always be treated as labels (consistent with DataFrame behavior). To access a value by position, use `ser.iloc[pos]`
  results_df.at[i, f'converged_{name}'] = converged_theta[j] if not np.isnan(converged_theta_vals[i, j]) else np.nan
Restart 5/5: Objective Value = 41.49573453661323, Theta = theta_1    212.683743
theta_2      0.064121
dtype: float64

Thanks!

sscini:
Note for reviewers: The current form of the code attempts to but cannot successfully rewrite the parameters for each iteration, and gives the same result each time. Advice on how to set this value from within theta_est_multistart() would be appreciated.

Update: Looks like something else. In the model pprint of the integrated ef_instance, the parameters are stale. So they are set but something else is happening. Looking into fix now.

sscini:
I added my current example, using simple multimodal model (described in file) as a temporary .ipynb that has the data gen, model formation, a manual for loop version of multistart and the current implemented version to show the output differences.

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +455,145 @@ def TotalCost_rule(model):
...
+    # This method will be used to generate the initial theta values for multistart
+    # optimization. It will take the theta names and the initial theta values
+    # and return a dictionary of theta names and their corresponding values.
+    def _generate_initial_theta(self, parmest_model=None, seed=None, n_restarts=None, multistart_sampling_method=None, user_provided=None):
+        """
+        Generate initial theta values for multistart optimization using selected sampling method.
+        """    
+        if n_restarts == 1:
+            # If only one restart, return an empty list
+            return print("No multistart optimization needed. Please use normal theta_est()")
```
djlaky: You should raise a warning/log something here instead of using a print statement. That way you can use a debugger to control whether the message is displayed.

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +455,145 @@ def TotalCost_rule(model):
...
+        
+        # Get the theta names and initial theta values
+        theta_names = self._return_theta_names()
+        initial_theta = [parmest_model.find_component(name)() for name in theta_names]
```
djlaky: Are we sure we can't do this with suffixes? It feels like you should be able to do all three of these (also for upper and lower bounds) using the suffixes on parmest_model?

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +455,145 @@ def TotalCost_rule(model):
...
+        lower_bound = np.array([parmest_model.find_component(name).lb for name in theta_names])
+        upper_bound = np.array([parmest_model.find_component(name).ub for name in theta_names])
+        # Check if the lower and upper bounds are defined
+        if any(bound is None for bound in lower_bound) and any(bound is None for bound in upper_bound):
+            raise ValueError(
+                "The lower and upper bounds for the theta values must be defined."
+            )
+        
+        # Check the length of theta_names and initial_theta, and make sure bounds are defined
+        if len(theta_names) != len(initial_theta):
```
djlaky: I think this is unnecessary. You populate `initial_theta` using `theta_names` so I would hope it's not possible to even design a test that could get here.

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +455,145 @@ def TotalCost_rule(model):
...
+            raise ValueError(
+                "The length of theta_names and initial_theta must be the same."
+            )
+        
+        if multistart_sampling_method == "uniform":
```
djlaky: Maybe `uniform_random` would be more descriptive?

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +455,145 @@ def TotalCost_rule(model):
...
+            samples = sampler.random(n=n_restarts)
+            # Resulting samples should be size (n_restarts, len(theta_names))
+
+        elif multistart_sampling_method == "sobol":
+            sampler = scipy.stats.qmc.Sobol(d=len(theta_names), seed=seed)
+            # Generate theta values using Sobol sampling
+            # The first value of the Sobol sequence is 0, so we skip it
+            samples = sampler.random(n=n_restarts+1)[1:]
+
+        elif multistart_sampling_method == "user_provided":
```
djlaky: I think maybe `user_provided` does not imply values. For instance, what if a user wants to provide their own random sample generator?

Probably should use `user_provided_values` to be more explicit.

Also, should have some comments at the beginning of this describing what the method does, just like the other options.

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +455,145 @@ def TotalCost_rule(model):
...
+                # Add user provided dataframe option
+            if user_provided is not None:
+
+                if isinstance(user_provided, np.ndarray):
+                    # Check if the user provided numpy array has the same number of rows as the number of restarts
```
djlaky: Make sure comments are not too long. Break up over multiple lines like you did above...

djlaky: Running black may help?

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +455,145 @@ def TotalCost_rule(model):
...
+                        )
+                    # Check if the user provided numpy array has the same number of columns as the number of theta names
+                    if user_provided.shape[1] != len(theta_names):
+                        raise ValueError(
+                            "The user provided numpy array must have the same number of columns as the number of theta names."
+                        )
+                    # Check if the user provided numpy array has the same theta names as the model
+                    # if not, raise an error
+                    # if not all(theta in theta_names for theta in user_provided.columns):
+                        raise ValueError(
```
djlaky: Should this be deleted? It seems to be unnecessary currently?

Is this for another case that is being added?

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +455,145 @@ def TotalCost_rule(model):
...
+                            "The user provided dataframe must have the same number of rows as the number of restarts."
+                        )
+                    # Check if the user provided dataframe has the same number of columns as the number of theta names
+                    if user_provided.shape[1] != len(theta_names):
+                        raise ValueError(
+                            "The user provided dataframe must have the same number of columns as the number of theta names."
+                        )
+                    # Check if the user provided dataframe has the same theta names as the model
+                    # if not, raise an error
+                    # if not all(theta in theta_names for theta in user_provided.columns):
```
djlaky: This should be uncommented...

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +455,145 @@ def TotalCost_rule(model):
...
+                        raise ValueError(
+                            "The user provided dataframe must have the same theta names as the model."
+                        )
+                # If all checks pass, return the user provided dataframe
+                theta_vals_multistart = user_provided.iloc[0: len(initial_theta)].values
+            else:
+                raise ValueError(
+                    "The user must provide a numpy array or pandas dataframe from a previous attempt to use the 'user_provided' method."
```
djlaky: Same thing with all these long output messages, make sure they don't exceed one line (break them up).

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +455,145 @@ def TotalCost_rule(model):
...
+                )
+
+        else:
+            raise ValueError(
```
djlaky, line 1667: This would probably be more consistent with other code (and other suggestions) if the options were using an Enum object. You can check the DoE code, or Shammah's PR. It just makes it so that the strings are attached to an object instead (safer).

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +455,145 @@ def TotalCost_rule(model):
...
+                "Invalid sampling method. Choose 'uniform', 'latin_hypercube', 'sobol'  or 'user_provided'."
+            )
+        
+        if multistart_sampling_method == "sobol" or multistart_sampling_method == "latin_hypercube":
+            # Scale the samples to the range of the lower and upper bounds for each theta in theta_names
```
djlaky, line 1675: You need to add comments to those methods above that indicate the sampling renders numbers in (0, 1] or [-1, 1], whichever it is.

pyomo/contrib/parmest/parmest.py
```diff
@@ -447,6 +455,145 @@ def TotalCost_rule(model):
...
+                raise ValueError(
+                    "The user must provide a numpy array or pandas dataframe from a previous attempt to use the 'user_provided' method."
+                )
+
+        else:
+            raise ValueError(
+                "Invalid sampling method. Choose 'uniform', 'latin_hypercube', 'sobol'  or 'user_provided'."
+            )
+        
+        if multistart_sampling_method == "sobol" or multistart_sampling_method == "latin_hypercube":
```
djlaky: I appreciate the flow of the code being:

1. Generate samples
2. Load into dataframe
3. Set up dataframe for success

However, I wonder if keeping each step inside it's own setup above makes the code more interpretable.

pyomo/contrib/parmest/parmest.py
```diff
@@ -921,6 +1068,171 @@ def theta_est(
...
+        """
+
+        # check if we are using deprecated parmest
+        if self.pest_deprecated is not None:
+            return print(
+                "Multistart is not supported in the deprecated parmest interface"
+            )
+
+        assert isinstance(n_restarts, int)
+        assert isinstance(multistart_sampling_method, str)
```
djlaky: Still need to change these into exception-throwing statements.

pyomo/contrib/parmest/parmest.py
```diff
@@ -921,6 +1068,171 @@ def theta_est(
...
+        assert isinstance(solver, str)
+        assert isinstance(return_values, list)
+        
+        if n_restarts > 1 and multistart_sampling_method is not None:
+            
+            # Find the initialized values of theta from the labeled parmest model
+            # and the theta names from the estimator object
+
+            # print statement to indicate multistart optimization is starting
+            print(f"Starting multistart optimization with {n_restarts} restarts using {multistart_sampling_method} sampling method.")
```
djlaky: Again, this should be a logging output, not a print statement.

pyomo/contrib/parmest/parmest.py
```diff
@@ -921,6 +1068,171 @@ def theta_est(
...
+                # Create a fresh model for each restart
+                parmest_model = self._create_parmest_model(experiment_number=0)
+                theta_vals_current = theta_vals.iloc[i, :].to_dict()
+
+                # Set current theta values in the model
+                for name, value in theta_vals_current.items():
+                    parmest_model.find_component(name).set_value(value)
+
+                # Optional: Print the current theta values being set
+                print(f"Setting {name} to {value}")
```
djlaky: Same comment about logging.

pyomo/contrib/parmest/parmest.py
```diff
@@ -921,6 +1068,171 @@ def theta_est(
...
+                for name in theta_names:
+                    current_value = parmest_model.find_component(name)()
+                    print(f"Current value of {name} is {current_value}")
```
djlaky: Same comment about logging

pyomo/contrib/parmest/parmest.py
```diff
@@ -921,6 +1068,171 @@ def theta_est(
...
+                parmest_model = self._create_parmest_model(experiment_number=0)
+                theta_vals_current = theta_vals.iloc[i, :].to_dict()
+
+                # Set current theta values in the model
+                for name, value in theta_vals_current.items():
+                    parmest_model.find_component(name).set_value(value)
+
+                # Optional: Print the current theta values being set
+                print(f"Setting {name} to {value}")
+                for name in theta_names:
```
djlaky: What is this for? Why is this not just included in the previous for loop?

pyomo/contrib/parmest/parmest.py
```diff
@@ -921,6 +1068,171 @@ def theta_est(
...
+                )
+
+                # Unpack results 
+                objectiveval, converged_theta = qopt_result
+        
+
+                # Since _Q_opt does not return the solver result object, we cannot check
+                # solver termination condition directly here. Instead, we can assume
+                # that if converged_theta contains NaN, the solve failed.
+                if converged_theta.isnull().any():
```
djlaky: This probably doesn't catch all instances of model failure. 

It would be better if you could track the termination condition of the result within `_Q_opt` and save it to the object. Then you can reset the field each time a solve happens. The field can even be related to multistart. Open to discussion but I'm fairly certain that bad solves can render normal values.

pyomo/contrib/parmest/parmest.py
```diff
@@ -921,6 +1068,171 @@ def theta_est(
...
+
+                # # Check if the objective value is better than the best objective value
+                # # Set a very high initial best objective value
+                # best_objectiveval = np.inf  
+                # best_theta = np.inf
+                # if final_objectiveval < best_objectiveval:
+                #     best_objectiveval = objectiveval
+                #     best_theta = thetavals
+                    
+                print(f"Restart {i+1}/{n_restarts}: Objective Value = {final_objectiveval}, Theta = {converged_theta}")
```
djlaky: Same as other logging comments.

pyomo/contrib/parmest/parmest.py
```diff
@@ -921,6 +1068,171 @@ def theta_est(
...
+                objectiveval, converged_theta = qopt_result
+        
+
+                # Since _Q_opt does not return the solver result object, we cannot check
+                # solver termination condition directly here. Instead, we can assume
+                # that if converged_theta contains NaN, the solve failed.
+                if converged_theta.isnull().any():
+                    solver_termination = "not successful"
+                    solve_time = np.nan
+                    thetavals = np.nan
```
djlaky: This is never used?

pyomo/contrib/parmest/parmest.py
```diff
@@ -921,6 +1068,171 @@ def theta_est(
...
+                # print(results_df)
+
+                # Add buffer to save the dataframe dynamically, if save_results is True
+                if save_results and (i + 1) % buffer == 0:
+                    mode = 'w' if i + 1 == buffer else 'a'
+                    header = i + 1 == buffer
+                    results_df.to_csv(
+                        file_name, mode=mode, header=header, index=False
+                    )
+                    print(f"Intermediate results saved after {i + 1} iterations.")
```
djlaky: Same comment about logging.

pyomo/contrib/parmest/parmest.py
```diff
@@ -921,6 +1068,171 @@ def theta_est(
...
+
+            # Final save after all iterations
+            if save_results:
+                results_df.to_csv(file_name, mode='a', header=False, index=False)
+                print("Final results saved.")
```
djlaky: Same comment about logging.

pyomo/contrib/parmest/parmest.py
```diff
@@ -921,6 +1068,171 @@ def theta_est(
...
+                if save_results and (i + 1) % buffer == 0:
+                    mode = 'w' if i + 1 == buffer else 'a'
+                    header = i + 1 == buffer
+                    results_df.to_csv(
+                        file_name, mode=mode, header=header, index=False
+                    )
+                    print(f"Intermediate results saved after {i + 1} iterations.")
+
+            # Final save after all iterations
+            if save_results:
```
djlaky: If this uses a buffer, will this not add the whole data to the intermediately saved data?

pyomo/contrib/parmest/parmest.py
```diff
@@ -921,6 +1068,171 @@ def theta_est(
...
+
+                # Add buffer to save the dataframe dynamically, if save_results is True
+                if save_results and (i + 1) % buffer == 0:
+                    mode = 'w' if i + 1 == buffer else 'a'
+                    header = i + 1 == buffer
+                    results_df.to_csv(
+                        file_name, mode=mode, header=header, index=False
+                    )
+                    print(f"Intermediate results saved after {i + 1} iterations.")
+
```
djlaky, line 2249: You need to append the remaining data with the buffer-based saving system when `n_restarts` is not a multiple of 10.

pyomo/contrib/parmest/parmest.py
djlaky: Also, using "out of" instead of "/" is probably a better choice.

pyomo/contrib/parmest/parmest.py
```diff
@@ -921,6 +1068,171 @@ def theta_est(
...
+                results_df.at[i, "final objective"] = objectiveval if 'objectiveval' in locals() else np.nan
+                results_df.at[i, "solver termination"] = solver_termination if 'solver_termination' in locals() else np.nan
+                results_df.at[i, "solve_time"] = solve_time if 'solve_time' in locals() else np.nan
+
+                # Diagnostic: print the table after each restart
+                # print(results_df)
+
+                # Add buffer to save the dataframe dynamically, if save_results is True
+                if save_results and (i + 1) % buffer == 0:
+                    mode = 'w' if i + 1 == buffer else 'a'
```
djlaky: Does this keep appending the whole data frame each time to the csv? It's not clear that this is actually efficient. I'm concerned this will make files that have an enormous amount of entries.

djlaky (commented):
Finished up my initial review. Mostly comments about logging, shifting to Enum instead of strings, wondering why we aren't leveraging the `suffixes` from the model, and concerns about the buffered saving.

pyomo/contrib/parmest/parmest.py
```diff
@@ -91,7 +91,7 @@ def ef_nonants(ef):
 
 
 def _experiment_instance_creation_callback(
-    scenario_name, node_names=None, cb_data=None
+    scenario_name, node_names=None, cb_data=None, fix_vars=False,
```
adowling2: I would support having utility methods `_fix_unknown_parameters` and `_fix_experiment_inputs` live within the `Experiment` class. These methods could have `fix=True` as a default, and `fix=False` would unfix.

sscini:
Plans this week:

- Address above comments
- Write a function to reinitialize the model using the suffix values to remove need for strings
- Finish "user_provided_)samples" method
- Finish new examples to confirm functionality
- Bring new questions to 6/17 dev meeting

sscini:
# Notes for Pyomo Dev Meeting 6/17/25:
- Plan to complete plan from previous message
**Context**
- Currently, Q_opt creates the parmest model using experiment_creation_callback (line 89) and then compares it to data.
- Only option is if ThetaVals are provided: Fix them
- I added an option to unfix them but provide them, which is needed for my multistart, but this requires strings, and I have heard a lot of work was spent moving away from strings.
**Problem:**
- No way to access that model or insert a prebuilt pyomo model into Q_opt
**Proposed solution**
- I would like to make a function to refactor or change the initialized values of the function and then send that model to Q_opt. This would prevent the need for strings to be used to provide ThetaVals to Qopt. Does this make sense?

adowling2:
Update: The ND Pyomo.DoE/ParmEst team decided to split these ideals into two PRs.
1. This first PR will add core functionality needed to enable a few other contributions, such as a sensitivity analysis routine from @smondal13
2. The second PR will add a clear user interface, documentation, examples, possibly parallelization, ...

pyomo/contrib/parmest/parmest.py
sscini: Added Enum class above, working to implement here. Making note to talk to shammah about this

pyomo/contrib/parmest/parmest.py
sscini: The idea is that the first iteration creates the file, and then the file is appended. I agree it is unclear. Will discuss more next week

sscini:
Will not be actively worked on before finalization of _Q_opt_blocks. Closing for now to resume in the future.

sscini:
Notes for 02/24/26 development meeting:
Will need to merge _Q_opt update (PR 3789) first
Variant of objective_at_theta that estimates parameters at each parameter value set.
Still need to:
- [x] Update implementation of sample generation using random sampling techniques
- [x] Define defaults for sampling technique, number of restarts, etc
- [ ] Generate examples from literature with known solutions.
- [ ] Testing
- [ ] Documentation

sscini:
Now that _Q_opt has been finalized and merged, going to revisit this, make adjustments as needed from final _Q_opt and regularization revisions, and then will make a regular PR.

sscini:
@mrmundt Hi Miranda! I had a quick testing question if you can help clarify something. It looks like if a test has been marked expensive, it does not count as covered (according to coverage reports from individual checks). If this is as intended, is the goal to change those tests long-term? Curious your thoughts. Thanks!

mrmundt:
> @mrmundt Hi Miranda! I had a quick testing question if you can help clarify something. It looks like if a test has been marked expensive, it does not count as covered (according to coverage reports from individual checks). If this is as intended, is the goal to change those tests long-term? Curious your thoughts. Thanks!

HELLO! Okay so, yes, expensive tests aren't run in the standard suite. We have a specific nightly job that exercises them on Jenkins (which does then upload a coverage report), but it only runs on `main`. The purpose behind that is, because they are expensive and we already have such a large test suite, we don't turn those on for individual PR testing in order to speed up the suite.

jsiirola:
(@mrmundt: actually, the Jenkins PR tests *do* run the expensive tests ... we need to have the `main` and PR tests run the same things in order to get meaningful coverage information)

mrmundt:
> (@mrmundt: actually, the Jenkins PR tests _do_ run the expensive tests ... we need to have the `main` and PR tests run the same things in order to get meaningful coverage information)

Oh is that new? Because that wasn't the case before. ALRIGHT. Then it seems like it's a bug that the coverage reports aren't showing coverage on the expensive suite, too.

EDIT: AHA! Right, they _only_ run on Jenkins, and yours doesn't have the Jenkins job yet :P

PR #3576, opened 2025-04-24
Update Supported NEOS Solvers

mrmundt:
## Summary/Motivation:
We started having test failures because we were not capturing that we can now support BARON through the NEOS interface. This fixes that (in that we actually test it and have it recorded now) as well as fixing some language in the docs.

## Changes proposed in this PR:
- Update tests and docs

### Legal Acknowledgement

By contributing to this software project, I have read the [contribution guide](https://pyomo.readthedocs.io/en/stable/contribution_guide.html) and agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #3577, opened 2025-04-24
Detailed IPOPT Log in `ipopt_v2`

### LINKED ISSUE #548: return final Ipopt values in solution object
It would be nice to get Iter, objective, inf_pr, inf_du, lg(mu),...lg(rg), ... etc. from Ipopt in the solution object that is returned. Just to be clear: the workaround is to parse the outupt and get these values.

### LINKED ISSUE #728: Promote ipopt_solver_wrapper to a general utility
In #706 an Ipopt solver wrapper was introduced that reads the Ipopt output and returns a number of useful convergence statistics including the number of iterations, solve time, checking for regularization at the solution, etc.. I think this is a super useful utility that should be moved out of `contrib/parmest` to somewhere like `pyomo/util` or `pyomo/solvers`.

See the wrapper here: https://github.com/Pyomo/pyomo/blob/main/pyomo/contrib/parmest/utils/ipopt_solver_wrapper.py

mrmundt:
## Summary/Motivation:
There is a utility wrapper in parmest that gathers helpful data from the IPOPT log that has never been promoted to a "real world citizen." This PR replaces the need for that in `ipopt_v2` by parsing the output and storing important data by default on the `Results` object.

NOTE: This PR contains #3576 so that one needs to be merged first.

- Update IPOPT parsing logic to capture iterations, various final scaled/unscaled results, the full solver log, etc.
- Update tests to ensure that we are actually capturing the output
- Fix a really sneaky bug that only appeared on older Python versions in Windows that stopped output from being captured

### Legal Acknowledgement


pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -431,16 +448,26 @@ def solve(self, model, **kwds):
                     results.termination_condition = TerminationCondition.error
                     results.solution_loader = SolSolutionLoader(None, None)
                 else:
-                    results.iteration_count = iters
-                    if ipopt_time_nofunc is not None:
+                    results.iteration_count = parsed_output_data['iters']
```
michaelbynum, line 451: We probably don't want to generate an exception if this dictionary is empty (e.g., no log file).

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -431,16 +448,26 @@ def solve(self, model, **kwds):
...
+                        results.timing_info.nlp_function_evaluations = (
+                            parsed_output_data['cpu_seconds']['func_time']
+                        )
+                    parsed_output_data.pop('cpu_seconds')
+                    results.extra_info = parsed_output_data
+                    # Set iteration_log visibility to ADVANCED_OPTION because it's
+                    # a lot to print out with `display`
+                    results.extra_info.get("iteration_log")._visibility = (
+                        ADVANCED_OPTION
+                    )
```
michaelbynum, line 470: I'm confused. What type of object does `results.extra_info.get("iteration_log")` return? From below, I thought it would be a list.

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -380,23 +393,27 @@ def solve(self, model, **kwds):
...
+                        )
+                        timer.stop('subprocess')
+                except OSError:
+                    err = sys.exc_info()[1]
+                    msg = 'Could not execute the command: %s\tError message: %s'
+                    raise ApplicationError(msg % (cmd, err))
+
+                # This is the stuff we need to parse to get the iterations
+                # and time
+                parsed_output_data = self._parse_ipopt_output(ostreams[0])
```
michaelbynum, line 416: I think we should have an option to turn off this parsing.

pyomo/contrib/solver/tests/solvers/test_ipopt.py
```diff
@@ -302,6 +383,21 @@ def test_ipopt_config(self):
         self.assertFalse(solver.config.tee)
         self.assertTrue(solver.config.executable.startswith('/path'))
 
-        # Change value on a solve call
-        # model = self.create_model()
-        # result = solver.solve(model, tee=True)
+    def test_ipopt_solve(self):
+        model = self.create_model()
+        ipopt.Ipopt().solve(model)
```
michaelbynum, line 388: Does this test just make sure the solver runs without an exception?

michaelbynum (changes requested):
A couple minor comments. Yay for improved type hints!

pyomo/contrib/solver/tests/solvers/test_ipopt.py
mrmundt: Yup! It's a gut-check test (because that model shouldn't fail to solve).

pyomo/contrib/solver/solvers/ipopt.py
mrmundt: Thinking out loud:

- For what reason? We have to parse anyway to get timing info, which we have agreed is the new universal standard for solver results.
- If we do want to only parse for timing and not for other stuff, what is the default (e.g., is the default `True` - parse or `False` - don't parse)?
- In what instance is it not helpful to have this information?

pyomo/contrib/solver/solvers/ipopt.py
mrmundt: Oop, good call. I had a check in here in earlier iterations, but I forgot, I removed it when I ran into the weird Windows bug because I was trying to diagnose how to fix it. Will add back in the earlier logic here!

pyomo/contrib/solver/solvers/ipopt.py
mrmundt: Correct - it's a list. (It can be a very _long_ list, because it captures all of the iterations. See https://github.com/IDAES/idaes-pse/issues/92#issuecomment-2828127621)

michaelbynum: I'm just confused about how it has the `_visibility` attribute...

michaelbynum (approved):

michaelbynum:
Oh, is there a copy of the log parsing code in parmest that needs to be removed?

pyomo/contrib/solver/solvers/ipopt.py
mrmundt: Oh, because it's an implicitly declared `ConfigValue`.

mrmundt:
> Oh, is there a copy of the log parsing code in parmest that needs to be removed?

I'm hesitant to remove anything from parmest because I don't know how it's being used outside of Pyomo (e.g., it might be "the thing" that people are using to parse IPOPT output in other places). I figure it's something we can either (a) leave because it does no harm or (b) remove in the future with a fair deprecation warning.

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -499,46 +535,136 @@ def solve(self, model, **kwds):
...
+
+        # Extract scaled and unscaled table
+        scaled_unscaled_match = re.findall(
+            r'Objective\.\.+:\s+([-+eE0-9.]+)\s+([-+eE0-9.]+).*?'
+            r'Dual infeasibility\.\.+:\s+([-+eE0-9.]+)\s+([-+eE0-9.]+).*?'
+            r'Constraint violation\.\.+:\s+([-+eE0-9.]+)\s+([-+eE0-9.]+).*?'
+            r'Complementarity\.\.+:\s+([-+eE0-9.]+)\s+([-+eE0-9.]+).*?'
+            r'Overall NLP error\.\.+:\s+([-+eE0-9.]+)\s+([-+eE0-9.]+)',
+            output,
+            re.DOTALL,
```
jsiirola, line 625: Does the use of DOTALL and `.*?` make this a harder/more expensive regex?  Could we do the following:
```python
 scaled_unscaled_match = re.findall(
            r'Objective\.*:\s*([-+eE0-9.]+)\s+([-+eE0-9.]+)\s*'
            r'Dual infeasibility\.*:\s*([-+eE0-9.]+)\s+([-+eE0-9.]+)\s*'
            r'Constraint violation\.*:\s*([-+eE0-9.]+)\s+([-+eE0-9.]+)\s*'
            r'Complementarity\.*:\s*([-+eE0-9.]+)\s+([-+eE0-9.]+)\s*'
            r'Overall NLP error\.*:\s*([-+eE0-9.]+)\s+([-+eE0-9.]+)',
            output,
)
```

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -499,46 +535,136 @@ def solve(self, model, **kwds):
...
+                "complementarity_error": float(scaled_unscaled_match[0][6]),
+                "overall_nlp_error": float(scaled_unscaled_match[0][8]),
+            }
+            unscaled = {
+                "incumbent_objective": float(scaled_unscaled_match[0][1]),
+                "dual_infeasibility": float(scaled_unscaled_match[0][3]),
+                "constraint_violation": float(scaled_unscaled_match[0][5]),
+                "complementarity": float(scaled_unscaled_match[0][7]),
+                "overall_nlp_error": float(scaled_unscaled_match[0][9]),
+            }
```
jsiirola, line 641: Could we use the following:
```python
fields = ["incumbent_objective",  "dual_infeasibility", "constraint_violation", "complementarity_error", "overall_nlp_error"]
scaled = {k: float(v) for k, v in zip(fields, scaled_unscaled_match[0][0:10:2])}
unscaled = {k: float(v) for k, v in zip(fields, scaled_unscaled_match[0][1:10:2])}
```

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -499,46 +535,136 @@ def solve(self, model, **kwds):
...
+        )
+        if cpu_time and len(cpu_time) >= 2:
+            parsed_data['cpu_seconds'] = {
+                'nofunc_time': float(cpu_time[0]),
+                'func_time': float(cpu_time[1]),
+            }
+        elif ipopt_seconds_match:
+            parsed_data['cpu_seconds'] = {
+                'total_time': float(ipopt_seconds_match.group(1))
+            }
```
jsiirola, line 661: Would the following be easier:
```python
parsed_data['cpu_seconds'] = {
    k.strip(): float(v) for k, v in 
    re.findall(r'Total(?: CPU)? sec(?:ond)?s in ([^=]+)=\s*([0-9.]+)')
}
```

pyomo/contrib/solver/tests/solvers/test_ipopt.py
```diff
@@ -162,6 +152,97 @@ def test_version_cache(self):
         self.assertIsNone(opt._version_cache[0])
         self.assertIsNone(opt._version_cache[1])
 
+    def test_parse_output(self):
```
jsiirola, line 155: Should there also be a test parsing an example of the "new-style" timing output?

jsiirola (commented):
Lots of parsing questions

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -431,16 +448,34 @@ def solve(self, model, **kwds):
...
+                            results.timing_info.nlp_function_evaluations = (
+                                parsed_output_data['cpu_seconds']['func_time']
+                            )
+                        parsed_output_data.pop('cpu_seconds')
+                        results.extra_info = parsed_output_data
+                        # Set iteration_log visibility to ADVANCED_OPTION because it's
+                        # a lot to print out with `display`
+                        results.extra_info.get("iteration_log")._visibility = (
+                            ADVANCED_OPTION
+                        )
```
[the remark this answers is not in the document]
mrmundt: Because it's implicitly declared. Is there a different way to do it in that scenario?

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -499,46 +534,136 @@ def solve(self, model, **kwds):
...
+        # Stop parsing if there is nothing to parse
+        if not output:
+            logger.log(
+                logging.WARNING,
+                "Returned output from ipopt was empty. Cannot parse for additional data.",
+            )
+            return parsed_data
+
+        # Extract number of iterations
+        iter_match = re.search(r'Number of Iterations\.\.\.\.:\s+(\d+)', output)
```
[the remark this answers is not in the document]
mrmundt: Probably not. I hate regex, as you know.

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -499,46 +535,136 @@ def solve(self, model, **kwds):
...
+                    # Extract restoration flag from 'iter'
+                    iter_val = iter_data["iter"]
+                    iter_match = re.match(r"(\d+)(r?)", iter_val)
+                    if iter_match:
+                        iter_data["restoration"] = bool(
+                            iter_match.group(2)
+                        )  # True if 'r' is present
+                    else:
+                        iter_data["restoration"] = False
+                    iter_data.pop('iter')
```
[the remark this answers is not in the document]
mrmundt: I added an assertion that ensures that the iteration number matches the number of parsed iterations. Will be in the next commit.

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -499,46 +535,136 @@ def solve(self, model, **kwds):
...
+                    # Separate alpha_pr into numeric part and optional tag
+                    alpha_pr_val = iter_data["alpha_pr"]
+                    match = re.match(r"([0-9.eE+-]+)([a-zA-Z]?)", alpha_pr_val)
+                    if match:
+                        iter_data["alpha_pr"] = match.group(1)
+                        iter_data["step_acceptance"] = (
+                            match.group(2) if match.group(2) else None
+                        )
+                    else:
+                        iter_data["step_acceptance"] = None
```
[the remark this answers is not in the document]
mrmundt: I actually want to modify this anyway. Looking back at it, that set of acceptable values is _way_ too big. It's only supposed to be fFhHkKnNRwstTr.

pyomo/contrib/solver/solvers/ipopt.py
mrmundt: It would make `cpu_seconds` look like this (ipopt 3.13):
```
{'IPOPT (w/o function evaluations)': 0.001, 'NLP function evaluations': 0.0}
```
Which is fine but then I have to do more logic in the actual `solve` to convert from this dict to the timer on the `Results` object. I'd prefer to keep it as it is because it's more clear.

pyomo/contrib/solver/tests/solvers/test_ipopt.py
mrmundt: Thanks for the suggestion - I actually caught a bug because of this. It turns out that the old-style and new-style tables aren't the same either. Wheeeeee.

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -380,23 +393,27 @@ def solve(self, model, **kwds):
...
+                        process = subprocess.run(
+                            cmd,
+                            timeout=timeout,
+                            env=env,
+                            universal_newlines=True,
+                            stdout=t.STDOUT,
+                            stderr=t.STDERR,
+                            check=False,
+                        )
+                        timer.stop('subprocess')
```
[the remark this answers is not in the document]
jsiirola: It's not a question of overhead: it's to ensure that the `subprocess` timer that is getting started is guaranteed to be stopped (even in the event of an exception).

pyomo/contrib/solver/solvers/ipopt.py
jsiirola: FWIW, as someone who has worked with Ipopt, I prefer `IPOPT (w/o function evaluations)` (et al.):  It has the advantages of being explanatory and familiar (we are not injecting our own terminology unnecessarily).

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -499,46 +539,140 @@ def solve(self, model, **kwds):
...
+            fields = [
+                "incumbent_objective",
+                "dual_infeasibility",
+                "constraint_violation",
+                "complementarity_error",
+                "overall_nlp_error",
+            ]
+            scaled = {
+                k: float(v)
+                for k, v in zip(fields, scaled_unscaled_match[0][0:10:2])
```
jsiirola, line 642: The slice upper bound is now incorrect: it needs to be `[0:12:2]` (or just `[0::2]`)

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -499,46 +539,140 @@ def solve(self, model, **kwds):
...
+                if v
+            }
+            unscaled = {
+                k: float(v)
+                for k, v in zip(fields, scaled_unscaled_match[0][1:10:2])
```
jsiirola, line 647: The slice upper bound is now incorrect: it needs to be `[1:12:2]` (or just `[1::2]`)

pyomo/contrib/solver/solvers/ipopt.py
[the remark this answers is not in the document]
jsiirola: We should probably make a public API for changing the visibility, then.

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -499,46 +539,140 @@ def solve(self, model, **kwds):
...
+                                    f"float:\n\t{sys.exc_info()[1]}\n\t{line}"
+                                )
+
+                    all_iterations.append(iter_data)
+                    iterations += 1
+
+            parsed_data['iteration_log'] = all_iterations
+
+        # Extract scaled and unscaled table
+        scaled_unscaled_match = re.findall(
```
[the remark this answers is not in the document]
mrmundt: `match` only checks to make sure that the beginning of the string matches, so I think `findall` or `search` are better options.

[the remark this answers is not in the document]
jsiirola: Right.  `search` then.

jsiirola (approved):

PR #3586, opened 2025-04-29
Rewrite `core.relax_integer_vars` transformation

emma58:
## Summary/Motivation:

This rewrites the `core.relax_integer_vars` transformation, modernizing it to our more recent transformation conventions. In particular, it:
- Deprecates the `undo` argument in favor of implementing `reverse` according to the design we introduced in #2809.
- Adds a `targets` argument
- Optionally finds Vars from active expressions rather than using `component_data_objects(Var)`

- Rewrites `core.relax_integer_vars` transformation
- Adds tests!
- Adds `VarCollector` enum to distinguish getting Vars from active expressions vs. using `component_data_objects(Var)`. Naming suggestions are welcome here...

### Legal Acknowledgement


pyomo/core/plugins/transform/discrete_vars.py
```diff
@@ -25,56 +41,170 @@
...
+            entire model is transformed. Note that if the transformation is done
+            out of place, the list of targets should be attached to the model before
+            it is cloned, and the list will specify the targets on the cloned
+            instance.""",
+        ),
+    )
+    CONFIG.declare(
+        'reverse',
+        ConfigValue(
+            default=None,
```
mrmundt, line 62: Is it possible to have a `domain` for this one?

pyomo/core/plugins/transform/discrete_vars.py
```diff
@@ -25,56 +41,170 @@
...
+        ConfigValue(
+            default=VarCollector.FromVarComponents,
+            domain=In(VarCollector),
+            description="The method for collection the Vars to relax. If "
+            "VarCollector.FromVarComponents (default), any Var component on "
+            "the active tree will be relaxed.",
+            doc="""
+            This specifies the method for collecting the Var components to relax.
+            The default, VarCollector.FromVarComponents, assumes that all relevant
+            Vars are on the active tree passed to the transformation. If this is
```
mrmundt, line 88: Something with this grammar is weird. I'm not sure what needs to be fixed, but it's not quite right.

pyomo/core/plugins/transform/discrete_vars.py
```diff
@@ -25,56 +41,170 @@
...
+            cases where some Vars may not be in the active tree, specify
+            VarCollector.FromExpressions to relax all Vars that appear in expressions
+            in the active tree.
+            """,
+        ),
+    )
+    CONFIG.declare(
+        'transform_deactivated_blocks',
+        ConfigValue(
+            default=True,
```
mrmundt, line 99: Potentially opening a can of worms... _Should_ it be true by default?

pyomo/core/plugins/transform/discrete_vars.py
emma58: Umm, yeah, I think so. It should either be `None` or a `ReverseTransformationToken`, so I can check that. Actually, we may even be able to check it's a *valid* token in the domain, right @jsiirola?

pyomo/core/plugins/transform/discrete_vars.py
emma58: I think absolutely not. This seems like very un-Pyomo behavior to me, given that the usual interpretation of deactivation is the user being allowed to hide stuff from us (and the solver). But this maintains backwards compatibility for now...

emma58: I deprecated it, wheee!

emma58:
This is ready for review again, @mrmundt  and @jsiirola

mrmundt (approved):

pyomo/core/plugins/transform/discrete_vars.py
```diff
@@ -25,56 +41,181 @@
...
+        'reverse',
+        ConfigValue(
+            default=None,
+            domain=IsInstance(ReverseTransformationToken),
+            description="The token returned by a (forward) call to this "
+            "transformation, if you wish to reverse the transformation.",
+            doc="""
+            This argument should be the reverse transformation token
+            returned by a previous call to this transformation to transform
+            fixed disjunctive state in the given model.
```
jsiirola, line 69: copy/paste error: `disjunctive` -> `discrete`

pyomo/core/plugins/transform/discrete_vars.py
```diff
@@ -25,56 +41,181 @@
...
+    )
+
     def __init__(self):
-        super(RelaxIntegerVars, self).__init__()
+        super().__init__()
 
     def _apply_to(self, model, **kwds):
-        options = kwds.pop('options', {})
-        if kwds.get('undo', options.get('undo', False)):
+        if not model.ctype in (Block, Disjunct):
```
jsiirola, line 119: This test differs slightly from the error message.  Do you mean `if model.ctype not in SubclassOf(Block):`?

pyomo/core/plugins/transform/discrete_vars.py
```diff
@@ -25,56 +41,181 @@
...
+            if not b.active:
+                if config.transform_deactivated_blocks:
+                    deprecation_warning(
+                        "The `transform_deactivated_blocks` arguments is deprecated. "
+                        "Either specify deactivated Blocks as targets to activate them "
+                        "if transforming them is the desired behavior.",
+                        version='6.9.3.dev0',
+                    )
+                else:
+                    continue
```
jsiirola, line 194: Some questions:
1. Why descend only over `Block` when you accept `Block` and `Disjunct` as targets?  Should this be `component_data_objects(SubclassOf(Block), ...`?
2. Why test for `active` inside the loop?  Wouldn't it be easier to `active = None if config.transform_deactivated_blocks else True`?  Is the idea to only issue the deprecation warning if model actually had a deactivated block?

jsiirola (commented):
Accidentally hit merge when I meant to update the branch from main.  This is fine, with three small comments (none of which would have prevented merging).

pyomo/core/plugins/transform/discrete_vars.py
emma58: Yes, all of this is to make sure we only issue the deprecation warning if the model actually has a deactivated block that we're transforming. When we remove the argument, all this complexity can go away and we can just do what you suggest above. I just didn't want to be loud in all the cases where this is fine because (annoyingly) the default value for `transform_deactivated_blocks` is `True`. So without this mess I would just be raising deprecation warnings when the user didn't actually *do* anything--they probably don't know the option exists.

PR #3592, opened 2025-05-02
Manually install setuptools on conda GHA jobs

mrmundt:
## Summary/Motivation:
For some reason, `setuptools` is no longer available in the base conda environment for Python 3.13 (causing an early failure). This adds manual installation to all conda jobs because why not.

- Add `conda install setuptools` before all other conda installations

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/test_branches.yml
+++ b/.github/workflows/test_branches.yml
@@ -337,6 +337,9 @@ jobs:
         conda list --show-channel-urls
         which python
         python --version
+        # As of May 2025, python 3.13/conda doesn't default have setuptools
+        # Manually install to bypass this issue
+        conda install setuptools
         # Note: some pypi packages are not available through conda
         PYOMO_DEPENDENCIES=`python setup.py dependencies \
             --extras "$EXTRAS" | tail -1`
--- a/.github/workflows/test_branches.yml
+++ b/.github/workflows/test_branches.yml
@@ -337,7 +337,7 @@ jobs:
         conda list --show-channel-urls
         which python
         python --version
-        # As of May 2025, python 3.13/conda doesn't default have setuptools
+        # As of May 2025, python 3.13/conda doesn't have setuptools by default
         # Manually install to bypass this issue
         conda install setuptools
         # Note: some pypi packages are not available through conda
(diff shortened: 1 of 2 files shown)


.github/workflows/test_branches.yml
```diff
@@ -337,6 +337,9 @@ jobs:
         conda list --show-channel-urls
         which python
         python --version
+        # As of May 2025, python 3.13/conda doesn't default have setuptools
```
blnicho, line 340:
```suggestion
        # As of May 2025, python 3.13/conda doesn't have setuptools by default
```

.github/workflows/test_pr_and_main.yml
```diff
@@ -389,6 +389,9 @@ jobs:
         conda list --show-channel-urls
         which python
         python --version
+        # As of May 2025, python 3.13/conda doesn't default have setuptools
```
blnicho, line 392:
```suggestion
        # As of May 2025, python 3.13/conda doesn't have setuptools by default
```

blnicho (approved):
Nitpicking grammar but otherwise this is fine.

.github/workflows/test_branches.yml
```diff
@@ -337,6 +337,9 @@ jobs:
...
+        # Manually install to bypass this issue
+        conda install setuptools
```
jsiirola, line 342: How does this interact with the Windows-specific setuptools install below?  Can we combine them into one?  Does setuptools need to be installed as a separate `conda install`, or can it be lumped into the rest of the dependencies (there was evidence in the past that it was *way* faster to install everything with a single `conda install` command where possible).

mrmundt: The problem is that we can't run line 344 (`python setup.py dependencies`) without setuptools installed, so this step needs to happen before anything else is able to run.

mrmundt: As for how it interacts with the Windows part - both lines are still necessary because we install whatever setuptools first so we can actually parse the dependencies, and then later we say, "Okay, if needed, please change to a different version."

mrmundt:
Appears to no longer be necessary!

PR #3593, opened 2025-05-03
CSpline - Fix segment function and add utils to constrain parameter calculations

eslickj:
## Fixes

There was a bug in the segment function in the parameter class. The function is just used for testing and plotting cubic splines, so it is not likely to have affected anything.  It's fixed and tests are added that would catch similar future issues.

## Summary/Motivation:

This fixes a bug and adds parameter calculation options that allow you to impose certain properties on the cubic spline.

- Fix bug in segments method in CsplineParameters class (use for plotting and testing)
- Add option to extend the curve to extrapolate linearly from the endpoints using endpoint slope
- Add function to add constraints to parameter calculation problem to encourage the cspline to be increasing
- Add function to add constraints to parameter calculation problem to encourage the cspline to be decreasing
- Add function to add constraints to parameter calculation problem to require the function to be convex
- Add function to add constraints to parameter calculation problem to require the function to be concave

### Legal Acknowledgement


pyomo/contrib/cspline_external/cspline_parameters.py
```diff
@@ -180,11 +206,18 @@ def segment(self, x):
...
-        return s
+        # If x is before the first knot use the first segment
+        if len(s) > 1:
+            s[s <= 0] = 1
+            s[s >= len(self.knots)] = len(self.knots) - 1
+        else:
+            if s <= 0:
+                return 0
+            if s >= len(self.knots):
+                # if the not extrapolates right it is 1 more than number of
```
mrmundt, line 217:
```suggestion
                # if the knot extrapolates right it is 1 more than number of
```
Is this supposed to be `knot`?

pyomo/contrib/cspline_external/cspline_parameters.py
```diff
@@ -198,6 +231,18 @@ def f(self, x):
...
+
+        Args:
+            x: location, numpy array float
+
+        Returns:
+            f(x) numpy array if x is numpy array or float
+        """
+        s = self.segment(x)
+        return self.a2[s] + 2 * self.a3[s] * x + 3 * self.a4[s] * x**2
+
```
mrmundt, line 245: `def f` and `def fx` both "get f(x)" but f(x) is different in both scenarios. That strikes me as confusing.

eslickj: fx is df/dx, so dfdx is probably a better name, and now I see I forgot to edit the doc string when I copied it.

pyomo/contrib/cspline_external/cspline_parameters.py
```diff
@@ -313,3 +359,71 @@ def yxx_endpoint_eqn(blk, s):
...
+    resulting curve or pair it with convex or concave constraints.
+    """
+
+    @m.Constraint(m.knt_idx)
+    def yx_ineq(blk, k):
+        if k >= len(m.knt_idx):
+            s = k - 1
+        else:
+            s = k
+        return _fx_cubic(m.x[k], m.alpha, s) <= -tol
```
jsiirola, line 380: It would be good for these functions to be declared at the module scope so that they would be picklable.  Given the use of the `tol` argument, maybe they should be functors:
```python
class _yx_ineq(object):
    def __init__(self, tol):
        self.tol = tol

    def __call__(self, m, k):
        if k >= len(m.knt_idx):
            s = k - 1
        else:
            s = k
        return _fx_cubic(m.x[k], m.alpha, s) <= -self.tol

def add_decreasing_constraints(m, tol=0):
    m.yx_ineq = Constraint(m.knt_idx, rule=_yx_ineq(tol))
```

pyomo/contrib/cspline_external/cspline_parameters.py
```diff
@@ -313,3 +359,71 @@ def yxx_endpoint_eqn(blk, s):
         else:
             j = s
         return _fxx_cubic(m.x[j], m.alpha, s) == 0
+
+
+def add_decreasing_constraints(m, tol=0):
```
jsiirola, line 364: These functions assume that `m` has a set `knt_idx`, and that they can create components with specific hardcoded names.  This feels like these should be methods on a custom Block class that disallows users to create components so you don't have to worry about conflicts and *know* that the required attributes are present.

eslickj: Ok, restructured it this way.

mrmundt:
FYI that there is a pretty silly false positive in the typos check right now that I am hoping they'll do a hot patch to fix. I'm monitoring the PR that their team opened (https://github.com/crate-ci/typos/pull/1311)

pyomo/contrib/cspline_external/cspline_parameters.py
```diff
@@ -11,6 +11,8 @@
 
 from pyomo.common.dependencies import numpy as np
 import pyomo.environ as pyo
+from pyomo.core.base.block import BlockData, declare_custom_block
+import idaes
```
mrmundt: We do not allow `idaes` to be imported in Pyomo because `idaes` depends on Pyomo. There are lots of potentially bad circular import issues that can occur.

mrmundt (changes requested):

pyomo/contrib/cspline_external/cspline_parameters.py
eslickj: Oops, put that there to set the solver path.  Meant to delete it.

mrmundt (approved):
The new class `CubicParametersModelData` probably should have some more unit tests, but I won't let that hold up this PR.

PR #3599, opened 2025-05-07
Suppress warmstart message from Xpress

bknueven:
## Summary/Motivation:
On `main`, when using `warmstart = True` with Xpress, you will always see the below message printed to the screen:
```
User solution (_) stored.
```
which is emitted by the `addmipsol` method of `xpress.problem`. This can be troublesome because (1) the message is printed to the screen unconditionally and (2) does not end up in a specified `logfile`.

- Delay calling `XpressDirect._warm_start` until the model is about to be solved, when the `logfile` and / or `tee` attributes have been properly configured.

### Legal Acknowledgement

### DIFF
--- a/pyomo/solvers/plugins/solvers/xpress_direct.py
+++ b/pyomo/solvers/plugins/solvers/xpress_direct.py
@@ -239,6 +239,10 @@ class XpressDirect(DirectSolver):
         # version of xpress is supported (and stored as a class attribute)
         del self._version
 
+        # xpress will apply the warmstart itself instead of
+        # DirectOrPersistentSolver._presolve
+        self._apply_warmstart = False
+
     def available(self, exception_flag=True):
         """True if the solver is available."""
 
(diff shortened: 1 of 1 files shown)


pyomo/solvers/plugins/solvers/xpress_direct.py
```diff
@@ -261,6 +265,13 @@ def available(self, exception_flag=True):
         finally:
             xpress.free()
 
+    def _presolve(self, *args, **kwds):
+        # we'll apply the warmstart in _solve_model so the
+        # message "User solution (_) stored" is wrote to the
+        # correct place, i.e., the console or the log or both
+        self._apply_warmstart = kwds.pop("warmstart", False)
+        return super(XpressDirect, self)._presolve(*args, **kwds)
```
mrmundt, line 273: This is Python2 style - you can do this instead:
```suggestion
        return super()._presolve(*args, **kwds)
```

jsiirola (approved):
Tests would be nice, but given that we are moving away from this interface, I don't think they are required.

PR #3601, opened 2025-05-08
Resolve intermittent `capture_output` deadlock 

jsiirola:
## Summary/Motivation:
We are continuing to see intermittent deadlocks or attempts to write to closed file descriptors in the HiGHS tests on main.  I think I have (finally) tracked down the root cause to the use of `capture_input()` in the *Gurobi* interface `__del__` method.  It looks like the GC will (randomly) interrupt a HiGHS test to run the collector (which, if it is collecting a Gurobi interface will call `capture_output`.  If this happens during another test's use of `capture_output.__enter__` or `.__exit__`, then the Gurobi interface's use of capture could happen when the underlying output streams are in a transitory state (and could be closed).

This PR resolves that by:
- rework the Gurobi interface to use it's own private Env (which suppresses the output to stdout and removes the need to use `capture_output` in `__del__`.
- Add thread locks so that one call to `capture_output` will not interrupts another fall to `__enter__ / `__exit__`
- improve some implementations within TeeStream / capture_output
- add additional tests that catch some edge cases identified during development

- (see above)

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/tee.py
+++ b/pyomo/common/tee.py
@@ -240,6 +240,8 @@ class capture_output(object):
 
     """
 
+    startup_shutdown = threading.Lock()
+
     def __init__(self, output=None, capture_fd=False):
         if output is None:
             output = io.StringIO()
@@ -285,6 +287,35 @@ class capture_output(object):
         return FAIL
 
     def __enter__(self):
+        if not capture_output.startup_shutdown.acquire(timeout=_poll_timeout_deadlock):
+            # This situation *shouldn't* happen.  If it does, it is
+            # unlikely that the user can fix it (or even debug it).
+            # Instead they should report it back to us.
+            #
+            # Breadcrumbs:
+            #
+            #   - The last time we hit this [5/2025], it was because we
+            #     were using capture_output in a solver's __del__.  This
+            #     led to the GC deleting the solver while another solver
+            #     was trying to start up / run (so the other solver held
+            #     the lock, but the GC interrupted that thread and
+            #     wouldn't let go).
+            raise DeveloperError("Deadlock starting capture_output")
+        try:
+            return self._enter_impl()
+        finally:
+            capture_output.startup_shutdown.release()
+
+    def __exit__(self, et, ev, tb):
+        if not capture_output.startup_shutdown.acquire(timeout=_poll_timeout_deadlock):
+            # See comments & breadcrumbs in __enter__() above.
+            raise DeveloperError("Deadlock closing capture_output")
+        try:
+            return self._exit_impl(et, ev, tb)
+        finally:
+            capture_output.startup_shutdown.release()
+
+    def _enter_impl(self):
         self.old = (sys.stdout, sys.stderr)
         old_fd = []
         for stream in self.old:
@@ -399,7 +430,7 @@ class capture_output(object):
             buf = buf[0]
         return buf
 
-    def __exit__(self, et, ev, tb):
+    def _exit_impl(self, et, ev, tb):
         # Check that we were nested correctly
         FAIL = []
         if self.tee._stdout is not None and self.tee.STDOUT is not sys.stdout:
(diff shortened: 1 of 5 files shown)


pyomo/common/tee.py
```diff
@@ -667,15 +705,18 @@ def close(self, in_exception=False):
...
                 th.join(_poll)
+                _timeout += _poll
             self._threads[:] = [th for th in self._threads if th.is_alive()]
             if not self._threads:
                 break
-            _poll *= 2
-            if _poll_timeout <= _poll < 2 * _poll_timeout:
+            if _poll < _poll_timeout:
+                _poll *= 2.0
+            if _poll_timeout * 0.5 <= _poll < _poll_timeout:
```
mrmundt, line 719: Something I have wondered for a while - why the factors of 2? Is it arbitrarily decided?

pyomo/common/tests/test_tee.py
```diff
@@ -559,6 +579,31 @@ def test_capture_output_stack_error(self):
             sys.stdout, sys.stderr = old
             logging.getLogger('pyomo.common.tee').handlers.clear()
 
+    def test_atomic_deadlock(self):
```
mrmundt, line 582: This is a nifty test, I like it.

pyomo/contrib/solver/solvers/highs.py
```diff
@@ -288,7 +288,7 @@ def set_instance(self, model):
                 f'({self.available()}).'
             )
 
-        with TeeStream(*ostreams) as t, capture_output(t.STDOUT, capture_fd=True):
+        with capture_output(TeeStream(*ostreams), capture_fd=True):
```
mrmundt, line 291: What is the difference between doing this and just doing `with TeeStream(*ostreams) as OUT`? Do we have a preference? (I ask because we don't seem to have decided on a "standard" and we probably should)

jsiirola: `capture_output` wraps the streams it is passed into a `TeeStream` in order to get handles it can give to stdout / stderr.  So the former implementation actually had 2 nested `TeeStream` objects.  If you just pass the `TeeStream` object to `capture_output`, it won't create a new `TeeStream`, so the new implementation only has a single `TeeStream`.

mrmundt: Sorry, let me rephrase. In ipopt currently, we do this:
```
                with TeeStream(*ostreams) as t:
                    timer.start('subprocess')
                    process = subprocess.run(
                        cmd,
                        timeout=timeout,
                        env=env,
                        universal_newlines=True,
                        stdout=t.STDOUT,
                        stderr=t.STDERR,
                        check=False,
                    )
                    timer.stop('subprocess')
                    # This is the stuff we need to parse to get the iterations
                    # and time
                    (iters, ipopt_time_nofunc, ipopt_time_func, ipopt_total_time) = (
                        self._parse_ipopt_output(ostreams[0])
                    )
```

In GUROBI, we do:
```
            with capture_output(TeeStream(*ostreams), capture_fd=False):
                gurobi_model = gurobipy.Model()

                timer.start('transfer_model')
                x = gurobi_model.addMVar(
                    len(repn.columns),
                    lb=lb,
                    ub=ub,
                    obj=repn.c.todense()[0] if repn.c.shape[0] else 0,
                    vtype=vtype,
                )
```

Which one is preferable?

jsiirola: They are both correct for their respective use cases:
- `subprocess.run` needs to be passed 2 TextIO objects, so it is appropriate (and necessary) to pass the raw TextIO objects from `TeeStream`.
- `capture_output` needs a single stream to send all output to (we have no added support for separate our/err targets (so internally it creates a TeeStream so that it can get raw file descriptors if it needs them (for `capture_fd=True`).  Here Passing a TeeStream directly avoids the need to create a second one that wraps the stream from the first.

mrmundt: Got it, thank you!

mrmundt (approved):

mrmundt:
Per @blnicho - we could merge this today if no one has time to do another review. Ergo, risking it for the biscuit and merging.

PR #3603, opened 2025-05-14
Modernize infrastructure to use `pyproject.toml`

mrmundt:
## Summary/Motivation:
There is an impending removal of support for `python setup.py develop` being used in our infrastructure. This PR moves the minimum amount of information possible from `setup.py` to the more modern `pyproject.toml`. I also apparently have some `black` issues but I have no idea where they are coming from.

- Introduce `pyproject.toml`
- Remove `setup.cfg`
- Make appropriate changes to `setup.py` to conform to [PEP 517](https://peps.python.org/pep-0517/)
- Update docs
- Update all workflows (tests and release) 
- Apply `black` for unknown reasons

**NOTE**: A passing Release distribution job is here: https://github.com/mrmundt/pyomo/actions/runs/15025422604

### Legal Acknowledgement


.github/workflows/test_branches.yml
```diff
@@ -654,14 +653,19 @@ jobs:
 
     - name: Install Pyomo
       run: |
+        # Turn on cython through env vars
+        if [[ "${{ matrix.other }}" == *cython* ]]; then
+            echo "Cython requested; turning on cythonization options in environment"
+            export PYOMO_SETUP_ARGS=--with-cython
+        fi
```
jsiirola, line 660: Why not just leave the `setup_options` matrix variable and then (unconditionally) add it to the `PYOMO_SETUP_ARGS` environment variable here?

.github/workflows/test_branches.yml
```diff
@@ -868,13 +872,18 @@ jobs:
...
```
jsiirola, line 879: Same question as above...

doc/OnlineDocs/getting_started/installation.rst
```diff
@@ -86,16 +90,30 @@ initialized.
...
+   pip install pyomo
+
+On Windows:
+
+::
+
+   # Via command prompt
+   set PYOMO_SETUP_ARGS=--with-cython
+   # Via powershell
+   $env:PYOMO_SETUP_ARGS="--with-cython"
```
jsiirola, line 108: Does it matter that you have he call for `pip install pyomo` in Linux, but not here?

jsiirola (commented):
Looks interesting.  A couple questions, though...

.github/workflows/test_branches.yml
mrmundt: Because the only reason we use `setup_options` is for cythonization. I don't really care either way, but it seemed redundant, so tell me what you'd prefer.

.github/workflows/test_branches.yml
mrmundt: Because the only reason we use `setup_options` is for cythonization. I don't really care either way, but it seemed redundant, so tell me what you'd prefer.

doc/OnlineDocs/getting_started/installation.rst
mrmundt: Whoops. That was a mistake. Fixed!

.github/workflows/test_branches.yml
jsiirola: In general, I would prefer to *not* key job behavior off the job name (this brings back bad memories of codes where the input deck filename could set some of the program arguments).  It seems fragile and not extensible.  (so yeah - if you are asking about my *preference*, it would be to have an explicit argument in the matrix definition.  :) )

pyproject.toml
```diff
@@ -0,0 +1,88 @@
...
+build-backend = "setuptools.build_meta"
+
+# project metadata
+
+[project]
+name = "pyomo"
+description = "The Pyomo optimization modeling framework"
+readme = "README.md"
+license = { file = "LICENSE.md" }
+authors = [{name = "Pyomo Development Team", email = "pyomo-developers@googlegroups.com"}]
```
blnicho, line 19:
```suggestion
authors = [{name = "Pyomo Development Team", email = "wg-pyomo@sandia.gov"}]
```
@jsiirola @mrmundt should we update the email address to match what we have in the README to request dev call info?

blnicho (approved):
I have one minor question but otherwise this looks great. I also checked the RTD build and verified that this PR picks up the pyproject.toml file.

pyproject.toml
mrmundt: @jsiirola @blnicho - I say "yes". But if so, we should check anywhere else that that one is used (e.g., `RELEASE.md`)

jsiirola: My only concern is that only the SNL developers receive mail sent to `wg-pyomo`.  While that does hit the bulk of the current "core dev team", it does not get to everyone (and I personally don't want to give people the untrue impression that you have to be at SNL to be a developer).  `pyomo-developers` is more public and can reach anyone who has developer interest.  

The justification for `wg-pyomo` for the dev call info is that the meeting is hosted by SNL, so it makes sense to request the info from an SNL address.

mrmundt: I accept that counter point - optics are important!

PR #3614, opened 2025-06-02
Add an option to skip checking the bounds of fixed variables

michaelbynum:
## Summary/Motivation:
Currently, the NL writer raises an exception if a variable is fixed to a value outside of its bounds. This PR adds an option to skip that check and error.

### Legal Acknowledgement

### DIFF
--- a/pyomo/repn/ampl.py
+++ b/pyomo/repn/ampl.py
@@ -1099,6 +1099,7 @@ class AMPLRepnVisitor(StreamBasedExpressionVisitor):
         symbolic_solver_labels,
         use_named_exprs,
         sorter,
+        check_fixed_variable_bounds=True,
     ):
         super().__init__()
         self.subexpression_cache = subexpression_cache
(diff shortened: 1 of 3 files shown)


jsiirola:
Design question: is it better to add a new option that is specific to this, or just expose the `TOL` to the user?

michaelbynum:
So just making TOL `inf` would do the same thing... Good question. My only concern with that is that TOL could be used for other things as well.

jsiirola:
The `TOL` in `ampl.py` (the walker) is *only* used for checking the feasibility of fixed variables - there is a separate `TOL` in `nl_writer.py` that is used for checking trivial constraints (when they are not just written to the NL file).  If we expose that to the user, we probably want to make it more explicit/descriptive (`trivial_constraint_abs_feasibility_tolerance`?).  [Actually, we probably want to make that more descriptive even if we *don't* expose it to the user!]

FWIW, the `TOL` in `ampl.py` was added for PyROS (@shermanjasonaf ), because some of the subproblems generated through the algorithm would be slightly infeasible due to tolerances in the solver used to generate them.  I believe that PyROS relaxes that tolerance (temporarily) within its algorithm.

michaelbynum:
I will modify this PR to expose that tolerance through the NL writer config.

Issue #323, opened 2018-01-25
Add sinc(x) function


whart222:
sinc(x) = 1 if x == 0 else sin(x)/x

While NL files do not support this intrinsic function, it can be supported by adding a variable z and a constraint:

  x*z - sin(x) == 0

whart222:
This is the "sampling function" that arises frequently in signal processing applications.  This is perhaps illustrative of a broad class of nonlinear functions that could be added and implicitly represented through transformations of Pyomo expressions.

jsiirola:
Why do you want to add this?  First, your example is not the sinc function, as z becomes free at x == 0.  The correct implementation is probably 
```
def sinc(x):
    return Expr_if(IF=x==0, THEN=1, ELSE=sin(x)/x)
```
The problem is that even that may have numerical issues (it is in C1, but not C2: if the ASL evaluates it at 0, the second derivative will be 0).  Maybe there is a better value for the `THEN` clause that would move it into C2?

Making things more confusing, there are two definitions for sinc (the other being `sin(pi*x)/(pi*x)`).  Which do you want to support?

The bigger question is if we want to start implementing a nearly infinite set of functions (much like the GSL?).  I suppose we could, but that is definitely not something we would want imported by `pyomo.environ`.  Maybe you are arguing for a `pyomo.math`?

carldlaird:
There are several issues here. the x*z-sin(x) transformation may cause numerical problems for the NLP solver (e.g., at a common initialization of x=0, z=0, but I haven't thought about this too much).

The Expr_if example above is heading in the right direction, but may have trouble because it is doing a exact comparison of a floating point, and sin(x)/x might be poorly behaved numerically for values of x that are very close to zero, but not zero. (Again, we would need to check into it.)

I agree that we should look at a package to add these kinds of functions as needed. However, for them to be useful, we need to ensure the appropriate numerical properties over the valid range.

mrmundt:
This functionality could likely be implemented using an `ExternalFunction` or black-box function. Either way, the progression of math libraries in Python and the desire to incur less technical debt has led us to decide to close this issue.

PR #3622, opened 2025-06-09
Add sgnsqr external function f(X) = sgn(x)*x^2

eslickj:
## Summary/Motivation:

This adds a signed square function to the contributed external function demonstration.  

### Here is an example of where this can be useful.

Say you have a pressure-flow relationship that says the mass flow is proportional to the square root of pressure drop, where pa is inlet pressure and pb is outlet pressure.  Like this: mdot^2 = k(pb - pa).

The direction of flow is forward if the pressure drop is positive or backward if the pressure drop is negative.  There are both positive and negative solutions to the equation.  If we use the proposed sgnsqr function, there is now only one solution to the equation, and the sign of flow matches the sign of pressure drop. 

The sign squared function is still smooth and replacing square with sgnsqr still gives us solutions to the original equation.  It just limits us to one correct solution instead of two possibilities.  I expect there could be many situations where it would be useful to limit the solutions of y = x^2 to a single solution.   

- add sgnsqr external function to the ampl_external_demo library.
-

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/ampl_function_demo/build.py
+++ b/pyomo/contrib/ampl_function_demo/build.py
@@ -15,10 +15,10 @@ from pyomo.common.cmake_builder import build_cmake_project
 
 def build_ampl_function_demo(user_args=[], parallel=None):
     return build_cmake_project(
-        targets=['src'],
-        package_name='asl_external_demo',
-        description='AMPL External function demo library',
-        user_args=['-DBUILD_AMPLASL_IF_NEEDED=ON'] + user_args,
+        targets=["src"],
+        package_name="asl_external_demo",
+        description="AMPL External function demo library",
+        user_args=["-DBUILD_AMPLASL_IF_NEEDED=ON"] + user_args,
         parallel=parallel,
     )
 
--- a/pyomo/contrib/ampl_function_demo/plugins.py
+++ b/pyomo/contrib/ampl_function_demo/plugins.py
@@ -14,4 +14,4 @@ from pyomo.contrib.ampl_function_demo.build import AMPLFunctionDemoBuilder
 
 
 def load():
-    ExtensionBuilderFactory.register('ampl_function_demo')(AMPLFunctionDemoBuilder)
+    ExtensionBuilderFactory.register("ampl_function_demo")(AMPLFunctionDemoBuilder)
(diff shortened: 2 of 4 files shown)


jsiirola:
Overall, I don't see a huge problem with supporting a library of useful external functions; however, some concerns with this implementation:
- the `ampl_function_demo` module is really meant to be lightweight documentation for how to write and build AMPL external functions that are compatible with the Pyomo build system.  I think if we are going to start supporting "real" external functions, then that should be in a separate module.
- this implementation is in C1 but not in C2.  That could cause some solvers (e.g., `ipopt`) that assume continuous second derivatives to have trouble.  @carldlaird should weigh in here.
- If we decide to go down the path of creating a library of "useful" external functions, we should consider including others (e.g., #323).
- I think the package would need appropriate user documentation (at the very least to include caveats so that users understand the limitations of the function implementations).

eslickj:
@jsiirola, I agree that if we decide that there are useful external functions we should add a new library.  I was thinking that but then decided to see if I could get away with a short cut.  The fact that the function isn't second order smooth is bothersome, but then again so is getting a solution with the wrong sign.  I'm still not sure if this is a good idea or not.  I know we could also do things with bounds and constraints, but I do a fair number of straight forward square simulations where bounds and constraints are harder to manage.

rschwarz:
I have a general question about this functionality: Would this even work with IPOPT currently, which is using the NL interface (via files) to communicate the problem?

jsiirola:
> I have a general question about this functionality: Would this even work with IPOPT currently, which is using the NL interface (via files) to communicate the problem?

Yes.  In fact, that is it's primary use case.  It should also work with any other local nonlinear solver that accepts NL file inputs.

rschwarz:
Thanks, I was unaware of the possibility and we have that same expression in our models, actually.

jsiirola:
@eslickj, we discussed this at yesterday's dev call.  The consensus was that we should move this to a separate contrib package.  The package could host multiple ASL external libraries, and this should go into one that explicitly calls out that it is not a function in C2.  We did not pick a name (although several were tossed around, like `c1_functions` or `numerically_questionable_functions`).

eslickj:
@jsiirola. thanks.  I'll close this and open a new PR soon.

PR #3625, opened 2025-06-13
Temporarily pinning the scikit-learn and Python 3.13 versions

blnicho:
## Summary/Motivation:
This sets an upper limit on the scikit-learn version to work around testing failures in `contrib.piecewise` stemming from the unmaintained `linear-tree` package. @emma58 is working on a better solution that will replace the `linear-tree` dependency with something that is being maintained. This PR is a temporary fix until then.

Upon opening this PR we also discovered a new testing issue with cpython 3.13.4. This new Python version introduces a slight change in the formatting of some logging messages which we've tracked to this resolved issue: https://github.com/python/cpython/issues/91555. We need feedback from @jsiirola on how to proceed.

- Temporarily bound scikit-learn version 
- Temporarily pin Python 3.13 version used for testing

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/test_branches.yml
+++ b/.github/workflows/test_branches.yml
@@ -373,7 +373,7 @@ jobs:
         fi
         # Note: this will fail the build if any installation fails (or
         # possibly if it outputs messages to stderr)
-        conda install --update-deps -q -y $CONDA_DEPENDENCIES
+        conda install --update-deps -q -y python="${{matrix.python}}" $CONDA_DEPENDENCIES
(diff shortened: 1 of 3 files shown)


jsiirola (approved):

jsiirola:
I think it is fine to merge this as-is to get tests passing and to free up the PR backlog.  I will start looking into what changed in Python 3.13.4, but that will take longer and I don't think we should hold up everything else waiting for it.

PR #3629, opened 2025-06-16
Modified _parse_gdx_results in GAMS.py to replace _parse_special_value

### LINKED ISSUE #3624: Updating the Import Statement to Use GAMS new API
## Summary

Updating the `attempt_import` statement in `pyomo\solvers\plugins\solvers\GAMS.py` from `gdxcc` to `gams.core.gdx` to use GAMS new API.

### Rationale

The `attempt_import` statement returns `gdxcc_available` as a boolean value that parses the solver result into a `.gdx` file if it is true or a `.dat` file if it is false. The data file format may affect parsing performance post-solved.

### Description

```
gdxcc, gdxcc_available = attempt_import('gdxcc') # old interface
gdxcc, gdxcc_available = attempt_import('gams.core.gdx') # new interface
```

### Additional information

We are currently comparing the data parsing time after solving between the `.dat` and `.gdx` data file formats. The model used for the test and the `.pkl` time data will be added below.

AnhTran01:
## Summary/Motivation:
When parsing the results of a model after it has been solved, the level and dual value are obtained through a series of `if` statements in `_parse_special_values` that may cause slowdowns. This PR added GAMS existing functions to handle data parser for these special values in `_parse_gdx_results`.

- Replaced `_parse_special_values` with GAMS special value parser in `_parse_gdx_results`.

### Legal Acknowledgement

### DIFF
--- a/doc/OnlineDocs/explanation/solvers/pyros.rst
+++ b/doc/OnlineDocs/explanation/solvers/pyros.rst
@@ -963,10 +963,10 @@ Observe that the log contains the following information:
    :linenos:
 
    ==============================================================================
-   PyROS: The Pyomo Robust Optimization Solver, v1.3.6.
-          Pyomo version: 6.9.2
-          Commit hash: 41cd797e0
-          Invoked at UTC 2025-03-13T16:20:31.105320+00:00
+   PyROS: The Pyomo Robust Optimization Solver, v1.3.8.
+          Pyomo version: 6.9.3dev0
+          Commit hash: unknown
+          Invoked at UTC 2025-05-05T00:00:00.000000+00:00
 
    Developed by: Natalie M. Isenberg (1), Jason A. F. Sherman (1),
                  John D. Siirola (2), Chrysanthos E. Gounaris (1)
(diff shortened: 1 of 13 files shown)


AnhTran01:
@boxblox
@abhosekar

boxblox:
this is a small thing, but we should add `gdx.gdxLibraryUnload()` after the `gdxFree(pgdx)` calls in both the results file and the stats file read operations

pyomo/solvers/plugins/solvers/GAMS.py
```diff
@@ -40,8 +40,10 @@
 )
 
 from pyomo.common.dependencies import attempt_import
+import numpy as np
+import struct
 
-gdxcc, gdxcc_available = attempt_import('gdxcc')
+gdxcc, gdxcc_available = attempt_import('gams.core.gdx')
```
jsiirola: I would suggest the following change to support both pre-GAMS-45 and current GAMS APIs:
```python
def _gams_importer():
    try:
        import gams.core.gdx as gdx
        return gdx
    except ImportError:
        try:
            # fall back to the pre-GAMS-45.0 API
            import gdxcc
            return gdxcc
        except:
            # suppress the error from the old API and reraise the current API import error
            pass
        raise

gdxcc, gdxcc_available = attempt_import('gdxcc', importer=_gams_importer) 
```

jsiirola:
It is unfortunate that e1da345 was committed... that introduced a bunch of non-functional changes that are confusing review of this PR.  For the future, we run (and test against) `black -C -S`.  If you want to make all the changes to the string quote characters, that's fine, but it would be better for that to be isolated to a separate PR that ONLY makes that change.

AnhTran01:
This is my bad @jsiirola. In this case, would the best practice is to make a new PR on a clean branch with the changes to the import statement?

jsiirola:
> This is my bad @jsiirola. In this case, would the best practice is to make a new PR on a clean branch with the changes to the import statement?

That is completely up to you.  You can just revert that commit on this branch and it is likely to be OK.  If that still leaves "a mess" then you can decide if it is easier to just recreate a new branch / PR.

PR #3634, opened 2025-06-16
Fix: handle HighsModelStatus.kSolutionLimit like kIterationLimit

### LINKED ISSUE #3632: HiGHS status kSolutionLimit is not handled
## Summary

When solving MIP problems with HiGHS and some active solver limits, eg, `mip_max_nodes=50`, the solver may stop prematurely, ie, before solving the problem to optimality. In that case, the status given by HiGHS is `kSolutionLimit`. But this is not handled in the `if/else` chain here:

https://github.com/Pyomo/pyomo/blob/dc503c2835ea2e41569b1a119caedb248ae2939d/pyomo/contrib/appsi/solvers/highs.py#L657-L690

Hence, the Pyomo user just gets the `result.solver`:

```
- Status: unknown
  Termination condition: unknown
  Termination message: TerminationCondition.unknown
```

Even though, looking at the solver log, we know quite a bit more, eg, that a feasible solution exists:

```
Solving report
  Status            Solution limit reached
  Primal bound      153.831642438
  Dual bound        154.134587306
  Gap               0.197% (tolerance: 0.01%)
  P-D integral      0.020659109043
  Solution status   feasible
                    153.831642438 (objective)
                    0 (bound viol.)
                    0 (int. viol.)
                    0 (row viol.)
  Timing            0.17 (total)
                    0.00 (presolve)
                    0.00 (solve)
                    0.00 (postsolve)
  Max sub-MIP depth 3
  Nodes             1
```

I can provide code to generate tricky Knapsack instances, if needed.

The important bit is to use `options=dict(mip_max_nodes=1)` or similar.

Pyomo version: 6.7.3 (but the relevant code is still the same on `main` branch)
Python version: 3.12.
Operating system: Ubuntu 24.04
Solver (if applicable): HiGHS v1.11.0

### Additional information

I'm happy to have a go at this in a PR, but might not get to it immediately.

rschwarz:
Was previously not handled at all, resulting in termination_condition of unknown, and no further information about whether a solution was found.

kIterationLimit may not be exact, but it's the best fit among the current enum values, I guess.

- add case for `kSolutionLimit`
- add warning for the `else` branch

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/appsi/solvers/highs.py
+++ b/pyomo/contrib/appsi/solvers/highs.py
@@ -684,9 +684,12 @@ class Highs(PersistentBase, PersistentSolver):
             results.termination_condition = TerminationCondition.maxTimeLimit
         elif status == highspy.HighsModelStatus.kIterationLimit:
             results.termination_condition = TerminationCondition.maxIterations
+        elif status == highspy.HighsModelStatus.kSolutionLimit:
+            results.termination_condition = TerminationCondition.maxIterations
         elif status == highspy.HighsModelStatus.kUnknown:
             results.termination_condition = TerminationCondition.unknown
         else:
+            logger.warning(f'Received unhandled {status=} from solver HiGHS.')
             results.termination_condition = TerminationCondition.unknown
 
         timer.start('load solution')


mrmundt:
So I understand - does `kSolutionLimit` exist for all versions of `highspy`, or is it only present in 1.5?

rschwarz:
My understanding is that it was added to HiGHS v1.5, and has remained to the present day.

EDIT: See this HiGHS test which is still on the main branch:

https://github.com/ERGO-Code/HiGHS/blob/364c83a51e44ba6c27def9c8fc1a49b1daf5ad5c/check/TestMipSolver.cpp#L62-L67

mrmundt:
Okay, then this is going to need a bit of work. This change would effectively pin us to a minimum version of highspy, and we try our best to maintain backwards compatibility as much as possible. Additionally, there are actually two different HiGHS interfaces - contrib/appsi and contrib/solver. We have a dev call tomorrow and will talk about it then. Thanks for submitting the PR! I'll let you know soon what we decide.

rschwarz:
If it helps, I could add in an additional version check for HiGHS in the code.

jsiirola:
A version check would be OK, but something like
```python
       elif status == getattr(highspy.HighsModelStatus, "kSolutionLimit", NOTSET):
            results.termination_condition = TerminationCondition.maxIterations
```
might be more performant and simpler.

rschwarz:
> A version check would be OK, but something like
> 
> ```python
>        elif status == getattr(highspy.HighsModelStatus, "kSolutionLimit", NOTSET):
>             results.termination_condition = TerminationCondition.maxIterations
> ```
> 
> might be more performant and simpler.

This looks good to me, but I'm not sure if this might cause some type problems, as `highspy.HighsModelStatus` is actually an `enum` at the C++ level.

Please let me know if which changes I should apply, if any.

PS: I'm now aware of the non-APPSI interface to HiGHS, which is affected by the same problem.

mrmundt:
Okay, we talked about this today, and we think using @jsiirola 's proposed solution above on both files should be good. Please make those changes and we'll happily move forward!

rschwarz:
I see that the newly added lines are not covered by the existing tests. Should I attempt to set up a MIP problem of moderate size where this status would result after a node limit of 1, or would that be overkill?

jsiirola:
That would be fantastic.  We (i.e., @mrmundt) are working on building a new set of "standard solver tests" that we can use to exercise all the solver interfaces in a semi-consistent manner (in particular to exercise the expected solver return status flags), and it would be great to include this in the suite.  Testing it for HiGHS specifically is fine for now -- at least it will be in the repo and can be pulled into the new test suite as it gets assembled.

rschwarz:
It's good that I tried to add tests, and for both interfaces of HiGHS: I was unaware that multiple versions of the `TerminationCondition` class exist, and that the field is sometimes called `maxIterations`, and sometimes `iterationLimit`.

I'm using a Knapsack instance based on random data, which is not solved to optimality in the root with my version of HiGHS and on my machine, so I hope this will also be the case for GitHub.

PR #3525, opened 2025-03-19
Pyomo.DoE: adding more verbose output for sensitivity analysis

smondal13:
## Summary/Motivation:
- Adding more verbose is helpful to analyze the problem
- Adding a new method to check the FIM in the utils.py can help in the general usage of the method.
- Adding a new standalone function in utils.py to calculate different metrics of the FIM is helpful in case the user does not want to build the `get_labeled_model()`.
- Adding arguments in the `reactor_example.py` will help with the testing.

## Changes proposed in this PR:
- added eigenvalues, determinant of FIM, and trace of FIM for sensitivity output in `compute_FIM_full_factorial()`
- added `check_FIM()` in utils.py by separating it from `check_model_FIM()` in doe.py
- added `compute_FIM_metrics()` in utils.py
- added `get_FIM_metrics()` in utils.py
- added arguments in the `reactor_example.py`

### Legal Acknowledgement

By contributing to this software project, I have read the [contribution guide](https://pyomo.readthedocs.io/en/stable/contribution_guide.html) and agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #3639, opened 2025-06-23
Pyomo. DoE: Sensitivity initialization

smondal13:
## Summary/Motivation:
- Initializing the model to a previously solved point in the design space can help the model converge at another point in the design space for sensitivity analysis
- Better error messages in the method can help the user in debugging
- Saving the file can be helpful if the model takes a long time to solve or if the user wants it.
- The pattern generator in the utils.py file can be used in other cases where it can help with the initialization
- Correlation matrix computation functionality will be helpful in computing correlation among the parameters

- Added a new method `compute_FIM_factorial()` in Pyomo.DoE for sensitivity initialization
- Added `snake_traversal_grid_sampling()` and `compute_correlation_matrix()` in utils.py

### Legal Acknowledgement


smondal13:
@adowling2  @djlaky, this draft PR is ready for initial review.

adowling2:
Does this PR depend on changes in #3575?

pyomo/contrib/doe/utils_updated.py
```diff
@@ -0,0 +1,267 @@
+#  ___________________________________________________________________________
```
adowling2: @smondal13 Why are your creating a new `utils` file instead of updating the current one? There might be a point of confusion regarding branches in git.

pyomo/contrib/doe/doe.py
```diff
@@ -1617,6 +1628,221 @@ def compute_FIM_full_factorial(
 
         return self.fim_factorial_results
 
+    def compute_FIM_factorial(
```
adowling2: Was this function already in Pyomo.DoE somewhere else?

pyomo/contrib/doe/doe.py
```diff
@@ -1617,6 +1628,221 @@ def compute_FIM_full_factorial(
...
+
+        # Get the design map keys that match the design_values keys
+        design_map_keys = [
+            k for k in model.experiment_inputs.keys() if k.name in design_values.keys()
+        ]
+        # This ensures that the order of the design_values keys matches the order of the
+        # design_map_keys so that design_point can be constructed correctly in the loop.
+        des_ranges = [design_values[k.name] for k in design_map_keys]
+        if change_one_design_at_a_time:
+            factorial_points = generate_snake_zigzag_pattern(*des_ranges)
```
adowling2: Let's brainstorm different sensitivity analysis sequences we might use. We can then define an enum.

pyomo/contrib/doe/utils.py
```diff
@@ -100,3 +100,43 @@ def rescale_FIM(FIM, param_vals):
 #             pass  # ToDo: Write error for suffix keys that aren't ParamData or VarData
 #
 #     return param_list
+
+
+def generate_snake_zigzag_pattern(*lists):
```
adowling2: Let's come up with a better name. I do not think of "snake" as a technical term.

pyomo/contrib/doe/utils.py
```diff
@@ -100,3 +100,43 @@ def rescale_FIM(FIM, param_vals):
...
+
+        # Enumerate to get the index `i` for the *next* recursive call's sum.
+        for i, value in enumerate(iterable):
+            # Recur for the next list, updating the index_sum.
+            for sub_pattern in _generate_recursive(depth + 1, index_sum + i):
+                # Prepend the current value to the results from deeper levels.
+                yield (value,) + sub_pattern
+
+    # Start the recursion at the first list (depth 0) with an initial sum of 0.
+    yield from _generate_recursive(depth=0, index_sum=0)
```
adowling2, line 362: How are we going to test this function? IMO, the next step is to develop some tests for this. That way, as you try to run the sensitivity analysis for an example, you are least know this part of the code is correct.

adowling2 (commented):
@smondal13 I left some initial comments.

pyomo/contrib/doe/doe.py
smondal13: No, I created this method. I do not see the same name anywhere else.

pyomo/contrib/doe/utils.py
smondal13: @adowling2 , I did check it manually before. I have recently added a test for this in `test_utils.py` for up to 3 variables, and also for different array-like structures. From the test, it works the way we want.

pyomo/contrib/doe/utils_updated.py
smondal13: This issue is resolved now. merged the utils file from my previous [PR # 3525](https://github.com/Pyomo/pyomo/pull/3525).

pyomo/contrib/doe/doe.py
smondal13: @adowling2, we have `compute_FIM_full_factorial()`. I have not changed that method, rather I have added a new method. Maybe we can show a deprecation warning for `compute_FIM_full_factorial()`

smondal13:
@adowling2 , @djlaky, @blnicho  This PR is ready for design review.

pyomo/contrib/doe/doe.py
```diff
@@ -1611,6 +1622,333 @@ def compute_FIM_full_factorial(
...
 
+    def compute_FIM_factorial(
+        self,
+        model=None,
+        design_vals: list = None,
+        abs_change: list = None,
+        rel_change: list = None,
+        n_designs: int = 5,
+        method="sequential",
+        df_settings=(True, None, None, 500),
```
smondal13: Should I use this tuple? I was reluctant to use 4 different arguments for the pd df setting / configuration

smondal13:
TODO: Create a new suffix `design_var`s and pass that to Stephen's `update_model_from_suffix()`

pyomo/contrib/doe/doe.py
adowling2: Yes, a depreciation warning sounds reasonable.

pyomo/contrib/doe/doe.py
```diff
@@ -1802,6 +1815,350 @@ def compute_FIM_full_factorial(
...
+        initialization_scheme="snake_traversal",
+        file_name: str = None,
+    ):
+        """Will run a simulation-based factorial exploration of the experimental input
+        space (i.e., a ``grid search`` or ``parameter sweep``) to understand how the
```
adowling2: Instead of "i.e." should this be "a.k.a." (also knowns as)

pyomo/contrib/doe/doe.py
```diff
@@ -1802,6 +1815,350 @@ def compute_FIM_full_factorial(
...
+                f"Initialization scheme '{initialization_scheme}' is not recognized. "
+                "Using `snake_traversal` as the default initialization scheme."
+            )
+            scheme_enum = SensitivityInitialization.snake_traversal
+
+        if scheme_enum == SensitivityInitialization.snake_traversal:
+            factorial_points = snake_traversal_grid_sampling(*design_values)
+        elif scheme_enum == SensitivityInitialization.nested_for_loop:
+            factorial_points = product(*design_values)
+
```
adowling2, line 2173: Does it make sense to split the function here? That way, we can do extensive testing on the results of `factorial_points` without running the rest of the function.

pyomo/contrib/doe/doe.py
```diff
@@ -1802,6 +1815,350 @@ def compute_FIM_full_factorial(
...
+        """Will run a simulation-based factorial exploration of the experimental input
+        space (i.e., a ``grid search`` or ``parameter sweep``) to understand how the
+        FIM metrics change as a function of the experimental design space. This method
+        can be used for both full factorial and fractional factorial designs.
+
+        Parameters
+        ----------
+        model : DoE model, optional
+            The model to perform the full factorial exploration on. Default: None
+        design_vals : dict,
```
adowling2: After reading this doc string, I do not fully appreciate all of options. I do not think we are looking at all of the edge cases with the tests.

adowling2 (changes requested):

adowling2:
@smondal13 As a next step, I suggest you create a Google Doc as a mock-up for the documentation you plan to write for this new feature. In the mock-up, you can provide example usage to show the programmatic interface. I would set the Google Doc to allow for anonymous commenting and post the link here. Alternatively, you could do this in a markdown `.md` file that you add to this PR. Then, at the next Pyomo dev meeting, you can present the draft documentation and get feedback. This will facilitate the design review process.

smondal13:
To see the Google Doc that shows the documentation and example of this feature, click [here](https://docs.google.com/document/d/1uWlLoITbQLL9QbhLZX-6ZopWtboOhgDNASXa3AVeGc8/edit?tab=t.0)

pyomo/contrib/doe/doe.py
```diff
@@ -1802,6 +1815,376 @@ def compute_FIM_full_factorial(
...
 
+    def compute_FIM_factorial(
+        self,
+        model=None,
+        design_vals: dict = None,
+        abs_change: list = None,
+        rel_change: list = None,
+        n_design_points: int = None,
+        method="sequential",
+        return_df=True,
```
smondal13: Since the method is printing the df depending on this flag, not returning it, does it make more sense to use `show_df` instead of `return_df`?

pyomo/contrib/doe/doe.py
```diff
@@ -1807,6 +1820,376 @@ def compute_FIM_full_factorial(
...
+        traversal_scheme="snake_traversal",
+        file_name: str = None,
+    ):
+        """Will run a simulation-based factorial exploration of the experimental input
```
adowling2: Change "will run" to "performs"

pyomo/contrib/doe/doe.py
```diff
@@ -1807,6 +1820,376 @@ def compute_FIM_full_factorial(
...
+        space (a.k.a., a ``grid search`` or ``parameter sweep``) to understand how the
+        FIM metrics change as a function of the experimental design space. This method
+        can be used for both full factorial and fractional factorial designs.
+
+        Parameters
+        ----------
+        model : DoE model, optional
+            The model to perform the full factorial exploration on. Default: None
+        design_vals : dict,
```
adowling2: @mrmundt @blnicho Do you have any suggestions for avoiding string-based names? I know that we want to avoid string-based names, but I do not have any good ideas of how to do that in a way that is convenient for the user.

pyomo/contrib/doe/doe.py
```diff
@@ -1807,6 +1820,376 @@ def compute_FIM_full_factorial(
...
+                "Please use one of the following: "
+                f"{[e.value for e in DesignSpaceTraversal]}"
+            )
+
+        if scheme_enum == DesignSpaceTraversal.snake_traversal:
+            factorial_points = snake_traversal_grid_sampling(*design_values)
+        elif scheme_enum == DesignSpaceTraversal.nested_for_loop:
+            factorial_points = product(*design_values)
+
+        # TODO: Add more DesignSpaceTraversal schemes
```
adowling2: Remove "TODO", comment could read "Additional DesignSpaceTraversal schemes can be added here as a future extension"

pyomo/contrib/doe/doe.py
```diff
@@ -1807,6 +1820,376 @@ def compute_FIM_full_factorial(
...
+
+        # Save the results to a json file based on the file_name provided
+        if file_name is not None:
+            with open(file_name + ".json", "w") as f:
+                json.dump(self.factorial_results, f, indent=4)
+                self.logger.info(f"Results saved to {file_name}.json.")
+
+        return self.factorial_results
+
     # TODO: Overhaul plotting functions to not use strings
```
adowling2, line 2430: Do we still want to overhaul the plotting functions to not use strings? Or do we just want this function to be consistent with our user interface above? In other words, if we use string-based names above, I think it it okay to use string-based names here.

pyomo/contrib/doe/doe.py
adowling2, line 2435: Let's update this user interface to match the function above. It appears that design variables as specified differently in these two functions. Likewise, this function should do design variable name error checking similar to the function above.

pyomo/contrib/doe/utils.py
```diff
@@ -233,3 +235,105 @@ def get_FIM_metrics(FIM):
         "log10(E-Optimality)": E_opt,
         "log10(Modified E-Optimality)": ME_opt,
     }
+
+
+def snake_traversal_grid_sampling(*array_like_args):
```
adowling2, line 313: Do you want the user to call this method directly? Or is this an internal or private method that our main interface function above uses? If this is a private method, the method name should start with a `_`

pyomo/contrib/doe/utils.py
adowling2: Sounds good. You can resolve this comment.

pyomo/contrib/doe/utils.py
```diff
@@ -233,3 +235,105 @@ def get_FIM_metrics(FIM):
...
+            for sub_pattern in _generate_recursive(depth + 1, index_sum + i):
+                # Prepend the current value to the results from deeper levels.
+                yield (value,) + sub_pattern
+
+    # Start the recursion at the first list (depth 0) with an initial sum of 0.
+    yield from _generate_recursive(depth=0, index_sum=0)
+
+
+def compute_correlation_matrix(
+    covariance_matrix, var_name: list = None, significant_digits=3
```
adowling2: Please make the default for significant digits None.

pyomo/contrib/doe/utils.py
```diff
@@ -233,3 +235,105 @@ def get_FIM_metrics(FIM):
...
+    Computes the correlation matrix from a covariance matrix.
+
+    Parameters
+    ----------
+    covariance_matrix : numpy.ndarray
+        2D array representing the covariance matrix.
+    var_name : list, optional
+        List of variable names corresponding to the rows/columns of the covariance matrix.
+    significant_digits : int, optional
+        Number of significant digits to round the correlation matrix to. Default: 3.
```
adowling2: Number of significant digits for rounding entries of the correlation matrix. Default: None (no rounding)

pyomo/contrib/doe/utils.py
```diff
@@ -233,3 +235,105 @@ def get_FIM_metrics(FIM):
...
+    covariance_matrix, var_name: list = None, significant_digits=3
+):
+    """
+    Computes the correlation matrix from a covariance matrix.
+
+    Parameters
+    ----------
+    covariance_matrix : numpy.ndarray
+        2D array representing the covariance matrix.
+    var_name : list, optional
```
adowling2, line 377: What happens if this optional argument is not provided? The doc string does not tell me what happens with this information or the consequences of not providing it.

pyomo/contrib/doe/utils.py
```diff
@@ -233,3 +235,105 @@ def get_FIM_metrics(FIM):
...
+
+def compute_correlation_matrix(
+    covariance_matrix, var_name: list = None, significant_digits=3
+):
+    """
+    Computes the correlation matrix from a covariance matrix.
+
+    Parameters
+    ----------
+    covariance_matrix : numpy.ndarray
```
adowling2: Can this be a square DataFrame? If yes, what type of error checking should we do on the column and row names?

pyomo/contrib/doe/utils.py
```diff
@@ -233,3 +235,105 @@ def get_FIM_metrics(FIM):
...
+    std_dev = np.sqrt(np.diag(covariance_matrix))
+
+    std_dev_matrix = np.outer(std_dev, std_dev)
+
+    correlation_matrix = covariance_matrix / std_dev_matrix
+
+    # Set the index to the variable names if provided,
+    corr_df = (
+        pd.DataFrame(correlation_matrix, index=var_name, columns=var_name)
+        if var_name
```
adowling2: What is this logic doing here?

adowling2 (changes requested):
Please see my requested changes.

pyomo/contrib/doe/utils.py
smondal13: I want it to be a public function, that's why I have added it in `doe/utils.py`

pyomo/contrib/doe/utils.py
smondal13: This ternary block returns a pd DataFrame if `var_name` is passed; otherwise, it returns a numpy array.

pyomo/contrib/doe/utils.py
smondal13: Added details in the docstring

pyomo/contrib/doe/doe.py
smondal13: Made the FIM metrics consistent in both `compute_FIM_full_factorial()` and `compute_FIM_factorial()`. Added design variable name error checking.

adowling2:
@smondal13 Were you able to address the comments from my review in early February? If yes, I can re-review. Reply here.

smondal13:
> @smondal13 Were you able to address the comments from my review in early February? If yes, I can re-review. Reply here.

Excluding the `string` related comments, the comments are addressed. You can probably wait until @blnicho or @mrmundt replies to that.

pyomo/contrib/doe/doe.py
mrmundt: I'll let @blnicho also chime in but my initial thoughts are:

The best general practice is to accept component objects / references, not names. It is convenient (for users) to optionally accept names and resolve them internally. So if it's possible, support both, with a clear canonical form.

pyomo/contrib/doe/tests/test_doe_errors.py
```diff
@@ -219,8 +219,7 @@ def test_reactor_check_bad_prior_negative_eigenvalue(self):
 
         with self.assertRaisesRegex(
             ValueError,
-            r"FIM provided is not positive definite. It has one or more "
-            r"negative eigenvalue\(s\) less than -{:.1e}".format(
+            r"Matrix provided is not positive definite. It has one or more negative eigenvalue\(s\) less than -{:.1e}".format(
```
jsiirola, line 223: This is a long line.  Can it be broken to stay within the 88-column limit?

pyomo/contrib/doe/doe.py
jsiirola: Actually, I am really confused: why did you create a new method *in the same file* that did effectively the same thing?  Does this replace the other method?  Is it a superset of the other method?  Do we need both?

Also, you might consider renaming this function: I don't know what a `FIM factorial` is (nor how to compute one).  Maybe something like `compute_FIM_from_full_factorial_sampling`?

pyomo/contrib/doe/utils.py
```diff
@@ -266,3 +273,155 @@ def get_FIM_metrics(FIM):
         "log10(E-Optimality)": E_opt,
         "log10(Modified E-Optimality)": ME_opt,
     }
+
+
+def snake_traversal_grid_sampling(*array_like_args):
+    """
+    Generates a multi-dimensional pattern for an arbitrary number of 1D array-like arguments.
+    This pattern is useful for generating patterns for sensitivity analysis when we want
+    to change one variable at a time. This function uses recursion and acts as a generator.
```
jsiirola: Please make sure these are wrapped within 88 columns.

pyomo/contrib/doe/utils.py
```diff
@@ -91,41 +96,43 @@ def rescale_FIM(FIM, param_vals):
     return scaled_FIM
 
 
-def check_FIM(FIM):
+def check_matrix(mat):
```
jsiirola: This should be renamed to indicate what it is checking.  Further, it is not checking and returning True/False, it is asserting (see, e.g. the difference between `check_optimal_termination` and `assert_optimal_termination`).

Maybe rename to something like `assert_matrix_symmetric_positive_definite` or `assert_symmetric_positive_definite`?

pyomo/contrib/doe/utils.py
```diff
@@ -91,41 +96,43 @@ def rescale_FIM(FIM, param_vals):
...
+        raise ValueError("argument mat must be a 2D square matrix")
 
-    # Compute the eigenvalues of the FIM
-    evals = np.linalg.eigvals(FIM)
+    # Compute the eigenvalues of the matrix
+    evals = np.linalg.eigvals(mat)
 
-    # Check if the FIM is positive definite
+    # Check if the matrix is positive definite
     if np.min(evals) < -_SMALL_TOLERANCE_DEFINITENESS:
```
jsiirola: Should these tolerances be passed as optional arguments that default to the global value if None?

pyomo/contrib/doe/utils.py
```diff
@@ -266,3 +273,155 @@ def get_FIM_metrics(FIM):
...
+        NumPy array is returned.
+    significant_digits : int, optional
+        Number of significant digits for rounding entries of the correlation matrix.
+        Default: None (no rounding).
+
+    Returns
+    -------
+    pandas.DataFrame/numpy.ndarray
+        If `var_name` is provided, returns a pandas DataFrame with the correlation matrix
+        and the specified variable names as both index and columns. If `var_name` is not
```
jsiirola: Please make sure these are wrapped <88 columns.

pyomo/contrib/doe/doe.py
```diff
@@ -2005,6 +2018,378 @@ def compute_FIM_full_factorial(
...
+                        )
+                    val = lb
+                    while val <= ub:
+                        des_val.append(val)
+                        val += del_val
+
+                else:
+                    raise ValueError(
+                        "Unexpected error in generating design values. Please check the"
+                        " input parameters."
```
jsiirola: Please make sure these are wrapped <88 columns.

pyomo/contrib/doe/doe.py
```diff
@@ -2005,6 +2018,378 @@ def compute_FIM_full_factorial(
...
+                )
+
+            # Get the design map keys that match the design_values keys
+            design_map_keys = [
+                k
+                for k in model.experiment_inputs.keys()
+                if k.name in design_vals.keys()
+            ]
+            # This ensures that the order of the design_values keys matches the order of the
+            # design_map_keys so that design_point can be constructed correctly in the loop.
```
jsiirola: Please make sure these are wrapped <88 columns.

pyomo/contrib/doe/doe.py
```diff
@@ -2005,6 +2018,378 @@ def compute_FIM_full_factorial(
...
+            Method used for FIM computation. Options: ``"kaug"`` and
+            ``"sequential"``. Default: ``"sequential"``.
+        return_df : bool, optional
+            If ``True``, print results as a pandas DataFrame.
+        traversal_scheme : str, optional
+            Which scheme to use for initializing the design variables.
+            Options are ``"snake_traversal"`` and ``"nested_for_loop"``.
+            If ``"snake_traversal"`` is used, the design variables will be initialized in a
+            snake-like pattern where only one design variables change at a time.
+            If ``"nested_for_loop"`` is used, the design variables will be initialized using a
```
jsiirola: Please make sure these are wrapped <88 columns.

pyomo/contrib/doe/doe.py
```diff
@@ -2005,6 +2018,378 @@ def compute_FIM_full_factorial(
...
+        # Optional: Check if at least one was provided, if that's a requirement
+        if num_des_provided == 0:
+            raise ValueError(
+                "At least one of the following arguments: "
+                "`design_vals`, `n_design_points`, or `abs_change` and/or `rel_change`"
+                " must be provided."
+            )
+
+        if design_vals is not None:
+            # Check whether the design_ranges keys are in the experiment_inputs
```
jsiirola: `design_ranges` or `design_vals`?  This looks like a copy-paste error.

pyomo/contrib/doe/doe.py
```diff
@@ -2005,6 +2018,378 @@ def compute_FIM_full_factorial(
...
+        Parameters
+        ----------
+        model : DoE model, optional
+            Reserved for API compatibility. A fresh labeled model clone is built
+            internally for the factorial exploration.
+        design_vals : dict, optional
+            Mapping of design variable names (strings) to 1D array-like values:
+            ``{"var_name": [values...]}``. Keys must be a subset of
+            ``model.experiment_inputs`` names. Variables not included are fixed at
+            their model values.
```
jsiirola: Using names is fragile and best avoided.  Why are you using them?  This instance has an experiment, which uses a proper ComponentMap, so you should accept the same API here.

pyomo/contrib/doe/doe.py
```diff
@@ -2005,6 +2018,378 @@ def compute_FIM_full_factorial(
...
+            ``design_vals`` and ``n_design_points`` are not provided. Length must
+            match the number of design variables in ``model.experiment_inputs``.
+            If provided without ``rel_change``, ``rel_change`` is treated as zeros.
+        rel_change : list, optional
+            Relative change for each design variable. Used only when
+            ``design_vals`` and ``n_design_points`` are not provided. Length must
+            match the number of design variables in ``model.experiment_inputs``.
+            If provided without ``abs_change``, ``abs_change`` is treated as zeros.
+        n_design_points : int, optional
+            Number of equally spaced points per design variable generated from
```
jsiirola:
```suggestion
            Number of equally-spaced points per design variable generated from
```

pyomo/contrib/doe/doe.py
```diff
@@ -2005,6 +2018,378 @@ def compute_FIM_full_factorial(
...
+            bounds via ``np.linspace``. Requires finite lower and upper bounds.
```
jsiirola:
```suggestion
            bounds via :func:`np.linspace`. Requires finite lower and upper bounds.
```

pyomo/contrib/doe/doe.py
```diff
@@ -2005,6 +2018,378 @@ def compute_FIM_full_factorial(
...
+        method : str, optional
+            Method used for FIM computation. Options: ``"kaug"`` and
+            ``"sequential"``. Default: ``"sequential"``.
```
jsiirola:
```suggestion
            Method used for FIM computation. Options: ``kaug`` and
            ``sequential``. Default: ``sequential``.
```
(Don't include the quotes, as that implies that the string contains quotes)

pyomo/contrib/doe/doe.py
jsiirola: Same here: the quotes are not part of the string.

pyomo/contrib/doe/doe.py
```diff
@@ -2005,6 +2018,378 @@ def compute_FIM_full_factorial(
...
+
+        # Make new model for factorial design
+        self.factorial_model = self.experiment.get_labeled_model(
+            **self.get_labeled_model_args
+        ).clone()
+        model = self.factorial_model
+
+        # Group the mutually exclusive arguments for passing design values
+        # abs_change, and rel_change can be provided together or separately
+        num_des_provided = (
```
jsiirola: This implies you are computing the number of design points provided.  A less confusing variable might be `num_des_args_provided`

pyomo/contrib/doe/doe.py
```diff
@@ -2005,6 +2018,378 @@ def compute_FIM_full_factorial(
...
+            (design_vals is not None)
+            + (n_design_points is not None)
+            + (abs_change is not None or rel_change is not None)
+        )
+
+        # Check the count and raise an error if more than one was provided
+        if num_des_provided > 1:
+            raise ValueError(
+                "Please provide only one of the following arguments: "
+                "`design_vals`, `n_design_points`, or `abs_change` and/or `rel_change`."
```
jsiirola: This is confusing.  I think you want to say, "design_vals, n_design_points, and abs_change/rel_change are mutually exclusive."

pyomo/contrib/doe/doe.py
```diff
@@ -2005,6 +2018,378 @@ def compute_FIM_full_factorial(
...
+            )
+
+        # Optional: Check if at least one was provided, if that's a requirement
+        if num_des_provided == 0:
+            raise ValueError(
+                "At least one of the following arguments: "
+                "`design_vals`, `n_design_points`, or `abs_change` and/or `rel_change`"
+                " must be provided."
```
jsiirola: This is confusing.  Something like "Missing required argument: specify one of design_vals, n_design_points, or abs_change/rel_change."?

pyomo/contrib/doe/doe.py
```diff
@@ -2005,6 +2018,378 @@ def compute_FIM_full_factorial(
...
+        abs_change : list, optional
+            Absolute change for each design variable. Used only when
+            ``design_vals`` and ``n_design_points`` are not provided. Length must
+            match the number of design variables in ``model.experiment_inputs``.
+            If provided without ``rel_change``, ``rel_change`` is treated as zeros.
+        rel_change : list, optional
+            Relative change for each design variable. Used only when
+            ``design_vals`` and ``n_design_points`` are not provided. Length must
+            match the number of design variables in ``model.experiment_inputs``.
+            If provided without ``abs_change``, ``abs_change`` is treated as zeros.
```
jsiirola: Should these also accept a single float, that we copy to make an appropriately-sized list?

I also think that the names are misleading.  Perhaps `abs_step` and `rel_step` would be more clear?

pyomo/contrib/doe/doe.py
```diff
@@ -2005,6 +2018,378 @@ def compute_FIM_full_factorial(
...
+                    raise ValueError(
+                        f"{comp.name} does not have a lower or upper bound."
+                    )
+
+                if abs_change is None and rel_change is None:
+                    des_val = np.linspace(lb, ub, n_design_points)
+
+                # if abs_change and/or rel_change is provided, generate design values
+                # using the formula:
+                # change_in_value = lower_bound * rel_change + abs_change
```
jsiirola: This formula seems suspect.  I would have expected it to be
```
change_in_value = (upper_bound - lower_bound) * rel_change + abs_change

File: pyomo/contrib/doe/doe.py [hunk 39]
```diff
@@ -2005,6 +2018,378 @@ def compute_FIM_full_factorial(
 
         return self.fim_factorial_results
 
+    def compute_FIM_factorial(
```
jsiirola: This function is a beast and should be broken up into helper functions.  At the very least, pull the (probably overly complex) logic for generating the design points into it's own (private) function.

pyomo/contrib/doe/doe.py
```diff
@@ -2005,6 +2018,378 @@ def compute_FIM_full_factorial(
...
+                        f"`{len(design_keys)}` as `design_keys`."
+                    )
+
+            if rel_change:
+                if len(rel_change) != len(design_keys):
+                    raise ValueError(
+                        "`rel_change` must have the same length of "
+                        f"`{len(design_keys)}` as `design_keys`."
+                    )
+
```
jsiirola, line 2091: The logic here is pretty convoluted.  You repeatedly test to see which "source" you are using for the design points.  I would recommend refactoring.

pyomo/contrib/doe/doe.py
```diff
@@ -2005,6 +2018,378 @@ def compute_FIM_full_factorial(
...
+            for i, comp in enumerate(design_keys):
+                lb = comp.lb
+                ub = comp.ub
+                # Check if the component has finite lower and upper bounds
+                if lb is None or ub is None:
+                    raise ValueError(
+                        f"{comp.name} does not have a lower or upper bound."
+                    )
+
+                if abs_change is None and rel_change is None:
```
jsiirola: Note that at this point you don't have to test both abs and rel: they are either both None or both non-None.

pyomo/contrib/doe/doe.py
```diff
@@ -2005,6 +2018,378 @@ def compute_FIM_full_factorial(
...
+                    del_val = comp.lb * rel_change[i] + abs_change[i]
+                    if del_val == 0:
+                        raise ValueError(
+                            f"Design variable {comp.name} has no change in value - "
+                            "check abs_change and rel_change values."
+                        )
+                    val = lb
+                    while val <= ub:
+                        des_val.append(val)
+                        val += del_val
```
jsiirola:
```suggestion
                    des_val = list(range(lb, ub, del_val))
```

pyomo/contrib/doe/doe.py
```diff
@@ -2005,6 +2018,378 @@ def compute_FIM_full_factorial(
...
+                    f"The code has run for {round(sum(time_set), 2)} seconds."
+                )
+                self.logger.info(
+                    "Estimated remaining time:  %s seconds",
+                    round(
+                        sum(time_set) / (curr_point) * (total_points - curr_point + 1),
+                        2,
+                    ),
+                )
+            except:
```
jsiirola: Bare `except:` is potentially problematic, as it catching everything including both solver errors and generic programming errors.  At a minimum, you should include the exception in the log so that a user can get an idea as to why things failed.

pyomo/contrib/doe/doe.py
```diff
@@ -2005,6 +2018,378 @@ def compute_FIM_full_factorial(
...
+                trace_cov,
+                trace_FIM,
+                E_vals,
+                _,
+                D_opt,
+                A_opt,
+                pseudo_A_opt,
+                E_opt,
+                ME_opt,
+            ) = compute_FIM_metrics(FIM)
```
jsiirola: This API should be re-thought: returning this many values is asking for a programming error (e.g., getting the order wrong).  I would strongly recommend changing the return value to a `namedtuple` (or a general class, if appropriate).

pyomo/contrib/doe/doe.py
```diff
@@ -2005,6 +2018,378 @@ def compute_FIM_full_factorial(
...
+            dict_for_df = {
+                k: v for k, v in factorial_results.items() if k not in exclude_keys
+            }
+            res_df = pd.DataFrame(dict_for_df)
+            print("\n\n=========Factorial results===========")
+            print("Total points:", total_points)
+            print("Success count:", success_count)
+            print("Failure count:", failure_count)
+            print("\n")
+            print(res_df)
```
jsiirola: We generally strongly discourage `print()` within Pyomo.  Instead, output should be channeled via loggers or global streams (like `sys.stdout`) that are easily replaceable.

pyomo/contrib/doe/doe.py
```diff
@@ -2081,23 +2474,63 @@ def draw_factorial_figure(
...
+        if len(set(sensitivity_design_variables)) != len(sensitivity_design_variables):
+            raise ValueError(
+                "Sensitivity design variables contain duplicates. Please provide "
+                "unique design variable names."
+            )
+
+        if not isinstance(fixed_design_variables, dict):
+            raise ValueError("``fixed_design_variables`` must be a dictionary.")
+        if any(not isinstance(name, str) for name in fixed_design_variables.keys()):
+            raise ValueError("Fixed design variable names must be strings.")
```
jsiirola: This kind of type checking is very "non-Pythonic" and something I would generally discourage.  Python encourages duck-typing: as long as the variable *behaves* like the desired type, then it should be OK.

pyomo/contrib/doe/doe.py
smondal13: @jsiirola This function is a superset of the already existing `compute_FIM_full_factorial()` method.  So, I am not sure if I should keep it under the same name since the "full" in the method name can be misleading. It can use a full factorial (changing all the decision variables) and a fractional factorial (changing some of the decision variables). This method basically analyzes the sensitivity.

@adowling2, @slilonfe5, @snarasi2, @sscini  Can we name it something like `compute_sensitivity_at_design_points()`? Or similar to the one @jsiirola suggested, `compute_FIM_from_factorial_sampling`?

pyomo/contrib/doe/doe.py
smondal13: Yes. I think that would be better.

pyomo/contrib/doe/doe.py
smondal13: yes, thank you. I have changed the name

pyomo/contrib/doe/doe.py
smondal13: Yes. I have updated the error message

pyomo/contrib/doe/doe.py
smondal13: Yes. I have updated the error message

pyomo/contrib/doe/doe.py
smondal13: Resolved

pyomo/contrib/doe/doe.py
smondal13: Done

pyomo/contrib/doe/doe.py
smondal13: I have refactored the code.

pyomo/contrib/doe/doe.py
smondal13: I did implement it as a relative change (step) of the lower bound. But using their range may be a better choice.

pyomo/contrib/doe/doe.py
smondal13: Added the error message inside the logger warning.

pyomo/contrib/doe/doe.py
smondal13: I have converted this to a named tuple. I think this is better since the returned object should be immutable.

pyomo/contrib/doe/doe.py
smondal13: I have repaced the `print` with `sys.stdout`

pyomo/contrib/doe/doe.py
smondal13: I have removed the strict type checks,

pyomo/contrib/doe/utils.py
smondal13: That's a very good suggestion. I have renamed the function name to make it more explicit.

pyomo/contrib/doe/utils.py
smondal13: I think passing these tolerances as optional argument is a good idea. I have added them as optional arguments

pyomo/contrib/doe/doe.py
mrmundt: +1 on this. This function is _huge_.

pyomo/contrib/doe/doe.py
smondal13: I have refactored and added private methods.

pyomo/contrib/doe/doe.py
smondal13: Great suggestion, and I implemented that split: traversal generation now lives in a dedicated helper (_materialize_factorial_points).
Current tests already cover traversal behavior through the public API (including invalid scheme errors and explicit nested-for-loop ordering on a 2x2 grid).

pyomo/contrib/doe/utils.py
smondal13: Done.

pyomo/contrib/doe/utils.py
smondal13: Done

pyomo/contrib/doe/utils.py
smondal13: Good point. We now support square DataFrame input and have added explicit label checks: row labels must be unique, column labels must be unique, and row/column labels must match in the same order. If any of those fail, we raise a `ValueError` with a specific message.

pyomo/contrib/doe/doe.py
smondal13: @jsiirola Thanks for the suggestion. I agree and have refactored this in the latest update.

pyomo/contrib/doe/doe.py
slilonfe5: @smondal13 If sensitivity is what you are doing, I think `compute_sensitivity_at_design_points` is better.

PR #3641, opened 2025-06-23
Enable parallel big-M calculation for gdp.mbigm transformation

sadavis1:
## Summary/Motivation:
The multiple big-M transformation tends to slow down as model size increases due to linear or quadratic growth in the number of subsolver runs required. This change parallelizes M calculation using Python's multiprocessing module. The threading module was tried at first, but due to previously discussed issues was switched to multiprocessing.

- Rework flow of gdp.mbigm transformation. Calculate M values all at once using a multiprocessing.Pool.starmap call.
- Switch gdp.mbigm big-M calculation from using the primal bound to using the dual bound, as the primal bound is not mathematically correct in the presence of numerical error (but any dual bound is valid).
- Add configuration option 'use_primal_bound' to reenable old behavior, in case a solver like ipopt that cannot provide a dual bound is used.
- Add configuration option 'threads' to control number of multiprocessing workers:
   - By default, use os.cpu_count() - 1 workers. Note that this is potentially harmful; for example, it may lead to using more Gurobi license tokens at once, also on Windows the model must now be pickleable (see below).
   - When set to 1 or 0, do not use multiprocessing and revert to fully single-threaded operation
- Add configuration option 'process_spawn_mechanism' to determine how we spawn processes.
   - Each mechanism described in the [Python docs](https://docs.python.org/3/library/multiprocessing.html#multiprocessing-start-methods) is available.
   - Dynamically choose the default option: 'spawn' on Windows, 'fork' on Unix, unless we can detect that there are multiple threads, in which case we use 'forkserver' instead.
   - When using 'spawn' or 'forkserver', models must be pickled to give to the other processes. Depend on `dill` in this case as models often contain nested functions and in my testing do not reliably pickle without it. I think this code leads to a nested pickle, but I'm not sure if anything can be done about it.
- Fix a bug in bigm_mixin.py (name 'logger' used without being defined)
- contrib/solver/factory.py: Set `name` class attribute of LegacySolverWrapper derived class to the solver's legacy_name. This is so that code like this:
```
from pyomo.environ import SolverFactory
solver = SolverFactory('gurobi_direct_v2')
solver_effective_copy = SolverFactory(solver.name, options=solver.options)
```
will behave as intended (as solvers do not reliably pickle even with dill). This does not affect code that gets the original contrib Solver directly from the pyomo.contrib.solver.common.factory.SolverFactory.
- Add check that we are not passing a contrib Solver object, unless it is a LegacySolverWrapper (it has a .solve attribute, but the code will choke later in several places if we do not reject it).
- Add tests to ensure that big-M calculation functions properly with the different start methods. These are essentially copy-pastes of `test_calculated_Ms_correct` with parameters to the solve() call altered.

Since this is a performance change, I ran a test on the medium-sized instance `gdp_col` from GDPlib, using baron as the subsolver.
test_gdp_col_fork
It kind of looks like `f(x) = 1/x`, if you squint in such a way that you cannot see the bottom of the chart. This instance transformed in 145 seconds on the current `main` branch, so this is not a regression in the single-threaded case. Naturally, things are slightly slower when using 'spawn', but I do at least make sure we only pickle the model once per thread (and hopefully only once total? I'm not sure how multiprocessing works on the inside, but it really should cache these).

I also tested the small instance `jobshop`, to ensure nothing horrible happened.
test_jobshop_fork
On the current main branch, this model transforms in 0.36 seconds so again there is no regression.

Finally, there seems to be a bug when using this transformation with `gurobi_direct` v1. It works fine with the other solvers I've tried, so I suspect it's a bug in that interface, but I haven't tracked it down yet. This combination also has errors on the current main branch, but they're different errors, so it's hard to know if I've changed anything in that regard.

### Legal Acknowledgement

### DIFF
--- a/pyomo/gdp/tests/test_mbigm.py
+++ b/pyomo/gdp/tests/test_mbigm.py
@@ -51,6 +51,10 @@ gurobi_available = (
     SolverFactory('gurobi').available(exception_flag=False)
     and SolverFactory('gurobi').license_is_valid()
 )
+gurobi_direct_available = (
+    SolverFactory('gurobi_direct_v2').available(exception_flag=False)
+    and SolverFactory('gurobi_direct_v2').license_is_valid()
+)
 exdir = normpath(join(PYOMO_ROOT_DIR, 'examples', 'gdp'))
 
 
@@ -970,7 +974,7 @@ class LinearModelDecisionTreeExample(CommonTests):
         self.assertStructuredAlmostEqual(mbm.get_all_M_values(m), self.get_Ms(m))
 
     # Make sure we don't choke on a LegacySolverWrapper
-    @unittest.skipUnless(gurobi_available, "Gurobi is not available")
+    @unittest.skipUnless(gurobi_direct_available, "Gurobi direct is not available")
     @unittest.skipUnless(dill_available, "Dill is not available")
     def test_calculated_Ms_legacy_solver_wrapper(self):
         m = self.make_model()


pyomo/contrib/solver/common/factory.py
```diff
@@ -31,6 +31,7 @@ def decorator(cls):
             class LegacySolver(LegacySolverWrapper, cls):
                 pass
 
+            LegacySolver.name = legacy_name
```
mrmundt: Why add this? Did you notice something not acting as intended?

pyomo/gdp/plugins/bigm_mixin.py
```diff
@@ -23,7 +27,7 @@ def _convert_M_to_tuple(M, constraint, disjunct=None):
         else:
             try:
                 M = (-M, M)
-            except:
+            except Exception:
```
mrmundt, line 30: You touched this line so now I get to snark. It's good practice, if catching a general exception, to raise the original exception as well so folks can directly inspect it.

pyomo/gdp/plugins/multiple_bigm.py
```diff
@@ -271,89 +341,297 @@ def _apply_to_impl(self, instance, **kwds):
...
-            raise GDP_Error(
-                "Cannot do multiple big-M reformulation for "
-                "Disjunction '%s' with OR constraint.  "
-                "Must be an XOR!" % obj.name
-            )
-
-        (transBlock, algebraic_constraint) = self._setup_transform_disjunctionData(
-            obj, root_disjunct
-        )
+    def _transform_disjunctionDatas(
```
mrmundt, line 381: Is it possible to break this up? This function is _huge_.

pyomo/contrib/solver/common/factory.py
sadavis1: Well, the goal here is for `SolverFactory(SolverFactory('gurobi_direct_v2').name)` to give me another `gurobi_direct_v2` LegacySolverWrapper object. But right now it gives me a gurobi direct v1 from pyomo.solvers.plugins.solvers, because the wrapped solver is returning its true name of `gurobi_direct`, rather than the pseudonym it goes by in legacy solver land.

The original reason behind all of this admitted nonsense is that the api for this transformation takes a constructed solver object, but I can't actually give it to the other process because solvers don't pickle reliably, so I'm making an effective copy on the other side by reconstructing it from the .name and .options attributes. Actually this hack would still be necessary even if threading worked and I didn't have to deal with processes, because solvers don't copy well either, nor work properly if you use the same solver multiple times concurrently.

pyomo/gdp/plugins/bigm_mixin.py
sadavis1: Don't we already do this? There is a `raise` statement after we log.

pyomo/gdp/plugins/multiple_bigm.py
sadavis1: It's even worse than it looks, because _setup_jobs_for_disjunction is basically just the inner body of this function that I transposed out so it would be less offensively indented (notice it mutates like 4 of its parameters). The problem is that the whole transformation is basically a big ball of state until it's done and I don't know if it can really not be that way. Emma's version was a lot nicer because it did the disjunctions one by one, but I can't do that if I want it to use threads effectively.

All that said, I will look and see if I can separate any more of this out in a reasonably clean way.

sadavis1:
After discussion today, I have reverted the change to the LegacySolverWrapper class names, and switched from sending the solver name to recreate the solver to sending the solver class to recreate the solver. This makes the assumption that all solver classes (besides contrib without the wrapper, which we reject for other reasons) can be correctly constructed with the single named argument `options`.

Also, using this on windows now depends on `dill` even more, because solvers can be instances of nested classes so that's another dill.dumps(). I went ahead and completed the trio by dill-ing the options parameter too -- who knows, maybe there's a way to pass a nested function into one.

pyomo/contrib/solver/common/factory.py
sadavis1: Reverted this change

pyomo/gdp/plugins/multiple_bigm.py
sadavis1: I moved the multiprocessing pool setup to its own instance method, slightly improving this situation

pyomo/gdp/plugins/bigm_mixin.py
mrmundt: Oooo, I didn't see it because it wasn't a "changed line" and was hidden from the diff view. My bad.

pyomo/gdp/plugins/bigm_mixin.py
```diff
@@ -9,12 +9,16 @@
...
 
+import logging
+
 from pyomo.gdp import GDP_Error
 from pyomo.common.collections import ComponentSet
 from pyomo.contrib.fbbt.expression_bounds_walker import ExpressionBoundsVisitor
 import pyomo.contrib.fbbt.interval as interval
 from pyomo.core import Suffix
 
+logger = logging.getLogger('pyomo.gdp.bigm')
```
jsiirola: Should we start getting into the habit of making this `getLogger(__name__)`?

pyomo/gdp/plugins/multiple_bigm.py
```diff
@@ -71,15 +69,38 @@
...
+_thread_local.config_use_primal_bound = None
+_thread_local.in_progress = False
+
 
 def Solver(val):
     if isinstance(val, str):
         return SolverFactory(val)
     if not hasattr(val, 'solve'):
         raise ValueError("Expected a string or solver object (with solve() method)")
+    if isinstance(val, NewSolverBase) and not isinstance(val, LegacySolverWrapper):
```
jsiirola, line 96: How does this admonition work with APPSI solvers?  Should you just check for not deriving off the old Solver base class?

pyomo/gdp/plugins/multiple_bigm.py
```diff
@@ -217,6 +238,53 @@ class MultipleBigMTransformation(GDP_to_MIP_Transformation, _BigM_MixIn):
...
+            one. If set to one or less, do not spawn processes, and revert to
+            single-threaded operation.
+            """,
+        ),
+    )
+    CONFIG.declare(
+        'process_start_method',
+        ConfigValue(
+            default=None,
+            domain=ProcessStartMethod,
```
jsiirola: Why is this a string and not an Enum?  I think for things like this, an Enum (and validation using `In`) would be better?

pyomo/gdp/plugins/multiple_bigm.py
```diff
@@ -217,6 +238,53 @@ class MultipleBigMTransformation(GDP_to_MIP_Transformation, _BigM_MixIn):
         """,
         ),
     )
+    CONFIG.declare(
+        'threads',
+        ConfigValue(
+            default=None,
+            domain=int,
```
jsiirola: Should the domain be `PositiveInt`?

pyomo/gdp/plugins/multiple_bigm.py
```diff
@@ -271,89 +349,242 @@ def _apply_to_impl(self, instance, **kwds):
...
-        # ordered here, but using a set because I will remove infeasible
-        # Disjuncts from it if I encounter them calculating M's.
-        active_disjuncts = ComponentSet(disj for disj in obj.disjuncts if disj.active)
-        # First handle the bound constraints if we are dealing with them
-        # separately
-        transformed_constraints = set()
-        if self._config.reduce_bound_constraints:
-            transformed_constraints = self._transform_bound_constraints(
-                active_disjuncts, transBlock, arg_Ms
+                active_disjuncts[t] = ComponentSet(
```
jsiirola, line 442: It is a little unfortunate that the refactor removed the comment about using a `ComponentSet` for determinism.

pyomo/gdp/plugins/multiple_bigm.py
```diff
@@ -271,89 +349,242 @@ def _apply_to_impl(self, instance, **kwds):
...
+            # map the DisjunctionData to its XOR constraint to mark it as
+            # transformed
+            disjunction._algebraic_constraint = weakref_ref(
+                algebraic_constraint[disjunction.index()]
             )
+            disjunction.deactivate()
+
+    # Do the inner work setting up M calculation jobs for one
+    # disjunction. This mutates the parameters Ms, jobs,
+    # transBlock._mbm_values, and also self.used_args.
```
jsiirola: It would be helpful to convert this to a docstring.  In particular, it would be good to document the data structures that are being created (lists? dicts? if a dict, from what to what, etc.).

jsiirola (commented):
A couple minor questions / nits, but overall looks good.

pyomo/gdp/plugins/multiple_bigm.py
sadavis1: To be honest, I only included this error check because it's an error I personally managed to run into. In general I don't think there's any way to tell if something is a legitimate old-style solver or not (ie, takes a model and produces a SolverResults), except maybe by passing a trivial model and seeing what comes back, which seems like a little too much work for an argument validation function. I think the "normal" solvers all derive from the base class, but there are plenty of weird ones like gdpopt, mindtpy, contrib.multistart, the pynumero scipy stuff, etc. that should be a legitimate argument here but don't inherit from anything common.

Maybe I should just add another special case for appsi? Surely the .solve()-able classes that do work here outnumber those that don't?

pyomo/gdp/plugins/bigm_mixin.py
sadavis1: I went ahead and made that change

pyomo/gdp/plugins/multiple_bigm.py
sadavis1: Python itself uses strings, so a user might expect them to work. I changed it to an enum type derived from str, and used InEnum as the domain validator. This automatically converts any string that does get passed, as best I can tell.

pyomo/gdp/plugins/multiple_bigm.py
sadavis1: Probably, unless we intend to do negative work and go back in time

pyomo/gdp/plugins/multiple_bigm.py
sadavis1: I added a new comment that mentions again that we're using it for determinism

pyomo/gdp/plugins/multiple_bigm.py
sadavis1: I replaced this comment with a docstring, but I'm not sure if it's properly formatted or if it contains all the things it should contain.

sadavis1:
I applied the docstrings from the arguments to apply_to using document_kwargs_from_configdict. Actually this seems to be the only place I can put it, and I confused myself earlier -- it can't go on CONFIG because decorators don't go on class attributes, and putting it on the class would make them look like constructor arguments. It also seems like this was only partially successful, because it added all the contents of `doc`, but for a lot of these you also need the strings from `description` to get the full picture...

pyomo/gdp/plugins/multiple_bigm.py
jsiirola: I wonder if the right solution is to add something like a `api_version` method to the various solver interfaces that you can query (something like 1 for legacy, 2 for APPSI, and 3 for `contrib.solver`)?

@mrmundt: thoughts?

jsiirola (approved):

pyomo/gdp/plugins/multiple_bigm.py
mrmundt: @jsiirola - This is a good question. We have a "version" of this in `pyomo.__future__` - `solver_factory_v3`. Could we use this system to our advantage somehow? (Basically, I am saying yes, but wondering if we can build on something we already have)

pyomo/gdp/plugins/multiple_bigm.py
```diff
@@ -271,89 +363,260 @@ def _apply_to_impl(self, instance, **kwds):
...
+        disjunction. Mutate the parameters Ms, jobs, transBlock._mbm_values,
+        and self.used_args.
+
+        Args:
+            self: self.used_args map from (constraint, disjunct) to M tuple
+                updated with the keys used from arg_Ms
+            disjunction: disjunction to set up the jobs for
+            active_disjuncts: map from disjunctions to ComponentSets of active
+                disjuncts
+            arg_Ms: user-provided map from (constraint, disjunct) to M value
```
mrmundt, line 578: `transformed_constraints` is missed in this list.

pyomo/gdp/plugins/multiple_bigm.py
jsiirola: It's different, though: The thing in `__future__` is selecting what the "default" solver factory is.  In this case, we want to know, "given a particular solver instance, what API does it support"?

pyomo/gdp/plugins/multiple_bigm.py
```diff
@@ -271,89 +363,260 @@ def _apply_to_impl(self, instance, **kwds):
...
+
+    def _setup_jobs_for_disjunction(
+        self,
+        disjunction,
+        active_disjuncts,
+        transformed_constraints,
+        arg_Ms,
+        Ms,
+        jobs,
+        transBlock,
```
mrmundt: Nitpick - `transBlock` is weird here in comparison to `active_disjuncts`, `args_Ms`. It's a switch from snake_case to camelCase for no discernible reason.

pyomo/gdp/plugins/multiple_bigm.py
mrmundt: Fair point. What about a property that we introduce based on an enum, e.g.,
```python
from enum import Enum

class SolverAPI(Enum):
    V1 = 1
    APPSI = 2
    V2 = 3
```

And then in the solver base classes (e.g., `OptSolver` and `SolverBase`), we make a property:
```python
@property
def api_version(self):
    return SolverAPI.V1
```

jsiirola: Something like that was what I was thinking.  I would have a slight preference for it to be a class attribute over a `@property` so that you can detect the API version without necessarily instantiating an instance.

jsiirola:
> I applied the docstrings from the arguments to apply_to using document_kwargs_from_configdict. Actually this seems to be the only place I can put it, and I confused myself earlier -- it can't go on CONFIG because decorators don't go on class attributes, and putting it on the class would make them look like constructor arguments. It also seems like this was only partially successful, because it added all the contents of `doc`, but for a lot of these you also need the strings from `description` to get the full picture...

This motivated #3667.  Now that it is merged, I have ported it here.  I also updated the `doc` fields to be self-contained (i.e., `description` is meant to be a short description of the field, and `doc` is the (standalone) full documentation).  I think this can be merged once we get the second review.

jsiirola:
I just slipped in a fix to the Config system that should get tests passing.

pyomo/gdp/plugins/multiple_bigm.py
```diff
@@ -165,10 +197,13 @@ class MultipleBigMTransformation(GDP_to_MIP_Transformation, _BigM_MixIn):
...
             domain=bool,
-            description="Flag indicating whether or not to handle disjunctive "
-            "constraints that bound a single variable in a single (tighter) "
-            "constraint, rather than one per Disjunct.",
+            description="Combine constraints in multiple disjuncts that "
+            "bound a single variable into a single constraint",
             doc="""
+        Flag indicating whether or not to handle disjunctive
+        constraints that bound a single variable in a single (tighter)
+        constraint, rather than one per Disjunct.
```
emma58: Sorry, looks like this is my mistake propagating, but this is wrong--it's not tighter. It's equivalent in relaxation. It is more compact however...

pyomo/gdp/plugins/multiple_bigm.py
```diff
@@ -217,6 +251,60 @@ class MultipleBigMTransformation(GDP_to_MIP_Transformation, _BigM_MixIn):
...
+            process pool is created, in which case we instead use 'forkserver'. On
+            Windows, the default and only possible option is 'spawn'. Note that if
+            'spawn' or 'forkserver' are selected, we depend on the `dill` module for
+            pickling, and model instances must be pickleable using `dill`. This option is
+            ignored if `threads` is set to 1.
+            """,
+        ),
+    )
+    CONFIG.declare(
+        'use_primal_bound',
```
emma58, line 292: Perhaps this should be `estimate_M_with_primal_bound`? Is that too long and annoying? @jsiirola

pyomo/gdp/plugins/multiple_bigm.py
emma58: Poor Soren, it's my bad habit. :P We should change it though.

emma58 (approved):
Sorry for the much-delayed review, @sadavis1. This looks good--I have a few minor nits and I think that Miranda is right in a couple places, but I'll just push some changes now so that we can get this into the release.

mrmundt:
Okay, the Jenkins test failure might actually be real because it's now repeated.

emma58:
Oh, yeah, that test checked if gurobi was available, but not gurobi_direct_v2, which is what it actually uses... Let's see if this works.

PR #3648, opened 2025-06-26
Add external function library

eslickj:
## Summary/Motivation:

Add and external function library for useful functions that are not easily implemented as Pyomo expressions.
- sgn(x)x^2, https://github.com/Pyomo/pyomo/pull/3622
- sinc(x) = sin(x)/x, https://github.com/Pyomo/pyomo/issues/323

- Add external function library to Pyomo extensions

### Legal Acknowledgement


pyomo/contrib/external_functions/tests/test_functions.py
```diff
@@ -0,0 +1,104 @@
...
+from pyomo.common.fileutils import find_library
+
+flib = find_library("external_functions")
+is_pypy = platform.python_implementation().lower().startswith("pypy")
+
+
+@unittest.skipUnless(flib, 'Could not find the "external" library')
+class TestAMPLExternalFunction(unittest.TestCase):
+
+    @unittest.skipIf(is_pypy, "Cannot evaluate external functions under pypy")
```
mrmundt: Why not move this to the class level rather than put it on every single test?

mrmundt (approved):

doc/OnlineDocs/explanation/modeling_utils/external_functions/index.rst
```diff
@@ -0,0 +1,112 @@
...
+sgnsqr(x)
+~~~~~~~~~
+
+This function is defined as:
+
+.. math::
+
+    \text{sgnsqr}(x) = \text{sgn}(x)x^2
+
+This function is only $C^1$ smooth because at 0 the second derivative is undefined and the jumps from -2 to 2.
```
blnicho:
```suggestion
This function is only :math:`C^1` smooth because at 0 the second derivative is undefined and the jumps from -2 to 2.  
```

doc/OnlineDocs/explanation/modeling_utils/external_functions/index.rst
```diff
@@ -0,0 +1,112 @@
...
+
+sgnsqr_c4(x)
+~~~~~~~~~~~~
+
+
+This function is defined as:
+
+.. math::
+
+    \text{sgnsqr}(x) = \begin{cases}
```
blnicho:
```suggestion
    \text{sgnsqr_c4}(x) = \begin{cases}
```

doc/OnlineDocs/explanation/modeling_utils/external_functions/index.rst
```diff
@@ -0,0 +1,112 @@
...
+
+
+sgnsqrt_c4(x)
+~~~~~~~~~~~~
+
+This function is a signed square root approximation defined as:
+
+.. math::
+
+    \text{sgnsqr}(x) = \begin{cases}
```
blnicho:
```suggestion
    \text{sgnsqrt_c4}(x) = \begin{cases}
```

doc/OnlineDocs/explanation/modeling_utils/external_functions/index.rst
```diff
@@ -0,0 +1,112 @@
...
+        \text{sgn}(x)|x|^{0.5} & \text{if } |x| \ge 0.1 \\
+        \sum_{i=0}^{11} c_i x^i & \text{if } |x| < 0.1
+    \end{cases}
+
+This function is $C^4$ smooth.  The region :math:`-0.1 < x < 0.1` is replaced by an 11th order polynomial approximates that :math:`\text{sgn}(x)|x|^{0.5}`.  The approximate function has a well behaved derivatives at :math:`x=0`.  If you need to use this function with very small numbers and high accuracy is important, you can scale the argument up (e.g. :math:`\text{sgnsqrt\_c4}(sx)/s^{0.5}`).
```
blnicho:
```suggestion
This function is :math:`C^4` smooth.  The region :math:`-0.1 < x < 0.1` is replaced by an 11th order polynomial that approximates :math:`\text{sgn}(x)|x|^{0.5}`.  This function has well behaved derivatives at :math:`x=0`.  If you need to use this function with very small numbers and high accuracy is important, you can scale the argument up (e.g. :math:`\text{sgnsqrt_c4}(sx)/s^{0.5}`).
```

doc/OnlineDocs/explanation/modeling_utils/external_functions/index.rst
```diff
@@ -0,0 +1,112 @@
...
+This function is defined as:
+
+.. math::
+
+    \text{sgnsqr}(x) = \begin{cases}
+        \text{sgn}(x)x^2 & \text{if } |x| \ge 0.1 \\
+        \sum_{i=0}^{11} c_i x^i & \text{if } |x| < 0.1
+    \end{cases}
+
+This function is $C^4$ smooth.  The region :math:`-0.1 < x < 0.1` is replaced by an 11th order polynomial approximates that :math:`\text{sgn}(x)x^2`.  The approximate function has a well behaved derivatives at :math:`x=0`.  If you need to use this function with very small numbers and high accuracy is important, you can scale the argument up (e.g. :math:`\text{sgnsqr\_c4}(sx)/s^2`).
```
blnicho:
```suggestion
This function is :math:`C^4` smooth. The region :math:`-0.1 < x < 0.1` is replaced by an 11th order polynomial that approximates :math:`\text{sgn}(x)x^2`. This function has well behaved derivatives at :math:`x=0`. If you need to use this function with very small numbers and high accuracy is important, you can scale the argument up (e.g. :math:`\text{sgnsqr_c4}(sx)/s^2`).
```

doc/OnlineDocs/explanation/modeling_utils/external_functions/index.rst
```diff
@@ -0,0 +1,112 @@
...
+.. math::
+
+    \text{sgnsqr}(x) = \begin{cases}
+        \text{sgn}(x)|x|^{0.5} & \text{if } |x| \ge 0.1 \\
+        \sum_{i=0}^{11} c_i x^i & \text{if } |x| < 0.1
+    \end{cases}
+
+This function is $C^4$ smooth.  The region :math:`-0.1 < x < 0.1` is replaced by an 11th order polynomial approximates that :math:`\text{sgn}(x)|x|^{0.5}`.  The approximate function has a well behaved derivatives at :math:`x=0`.  If you need to use this function with very small numbers and high accuracy is important, you can scale the argument up (e.g. :math:`\text{sgnsqrt\_c4}(sx)/s^{0.5}`).
+
+These figures show the signed square root function compared to the smooth approximation sgnsqr_c4(x).
```
blnicho:
```suggestion
These figures show the signed square root function compared to the smooth approximation sgnsqrt_c4(x).
```

pyomo/contrib/external_functions/build.py
```diff
@@ -0,0 +1,32 @@
...
+
+import sys
+from pyomo.common.cmake_builder import build_cmake_project
+
+
+def build_external_functions(user_args=[], parallel=None):
+    return build_cmake_project(
+        targets=["src"],
+        package_name="external_functions",
+        description="Useful external function",
```
blnicho:
```suggestion
        description="Useful AMPL external functions",
```

pyomo/contrib/external_functions/src/CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+#  Copyright (c) 2008-2025
+#  National Technology and Engineering Solutions of Sandia, LLC
+#  Under the terms of Contract DE-NA0003525 with National Technology and
+#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain
+#  rights in this software.
+#  This software is distributed under the 3-clause BSD License.
+#  ___________________________________________________________________________
+
+cmake_minimum_required(VERSION 3.0...3.31)
+# This was developed against CMake 3.0, and appears to comply with 3.5
```
blnicho: CMake 3.5 is mentioned in the comment but the upper limit on the version in the previous line is 3.31. Is that a typo?

eslickj:
> @eslickj could you change the name of this contrib package (and associated references) from `contrib.external_functions` to `contrib.libaslfunctions`? We think `external_functions` is a bit too generic and want to make it clear that these are AMPL external functions.

@blnicho, thanks! I was trying to come up with a better name, and this sounds good to me.

pyomo/contrib/external_functions/src/CMakeLists.txt
eslickj: I'm not really sure.  Original source of this is the Pynumero cmake file and this is the same.  I'm not sure what to do here. As a whole, I guess, it implies this should work with cmake 3.5, but we aren't taking any chances.  The easiest thing would be to delete the comment.

jsiirola: It’s not a typo. 3.5 changed a bunch of things, and cmake started complaining loudly if you don't explicitly state you are compatible.  Basically, newer versions (through 3.31 - which was current when this change was made) only promise compatibility back to the syntax / libraries / behavior provided in 3.5. 

Also, “min…max” is not the correct way to read that. It is closer to “min…guaranteed compatible through”.  So, what we are saying is that this needs at least cmake 3, and should work with any cmake that is compatible back to version 3.31 (so cmake, if you are newer than 3.31, please behave like 3.31).

blnicho (approved):
Looks good pending passing tests

blnicho:
@eslickj your last commit seems to have fixed the pdflatex build in the linux/3.11/rtd job but it breaks the math/text rendering on ReadtheDocs. We should look for another formatting solution that works with both.

image

mrmundt:
> @eslickj your last commit seems to have fixed the pdflatex build in the linux/3.11/rtd job but it breaks the math/text rendering on ReadtheDocs. We should look for another formatting solution that works with both.
>

I fixed this by switching to `operatorname`.

Corrected math that no longer shows backslashes on RTD

eslickj:
@mrmundt, thanks!

blnicho:
I noticed that the tests are being skipped on the osx and linux GHA jobs even though the extension appears to build successfully. We should figure out why the tests aren't running before this gets merged.

mrmundt:
Let me try to summarize the weirdness that @blnicho and I just discovered. When this library is named `libaslfunctions` and it is built on Unix systems, it ends up with the name `liblibaslfunctions.*`. The double-`lib` either causes problems in `ctypes` or in our implementation of `find_library`. It's either not pre-pending enough libs when looking for the library (e.g., `libaslfunctions`), or it is prepending too many (e.g., `libliblibaslfunctions`).

If it does NOT start with `lib`, everything is fine. We aren't sure what the right path forward is here because it's not clear which item is the actual root cause.

I personally vote for, "We implement a standard that says we don't allow external functions that go into Pyomo to start with `lib`."

@jsiirola - thoughts?

jsiirola:
The project name generally shouldn't start with "lib" (as I believe cmake will add the "lib" on *NIX).  I would vote for `aslfunctions` (in general, libraries on Linux don't have underscores in the name, plus `libasl_functions_lib.so` is excessively redundant)

eslickj:
Ok, I renamed the library aslfunctions.  On Unix this becomes libaslfunctions.  ~~The find_library function strips the "lib" off on windows, so the right way to find it in general is find_library(libaslfunctions).  Hopefully I changed the names right, but I may have missed something.~~

I misunderstood the logic in find_library and got it backward. It should actually be ``find_library(aslfunctions)`` The "lib" gets removed on non-Windows platforms.  So, it works either way on Unix-like, but only ``find_library(aslfunctions)`` works on Windows.  I'll fix this now.

jsiirola (approved):

PR #3650, opened 2025-06-27
Add update_model utility to utils.py for updating suffix values in Pyomo.DoE

sscini:
## Summary/Motivation:
Currently, there is no way internally to update values for a suffix in a labeled model for parmest or DoE. This PR adds a new utility function to update suffix values when provided the model, the suffix, and the new values. 

- Addition of a utility function for Pyomo.DoE to update model suffix values.

### Legal Acknowledgement


sscini:
@adowling2 @djlaky @blnicho 
Hello, this PR is ready for initial review. Currently I have only added the new function, working on an example and tests currently to add as well. All feedback appreciated.

sscini:
This PR adds a new functionality that will be used in other PRs currently being drafted (#3575, #3639).
Early next week I plan to start adding:

Examples
- Parmest example for modifying unknown parameters and experiment outputs
- DoE example for modifying experiment inputs, experiment outputs, and measurement error

Testing
- Tests for new examples
- Test to show value is changed with each of 4 suffix names
- Test to show suffix itself maintains its original value
- Tests for all errors

sscini:
@mrmundt Absolutely, I'm working on a small example and adding checks to show it works and make sure it catches all the errors. Thank you!

sscini:
Main talking points for meeting 7/1:

- Main utility made
- Use strings or an Enum to provide function?

Finished example using unknown parameters, are more needed?
Working on tests as we speak

sscini:
Action items from July 1 meeting:
- Simplify input to just provide the model.suffix directly
- Modify example to reflect this
- Add tests

sscini:
@blnicho @mrmundt @djlaky @adowling2

**Context:** I developed the utility in doe/utils.py, and added tests to the doe/tests folder. 

**Problem:** DoE tests are failing with current example for experiment outputs and measurement error, says they are uninitialized.

**Proposed Solution:** I made an additional example (with a parmest model) to show it works with unknown parameters, experiment outputs, and measurement_error.

**Question:** Would it be better for me to keep everything in contrib/doe, or can I add testing and an example in parmest using a function from doe? Trying to minimize new imports needed. Thoughts? 

**TLDR:** Does it matter if I add items to doe and parmest contrib folders in same PR? Imports will overlap

sscini:
Note from 7/8 meeting, would make more sense in parmest by default, where experiment class lives. Moving to parmest utility and fixing tests.

sscini:
@adowling2 @djlaky @mrmundt @blnicho 
This is ready for final review. Thank you.

pyomo/contrib/parmest/utils/model_utils.py
```diff
@@ -201,3 +201,48 @@ def convert_params_to_vars(model, param_names=None, fix_vars=False):
...
+
+    # Add a check for measurement error suffix
+    is_me_err = suffix_name == "measurement_error"
+
+    # Iterate through the items in the suffix and update their values
+    # Note: items are tuples of (component, suffix_value)
+    for (comp, _), new_val in zip(items, values):
+
+        # update the component value
+        # Measurement error is only stored in the suffix, not in the model, so it needs to be set directly
```
djlaky: Can you make this comment two lines so it doesn't wrap?

pyomo/contrib/parmest/tests/test_examples.py
```diff
@@ -194,6 +194,13 @@ def test_datarec_example(self):
 
         datarec_example.main()
 
+    def test_reactor_design_updatesuffix(self):
```
djlaky: Just to be thorough, can you add a test for the new code introduced to DoE as well, please?

pyomo/contrib/parmest/tests/test_utils.py
```diff
@@ -9,14 +9,37 @@
...
+
+from pyomo.common.fileutils import this_file_dir
 import pyomo.common.unittest as unittest
+
 import pyomo.contrib.parmest.parmest as parmest
 from pyomo.opt import SolverFactory
 
-ipopt_available = SolverFactory("ipopt").available()
+from pyomo.contrib.parmest.utils.model_utils import update_model_from_suffix
+from pyomo.contrib.doe.examples.reactor_example import (
```
djlaky, line 31: Should we avoid importing from a different contributed package here? I'm not sure what the convention is @mrmundt

If so, we probably should be using a parmest-specific example to do the testing.

pyomo/contrib/parmest/utils/model_utils.py
```diff
@@ -201,3 +201,48 @@ def convert_params_to_vars(model, param_names=None, fix_vars=False):
     # solver.solve(model)
 
     return model
+
+
+# Adding utility to update parameter values in a model based on the suffix
+def update_model_from_suffix(suffix_obj: pyo.Suffix, values):
+    """
+    Overwrite each variable/parameter referenced by ``suffix_obj`` with the
+    corresponding value in ``values``.
```
djlaky: Might want to add a comment about assuming insertion order from when the suffix was created.

djlaky (changes requested):
Just a few general comments. One that @mrmundt can comment on if we want to keep the DoE library away from anything in parmest for testing.

pyomo/contrib/parmest/tests/test_examples.py
sscini: Will do. That would go in doe/tests/test_doe_build.py correct? Currently the next comment area is where I used the doe example and tested it with measurement error since that is the only example currently that contains measurement error. If preferred, I can add measurement error to a parmest example

sscini:
@blnicho @djlaky During meeting today, let's also finalize changes needed to merge this PR, which we believe will be integrated into the multistart design. Thanks!

pyomo/contrib/doe/examples/reactor_updatesuffix.py
```diff
@@ -0,0 +1,67 @@
...
+    reactor_model.pprint()
+    # Update the model to change the values of the desired component
+    # Here we will update the unknown parameters of the reactor model
+    example_suffix = "measurement_error"
+    suffix_obj = reactor_model.measurement_error
+    me_vars = list(suffix_obj.keys())  # components
+    orig_vals = np.array([suffix_obj[v] for v in me_vars])
+
+    # Original values
+    print("Original σ values:", orig_vals)
```
mrmundt: It's probably better to type out `sigma`, not use the symbol.

pyomo/contrib/doe/examples/reactor_updatesuffix.py
```diff
@@ -0,0 +1,67 @@
...
+    # Update the suffix with new values
+    new_vals = orig_vals + 1
+    # Here we are updating the values of the unknown parameters
+    # You must know the length of the list and order of the suffix items to update them correctly
+    update_model_from_suffix(suffix_obj, new_vals)
+
+    # Updated values
+    print("Updated σ values :", [suffix_obj[v] for v in me_vars])
```
mrmundt: It's probably better to type out `sigma`, not use the symbol.

pyomo/contrib/parmest/tests/test_utils.py
mrmundt: That's not an issue as long as there aren't circular imports introduced. We do not, however, import from, e.g., idaes, because it requires pyomo. _That_ is a circular import nightmare.

sscini:
@mrmundt Thank you for your comments. With these changes would it be approved or is Bethany still planning to suggest how organize the files? Should I add the additional test Dan recommended earlier?

mrmundt:
> @mrmundt Thank you for your comments. With these changes would it be approved or is Bethany still planning to suggest how organize the files? Should I add the additional test Dan recommended earlier?

I assume @blnicho still has suggestions, and yes, please do add Dan's recommendations.

blnicho:
@sscini I'm fine with the current file organization. I will have a few more minor review comments for you in a couple hours.

sscini:
@mrmundt Applied changes from your comments. Assuming checks pass, once @blnicho adds final comments will request final review. Thank you!

pyomo/contrib/doe/examples/updatesuffix_doe_example.py
```diff
@@ -0,0 +1,67 @@
...
+
+    # Call the experiment's model using get_labeled_model
+    reactor_model = experiment.get_labeled_model()
+
+    # Show the model
+    reactor_model.pprint()
+    # Update the model to change the values of the desired component
+    # Here we will update the unknown parameters of the reactor model
+    example_suffix = "measurement_error"
+    suffix_obj = reactor_model.measurement_error
```
blnicho: The comment doesn't seem to match the implementation. It says "unknown parameters" are being updated but the implementation is updating "measurement_error"

pyomo/contrib/doe/examples/updatesuffix_doe_example.py
```diff
@@ -0,0 +1,67 @@
...
+    experiment = ReactorExperiment(data=data_ex, nfe=10, ncp=3)
+
+    # Call the experiment's model using get_labeled_model
+    reactor_model = experiment.get_labeled_model()
+
+    # Show the model
+    reactor_model.pprint()
+    # Update the model to change the values of the desired component
+    # Here we will update the unknown parameters of the reactor model
+    example_suffix = "measurement_error"
```
blnicho: Can this line be removed? It doesn't look like `example_suffix` is used anywhere.

pyomo/contrib/doe/examples/updatesuffix_doe_example.py
```diff
@@ -0,0 +1,67 @@
...
+    suffix_obj = reactor_model.measurement_error
+    me_vars = list(suffix_obj.keys())  # components
+    orig_vals = np.array([suffix_obj[v] for v in me_vars])
```
blnicho:
```suggestion
    orig_vals = np.array(list(suffix_obj.values()))
```
Suggestion for simplifying this a bit

pyomo/contrib/doe/examples/updatesuffix_doe_example.py
```diff
@@ -0,0 +1,67 @@
...
+
+    # Original values
+    print("Original sigma values:", orig_vals)
+    # Update the suffix with new values
+    new_vals = orig_vals + 1
+    # Here we are updating the values of the unknown parameters
```
blnicho: Again this comment seems out-of-date. It's referencing "unknown parameters" which isn't the suffix being updated.

pyomo/contrib/doe/examples/updatesuffix_doe_example.py
```diff
@@ -0,0 +1,67 @@
...
+    reactor_model.pprint()
+    # Update the model to change the values of the desired component
+    # Here we will update the unknown parameters of the reactor model
+    example_suffix = "measurement_error"
+    suffix_obj = reactor_model.measurement_error
+    me_vars = list(suffix_obj.keys())  # components
+    orig_vals = np.array([suffix_obj[v] for v in me_vars])
+
+    # Original values
+    print("Original sigma values:", orig_vals)
```
blnicho: Using the `display` or `pprint` method on the suffix might be a little more clear here.

    print("Original sigma values")
    print("---------------------")
    suffix_obj.display()

pyomo/contrib/doe/examples/updatesuffix_doe_example.py
```diff
@@ -0,0 +1,67 @@
...
+    # Update the suffix with new values
+    new_vals = orig_vals + 1
+    # Here we are updating the values of the unknown parameters
+    # You must know the length of the list and order of the suffix items to update them correctly
+    update_model_from_suffix(suffix_obj, new_vals)
+
+    # Updated values
+    print("Updated sigma values :", [suffix_obj[v] for v in me_vars])
```
blnicho:
```suggestion
    print("Updated sigma values :", list(suffix_obj.values()))
```
Using `values` is a little more concise but I also have the same comment here about using `display` or `pprint` to provide a little more context beyond a list of numbers.

pyomo/contrib/doe/tests/test_doe_build.py
```diff
@@ -474,6 +474,15 @@ def test_generate_blocks_without_model(self):
...
 
+    def test_reactor_update_suffix_items(self):
+        """Test the reactor example with updating suffix items."""
+        from pyomo.contrib.doe.examples.updatesuffix_doe_example import (
+            run_reactor_update_suffix_items,
+        )
+
+        # Run the reactor update suffix items example
+        run_reactor_update_suffix_items()
+
```
blnicho: A test should have an assert statement. Maybe you could modify the `run_reactor_update_suffix_items` function to return the original and updated suffix values and/or the model/suffix and verify that the values are correct.

pyomo/contrib/parmest/tests/test_examples.py
```diff
@@ -194,6 +194,11 @@ def test_datarec_example(self):
...
```
blnicho: This test needs an assert statement.

pyomo/contrib/parmest/tests/test_utils.py
```diff
@@ -60,6 +83,120 @@ def test_convert_param_to_var(self):
...
             self.assertTrue(c in m_vars.unknown_parameters)
 
+    def test_update_model_from_suffix_unknown_parameters(self):
+        experiment = FullReactorExperiment(data_ex, 10, 3)
+        test_model = experiment.get_labeled_model()
+
+        suffix_obj = test_model.unknown_parameters  # a Suffix
+        var_list = list(suffix_obj.keys())  # components only
+        orig_var_vals = np.array([pyo.value(v) for v in var_list])  # numeric var values
+        orig_suffix_val = np.array([tag for _, tag in suffix_obj.items()])
```
blnicho:
```suggestion
        orig_suffix_val = np.array(list(suffix_obj.values()))
```

pyomo/contrib/parmest/tests/test_utils.py
```diff
@@ -60,6 +83,120 @@ def test_convert_param_to_var(self):
...
+        new_vals = orig_var_vals + 10
+
+        # Update the model from the suffix
+        update_model_from_suffix(suffix_obj, new_vals)
+
+        # ── Check results ────────────────────────────────────────────────────
+        new_var_vals = np.array([pyo.value(v) for v in var_list])
+        new_suffix_val = np.array([tag for _, tag in suffix_obj.items()])
```
blnicho:
```suggestion
        new_suffix_val = np.array(list(suffix_obj.values()))
```

pyomo/contrib/parmest/tests/test_utils.py
```diff
@@ -60,6 +83,120 @@ def test_convert_param_to_var(self):
...
+        m = experiment.get_labeled_model()
+        # Only ONE new value for TWO suffix items ➜ should raise
+        with self.assertRaisesRegex(
+            ValueError, "values length does not match suffix length"
+        ):
+            update_model_from_suffix(m.unknown_parameters, [42])
+
+    def test_update_model_from_suffix_unsupported_component(self):
+        experiment = FullReactorExperiment(data_ex, 10, 3)
+        m = experiment.get_labeled_model()
```
blnicho:
```suggestion
        m = pyo.ConcreteModel()
```
I don't think you need the experiment model for this test.

pyomo/contrib/parmest/tests/test_utils.py
```diff
@@ -60,6 +83,120 @@ def test_convert_param_to_var(self):
...
+
+    def test_update_model_from_suffix_length_mismatch(self):
+
+        experiment = FullReactorExperiment(data_ex, 10, 3)
+        m = experiment.get_labeled_model()
+        # Only ONE new value for TWO suffix items ➜ should raise
+        with self.assertRaisesRegex(
+            ValueError, "values length does not match suffix length"
+        ):
+            update_model_from_suffix(m.unknown_parameters, [42])
```
blnicho: You don't need a full experiment model for this test. I would recommend testing this exception using a basic Pyomo model instead.

pyomo/contrib/parmest/utils/model_utils.py
```diff
@@ -201,3 +201,51 @@ def convert_params_to_vars(model, param_names=None, fix_vars=False):
...
+    created.
+
+    Parameters
+    ----------
+    suffix_obj : pyomo.core.base.suffix.Suffix
+        The suffix whose *keys* are the components you want to update.
+        Call like ``update_from_suffix(model.unknown_parameters, vals)``.
+    values : iterable of numbers
+        New numerical values for the components referenced by the suffix.
+        Must be the same length as ``suffix_obj``.
```
blnicho, line 220: This docstring doesn't mention that the `measurement_error` suffix has special handling

pyomo/contrib/parmest/tests/test_utils.py
```diff
@@ -60,6 +83,120 @@ def test_convert_param_to_var(self):
...
+        suffix_obj = test_model.measurement_error  # a Suffix
+        var_list = list(suffix_obj.keys())  # components
+        orig_var_vals = np.array([suffix_obj[v] for v in var_list])
+        new_vals = orig_var_vals + 0.5
+        # Update the model from the suffix
+        update_model_from_suffix(suffix_obj, new_vals)
+        # ── Check results ────────────────────────────────────────────────────
+        new_var_vals = np.array([suffix_obj[v] for v in var_list])
+        # (1) Variables have been overwritten with `new_vals`
+        self.assertTrue(np.allclose(new_var_vals, new_vals))
```
blnicho: It doesn't seem like these tests are actually testing different aspects of your utility function, they are just trying out different suffix names. You're already testing the interaction between your utility function and an Experiment class in `test_examples.py` so I would recommend moving one set of these assert statements to that test and getting rid of the rest of these Experiment tests here. Your utility function doesn't rely on or require an Experiment class so most of the unit tests for it should only require basic Pyomo models. You have special handling for the `measurement_error` suffix but that can be tested without an Experiment class. To rigorously test your utility function you should think about edge cases in how the model/suffix might be structured. What happens if the Suffix has indexed components for keys? What if the Suffix has a mixture of component objects and component data objects? What if the Suffix has a mixture of component types, some that you support and some that you don't?

pyomo/contrib/parmest/utils/model_utils.py
```diff
@@ -201,3 +201,51 @@ def convert_params_to_vars(model, param_names=None, fix_vars=False):
     # solver.solve(model)
 
     return model
+
+
+# Adding utility to update parameter values in a model based on the suffix
```
blnicho:
```suggestion
```
This comment is captured in the docstring

pyomo/contrib/parmest/utils/model_utils.py
```diff
@@ -201,3 +201,51 @@ def convert_params_to_vars(model, param_names=None, fix_vars=False):
...
+        The suffix whose *keys* are the components you want to update.
+        Call like ``update_from_suffix(model.unknown_parameters, vals)``.
+    values : iterable of numbers
+        New numerical values for the components referenced by the suffix.
+        Must be the same length as ``suffix_obj``.
+    """
+    # Check that the length of values matches the suffix length
+    items = list(suffix_obj.items())
+    if len(items) != len(values):
+        raise ValueError("values length does not match suffix length")
```
blnicho: What is the expected behavior when indexed components are used as keys in the suffix?

pyomo/contrib/parmest/utils/model_utils.py
```diff
@@ -201,3 +201,51 @@ def convert_params_to_vars(model, param_names=None, fix_vars=False):
...
+
+    # Robust way to get the component’s own name
+    suffix_name = getattr(suffix_obj, "local_name", suffix_obj.name)
+
+    # Add a check for measurement error suffix
+    is_me_err = suffix_name == "measurement_error"
```
blnicho:
```suggestion
    # Add a check for measurement error suffix
    is_me_err = "measurement_error" in suffix_obj.name
```
What do you think about this as a more concise alternative?

pyomo/contrib/parmest/utils/model_utils.py
```diff
@@ -201,3 +201,51 @@ def convert_params_to_vars(model, param_names=None, fix_vars=False):
...
+    ----------
+    suffix_obj : pyomo.core.base.suffix.Suffix
+        The suffix whose *keys* are the components you want to update.
+        Call like ``update_from_suffix(model.unknown_parameters, vals)``.
+    values : iterable of numbers
+        New numerical values for the components referenced by the suffix.
+        Must be the same length as ``suffix_obj``.
+    """
+    # Check that the length of values matches the suffix length
+    items = list(suffix_obj.items())
```
blnicho:
```suggestion
    items = list(suffix_obj.keys())
```
Using `keys` instead of items here would simplify the for-loop below a bit.

pyomo/contrib/parmest/utils/model_utils.py
```diff
@@ -201,3 +201,51 @@ def convert_params_to_vars(model, param_names=None, fix_vars=False):
...
+            suffix_obj[comp] = float(new_val)
+        # Check if the component is a VarData or ParamData
+        elif isinstance(comp, (VarData, ParamData)):
+            comp.set_value(new_val)
+
+        # If the component is not a VarData or ParamData, raise an error
+        else:
+            raise TypeError(
+                f"Unsupported component type {type(comp)}; expected VarData or ParamData."
+            )
```
blnicho: If the suffix contains a mixture of component types you could end up with an edge case where some values are updated and some are not if an incompatible type is encountered in the middle. You might want to do the type validation in a separate loop before updating any values just to be safe.

sscini: @blnicho Thank you very much for your detailed comments. 

For this case, the issue would be in the labeling of the model, no? From my understanding, that would require the user labeling the model with something outside of the pyomo object? I need to look into what would not be considered a VarData or ParamData. Either way, definitely important to address and then update the testing as you describe.

@adowling2 It would be helpful if we discuss this later this morning.

pyomo/contrib/parmest/utils/model_utils.py
sscini: The expected behavior is that the user will know the length of the item and adjust the list provided accordingly, OR they print the suffix original values as shown as the example and modify the list as needed. Please let me know if this does not answer your question.

sscini:
@mrmundt @blnicho Still working on final testing but I noticed a problem?
I synced my fork and now it is showing the lint and style check is failing. Ran black on doe and parmest to confirm and all files were unchanged. Check passed prior to syncing. Could there be a styling issue from a recent merge or could it be something in my code other than black?

Thank you.

mrmundt:
> @mrmundt @blnicho Still working on final testing but I noticed a problem? I synced my fork and now it is showing the lint and style check is failing. Ran black on doe and parmest to confirm and all files were unchanged. Check passed prior to syncing. Could there be a styling issue from a recent merge or could it be something in my code other than black?
> 
> Thank you.

@sscini - I think you ran it right in between when they were cutting a mini-patch release for the spell checker (they released 1.35.0 this morning, and then it had some flaws, so they almost immediately thereafter released 1.35.1). I just re-ran it which is hopefully good enough, but if not, I suspect that this is a "wait until whatever background processes need to happen" fix.

pyomo/contrib/doe/examples/updatesuffix_doe_example.py
sscini: If I use display or pprint, they both show:
```
Original sigma values
----------------------
unknown_parameters : Direction=LOCAL, Datatype=FLOAT
    Key : Value
     k1 :    k1
     k2 :    k2
     k3 :    k3
```

The original print method shows the numeric values. Should I leave it as is / hard-code improved formatting? Or would you expect pprint/display to show something else?

pyomo/contrib/doe/examples/updatesuffix_doe_example.py
sscini: When running testing, noticed an issue from this change. It changes the values to non-numeric values showing the variable names, and an error happens when I try to add 1 saying "+" does not work between a float and a ComponentUID. Any suggestions or may I leave in original state?

pyomo/contrib/doe/examples/updatesuffix_doe_example.py
sscini: Clarification. This works well in the doe example. Which is where you requested this update. I tried to apply it to the parmest example as well, and it did not work, instead showing above because I am updating the variable not the suffix value. I am leaving original in the parmest example, but the .display() in the doe example. Please let me know if other changes are needed.

sscini:
@blnicho @mrmundt @djlaky 
Thank you for your patience. Ready for final review pending final checks.

pyomo/contrib/doe/examples/updatesuffix_doe_example.py
```diff
@@ -0,0 +1,72 @@
+#  ___________________________________________________________________________
```
mrmundt: I fully accept all grumpiness about the two comments I am about to make. In terms of screen reader accessibility, `updatesuffix` isn't the best. There should be an underscore between them: `update_suffix`. I recommend changing the filename to `update_suffix_doe_example.py`

pyomo/contrib/parmest/examples/reactor_design/updatesuffix_example.py
```diff
@@ -0,0 +1,67 @@
+#  ___________________________________________________________________________
```
mrmundt: Second grump-inducing comment. In terms of screen reader accessibility, `updatesuffix` isn't the best. There should be an underscore between them: `update_suffix`. I recommend changing the filename to `update_suffix_example.py`

pyomo/contrib/parmest/examples/reactor_design/updatesuffix_example.py
```diff
@@ -0,0 +1,67 @@
...
+import pyomo.environ as pyo
+from pyomo.contrib.parmest.utils.model_utils import update_model_from_suffix
+
+
+def main():
+    # Read in file
+    # Read in data
+    file_dirname = dirname(abspath(str(__file__)))
+    file_name = abspath(join(file_dirname, "reactor_data.csv"))
+    data = pd.read_csv(file_name)
```
mrmundt: This isn't a deal-breaker, but the inconsistency with how the files are loaded between the two examples makes me squint. In `update_suffix_doe_example`, you do:
```python
    # Read in file
    DATA_DIR = pathlib.Path(__file__).parent
    file_path = DATA_DIR / "result.json"
```

Here you do the more "classic" `dirname`, `abspath`, `join` routine. It just kind of looks weird, if that makes sense?

pyomo/contrib/parmest/examples/reactor_design/updatesuffix_example.py
```diff
@@ -0,0 +1,67 @@
...
+import pyomo.contrib.parmest.parmest as parmest
+from pyomo.contrib.parmest.examples.reactor_design.reactor_design import (
+    ReactorDesignExperiment,
+)
+
+import pyomo.environ as pyo
+from pyomo.contrib.parmest.utils.model_utils import update_model_from_suffix
+
+
+def main():
```
mrmundt: One example driver is called `main`, and the other one is called `run_reactor_update_suffix_items`. It's inconsistent. (Similar to other comment, not a deal-breaker, just a bit odd.)

pyomo/contrib/parmest/examples/reactor_design/updatesuffix_example.py
sscini: Thank you for catching this. I believe I inherited the read in method from other examples in each contrib. To confirm, you would prefer to see both using the "classic" method in the parmest reactor_design example?

pyomo/contrib/parmest/examples/reactor_design/updatesuffix_example.py
sscini: Similar to before, I was trying to follow conventions seen in each of the contrib libraries. All parmest examples have the driver called "main", as seen in the test_examples file, and the other doe example files have the driver as a descriptive name. Would your preference be for them both to be main?

pyomo/contrib/parmest/examples/reactor_design/updatesuffix_example.py
mrmundt: I have no strong preference - pick whichever one brings you joy.

pyomo/contrib/parmest/examples/reactor_design/updatesuffix_example.py
mrmundt: I have no strong preference - pick whichever one brings you joy.

mrmundt (approved):
Thanks for doing all this work!

sscini:
@mrmundt @blnicho Please let me know if there are any tips to fix the failing tests, but I ran the tests on my machine for parmest and doe, all passed. And it looks like it might be related to something else

mrmundt:
> @mrmundt @blnicho Please let me know if there are any tips to fix the failing tests, but I ran the tests on my machine for parmest and doe, all passed. And it looks like it might be related to something else

Has nothing whatsoever to do with your changes. An upstream dependency released a new version that introduced this. We're looking into it.

pyomo/contrib/doe/examples/updatesuffix_doe_example.py
blnicho: Thanks for testing this. I didn't realize the values in the Suffix were components and not numeric values. You can ignore my suggestion.

pyomo/contrib/parmest/tests/test_utils.py
```diff
@@ -60,6 +83,125 @@ def test_convert_param_to_var(self):
...
+        m.unknown_parameters[m.x] = 0.0  # tag a Var
+        m.unknown_parameters[m.y] = bad_value  # tag a Var with a bad value
+        # Attempt to update with a list of mixed types
+        # This should raise an error because the suffix expects VarData or ParamData
+
+        with self.assertRaisesRegex(
+            ValueError, f"could not convert string to float: '{bad_value}'"
+        ):
+            # Attempt to update with a list of different length
+            update_model_from_suffix(m.unknown_parameters, [42, bad_value])
```
blnicho:
```suggestion
        # This should raise an error because this utility only allows numeric values
        # in the model to be updated.

        with self.assertRaisesRegex(
            ValueError, f"could not convert string to float: '{bad_value}'"
        ):
            # Attempt to update with a non-numeric value
            update_model_from_suffix(m.unknown_parameters, [42, bad_value])
```

pyomo/contrib/parmest/utils/model_utils.py
```diff
@@ -201,3 +201,46 @@ def convert_params_to_vars(model, param_names=None, fix_vars=False):
...
+        The suffix whose *keys* are the components you want to update.
+        Call like ``update_from_suffix(model.unknown_parameters, vals)``.
+    values : iterable of numbers
+        New numerical values for the components referenced by the suffix.
+        Must be the same length as ``suffix_obj``.
+
+    Notes
+    -----
+    Measurement error is a special case: instead of updating the value of the
+    keys (variables/parameters), it updates the value stored in the suffix itself.
```
blnicho:
```suggestion
    The measurement_error suffix is a special case: instead of updating the value of the
    keys (variables/parameters), it updates the value stored in the suffix itself.
```

pyomo/contrib/parmest/utils/model_utils.py
```diff
@@ -201,3 +201,46 @@ def convert_params_to_vars(model, param_names=None, fix_vars=False):
...
+        if not isinstance(comp, (VarData, ParamData)):
+            raise TypeError(
+                f"Unsupported component type {type(comp)}; expected VarData or ParamData."
+            )
+    # Second loop: adjust the values
+    for (comp, _), new_val in zip(items, values):
+        if is_me_err:
+            suffix_obj[comp] = float(new_val)
+        else:
+            comp.set_value(float(new_val))
```
blnicho:
```suggestion
    comps = list(suffix_obj.keys())
    if len(comps) != len(values):
        raise ValueError("values length does not match suffix length")

    # Add a check for measurement error suffix
    is_me_err = "measurement_error" in suffix_obj.name
    # Iterate through the keys in the suffix and update their values
    # First loop: check all values are the right type
    for comp in comps:
        if not isinstance(comp, (VarData, ParamData)):
            raise TypeError(
                f"Unsupported component type {type(comp)}; expected VarData or ParamData."
            )
    # Second loop: adjust the values
    for comp, new_val in zip(comps, values):
        if is_me_err:
            suffix_obj[comp] = float(new_val)
        else:
            comp.set_value(float(new_val))
```

blnicho (approved):

PR #3651, opened 2025-06-27
Refactor linear/quadratic expression compilers

jsiirola:
## Summary/Motivation:
The linear / quadratic / parameterized / templated expression walkers and compilers contain a large amount of repeated code.  This PR refactors the walkers so that the bulk of repeated code can be deleted.  The key changes are:
- Move the "fixed" argument handlers from the parameterized walkers into the base (linear/quadratic) walkers
- Don't assume that coefficients (that are floats in the base walkers and potentially expressions in other walkers) can be implicitly be case to bool or compared to 1.  This change imparts a small performance hit (~1% for the LP writer; no noticeable effect on the other writers).  

Other changes in this PR
- Parameterized walkers are simple enough that they are now together in a single "parameterized.py" module
- A slight change to how we process bounds in constraints and variables makes the code simpler and slightly more efficient (or at least no more inefficient)
- The introduction of TemplateDataMixin classes for Constraint and Objectives allow us to templatize scalar components in addition to indexed components
- Simplify the ScalarObjective class, and introduce an AbstractScalarObjective class (with the use of the `@disable_methods` decorator).  This follows the pattern used elsewhere (e.g., in Constriant)
- Fix a bug in how we pass InvalidNumbers through unary operators
- Test the LinearStandardForm compiler on regular and templatized models
- Fix a bug in the TemplateVarRecorder when the user provides a (partial) variable ordering.

While this PR will slow down the LP writer a little (1-2%), it resolves enough issues in the Parameterized and Templatized walkers that I think we should adopt it.  There are a number of future developments in the works (that build on this PR) that should more than make up for this performance hit [moving the LP writer to leverage the standard form compiler to eliminate the need for sorting in the lp_writer and the use of dicts in the compiler for coefficient deduplication).

- (see above)

### Legal Acknowledgement


pyomo/repn/linear.py
```diff
@@ -789,48 +910,58 @@ def exitNode(self, node, data):
...
+        mult_flag = ans.multiplier_flag(ans.multiplier)
+        if mult_flag == 1:
+            # mult is identity: only thing to do is filter out zero coefficients
+            self._filter_zeros(ans)
             return ans
-        ans = self.Result()
-        assert result[0] is _CONSTANT
-        ans.constant = result[1]
+        elif not mult_flag:
+            # the mulltiplier has cleared out the entire expression.
```
mrmundt: Reported missed typo upstream.
```suggestion
            # the multiplier has cleared out the entire expression.
```

pyomo/repn/parameterized.py
```diff
@@ -0,0 +1,170 @@
...
+            return val
+        return 2  # something not 0 or 1
+
+    @staticmethod
+    def multiplier_flag(val):
+        if val.__class__ in native_numeric_types:
+            if not val:
+                return 2
+            return val
+        return 2  # something not 0 or 1
```
mrmundt, line 50: Should these both return 2?

pyomo/repn/quadratic.py
```diff
@@ -130,40 +144,99 @@ def append(self, other):
...
+            if other.quadratic:
+                for vid, coef in other.quadratic.items():
+                    if coef != coef:
+                        if not self.quadratic:
+                            self.quadratic = {}
+                        if vid in self.quadratic:
+                            self.quadratic[vid] += mult * coef
+                        else:
+                            self.quadratic[vid] = mult * coef
+        else:
```
mrmundt: There are a _lot_ of missed lines here w.r.t. tests. Might be worth adding at least one test.

pyomo/repn/tests/test_parameterized_quadratic.py
```diff
@@ -976,7 +1016,7 @@ def test_sum_nonlinear_custom_multiplier(self):
         self.assertEqual(repn.multiplier, 1)
         assertExpressionsEqual(self, repn.constant, 2 + 2 * (m.y + m.y**2))
         self.assertEqual(repn.linear, {})
-        self.assertIsNone(repn.quadratic)
+        self.assertEqual(repn.quadratic, None)
```
mrmundt, line 1019: Why not just use `assertIsNone`?

pyomo/repn/tests/test_parameterized_quadratic.py
```diff
@@ -955,7 +995,7 @@ def test_factor_multiplier_simplify_coefficients(self):
         self.assertEqual(cfg.var_order, {id(m.x): 0, id(m.z): 1})
         self.assertEqual(repn.multiplier, 1)
         self.assertIsNone(repn.nonlinear)
-        self.assertEqual(repn.quadratic, {})
+        self.assertEqual(repn.quadratic, None)
```
mrmundt, line 998: Why not just use `assertIsNone`?

pyomo/repn/tests/test_parameterized_quadratic.py
```diff
@@ -936,7 +976,7 @@ def test_finalize_simplify_coefficients(self):
         self.assertEqual(repn.multiplier, 1)
         assertExpressionsEqual(self, repn.constant, 2 * m.y**2)
         self.assertEqual(repn.linear, {id(m.z): -1})
-        self.assertEqual(repn.quadratic, {})
+        self.assertEqual(repn.quadratic, None)
```
mrmundt, line 979: Why not just use `assertIsNone`?

pyomo/repn/parameterized.py
jsiirola: Sure.  It really doesn't matter.  The point (as indicated by the comment) is that it is just anything other 0 or 1.  Because "2" is not unique (the `val` could actually be `float(2)`), there really isn't a driver to differentiate between 0, 2, and non-native numeric types.

pyomo/repn/tests/test_parameterized_quadratic.py
jsiirola: This was part of leftover debugging.  The error you get when `repn.quadratic` is not None is remarkable unhelpful, whereas using `assertEqual` gives an error message that says *what* `repn.quadratic` actually is.

blnicho (approved):

pyomo/core/base/constraint.py
```diff
@@ -552,17 +552,34 @@ def template_expr(self):
         return self._expr
 
     def set_value(self, expr):
-        self.__class__ = ConstraintData
+        self.__class__ = self.__class__.__mro__[
+            self.__class__.__mro__.index(TemplateDataMixin) + 1
+        ]
```
emma58, line 564: Could you perhaps put a comment in explaining this--that you're getting the data class that this is pretending to be?

pyomo/core/base/objective.py
```diff
@@ -249,11 +256,11 @@ class Objective(ActiveIndexedComponent):
 
     def __new__(cls, *args, **kwds):
         if cls != Objective:
-            return super(Objective, cls).__new__(cls)
+            return super().__new__(cls)
         if not args or (args[0] is UnindexedComponent_set and len(args) == 1):
-            return ScalarObjective.__new__(ScalarObjective)
+            return super().__new__(AbstractScalarObjective)
```
emma58, line 278: Why is this now an `AbstractScalarObjective` instead of just a `ScalarObective`? Maybe the same question or maybe a different one: Why is this special for Objective and not happening for other components?

pyomo/core/base/objective.py
```diff
@@ -165,7 +166,27 @@ class _GeneralObjectiveData(metaclass=RenamedClass):
...
+        self.set_value(self.parent_component().rule(self.parent_block(), self.index()))
+        return self._args_
+
+    def template_expr(self):
+        return self._args_
+
+    def set_value(self, expr):
+        self.__class__ = self.__class__.__mro__[
+            self.__class__.__mro__.index(TemplateDataMixin) + 1
+        ]
```
emma58, line 192: The same comment as above might help future us here! :)

pyomo/repn/linear.py
```diff
@@ -161,32 +181,60 @@ def append(self, other):
         # change). Omitting the assertion for efficiency.
         # assert self.multiplier == 1
         _type, other = other
-        if _type is _CONSTANT:
+        if _type <= _FIXED:
```
emma58: Is this a good idea? Are we going to remember the implications of where things are in the Enum? Maybe could you at least add a comment here for what you are catching/not catching?

pyomo/repn/linear.py
```diff
@@ -161,32 +181,60 @@ def append(self, other):
...
-                self.nonlinear = nl
+                self.nonlinear = mult * other.nonlinear
             else:
-                self.nonlinear += nl
+                self.nonlinear += mult * other.nonlinear
 
 
 def to_expression(visitor, arg):
-    if arg[0] is _CONSTANT:
+    if arg[0] <= _VARIABLE:
```
emma58: Same comment here: This is harder to read, but maybe add a comment to clarify.

pyomo/repn/linear.py
```diff
@@ -789,48 +914,60 @@ def exitNode(self, node, data):
...
+            if ans.constant.__class__ is InvalidNumber or any(
+                c.__class__ is InvalidNumber for c in ans.linear.values()
+            ):
+                deprecation_warning(
+                    f"Encountered {ans.multiplier}*nan in expression tree.  "
+                    "Mapping the NaN result to 0 for compatibility "
+                    "with the lp_v1 writer.  In the future, this NaN "
+                    "will be preserved/emitted to comply with IEEE-754.",
+                    version='6.6.0',
+                )
```
emma58, line 942: Do we have removing this on the Pyomo 7 list yet, by the way?

pyomo/repn/parameterized.py
```diff
@@ -0,0 +1,170 @@
...
+            visitor.var_recorder.add(child)
+        ans = visitor.Result()
+        ans.linear[_id] = 1
+        return False, (ExprType.LINEAR, ans)
+
+
+def _handle_product_constant_constant(visitor, node, arg1, arg2):
+    # [ESJ 5/22/24]: Overriding this handler to exclude the deprecation path for
+    # 0 * nan. It doesn't need overridden when that deprecation path goes away.
+    return _CONSTANT, arg1[1] * arg2[1]
```
emma58, line 102: Does this mean we are bypassing the deprecation path for templatized things too, or are they not using this walker? And if so, do we care?

emma58 (commented):
I like this very much! A few questions in the comments, but I think this looks good (and I've actually been using it for awhile because I raided the cookie jar...)

pyomo/core/base/objective.py
jsiirola: This is consistent with other AbstractScalar objects (in particular AbstractScalarConstraint)

pyomo/repn/linear.py
jsiirola: I am pretty sure we don't document that old behavior in The Book.  We can probably remove this before Pyomo 7.

pyomo/repn/parameterized.py
jsiirola: Currently, the Templatized walker does not use the Parameterized walker, so No.  The idea was that it should behave the same as the regular linear walker.

pyomo/repn/linear.py
jsiirola: Added a comment.

pyomo/repn/linear.py
emma58: :D

pyomo/repn/parameterized.py
emma58: That makes sense!

PR #3656, opened 2025-07-08
Introduce Retries into `download-extensions`

mrmundt:
## Summary/Motivation:
I'm personally losing my mind w.r.t. `download-extensions` because `gjh` is somewhat unreliable lately when downloading (but then will sometimes be fine if you try again). This introduces retries (default 3) into `download-extensions` with a short sleep of 1 second between each one.

- 3 retries if there is a failure in `download-extension`
- 1 second sleep between retries

### Legal Acknowledgement


jsiirola:
I realize this is merged, but is this the right place to put the retry? I would have thought a better place to put retries is in our test drivers, under the theory that the downloaded should just download and any retries should be left to the driver (much like curl and wget don’t retry automatically).

PR #3657, opened 2025-07-09
Solution Pool

whart222:
## Summary/Motivation:

The alternative_solutions contrib package includes a rudimentary solution pool object. This PR includes a more comprehensive capability for defining and managing solution pools.

- PyomoPoolManager - Manages pools for Pyomo solutions
- SolutionPool* - Classes used to define various solution pools
- Functions in pyomo.contrib.alternative_solutions now use and return PyomoPoolManager instances.
- PyomoSolution - Function that returns a Solution object given Pyomo model data 

NOTE: This is a WIP PR.  I'm submitting this now to help define expectations for finalizing this new capability.

### Legal Acknowledgement


whart222:
Note that this PR now includes updates to the Bunch class defined in pyomo.common.  I have been using the public Munch class, but these revisions align the Bunch API with Munch.

whart222:
@emma58 @michaelbynum FYI, this PR is ready for review.

mrmundt:
@whart222 - Can you please resolve the merge conflicts?

whart222:
@mrmundt @emma58 @michaelbynum Merge conflicts have been resolved.

whart222:
@emma58 @michaelbynum FYI, a CMU collaborator recently highlighted that the approval of this PR is blocking CMU software commits to snoglode that would leverage this capability.

pyomo/contrib/alternative_solutions/tests/test_balas.py
```diff
@@ -39,6 +39,16 @@ def test_bad_solver(self, mip_solver):
...
 
+    def test_non_positive_num_solutions(self, mip_solver):
+        """
+        Confirm that an exception is thrown with a non-positive num solutions
+        """
+        m = tc.get_triangle_ip()
+        try:
+            enumerate_binary_solutions(m, num_solutions=-1, solver=mip_solver)
+        except AssertionError as e:
+            pass
```
mrmundt: This test doesn't really make sense to me. If you want to actually check that an exception is thrown, wouldn't it be better to use something like:
```
with self.assertRaises(WhateverException):
   --- do the thing ---
```

pyomo/contrib/alternative_solutions/tests/test_gurobi_lp_enum.py
```diff
@@ -31,13 +31,23 @@
...
 
+    def test_non_positive_num_solutions(self):
+        """
+        Confirm that an exception is thrown with a non-positive num solutions
+        """
+        n = tc.get_pentagonal_pyramid_mip()
+        try:
+            gurobi_enumerate_linear_solutions(n, num_solutions=-1)
+        except AssertionError as e:
+            pass
```
mrmundt: Same comment as above.

pyomo/contrib/alternative_solutions/tests/test_gurobi_solnpool.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def test_search_mode(self):
+        """
+        Confirm that an exception is thrown with pool_search_mode not in [1,2]
+        """
+        m = tc.get_triangle_ip()
+        try:
+            gurobi_generate_solutions(m, pool_search_mode=0)
+        except AssertionError as e:
+            pass
```
mrmundt: Same comment as above

pyomo/contrib/alternative_solutions/tests/test_lp_enum.py
```diff
@@ -42,6 +42,16 @@ def test_bad_solver(self, mip_solver):
...
 
+    def test_non_positive_num_solutions(self, mip_solver):
+        """
+        Confirm that an exception is thrown with a non-positive num solutions
+        """
+        m = tc.get_3d_polyhedron_problem()
+        try:
+            lp_enum.enumerate_linear_solutions(m, num_solutions=-1, solver=mip_solver)
+        except AssertionError as e:
+            pass
```
mrmundt: Same comment as above.

pyomo/contrib/alternative_solutions/tests/test_solnpool.py
```diff
@@ -1,210 +1,980 @@
-#  ___________________________________________________________________________
-#
-#  Pyomo: Python Optimization Modeling Objects
-#  Copyright (c) 2008-2025
-#  National Technology and Engineering Solutions of Sandia, LLC
-#  Under the terms of Contract DE-NA0003525 with National Technology and
-#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain
-#  rights in this software.
-#  This software is distributed under the 3-clause BSD License.
-#  ___________________________________________________________________________
```
mrmundt, line 10: Copyright was accidentally deleted.

pyomo/contrib/alternative_solutions/tests/test_solnpool.py
```diff
@@ -1,210 +1,980 @@
...
-        # Use quiet=False to test error message
-        results = gurobi_generate_solutions(
-            m, num_solutions=100, solver_options={"TimeLimit": 0.0}, quiet=False
-        )
-        assert len(results) == 0
-
-
-if __name__ == "__main__":
-    unittest.main()
+import pytest
```
mrmundt: Can you please do `from pyomo.common.unittest import pytest` instead?

Also, generally speaking, this test file doesn't match the "normal" Pyomo format. It'd be better to use the unittest.TestCase thing we normally do.

pyomo/contrib/alternative_solutions/tests/test_solnpool.py
```diff
@@ -1,210 +1,980 @@
...
+    except ValueError as e:
+        pass
+
+
+def test_invalid_policy_2():
+    pm = PoolManager()
+    try:
+        pm.add_pool(name="pool", policy="invalid_policy", max_pool_size=-2)
+    except ValueError as e:
+        pass
```
mrmundt: Same comment as above re: `with self.assertRaises`

pyomo/contrib/alternative_solutions/tests/test_solnpool.py
```diff
@@ -1,210 +1,980 @@
...
+
+
+def test_keeplatest_bad_max_pool_size():
+    pm = PoolManager()
+    try:
+        pm.add_pool(name="pool", policy="keep_latest", max_pool_size=-2)
+    except AssertionError as e:
+        pass
```
mrmundt: Same comment as above re: `with self.assertRaises`

pyomo/contrib/alternative_solutions/gurobi_lp_enum.py
```diff
@@ -33,6 +33,7 @@ def __init__(
         all_variables,
         orig_objective,
         num_solutions,
+        poolmanager,
```
mrmundt: It would be more accessible to have this as `pool_manager` instead of `poolmanager`.

pyomo/contrib/alternative_solutions/balas.py
```diff
@@ -71,16 +72,21 @@ def enumerate_binary_solutions(
         Boolean indicating that the solver output should be displayed.
     seed : int
         Optional integer seed for the numpy random number generator
+    poolmanager : None
```
mrmundt: It would be more accessible to have this as `pool_manager` instead of `poolmanager`.

pyomo/contrib/alternative_solutions/gurobi_lp_enum.py
```diff
@@ -116,14 +126,24 @@ def enumerate_linear_solutions_soln_pool(
...
         Boolean indicating that the solver output should be displayed.
+    poolmanager : None
+        Optional pool manager that will be used to collect solution
 
     Returns
     -------
-    solutions
-        A list of Solution objects.
-        [Solution]
+    poolmanager
```
mrmundt: It would be more accessible to have this as `pool_manager` instead of `poolmanager`.

pyomo/contrib/alternative_solutions/gurobi_solnpool.py
```diff
@@ -0,0 +1,130 @@
...
+
+def gurobi_generate_solutions(
+    model,
+    *,
+    num_solutions=10,
+    rel_opt_gap=None,
+    abs_opt_gap=None,
+    solver_options={},
+    tee=False,
+    poolmanager=None,
```
mrmundt: It would be more accessible to have this as `pool_manager` instead of `poolmanager`.

pyomo/contrib/alternative_solutions/gurobi_solnpool.py
```diff
@@ -0,0 +1,130 @@
...
+#  ___________________________________________________________________________
+
+import logging
+
+logger = logging.getLogger(__name__)
+
+from pyomo.common.dependencies import attempt_import
+from pyomo.common.errors import ApplicationError
+
+from pyomo.contrib import appsi
```
mrmundt, line 19: It's probably better, if you are doing this work anyway, to NOT pull from APPSI. That's basically deprecated at this point in favor of `pyomo.contrib.solver`.

pyomo/contrib/alternative_solutions/lp_enum.py
```diff
@@ -35,9 +30,11 @@ def enumerate_linear_solutions(
     solver_options={},
     tee=False,
     seed=None,
+    poolmanager=None,
```
mrmundt: It would be more accessible to have this as `pool_manager` instead of `poolmanager`.

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -1,127 +1,1001 @@
-#  ___________________________________________________________________________
```
mrmundt, line 1: Copyright accidentally deleted!

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -1,127 +1,1001 @@
...
-    Changes to this mode can be made by included PoolSearchMode set to the intended value
-    in solver_options.
+import heapq
+import collections
+import dataclasses
+import json
+import weakref
+
+from .aos_utils import MyMunch, _to_dict
+from .solution import Solution, PyomoSolution
```
mrmundt: We generally prefer absolute paths.

pyomo/contrib/alternative_solutions/solution.py
```diff
@@ -1,158 +1,282 @@
-#  ___________________________________________________________________________
```
mrmundt, line 1: Accidentally deleted copyright!

mrmundt (changes requested):
This is by no means comprehensive, but I wanted to provide at least some feedback to start.

pyomo/contrib/alternative_solutions/gurobi_solnpool.py
whart222: There is *a lot* of code here that uses APPSI.  Changing this is *not* going to be easy.

I did a quick check if I could drop-in the pyomo.contrib.solver.solvers.gurobi_direct.GurobiDirect() solver, and it failed. :(

whart222: I really think that removing APPSI needs to be a separate PR.

blnicho: @whart222 we talked about this at the dev call and are fine with this being a separate PR as you suggested.

doc/OnlineDocs/explanation/analysis/alternative_solutions.rst
```diff
@@ -2,233 +2,359 @@
...
-  are realized in later stages of model analysis.
+* alternative optimal solutions can be used to assess trade-offs between
+  competing objectives; 
+ 
+* comparisons amongst alternative solutions provide 
+  insights into the efficacy of model predictions with 
+  inaccurate or untrusted optimization formulations; or 
+ 
+* alternative solutions can be identified to support the future analysis of model revisions (e.g. to 
+  account for previously unexpressed constraints).
```
emma58: It seems like there is also just the obvious use case of wanting to enable a decision maker to explore the optimal space: for example, for a design optimization.

doc/OnlineDocs/explanation/analysis/alternative_solutions.rst
```diff
@@ -2,233 +2,359 @@
...
 
 The *alternative-solutions library* provides a variety of functions that
 can be used to generate optimal or near-optimal solutions for a pyomo
 model. Conceptually, these functions are like pyomo solvers. They can
-be configured with solver names and options, and they return a list of
+be configured with solver names and options, and they return a pool of
 solutions for the pyomo model. However, these functions are independent
-of pyomo's solver interface because they return a custom solution object.
+of pyomo's solver interface because they return a custom pool manager object.
```
emma58: typo: interface -> interfaces

doc/OnlineDocs/explanation/analysis/alternative_solutions.rst
```diff
@@ -2,233 +2,359 @@
...
 
-* ``enumerate_binary_solutions``
+* :py:func:`enumerate_binary_solutions`
 
     * Finds alternative optimal solutions for a binary problem using no-good cuts.
 
-* ``enumerate_linear_solutions``
+* :py:func:`enumerate_linear_solutions`
 
     * Finds alternative optimal solutions for a (mixed-integer) linear program.
```
emma58: This is the only one that doesn't say *how* it generates alternative optimal solutions.

doc/OnlineDocs/explanation/analysis/alternative_solutions.rst
```diff
@@ -2,233 +2,359 @@
...
-   }
+The previous example enumerated all feasible solutions. However optimization models are typically
+used to identify optimal or near-optimal solutions.  The ``abs_opt_gap`` and ``rel_opt_gap``
+arguments are used to limit the search to these solutions:
 
+* ``rel_opt_gap`` : non-negative float or None
 
-Gap Usage Example
------------------
+  * The relative optimality gap for allowable alternative solutions.  None implies that there is no limit on the relative optimality gap (i.e. that any feasible solution can be considered).
```
emma58:
```suggestion
  * The relative optimality gap for allowable alternative solutions.  Specifying a gap of ``None`` implies that there is no limit on the relative optimality gap (i.e. that any feasible solution can be considered).
```

doc/OnlineDocs/explanation/analysis/alternative_solutions.rst
```diff
@@ -2,233 +2,359 @@
...
 
-When we only want some of the solutions based off a tolerance away from
-optimal, this can be done using the ``abs_opt_gap`` parameter. This is
-shown in the following simple knapsack examples where the weights and
-values are the same.
+* ``abs_opt_gap`` : non-negative float or None
+
+  * The absolute optimality gap for allowable alternative solutions.  None implies that there is no limit on the absolute optimality gap (i.e. that any feasible solution can be considered).
```
emma58:
```suggestion
  * The absolute optimality gap for allowable alternative solutions.  Specifying a gap of ``None`` implies that there is no limit on the absolute optimality gap (i.e. that any feasible solution can be considered).
```

doc/OnlineDocs/explanation/analysis/alternative_solutions.rst
```diff
@@ -2,233 +2,359 @@
...
+The *alternative-solutions library* uses solution pools to filter and store solutions generated by an optimizer.
+The following types of solution pools are currently supported:
+
+* ``keep_all`` : This pool stores all solutions. No solutions are filtered out.
+
+* ``keep_latest`` : This pool stores the latest ``max_pool_size`` solutions that are added to the pool.
+
+  * ``max_pool_size`` (non-negative integer) : The maximum number of solutions that are stored.
+
+* ``keep_latest_unique`` : This pool stores the latest ``max_pool_size`` unique solutions that are added to the pool.
```
emma58: What's the use case for `keep_latest` rather than `keep_latest_unique`? It seems like having the same solution in the pool repeatedly isn't desirable since they aren't really "alternatives"?

doc/OnlineDocs/explanation/analysis/alternative_solutions.rst
```diff
@@ -2,233 +2,359 @@
...
+
+  * ``max_pool_size`` (non-negative integer) : The maximum number of solutions that are stored.
+
+  * ``objective`` (function) : A user-specified function that computes the objective value used for comparisons.
+    
+  * ``abs_tolerance`` (non-negative float) : The absolute tolerance that is used to filter solutions.
+
+  * ``rel_tolernace`` (non-negative float) : The relative tolerance that is used to filter solutions.
+
+  * ``sense_is_min`` (bool) : If True, then the pool will keep solutions with the minimal objective values.
```
emma58, line 160: Just for the sake of API consistency, I would vote for calling this `sense` and having it accept values from the `ObjectiveSense` Enum in pyomo.core.

doc/OnlineDocs/explanation/analysis/alternative_solutions.rst
```diff
@@ -2,233 +2,359 @@
...
+    
+  * ``best_value`` (float) : If specified, then this value is used to filter solutions when the absolute or relative tolerances are specified.
```
emma58: I don't quite understand what this does yet, from the description?

doc/OnlineDocs/explanation/analysis/alternative_solutions.rst
```diff
@@ -2,233 +2,359 @@
...
+
+A pool manager class is used to manage one-or-more solution pools. This allows for flexible collection of solutions with different criteria. For example, the
+the best solutions might be stored along with all per-iteration solutions in an optimization solver.  The solution generation functions 
+in the *alternative-solutions library* return a :py:class:`PyomoPoolManager`.  By default, this pool manager uses a solution pool that keeps the best solutions.
+However, the user can provide a pool manager that is used to store solutions.
+
+For example, we can explicit create a pool manager that keeps the latest solutions.  Consider the previous example, where all
```
emma58:
```suggestion
For example, we can explicitly create a pool manager that keeps the latest solutions.  Consider the previous example, where all
```

doc/OnlineDocs/explanation/analysis/alternative_solutions.rst
```diff
@@ -2,233 +2,359 @@
...
+   >>> print( [soln.id for soln in solns] )
+   [6, 7, 8]
+
+The default solution pool has policy ``keep_best`` with name ``None``.
+If a new Solution pool is added without a name, then the ``None``
+pool is replaced.  Otherwise, if a solution pool is added with an
+existing name an error occurs.
+
+The pool manager always has an active pool.  The pool manager has the
+same API as a solution pool, and the envelope design pattern is used
```
emma58: What is "the envelope design pattern"?

doc/OnlineDocs/explanation/analysis/alternative_solutions.rst
```diff
@@ -2,233 +2,359 @@
...
+               "fixed": false,
+               "index": 3,
+               "name": "x[3]",
+               "suffix": {},
+               "value": 1
+           }
+       ]
    }
 
+Further, variable and objective values can be retrieved either using an integer index or the corresponding name:
```
emma58, line 325: Would it be hard to allow retrieving them using the Var or Objective object itself?

pyomo/contrib/alternative_solutions/aos_utils.py
```diff
@@ -9,11 +9,12 @@
 #  This software is distributed under the 3-clause BSD License.
 #  ___________________________________________________________________________
 
+from pyomo.common.collections import Bunch as Munch
```
emma58, line 12: Out of curiosity, why do you rename it to `Munch`?

pyomo/contrib/alternative_solutions/aos_utils.py
```diff
@@ -302,3 +303,21 @@ def get_model_variables(
...
+class MyMunch(Munch):
+    # WEH, MPV needed to add a to_dict since Bunch did not have one
+    def to_dict(self):
+        return _to_dict(self)
+
+
+def _to_dict(x):
+    xtype = type(x)
+    if xtype in [float, int, complex, str, list, bool] or x is None:
+        return x
```
emma58: Will there be any need to handle numpy types or any other numeric types?

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1004 @@
...
 
-import logging
+import heapq
+import collections
+import dataclasses
+import json
+import weakref
 
-logger = logging.getLogger(__name__)
+from pyomo.contrib.alternative_solutions.aos_utils import MyMunch, _to_dict
```
emma58: Should `_to_dict` be private if you need to import it here (and elsewhere)?

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1004 @@
...
 
-from pyomo.contrib import appsi
-import pyomo.contrib.alternative_solutions.aos_utils as aos_utils
-from pyomo.contrib.alternative_solutions import Solution
 
+def _as_solution(*args, **kwargs):
+    if len(args) == 1 and len(kwargs) == 0:
+        assert type(args[0]) is Solution, "Expected a single solution"
+        return args[0]
+    return Solution(*args, **kwargs)
```
emma58: Could the `Solution` constructor handle this check?

pyomo/contrib/alternative_solutions/solution.py
```diff
@@ -9,150 +9,285 @@
...
+    value : float
+        The value of the variable.
+    fixed : bool
+        If True, then the variable was fixed during optimization.
+    name : str
+        The name of the variable.
+    index : int
+        The unique identifier for this variable.
+    discrete : bool
+        If True, then this is a discrete variable
```
emma58: Would it be more general to just store the Var's domain, and then have a utility to check it it's discrete?

pyomo/contrib/alternative_solutions/solution.py
```diff
@@ -9,150 +9,285 @@
...
+    suffix : dict
```
emma58, line 53: For this and ObjectiveInfo, maybe you could actually store a reference to the original object, and then the only extra things you need are `value_in_soln`, `fixed_in_soln` (or something like that), and perhaps your `suffix` attribute. Everything else (and more) will be on the `Var` object itself. I'm thinking about this because it looks hard to find the original `Var` using a `VariableInfo`--you'd have to use the name, which isn't so fun. And given that, questions your current API doesn't answer but might someday be relevant (for example, was the Var binary?) will be easy to answer without any design changes.

pyomo/contrib/alternative_solutions/solution.py
```diff
@@ -9,150 +9,285 @@
...
+    ----------
+    value : float
+        The objective value.
+    name : str
+        The name of the objective.
+    index : int
+        The unique identifier for this objective.
+    suffix : dict
+        Other information about this objective.
+    """
```
emma58, line 84: See my comment above, but wondering if storing a reference to the original Objective might not be a good thing here--that way someone could get the expression or the sense or whatever they wanted.

pyomo/contrib/alternative_solutions/solution.py
```diff
@@ -9,150 +9,285 @@
...
+    An object that describes an optimization solution.
+
+    Parameters
+    -----------
+    variables : None or list
+        A list of :py:class:`VariableInfo` objects. (default is None)
+    objective : None or :py:class:`ObjectiveInfo`
+        A :py:class:`ObjectiveInfo` object. (default is None)
+    objectives : None or list
+        A list of :py:class:`ObjectiveInfo` objects. (default is None)
```
emma58, line 107: Why are there both `objective` and `objectives`? Couldn't `objectives` just have length 1 if there is only one?

pyomo/contrib/alternative_solutions/solution.py
```diff
@@ -9,150 +9,285 @@
...
+    kwds : dict
+        A dictionary of auxiliary data that is stored with the core solution values.
```
emma58: Is `kwds` the right name for this? Maybe `aux_info`? Or maybe `suffix` to match your other naming?

pyomo/contrib/alternative_solutions/solution.py
```diff
@@ -9,150 +9,285 @@
...
+    def __init__(self, *, variables=None, objective=None, objectives=None, **kwds):
+        self.id = None
+
+        self._variables = []
+        self.name_to_variable = {}
+        self.fixed_variable_names = set()
+        if variables is not None:
+            self._variables = variables
+            for v in variables:
+                if v.name is not None:
```
emma58, line 127: What happens if `v.name` is `None`?

pyomo/contrib/alternative_solutions/solution.py
```diff
@@ -9,150 +9,285 @@
...
+                    self.name_to_variable[v.name] = v
+
+        self._objectives = []
+        self.name_to_objective = {}
+        if objective is not None:
+            objectives = [objective]
+        if objectives is not None:
+            self._objectives = objectives
+            for o in objectives:
+                if o.name is not None:
```
emma58, line 139: Same question here.

pyomo/contrib/alternative_solutions/solution.py
```diff
@@ -9,150 +9,285 @@
...
+                    self.name_to_objective[o.name] = o
+
+        if "suffix" in kwds:
+            self.suffix = MyMunch(kwds.pop("suffix"))
+        else:
+            self.suffix = MyMunch(**kwds)
```
emma58: This might be worth mentioning in the docstring, that if you give suffix as a kwd arg, you will lose anything else that is bonus, but if you don't give it, you keep everything.

pyomo/contrib/alternative_solutions/solution.py
```diff
@@ -9,150 +9,285 @@
...
-            for var, val in self.variables.items()
-            if include_fixed or not var in self.fixed_vars
-        }
+        return json.dumps(self.to_dict(), sort_keys=sort_keys, indent=indent)
 
-    def get_fixed_variable_names(self):
-        """
-        Get a list of fixed-variable names.
+    def __str__(self):
+        return self.to_string()
```
emma58, line 233: If I'm understanding everything correctly, this is potentially a *very* long string if someone innocently prints something with several solutions to a large model, right, because every variable will be printed? I wonder if a more succinct output might be more practical in practice?

pyomo/contrib/alternative_solutions/solution.py
```diff
@@ -9,150 +9,285 @@
...
+
+        We use string names if possible, because they more explicit than the integer index values.
         """
-        return value if not round_discrete or variable.is_continuous() else round(value)
+        if len(self.name_to_variable) == len(self._variables):
+            return tuple(
+                tuple([k, var.value]) for k, var in self.name_to_variable.items()
+            )
+        else:
+            return tuple(tuple([k, var.value]) for k, var in enumerate(self._variables))
```
emma58, line 248: Do you hit this case when some of the `VariableInfo`s had `None` as the name? Or if not, why can this happen?

pyomo/contrib/alternative_solutions/solution.py
```diff
@@ -9,150 +9,285 @@
...
+
+    def __eq__(self, soln):
+        if not isinstance(soln, Solution):
+            return NotImplemented
+        return self._tuple_repn() == soln._tuple_repn()
+
+    def __lt__(self, soln):
+        if not isinstance(soln, Solution):
+            return NotImplemented
+        return self._tuple_repn() <= soln._tuple_repn()
```
emma58, line 258: Couldn't this have unexpected behavior if someone accidentally compares solutions that are not from the same model, since the names are in this tuple? For example, it is true that `('bye', 0) <= ('hi', 0)`.

pyomo/contrib/alternative_solutions/solution.py
```diff
@@ -9,150 +9,285 @@
...
+
+
+def PyomoSolution(*, variables=None, objective=None, objectives=None, **kwds):
```
emma58: Naming-wise this looks like a class. I would suggest something like `construct_solution_from_pyomo_components`, but to that end, it feels like what you really want is overloaded constructors for `Solution`. Could you accomplish this by making `pyo_variables `and `pyo_objectives` optional arguments to the `Solution` constructor and moving this logic there? I think it would reduce some complexity and duplicated code in `solnpool.py`.

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1004 @@
...
-        opt.gurobi_options["PoolGapAbs"] = abs_opt_gap
-    for parameter, value in solver_options.items():
-        opt.gurobi_options[parameter] = value
-    if "PoolSearchMode" not in opt.gurobi_options:
-        opt.gurobi_options["PoolSearchMode"] = 2
-    elif opt.gurobi_options["PoolSearchMode"] == 0:
-        logger.warning(
-            "Running gurobi_solnpool with PoolSearchMode=0, this is single search mode and not the intended use case for gurobi_generate_solutions"
+
+    def __init__(self, name, as_solution, counter, policy="unspecified"):
```
emma58: It might be nice to make a Policy Enum rather than using strings, then you can have nice error checking on that.

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1004 @@
...
+        Property to return solution conversion method.
+        """
+        return self._as_solution
+
+    @property
+    def pool_config(self):
+        """
+        Property to return SolutionPool class specific configuration data.
+        """
+        return dict()
```
emma58, line 148: Should this return an empty dictionary, or a `NotImplementedError`? Seems like you are assuming that the subclasses will implement this?

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1004 @@
...
+        int
+            The ID value to match the added solution from the solution pool's PoolCounter.
+            The ID value is also the pool dictionary key for this solution.
+        """
+        soln = self._as_solution(*args, **kwargs)
+        #
+        soln.id = self._next_solution_counter()
+        assert (
+            soln.id not in self._solutions
+        ), f"Solution id {soln.id} already in solution pool context '{self._context_name}'"
```
emma58: Perhaps make this a `DeveloperError`? This comment applies to a the other assertions of this ilk too.

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1004 @@
...
+        dict
+            Dictionary with three keys: 'metadata', 'solutions', 'pool_config'
+            'metadata' contains a dictionary of information about SolutionPools that is always present.
+            'solutions' contains a dictionary of the pool's solutions.
+            'pool_config' contains a dictionary of details conditional to the SolutionPool type.
+        """
+        return dict(
+            metadata=_to_dict(self.metadata),
+            solutions=_to_dict(self._solutions),
+            pool_config=_to_dict(self.pool_config),
```
emma58: Why is this method not implemented in the base class? It looks to be the same for most of these.

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1004 @@
...
+    counter : PoolCounter or None
+        PoolCounter object to manage solution indexing.
+        A value of None will result in a new PoolCounter object being created and used.
+    max_pool_size : None or int
+        Value of None results in no max pool limit based on number of solutions.
+        If not None, the value must be a positive integer.
+        The max_pool_size is the K value for keeping the latest K solutions.
+    objective : None or Function
+        The function to compare solutions based on.
+        None makes the objective be the constant function 0.
```
emma58: What is the expected input for this function? A `Solution` object?

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1004 @@
...
+        soln = self._as_solution(*args, **kwargs)
+        #
+        # Return None if the solution has already been added to the pool
+        #
+        tuple_repn = soln._tuple_repn()
+        if tuple_repn in self._unique_solutions:
+            return None
+        self._unique_solutions.add(tuple_repn)
+        #
+        value = soln.objective(self.objective).value
```
emma58, line 516: I'm confused: In the default case, this makes sense since `self.objective` is `0` and that in turn gets the 0th objective from the `ObjectiveInfo` object on the `Solution` class, so you get its value. But the docstring on this class says that `self.objective` could be a function, which doesn't seem like it would work here, or I'm missing something?

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1004 @@
...
+                (self.rel_tolerance is None)
+                or (
+                    diff / min(math.fabs(value), math.fabs(self.best_value))
+                    <= self.rel_tolerance
+                )
+            ):
+                # Keep if the absolute or relative difference with the best value is small enough
+                keep = True
+
+        if keep:
```
emma58: Minor nit, but rather than defining and checking `keep`, I think you can just return `None` whenever you would set `keep` to `False` and then you don't need this if statement.

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1004 @@
...
+            if self.max_pool_size is None or len(self._heap) < self.max_pool_size:
+                # There is room in the pool, so we just add it
+                heapq.heappush(self._heap, item)
+            else:
+                # We add the item to the pool and pop the worst item in the pool
+                item = heapq.heappushpop(self._heap, item)
+                del self._solutions[item.id]
+
+            if new_best_value:
+                # We have a new best value, so we need to check that all existing solutions are close enough and re-heapify
```
emma58: This makes sense, but it is a lot of work in the worst case. Would it be possible and maybe more efficient to have this implementation just keep the k best solutions and only check tolerance and prune out ones that are outside when asked for its solutions?

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1004 @@
...
+    def __init__(self):
+        self._name = None
+        self._pools = {}
+        self.add_pool(name=self._name)
+        self._solution_counter = 0
+
+    #
+    # The following methods give the PoolManager the same API as a pool.
+    # These methods pass-though and operate on the active pool.
+    #
```
emma58, line 628: Oh, I see, this is what "envelope design" is, right?

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1004 @@
...
+        name : String
+            name key to pick the SolutionPool in the PoolManager object to the active pool
+            If name not a valid key then assertion error thrown
+        Returns
+        ----------
+        dict
+            Metadata attribute of the now active SolutionPool
+
+        """
+        assert name in self._pools, f"Unknown pool '{name}'"
```
emma58: This should probably be a `ValueError`

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1004 @@
...
+
+        Parameters
+        ----------
+        json_filename : path-like
+            File name to read in as SolutionPools for this PoolManager
+            If corresponding file does not exist, throws assertion error
+
+        """
+        # TODO: this does not set an active pool, should we do that?
+        # TODO: this does not seem to update the counter value, possibly leading to non-unique ids
```
emma58: Is this still true? Does that break any of your assumptions in this class?

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1004 @@
...
+        self._solution_counter = value
+
+
+class PyomoPoolManager(PoolManager):
+    """
+    A subclass of PoolManager for handing pools of Pyomo solutions.
+
+    This class redefines the add_pool method to use the _as_pyomo_solution method to construct Solution objects.
+    Otherwise, this class inherits from PoolManager.
+    """
```
emma58, line 913: Why not make the converstion-to-solution method an argument to the `PoolManager` constructor, and then you don't need this class at all--that can just be changed to `_as_pyomo_solution` to get this functionality.

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1004 @@
...
-
-    Returns
-    -------
-    solutions
-        A list of Solution objects.  [Solution]
+    name : String
+        String name to describe the pool.
+    as_solution : Function or None
+        Method for converting inputs into Solution objects.
+        A value of None will result in the default _as_solution function being used.
```
emma58: I like this callback design for its extensibility (and because in a lot of contexts we really only care about part of a solution, not every Var in the model, and this will allow for defining that as a "solution" for AOS purposes.). Can you document somewhere what the expected input and output for this function are? It seems like it is assumed it returns a `Solution` object, right? And it can actually be called given anything because it will pass through what is given to `add`?

pyomo/contrib/alternative_solutions/balas.py
```diff
@@ -71,16 +72,21 @@ def enumerate_binary_solutions(
         Boolean indicating that the solver output should be displayed.
     seed : int
         Optional integer seed for the numpy random number generator
+    pool_manager : None
+        Optional pool manager that will be used to collect solution
```
emma58:
```suggestion
        Optional pool manager that will be used to collect solutions
```

pyomo/contrib/alternative_solutions/tests/test_gurobi_lp_enum.py
```diff
@@ -31,15 +30,20 @@
 @unittest.skipUnless(numpy_available, "NumPy not found")
 class TestLPEnumSolnpool(unittest.TestCase):
 
+    def test_non_positive_num_solutions(self):
+        """
+        Confirm that an exception is thrown with a non-positive num solutions
+        """
+        n = tc.get_pentagonal_pyramid_mip()
+        with self.assertRaises(AssertionError):
+            gurobi_enumerate_linear_solutions(n, num_solutions=-1)
```
emma58: This should be a `ValueError`, I think. Though what matters most is that it *is* an error! :)

emma58 (commented):
Sorry for the huuuuge number of comments, but there's a lot here! Many are minor, but there are a few design questions too: The biggest one is that I'm wondering if the `VariableInfo` and `ObjectiveInfo` classes could store a reference to the original pyomo component rather than copying a lot but not all of the same data. It seems like that would make it *much* easier to map solutions in the pool to the original model, which I imagine will be a common user need.

pyomo/contrib/alternative_solutions/solnpool.py
whart222: I agree that this needs better documentation.

I've renamed _as_solution to default_as_solution, and I've added documentation there.

I've also simplified the API, so that this function *only* is responsible for constructing and returning a solution object.  I moved the checks for the special case where a Solution is passed-in to the add() method.

pyomo/contrib/alternative_solutions/solnpool.py
whart222: This is an interesting question.  Would we ever want to have a pool manager manage pools with different as_solution functions?  Right now, that's supported.  But I'm not sure whether that makes sense.  I'll discuss this with the devs.

whart222: There still seems a motivation for PyomoPoolManager, to avoid exposing the _as_pyomo_solution function.  This is a matter of design, though.  I think it's simpler for an end-user to use a class with a pre-defined configuration like this, than to use a base class with explicit arguments.  In absolute terms, the same # of symbols are exposed to the user, but the semantics are simpler for PyomoPoolManager (i.e. they don't need to think about the as_solution option).

pyomo/contrib/alternative_solutions/solnpool.py
whart222: Yes

pyomo/contrib/alternative_solutions/solnpool.py
whart222: This is a good idea.  But it requires an abstraction that's not currently supported:  asking for the solutions from a pool.  Right now, the code just assumes that they are available within the class.  But a get_solutions() method would be used by methods like to_dict() to get the solutions and do some last-minute processing.

This is a performance optimization that should be moved to a subsequent PR.  I've made a note of this in the issue tracker I use for solution pools.

pyomo/contrib/alternative_solutions/solnpool.py
whart222: Yes, this is a problem.  I need to think about this. :(

pyomo/contrib/alternative_solutions/solnpool.py
whart222: That change would require a complete rework of the conditional logic for setting keep.  It's not obvious how to make that change correctly.

BUT, we can avoid the "if keep" block and return None here.  I made that change, which simplifies the rest of the function.

pyomo/contrib/alternative_solutions/solnpool.py
whart222: The docstring was wrong.  And it's easier just to make the default value of the 'objective' argument zero.

doc/OnlineDocs/explanation/analysis/alternative_solutions.rst
whart222: This is not supported, and it would be a nontrivial change to create the indexing for model components.  I've been treating the solution pools as Pyomo-agnostic to facilitate loading and storing pools.  I'm open to revisiting this decision, but there will be design implications that we need to work through.

pyomo/contrib/alternative_solutions/aos_utils.py
whart222: Munch is the public package that supercedes Bunch capabilities in virtually all ways.  We started using Munch in forestlib development, so this is a hack to back-port to use Pyomo's Bunch without changing all of the forestlib code...

doc/OnlineDocs/explanation/analysis/alternative_solutions.rst
emma58: You wouldn't have to create indexing if you used `ComponentMap` to do this, I don't think?

pyomo/contrib/alternative_solutions/solnpool.py
whart222: Yep.  Done.

pyomo/contrib/alternative_solutions/aos_utils.py
whart222: I reworked the order of data checks, and now the logic is agnostic of numeric types!

whart222:
@emma58  Thanks for all the feedback!  I've gone through some of your comments, but I'll get back to this sometime next week.

doc/OnlineDocs/explanation/analysis/alternative_solutions.rst
whart222: keep_latest_unique is probably what most users want.  But it does require a tuplization of the solution and a check for uniqueness.  Those aren't trivial computations in high dimensions, so I added both alternatives for the solution pool.

More generally, we should recognize that these are just some candidate solution pools that we thought of.  We haven't tried doing a careful vetting of use cases for all of these...

doc/OnlineDocs/explanation/analysis/alternative_solutions.rst
whart222: I revised this to provide more detail.  A user may know what the value of the optimum is, to guide thresholding for absolute or relative tolerances.

doc/OnlineDocs/explanation/analysis/alternative_solutions.rst
whart222: I reworked this to avoid this terminology.  I struggled to find this concept documented well on the internet.

doc/OnlineDocs/explanation/analysis/alternative_solutions.rst
whart222: So the solution pool data would depend on links to an active model?  What if I loaded a solution pool from a file?

pyomo/contrib/alternative_solutions/solution.py
whart222: Our design explicitly disconnects from the structure of a pyomo model and the model component data.  This is intentional, and it is designed to avoid inter-dependencies between solution pools and pyomo models.  Thus, we're not concerns with how to load a model from a pool solution using links to model components, since that requires that we *always* have those links instantiated.

doc/OnlineDocs/explanation/analysis/alternative_solutions.rst
whart222: I am going to leave this functionality to a follow-on PR.  The design of solution pools intentionally omits references to model component data.

pyomo/contrib/alternative_solutions/solution.py
whart222: I'm leaving this as a topic for a follow-on PR.

pyomo/contrib/alternative_solutions/solution.py
whart222: See my note above that this would require a separate PR.  The Solution object, and its constituent components, was designed to be agnostic of Pyomo.

pyomo/contrib/alternative_solutions/solution.py
whart222: The 'objective' argument is used provided to avoid forcing single-objective models from being forced to specify lists of objectives with length one.

I could probably omit this as an argument to Solution without difficulty, but I *really* think it makes sense for the Solution API to have both an objective() and objectives() method.  Hence, the symmetry in the constructor is well-motivated.

pyomo/contrib/alternative_solutions/solution.py
whart222: I changed this variable to 'kwargs', which matches the standard Python convention for capturing all other keyword arguments.  That really is the intent here.

pyomo/contrib/alternative_solutions/solution.py
whart222: Same as above. I want to make it convenient to specify objective data in a manner that best suits their application: single-objective of multi-objective.

pyomo/contrib/alternative_solutions/solution.py
whart222: Done

pyomo/contrib/alternative_solutions/solution.py
whart222: Yes this could be long, but the to_string() method calls json.dumps()  using an indent option, so the default behavior is not one long string with no linebreaks, but a long string with many line breaks.  I think that's reasonable/appropriate.

I'm not sure I know of a suitable alternative that provides solution details for the user.

whart222: I could imagine a future PR specifying a pretty-print option.  Maybe that could be used to control the level of detail in the print?

pyomo/contrib/alternative_solutions/solution.py
whart222: This case arises when a user adds a solution using integer variable indices and not variable names.  Clearly, Pyomo models would have names.  But the core of the solution pool is Pyomo-agnostic, and there are plenty of modeling environments that serialize variables into arrays without names.

whart222: For example, we're anticipating the integration of solution pools into our mpi-sppy interface, and mpi-sppy serializes variables into a single array.

doc/OnlineDocs/explanation/analysis/alternative_solutions.rst
whart222: There are some subtleties here, so I'm punting this to a future PR.  The solution pool core is Pyomo agnostic, and I want to make sure that we're propagating the appropriate information.

pyomo/contrib/alternative_solutions/solution.py
whart222: If v.name is None, then no variable name was specified by the user.  Hence, we don't cache indexing values associated with the variable name.

pyomo/contrib/alternative_solutions/solution.py
whart222: Well, the operator is well defined even in this case.  So what is the problem?  

Solutions from multiple models seems like a reasonable use case.  We're not trying to enforce model consistency here.  Rather, we're just adding methods that are needed to allow these objects be compared (and used in dictionary keys, I think).

pyomo/contrib/alternative_solutions/solution.py
whart222: Keeping the core functionality differentiated from Pyomo was a core design principle.  I suggest we revisit that as a PR.  I'm not convinced this would reduce the complexity of the code.

We are imagining use cases where we have different solution types, like GTEP solutions.  Hence, keeping this separation seems sensible ... at least for now.

I've added a PyomoSolution class whose base class is Solution.  Semantically, that's more consistent with what we imagine doing with GTEP and other applications.

pyomo/common/tests/test_bunch.py
```diff
@@ -96,6 +97,19 @@ def test_Bunch1(self):
         ):
             Bunch('a=5 foo = 6')
 
+    def test_Bunch2(self):
```
jsiirola: Please use a more descriptive test name
```suggestion
    def test_Bunch_fromGenerator(self):
```

pyomo/contrib/alternative_solutions/aos_utils.py
```diff
@@ -302,3 +303,21 @@ def get_model_variables(
                 )
 
     return variable_set
+
+
+class MyMunch(Munch):
+    # WEH, MPV needed to add a to_dict since Bunch did not have one
+    def to_dict(self):
+        return to_dict(self)
```
jsiirola, line 308: Why did you add `toDict()` to Bunch and then add `to_dict()` to `MyMunch`?  I don't understand the need for `MyMunch`...

pyomo/contrib/alternative_solutions/solnpool.py
jsiirola: Agree with making `policy` an Enum

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1026 @@
...
+        ----------
+        Input needs to match as_solution format from pool initialization.
+
+        Returns
+        ----------
+        int
+            The ID value to match the added solution from the solution pool's PoolCounter.
+            The ID value is also the pool dictionary key for this solution.
+        """
+        if len(args) == 1 and len(kwargs) == 0 and type(args[0]) is Solution:
```
jsiirola: Some suggestions:
- `not kwargs` instead of calling `len()` and then comparing the resulting int [efficiency]
- use isinstance (e.g., so you can add a PyomoSolution)?
```suggestion
        if not kwargs and len(args) == 1 and isinstance(args[0], Solution):
```

You use this idiom a lot, so any changes here obviously apply there, too.

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1026 @@
...
+        #
+        soln.id = self._next_solution_counter()
+        assert (
+            soln.id not in self._solutions
+        ), f"Solution id {soln.id} already in solution pool context '{self._context_name}'"
+        #
+        self._solutions[soln.id] = soln
+        return soln.id
+
+    def to_dict(self):
```
jsiirola: What is the intended use of this API?  Is it just for serialization?

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1026 @@
...
+
+    The default solution pool has policy ``keep_best`` with name ``None``.
+    If a new Solution pool is added without a name, then the ``None``
+    pool is replaced.  Otherwise, if a solution pool is added with an
+    existing name an error occurs.
+
+    The pool manager always has an active pool.  The pool manager has the
+    same API as a solution pool, and the envelope design pattern is used
+    to expose the methods and data for the active pool.  The active pool
+    defaults to the pool that was most recently added to the pool manager.
```
jsiirola, line 605: I am concerned by this design.  It is inherently not thread safe and will likely lead to unintentional behavior if solution pools are ever used in a hierarchical context.  Further, where we have used it in the past (e.g., in the TempFileManager), we have run into problems (see #2514, #2475, #1125) .  

I strongly recommend a design where we use the manager to add / get pools, and then the clients directly work on pool objects (and not assuming that the pool manager's active pool hasn't changed).  I would go as far to strongly encourage *removing* the Pool API from the PoolManager class.

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1026 @@
...
+        """
+        Initializes a new solution pool and adds it to this pool manager.
+
+        The method expects parameters for the constructor of the corresponding solution pool.
+        Supported pools are `keep_all`, `keep_best`, `keep_latest`, and `keep_latest_unique`.
+
+        Parameters
+        ----------
+        name : String
+            The name of the solution pool.  If name is already used then, then an error is generated.
```
jsiirola: Is there a reason that the name needs to be a string?  Couldn't any hashable object be acceptable?

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1026 @@
...
-    solver_options : dict
-        Solver option-value pairs to be passed to the Gurobi solver.
-    tee : boolean
-        Boolean indicating that the solver output should be displayed.
-
-    Returns
-    -------
-    solutions
-        A list of Solution objects.  [Solution]
+    name : String
```
jsiirola: the type should be valid types (or things from `typing`)
```suggestion
    name : str
```

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1026 @@
...
+        String name to describe the pool.
+    as_solution : Function or None
```
jsiirola:
```suggestion
    as_solution : Callable | None
```
(Ideally, you would specify the API of the callable)

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1026 @@
...
+        """
+        if name is None and None in self._pools:
+            del self._pools[None]
+
+        if name not in self._pools:
+            # Delete the 'None' pool if it isn't being used
+            if name is not None and None in self._pools and len(self._pools[None]) == 0:
+                del self._pools[None]
+
+            if policy == "keep_all":
```
jsiirola: This "if tree" could be replaced by a dispatcher:
```python
class PoolManager:
    _policy_dispatcher = {
        "keep_all": SolutionPool_KeepAll,
        "keep_best": SolutionPool_KeepBest,
        "keep_latest": SolutionPool_KeepLatest,
        "keep_latest_unique": SolutionPool_KeepLatestUnique,
    }
    # ...
    def add_pool(self, *, name=None, policy="keep_best", as_solution=None, **kwds):
        # ...
        self._pool[name] = self._policy_dispatcher[policy](name, as_solution, weakref.proxy(self), **kwds)
```

This both makes the code simpler and also more extensible.

pyomo/contrib/alternative_solutions/solnpool.py
```diff
@@ -9,119 +9,1026 @@
...
+    @property
+    def solution_counter(self):
+        return self._solution_counter
+
+    @solution_counter.setter
+    def solution_counter(self, value):
+        self._solution_counter = value
+
+
+class PyomoPoolManager(PoolManager):
```
jsiirola, line 906: Why does this need to be a separate class?  You could just make the default `as_solution` be an instance attribute on the pool manager?

pyomo/contrib/alternative_solutions/balas.py
```diff
@@ -71,16 +72,21 @@ def enumerate_binary_solutions(
...
-    solutions
-        A list of Solution objects.
-        [Solution]
+    pool_manager
+        A PyomoPoolManager object
 
     """
     logger.info("STARTING NO-GOOD CUT ANALYSIS")
 
+    assert num_solutions >= 1, "num_solutions must be positive integer"
```
jsiirola: Argument checking of public APIs should raise exceptions and not use assertions (running python in optimized mode will skip *all* assertions).

pyomo/contrib/alternative_solutions/gurobi_lp_enum.py
```diff
@@ -116,14 +126,24 @@ def enumerate_linear_solutions_soln_pool(
...
     -------
-    solutions
-        A list of Solution objects.
-        [Solution]
+    pool_manager
+        A PyomoPoolManager object
     """
     logger.info("STARTING LP ENUMERATION ANALYSIS USING GUROBI SOLUTION POOL")
+
+    assert num_solutions >= 1, "num_solutions must be positive integer"
```
jsiirola: Argument checking of public APIs should raise exceptions and not use assertions (running python in optimized mode will skip *all* assertions).

pyomo/contrib/alternative_solutions/balas.py
```diff
@@ -71,16 +72,21 @@ def enumerate_binary_solutions(
...
+    pool_manager : None
+        Optional pool manager that will be used to collect solutions
 
     Returns
     -------
-    solutions
-        A list of Solution objects.
-        [Solution]
+    pool_manager
+        A PyomoPoolManager object
```
jsiirola, line 81: Wouldn't it make more sense for this to return the *pool* and not the *pool manager*?  You are implicitly trusting that the "active pool" stays constant not only throughout this routine, but also after it exits.

(This comment applies to all AOS routines.)

pyomo/contrib/alternative_solutions/balas.py
```diff
@@ -71,16 +72,21 @@ def enumerate_binary_solutions(
         Boolean indicating that the solver output should be displayed.
     seed : int
         Optional integer seed for the numpy random number generator
+    pool_manager : None
+        Optional pool manager that will be used to collect solutions
```
jsiirola, line 76: Wouldn't it make more sense to pass in the specific *pool* the user wants to the routine to use and not the whole *pool manager*?

pyomo/contrib/alternative_solutions/gurobi_solnpool.py
```diff
@@ -0,0 +1,130 @@
+#  ___________________________________________________________________________
```
jsiirola, line 1: This module name is misleading: I expected this to define a derived SolutionPool class that was specific to the gurobipy interface.

pyomo/contrib/alternative_solutions/lp_enum.py
```diff
@@ -77,15 +74,20 @@ def enumerate_linear_solutions(
...
     -------
-    solutions
-        A list of Solution objects.
-        [Solution]
+    pool_manager
+        A PyomoPoolManager object
     """
     logger.info("STARTING LP ENUMERATION ANALYSIS")
 
+    assert num_solutions >= 1, "num_solutions must be positive integer"
```
jsiirola: Argument checking of public APIs should raise exceptions and not use assertions (running python in optimized mode will skip *all* assertions).

pyomo/contrib/alternative_solutions/solution.py
jsiirola: I find the duplicative API confusing.  It would be simpler and less error-prone for the `__init__` to take a single argument, and then in the constructor map a single ObjectiveData to a list of length 1 (you can also then accept an IndexedObjective and map it to a list of Objectives!)

(To the point about "error prone": note that the current implementation for processing the "2 argument solution" has logic errors and will silently drop objectives if both `objective=` and `objectives=` are specified.)

pyomo/contrib/alternative_solutions/solution.py
jsiirola: Are there any applications using a Solution (Pool) from Pyomo that are making use of the feature that the pool is disconnected from Pyomo?  This seems like a design decision that is adding a lot of complexity and potential for logic errors for a use case that doesn't exist?

For example, you are re-inventing solutions for caching Pyomo objects that have already been well-defined and worked out / debugged elsewhere (e.g., `pyomo.common.collections`; `pyomo.contrib.solver`).

Further, the dev team has been spending a significant amount of effort to *remove* the use of string names from other parts of Pyomo (e.g., `parmest`), so going and adding a new API that is string-based seems to run counter to that effort.

pyomo/contrib/alternative_solutions/solution.py
jsiirola: That change does not appear to have been pushed to this PR?

pyomo/contrib/alternative_solutions/solnpool.py
whart222: Serialization, debugging, testing, etc

pyomo/contrib/alternative_solutions/solnpool.py
whart222: This design doesn't involve a singleton pool, so I don't see why there would be thread safety issues

pyomo/contrib/alternative_solutions/solnpool.py
whart222: Maybe?  Is there a use case that would motivate a hashable name?  I'm not sure what motivates this comment.

I'm not sure what it would mean to serialize a hashable object.  The code certainly doesn't support that right now, so I'm inclined to treat this generalization as a separate PR.

pyomo/contrib/alternative_solutions/solnpool.py
whart222: Done

pyomo/contrib/alternative_solutions/solnpool.py
whart222: Done

pyomo/contrib/alternative_solutions/solnpool.py
whart222: Done

pyomo/contrib/alternative_solutions/solution.py
whart222: I pushed this today. I've been working on this in my spare time, and I forgot to push last time.

pyomo/contrib/alternative_solutions/solnpool.py
whart222: No, I think the right semantics are that the subclasses do not need to implement this method if they do not have configuration options.

pyomo/contrib/alternative_solutions/solnpool.py
whart222: Done

pyomo/contrib/alternative_solutions/solnpool.py
whart222: I'm not sure.  I moved it.

pyomo/contrib/alternative_solutions/solnpool.py
whart222: I corrected this.  The value is an int.

pyomo/contrib/alternative_solutions/aos_utils.py
whart222: Adding toDict() to Bunch supports API compatibility with Munch.  But it turns out I'm not using that so I removed it.

MyMunch is an alias that allows pyomo.contrib.alternative_solutions to use a Munch object without importing the 'munch' package (because Bunch is used on the backend).  And further it adds a to_dict() method, for compatibility with the naming convention used in pyomo.contrib.alternative_solutions.

pyomo/contrib/alternative_solutions/balas.py
whart222: Done (except for a few assertions buried in old code that I'm not sure how to rework).

pyomo/contrib/alternative_solutions/balas.py
whart222: I don't think so.  An AOS method can populate multiple pools.  Hence, a pool manager seems like the right level of abstraction.

pyomo/contrib/alternative_solutions/balas.py
whart222: See above

pyomo/contrib/alternative_solutions/gurobi_solnpool.py
whart222: What would you suggest?  Something like gurobi_with_native_solution_pool.py seems verbose.

pyomo/contrib/alternative_solutions/lp_enum.py
whart222: Done

pyomo/contrib/alternative_solutions/solution.py
whart222: I think this feedback is devolving into a design discussion that is out of scope for this PR.  This class does not accept Pyomo Objective data classes, and I don't think it's feasible to consider that within the current PR.

pyomo/contrib/alternative_solutions/solnpool.py
whart222: Yep, we could get rid of PyomoSolution and bury all of this logic into the pyomo-specific as_solution function.

I think we should leave this as-is for now, since this design allows for a pyomo-specific solution object that is differentiated from the core Solution object.  We're not leveraging that distinction today, but I'm still trying to understand the requirements for serialization of solution pools.

pyomo/contrib/alternative_solutions/gurobi_lp_enum.py
whart222: Done

whart222:
It's not obvious to me why the linux slim tests are failing.  Do we need to conditionally import numpy?

pyomo/contrib/alternative_solutions/solnpool.py
whart222: FYI, I've removed the read/write semantics.  It turns out that there are some deeper issues that need to be addressed to support this, so I'm deferring this to a subsequent PR.

blnicho:
> It's not obvious to me why the linux slim tests are failing. Do we need to conditionally import numpy?

@whart222 yes, numpy needs to be treated as an optional dependency so tests relying on numpy need to be skipped when it's not available.

mrmundt:
@whart222 - Should this PR be closed, given the followup to our discussion last month?

PR #3660, opened 2025-07-15
Add project-level default configuration for black to pyproject.toml

sadavis1:
## Summary/Motivation:
Make it easier to use black and harder to accidentally mess it up

Add [tool.black] section to pyproject.toml. This will cause a command-line invocation like `black .` to correctly use Pyomo's black options without needing to set them yourself.

Notes:
 - Black walks up the filesystem looking for this file starting from the file being formatted, so it will still work if you run it from an unrelated working directory
 - You can also set the intended python version, but it infers this anyways based on the requires-python field under [project] above.
 - The difference between exclude and extend-exclude is that with extend-exclude it also excludes gitignored files (which are the default excluded files). This is meaningful because some of our tests generate python code and leave the gitignored, non-black-compliant files behind.
 - It's still possible to mess this up if you set conflicting command-line flags, as they override this. This is most likely to happen if you are using an ide integration and you set a conflicting option there (as this is almost certainly implemented by changing the flags).

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/test_branches.yml
+++ b/.github/workflows/test_branches.yml
@@ -44,9 +44,10 @@ jobs:
         python-version: '3.10'
     - name: Black Formatting Check
       run: |
-        # Note v24.4.1 fails due to a bug in the parser
+        # Note v24.4.1 fails due to a bug in the parser. Project-level
+        # configuration is inherited from pyproject.toml.
         pip install 'black!=24.4.1'
-        black . -S -C --check --diff --exclude examples/pyomobook/python-ch/BadIndent.py
+        black . --check --diff
     - name: Spell Check
       uses: crate-ci/typos@master
       with: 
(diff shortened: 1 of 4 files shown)


jsiirola:
This is great.  One question: does it also mean that we can remove the command line options from the GHA testing workflows?

sadavis1:
I tried on another branch and it looks like it works correctly. Should I add that to this PR too? I guess it would be redundant not to, but on the other hand there's no harm in being excessively verbose in a script.

jsiirola:
@sadavis1: yes: I think adding that to this PR makes sense.  Leaving it in the driver scripts runs the risk of inconsistency (where someone later changes one and forgets to change the other).

sadavis1:
Sure. I'll also edit contribution_guide.rst in the docs so we don't instruct people to type more than they need to.

mrmundt (approved):
I love this, thank you!

PR #3672, opened 2025-08-01
Bugfixes for `core.lp_dual` transformation

emma58:
## Summary/Motivation:

NOTE: This is built on top of #3651, so it will be impossible to review until that is merged.

This fixes a couple bugs in the `core.lp_dual` transformation. In particular:
1) The Parameterized Standard Form walker now checks if a Var is being treated as a parameter *before* checking if it is fixed--this is very important because it mean that we were losing fixed Vars we were treating as data, leading to unexpected behavior in dualize-and-combine methods.
2) We now raise an error for completely empty primal models because we don't actually do the right thing in the sense that strong duality won't hold (whatever that means), so I'd rather just not do anything
3) We now raise an error when parameterization causes a primal Constraint to become trivial because, unlike with trivial constraints not due to parameterization, we cannot just ignore them, but the behavior from taking the dual with them in 
(whatever that means) is rather unexpected, I think.
4) We now raise an error when the primal model has discrete variables (or other domains we can't handle).

- Fixing the three bugs above, the first of which is actually a change to parameterized standard form
- Adding tests

### Legal Acknowledgement


pyomo/core/plugins/transform/lp_dual.py
```diff
@@ -141,17 +170,27 @@ def _take_dual(self, model, std_form):
                 primal_row = A.indices[j]
                 lhs += coef * dual.x[primal_row]
 
-            if primal.domain is Reals:
+            domain = primal.domain
+            lb, ub = domain.bounds()
+            if not domain == RangeSet(*domain.bounds(), 0):
```
jsiirola, line 176: It would be nice to add a comment that this test is checking for compactness
```suggestion
            # Note: the following checks the domain for continuity and compactness:
            if not domain == RangeSet(*domain.bounds(), 0):
```

pyomo/repn/plugins/parameterized_standard_form.py
```diff
@@ -224,6 +224,64 @@ def todense(self):
 
         return dense
 
+    def tocsr(self):
```
jsiirola, line 227: We ought to follow scipy's lead and only implement a single method that `tocsr` and `tocsc` both use (instead of virtually duplicate code)

jsiirola (approved):
Looks good.  Two comments that we should consider (but it is fine to leave them for a future PR).

pyomo/core/plugins/transform/lp_dual.py
```diff
@@ -121,7 +128,29 @@ def _take_dual(self, model, std_form):
...
+            # parameterization, which we can detect if the row is all 0's.
+            if A_csr.indptr[j] == A_csr.indptr[j + 1]:
+                # All 0's in the coefficient matrix: check what's on the RHS
+                b = std_form.rhs[j]
+                if type(b) not in native_numeric_types:
+                    # The parameterization made this trivial. I'm not sure what's
+                    # really expected here, so maybe we just scream? Or we leave
+                    # the constraint in the model as it is written...
+                    raise ValueError(
+                        f"The primal model contains a constraint that that the "
```
blnicho:
```suggestion
                        f"The primal model contains a constraint that the "
```

pyomo/core/tests/unit/test_lp_dual.py
```diff
@@ -398,3 +400,45 @@ def test_dual_var_map_error(self):
...
             thing = lp_dual.get_dual_var(m, m.c_new)
+
+    def test_parameterization_makes_constraint_trivial(self):
+        m = self.get_bilevel_model()
+        m.budgetish = Constraint(expr=m.outer[2] + m.outer[3] == 1)
+
+        lp_dual = TransformationFactory('core.lp_dual')
+        with self.assertRaisesRegex(
+            ValueError,
+            "The primal model contains a constraint that that the "
```
blnicho:
```suggestion
            "The primal model contains a constraint that the "
```

pyomo/repn/plugins/parameterized_standard_form.py
```diff
@@ -224,6 +224,64 @@ def todense(self):
...
+                # Note that the data changes order because now we are looking
+                # for nonzeros through the columns rather than through the rows.
+                csr_data[dest] = csc_data[j]
+
+                row_index_ptr[row] += 1
+
+        # Fix the row index pointer by inserting 0 at the beginning. The
+        # row_index_ptr currently holds pointers to 1 past the last element of
+        # each row, which is really the starting index for the next
+        # row. Inserting the 0 (the starting index for the firsst column)
```
blnicho:
```suggestion
        # row. Inserting the 0 (the starting index for the first column)
```

blnicho (approved):
I found a few typos but otherwise this looks fine

Issue #2962, opened 2023-08-22
[PEP] Process for handling/testing examples


blnicho:
## Summary

Pyomo examples are inconsistent in how they are implemented and tested and where they are located. We should come up with, implement, and document a process for how examples should be handled within repositories under the Pyomo GitHub organization. 

### Rationale

A coherent and standardized process will help us keep existing examples organized and maintained and will help future developers who wish to contribute examples. 

Relevant comment from @bernalde from #2932: 
> Where should we draw the line between a set of examples, tests, and supported models? In particular, for GDP, we have GDPLib https://github.com/SECQUOIA/gdplib, which could host these two examples. What's the motivation for adding examples to the main Pyomo repository?

### Description

Related issues/tasks:

- [ ] https://github.com/Pyomo/pyomo/issues/101
- [ ] https://github.com/Pyomo/pyomo/issues/1026
- [ ] https://github.com/Pyomo/pyomo/issues/1950

PR #3678, opened 2025-08-05
Bugfix: lp_writer SOS constraints and row_order

mrmundt:
## Summary/Motivation:
In starting work on the new `bar_writer`, I stumbled across a few bugs in `lp_v2` that no one has hit yet but definitely don't work. Primarily, the bugs were:

- Missing import statement
- Accidentally misused letter
- Entirely skipped edge case to correct handle when a model has SOS constraints and is using `row_order` in the writer config.

- Fix two small bugs
- Refactor `ordered_active_constraints` to pull out `row_map` logic (for reusability)
- Bugfix using new `row_order2row_map` function
- `index_set` is now a reserved name, so SOS example fixed so it'll actually run
- Test using SOS example as a baseline

### Legal Acknowledgement

### DIFF
--- a/examples/pyomo/sos/sos2_piecewise.py
+++ b/examples/pyomo/sos/sos2_piecewise.py
@@ -22,7 +22,7 @@ import pyomo.environ as pyo
 
 model = pyo.ConcreteModel()
 
-model.index_set = pyo.Set(initialize=[1, 2])
+model.idx_set = pyo.Set(initialize=[1, 2])
 DOMAIN_PTS = {1: [1, 2, 3], 2: [1, 2, 3]}
 F = {1: [1, 4, 9], 2: [1, 4, 9]}
 # Note we can also implement this like below
@@ -37,18 +37,18 @@ def SOS_indices_init(model, t):
 
 
 model.SOS_indices = pyo.Set(
-    model.index_set, dimen=2, ordered=True, initialize=SOS_indices_init
+    model.idx_set, dimen=2, ordered=True, initialize=SOS_indices_init
 )
 
 
 def sos_var_indices_init(model):
-    return [(t, i) for t in model.index_set for i in range(len(DOMAIN_PTS[t]))]
+    return [(t, i) for t in model.idx_set for i in range(len(DOMAIN_PTS[t]))]
 
 
 model.sos_var_indices = pyo.Set(ordered=True, dimen=2, initialize=sos_var_indices_init)
 
-model.x = pyo.Var(model.index_set)  # domain variable
-model.Fx = pyo.Var(model.index_set)  # range variable
+model.x = pyo.Var(model.idx_set)  # domain variable
+model.Fx = pyo.Var(model.idx_set)  # range variable
 model.y = pyo.Var(model.sos_var_indices, within=pyo.NonNegativeReals)  # SOS2 variable
 
 model.obj = pyo.Objective(expr=pyo.sum_product(model.Fx), sense=pyo.maximize)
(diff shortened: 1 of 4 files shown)


examples/pyomo/sos/sos2_piecewise.py
```diff
@@ -22,7 +22,7 @@
 
 model = pyo.ConcreteModel()
 
-model.index_set = pyo.Set(initialize=[1, 2])
+model.idx_set = pyo.Set(initialize=[1, 2])
```
jsiirola, line 25: This example clearly was not being tested.  Is it now?

pyomo/repn/util.py
```diff
@@ -768,16 +768,26 @@ def ordered_active_constraints(model, config):
...
+
+    # Map the implicit dict ordering to an explicit 0..n ordering
+    return {_id: i for i, _id in enumerate(row_map)}
+
+
+def ordered_active_constraints(model, config):
+    sorter = FileDeterminism_to_SortComponents(config.file_determinism)
+    constraints = model.component_data_objects(Constraint, active=True, sort=sorter)
+
+    row_map = row_order2row_map(config)
```
jsiirola, line 780: While it is good to avoid the duplicate code, from an efficiency / consistency standpoint, this is less than ideal (`row_order2row_map` is being called twice.  I think I would add `row_map` as an argument of this function.

pyomo/repn/util.py
```diff
@@ -768,16 +768,26 @@ def ordered_active_constraints(model, config):
...
         return constraints
-    # map the implicit dict ordering to an explicit 0..n ordering
-    row_map = {_id: i for i, _id in enumerate(row_map)}
     # sorted() is stable (per Python docs), so we can let all
     # unspecified rows have a row number one bigger than the
     # number of rows specified by the user ordering.
     _n = len(row_map)
-    _row_getter = row_map.get
-    return sorted(constraints, key=lambda x: _row_getter(id(x), _n))
+    return sorted(constraints, key=lambda x: row_map.get(id(x), _n))
```
jsiirola: The definition / use of `_row_getter` was intentional: it saves an attribute lookup within the sorter (and was measurable for large models).

examples/pyomo/sos/sos2_piecewise.py
mrmundt: I suspect a large chunk of our examples are not being tested. That's a bigger effort than this PR. If you want me to cobble something together for this exact example, I can, but I'd recommend that we turn it into a targeted PEP instead.

mrmundt: 'Eyyyy, this is, in fact, already a PEP: https://github.com/Pyomo/pyomo/issues/2962

pyomo/repn/tests/cpxlp/test_lpv2.py
```diff
@@ -20,6 +20,72 @@
 from pyomo.repn.plugins.lp_writer import LPWriter
 
 
+def create_sos_model():
```
jsiirola, line 23: This is a weak test case.  For verifying row order, a better case would have a Scalar and an Indexed SOSConstraint, (and the same for Constraint).  The we can test things like setting the order by passing the whole component, and by passing component datas separately. 

The 3 scalar constraints are redundant and unnecessary.

pyomo/repn/util.py
```diff
@@ -768,16 +768,26 @@ def ordered_active_constraints(model, config):
                 row_map[id(c)] = c
         else:
             row_map[id(con)] = con
+
+    if not row_map:
+        return {}
```
jsiirola: This logic is unnecessary: if `row_map` is empty, the following return statement will return an empty dict

pyomo/repn/tests/cpxlp/test_lpv2.py
```diff
@@ -139,3 +205,180 @@ def test_deterministic_unordered_sets(self):
...
+
+end
+"""
+        model = create_sos_model()
+        with LoggingIntercept() as OUT1:
+            LPWriter().write(model, OUT1)
+        self.assertEqual(OUT1.getvalue(), ref_no_row_order)
+        with LoggingIntercept() as OUT2:
+            LPWriter().write(model, OUT2, row_order=[model.constraint3])
+        self.assertEqual(OUT2.getvalue(), ref_row_order)
```
jsiirola: Please turn on `symbolic_solver_labels` so we can make heads or tails of the baselines.

jsiirola (changes requested):

pyomo/repn/util.py
jsiirola: I would like to go through the standard form, LP, and NL writers and refactor this logic to remove redundancy & enforce uniformity (e.g., I am pretty sure that the NLv2 is ignoring row_order!).  But, I am comfortable merging this PR first.  The redundant call to `row_order2row_map` only hits users with both SOSConstraint components *and* who are specifying a `row_order` (and realistically, it's not that big of a hit).

jsiirola (approved):

PR #3693, opened 2025-08-12
Ipopt_v2: update options processing

jsiirola:
## Summary/Motivation:
@eslickj pointed out two problems with the new `contrib.solver` Ipopt interface:
- naming your model anything with a space in it would cause ipopt to fail to parse any options from the option file
- The LegacySolver wrapper failed to process the `OF_` options that the original ipopt interface supported.  

This PR resolves both of those issues:
- Update how we generate the base file name to avoid the quotes that can be added when retrieving the model's local name.  In addition, we process the model name to ensure that it is a valid (allowable by the OS) file name.  Finally, we ensure that all string options passed to the solver on the command line are properly quoted.
- Create a custom Legacy wrapper for Ipopt that will correctly handle the old `OF_*` options
- Define a new utility for ensuring strings are valid file names
- Ensure that temporary filenames are more likely to be unique (by including the PID and thread id)
- Check to make sure the interface doesn't overwrite any existing files
- The standard Pyomo `time_limit` option overrides the ipopt-specific `max_cpu_time` option

- (see above)

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/solver/solvers/ipopt.py
+++ b/pyomo/contrib/solver/solvers/ipopt.py
@@ -38,6 +38,7 @@ from pyomo.core.staleflag import StaleFlagManager
 from pyomo.repn.plugins.nl_writer import NLWriter, NLWriterInfo
 from pyomo.contrib.solver.common.base import SolverBase, Availability
 from pyomo.contrib.solver.common.config import SolverConfig
+from pyomo.contrib.solver.common.factory import LegacySolverWrapper
 from pyomo.contrib.solver.common.results import (
     Results,
     TerminationCondition,
@@ -710,3 +711,15 @@ class Ipopt(SolverBase):
             )
 
         return res
+
+
+class LegacyIpoptSolver(LegacySolverWrapper, Ipopt):
+    def _verify_ipopt_options(self, config: IpoptConfig) -> None:
+        # The old Ipopt solver would map solver_options starting with
+        # "OF_" to the options file.  That is no longer needed, so we
+        # will strip off any "OF_" that we find
+        for opt, val in list(config.solver_options.items()):
+            if opt.startswith('OF_'):
+                config.solver_options[opt[3:]] = val
+                del config.solver_options[opt]
+        return super()._verify_ipopt_options(config)
--- a/pyomo/contrib/solver/solvers/ipopt.py
+++ b/pyomo/contrib/solver/solvers/ipopt.py
@@ -369,7 +369,14 @@ class Ipopt(SolverBase):
                 dname = config.working_dir
             if not os.path.exists(dname):
                 os.mkdir(dname)
-            basename = to_legal_filename(model.name)
+            # Because we are just "making up" a file name, it is better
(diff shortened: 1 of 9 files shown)


pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -349,11 +367,29 @@ def solve(self, model, **kwds) -> Results:
...
+            basename = to_legal_filename(model.name, universal=True)
+            # Strip off quotes - the command line parser will re-add them
+            if basename[0] in "'\"" and basename[0] == basename[-1]:
+                basename = basename[1:-1]
+            # The base file name for this interface is "model_name + PID
+            # + thread id", so that this is reasonably unique in both
+            # parallel and threaded environments (even when working_dir
+            # is set to a persistent directory)
+            basename = os.path.join(
+                dname, f"{basename}.{os.getpid()}.{threading.get_ident()}"
```
mrmundt, line 388: We don't actually have support for threads turned on in IPOPT, BTW. If you can work that into this PR... That'd be great.

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -349,11 +367,29 @@ def solve(self, model, **kwds) -> Results:
...
+            )
+            for ext in ('.nl', '.row', '.col', '.sol', '.opt'):
+                if os.path.exists(basename + ext):
+                    raise RuntimeError(
+                        f"Solver interface file {basename + ext} already exists!"
```
mrmundt, line 393: Where is the funny/snarky message about cowardly abandoning all hope?!

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -680,3 +717,15 @@ def _parse_solution(
             )
 
         return res
+
+
+class LegacyIpoptSolver(LegacySolverWrapper, Ipopt):
```
mrmundt, line 724: This is a neat idea - is it worth documenting "more publicly" that this is something that can be done for new solver interfaces?

mrmundt (approved):
I think this is perfectly fine as-is but had some comments that Tuesday-morning-brain needed to get out.

pyomo/contrib/solver/solvers/ipopt.py
jsiirola: Are you asking for threading support, or a comment that we don't yet support threading, but are including the the thread_id as a bit of future-proofing?

mrmundt: We talked about this in the dev call and settled on "the latter".

pyomo/contrib/solver/solvers/ipopt.py
jsiirola: I think it's more "cowardly refusing to proceed" than "abandoning all hope".

mrmundt: Yes, that one ;)

Issue #3454, opened 2025-01-15
add multi-objective support for highs solver release 1.9.0


shuaiwang88:
## Summary
The highs solver[ release 1.9.0](https://github.com/ERGO-Code/HiGHS/releases/tag/v1.9.0) started to support multiple objectives. Please add this functionality.

quantresearch1:
Hi @shuaiwang88, I just had a look at the HiGHs repo and the HighsLinearObjective data structure as well as the AddLinearObjective method are not available in highspy, it seems it's only available in C/C++. 
see : https://github.com/search?q=repo%3AERGO-Code%2FHiGHS%20addLinearObjective&type=code
I have opened an issue in the HiGHs repo to get clarity on the above: https://github.com/ERGO-Code/HiGHS/issues/2263

On the pyomo side, it's worth discussing if there's an interest to support this more broadly with new priority and tolerance fields for the objective ?
schematically 
class ObjectiveData(NamedExpressionData, ActiveComponentData):
    """This class defines the data for a single objective."""

    __slots__ = ("_args_", "_sense", "_priority", "_tolerance")

    def __init__(self, expr=None, sense=minimize, priority=0, tolerance=0.0, component=None):
        # Inlining NamedExpressionData.__init__
        self._args_ = (expr,)
        # Inlining ActiveComponentData.__init__
        self._component = weakref_ref(component) if (component is not None) else None
        self._index = NOTSET
        self._active = True
        self._sense = ObjectiveSense(sense)
        self._priority = priority
        self._tolerance = tolerance

gurobi_direct (and probably other solvers I don't know of) could also make uses of these https://docs.gurobi.com/projects/optimizer/en/current/features/multiobjective.html

        
Thoughts @mrmundt @michaelbynum ?

jsiirola:
@quantresearch1, I am a little hesitant to add those as fields for the objective, as that is a pretty solver-specific thing.  What would you think about using Suffixes to provide that information?  Note that some solver interfaces already support a `priority` Suffix for specifying branching priorities, so this would seem to fit into the same idea.

quantresearch1:
Thanks @jsiirola  for your input. Good to know that's already done through suffixes, I was not aware of that. It makes sense to continue with suffixes. 
Eventually, I had in mind to not make lexicographic optimization solver specific by:
1) If a solver supports lexicographic optimization, pass on priority and rel/abs_tolerance as required
[...]

PR #3695, opened 2025-08-12
Model Observer

michaelbynum:
## Summary/Motivation:
This PR extracts the code for detecting changes in models from the persistent solver interfaces to its own, independent functionality. I will have a separate PR shortly that removes the code from the solver interfaces and updates them to use this code.

- add a module to contrib called `observer` that can be used to inform other classes of what changed in a pyomo model.

### Legal Acknowledgement

### DIFF
--- a/examples/pyomo/piecewise/indexed.py
+++ b/examples/pyomo/piecewise/indexed.py
@@ -9,7 +9,7 @@
 #  This software is distributed under the 3-clause BSD License.
 #  ___________________________________________________________________________
 
-# A piewise approximiation of a nonconvex objective function.
+# A piecewise approximation of a nonconvex objective function.
 
 import pyomo.environ as pyo
 
--- a/pyomo/contrib/doe/tests/test_doe_solve.py
+++ b/pyomo/contrib/doe/tests/test_doe_solve.py
@@ -251,7 +251,9 @@ class TestReactorExampleSolving(unittest.TestCase):
         # Make sure FIM and Q.T @ sigma_inv @ Q are close (alternate definition of FIM)
         self.assertTrue(np.all(np.isclose(FIM, Q.T @ sigma_inv @ Q)))
 
-    def test_reactor_obj_cholesky_solve_bad_prior(self):
+    def DISABLE_test_reactor_obj_cholesky_solve_bad_prior(self):
+        # [10/2025] This test has been disabled because it frequently
+        # (and randomly) returns "infeasible" when run on Windows.
         from pyomo.contrib.doe.doe import _SMALL_TOLERANCE_DEFINITENESS
 
         fd_method = "central"
--- a/pyomo/contrib/solver/common/results.py
+++ b/pyomo/contrib/solver/common/results.py
@@ -197,14 +197,6 @@ class Results(ConfigDict):
                 description="A tuple representing the version of the solver in use.",
             ),
         )
-        self.iteration_count: Optional[int] = self.declare(
-            'iteration_count',
-            ConfigValue(
-                domain=NonNegativeInt,
-                default=None,
-                description="The total number of iterations.",
-            ),
-        )
         self.timing_info: ConfigDict = self.declare(
             'timing_info', ConfigDict(implicit=True)
         )
--- a/pyomo/contrib/solver/solvers/gurobi_direct.py
+++ b/pyomo/contrib/solver/solvers/gurobi_direct.py
@@ -14,6 +14,7 @@ import io
 import math
 import operator
 import os
+import logging
 
 from pyomo.common.collections import ComponentMap, ComponentSet
 from pyomo.common.config import ConfigValue
@@ -43,6 +44,7 @@ from pyomo.contrib.solver.common.results import (
 )
 from pyomo.contrib.solver.common.solution_loader import SolutionLoaderBase
 
+logger = logging.getLogger(__name__)
 
 gurobipy, gurobipy_available = attempt_import('gurobipy')
 
@@ -442,7 +444,9 @@ class GurobiDirect(GurobiSolverMixin, SolverBase):
             results.incumbent_objective = None
             results.objective_bound = None
 
-        results.iteration_count = grb_model.getAttr('IterCount')
+        results.extra_info.IterCount = grb_model.getAttr('IterCount')
+        results.extra_info.BarIterCount = grb_model.getAttr('BarIterCount')
+        results.extra_info.NodeCount = grb_model.getAttr('NodeCount')
 
         timer.start('load solution')
         if config.load_solutions:
--- a/pyomo/contrib/solver/solvers/gurobi_persistent.py
+++ b/pyomo/contrib/solver/solvers/gurobi_persistent.py
@@ -857,7 +857,9 @@ class GurobiPersistent(
             ):
                 results.incumbent_objective = None
 
-        results.iteration_count = gprob.getAttr('IterCount')
+        results.extra_info.IterCount = gprob.getAttr('IterCount')
+        results.extra_info.BarIterCount = gprob.getAttr('BarIterCount')
+        results.extra_info.NodeCount = gprob.getAttr('NodeCount')
 
         timer.start('load solution')
         if config.load_solutions:
--- a/pyomo/contrib/solver/solvers/highs.py
+++ b/pyomo/contrib/solver/solvers/highs.py
@@ -750,7 +750,15 @@ class Highs(PersistentSolverMixin, PersistentSolverUtils, PersistentSolverBase):
                     results.objective_bound = None
             else:
                 results.objective_bound = info.mip_dual_bound
-            results.iteration_count = info.simplex_iteration_count
+
+            if info.valid:
+                results.extra_info.simplex_iteration_count = (
+                    info.simplex_iteration_count
+                )
+                results.extra_info.ipm_iteration_count = info.ipm_iteration_count
+                results.extra_info.mip_node_count = info.mip_node_count
+                results.extra_info.pdlp_iteration_count = info.pdlp_iteration_count
+                results.extra_info.qp_iteration_count = info.qp_iteration_count
 
         if config.load_solutions:
             if has_feasible_solution:
--- a/pyomo/contrib/solver/solvers/ipopt.py
+++ b/pyomo/contrib/solver/solvers/ipopt.py
@@ -473,7 +473,7 @@ class Ipopt(SolverBase):
                 results = Results()
                 results.termination_condition = TerminationCondition.provenInfeasible
                 results.solution_loader = SolSolutionLoader(None, None)
-                results.iteration_count = 0
+                results.extra_info.iteration_count = 0
                 results.timing_info.total_seconds = 0
             elif len(nl_info.variables) == 0:
                 if len(nl_info.eliminated_vars) == 0:
@@ -487,7 +487,7 @@ class Ipopt(SolverBase):
                     )
                     results.solution_status = SolutionStatus.optimal
                     results.solution_loader = SolSolutionLoader(None, nl_info=nl_info)
-                    results.iteration_count = 0
+                    results.extra_info.iteration_count = 0
                     results.timing_info.total_seconds = 0
             else:
                 if os.path.isfile(basename + '.sol'):
@@ -503,7 +503,9 @@ class Ipopt(SolverBase):
                     results.solution_loader = SolSolutionLoader(None, None)
                 else:
                     try:
-                        results.iteration_count = parsed_output_data.pop('iters')
+                        results.extra_info.iteration_count = parsed_output_data.pop(
+                            'iters'
+                        )
                         cpu_seconds = parsed_output_data.pop('cpu_seconds')
                         for k, v in cpu_seconds.items():
                             results.timing_info[k] = v
--- a/pyomo/contrib/solver/solvers/knitro/base.py
+++ b/pyomo/contrib/solver/solvers/knitro/base.py
@@ -82,7 +82,7 @@ class KnitroSolverBase(SolutionProvider, PackageChecker, SolverBase):
         if config.restore_variable_values_after_solve:
             self._save_var_values()
 
-        with capture_output(TeeStream(self._stream, *config.tee), capture_fd=False):
+        with capture_output(TeeStream(self._stream, *config.tee), capture_fd=True):
             self._solve(config, timer)
 
         if config.restore_variable_values_after_solve:
@@ -141,7 +141,7 @@ class KnitroSolverBase(SolutionProvider, PackageChecker, SolverBase):
         results.solution_status = self._get_solution_status(status)
         results.termination_condition = self._get_termination_condition(status)
         results.incumbent_objective = self._engine.get_obj_value()
-        results.iteration_count = self._engine.get_num_iters()
+        results.extra_info.iteration_count = self._engine.get_num_iters()
         results.timing_info.solve_time = self._engine.get_solve_time()
         results.timing_info.timer = timer
 
--- a/pyomo/contrib/solver/tests/solvers/test_ipopt.py
+++ b/pyomo/contrib/solver/tests/solvers/test_ipopt.py
@@ -594,13 +594,13 @@ class TestIpopt(unittest.TestCase):
         result = ipopt.Ipopt().solve(model, solver_options={'print_level': 0})
         # IPOPT doesn't tell us anything about the iters if the print level
         # is set to 0
-        self.assertIsNone(result.iteration_count)
+        self.assertFalse(hasattr(result.extra_info, 'iteration_count'))
         self.assertFalse(hasattr(result.extra_info, 'iteration_log'))
         model = self.create_model()
         result = ipopt.Ipopt().solve(model, solver_options={'print_level': 3})
         # At a slightly higher level, we get some of the info, like
         # iteration count, but NOT iteration_log
-        self.assertEqual(result.iteration_count, 11)
+        self.assertEqual(result.extra_info.iteration_count, 11)
         self.assertFalse(hasattr(result.extra_info, 'iteration_log'))
 
     def test_ipopt_loud_print_level(self):
@@ -609,13 +609,13 @@ class TestIpopt(unittest.TestCase):
             result = ipopt.Ipopt().solve(model, solver_options={'print_level': 8})
             # Nothing unexpected should be in the results object at this point,
             # except that the solver_log is significantly longer
-            self.assertEqual(result.iteration_count, 11)
+            self.assertEqual(result.extra_info.iteration_count, 11)
             self.assertEqual(result.incumbent_objective, 7.013645951336496e-25)
             self.assertIn('Optimal Solution Found', result.extra_info.solver_message)
             self.assertTrue(hasattr(result.extra_info, 'iteration_log'))
             model = self.create_model()
             result = ipopt.Ipopt().solve(model, solver_options={'print_level': 12})
-            self.assertEqual(result.iteration_count, 11)
+            self.assertEqual(result.extra_info.iteration_count, 11)
             self.assertEqual(result.incumbent_objective, 7.013645951336496e-25)
             self.assertIn('Optimal Solution Found', result.extra_info.solver_message)
             self.assertTrue(hasattr(result.extra_info, 'iteration_log'))
@@ -624,7 +624,7 @@ class TestIpopt(unittest.TestCase):
         model = self.create_model()
         results = ipopt.Ipopt().solve(model)
         self.assertEqual(results.solver_name, 'ipopt')
-        self.assertEqual(results.iteration_count, 11)
+        self.assertEqual(results.extra_info.iteration_count, 11)
         self.assertEqual(results.incumbent_objective, 7.013645951336496e-25)
         self.assertIn('Optimal Solution Found', results.extra_info.solver_message)
 
--- a/pyomo/contrib/solver/tests/solvers/test_knitro_direct.py
+++ b/pyomo/contrib/solver/tests/solvers/test_knitro_direct.py
@@ -9,6 +9,9 @@
 #  This software is distributed under the 3-clause BSD License.
 #  ___________________________________________________________________________
 
+import io
+from contextlib import redirect_stdout
+
 import pyomo.common.unittest as unittest
 
 from pyomo.contrib.solver.solvers.knitro.config import KnitroConfig
@@ -86,6 +89,32 @@ class TestKnitroDirectSolverInterface(unittest.TestCase):
 class TestKnitroDirectSolver(unittest.TestCase):
     def setUp(self):
         self.opt = KnitroDirectSolver()
+
+    def test_solve_tee(self):
+        m = pyo.ConcreteModel()
+        m.x = pyo.Var(initialize=1.5, bounds=(-5, 5))
+        m.y = pyo.Var(initialize=1.5, bounds=(-5, 5))
+        m.obj = pyo.Objective(
+            expr=(1.0 - m.x) + 100.0 * (m.y - m.x), sense=pyo.minimize
+        )
+        stream = io.StringIO()
+        with redirect_stdout(stream):
+            self.opt.solve(m, tee=True)
+        output = stream.getvalue()
+        self.assertTrue(bool(output.strip()))
+
+    def test_solve_no_tee(self):
+        m = pyo.ConcreteModel()
+        m.x = pyo.Var(initialize=1.5, bounds=(-5, 5))
+        m.y = pyo.Var(initialize=1.5, bounds=(-5, 5))
+        m.obj = pyo.Objective(
+            expr=(1.0 - m.x) + 100.0 * (m.y - m.x), sense=pyo.minimize
+        )
+        stream = io.StringIO()
+        with redirect_stdout(stream):
+            self.opt.solve(m, tee=False)
+        output = stream.getvalue()
+        self.assertFalse(bool(output.strip()))
 
     def test_solve(self):
         m = pyo.ConcreteModel()
--- a/pyomo/contrib/solver/tests/solvers/test_solvers.py
+++ b/pyomo/contrib/solver/tests/solvers/test_solvers.py
@@ -604,9 +604,6 @@ class TestSolvers(unittest.TestCase):
         for v in res.solver_version:
             self.assertIsInstance(v, int)
 
-        # iteration_count is nonnegative
-        self.assertGreaterEqual(res.iteration_count, 0)
-
         # timing_info should exist
         self.assertIsNotNone(res.timing_info)
 
--- a/pyomo/contrib/solver/tests/unit/test_results.py
+++ b/pyomo/contrib/solver/tests/unit/test_results.py
@@ -157,7 +157,6 @@ class TestResults(unittest.TestCase):
         expected_declared = {
             'extra_info',
             'incumbent_objective',
-            'iteration_count',
             'objective_bound',
             'solution_loader',
             'solution_status',
@@ -182,7 +181,6 @@ class TestResults(unittest.TestCase):
         self.assertEqual(res.solution_status, results.SolutionStatus.noSolution)
         self.assertIsNone(res.solver_name)
         self.assertIsNone(res.solver_version)
-        self.assertIsNone(res.iteration_count)
         self.assertIsInstance(res.timing_info, ConfigDict)
         self.assertIsInstance(res.extra_info, ConfigDict)
         self.assertIsNone(res.timing_info.start_timestamp)
@@ -198,7 +196,6 @@ incumbent_objective: None
 objective_bound: None
 solver_name: None
 solver_version: None
-iteration_count: None
 timing_info:
   start_timestamp: None
   wall_time: None


pyomo/contrib/observer/__init__.py
mrmundt, line 1: Needs the copyright statement!

pyomo/contrib/observer/tests/__init__.py
mrmundt, line 1: Needs the copyright statement!

pyomo/contrib/observer/tests/test_change_detector.py
```diff
@@ -0,0 +1,225 @@
+from pyomo.core.base.constraint import ConstraintData
```
mrmundt, line 16: Needs the copyright statement!

pyomo/contrib/observer/__init__.py
michaelbynum: Good catch. Thanks and done!

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,742 @@
...
+  - modified constraint expressions (relies on expressions being immutable)
+  - modified objective expressions (relies on expressions being immutable)
+  - modified objective sense
+  - changes to variable bounds, domains, and "fixed" flags
+  - changes to named expressions (relies on expressions being immutable)
+  - changes to parameter values and fixed variable values
+"""
+
+
+class AutoUpdateConfig(ConfigDict):
```
mrmundt, line 78: Is the intent that this lives here, not in `pyomo.contrib.solver.common.config`? If so, delete it there; otherwise, import it.

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,742 @@
...
+                description="""
+                If False, changes to objectives will not be automatically detected on 
+                subsequent solves. This includes the expr and sense attributes of objectives. Use 
+                False only when manually updating the solver with opt.set_objective() or when you are 
+                certain objectives are not being modified.""",
+            ),
+        )
+
+
+class Observer(abc.ABC):
```
mrmundt: Channeling John here - does this _need_ to inherit from `abc.ABC`? What does it get us that `NotImplemented` doesn't deliver?

Also, this is basically the same as part of `PersistentSolverBase`. Again, what is the long term intent? Will `PersistentSolverBase` end up inheriting from this?

pyomo/contrib/observer/tests/test_change_detector.py
```diff
@@ -0,0 +1,236 @@
...
+    d = {'add': 0, 'remove': 0, 'update': 0, 'set': 0}
+    return d
+
+
+class ObserverChecker(Observer):
+    def __init__(self):
+        super().__init__()
+        self.counts = ComponentMap()
+        """
+        counts is be a mapping from component (e.g., variable) to another
```
mrmundt: Grammar weirdness

pyomo/contrib/observer/tests/test_change_detector.py
```diff
@@ -0,0 +1,236 @@
...
+        # the constraint is removed and then
+        # added again when the constraint is added
+        expected[m.x]['update'] += 1
+        expected[m.x]['remove'] += 1
+        expected[m.x]['add'] += 1
+        expected[m.p]['remove'] += 1
+        expected[m.p]['add'] += 1
+        obs.check(expected)
+
+    def test_vars_and_params_elsewhere(self):
```
mrmundt, line 227: Incomplete intentionally?

pyomo/contrib/observer/model_observer.py
michaelbynum: All of the deletions from `pyomo.contrib.solver` related to this will be in a separate PR after I refactor those interfaces to use this. For now, it will have to be in both places.

pyomo/contrib/observer/model_observer.py
michaelbynum: I will eventually delete `PersistentSolverBase`.

michaelbynum: But not quite yet.

michaelbynum: `ABC` forces all derived classes to implement the specified methods.

michaelbynum: I really don't understand the dislike of `ABC`

pyomo/contrib/observer/tests/test_change_detector.py
michaelbynum: fixed!

pyomo/contrib/observer/tests/test_change_detector.py
michaelbynum: Oops. No. I'll add some tests here.

michaelbynum: done

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,745 @@
...
+            pid = id(p)
+            if pid in self._referenced_params:
+                raise ValueError(f'Parameter {p.name} has already been added')
+            self._referenced_params[pid] = [{}, {}, None]
+            self._params[id(p)] = (p, p.value)
+        for obs in self._observers:
+            obs.add_parameters(params)
+
+    def _check_for_new_vars(self, variables: List[VarData]):
+        new_vars = {}
```
yemine0x01: Isn't enough to make new_vars a list and append non referenced variables?

michaelbynum: Yes, you are right!

yemine0x01: I think also this pattern is repeated multiple times.

pyomo/contrib/observer/model_observer.py
mrmundt: We talked about this verbally last week, but TL;DR - it makes refactoring hard if there is repeated code. We can leave it alone for now but reserve the right to change it in the future if it causes issues.

pyomo/contrib/observer/tests/test_change_detector.py
```diff
@@ -0,0 +1,365 @@
...
+from pyomo.core.base.param import ParamData
+from pyomo.core.base.sos import SOSConstraintData
+from pyomo.core.base.var import VarData
+import pyomo.environ as pyo
+from pyomo.common import unittest
+from typing import List
+from pyomo.contrib.observer.model_observer import (
+    Observer,
+    ModelChangeDetector,
+    AutoUpdateConfig,
```
mrmundt, line 25: There aren't actually any tests that exercise `AutoUpdateConfig`. There also isn't any documentation that shows how to use it. Can you add a test (that can act as documentation)?

pyomo/contrib/observer/component_collector.py
```diff
@@ -0,0 +1,80 @@
...
+    ParamData: handle_param,
+    ScalarParam: handle_param,
+    ExpressionData: handle_named_expression,
+    ScalarExpression: handle_named_expression,
+    ExternalFunctionExpression: handle_external_function,
+    NPV_ExternalFunctionExpression: handle_external_function,
+}
+
+
+class _ComponentFromExprCollector(StreamBasedExpressionVisitor):
```
jsiirola, line 87: This implementation is fine, but you could consider a slightly more performant version might be:

```python
class _ComponentFromExprCollector(StreamBasedExpressionVisitor):
    def __init__(self, **kwds):
        self.named_expressions = {}
        self.variables = {}
        self.params = {}
        self.external_functions = {}
        self.handlers = {
            VarData: self.variables.__setitem__,
            ScalarVar: self.variables.__setitem__,
            ParamData: self.params.__setitem__,
            ScalarParam: self.params.__setitem__,
            ExpressionData: self.named_expressions.__setitem__,
            ScalarExpression: self.named_expressions.__setitem__,
            ExternalFunctionExpression: self.external_functions.__setitem__,
            NPV_ExternalFunctionExpression: self.external_functions.__setitem__,
        }
        super().__init__(**kwds)

    def exitNode(self, node, data):
        nt = type(node)
        if nt in self.handlers:
            self.handlers[nt](id(node), node)
        return None
```
...in that it saves several function calls.

Alternatively, you might consider moving to a "full dispatcher" model:
```python
    def exitNode(self, node, data):
        collector_handlers[node.__class__](self, node)
```

The dispatcher would then be based off a `defaultdict` and would end up with entries for native_types and (known) expressions.  The advantage with the "full dispatcher" model is that it will help "future-proof" the interface so that it doesn't silently ignore unknown types.

pyomo/contrib/observer/component_collector.py
```diff
@@ -0,0 +1,80 @@
...
+    ExternalFunctionExpression,
+    NPV_ExternalFunctionExpression,
+)
+from pyomo.core.base.var import VarData, ScalarVar
+from pyomo.core.base.param import ParamData, ScalarParam
+from pyomo.core.base.expression import ExpressionData, ScalarExpression
+
+
+def handle_var(node, collector):
+    collector.variables[id(node)] = node
```
jsiirola: This is basically reimplementing a version of `ComponentSet`.  Is there a reason not to re-use that object?

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,742 @@
...
+  - modified constraint expressions (relies on expressions being immutable)
+  - modified objective expressions (relies on expressions being immutable)
+  - modified objective sense
+  - changes to variable bounds, domains, and "fixed" flags
+  - changes to named expressions (relies on expressions being immutable)
+  - changes to parameter values and fixed variable values
+"""
+
+
+class AutoUpdateConfig(ConfigDict):
```
jsiirola, line 78: Leverage the `pyomo.common.config.document_config` decorator to add full documentation of the ConfigDict to the class:

```suggestion
@document_config
class AutoUpdateConfig(ConfigDict):
```

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,742 @@
...
+            doc = 'Configuration options to detect changes in model between solves'
+        super().__init__(
+            description=description,
+            doc=doc,
+            implicit=implicit,
+            implicit_domain=implicit_domain,
+            visibility=visibility,
+        )
+
+        self.check_for_new_or_removed_constraints: bool = self.declare(
```
jsiirola, line 102: Adding a short description (with the "magic" `#:` comment) will cause this to be listed in an `Attributes` table in the Sphinx docs:
```suggestion
        #: Automatically detect new/removed constraints on subsequent solves
        self.check_for_new_or_removed_constraints: bool = self.declare(
```

(If you accept this, you will want to add "short docstring" comments to the other config entries, too)

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,742 @@
...
+
+
+class ModelChangeDetector:
+    def __init__(self, observers: Sequence[Observer], **kwds):
+        """
+        Parameters
+        ----------
+        observers: Sequence[Observer]
+            The objects to notify when changes are made to the model
+        """
```
jsiirola, line 491: It would be helpful to expand this documentation.  In particular, to discuss the overall intent of the class, and what the "entry points" into the class are (for help in understanding the expected control flow).

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,742 @@
...
+        self._vars_referenced_by_con = {}
+        self._vars_referenced_by_obj = []
+        self._params_referenced_by_con = {}
+        self._params_referenced_by_obj = []
+        self._expr_types = None
+        self.config: AutoUpdateConfig = AutoUpdateConfig()(
+            value=kwds, preserve_implicit=True
+        )
+
+    def set_instance(self, model):
```
jsiirola: Doesn't this event need to be passed on to the observers?  How will they know to release the old model instance (if present)?

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,742 @@
...
+        config: AutoUpdateConfig = self.config(value=kwds, preserve_implicit=True)
+
+        added_cons = set()
+        added_sos = set()
+
+        if config.check_for_new_or_removed_constraints:
+            timer.start('sos')
+            new_sos, old_sos = self._check_for_new_or_removed_sos()
+            self._add_sos_constraints(new_sos)
+            self._remove_sos_constraints(old_sos)
```
jsiirola: Wouldn't it make sense to only call those methods if `new_sos` (or `old_sos`) were nonempty?

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,742 @@
...
+        self._active_constraints = {}  # maps constraint to expression
+        self._active_sos = {}
+        self._vars = {}  # maps var id to (var, lb, ub, fixed, domain, value)
+        self._params = {}  # maps param id to param
+        self._objective = None
+        self._objective_expr = None
+        self._objective_sense = None
+        self._named_expressions = (
+            {}
+        )  # maps constraint to list of tuples (named_expr, named_expr.expr)
```
jsiirola: Black did some really silly formatting here.  Would you object to moving these comments to the line before the declaration (both here and a couple below)?

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,742 @@
...
+        return update_obj
+
+    def update(self, timer: Optional[HierarchicalTimer] = None, **kwds):
+        if timer is None:
+            timer = HierarchicalTimer()
+        config: AutoUpdateConfig = self.config(value=kwds, preserve_implicit=True)
+
+        added_cons = set()
+        added_sos = set()
+
```
jsiirola: Because of the large number of objects that are getting created / released (particularly during the first model build), it might be useful to pause the garbage collector for the duration of this method.

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,742 @@
...
+        if self._objective is None:
+            return update_obj
+        if self._objective.expr is not self._objective_expr:
+            update_obj = True
+        elif self._objective.sense != self._objective_sense:
+            # we can definitely do something faster here than resetting the whole objective
+            update_obj = True
+        return update_obj
+
+    def update(self, timer: Optional[HierarchicalTimer] = None, **kwds):
```
jsiirola, line 1203: It would be useful to add some documentation about the design philosophy here.  For example, why add before remove, and what order you look at the various modeling components.

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,742 @@
...
+            self._objective = None
+            self._objective_expr = None
+            self._objective_sense = None
+            self._obj_named_expressions = []
+        for obs in self._observers:
+            obs.set_objective(obj)
+        self._check_to_remove_vars(vars_to_remove_check)
+        self._check_to_remove_params(params_to_remove_check)
+
+    def _add_block(self, block):
```
jsiirola: This method name is a bit misleading (it is not called for every block, but only when we are setting a new instance).

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,742 @@
...
+    def add_constraints(self, cons: List[ConstraintData]):
+        pass
+
+    @abc.abstractmethod
+    def add_sos_constraints(self, cons: List[SOSConstraintData]):
+        pass
+
+    @abc.abstractmethod
+    def set_objective(self, obj: Optional[ObjectiveData]):
+        pass
```
jsiirola: Explicit in this is that you will only support single-objective problems.  It would be better if we support multiple objectives as it is often requested (for solvers we already support), e.g. #3454, [here](https://stackoverflow.com/questions/50742999/multi-objective-optimization-example-pyomo), [here](https://www.linkedin.com/pulse/multi-objective-optimization-pyomo-alireza-soroudi), ...

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,742 @@
...
+
+    def _add_constraints(self, cons: List[ConstraintData]):
+        vars_to_check = []
+        params_to_check = []
+        for con in cons:
+            if con in self._active_constraints:
+                raise ValueError(f'Constraint {con.name} has already been added')
+            self._active_constraints[con] = con.expr
+            tmp = collect_components_from_expr(con.expr)
+            named_exprs, variables, parameters, external_functions = tmp
```
jsiirola: Ideally these two lines would be combined (it saves an assignment and fetch from the locals()).

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,742 @@
...
+            vars_to_check.extend(variables)
+            params_to_check.extend(parameters)
+            self._named_expressions[con] = [(e, e.expr) for e in named_exprs]
+            if len(external_functions) > 0:
```
jsiirola: The following is slightly more performant (and more clear):

```suggestion
            if external_functions:
```

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,742 @@
...
+        params_to_check = []
+        for con in cons:
+            if con in self._active_constraints:
+                raise ValueError(f'Constraint {con.name} has already been added')
+            self._active_constraints[con] = con.expr
+            tmp = collect_components_from_expr(con.expr)
+            named_exprs, variables, parameters, external_functions = tmp
+            vars_to_check.extend(variables)
+            params_to_check.extend(parameters)
+            self._named_expressions[con] = [(e, e.expr) for e in named_exprs]
```
jsiirola: Is it worth following the example of `external_expressions` and only add an entry if there actually are named expressions?  As written you will allocate a list and an entry in the dict for every constraint in the model (that could be a lot of memory...).

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,742 @@
...
+
+    def _add_variables(self, variables: List[VarData]):
+        for v in variables:
+            if id(v) in self._referenced_variables:
+                raise ValueError(f'Variable {v.name} has already been added')
+            self._referenced_variables[id(v)] = [{}, {}, None]
+            self._vars[id(v)] = (
+                v,
+                v._lb,
+                v._ub,
```
jsiirola: We should consider adding a public `raw_bound_exprs` (or similar) to VarData so that this doesn't need to access the private data?

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,742 @@
...
+        self.__init__(observers=self._observers)
+        self.config = saved_config
+        self._model = model
+        self._add_block(model)
+
+    def _add_variables(self, variables: List[VarData]):
+        for v in variables:
+            if id(v) in self._referenced_variables:
+                raise ValueError(f'Variable {v.name} has already been added')
+            self._referenced_variables[id(v)] = [{}, {}, None]
```
jsiirola: Is it worth making the default value be `[{}, None, None]`?  SOS components aren't terribly common, so this could potentially save a lot of memory (and object allocations).

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,742 @@
...
+            )
+        for obs in self._observers:
+            obs.add_variables(variables)
+
+    def _add_parameters(self, params: List[ParamData]):
+        for p in params:
+            pid = id(p)
+            if pid in self._referenced_params:
+                raise ValueError(f'Parameter {p.name} has already been added')
+            self._referenced_params[pid] = [{}, {}, None]
```
jsiirola: Is it worth making the default value be `[{}, None, None]`?  SOS components aren't terribly common, so this could potentially save a lot of memory (and object allocations).

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,742 @@
...
+        )  # maps constraint to list of tuples (named_expr, named_expr.expr)
+        self._external_functions = ComponentMap()
+        self._obj_named_expressions = []
+        self._referenced_variables = (
+            {}
+        )  # var_id: [dict[constraints, None], dict[sos constraints, None], None or objective]
+        self._referenced_params = (
+            {}
+        )  # param_id: [dict[constraints, None], dict[sos constraints, None], None or objective]
+        self._vars_referenced_by_con = {}
```
jsiirola: I wonder if a more efficient data structure would be:
```python
# var_id: [list[tuple[constraint, int]], list[tuple[sos constraints, int]], None or objective]
self._referenced_variables = {}
# constraint: list[tuple[var_id, int]]
self._vars_referenced_by_con = {}
```
The idea is that the `int` in each would point to the index in the other list.  Deletion could then be done by setting that entry in the list to `None`.  The performance improvement would be avoiding the need to hash things, plus saving the ~1k for each nonempty dict.

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,742 @@
...
+            if v_id not in self._referenced_variables:
+                new_vars[v_id] = v
+        self._add_variables(list(new_vars.values()))
+
+    def _check_to_remove_vars(self, variables: List[VarData]):
+        vars_to_remove = {}
+        for v in variables:
+            v_id = id(v)
+            ref_cons, ref_sos, ref_obj = self._referenced_variables[v_id]
+            if len(ref_cons) == 0 and len(ref_sos) == 0 and ref_obj is None:
```
jsiirola: The following would be more performant (and a little more clear):
```suggestion
            if not ref_cons and not ref_sos and ref_obj is None:
```

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,742 @@
...
+                if self._referenced_variables[vid][2] is not None:
+                    update_obj = True
+            elif v._lb is not _lb:
+                vars_to_update.append(v)
+            elif v._ub is not _ub:
+                vars_to_update.append(v)
+            elif _domain_interval != v.domain.get_interval():
+                vars_to_update.append(v)
+            elif v.value != _value:
+                vars_to_update.append(v)
```
jsiirola: Don't you also have to update a constraint if the var is fixed?

pyomo/contrib/observer/component_collector.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+def handle_param(node, collector):
+    collector.params[id(node)] = node
+    return None
+
+
+def handle_named_expression(node, collector):
+    collector.named_expressions[id(node)] = node
+    return None
```
jsiirola, line 51: Named expressions are frequently nested (particularly for things like IDAES models).  It might be good to preserve that nested structure so that we don't have to process the same subexpression twice.

(Anecdotally, doing that in the NL writer made for a significant performance improvement for IDAES models)

jsiirola (commented):
Overall, this is great.  I have lots of questions, but very few should prohibit merging this.

jsiirola:
Additional question: who is responsible for verifying that the model does not contain any active components that the Solver/Observer doesn't know how to handle?  This appears to assume that the responsibility lies with the Solver, but does that mean that the solver needs to also implement it's own observer to make sure that "unallowable" active components aren't added between solves?  That makes me think that maybe the responsibility needs to lie in this observer...

pyomo/contrib/observer/component_collector.py
michaelbynum: I have found that doing this manually is faster. That is the only reason. I'm not sure if the difference is enough to justify or not.

michaelbynum: I'll try to get some numbers.

pyomo/contrib/observer/component_collector.py
michaelbynum: Ahh. Good idea. I'll try that out.

pyomo/contrib/observer/component_collector.py
michaelbynum: I have not looked at the dispatcher before. Would the use case here be if someone created a class that derived from `VarData`?

pyomo/contrib/observer/model_observer.py
michaelbynum: This is so cool!

pyomo/contrib/observer/component_collector.py
jsiirola: You are basically implementing a dispatcher in exitNode:
```python
    def exitNode(self, node, data):
        nt = type(node)
        if nt in collector_handlers:
            return collector_handlers[nt](node, self)
        return None
```

The thing I am worried about is really the "if without an else": you are looking for a specific set of objects and implicitly ignoring the rest.  If new objects are added to the tree that you aren't currently looking for, then you will (silently) miss them.  By passing everything through the dispatcher, new / unexpected types won't be silently ignored.

A good example you could copy is `pyomo.repn.util.BeforeChildDispatcher`: basically derive from `defaultdict` and implement `__missing__`.

Incidentally, it will be more performant to implement this in `beforeChild` instead of `exitNode`: at least 1/2 the nodes in the tree are leaf nodes, and beforeChild lets you short-circuit the traversal and avoid incurring the overhead of entering / leaving the leaf node contexts.

michaelbynum: Thanks and done!

pyomo/contrib/observer/component_collector.py
michaelbynum: This seems to be about twice as fast.

pyomo/contrib/observer/model_observer.py
michaelbynum: Done.

pyomo/contrib/observer/model_observer.py
michaelbynum: Done.

pyomo/contrib/observer/model_observer.py
michaelbynum: Good call. I removed this method. The model must now be passed to the constructor, and a new change detector has to get created if there is a new model.

pyomo/contrib/observer/model_observer.py
michaelbynum: Good call. Done.

pyomo/contrib/observer/model_observer.py
michaelbynum: Done.

pyomo/contrib/observer/model_observer.py
michaelbynum: Done.

pyomo/contrib/observer/model_observer.py
michaelbynum: Done.

pyomo/contrib/observer/model_observer.py
michaelbynum: Fixed.

pyomo/contrib/observer/model_observer.py
michaelbynum: Fixed.

pyomo/contrib/observer/model_observer.py
michaelbynum: Done.

pyomo/contrib/observer/model_observer.py
michaelbynum: Done.

pyomo/contrib/observer/model_observer.py
michaelbynum: Done.

pyomo/contrib/observer/model_observer.py
michaelbynum: Yes, this would be excellent!

pyomo/contrib/observer/model_observer.py
michaelbynum: In order to handle multiple objectives, I had to make the third element a dictionary as well. If I make it None, then I have to check repeatedly if any constraints/objectives use the variable. Instead, I made the list a tuple (tuples take far less memory than lists) as a compromise.

pyomo/contrib/observer/model_observer.py
michaelbynum: Same as above.

pyomo/contrib/observer/model_observer.py
michaelbynum: I'm going to push this off to future work.

pyomo/contrib/observer/model_observer.py
michaelbynum: I think I fixed all of these...

pyomo/contrib/observer/model_observer.py
michaelbynum: I think this is up to the observer... Worth a discussion.

michaelbynum:
> Additional question: who is responsible for verifying that the model does not contain any active components that the Solver/Observer doesn't know how to handle? This appears to assume that the responsibility lies with the Solver, but does that mean that the solver needs to also implement it's own observer to make sure that "unallowable" active components aren't added between solves? That makes me think that maybe the responsibility needs to lie in this observer...

I added a check for unrecognized active components.

pyomo/contrib/observer/model_observer.py
michaelbynum: I'm not sure what the best solution is for this. I think we need more use cases. I vote we leave it for now. Here are a couple of examples that make this challenging:

```
m.x * (m.y + m.z) == 0
m.x.fix(1)
opt.solve(m)
m.x.value = 2
```

Right now, this works fine for the existing persistent solver interfaces because changing the value of `m.x` does not change the "structure" of the constraint (i.e., it remains linear). Essentially, we create a "mutable linear coefficient" and update that coefficient when the value of the variable changes. Furthermore, this does not require reconstructing the constraint in any way. However, 

```
(m.y + m.z) ** m.x == 0
m.x.fix(1)
```

will currently just result in an exception for both gurobi persistent and highs persistent (scip is fine).

pyomo/contrib/observer/model_observer.py
michaelbynum: Done.

pyomo/contrib/observer/tests/test_change_detector.py
michaelbynum: I added some documentation in the docstring for `ModelChangeDetector`, but I'm still working on tests.

pyomo/contrib/observer/model_observer.py
michaelbynum: Done.

pyomo/contrib/observer/component_collector.py
michaelbynum: Done.

pyomo/contrib/observer/tests/test_change_detector.py
michaelbynum: Done!

michaelbynum:
I think I have addressed everything, @jsiirola and @mrmundt. Should be ready to go.

pyomo/contrib/observer/model_observer.py
michaelbynum: I've been thinking about this more. Maybe we can add some public methods to `ModelChangeDetector` to provide more information if necessary. For example, we could add a method to get all of the constraints used by a variable (the change detector already has that data).

pyomo/contrib/observer/model_observer.py
mrmundt: You missed the "magic" part - `#:`

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1003 @@
...
+        )
+
+
+class Observer(abc.ABC):
+    def __init__(self):
+        pass
+
+    @abc.abstractmethod
+    def add_variables(self, variables: List[VarData]):
+        pass
```
mrmundt: You may want to put docstrings on all of these methods. It'll make them nicer looking when the API docs are rendered. Right now the table is just so... blank: https://pyomo--3695.org.readthedocs.build/en/3695/api/pyomo.contrib.observer.model_observer.Observer.html#pyomo.contrib.observer.model_observer.Observer

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1003 @@
...
+    changes to the model are made (but only when ModelChangeDetector.update
+    is called). An example use case is for the persistent solver interfaces.
+
+    The ModelChangeDetector considers the model to be defined by its set of
+    active components and any components used by those active components. For
+    example, the observers will not be notified of the addition of a variable
+    if that variable is not used in any constraints.
+
+    The Observer/ModelChangeDetector are most useful when a small number
+    of changes are being relative to the size of the model. For example,
```
mrmundt:
```suggestion
    of changes are being made relative to the size of the model. For example,
```
Grammar weirdness - I think this is what you mean?

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1003 @@
...
+        #     dict[sos constraints, None],
+        #     dict[objectives, None],
+        # )
+        self._referenced_variables = {}
+
+        # param_id: (
+        #     dict[constraints, None],
+        #     dict[sos constraints, None],
+        #     dict[objectives, None],
+        # )
```
mrmundt: Are these comments here for reference, or are they leftover code?

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1003 @@
...
+        Check for changes to the model and notify the observers.
+
+        Parameters
+        ----------
+        timer: Optional[HierarchicalTimer]
+            The timer to use for tracking how much time is spent detecting
+            different kinds of changes
+        """
+
+        """
```
mrmundt, line 1214: This feels like a leftover weirdness. Remove?

pyomo/contrib/observer/tests/test_component_collector.py
```diff
@@ -0,0 +1,20 @@
+import pyomo.environ as pyo
```
mrmundt, line 12: Need copyright statement please!

pyomo/contrib/observer/model_observer.py
michaelbynum: Oh, these describe the data structures. I suppose I could actually use type hinting.

michaelbynum: Well, type hinting here would actually be less informative

pyomo/contrib/observer/model_observer.py
michaelbynum: This is intentional. I wanted to separate the docstring meant for users from the comments that are meant for developers below.

pyomo/contrib/observer/model_observer.py
michaelbynum: Thanks! I definitely missed that there was supposed to be a colon.

michaelbynum: Fixed!

pyomo/contrib/observer/model_observer.py
michaelbynum: Done!

pyomo/contrib/observer/model_observer.py
michaelbynum: Fixed!

mrmundt (approved):
This looks good! Thanks for doing all the fixes.

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1131 @@
...
+            if not issubclass(ctype, ActiveComponent):
+                continue
+            if ctype in self._known_active_ctypes:
+                continue
+            for comp in self._model.component_data_objects(
+                ctype, active=True, descend_into=True
+            ):
+                raise NotImplementedError(
+                    f'ModelChangeDetector does not know how to '
+                    'handle compents with ctype {ctype}'
```
yemine0x01: typo: shouldn't it be components instead of compents?

mrmundt: Ooo, reporting upstream, thank you.

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1131 @@
...
+            self._check_for_unknown_active_components()
+
+            self._add_constraints(
+                list(
+                    self._model.component_data_objects(
+                        Constraint, descend_into=True, active=True
+                    )
+                )
+            )
+            self._add_sos_constraints(
```
yemine0x01: Some solvers do not support sos constraint. How about making this optional?

yemine0x01:
LGTM. One minor comment but not that important.

pyomo/contrib/observer/model_observer.py
michaelbynum: @jsiirola and I discussed this further and agreed that expanding the interface was a good solution. I added to the docstring explaining that we can expand the interface, but we are waiting to do so until we have a need.

pyomo/contrib/observer/model_observer.py
michaelbynum: Fixed. Thanks!

pyomo/contrib/observer/model_observer.py
michaelbynum: I think that the derived classes can choose to raise `NotImplementedError`. I'd rather they do so consciously.

michaelbynum:
Okay, if tests pass, I think we can merge this.

pyomo/contrib/observer/component_collector.py
jsiirola: OK - but we ought to see *why* ComponentSet is slower and try to fix it there.

pyomo/contrib/observer/component_collector.py
```diff
@@ -0,0 +1,114 @@
...
+        self.named_expressions = {}
+        self.variables = {}
+        self.params = {}
+        self.external_functions = {}
+        super().__init__(**kwds)
+
+    def exitNode(self, node, data):
+        return collector_handlers[node.__class__](node, self)
+
+    def beforeChild(self, node, child, child_idx):
```
jsiirola, line 98: (I mentioned this in a [previous comment](https://github.com/Pyomo/pyomo/pull/3695#discussion_r2319033676), so repeating here so it is not lost for future development / reference)

For performance reasons, it may be good to add a BeforeChildDispatcher as well (especially because this collector is specifically looking for leaf nodes).  

That said, this improvement can be deferred to a subsequent PR.

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1141 @@
...
+from pyomo.core.base.component import ActiveComponent
+from pyomo.common.collections import ComponentMap
+from pyomo.common.timing import HierarchicalTimer
+from pyomo.contrib.solver.common.util import get_objective
+from pyomo.contrib.observer.component_collector import collect_components_from_expr
+from pyomo.common.numeric_types import native_numeric_types
+import gc
+
+
+"""
```
jsiirola: If you either move this before the imports, or make the following change,this string will be picked up by the RTD API docs:
```suggestion
__doc__ = """
```

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1141 @@
...
+            if id(v) not in self._referenced_variables:
+                new_vars.append(v)
+        self._add_variables(new_vars)
+
+    def _check_to_remove_vars(self, variables: List[VarData]):
+        vars_to_remove = []
+        for v in variables:
+            v_id = id(v)
+            ref_cons, ref_sos, ref_obj = self._referenced_variables[v_id]
+            if not ref_cons and not ref_sos and not ref_obj:
```
jsiirola: I realize that I actually suggested the current syntax, but thinking more on it, it might nice to consider (in the future) the following:
```python
            if not any(self._referenced_variables[v_id]):
```
(same functionality, possible clearer, and avoids the creation / access / destruction of unnecessary entries in `locals()`;  unknown if saving the intermediate variables is faster / slowed than the call to `any()`)

This idiom can be used here and numerous places below.

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1141 @@
...
+            ):
+                raise NotImplementedError(
+                    f'ModelChangeDetector does not know how to '
+                    'handle components with ctype {ctype}'
+                )
+
+    def _set_instance(self):
+
+        is_gc_enabled = gc.isenabled()
+        gc.disable()
```
jsiirola: Please use the PauseGC context manager (instead of re-implementing that logic here)

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1141 @@
...
+                continue
+            if ctype in self._known_active_ctypes:
+                continue
+            for comp in self._model.component_data_objects(
+                ctype, active=True, descend_into=True
+            ):
+                raise NotImplementedError(
+                    f'ModelChangeDetector does not know how to '
+                    'handle components with ctype {ctype}'
+                )
```
jsiirola: This is (could be) redundant: `collect_ctypes(active=True)` already contains this check.

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1141 @@
...
+            del self._objectives[obj_id]
+            self._obj_named_expressions.pop(obj_id, None)
+            self._external_functions.pop(obj, None)
+            del self._vars_referenced_by_obj[obj_id]
+            del self._params_referenced_by_obj[obj_id]
+        self._check_to_remove_vars(vars_to_check)
+        self._check_to_remove_params(params_to_check)
+
+    def _check_for_unknown_active_components(self):
+        for ctype in self._model.collect_ctypes():
```
jsiirola:
```suggestion
        for ctype in self._model.collect_ctypes(active=True, descend_into=True):
```

pyomo/contrib/observer/model_observer.py
jsiirola: In that case, I would make this a big comment block and not a second multi-line string

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1141 @@
...
+        We check for changes to constraints/objectives before variables/parameters 
+        so that we don't waste time updating a variable/parameter that is going to 
+        get removed.
+        """
+        if timer is None:
+            timer = HierarchicalTimer()
+        config: AutoUpdateConfig = self.config(value=kwds, preserve_implicit=True)
+
+        is_gc_enabled = gc.isenabled()
+        gc.disable()
```
jsiirola: Again, I would recommend (re)using `PauseGC`

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1141 @@
...
+  - new constraints that have been added to the model
+  - constraints that have been removed from the model
+  - new variables that have been detected in new or modified constraints/objectives
+  - old variables that are no longer used in any constraints/objectives
+  - new parameters that have been detected in new or modified constraints/objectives
+  - old parameters that are no longer used in any constraints/objectives
+  - new objectives that have been added to the model
+  - objectives that have been removed from the model
+  - modified constraint expressions (relies on expressions being immutable)
+  - modified objective expressions (relies on expressions being immutable)
```
jsiirola: It is still weird to me that we make the client remember the mapping from Variable / Param values if they need to manually update constraints (e.g., due to fixed variables / Params changing values).  This is implicit in this documentation, too: we will update the user on *structural* changes to the expressions, but not numeric changes ... and if a numeric change causes a structural change, that is (hopefully) caught by the client (if noly they know they need to look for it)?

Can we at least make these mappings public so that the client can reuse them?

pyomo/contrib/observer/component_collector.py
michaelbynum: I don't disagree, but I don't think that needs to be part of the PR. We can always change the underlying implementation of this later. None of this is exposed to the user.

pyomo/contrib/observer/model_observer.py
michaelbynum: Does this need to be part of this PR? I have no doubt that change will require a great deal of discussion on its own.

pyomo/contrib/observer/component_collector.py
michaelbynum: I would prefer to do this as a separate PR.

pyomo/contrib/observer/model_observer.py
michaelbynum: I thought we discussed this? We can expand the API as needed to support this.

michaelbynum: I don't think we can make the mappings public because they should not be modified. Instead we can add methods for clients to request the information they need. I don't think that needs done as part of this PR.

pyomo/contrib/observer/model_observer.py
michaelbynum: I did not intend this to be user documentation. The docstring for `ModelChangeDetector` is mean to be the documentation.

pyomo/contrib/observer/model_observer.py
michaelbynum: Good catch!

pyomo/contrib/observer/model_observer.py
michaelbynum: Done.

pyomo/contrib/observer/model_observer.py
michaelbynum: Done.

pyomo/contrib/observer/model_observer.py
michaelbynum: Done.

michaelbynum:
Okay, should be good to go, @jsiirola (assuming tests pass).

michaelbynum:
Okay, I did a fairly substantial refactor of this. The `ModelChangeDetector` now also tells the `Observer` the reason for the change instead of just the fact that something changed. For example, if a variable was modified, there is an enum to specify what part of the variable changed (e.g., bounds, domain, etc.). This refactor also makes some improvements to the logic so that we don't remove and add a modified constraint (in the past, that could potentially also lead to removing and re-adding variables unnecessarily). 

I'm still working on adding more tests, but I think this should be ready for review.

michaelbynum (commented):
~~Add a test for nested Expressions~~ (done)

michaelbynum:
I think I am finally done with this PR.

michaelbynum:
I think that codecov is confused. Coverage should be over 90%.

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1294 @@
...
+
+        self._model: BlockData = model
+        self._set_instance()
+
+    def _add_variables(self, variables: Collection[VarData]):
+        for v in variables:
+            if v in self._referenced_variables:
+                raise ValueError(f'Variable {v.name} has already been added')
+            self._updates.vars_to_update[v] |= Reason.added
+            self._referenced_variables[v] = (OrderedSet(), OrderedSet(), ComponentSet())
```
jsiirola, line 573: The code will be more readable / maintainable if we use a `namedtuple` here

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1294 @@
...
+                self._check_for_new_params(ref_params)
+                for p in ref_params:
+                    self._referenced_params[p][3].add(v)
+
+    def _add_parameters(self, params: Collection[ParamData]):
+        for p in params:
+            if p in self._referenced_params:
+                raise ValueError(f'Parameter {p.name} has already been added')
+            self._updates.params_to_update[p] |= Reason.added
+            self._referenced_params[p] = (
```
jsiirola, line 606: The code will be more readable / maintainable if we use a `namedtuple` here

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1294 @@
...
+class _Updates:
+    def __init__(self, observers: Collection[Observer]) -> None:
+        self.vars_to_update = DefaultComponentMap(_default_reason)
+        self.params_to_update = DefaultComponentMap(_default_reason)
+        self.cons_to_update = defaultdict(_default_reason)
+        self.sos_to_update = defaultdict(_default_reason)
+        self.objs_to_update = DefaultComponentMap(_default_reason)
+        self.observers = observers
+
+    def run(self):
```
jsiirola, line 309: Stylistic nit, but would `notify()` be more descriptive than `run()`?

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1294 @@
...
+    def _update_variables(self, variables: Optional[Collection[VarData]] = None):
+        if variables is None:
+            variables = self._vars
+        for v in variables:
+            _lb, _ub, _fixed, _domain_interval, _value = self._vars[v]
+            lb, ub, fixed, domain_interval, value = (
+                v._lb,
+                v._ub,
+                v.fixed,
+                v.domain.get_interval(),
```
jsiirola, line 907: This is OK for now, but:
- will break down as soon as we have solver interfaces that support anything other than simple bounded domains (e.g., this won't work for semi-continuous domains)
- `get_interval()` can be inefficient for custom sets

The long-term solution is to probably maintain a list of known domains and perform change detection on them (much like we do for Params).  The short-term solution is probably to document this issue in a comment here.

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1294 @@
...
+        if params is None:
+            params = self._params
+        for p in params:
+            _val = self._params[p]
+            val = p.value
+            reason = Reason.no_change
+            if _val != val:
+                reason |= Reason.value
+            if reason:
+                self._updates.params_to_update[p] |= reason
```
jsiirola, line 939: Cant this all be simplified to:
```suggestion
            if _val != val:
                self._updates.params_to_update[p] |= Reason.value
```

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1294 @@
...
+                self._params[p] = val
+
+    def update_parameters(self, params: Optional[Collection[ParamData]]):
+        self._update_parameters(params)
+        self._updates.run()
+
+    def _update_con(self, con: ConstraintData):
```
jsiirola, line 946: Is `update` the right verb here?  Would `_record_constraint` be more descriptive, and differentiate this method from `_update_constraints` below?

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1294 @@
...
+
+    def _update_constraints(self, cons: Optional[Collection[ConstraintData]] = None):
+        if cons is None:
+            cons = self._active_constraints
+        for c in cons:
+            reason = Reason.no_change
+            if c.expr is not self._active_constraints[c]:
+                reason |= Reason.expr
+            if reason:
+                self._updates.cons_to_update[c] |= reason
```
jsiirola, line 997: Can't this be simplified to:
```suggestion
            if c.expr is not self._active_constraints[c]:
                self._updates.cons_to_update[c] |= Reason.expr
```

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1294 @@
...
+                )
+            self._updates.params_to_update[p] |= Reason.removed
+            if any(self._referenced_params[p]):
+                raise ValueError(
+                    f'Cannot remove parameter {p.name} - it is still being used by constraints/objectives'
+                )
+            self._referenced_params.pop(p)
+            self._params.pop(p)
+
+    def _update_var_bounds(self, v: VarData):
```
jsiirola, line 859: Is `update` the right verb here?  Would `_record_var_bounds` be more descriptive?

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1294 @@
...
+                    if v1 is not v2:
+                        needs_update = True
+                        break
+                for p1, p2 in zip(_plist, plist):
+                    if p1 is not p2:
+                        needs_update = True
+                        break
+            if needs_update:
+                reason |= Reason.sos_items
+                self._updates.sos_to_update[c] |= reason
```
jsiirola, line 1070: As above, can't this be simplified to remove the `reason` local variable:
```suggestion
                self._updates.sos_to_update[c] |= Reason.sos_items
```

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1294 @@
...
+                self._update_sos_con(c)
+
+    def update_sos_constraints(
+        self, cons: Optional[Collection[SOSConstraintData]] = None
+    ):
+        self._update_sos_constraints(cons)
+        self._updates.run()
+
+    def _update_obj_expr(self, obj: ObjectiveData):
```
jsiirola, line 1079: As above, should the verb be changed from `update` to `record`?

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1294 @@
...
+
+        # maps param to value
+        self._params: MutableMapping[ParamData, float] = ComponentMap()
+
+        self._objectives: MutableMapping[
+            ObjectiveData, Tuple[Union[NumericValue, float, int, None], ObjectiveSense]
+        ] = ComponentMap()  # maps objective to (expression, sense)
+
+        # maps constraints/objectives to list of tuples (named_expr, named_expr.expr)
+        self._named_expressions: MutableMapping[
```
jsiirola, line 512: Why is this `_named_expressions` and not `_con_named_expressions`?

pyomo/contrib/observer/model_observer.py
```diff
@@ -0,0 +1,1294 @@
...
+        self._vars: MutableMapping[VarData, Tuple] = ComponentMap()
+
+        # maps param to value
+        self._params: MutableMapping[ParamData, float] = ComponentMap()
+
+        self._objectives: MutableMapping[
+            ObjectiveData, Tuple[Union[NumericValue, float, int, None], ObjectiveSense]
+        ] = ComponentMap()  # maps objective to (expression, sense)
+
+        # maps constraints/objectives to list of tuples (named_expr, named_expr.expr)
```
jsiirola, line 511: I think this implementation is OK for now, but named expressions logically form a tree.  By flattening the tree here the observer will end up performing repeated work any time an expression is shared.  Maybe not a huge issue for Gurobi, but this will be a bigger deal for things like IDAES / ipopt models.

I would recommend adding a TODO to the code to remind us that eventually we should consider updating the internals to store that hierarchy so that we never need to walk any (sub) expression more than once.

jsiirola (approved):
Overall, this looks really good.  Some minor questions / nits that would be nice to discuss / implement, but nothing that needs to block merging.

michaelbynum:
Thanks for the feedback, @jsiirola. These are all great suggestions. As we discussed, I am going to go ahead an hit merge, so that #3698 can be reviewed more easily. I'll start working on these changes in a separate PR.

PR #3698, opened 2025-08-12
Observer gurobi refactor

michaelbynum:
## Summary/Motivation:
This PR refactors the gurobi interfaces to achieve two goals:
- reduce duplicate code
- use the new contrib.observer package

- create a directory for gurobi now that there are 4 files
- create a base class for all of the interfaces that use gurobipy, making the direct and persistent interfaces more consistent (e.g., tc_map, postsolve, solve, mipstart)
- modify the persistent interface to use the contrib.observer package
- add mipstart functionality to gurobi direct and gurobi minlp
- add a minimum gurobipy version to the direct interface (addMConstr was not added until version 9)
- simplified some of the persistent interface

### Legal Acknowledgement


pyomo/contrib/solver/plugins.py
```diff
@@ -39,7 +39,7 @@ def load():
         doc='Direct interface to Gurobi accommodating general MINLP',
     )(GurobiDirectMINLP)
     SolverFactory.register(
-        name="highs", legacy_name="highs", doc="Persistent interface to HiGHS"
+        name="highs", legacy_name="highs_v2", doc="Persistent interface to HiGHS"
```
jsiirola: Why is this being renamed?  There wasn't a `highs`, so it doesn't make a lot of sense to change this to `v2`?

pyomo/contrib/solver/solvers/gurobi/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .gurobi_direct import GurobiDirect
+from .gurobi_persistent import GurobiPersistent
+from .gurobi_direct_minlp import GurobiDirectMINLP
```
jsiirola: We have had issues with relative imports; it would probably be best to convert these to absolute imports...

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
```diff
@@ -0,0 +1,529 @@
...
+        if GurobiDirectBase._num_gurobipy_env_clients:
+            logger.warning(
+                "Call to GurobiDirectBase.release_license() with %s remaining "
+                "environment clients." % (GurobiDirectBase._num_gurobipy_env_clients,)
+            )
+        GurobiDirectBase._gurobipy_env.close()
+        GurobiDirectBase._gurobipy_env = None
+
+    @staticmethod
+    def env():
```
jsiirola, line 267: We should provide a mechanism for users to construct their own Env and pass it to us.  This is necessary for thing like working with a trial license.
```suggestion
    def env(env=None):
```

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
```diff
@@ -0,0 +1,529 @@
...
+        finally:
+            model.dispose()
+
+    def version(self):
+        version = (
+            gurobipy.GRB.VERSION_MAJOR,
+            gurobipy.GRB.VERSION_MINOR,
+            gurobipy.GRB.VERSION_TECHNICAL,
+        )
+        return version
```
jsiirola, line 310: Should this return None if gurobipy is unavailable?

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
```diff
@@ -0,0 +1,529 @@
...
+    def solve(self, model, **kwds) -> Results:
+        start_timestamp = datetime.datetime.now(datetime.timezone.utc)
+        orig_config = self.config
+        orig_cwd = os.getcwd()
+        try:
+            config = self.config(value=kwds, preserve_implicit=True)
+
+            # hack to work around legacy solver wrapper __setattr__
+            # otherwise, this would just be self.config = config
+            object.__setattr__(self, 'config', config)
```
jsiirola: The whole point of ephemeral options is to *not* change the instance config.  I would strongly recommend passing the local `config` to `_create_solver_model` and `_postsolve`

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
```diff
@@ -0,0 +1,529 @@
...
+
+            if not self.available():
+                c = self.__class__
+                raise ApplicationError(
+                    f'Solver {c.__module__}.{c.__qualname__} is not available '
+                    f'({self.available()}).'
+                )
```
jsiirola: I don't think that we should call `available()` within `solve()`... and instead just let any underlying exceptions (e.g., DeferredImportError) percolate up.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
```diff
@@ -0,0 +1,529 @@
...
+                results.solution_loader.load_vars()
+            else:
+                raise NoFeasibleSolutionError()
+        self.config.timer.stop('load solution')
+
+        # self.config gets copied a the beginning of
+        # solve and restored at the end, so modifying
+        # results.solver_config will not actually
+        # modify self.config
+        results.solver_config = self.config
```
jsiirola: We should either:

  -  duplicate the `config`
  -  (better yet), just store the ephemeral `config` that we created for this `solve()` - provided that we do not assign it back to `self.config` [see the earlier comment]

pyomo/contrib/solver/solvers/gurobi/gurobi_direct.py
```diff
@@ -0,0 +1,134 @@
...
+    gurobipy,
+    GurobiDirectSolutionLoaderBase,
+)
+import logging
+
+
+logger = logging.getLogger(__name__)
+
+
+class GurobiDirectSolutionLoader(GurobiDirectSolutionLoaderBase):
```
jsiirola, line 36: Why is this derived class needed?  Can't it be merged into the base class (and the base class renamed `GurobiDirectSolutionLoader`)?

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
```diff
@@ -0,0 +1,529 @@
...
+    solver_model: gurobipy.Model
+    var_map: Mapping[VarData, gurobipy.Var]
+        Maps the pyomo variable to the gurobipy variable
+    vars_to_load: List[VarData]
+    """
+    if solver_model.Status != gurobipy.GRB.OPTIMAL:
+        raise NoReducedCostsError()
+    if solver_model.IsMIP:
+        # this will also return True for continuous, nonconvex models
+        raise NoDualsError()
```
jsiirola: Shouldn't this also be a `NoReducedCostsError`?

pyomo/contrib/solver/solvers/gurobi/gurobi_direct.py
```diff
@@ -0,0 +1,134 @@
...
+            self._solver_model = None
+
+
+class GurobiDirect(GurobiDirectBase):
+    _minimum_version = (9, 0, 0)
+
+    def __init__(self, **kwds):
+        super().__init__(**kwds)
+        self._gurobi_vars = None
+        self._pyomo_vars = None
```
jsiirola, line 71: Why are these being stored on the solver interface?  Direct is supposed to be non-persistent, and adding this seems to violate that principle.

This seems to only be needed for `_pyomo_gurobi_var_iter()`, which is in turn only needed by `_mipstart`.  I propose that both the attributes and `_pyomo_gurobi)var_iter()` be removed, and instead `_mipstart` be updated to accept the SolutionLoader as an argument.  As the SolutionLoader is responsible for remembering the mapping between Pyomo and Gurobi, that feels like the natural place to query to get the mapping within `_mipstart`.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
```diff
@@ -0,0 +1,529 @@
...
+            ConfigValue(
+                default=False,
+                domain=bool,
+                description="If True, the current values of the integer variables "
+                "will be passed to Gurobi.",
+            ),
+        )
+
+
+def _load_suboptimal_mip_solution(solver_model, var_map, vars_to_load, solution_number):
```
jsiirola: Should this (and other loaders) be updated so that instead of `var_map`, it accepts an iterable of `(pyomo_var, gurobi_var)` tuples?  That way interfaces that do not generate the mapping can still make use of this method?

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
```diff
@@ -0,0 +1,529 @@
...
+        solver_model.getAttr('NumIntVars') == 0
+        and solver_model.getAttr('NumBinVars') == 0
+    ):
+        raise ValueError('Cannot obtain suboptimal solutions for a continuous model')
+    original_solution_number = solver_model.getParamInfo('SolutionNumber')[2]
+    solver_model.setParam('SolutionNumber', solution_number)
+    gurobi_vars_to_load = [var_map[v] for v in vars_to_load]
+    vals = solver_model.getAttr("Xn", gurobi_vars_to_load)
+    res = ComponentMap(zip(vars_to_load, vals))
+    solver_model.setParam('SolutionNumber', original_solution_number)
```
jsiirola: These implementations all seem to be geared around "partial loads", and are actually quite inefficient for "full loads".  As I would expect that the vast majority of users / use cases will actually be doing a full load, I would suggest redesigning these methods so that full loads are efficient, and partial loads are efficient if the number of objects to load is small.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
```diff
@@ -0,0 +1,529 @@
...
+                grb.TIME_LIMIT: tc.maxTimeLimit,
+                grb.SOLUTION_LIMIT: tc.unknown,
+                grb.INTERRUPTED: tc.interrupted,
+                grb.NUMERIC: tc.unknown,
+                grb.SUBOPTIMAL: tc.unknown,
+                grb.USER_OBJ_LIMIT: tc.objectiveLimit,
+            }
+        return GurobiDirectBase._tc_map
+
+    def _postsolve(self, grb_model, solution_loader, has_obj):
```
jsiirola: I would suggest renaming this something more descriptive like `_populate_results`

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
```diff
@@ -0,0 +1,1329 @@
...
+logger = logging.getLogger(__name__)
+
+
+class GurobiPersistentSolutionLoader(GurobiDirectSolutionLoaderBase):
+    def __init__(self, solver_model, var_map, con_map) -> None:
+        super().__init__(solver_model, var_map, con_map)
+        self._valid = True
+
+    def invalidate(self):
+        self._valid = False
```
jsiirola, line 68: Doesn't the Direct interface also need to worry about invalidating the last loader before the next solve?  Does this really need to be a specialization for the persistent interface?

jsiirola (commented):
Some initial comments and questions.  I haven't gotten through the persistent interface yet.  I have lots of design questions there -- it might be helpful to discuss offline / at a dev call.

pyomo/contrib/solver/plugins.py
michaelbynum: I don't remember why I did this. Maybe I was just going for consistency? I can revert it.

pyomo/contrib/solver/solvers/gurobi/__init__.py
michaelbynum: Will do.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: I agree.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: That makes sense to me. I'll update it.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: What makes you say this implementation is inefficient?

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: Yes. Good catch.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: That might work for the direct interfaces, but this way simplifies things tremendously for the persistent interface. This is a temporary change that gets reverted below.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: Sounds good.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: Will do.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: The way this is currently implemented, this is the ephemeral version.

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
michaelbynum: The direct interface solution loader should still be valid after subsequent solves, I think???

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
```diff
@@ -0,0 +1,529 @@
...
+        for pyomo_var, gurobi_var in self._pyomo_gurobi_var_iter():
+            if pyomo_var.is_integer() and pyomo_var.value is not None:
+                gurobi_var.setAttr('Start', pyomo_var.value)
+
+    def solve(self, model, **kwds) -> Results:
+        start_timestamp = datetime.datetime.now(datetime.timezone.utc)
+        orig_config = self.config
+        orig_cwd = os.getcwd()
+        try:
+            config = self.config(value=kwds, preserve_implicit=True)
```
michaelbynum, line 330: @jsiirola, the duplication happens here.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
```diff
@@ -0,0 +1,529 @@
...
+
+            res = self._postsolve(
+                grb_model=gurobi_model, solution_loader=solution_loader, has_obj=has_obj
+            )
+        finally:
+            os.chdir(orig_cwd)
+
+            # hack to work around legacy solver wrapper __setattr__
+            # otherwise, this would just be self.config = orig_config
+            object.__setattr__(self, 'config', orig_config)
```
michaelbynum: @jsiirola, it gets reverted here.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: I take it back. I think this used to be true, but it is not anymore. I will change this like you suggest.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct.py
michaelbynum: In the case of the persistent solution loader, we do not want to dispose of the gurobipy model when the solution loader is garbage collected. The solver interface needs to hang on to it for future solves.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct.py
michaelbynum: I like that suggestion. Will do.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: I think I have to iterate over it twice...

pyomo/contrib/solver/solvers/gurobi/gurobi_direct.py
emma58: Is this going to work for callbacks too? They have the same issue, that they need this state (and actually more) to work... So I think there may be a bigger design question lurking here: For example, they need the gurobipy model to be somewhere they can get to: In persistent, that is currently also sitting on the solver interface.

emma58: (I was originally thinking that this kind of state could be added in a contextmanager kind of style, and removed before `solve` is over, regardless of what happens. But that certainly might not be the right answer.)

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
jsiirola: At least in the case of the direct interface, we already have the list of gurobi vars and pyomo vars in the order in which they were given to Gurobi, so if you are getting everything you could get away with:
```python
return ComponentMap(zip(pyomo_vars, solver_model.getAttr('Xn')))
```
and avoid iterating over things twice, and forming an extra list, and performing *N* COmponentMap lookups (plus unknown lookups within Gurobi).

michaelbynum: Let me do some profiling and see if this even matters.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct.py
michaelbynum: Excellent points, @emma58. I think this is worth further discussion.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
jsiirola: A good model to use in profiling is @emma58's 400M var LP.

pyomo/contrib/solver/solvers/gurobi/__init__.py
michaelbynum: Oh, I think this has to be a relative import. This is in the `__init__.py` file.

pyomo/contrib/solver/plugins.py
michaelbynum: Done.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: Done

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: Done.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: Done.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: Done.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: Fixed.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: Done.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: I refactored the solution loader. Let me know what you think.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: So `ComponentMap` was indeed causing problems. I refactored the solution loader so that the mapping only gets created/used if it is needed. I tested the new implementation on a linear assignment problem with about a million variables (so not huge, but also not small). Loading the solution was over twice as fast as on `main` for `GurobiDirect`, and the overall timing was about 8% faster (excluding solve time). Let me know what you think.

michaelbynum: I still need to address some things with constraints. Constraints/duals are a bit more tricky. However, I think I am done with the variables.

michaelbynum: Okay, I think the constraint map is not a performance problem, so maybe this implementation is okay.

pyomo/contrib/solver/solvers/gurobi/__init__.py
jsiirola: I don't think there is *any* requirement for relative imports in `__init__.py` ... we generally don't use them elsewhere.

michaelbynum: You are correct. I got myself confused. Should be fixed now.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct.py
jsiirola: I am struggling with the design here.  I think that the SolutionLoader is filling multiple roles:
- it is the definitive "mapping object" between the Pyomo and Gurobi models
- it knows something about it's own validity (for persistent)
- it has the logic for translating the values back and forth
- it knows when to delete the Gurobi model

I feel like the the persistent interface should just hold on to the SolutionLoader as the definitive "mapping object" between the Pyomo and Gurobi models?  That would resolve the issue around premature deletion of the Gurobi model.  But it would make invalidation more complicated (because each solution would have the same loader?).  

I wonder: would a better design be for a mapping object that currently does 99% of the current SolutionLoader?  It would be the "owner" of the Gurobi model reference, know the translation keys, and know an "id" for the current solution.  The SolutionLoader would then be a thin wrapper that stored the `(mapper, solution_id)`.  Invalidating a SolutionLoader would only require incrementing the solution id in the mapper object (much in the same way that we handle `stale` for variables).  The mapping object could also be used for things like callbacks, too (see @emma58's comment below).

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
```diff
@@ -0,0 +1,546 @@
...
+
+    def _get_con_map(self):
+        raise NotImplementedError('should be implemented by derived classes')
+
+    def __del__(self):
+        # Release the gurobi license if this is the last reference to
+        # the environment (either through a results object or solver
+        # interface)
+        GurobiDirectBase._release_env_client()
+
```
jsiirola, line 103: I am struggling with the amount of copy-paste in this interface.  Is there a reason not to implement a single method:
```python
def _load_vars(self, vars_to_load=None, solution_number=0):
    if vars_to_load is None:
        pvars, gvars = self._get_var_lists()
    else:
        pvars = vars_to_load
        gvars = list(map(self._get_var_map().__getitem__, vars_to_load))
    if solution_number:
        if (
            self._solver_model.getAttr('NumIntVars') == 0
            and self._solver_model.getAttr('NumBinVars') == 0
        ):
            raise ValueError(
                'Cannot obtain suboptimal solutions for a continuous model'
            )
        original_solution_number = self._solver_model.getParamInfo('SolutionNumber')[2]
        self._solver_model.setParam('SolutionNumber', solution_number)
        grbFcn = "Xn"
    else:
        grbFcn = "X"
    try:
        vals = self._solver_model.getAttr(grbFcn, gvars)
        for pv, val in zip(pvars, vals):
            pv.set_value(val, skip_validation=True)
    finally:
        if solution_number:
            self._solver_model.setParam('SolutionNumber', original_solution_number)
```

(Note that this moves away from `_var_pair_iter()` and instead uses:
```python
def _get_var_lists(self):
    return self._var_map.keys(), self._var_map.values()
```
or
```python
def _get_var_lists(self):
    return self._pyomo_vars, self._gurobi_vars
```
depending on the interface)

jsiirola (commented):
Finally getting back to this.  I think there are still a number of open questions from previous reviews, plus a new one:

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: This is a great suggestion. I'll have this done shortly.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: This is unrelated to this PR, so I created an issue to track it. I mostly just don't have time to write the tests right now.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: Done.

michaelbynum:
I created issues for the unresolved comments. One is minor (handling `Env`), but will require writing some new tests. The other has to do with the mapper objects. That will require some major changes that I think we should do as a separate PR.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
```diff
@@ -84,105 +66,171 @@ def __init__(
...
+    def _get_rc_all_vars(self):
+        pvars, gvars = self._get_var_lists()
+        vals = self._solver_model.getAttr("Rc", gvars)
+        return ComponentMap((i, val) for i, val in zip(pvars, vals))
+
+    def _get_rc_subset_vars(self, vars_to_load):
+        var_map = self._get_var_map()
+        gvars = [var_map[i] for i in vars_to_load]
+        vals = self._solver_model.getAttr("Rc", gvars)
+        return ComponentMap(zip(vars_to_load, vals))
```
jsiirola, line 164: Does it make sense to merge these into `get_reduced_costs` similar to what was done with `get_primals`?

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
```diff
@@ -0,0 +1,1340 @@
...
+        self._assert_solution_still_valid()
+        return super().get_duals(cons_to_load)
+
+    def get_reduced_costs(
+        self, vars_to_load: Sequence[VarData] | None = None
+    ) -> Mapping[VarData, float]:
+        self._assert_solution_still_valid()
+        return super().get_reduced_costs(vars_to_load)
+
+
```
jsiirola, line 98: Should all these classes be slotized?  I think there could end up being *a lot* of them, so the memory / performance improvements of slotization might be significant.

I am fine deferring that to subsequent PRs.

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
```diff
@@ -0,0 +1,1340 @@
...
+            raise RuntimeError('must call set_instance or solve before update')
+        timer.start('update')
+        if self._needs_updated:
+            self._update_gurobi_model()
+        self._change_detector.update(timer=timer, **config.auto_updates)
+        timer.stop('update')
+
+    def _get_expr_from_pyomo_repn(self, repn):
+        if repn.nonlinear_expr is not None:
+            raise IncompatibleModelError(
```
jsiirola: Is this an `IncompatibleModelError` or should it be a `MouseTrap` (special case of `NotImplementedError`)?  Support for NL models would be desirable in the future...

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
```diff
@@ -0,0 +1,1340 @@
...
+        self._invalidate_last_results()
+        gurobi_expr_list = []
+        for ndx, con in enumerate(cons):
+            lb, body, ub = con.to_bounded_expression(evaluate_bounds=False)
+            repn = generate_standard_repn(body, quadratic=True, compute_values=False)
+            gurobi_expr = self._get_expr_from_pyomo_repn(repn)
+            mutable_constant = None
+            if lb is None and ub is None:
+                raise ValueError(
+                    "Constraint does not have a lower " f"or an upper bound: {con} \n"
```
jsiirola: Black did something weird here...

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
```diff
@@ -0,0 +1,1340 @@
...
+
+    def _add_constraints(self, cons: List[ConstraintData]):
+        self._invalidate_last_results()
+        gurobi_expr_list = []
+        for ndx, con in enumerate(cons):
+            lb, body, ub = con.to_bounded_expression(evaluate_bounds=False)
+            repn = generate_standard_repn(body, quadratic=True, compute_values=False)
+            gurobi_expr = self._get_expr_from_pyomo_repn(repn)
+            mutable_constant = None
+            if lb is None and ub is None:
```
jsiirola: It is *slightly* more efficient to put this test farther down to prevent testing the lb twice:
```python
if lb is None:
    if ub is None:
        raise ValueError()
    rhs_expr = ...
```

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
```diff
@@ -0,0 +1,1340 @@
...
+                Start
+                VarHintVal
+                VarHintPri
+                BranchPriority
+                VBasis
+                PStart
+        val: any
+            See gurobi documentation for acceptable values.
+        """
+        if attr in {'LB', 'UB', 'VType', 'VarName'}:
```
jsiirola: Are the Gurobi attributes case insensitive (I think they are)?  Should you map `attr` to lowercase before testing?

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
```diff
@@ -0,0 +1,1340 @@
...
+                f'Var attr {attr} cannot be set with'
+                ' the set_var_attr method. Please use'
+                ' the update_var method.'
+            )
+        if attr == 'Obj':
+            raise ValueError(
+                'Var attr Obj cannot be set with'
+                ' the set_var_attr method. Please use'
+                ' the set_objective method.'
+            )
```
jsiirola: It would be more efficient to create the set once (at the class / globals scope) and just interrogate it.  You could also make it a dict, something like:
```python
_disallow_set_var_attr = {
    'LB' = 'update_var',
    ...
    'Obj' = 'set_objective',
}
...

def set_var_attr()
   if attr in _disallow_set_var_attr:
        raise ValueError(
            f'Var attr {attr} cannot be set with'
            ' the set_var_attr method. Please use'
            f' the {_disallow_set_var_attr[attr]} method.'
        )

File: pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py [hunk 25]
```diff
@@ -0,0 +1,1340 @@
...
+        self._pyomo_var_to_solver_var_map[id(var)].setAttr(attr, val)
```
jsiirola: Shouldn't this be `[var]` and not `[id(var)]`?

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
```diff
@@ -0,0 +1,1340 @@
...
+        ----------
+        var: pyomo.core.base.var.VarData
+            The pyomo var for which the corresponding gurobi var attribute
+            should be retrieved.
+        attr: str
+            The attribute to get. See gurobi documentation
+        """
+        if self._needs_updated:
+            self._update_gurobi_model()
+        return self._pyomo_var_to_solver_var_map[id(var)].getAttr(attr)
```
jsiirola: Shouldn't this be `[var]` and not `[id(var)]`?

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
```diff
@@ -0,0 +1,1340 @@
...
+
+    def cbGetNodeRel(self, variables):
+        """
+        Parameters
+        ----------
+        variables: Var or iterable of Var
+        """
+        if not isinstance(variables, Iterable):
+            variables = [variables]
+        gurobi_vars = [self._pyomo_var_to_solver_var_map[id(i)] for i in variables]
```
jsiirola: Shouldn't this be `[var]` and not `[id(var)]`?

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
```diff
@@ -0,0 +1,1340 @@
...
+
+    def cbGetSolution(self, variables):
+        """
+        Parameters
+        ----------
+        variables: iterable of vars
+        """
+        if not isinstance(variables, Iterable):
+            variables = [variables]
+        gurobi_vars = [self._pyomo_var_to_solver_var_map[id(i)] for i in variables]
```
jsiirola: Shouldn't this be `[var]` and not `[id(var)]`?

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
```diff
@@ -0,0 +1,1340 @@
...
+            )
+        else:
+            raise ValueError(
+                f'Constraint does not have a lower or an upper bound {con} \n'
+            )
+
+    def cbSetSolution(self, variables, solution):
+        if not isinstance(variables, Iterable):
+            variables = [variables]
+        gurobi_vars = [self._pyomo_var_to_solver_var_map[id(i)] for i in variables]
```
jsiirola: Shouldn't this be `[var]` and not `[id(var)]`?

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
```diff
@@ -0,0 +1,1340 @@
...
+
+
+class _MutableLowerBound:
+    def __init__(self, var_id, expr, var_map):
+        self.var_id = var_id
+        self.expr = expr
+        self.var_map = var_map
+
+    def update(self):
+        self.var_map[self.var_id].setAttr('lb', value(self.expr))
```
jsiirola: I think this is a problem (here and in the other `_Mutable*` classes): the `var_map` is a ComponentMap, so you should be storing and looking things up by the `var` and not the `var_id`

jsiirola (commented):
I was going to approve and leave all these comments to a future PR, but I think there is a significant bug that should be resolved before merging: there is inconsistent use `var` vs `id(var)` in `gurobi_persistent`: it looks liek the `_pyomo_var_to_solver_var_map` has moved to a `ComponentMap`, but there are still numerous cases where we are storing / looking things up by `id()`.

michaelbynum:
> I was going to approve and leave all these comments to a future PR, but I think there is a significant bug that should be resolved before merging: there is inconsistent use `var` vs `id(var)` in `gurobi_persistent`: it looks liek the `_pyomo_var_to_solver_var_map` has moved to a `ComponentMap`, but there are still numerous cases where we are storing / looking things up by `id()`.

It's not actually a bug. With `ComponentMap`, you can key by variable or by id of the variable, and you get the same thing:

```

In [2]: from pyomo.common.collections import ComponentMap

In [3]: m = pe.ConcreteModel()

In [4]: m.x = pe.Var()

In [5]: d = ComponentMap()

In [6]: d[m.x] = 'foo'

In [7]: d[id(m.x)]
Out[7]: 'foo'
```

That being said, it does need fixed, so I'll fix it now. Tests would have caught most of these if it had been a real bug.

jsiirola:
> It's not actually a bug. With `ComponentMap`, you can key by variable or by id of the variable, and you get the same thing:

Actually, it is a bug - with `ComponentMap`.  That is actually a hash collision between unhashable types (like Var) and ints.  Note also that it only happens for unhashable types.  Hashable types (like Constraint / Model) stored as `ComponentMap` keys cannot be looked up by `id()`.  

I had found/fixed it back when we were working on performance in `ComponentMap` / `ComponentSet`, but apparently that is still living in a stash somewhere.  I will try and dig it up and put up a PR.

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
michaelbynum: fixed

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
michaelbynum: fixed

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
michaelbynum: fixed

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
michaelbynum: fixed

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
michaelbynum: fixed

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
michaelbynum: fixed

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: I don't think so. There is not really any duplicate code for the reduced costs.

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
michaelbynum: Great suggestion. Done.

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
michaelbynum: Good point. Done.

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
michaelbynum: It sure did. Fixed.

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
michaelbynum: I think it is easier to read this way.

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
michaelbynum: Great suggestion. Done.

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
michaelbynum: Ah, maybe not. I'll give it a go and see.

michaelbynum: Oh, yeah. I'm crazy. Forget I said anything. I was thinking of the wrong thing.

michaelbynum: Fixed.

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
michaelbynum: Done.

michaelbynum:
Okay, this should be good to go, assuming tests pass.

jsiirola (approved):

PR #3701, opened 2025-08-13
Rework SolutionLoader

michaelbynum:
## Summary/Motivation:
This PR updates the solution loader in pyomo/contrib/solver based on recent design discussions.

- add `get_number_of_solutions` method
- add `get_solution_ids` method
- add `load_solution` method
- add `load_import_suffixes` method
- add `solution()` to provide (through a view or a context manager) access to solutions from the solution pool

### Legal Acknowledgement


pyomo/contrib/solver/common/solution_loader.py
```diff
@@ -9,11 +9,34 @@
...
+    for suffix in pyomo_model.component_objects(Suffix, descend_into=True, active=True):
+        if not suffix.import_enabled():
+            continue
+        if suffix.local_name == 'dual':
+            dual_suffix = suffix
+        elif suffix.local_name == 'rc':
+            rc_suffix = suffix
+    if dual_suffix is not None:
+        for k, v in solution_loader.get_duals(solution_id=solution_id).items():
+            dual_suffix[k] = v
```
jsiirola: Some questions:
- Should the Suffix be cleared before loading?
- Could this just be:

```suggestion
        dual_suffix.update(solution_loader.get_duals(solution_id=solution_id))
```

pyomo/contrib/solver/common/solution_loader.py
```diff
@@ -50,18 +119,21 @@ def get_primals(
         vars_to_load: list
             A list of the variables whose solution value should be retrieved. If vars_to_load
             is None, then the values for all variables will be retrieved.
+        solution_id: Optional[Any]
```
jsiirola: Instead of `Optional[Type]`, I believe the current "preferred Python" annotation is `Type | None`.  In this case, since `Any` includes `None`, I think you could just do `Any` (although `Any | None` is a little more explicit).

pyomo/contrib/solver/solvers/asl_sol_reader.py
```diff
@@ -55,16 +58,34 @@ class ASLSolFileSolutionLoader(SolutionLoaderBase):
...
+    def get_number_of_solutions(self) -> int:
+        if self._nl_info is None:
+            return 0
+        return 1
+
+    def get_solution_ids(self) -> List[Any]:
+        return [None]
+
+    def load_import_suffixes(self, solution_id=None):
+        load_import_suffixes(self._pyomo_model, self, solution_id=solution_id)
```
jsiirola: This only loads `duals` and `rc` ... the ASL supports arbitrary inpirt suffixes, which I think need to be supported here.

jsiirola (changes requested):
A couple minor questions that don't need to hold up the PR and one significant one that probably does (it look like the ASL solution loader is not loading all import suffixes).

pyomo/contrib/solver/solvers/asl_sol_reader.py
michaelbynum: Can you give me an example of what `ASLSolFileData` would look like if it had other suffixes. I see that it has attributes like `var_suffixes`. I assume the keys are the names of the suffixes, and the values map the variable indices to the suffix values? Any suggestions on how to test this?

jsiirola: Yes: Suffix data comes back from the ASL flagged for "variables", "constraints", "objectives", and the "problem".  We blindly parse that into the corresponding attributes (`var_suffixes`, etc).  
- For `var`, `con`, and `obj`, the dict maps suffix name to a dict that maps the 0-based integer index of the relevant variable / constraint / objective in the corresponding list in the `NLWriterInfo` returned by the writer tot he suffix value.  
- For the problem, since there is only one problem, the `problem_suffixes` dict just maps suffix name to value.

pyomo/contrib/solver/common/solution_loader.py
michaelbynum: Done.

pyomo/contrib/solver/common/solution_loader.py
michaelbynum: Done.

pyomo/contrib/solver/solvers/asl_sol_reader.py
michaelbynum: Done - kind of. I just raise an error if any variables are eliminated or any scaling is done.

michaelbynum:
@AnhTran01 , @boxblox  - could you please take a look at the changes to the gams solution loader?

boxblox:
> @AnhTran01 , @boxblox - could you please take a look at the changes to the gams solution loader?

yep, is there anything in particular you need me to focus in on?  on my first pass it looks like a lot of internal organization.  Not sure I'd have much to add on that, but can look for an eye for clarity, if that's helpful.

michaelbynum:
> > @AnhTran01 , @boxblox - could you please take a look at the changes to the gams solution loader?
> 
> yep, is there anything in particular you need me to focus in on? on my first pass it looks like a lot of internal organization. Not sure I'd have much to add on that, but can look for an eye for clarity, if that's helpful.

I mostly wanted to make sure I understood when a solution is available for the purposes of `get_number_of_solutions`. If `_gms_info` is not `None`, then it should be safe to assume that there is a solution, correct? That's what it looked like from `load_vars`.

michaelbynum:
@jsiirola - this should be ready for a follow-up review. One thing that is probably worth further discussion: I added a method to the `SolutionLoader` to get the number of solutions and the solution ids, but this does not include a way to indicate the status (optimal, feasible, etc.) of each of those solutions. Thoughts?

michaelbynum:
The Jenkins failures seem to be unrelated to this PR. They look like timeout error during calls to `available`. We probably want to increase the default timeout for those. I can create a separate PR for that.

pyomo/contrib/solver/common/solution_loader.py
```diff
@@ -7,11 +7,34 @@
 # software.  This software is distributed under the 3-clause BSD License.
 # ____________________________________________________________________________________
 
-from typing import Sequence, Dict, Optional, Mapping
+from __future__ import annotations
+
+from typing import Sequence, Dict, Optional, Mapping, List, Any
```
jsiirola: Do you know what you need from `__future__` here?

Also, as we are Python>-3.10, you don't need `List` or `Dict` (and can just use `list` and `dict`)

pyomo/contrib/solver/solvers/knitro/solution.py
```diff
@@ -8,7 +8,7 @@
 # ____________________________________________________________________________________
 
 from collections.abc import Mapping, Sequence
-from typing import Protocol
+from typing import Any, List, Protocol
```
jsiirola: Again, you don't need to import `List` and can just use `list`

pyomo/contrib/solver/common/solution_loader.py
```diff
@@ -7,11 +7,34 @@
...
 
 from pyomo.core.base.constraint import ConstraintData
 from pyomo.core.base.var import VarData
 from pyomo.core.staleflag import StaleFlagManager
+from pyomo.core.base.suffix import Suffix
+
+
+def load_import_suffixes(
```
jsiirola: There's a couple problems with this implementation:
- this doesn't clear any import suffixes other than `dual` and `rc`
- if there are multiple dual suffixes, only the last one found is cleared (and that is usually the deeper one in the tree)

See the next comment for a proposed solution

pyomo/contrib/solver/common/solution_loader.py
```diff
@@ -88,45 +163,70 @@ def get_reduced_costs(
...
+
+    def load_import_suffixes(self, solution_id=None):
+        """
+        Parameters
+        ----------
+        solution_id: Optional[Any]
+            If there are multiple solutions, this specifies which solution
+            should be loaded. If None, the default solution will be used.
+        """
+        return NotImplemented
```
jsiirola: I propose the following default implementation:
```suggestion
    def load_import_suffixes(self, solution_id=None):
        """Clear import suffixes on the model and load data returned by the solver.
        
        Parameters
        ----------
        solution_id: Optional[Any]
            If there are multiple solutions, this specifies which solution
            should be loaded. If None, the default solution will be used.
        """
        suffixes = self._clear_import_suffixes()
        if 'dual' in known_suffixes:
            suffixes['dual'].update(solution_loader.get_duals(solution_id=solution_id))
        if 'rc' in known_suffixes:
            suffixes['rc'].update(
                solution_loader.get_reduced_costs(solution_id=solution_id)
            )

    def _clear_import_suffixes(self):
        """Clear and return all import suffixes on the model

        This walks the Pyomo model and clears all :class:`Suffix`
        components that are flagged to import values from the solver
        (this includes :attr:`Suffix.IMPORT` and
        :attr:`Suffix.IMPORT_EXPORT`).  It returns a :class:`dict`
        mapping the :attr:`Suffix.local_name` to the :class:`Suffix`
        closest to the root block.

        Returns
        -------
        known_suffixes : dict[str, Suffix]
        """
        known_suffixes = {}
        for suffix in self.pyomo_model.component_objects(
            Suffix,
            active=True,
            descend_into=True,
            descent_order=TraversalStrategy.BreadthFirstSearch,
        ):
            if not suffix.import_enabled():
                continue
            suffix.clear()
            known_suffixes.setdefault(suffix.local_name, suffix)
        return known_suffixes
```

This resolves issues with not clearing all import suffixes and handling multiple Suffix components with the same name.  It also makes this a proper method of the loader (the "public utility function" seemed a bit strange).

This does imply a more significant change: it requires all loaders to keep an explicit reference to the Pyomo model.  It is an open question if `pyomo_model` should just be a public attribute or a property...  Either way, I think that is actually a positive change, but it is a bigger one that we should at least acknowledge/discuss.

jsiirola: An alternate implementation could be to force the solver to pre-collect the import Suffixes and just record that list on the SolutionLoader, but that seems more clunky and repetitive (each solver would have to remember to implement it)?

pyomo/contrib/solver/common/solution_loader.py
```diff
@@ -21,24 +44,70 @@ class SolutionLoaderBase:
...
+    def get_number_of_solutions(self) -> int:
+        """
+        Returns
+        -------
+        num_solutions: int
+            Indicates the number of solutions found
+        """
+        return NotImplemented
+
+    def load_solution(self, solution_id=None):
```
jsiirola: I have a general question about this design: do we know the "cost" of switching the "active solution"?  As implemented, `load_solution` will set and clear the active solution up to 3 times (for vars, duals, and reduced costs).  Do we know if that is a problem.

An alternate solution could be to implement a context manager, so users could do things like:
```python
results = solver.solve(m)
# ...
solution_ids = results.get_solution_ids()
with results.solution(solution_ids[2]):
    results.load_solution()
    # or
    vars = results.get_vars()
    duals = results.get_duals()
```

I am not proposing a change (yet) ... just interested in your thoughts.

pyomo/contrib/solver/solvers/asl_sol_reader.py
```diff
@@ -53,16 +57,83 @@ class ASLSolFileSolutionLoader(SolutionLoaderBase):
...
-    def load_vars(self, vars_to_load: Optional[Sequence[VarData]] = None) -> None:
+        self._pyomo_model = pyomo_model
+
+    def get_number_of_solutions(self) -> int:
+        if self._nl_info is None:
+            return 0
+        return 1
+
+    def get_solution_ids(self) -> List[Any]:
+        return [None]
```
jsiirola: Should this return `[]` if `self._nl_info is None`?

pyomo/contrib/solver/solvers/asl_sol_reader.py
```diff
@@ -53,16 +57,83 @@ class ASLSolFileSolutionLoader(SolutionLoaderBase):
...
+                        'all suffixes.'
+                    )
+                if self._nl_info.scaling:
+                    raise MouseTrap(
+                        'General suffixes (other than duals and reduced costs) '
+                        'are not available when the model has been scaled. Turn '
+                        'scaling off in the NL writer '
+                        '(solver.config.writer_config.scale_model=False) to get '
+                        'all suffixes.'
+                    )
```
jsiirola: I have been thinking about this behavior: I think we should *not* raise an exception here and instead only log a warning (and possibly some day include a configuration option for suppressing the warning).  My thought process is something like "well, this is not an unrecoverable error ... we *can* load the suffixes that the solver returns.  It is just the user's responsibility to think through how to best interpret them in the context of scaling or presolve."  

The concrete example is something like getting and setting the bound multipliers for warmstarting Ipopt.  If we know we are not changing scaling or the model structure between solves, then the bound multipliers are still valid for the next warmstart.  If we raise an exception, not only does this prevent the user from warmstarting Ipopt, it prevents loading solutions from the previous run (if they had declared the Suffix).

pyomo/contrib/solver/solvers/highs.py
```diff
@@ -760,19 +774,25 @@ def _postsolve(self, stream: io.StringIO):
...
                 raise NoFeasibleSolutionError()
         timer.stop('load solution')
 
         return results
 
-    def _load_vars(self, vars_to_load=None):
+    def _load_vars(self, vars_to_load=None, solution_id=None):
+        assert (
+            solution_id is None
+        ), 'highs interface does not currently support multiple solutions'
```
jsiirola: This is validating a user input and not a programming assumption.  It should not be bypassed when running Python with optimization enabled, so therefore should be an "if ... raise ValueError()".

(I just caught this, but it applies to GAMS and ASL loaders, too)

pyomo/contrib/solver/common/solution_loader.py
jsiirola: One advantage of the context manager approach is we would only need to validate the `solution_id` in one place (instead of repeating it in *every* load / get method).

pyomo/contrib/solver/solvers/asl_sol_reader.py
```diff
@@ -53,16 +57,83 @@ class ASLSolFileSolutionLoader(SolutionLoaderBase):
...
+                suffix = suffixes_to_load[suffix_name]
+                suffix.clear()
+                for comp_ndx, val in suffix_vals.items():
+                    comp = comp_list[comp_ndx]
+                    suffix[comp] = val
```
jsiirola: This is a general note on efficiency... this is measurably slower than just 
```python
suffix[comp_list[comp_ndx]] = val
```
(Python has to create the "comp" entry in `locals()` and then immediately has to look it up).

In general this isn't an issue, but in tight loops (like the solution loader), silly things like creating an unnecessary variable are actually measurable.  I'm not sure this needs to be changed for this PR, but something to keep in mind as we get closer to finalizing this interface.

pyomo/contrib/solver/solvers/asl_sol_reader.py
michaelbynum: Oh, good catch. I need to add a test for this, too.

pyomo/contrib/solver/solvers/asl_sol_reader.py
michaelbynum: Done.

pyomo/contrib/solver/solvers/highs.py
michaelbynum: Done.

pyomo/contrib/solver/solvers/asl_sol_reader.py
michaelbynum: done

pyomo/contrib/solver/solvers/asl_sol_reader.py
michaelbynum: I did modify this. There are some lingering issues, but I would prefer to address them after #3703 is merged. I added an issue to track this (#3917). Right now, I don't think it is possible to hit the case where `_nl_info` is `None`. What happens is that `results.solution_loader` ends up being `None` instead. I personally would find that unexpected...

pyomo/contrib/solver/common/solution_loader.py
michaelbynum: Currently, the only interfaces that actually support solution pools are the gurobi interfaces. With those, I don't think there is really any cost. That being said, I still like your suggestion - particularly the bit about not checking the solution id in *every* method...

jsiirola (approved):

jsiirola:
@michaelbynum, @mrmundt: I implemented the things that @michaelbynum and I discussed. Once either/both of you look it over, then I think we can merge.

pyomo/contrib/solver/common/solution_loader.py
```diff
@@ -7,123 +7,364 @@
...
+        closest to the root block.
+
+        Returns
+        -------
+        import_suffixes : dict[str, Suffix]
 
-class PersistentSolutionLoader(SolutionLoaderBase):
+        """
+        import_suffixes = {}
+        for suffix in self._pyomo_model.component_objects(
```
michaelbynum, line 241: We should probably make `_pyomo_model` an attribute on the `SolutionLoader` class. Right now, this just assumes the derived classes create the attribute.

pyomo/contrib/solver/solvers/knitro/solution.py
```diff
@@ -50,13 +50,14 @@ def __init__(
         self.has_reduced_costs = has_reduced_costs
         self.has_duals = has_duals
 
+    def get_solution_ids(self) -> list[Any]:
+        if self.get_number_of_solutions() == 0:
+            return []
+        return [None]
```
michaelbynum: I think we can delete this method? The base class has the same implementation.

pyomo/contrib/solver/solvers/gms_sol_reader.py
```diff
@@ -8,47 +8,43 @@
...
-
-    def __init__(self) -> None:
-        self.primals: List[float] = []
-        self.duals: List[float] = []
-        self.var_suffixes: Dict[str, Dict[int, Any]] = {}
-        self.con_suffixes: Dict[str, Dict[Any]] = {}
-        self.obj_suffixes: Dict[str, Dict[int, Any]] = {}
-        self.problem_suffixes: Dict[str, List[Any]] = {}
-        self.other: List[str] = []
-
```
michaelbynum, line 40: I don't remember deleting this. Did you change this, @jsiirola?

michaelbynum (commented):
I think this looks great. Thanks for all the effort you put into this, @jsiirola. I love the new solution loader view and context manager. That was a great idea. I'm also glad you updated the ASL and Ipopt solution loaders to generalize the reduced costs.

pyomo/contrib/solver/solvers/knitro/solution.py
michaelbynum: I'm going to go ahead and delete this.

michaelbynum: Done.

michaelbynum:
Okay, I'm happy to hit merge if tests pass.

pyomo/contrib/solver/solvers/gms_sol_reader.py
michaelbynum: I see - this class was not being used. Thanks.

jsiirola: Right - I was going through and updating type hints to Python 3.10+, which caused those lines to be added to the PR's "patch" - and they were uncovered.  I dug around, and that class is never used (I think it was left over from when the LP writer was used as a base for the rewrite of the GAMS writer).  Removing it seemed right.

PR #3703, opened 2025-08-14
Add tests for trivial constraints in pyomo/contrib/solver

michaelbynum:
## Summary/Motivation:
This PR adds a test for trivial constraints (both feasible and infeasible) to the new solver tests. It also fixes some related bugs.

### Legal Acknowledgement


pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
```diff
@@ -0,0 +1,1428 @@
...
+    @property
+    def var1(self):
+        return self.var_map[self.v1id]
+
+    @property
+    def var2(self):
+        return self.var_map[self.v2id]
+
+
+class GurobiDirectQuadratic(GurobiDirectBase):
```
yemine0x01: Shouldn't this class be named GurobiConsistentQuadratic? In GurobiObserver this is the used name.

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
```diff
@@ -0,0 +1,1428 @@
...
+                raise ValueError(f'Objective sense is not recognized: {obj.sense}')
+
+            repn = generate_standard_repn(obj.expr, quadratic=True, compute_values=True)
+            gurobi_expr = self._get_expr_from_pyomo_repn(repn)
+            repn_constant = repn.constant
+
+        self._solver_model.setObjective(gurobi_expr + repn_constant, sense=sense)
+
+
+class _GurobiObserver(Observer):
```
yemine0x01: Since this class is just calling the "same" functions on the opt field. Why not just inherits Observer in GurobiPersistantSolver?

michaelbynum: It has to be done this way in order for "manual mode" of the persistent solvers to work. If someone calls `opt.add_constraint` directly on the solver interface, we need the observer to be updated so things stay synchronized.

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
michaelbynum: I'm confused. Can you point me to to the place in GurobiObserver you are referring to?

yemine0x01: I was referring to the field opt of the class _GurobiObserver.

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
yemine0x01: You’re probably right, and I might be missing something. However, I don’t see why adding Observer as a superclass of GurobiPersistent wouldn't mean the observer is always updated, since it’s essentially the same as the persistent object.

michaelbynum: The important thing is that the `_change_detector` methods get called. This makes my head hurt a little. Let me think on it.

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
michaelbynum: This class does not use the observer. This one is not persistent. Maybe I am missing something?

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
michaelbynum: So if we did something like

```
class GurobiPersistent(GurobiDirectQuadratic, Observer):
    def add_constraints(self, cons):
        self._change_detector.add_constraints(cons)
```

Then there would not be a way to call `_add_constraints`. It would just be an infinte loop. If `GurobiPersistent.add_constraints` gets called, it would call the `_change_detector`, and the `_change_detector` would again call `add_constraints` on the observer, which is `GurobiPersistent.add_constraints`. 

That is convoluted, but did it make any sense?

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
yemine0x01: Sorry, maybe I wasn’t clear enough. From what I see, in _GurobiObserver you are requesting opt as a GurobiPersistantQuadratic, but this class doesn’t appear to be defined. However, GurobiPersistant inherits from GurobiDirectQuadratic and is what gets passed to _GurobiObserver. So either I’m missing something, or there’s a naming issue.

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
yemine0x01: With the current design, this will always happen: the solver calls the detector, which calls the observer, which in turn calls the solver (Opt) again. Or am I missing something?

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
michaelbynum: Good catch! Thank you. The observer is expecting `GurobiPersistent` not `GurobiPersistentQuadratic`. I will fix that.

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
michaelbynum: If `GurobiPersistent.add_constraints` gets called, then that will call `ModelChangeDetector.add_constraints`, which will then call `_GurobiObserver.add_constraints`, which then calls `GurobiPersistent._add_constraints` and *not* `GurobiPersistent.add_constraints`. Very subtle but important difference there. However, this conversation is making me realize how convoluted this is. Maybe someone has a better idea?

yemine0x01: Okay, understood! One possible fix would be to give all `Observer` functions a common prefix (e.g., `on_*`) and make GurobiPersistantSolver a subclass of `Observer`. We could then reimplement the functions accordingly.

pyomo/contrib/solver/common/solution_loader.py
```diff
@@ -90,45 +166,107 @@ def get_reduced_costs(
...
+        return []
+
+    def get_number_of_solutions(self) -> int:
+        return 0
+
+    def load_solution(self, solution_id=None):
+        raise NoSolutionError()
+
+    def load_vars(
+        self, vars_to_load: Sequence[VarData] | None = None, solution_id=None
```
yemine0x01: Shouldn't we use Optional here?

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
```diff
@@ -375,6 +376,8 @@ def solve(self, model, **kwds) -> Results:
                 has_obj=has_obj,
                 config=config,
             )
+        except InfeasibleConstraintException:
+            res = self._get_infeasible_results(config=config)
```
jsiirola: This exception should have information about the infeasible constraint in its message.  That should be preserved and added to the NoSolutionError / NoOptimalSolutionError / NoFeasibleSolutionError exceptions

pyomo/contrib/solver/common/solution_loader.py
```diff
@@ -331,6 +332,41 @@ def load_import_suffixes(self):
             return self._loader.load_import_suffixes()
 
 
+class NoSolutionSolutionLoader(SolutionLoader):
+    def __init__(self) -> None:
+        pass
+
+    def get_solution_ids(self) -> List[Any]:
```
jsiirola: This is redundant (covered by the base class)

pyomo/contrib/solver/common/solution_loader.py
```diff
@@ -331,6 +332,41 @@ def load_import_suffixes(self):
...
+        return []
+
+    def get_number_of_solutions(self) -> int:
+        return 0
+
+    def load_solution(self):
```
jsiirola: This is redundant (the base class will call `get_vars()`, which will raise the exception)

pyomo/contrib/solver/common/solution_loader.py
```diff
@@ -331,6 +332,41 @@ def load_import_suffixes(self):
...
+        raise NoSolutionError()
+
+    def load_vars(self, vars_to_load: Sequence[VarData] | None = None) -> None:
```
jsiirola: This is redundant (the base class will call `get_vars()`, which will raise the exception)

pyomo/contrib/solver/common/solution_loader.py
```diff
@@ -331,6 +332,41 @@ def load_import_suffixes(self):
...
+        self, cons_to_load: Sequence[ConstraintData] | None = None
+    ) -> Dict[ConstraintData, float]:
+        raise NoSolutionError()
+
+    def get_reduced_costs(
+        self, vars_to_load: Sequence[VarData] | None = None
+    ) -> Mapping[VarData, float]:
+        raise NoSolutionError()
+
+    def load_import_suffixes(self):
```
jsiirola: Is this is redundant? If the model has `dual` / `rc` Suffixes, then the base class implementation will call `get_duals()` / `get_reduced_costs()`, which will raise the exception. If those suffixes aren't defined, then nothing will happen ... but that might be consistent with the behavior implied by the base class implementation?

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
```diff
@@ -16,6 +16,7 @@
 from pyomo.common.errors import PyomoException
 from pyomo.common.shutdown import python_is_shutting_down
 from pyomo.common.timing import HierarchicalTimer
+from pyomo.common.errors import InfeasibleConstraintException
```
jsiirola: Why was this added?  It appears to be unused.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
```diff
@@ -407,6 +410,24 @@ def _get_tc_map(self):
             }
         return GurobiDirectBase._tc_map
 
+    def _get_infeasible_results(self, config):
```
jsiirola: Should this be promoted to a standard method (in results / util) that is specialized here?

jsiirola (changes requested):
Some minor nits (we can debate those), and one significant one (we should really propagate information about the infeasible constraint out through the `NoSolutionError`).

pyomo/contrib/solver/solvers/gurobi/gurobi_persistent.py
michaelbynum: Accident. Thank you. I removed it.

pyomo/contrib/solver/common/solution_loader.py
michaelbynum: Removed.

pyomo/contrib/solver/common/solution_loader.py
michaelbynum: Removed.

pyomo/contrib/solver/common/solution_loader.py
michaelbynum: Removed.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: I implemented this. Let me know what you think.

michaelbynum:
- I still have one comment left to address.
- There were several failing tests that I had to fix. I think there is still one more. These were real failures. There was a spot in the highs interface where I was using the wrong attribute to determine if there was a feasible solution or not. It works in most cases, but not all. I think I have that fixed. There was also an edge case in the gurobi persistent interface where a trivially infeasible constraint did not get caught if it started as a trivially feasible constraint and changed because of a mutable parameter or fixed variable value.

michaelbynum:
Tests are not running. I'm going to mark this as ready and see if that helps.

pyomo/contrib/solver/solvers/gurobi/gurobi_direct_base.py
michaelbynum: Done

pyomo/repn/plugins/gams_writer_v2.py
```diff
@@ -386,9 +387,14 @@ def write(self, model):
...
+                if (lb is not None and lb > offset) or (ub is not None and ub < offset):
+                    raise InfeasibleConstraintException(
+                        f'detected a trivially infeasible constraint: {con}'
+                    )
                 if (
-                    skip_trivial_constraints
-                    and (lb is None or lb <= offset)
+                    # skip_trivial_constraints
+                    # and (lb is None or lb <= offset)
+                    (lb is None or lb <= offset)
```
michaelbynum: Not skipping trivially feasible constraints does not seem to work. When I create a trivially feasible constraint and do not skip them, I get exceptions (https://github.com/Pyomo/pyomo/actions/runs/25291155719/job/74143137773). I checked codecov, and this bit of code is not covered on the `main` branch. I think we should just always skip trivially feasible constraints in the gams writer. Thoughts? If everyone agrees, I can deprecate the option.

michaelbynum:
The 13 uncovered lines in gams.py are not related to this PR. I just indented that code inside of a try/except block.

pyomo/repn/plugins/gams_writer_v2.py
jsiirola: The error from GAMS is because we are outputting an invalid expression for trivial constraints:
```
c3_lo..  =G= -1;
```
This is missing *any* output for the body; i.e., it should be 
```
c3_lo.. 0 =G= -1;
```
This is arguably a bug in the GAMS writer.  That said, it was never quite clear what the motivation for `slip_trivial_constraints=False` was.  Should we deprecate it for *all* interfaces?

(Note, even if we deprecate it, we probably still need to fix the GAMS writer)

jsiirola (approved):
I think I have addressed the outstanding comments

pyomo/repn/plugins/gams_writer_v2.py
jsiirola: fixed.

pyomo/contrib/solver/tests/unit/test_solution.py
```diff
@@ -391,3 +393,38 @@ def test_load_vars(self):
...
+            loader.get_reduced_costs()
+        with self.assertRaisesRegex(NoSolutionError, "^error message$"):
+            loader.load_solution()
+        with self.assertRaisesRegex(NoSolutionError, "^error message$"):
+            loader.load_vars()
+        # If there are no suffixes declared on the model, then there
+        # should be no error (because there is nothing to import)
+        self.assertEqual(loader.load_import_suffixes(), None)
+        # non-standard suffixes are ignored
+        # TODO: is this "good" behavior??
```
michaelbynum, line 419: Great question.

michaelbynum (commented):
Thanks for your updates, @jsiirola. This looks good to me.

PR #3716, opened 2025-09-03
Set LogToConsole instead of OutputFlag to handle "tee"

### LINKED ISSUE #3589: GurobiDirect does not save logfiles when requested
## Summary
The `logfile` keyword argument to `solve` does not work correctly for `GurobiDirect`. 

```python
from pyomo.environ import *
m = ConcreteModel()
m.x = Var()
m.c = Constraint(expr=(0.0, m.x, 1.0))
m.o = Objective(expr=m.x)

s = SolverFactory("gurobi_direct")
s.solve(m, logfile="gurobilog.out")
```

`gurobilog.out` is nowhere to be found. This seems to work correctly for other solvers, e.g., xpress, ipopt, cbc,..

Pyomo version: 32591c154dd9268a62552969afc3f0ea2079db42
Python version: 3.12.9
Operating system: macOS
Solver (if applicable): Gurobi

rschwarz:
## Summary/Motivation:

In Gurobi, there are three parameters related to logging:

[LogToConsole](https://docs.gurobi.com/projects/optimizer/en/current/reference/parameters.html#logtoconsole) is a boolean flag that toggles whether the solver output is logged to the console. (This fits the intention behind setting `tee` in Pyomo.)

LogFile is a string that refers to a file name. If non-empty, solver output will be appended to that file.

Finally, OutputFlag is a combination of the two: If set to 0, both LogToConsole is set to 0, as well as LogFile set to an empty string.

- Use LogToConsole instead of OutputFlag
- In the old implementation, if a user has already specified a LogFile via the `options`, but then calls `solve(tee=False)`, the log files will not be written. This change makes sure that `tee` only affects what happens to the console.

## Other notes

By the way: Only GurobiDirect and the derived GurobiPersistent have this logic implemented that way. For the `appsi_gurobi` implementation, `tee` already had no effect on what happened with log files.

### Legal Acknowledgement


rschwarz:
There is still one annoying aspect, even after this change, which is that the line
`Set parameter LogToConsole to value 0`
is printed once per call to `solve`, so we see it a lot when solving models in a loop.

I think this can be improved by only calling `_solver_model.setParam` when it would actually change the value, which I will add.

The Gurobi-related tests pass locally for me, but I think there are some failing checks on the `main` branch already, right?

pyomo/solvers/plugins/solvers/gurobi_direct.py
```diff
@@ -240,9 +240,9 @@ def _apply_solver(self):
         StaleFlagManager.mark_all_as_stale()
 
         if self._tee:
-            self._solver_model.setParam('OutputFlag', 1)
+            self._solver_model.setParam('LogToConsole', 1)
         else:
-            self._solver_model.setParam('OutputFlag', 0)
```
quantresearch1, line 245: Why should this be the new default ? It seems that what you want can be achieved by adding LogToConsole=1 to the params you pass in the inputs

rschwarz: No, I actually want (in this case) `LogToConsole=0`, which is implied by `OutputFlag=0`, but the latter also disables writing to log files.

But I want to be able to have all 4 combinations of console and log files for different use cases.

My understand was that the `tee` option refers to what is shown on the console, so it should not interfere with log files, which `OutputFlag` does.

rschwarz:
> There is still one annoying aspect, even after this change, which is that the line `Set parameter LogToConsole to value 0` is printed once per call to `solve`, so we see it a lot when solving models in a loop.
> 
> I think this can be improved by only calling `_solver_model.setParam` when it would actually change the value, which I will add.

Conditionally calling `setParam` does not actually avoid printing that line, since the default value of `LogToConsole` is actually 1.

To fix that problem (for me), I had to set `LogToConsole` to 0 at the level of the environment, ie, in the call to `SolverFactory`.

Thus, no more changes are planned from my side on this.

rschwarz:
I looked into the failing tests, and apparently many of them fail because the log file now contains the additional line ```Set parameter LogToConsole to value 0```.

I do not know how to avoid that, but have accepted it in our application.

Maybe temporarily setting `OutputFlag = 0` (and then back) or similar would help.

pyomo/solvers/plugins/solvers/gurobi_direct.py
```diff
@@ -240,9 +240,9 @@ def _apply_solver(self):
...
+            self._solver_model.setParam('LogToConsole', 0)
```
jsiirola: The following should resolve the test failures
```python
        # We would like to just set OutputFlag to turn off the console
        # log, but that prevents users from using the Gurobi LogFile
        # option (see #3589 / #3716).  BUT just setting LogToConsole
        # triggers Gurobi to write a message to the console.  We will
        # capture that message here:
        with capture_output(capture_fd=True):
            self._solver_model.setParam('LogToConsole', int(bool(self._tee)))
```

rschwarz: This is replacing the existing code, right?

rschwarz: I added a commit using this version :crossed_fingers: 

(I also found that I had a partial fix locally, which I forgot to push, apparently. I added this as well, but reverted it, for the sake of documenting this option.)

jsiirola (approved):

mrmundt (approved):

PR #3699, opened 2025-08-12
Create API version attribute for different solver generations

mrmundt:
## Summary/Motivation:
We want an easier way for users/devs to query which API version of solvers they are using.

## Changes proposed in this PR:
- Create a standard `SolverAPIVersion` enum
- Add the enum as a classmethod to all solver base classes
- Update some tests

### Legal Acknowledgement

By contributing to this software project, I have read the [contribution guide](https://pyomo.readthedocs.io/en/stable/contribution_guide.html) and agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #3717, opened 2025-09-03
Validate an argument using the new api_version() method, also implement that method on wrapper classes

sadavis1:
## Summary/Motivation:
#3699 created a new api_version() method that can be queried to determine which api version a solver object supports. Use that method to validate a solver object argument in multiple_bigm.py, rather than using `isinstance` in a questionable way.

Along the way I noticed that contrib LegacySolverWrapper and appsi LegacySolverInterface class instances were returning V2 and APPSI as their api versions which is not correct, so I added an override to api_version() there.

- Improve argument validation
- Override api_version() in wrapper classes as presenting the v1 api is their whole purpose

### Legal Acknowledgement


mrmundt (approved):
We had a quick chat about this at the dev call, and the implementation is perfectly fine. I just had a residual question that I wanted answered.

emma58 (approved):

PR #3722, opened 2025-09-08
Bugfix: `ConfigDict.user_values()` should return only child entries, not instance itself

### LINKED ISSUE #3721: Clarify contents of `ConfigDict().__call__().user_values()`
## Summary

According to the [Pyomo 6.9.3 documentation of the Pyomo configuration system](https://pyomo.readthedocs.io/en/stable/explanation/developer_utils/config.html), the `ConfigDict` instance method `user_values()` can be used to iterate over the entries of a `ConfigDict`  instance that have been explicitly set by the user. However, if a `ConfigDict` instance is obtained using the `__call__` method of another `ConfigDict` instance, then the first iterate of the `user_values()` method of the new `ConfigDict` instance may unexpectedly be the new `ConfigDict` instance itself. This occurs when the argument `value` to `ConfigDict.__call__()` is set to a nonempty `dict`.

Set up a simple `ConfigDict`:

```python
>>> config = ConfigDict()
>>> config.declare(
...     "filename",
...     ConfigValue(
...         default=None,
...         domain=str,
...         description="Input file name",
...     ),
... )
```

Obtain a `ConfigDict` by invoking the `__call__` method with a valid nonempty `dict`:

```python
>>> cfg = config(value={"filename": "example.txt"})
```

Check user values for the new `ConfigDict`:

```python
>>> print("User values:")
>>> for val in cfg.user_values():
...     print(f" name={val.name()!r}, value={val.value()!r}, is cfg={val is cfg}")
User values:
 name='', value={'filename': 'example.txt'}, is cfg=True
 name='filename', value='example.txt', is cfg=False
```

Notice that the first iterate of `cfg.user_values()` is `cfg` itself. Our expectation was that the only iterate of `cfg.user_values()` would be the `ConfigValue` object associated with the entry called `"filename"`.

However, if a `ConfigDict` is obtained using `config.__call__()` with default arguments or with `value` set to an empty `dict`, then the `user_values()` method of the new `ConfigDict` does not contain the new `ConfigDict` itself (regardless of whether entries of the `ConfigDict` are subsequently set by the user):

```python
>>> cfg2 = config({})
>>> cfg2["filename"] = "example.txt"
>>> print("User values:")
>>> for val in cfg2.user_values():
...     print(f" name={val.name()!r}, value={val.value()!r}, is cfg2={val is cfg2}")
User values:
 name='filename', value='example.txt', is cfg2=False
```

Pyomo version: 6.9.5dev0 (main @ 613f2ad1f)
Python version: 3.12.7
Operating system: Ubuntu 20.04

mrmundt:
## Summary/Motivation:
@shermanjasonaf reported an issue with `ConfigDict.user_values()`. This PR fixes that problem by introducing a new private flag and adds two tests that mimic the examples given in the Issue.

- When making a ConfigDict from a dict, only the entries are marked as user-set, not the instance itself
- Tests

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/config.py
+++ b/pyomo/common/config.py
@@ -1600,13 +1600,12 @@ class ConfigBase:
 
         # ... and set the value, if appropriate
         if value is not NOTSET:
-            # Change in response to Pyomo/pyomo#3721
-            # If setting a ConfigDict from a (nonempty) dict during __call__,
-            # do not mark the container itself as user-set; only the children.
-            if isinstance(ans, ConfigDict) and isinstance(value, dict):
-                ans.set_value(value, _mark_container_userSet=False)
-            else:
-                ans.set_value(value)
+            # Note that because we are *creating* a new Config object,
+            # we do not want set_value() to change the current (default)
+            # userSet flag for this object/container (see #3721).
+            tmp = ans._userSet
+            ans.set_value(value)
+            ans._userSet = tmp
         return ans
 
     def name(self, fully_qualified=False):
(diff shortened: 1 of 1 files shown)


mrmundt:
Jenkins failures were related to #3750 , unrelated to anything else. So this is technically good to go.

PR #3723, opened 2025-09-08
Pyomo.DoE - Measurement Error Convention Correction

djlaky:
## Fixes # .
Fixed measurement error consideration during model construction to be squared.

## Summary/Motivation:
The convention for measurement error mathematically is to provide measurement error for each measurement, and then take the square of this value when construction the FIM. The squaring of this error was mistakenly left off. This PR will update the DoE problem construction to use the square of each experimental output's measurement error.

- Update measurement error to be squared
- Update test(s) to align with changes

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/test_branches.yml
+++ b/.github/workflows/test_branches.yml
@@ -352,10 +352,6 @@ jobs:
         if test "${{matrix.TARGET}}" == linux; then
             EXCLUDE="casadi numdifftools $EXCLUDE"
         fi
-        if [[ "${{matrix.TARGET}}" == win && "${{matrix.python}}" == 3.13* ]]; then
-            # As of Nov 7, 2024, qtconsole is not compatible with python 3.13 on win
-            EXCLUDE="qtconsole $EXCLUDE"
-        fi
         EXCLUDE=`echo "$EXCLUDE" | xargs`
         if test -n "$EXCLUDE"; then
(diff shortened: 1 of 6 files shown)


pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -551,15 +553,9 @@ def test_doe_full_factorial(self):
         self.assertStructuredAlmostEqual(
```
adowling2, line 553: Is there any reason why these tests specify `abstol` but the new ones below do not? If there is a technical reason, just add a comment.

adowling2 (approved):
This PR fixes a bug with the measurement error specification.

Does the document clearly state that the elements of `measurement_error` are expected to be the standard deviation?

pyomo/contrib/doe/tests/test_doe_solve.py
djlaky: Added a reason. More or less, it is because the squaring of the error has caused the absolute magnitude to exceed 1e10, meaning an abstol of 1e-4 is smaller than the precision of the methods to compute these metrics (e.g., determinant, minimum eigenvalue, and trace). The log values of these metrics can still be computed with the abstol of 1e-4.

blnicho (approved):

mrmundt (approved):

PR #3730, opened 2025-09-16
Separate `solver_available` and `license_available`

mrmundt:
## Summary/Motivation:
As part of the solver redesign, we decided to separate the checks for `solver_available` (is the solver findable, e.g., through importing, on the path, etc.) and `license_is_valid` (does the user have a valid license to use the solver).

This PR introduces a fundamental change in which each derived class is now expected to create both `solver_available` and `license_available`. The combination of these two results is then propagated to `available` (which overall returns a bool stating whether a solver is both findable and licensed).

- Create `pyomo.contrib.solver.common.availability` for `SolverAvailability` and `LicenseAvailability` enums
- Add `license_available` and `solver_available` to `SolverBase`
- Update all existing solvers to implement `license_available` and `solver_available` (and properly do caching for version and availability)
   - @michaelbynum - I'd love your eyes on the changes in gurobi in particular; they work but you might want it implemented differently
- Remove `config` as an argument in IPOPT
- Update / add tests

### Legal Acknowledgement


pyomo/contrib/solver/solvers/gurobi_direct.py
```diff
@@ -174,83 +178,169 @@ class GurobiSolverMixin:
...
+    def version(self) -> Optional[Tuple[int, int, int]]:
+        if self._version_cache is None:
+            self.__class__._version_cache = (
+                gurobipy.GRB.VERSION_MAJOR,
+                gurobipy.GRB.VERSION_MINOR,
+                gurobipy.GRB.VERSION_TECHNICAL,
+            )
+        return self._version_cache
+
+    def license_available(
```
michaelbynum, line 219: Should we separate `acquire_license` and `license_available`?

mrmundt: Probably. What is the right way to acquire a license in gurobi? Could you help with that?

michaelbynum: For anyone following along, we are discussing this offline.

pyomo/contrib/solver/solvers/gurobi_direct.py
```diff
@@ -174,83 +182,172 @@ class GurobiSolverMixin:
...
+        Responses:
+          - FullLicense   : can optimize a small model with >2000 vars
+          - LimitedLicense: can optimize only up to demo/community limits
+          - NotAvailable  : gurobi license not present/denied
+          - Timeout       : waited but could not check out
+          - BadLicense    : clearly invalid/corrupt license
+          - Unknown       : unexpected error states
+        """
+        if not self._gurobipy_available:
+            return LicenseAvailability.NotAvailable
```
michaelbynum: I think this should return unknown because we didn't actually check the license.

pyomo/contrib/solver/solvers/gurobi_direct.py
```diff
@@ -174,83 +182,172 @@ class GurobiSolverMixin:
...
+        # Quick check - already have license
+        if cls._gurobipy_env is not None:
+            return cls._gurobipy_env
+        if not timeout:
+            try:
+                cls._gurobipy_env = gurobipy.Env()
+            except:
+                pass
+            if cls._gurobipy_env is not None:
+                return cls._gurobipy_env
```
michaelbynum: Couldn't this just be `return cls._gurobipy_env` without the if statement?

pyomo/contrib/solver/solvers/gurobi_direct.py
```diff
@@ -174,83 +182,172 @@ class GurobiSolverMixin:
...
+            return LicenseAvailability.NotAvailable
+        if not recheck and self._license_cache is not None:
+            return self._license_cache
+
+        with capture_output(capture_fd=True):
+            # Try to bring up an environment (this is where a license is often checked)
+            try:
+                env = self.acquire_license(timeout=timeout)
+                if env is None:
+                    self._license_cache = LicenseAvailability.Timeout
```
michaelbynum, line 242: It is hard to distinguish between a timeout and no license... It looks like it is possible to return Timeout when the result should actually be NotAvailable?

pyomo/contrib/solver/solvers/gurobi_direct.py
```diff
@@ -174,83 +182,172 @@ class GurobiSolverMixin:
...
+                    return self._license_cache
+            except gurobipy.GurobiError as acquire_error:
```
michaelbynum, line 244: Is it possible to hit this exception? It looks like `acquire_license` is guarded with `try` statements in a way that this code cannot be reached?

pyomo/contrib/solver/solvers/gurobi_direct.py
```diff
@@ -174,83 +182,172 @@ class GurobiSolverMixin:
...
+                # Distinguish timeout vs unavailable vs bad license
+                status = getattr(acquire_error, "errno", None)
+                msg = str(acquire_error).lower()
+                if "queue" in msg or "timeout" in msg:
```
michaelbynum, line 248: I am not familiar with all of the gurobi errors. Did you get these from documentation or testing?

pyomo/contrib/solver/common/availability.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    Runtime status for licensing. Independent from
+    overall solver availability. A return value > 0 is "usable in some form".
+    """
+
+    FullLicense = 3
+    LimitedLicense = 2
+    NotApplicable = 1
+    NotAvailable = 0
+    BadLicense = -1
```
michaelbynum, line 46: What is the intended difference between NotAvailable and BadLicense?

pyomo/contrib/solver/solvers/gurobi_direct.py
```diff
@@ -174,83 +182,172 @@ class GurobiSolverMixin:
...
+                elif "queue" in msg or "timeout" in msg:
+                    self._license_cache = LicenseAvailability.Timeout
+                elif (
+                    "no gurobi license" in msg
+                    or "not licensed" in msg
+                    or status in (10009,)
+                ):
+                    self._license_cache = LicenseAvailability.NotAvailable
+                else:
+                    self._license_cache = LicenseAvailability.BadLicense
```
michaelbynum: Once we have gotten here, don't we know that the result should be LimitedLicense? Is anything else possible?

pyomo/contrib/solver/solvers/gurobi_persistent.py
```diff
@@ -272,13 +272,24 @@ def __init__(self, **kwds):
         self._vars_added_since_update = ComponentSet()
         self._last_results_object: Optional[Results] = None
 
+    def close(self):
```
michaelbynum, line 275: What motivated this new method?

pyomo/contrib/solver/tests/solvers/test_gurobi_direct.py
```diff
@@ -0,0 +1,227 @@
+#  ___________________________________________________________________________
```
michaelbynum, line 1: This file is probably going to cause major conflicts with #3698...

pyomo/contrib/solver/common/availability.py
mrmundt: I consider `NotAvailable` to be "I can't find a license" and `BadLicense` to be "There is something wrong with your license (e.g., expired)"

pyomo/contrib/solver/solvers/gurobi_direct.py
mrmundt: I agree - good call.

pyomo/contrib/solver/solvers/gurobi_direct.py
mrmundt: I messed with this one a lot and am 1000% not sure what the right thing to do is. So basically, if acquire_license times out, it is guaranteed to return `None`. I think otherwise, it will error. But I was having a hard time testing to ensure that that did actually happen.

pyomo/contrib/solver/solvers/gurobi_direct.py
mrmundt: There isn't a specific error message that I could find for "timeout" on Gurobi's side, but there is one related to queuing: https://docs.gurobi.com/projects/optimizer/en/current/reference/numericcodes/errors.html <- see "JOB_REJECTED". I couldn't actually create a response like that when testing, so I have _no_ idea what the error message says. Ergo, this is my "best guess" that it may have something to do with queue or timeout.

pyomo/contrib/solver/solvers/gurobi_persistent.py
mrmundt: See the comment below on `release_license` - because I made a classmethod version, there was a naming clash, so I needed a bit of a hack to get around it.

pyomo/contrib/solver/tests/solvers/test_gurobi_direct.py
mrmundt: I don't think so! You touched `test_gurobi_persistent.py`; this is a brand-new file.

pyomo/contrib/solver/solvers/gurobi_direct.py
mrmundt: Yup, totally right. Fixed!

pyomo/contrib/solver/solvers/gurobi_direct.py
mrmundt: Ooo. That's a good point. What about this instead?
```python
        if not timeout:
            try:
                cls._gurobipy_env = gurobipy.Env()
            except gurobipy.GurobiError:
                # Re-raise so license_available can inspect further
                # or so users can explicitly view the error
                raise
            return cls._gurobipy_env
```

pyomo/contrib/solver/solvers/gurobi_direct.py
mrmundt: I think a timeout or "some other unknown thing" is possible. But not `NotAvailable`, you're right. What about this?
```python
            try:
                # We try a 'big' model (more than 2000 vars).
                # This should give us all the information we need
                # about the license status.
                large_model = gurobipy.Model(env=env)
                large_model.addVars(range(2001))
                large_model.optimize()
                self._license_cache = LicenseAvailability.FullLicense
            except gurobipy.GurobiError as large_error:
                msg = str(large_error).lower()
                status = getattr(large_error, "errno", None)
                if "too large" in msg or status in (10010,):
                    self._license_cache = LicenseAvailability.LimitedLicense
                elif "queue" in msg or "timeout" in msg:
                    self._license_cache = LicenseAvailability.Timeout
                else:
                    self._license_cache = LicenseAvailability.Unknown
            finally:
                large_model.dispose()
```

pyomo/contrib/solver/solvers/gurobi_direct.py
michaelbynum: That looks good, but I'm not sure there is a point to the `try` statement at this point???

pyomo/contrib/solver/solvers/gurobi_direct.py
michaelbynum: Looks good, I think.

pyomo/contrib/solver/tests/solvers/test_gurobi_direct.py
michaelbynum: Oh, you are right.

pyomo/contrib/solver/solvers/gurobi_direct.py
mrmundt: The only thing I see that might be useful here is, "What if there is some _other_ type of exception that isn't a GurobiError?" - In that case, we are just returning `None`. But do we want to do that? Is it better if we just let it explode?

michaelbynum: Sorry - I was not clear. I mean the `try` statement in `acquire_license`, not the `try` statement here.

pyomo/contrib/solver/solvers/gurobi_direct.py
```diff
@@ -174,83 +182,175 @@ class GurobiSolverMixin:
...
+        # Quick check - already have license
+        if cls._gurobipy_env is not None:
+            return cls._gurobipy_env
+        if not timeout:
+            try:
+                cls._gurobipy_env = gurobipy.Env()
+            except gurobipy.GurobiError:
+                # Re-raise so license_available can inspect further
+                # or so users can explicitly view the error
+                raise
```
michaelbynum, line 297: This is what I am talking about. This `try` statement does not seem to do anything except raise any exception that gets raised???

michaelbynum (approved):
My remaining comments don't really matter. I'm happy at this point.

mrmundt:
I'm re-working this one, so closing

PR #3735, opened 2025-09-29
Restore Xpress workflow

XPRSc4v4:
Hi!

I looked into the `xpress.init()` hanging issue, and it seems to have been resolved with newer versions of Xpress. 

This PR simply removes the specialized code that was added to avoid the hanging scenario.

Feel free to let me know if you still see issues with the workflow!

Thanks,
Francesco

### DIFF
--- a/.github/workflows/test_branches.yml
+++ b/.github/workflows/test_branches.yml
@@ -375,22 +375,10 @@ jobs:
         # possibly if it outputs messages to stderr)
         conda install --update-deps -q -y python="${{matrix.python}}" $CONDA_DEPENDENCIES
         if test -z "${{matrix.slim}}"; then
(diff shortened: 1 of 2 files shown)


mrmundt (approved):
@jsiirola and I had an offline discussion about this. We like the breadcrumbs, but this extra code truly is defunct (and has been for a while). That it exists in the history is probably good enough at this point.

mrmundt:
@XPRSc4v4 - It turns out this code isn't as defunct as we thought. It was preventing xpress from being installed on win/3.10. With this change, 9.1.0 is now installed, and this causes an infinite hang when attempting to report plugin information. C'est la vie. When we eventually retire Python 3.10 (after October 2026), we can remove this.

PR #3738, opened 2025-09-30
Bugfix: IPOPT log parser and no objective case

### LINKED ISSUE #3736: Issues with new solver interface for IPOPT
## Summary
Recently, when updating a contribution to the IDAES-Examples repo to use the new solver interface, we encountered several errors. 

1) The first issue is an error in the IPOPT output log parser. When the option `"OF_print_info_string": "yes",` is passed, there is an additional column in the IPOPT output:

```console
iter    objective    inf_pr   inf_du lg(mu)  ||d||  lg(rg) alpha_du alpha_pr  ls
   0  4.3126674e+00 1.34e+00 1.00e+00  -5.0 0.00e+00    -  0.00e+00 0.00e+00   0
Reallocating memory for MA57: lfact (2247250)
   1r 4.3126674e+00 1.34e+00 9.99e+02   0.1 0.00e+00  -4.0 0.00e+00 3.29e-10R  2
   2r 3.0519246e+08 1.13e+00 9.90e+02   0.1 2.30e+02    -  2.60e-02 9.32e-03f  1
   3r 2.2712595e+09 1.69e+00 9.73e+02   0.1 2.23e+02    -  2.54e-02 1.71e-02f  1 Nhj
   4  2.2712065e+09 1.69e+00 1.37e+09  -5.0 3.08e+03    -  1.32e-05 1.17e-05f  1 q
   5  1.9062986e+09 1.55e+00 1.25e+09  -5.0 5.13e+03    -  1.19e-01 8.38e-02f  1
```

This leads to an error [in the log parser](https://github.com/Pyomo/pyomo/blob/cb52a0285d06f78c6e376265b1686f10b5e07968/pyomo/contrib/solver/solvers/ipopt.py#L614C1-L617C29), because any lines in the log that don't have the expected number of columns are skipped:

```
for line in iter_table:
    tokens = line.strip().split()
    if len(tokens) != len(columns):
        continue
```

2) The second issue is an error in the solver reader. When dual variables are requested for a problem without an objective function, [this line](https://github.com/Pyomo/pyomo/blob/cb52a0285d06f78c6e376265b1686f10b5e07968/pyomo/contrib/solver/solvers/sol_reader.py#L141) causes an exception:

```
            obj_scale = self._nl_info.scaling.objectives[0]
```

For problems without objectives, either 1) the dual variables are zero or 2) the problem is infeasible. Nevertheless, a better error message should be raised.

### Error Message

IPOPT output log parser error:
```console
Solving controller model at time = 0.000000
Ipopt 3.13.2: linear_solver="ma57"
max_iter=1000
nlp_scaling_method="gradient-based"
tol=0.0001
halt_on_ampl_error="no"
mu_init=1e-05
bound_push=1e-06
bound_relax_factor=1e-06
ma57_pivtol=1e-06
warm_start_init_point="yes"
warm_start_mult_bound_push=1e-06
warm_start_bound_push=1e-06
option_file_name="C:\Users\dallan\AppData\Local\Temp\tmps0xkth8o\unknown.26400.33444.opt"

Using option file "C:\Users\dallan\AppData\Local\Temp\tmps0xkth8o\unknown.26400.33444.opt".


******************************************************************************
This program contains Ipopt, a library for large-scale nonlinear optimization.
 Ipopt is released as open source code under the Eclipse Public License (EPL).
         For more information visit http://projects.coin-or.org/Ipopt

This version of Ipopt was compiled from source code available at
    https://github.com/IDAES/Ipopt as part of the Institute for the Design of
    Advanced Energy Systems Process Systems Engineering Framework (IDAES PSE
    Framework) Copyright (c) 2018-2019. See https://github.com/IDAES/idaes-pse.

This version of Ipopt was compiled using HSL, a collection of Fortran codes
    for large-scale scientific computation.  All technical papers, sales and
    publicity material resulting from use of the HSL codes within IPOPT must
    contain the following acknowledgement:
        HSL, a collection of Fortran codes for large-scale scientific
        computation. See http://www.hsl.rl.ac.uk/.
******************************************************************************

This is Ipopt version 3.13.2, running with linear solver ma57.

Number of nonzeros in equality constraint Jacobian...:    77541
Number of nonzeros in inequality constraint Jacobian.:        0
Number of nonzeros in Lagrangian Hessian.............:    51855

Total number of variables............................:    15468
                     variables with only lower bounds:     3491
                variables with lower and upper bounds:     5026
                     variables with only upper bounds:      186
Total number of equality constraints.................:    15417
Total number of inequality constraints...............:        0
        inequality constraints with only lower bounds:        0
   inequality constraints with lower and upper bounds:        0
        inequality constraints with only upper bounds:        0

iter    objective    inf_pr   inf_du lg(mu)  ||d||  lg(rg) alpha_du alpha_pr  ls
   0  4.3126674e+00 1.34e+00 1.00e+00  -5.0 0.00e+00    -  0.00e+00 0.00e+00   0
Reallocating memory for MA57: lfact (2247250)
   1r 4.3126674e+00 1.34e+00 9.99e+02   0.1 0.00e+00  -4.0 0.00e+00 3.29e-10R  2
   2r 3.0519246e+08 1.13e+00 9.90e+02   0.1 2.30e+02    -  2.60e-02 9.32e-03f  1
   3r 2.2712595e+09 1.69e+00 9.73e+02   0.1 2.23e+02    -  2.54e-02 1.71e-02f  1 Nhj
   4  2.2712065e+09 1.69e+00 1.37e+09  -5.0 3.08e+03    -  1.32e-05 1.17e-05f  1 q
   5  1.9062986e+09 1.55e+00 1.25e+09  -5.0 5.13e+03    -  1.19e-01 8.38e-02f  1
   6  1.7041594e+09 1.46e+00 1.18e+09  -5.0 5.66e+03    -  7.06e-02 5.45e-02f  1
   7  1.4763158e+09 1.36e+00 1.10e+09  -5.0 3.94e+03    -  2.30e-01 6.92e-02f  1
   8  8.5873108e+08 1.04e+00 8.41e+08  -5.0 2.38e+05    -  3.49e-06 2.37e-01f  1
   9  4.4215572e+08 7.45e-01 6.03e+08  -5.0 1.63e+06    -  7.97e-02 2.82e-01f  1
iter    objective    inf_pr   inf_du lg(mu)  ||d||  lg(rg) alpha_du alpha_pr  ls
  10  5.0251884e+01 1.65e-01 1.57e+04  -5.0 1.24e+06    -  3.92e-05 1.00e+00f  1
  11  4.9121733e+01 4.97e-02 4.68e+03  -5.0 8.11e+04    -  4.31e-02 7.01e-01h  1
  12  4.1483985e+01 2.24e-02 5.97e+03  -5.0 1.15e+06    -  5.93e-02 1.00e+00f  1
  13  3.5762585e+01 1.75e-02 5.00e+03  -5.0 1.03e+06    -  1.25e-01 1.00e+00f  1
  14  3.2291014e+01 1.08e-02 3.51e+03  -5.0 8.25e+05    -  6.68e-01 1.00e+00f  1
  15  3.2274630e+01 3.31e-05 1.17e+00  -5.0 4.26e+04    -  9.92e-01 1.00e+00h  1
  16  3.2274631e+01 7.45e-09 2.71e-03  -5.0 6.11e+02    -  8.97e-01 1.00e+00h  1
  17  3.2274635e+01 7.45e-09 2.35e-03  -5.0 2.71e+04    -  1.32e-01 1.00e+00f  1
  18  3.2274635e+01 7.45e-09 1.15e-04  -5.0 5.53e+03    -  9.51e-01 1.00e+00h  1
  19  3.2274635e+01 7.45e-09 2.84e-05  -5.0 4.41e+04    -  7.54e-01 1.00e+00f  1
iter    objective    inf_pr   inf_du lg(mu)  ||d||  lg(rg) alpha_du alpha_pr  ls
  20  3.2274635e+01 7.45e-09 8.54e-07  -5.0 1.83e+04    -  1.00e+00 1.00e+00h  1

Number of Iterations....: 20

                                   (scaled)                 (unscaled)
Objective...............:   3.2274635418964841e+01    3.2274635418964841e+01
Dual infeasibility......:   8.5365078678328669e-07    8.5365078678328669e-07
Constraint violation....:   8.0780625068607930e-13    7.4505805969238281e-09
Complementarity.........:   1.2275904566414160e-05    1.2275904566414160e-05
Overall NLP error.......:   1.2275904566414160e-05    1.2275904566414160e-05


Number of objective function evaluations             = 23
Number of objective gradient evaluations             = 20
Number of equality constraint evaluations            = 23
Number of inequality constraint evaluations          = 0
Number of equality constraint Jacobian evaluations   = 22
Number of inequality constraint Jacobian evaluations = 0
Number of Lagrangian Hessian evaluations             = 20
Total CPU secs in IPOPT (w/o function evaluations)   =     10.450
Total CPU secs in NLP function evaluations           =      1.651

EXIT: Optimal Solution Found.
AssertionError                            Traceback (most recent call last)
File ~\Repos\scratch\SOC-Jupyternotebook\rSOC.py:475
    469 # ## Run NMPC
    470
    471 # In[8]:
    474 economic = True
--> 475 data_dict_updated_v2 = run_nmpc_v2(economic= economic, sim_time_point = sim_time_point)
    478 # In[ ]:
    481 data_dict_updated_v1 = run_nmpc_v1(economic= economic, sim_time_point = sim_time_point)

File ~\Repos\scratch\SOC-Jupyternotebook\rSOC.py:422, in run_nmpc_v2(economic, sim_time_point)
    420 # Solve the scaled controller model
    421 print("Solving controller model at time = %f" %t_base)
--> 422 controller_results = solver_v2.solve(m_controller, tee= True)
    423 pyo.assert_optimal_termination(controller_results)
    425 # Apply control actions to the plant model

File ~\miniforge3\envs\idaes\lib\site-packages\pyomo\contrib\solver\common\base.py:685, in LegacySolverWrapper.solve(self, model, tee, load_solutions, logfile, solnfile, timelimit, report_timing, solver_io, suffixes, options, keepfiles, symbolic_solver_labels, raise_exception_on_nonoptimal_result, solver_options, writer_config)
    682 filtered_args = {k: loc[k] for k in map_args if loc.get(k, None) is not None}
    683 self._map_config(**filtered_args)
--> 685 results: Results = super().solve(model)
    686 legacy_results, legacy_soln = self._map_results(model, results)
    687 legacy_results = self._solution_handler(
    688     load_solutions, model, results, legacy_results, legacy_soln
    689 )

File ~\miniforge3\envs\idaes\lib\site-packages\pyomo\contrib\solver\solvers\ipopt.py:467, in Ipopt.solve(self, model, **kwds)
    463         timer.stop('subprocess')
    465     # This is the data we need to parse to get the iterations
    466     # and time
--> 467     parsed_output_data = self._parse_ipopt_output(ostreams[0])
    469 if proven_infeasible:
    470     results = Results()

File ~\miniforge3\envs\idaes\lib\site-packages\pyomo\contrib\solver\solvers\ipopt.py:645, in Ipopt._parse_ipopt_output(self, output)
    639             except (ValueError, TypeError):
    640                 logger.warning(
    641                     "Error converting Ipopt log entry to "
    642                     f"float:\n\t{sys.exc_info()[1]}\n\t{line}"
    643                 )
--> 645     assert len(iterations) == iter_data.pop('iter'), (
    646         f"Parsed row in the iterations table\n\t{line}\ndoes not "
    647         f"match the next expected iteration number ({len(iterations)})"
    648     )
    649     iterations.append(iter_data)
    651 parsed_data['iteration_log'] = iterations

AssertionError: Parsed row in the iterations table
           5  1.9062986e+09 1.55e+00 1.25e+09  -5.0 5.13e+03    -  1.19e-01 8.38e-02f  1
does not match the next expected iteration number (3)
```

Dual variable interface error:

```console
Solving plant model at time = 0.000000
Ipopt 3.13.2: linear_solver="ma57"
max_iter=1000
nlp_scaling_method="gradient-based"
tol=0.0001
halt_on_ampl_error="no"
mu_init=1e-05
bound_push=1e-06
bound_relax_factor=1e-06
ma57_pivtol=1e-06
warm_start_init_point="yes"
warm_start_mult_bound_push=1e-06
warm_start_bound_push=1e-06
option_file_name="C:\Users\dallan\AppData\Local\Temp\tmp4k0guac4\unknown.26340.17868.opt"

Using option file "C:\Users\dallan\AppData\Local\Temp\tmp4k0guac4\unknown.26340.17868.opt".


******************************************************************************
This program contains Ipopt, a library for large-scale nonlinear optimization.
 Ipopt is released as open source code under the Eclipse Public License (EPL).
         For more information visit http://projects.coin-or.org/Ipopt

This version of Ipopt was compiled from source code available at
    https://github.com/IDAES/Ipopt as part of the Institute for the Design of
    Advanced Energy Systems Process Systems Engineering Framework (IDAES PSE
    Framework) Copyright (c) 2018-2019. See https://github.com/IDAES/idaes-pse.

This version of Ipopt was compiled using HSL, a collection of Fortran codes
    for large-scale scientific computation.  All technical papers, sales and
    publicity material resulting from use of the HSL codes within IPOPT must
    contain the following acknowledgement:
        HSL, a collection of Fortran codes for large-scale scientific
        computation. See http://www.hsl.rl.ac.uk.
******************************************************************************

This is Ipopt version 3.13.2, running with linear solver ma57.

Number of nonzeros in equality constraint Jacobian...:    34545
Number of nonzeros in inequality constraint Jacobian.:        0
Number of nonzeros in Lagrangian Hessian.............:    21570

Total number of variables............................:     8896
                     variables with only lower bounds:     1441
                variables with lower and upper bounds:     3413
                     variables with only upper bounds:       60
Total number of equality constraints.................:     8896
Total number of inequality constraints...............:        0
        inequality constraints with only lower bounds:        0
   inequality constraints with lower and upper bounds:        0
        inequality constraints with only upper bounds:        0

iter    objective    inf_pr   inf_du lg(mu)  ||d||  lg(rg) alpha_du alpha_pr  ls
   0  0.0000000e+00 1.82e-01 1.00e+00  -5.0 0.00e+00    -  0.00e+00 0.00e+00   0
Reallocating memory for MA57: lfact (707336)
   1  0.0000000e+00 5.57e-03 8.49e+00  -5.0 7.33e-02    -  5.67e-01 1.00e+00h  1
   2  0.0000000e+00 8.06e-06 1.47e-01  -5.0 7.33e-02    -  9.95e-01 1.00e+00h  1

Number of Iterations....: 2

                                   (scaled)                 (unscaled)
Objective...............:   0.0000000000000000e+00    0.0000000000000000e+00
Dual infeasibility......:   0.0000000000000000e+00    0.0000000000000000e+00
Constraint violation....:   8.0648345965883550e-06    8.0648345965883550e-06
Complementarity.........:   0.0000000000000000e+00    0.0000000000000000e+00
Overall NLP error.......:   8.0648345965883550e-06    8.0648345965883550e-06


Number of objective function evaluations             = 3
Number of objective gradient evaluations             = 3
Number of equality constraint evaluations            = 3
Number of inequality constraint evaluations          = 0
Number of equality constraint Jacobian evaluations   = 3
Number of inequality constraint Jacobian evaluations = 0
Number of Lagrangian Hessian evaluations             = 2
Total CPU secs in IPOPT (w/o function evaluations)   =      0.409
Total CPU secs in NLP function evaluations           =      0.046

EXIT: Optimal Solution Found.
IndexError                                Traceback (most recent call last)
File ~\Repos\scratch\SOC-Jupyternotebook\rSOC.py:475
    469 # ## Run NMPC
    470
    471 # In[8]:
    474 economic = True
--> 475 data_dict_updated_v2 = run_nmpc_v2(economic= economic, sim_time_point = sim_time_point)
    478 # In[ ]:
    481 data_dict_updated_v1 = run_nmpc_v1(economic= economic, sim_time_point = sim_time_point)

File ~\Repos\scratch\SOC-Jupyternotebook\rSOC.py:443, in run_nmpc_v2(economic, sim_time_point)
    441 # Solve plant model
    442 print("Solving plant model at time = %f" %t_base)
--> 443 plant_results = solver_v2.solve(m_plant, tee= True)
    444 pyo.assert_optimal_termination(plant_results)
    446 # Shift time points in plant

File ~\miniforge3\envs\idaes\lib\site-packages\pyomo\contrib\solver\common\base.py:685, in LegacySolverWrapper.solve(self, model, tee, load_solutions, logfile, solnfile, timelimit, report_timing, solver_io, suffixes, options, keepfiles, symbolic_solver_labels, raise_exception_on_nonoptimal_result, solver_options, writer_config)
    682 filtered_args = {k: loc[k] for k in map_args if loc.get(k, None) is not None}
    683 self._map_config(**filtered_args)
--> 685 results: Results = super().solve(model)
    686 legacy_results, legacy_soln = self._map_results(model, results)
    687 legacy_results = self._solution_handler(
    688     load_solutions, model, results, legacy_results, legacy_soln
    689 )

File ~\miniforge3\envs\idaes\lib\site-packages\pyomo\contrib\solver\solvers\ipopt.py:538, in Ipopt.solve(self, model, **kwds)
    532 results.solution_loader.load_vars()
    533 if (
    534     hasattr(model, 'dual')
    535     and isinstance(model.dual, Suffix)
    536     and model.dual.import_enabled()
    537 ):
--> 538     model.dual.update(results.solution_loader.get_duals())
    539 if (
    540     hasattr(model, 'rc')
    541     and isinstance(model.rc, Suffix)
    542     and model.rc.import_enabled()
    543 ):
    544     model.rc.update(results.solution_loader.get_reduced_costs())

File ~\miniforge3\envs\idaes\lib\site-packages\pyomo\contrib\solver\solvers\sol_reader.py:141, in SolSolutionLoader.get_duals(self, cons_to_load)
    139 else:
    140     scale_list = self._nl_info.scaling.constraints
--> 141     obj_scale = self._nl_info.scaling.objectives[0]
    142 if cons_to_load is None:
    143     cons_to_load = set(self._nl_info.constraints)

IndexError: list index out of range
```

Pyomo version: 6.9.4
Python version: 3.10
Operating system: Windows 11
How Pyomo was installed: IDAES advanced user installation
Solver (if applicable): IPOPT_v2

mrmundt:
## Summary/Motivation:
The issue above mentions two bugs that are related to IPOPT/parsing both the output and the `sol` files. This PR fixes both of those bugs and adds tests.

- Make IPOPT parser more robust
- Set duals and RCs to 0 when no objective is supplied
- Add tests!

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/solver/solvers/ipopt.py
+++ b/pyomo/contrib/solver/solvers/ipopt.py
@@ -63,43 +63,6 @@ logger = logging.getLogger(__name__)
 # Acceptable chars for the end of the alpha_pr column
 # in ipopt's output, per https://coin-or.github.io/Ipopt/OUTPUT.html
 _ALPHA_PR_CHARS = set("fFhHkKnNRwSstTr")
-_DIAGNOSTIC_TAGS = set(
-    {
-        "!",
-        "A",
-        "a",
-        "C",
-        "Dh",
-        "Dhj",
-        "Dj",
-        "dx",
-        "e",
-        "F-",
-        "F+",
-        "L",
-        "l",
-        "M",
-        "Nh",
-        "Nhj",
-        "Nj",
-        "NW",
-        "q",
-        "R",
-        "S",
-        "s",
-        "Tmax",
-        "W",
-        "w",
-        "Wb",
-        "We",
-        "Wp",
-        "Wr",
-        "Ws",
-        "WS",
-        "y",
-        "z",
-    }
-)
 
 
 class IpoptConfig(SolverConfig):
@@ -505,7 +468,6 @@ class Ipopt(SolverBase):
                 # This is the data we need to parse to get the iterations
                 # and time
                 parsed_output_data = self._parse_ipopt_output(ostreams[0])
-                parsed_output_message = f"Parsed solver data: {parsed_output_data}"
 
             if proven_infeasible:
                 results = Results()
@@ -554,7 +516,7 @@ class Ipopt(SolverBase):
                             logging.WARNING,
                             "The solver output data is empty or incomplete.\n"
                             f"Full error message: {e}\n"
-                            f"{parsed_output_message}\n",
+                            f"Parsed solver data: {parsed_output_data}\n",
                         )
         if (
             config.raise_exception_on_nonoptimal_result
@@ -648,8 +610,6 @@ class Ipopt(SolverBase):
                 "alpha_pr",
                 "ls",
             ]
-            numerical_columns = set(columns)
-            numerical_columns.remove('iter')
             iterations = []
             n_expected_columns = len(columns)
 
@@ -694,17 +654,11 @@ class Ipopt(SolverBase):
 
                 # Capture optional IPOPT diagnostic tags if present
                 if extra_tokens:
-                    if all(tok in _DIAGNOSTIC_TAGS for tok in extra_tokens):
-                        iter_data['diagnostic_tags'] = "".join(extra_tokens)
-                    else:
-                        logger.warning(
-                            f"Unrecognized Ipopt diagnostic tags {extra_tokens} on line: {line}"
-                        )
+                    iter_data['diagnostic_tags'] = " ".join(extra_tokens)
 
                 # Attempt to cast all values to float where possible
-                for key, val in iter_data.items():
-                    if key not in numerical_columns:
-                        continue
+                for key in columns[1:]:
+                    val = iter_data[key]
                     if val == '-':
                         iter_data[key] = None
                     else:


pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -115,6 +115,12 @@ def get_reduced_costs(
         self, vars_to_load: Optional[Sequence[VarData]] = None
     ) -> Mapping[VarData, float]:
         self._error_check()
+        # If the NL instance has no objectives, report zeros
+        if not getattr(self._nl_info, "objectives", None):
+            vars_ = (
+                vars_to_load if vars_to_load is not None else self._nl_info.variable_map
+            )
+            return {v: 0.0 for v in vars_}
```
dallan-keylogic: This fix is suitable if IPOPT converges to a feasible point, but if it converges to a point of local infeasibility, the values of the dual variables convey useful information about which bounds may be active at a solution.

mrmundt: Per your suggestion, though, if there are no objectives, these should be set to 0. So I don't think we change this. We capture the raw logs anyway, so folks can inspect those if they want more details.

jsiirola: Would it be better / more appropriate to return `{}` here?  Unspecified duals / reduced costs are assumed to be 0, right?

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -610,41 +616,68 @@ def _parse_ipopt_output(self, output: Union[str, io.StringIO]) -> Dict[str, Any]
...
+                if step_acceptance_tag in _ALPHA_PR_CHARS:
+                    iter_data['step_acceptance'] = step_acceptance_tag
                     iter_data['alpha_pr'] = iter_data['alpha_pr'][:-1]
                 else:
                     iter_data['step_acceptance'] = None
 
+                # Capture optional IPOPT info string if present
+                if extra_tokens:
+                    # Typically a single string like 'q' or 'Nhj'; join just in case
+                    iter_data['info_string'] = " ".join(extra_tokens)
```
jsiirola: The Ipopt documentation refers to these as "diagnostic tags": https://coin-or.github.io/Ipopt/OUTPUT.html

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -610,41 +616,68 @@ def _parse_ipopt_output(self, output: Union[str, io.StringIO]) -> Dict[str, Any]
...
+                # Require at least the base columns
+                # Additionally allow optional trailing info-string(s)
+                if len(tokens) < n_expected_columns:
                     continue
-                iter_data = dict(zip(columns, tokens))
+
+                expected_column_tokens = tokens[:n_expected_columns]
+                extra_tokens = tokens[n_expected_columns:]
+
+                iter_data = dict(zip(columns, expected_column_tokens))
```
jsiirola: You don't need to extract the expected list.  zip will stop when the first iterable is exhausted.


```suggestion
                iter_data = dict(zip(columns, tokens))
                extra_tokens = tokens[n_expected_columns:]
```

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -610,41 +616,68 @@ def _parse_ipopt_output(self, output: Union[str, io.StringIO]) -> Dict[str, Any]
...
             iterations = []
+            n_expected_columns = len(columns)
 
             for line in iter_table:
                 tokens = line.strip().split()
-                if len(tokens) != len(columns):
+                # Require at least the base columns
+                # Additionally allow optional trailing info-string(s)
+                if len(tokens) < n_expected_columns:
                     continue
```
jsiirola: Are there any lines that match this condition?

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -115,6 +115,12 @@ def get_reduced_costs(
         self, vars_to_load: Optional[Sequence[VarData]] = None
     ) -> Mapping[VarData, float]:
         self._error_check()
+        # If the NL instance has no objectives, report zeros
+        if len(self._nl_info.objectives) == 0:
+            vars_ = (
+                vars_to_load if vars_to_load is not None else self._nl_info.variables
+            )
+            return ComponentMap((v, 0.0) for v in vars_)
```
jsiirola: Is it more correct to just return an empty ComponentMap?  Aren't missing entries assumed to be 0?

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -610,41 +616,68 @@ def _parse_ipopt_output(self, output: Union[str, io.StringIO]) -> Dict[str, Any]
...
-                iter_data = dict(zip(columns, tokens))
+
+                expected_column_tokens = tokens[:n_expected_columns]
+                extra_tokens = tokens[n_expected_columns:]
+
+                iter_data = dict(zip(columns, expected_column_tokens))
 
                 # Extract restoration flag from 'iter'
-                iter_data['restoration'] = iter_data['iter'].endswith('r')
+                iter_str = iter_data['iter']
```
jsiirola: Why don't you pop the iter off here?  There doesn't appear to be a good reason to keep it around.  I would suggest something like:
```python
iter = iter_data.pop('iter')
iter_data['restoration'] = r = (iter[-1] == 'r')
if r:
    iter = int(iter[:-1])
else:
    iter = int(iter)
```

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -610,41 +616,68 @@ def _parse_ipopt_output(self, output: Union[str, io.StringIO]) -> Dict[str, Any]
...
                 else:
                     iter_data['step_acceptance'] = None
 
+                # Capture optional IPOPT info string if present
+                if extra_tokens:
+                    # Typically a single string like 'q' or 'Nhj'; join just in case
+                    iter_data['info_string'] = " ".join(extra_tokens)
+
                 # Attempt to cast all values to float where possible
                 for key in columns:
```
jsiirola: Why is this not `for key, val in iter_data.items()`?

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -610,41 +616,68 @@ def _parse_ipopt_output(self, output: Union[str, io.StringIO]) -> Dict[str, Any]
...
+                    iter_num = int(iter_data.pop('iter'))
+                except ValueError:
+                    logger.warning(
+                        "Could not parse IPOPT iteration number from line:\n\t%s", line
+                    )
+                    continue
+
+                assert len(iterations) == iter_num, (
+                    f"Parsed row in the iterations table\n\t{line}\n"
+                    f"does not match the next expected iteration number ({len(iterations)})"
```
jsiirola: Performance question: does Python form / format the message string even when the assertion passes?  If it were a function call it definitely would (and that would be a performance hit), but I don't know if the interpreter has special handling for assertions...

(answering my own question...
```python
>>> assert 0 == 0, f"{xyz}"
>>> assert 1 == 0, f"{xyz}"
NameError: name 'xyz' is not defined
```
)

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -610,41 +616,68 @@ def _parse_ipopt_output(self, output: Union[str, io.StringIO]) -> Dict[str, Any]
...
 
-                assert len(iterations) == iter_data.pop('iter'), (
-                    f"Parsed row in the iterations table\n\t{line}\ndoes not "
-                    f"match the next expected iteration number ({len(iterations)})"
+                try:
+                    iter_num = int(iter_data.pop('iter'))
+                except ValueError:
+                    logger.warning(
+                        "Could not parse IPOPT iteration number from line:\n\t%s", line
+                    )
```
jsiirola, line 677: Is this only a warning?  I would assume it is an error.

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -610,41 +616,68 @@ def _parse_ipopt_output(self, output: Union[str, io.StringIO]) -> Dict[str, Any]
...
+                    iter_data['step_acceptance'] = step_acceptance_tag
                     iter_data['alpha_pr'] = iter_data['alpha_pr'][:-1]
                 else:
                     iter_data['step_acceptance'] = None
 
+                # Capture optional IPOPT info string if present
+                if extra_tokens:
+                    # Typically a single string like 'q' or 'Nhj'; join just in case
+                    iter_data['info_string'] = " ".join(extra_tokens)
+
```
jsiirola, line 658: Design question: you have set this up to admit any number of extra tokens.  Is that being "too flexible"?  Wouldn't we rather an error if more than a single extra token shows up so that we can correctly update the parser?

pyomo/contrib/solver/solvers/sol_reader.py
```diff
@@ -122,6 +122,12 @@ def get_duals(
                 'Solution loader does not currently have a valid solution. Please '
                 'check results.termination_condition and/or results.solution_status.'
             )
+        # If the NL instance has no objectives, report zeros
+        if len(self._nl_info.objectives) == 0:
```
jsiirola: Style: I would generally prefer (and it is measurably faster):

```suggestion
        if not self._nl_info.objectives:
```

pyomo/contrib/solver/solvers/sol_reader.py
```diff
@@ -122,6 +122,12 @@ def get_duals(
...
+            cons = (
+                cons_to_load if cons_to_load is not None else self._nl_info.constraints
+            )
+            return {c: 0.0 for c in cons}
```
jsiirola, line 130: As with RC, is it more correct to return {}?

blnicho (approved):
It might be worth adding some additional tests for the log parser with different Ipopt `print_level`. We might just want to detect that a non-default `print_level` was used and add logic to not try to parse it. This doesn't have to hold up this PR but I was thinking about it in the context of making the parser more robust.

pyomo/contrib/solver/solvers/sol_reader.py
```diff
@@ -133,12 +139,19 @@ def get_duals(
                 "have happened. Report this error to the Pyomo Developers."
             )
         res = {}
-        if self._nl_info.scaling is None:
-            scale_list = [1] * len(self._nl_info.constraints)
-            obj_scale = 1
+        # Set a default scaling
```
jsiirola: I have concerns about this implementation.  99% of the time, the user us going to load everything, and 95% of the non-IDAES users will never use scaling.  Unfortunately, this implementation incurs the overhead of the filter and the division regardless of what the user is doing.  I would suggest a different approach:
```python
scaling = self._nl_info.scaling
if scaling:
    _iter = zip(self._nl_info.constraints, self._sol_data.duals, scaling.constraints)
    obj_scale = self._nl_info.scaling.objectives[0]
else:
    _iter = zip(self._nl_info.constraints, self._sol_data.duals)
if cons_to_load is not None:
    _iter = filter(lambda x: x[0] in cons_to_load, _iter)
if scaling:
    res = {con: val * scale / obj_scale for con, val, scale in _iter}
else:
    res = {con: val for con, val in _iter}

File: pyomo/contrib/solver/solvers/ipopt.py (hunk 4 above)
Comment (mrmundt) [reply pyomo/contrib/solver/solvers/ipopt.py]: I do not know - I am trying to balance being "flexible" with column length but also not allowing us to process any lines that are weird. I can remove it and see what happens...

File: pyomo/contrib/solver/solvers/ipopt.py (hunk 7 above)
Comment (mrmundt) [reply pyomo/contrib/solver/solvers/ipopt.py]: Because `iter_data` includes things like `restoration`, `step_acceptance`, etc., and I don't want to try to cast those or do anything else with them. So I am only going over "the values I care about", e.g., the columns.

File: pyomo/contrib/solver/solvers/ipopt.py (hunk 8 above)
Comment (mrmundt) [reply pyomo/contrib/solver/solvers/ipopt.py]: I think the answer is, "Yes, it formats it upfront." But it shouldn't be a huge performance hit because f-strings are pretty efficient.

File: pyomo/contrib/solver/solvers/ipopt.py (hunk 10 above)
Comment (mrmundt) [reply pyomo/contrib/solver/solvers/ipopt.py:658]: Good point - I forgot that I already have a list of "allowed" tokens (`_ALPHA_PR_CHARS`). I'll compare against that.

Comment (dallan-keylogic) [reply pyomo/contrib/solver/solvers/ipopt.py:658]: If you want to raise an error to update the parser, the error message should actually say "Open an issue for Pyomo so that we can update the parser". The grad students who initially got the assertion error about the iterations not matching had no idea what was going on.

Comment (mrmundt) [reply pyomo/contrib/solver/solvers/ipopt.py:658]: Yup, I have that going locally right meow. I am trying to debug something else, but it'll be here soon.

File: pyomo/contrib/solver/solvers/ipopt.py (hunk 5 above)
Comment (mrmundt) [reply pyomo/contrib/solver/solvers/ipopt.py]: So I did this (both here and for `get_duals`), and it actually goes against what we say is a norm, surprisingly enough:
```
________________________________________ TestSolvers.test_no_objective_3_ipopt _________________________________________

a = (<pyomo.contrib.solver.tests.solvers.test_solvers.TestSolvers testMethod=test_no_objective_3_ipopt>,), kw = {}

    @wraps(func)
    def standalone_func(*a, **kw):
>       return func(*(a + p.args), **p.kwargs, **kw)

../venv-pyomo/lib/python3.12/site-packages/parameterized/parameterized.py:620: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pyomo.contrib.solver.tests.solvers.test_solvers.TestSolvers testMethod=test_no_objective_3_ipopt>
name = 'ipopt', opt_class = <class 'pyomo.contrib.solver.solvers.ipopt.Ipopt'>, use_presolve = False

    @parameterized.expand(input=_load_tests(all_solvers))
    def test_no_objective(
        self, name: str, opt_class: Type[SolverBase], use_presolve: bool
    ):
        opt: SolverBase = opt_class()
        if not opt.available():
            raise unittest.SkipTest(f'Solver {opt.name} not available.')
        check_duals = True
        if any(name.startswith(i) for i in nl_solvers_set):
            if use_presolve:
                check_duals = False
                opt.config.writer_config.linear_presolve = True
            else:
                opt.config.writer_config.linear_presolve = False
        m = pyo.ConcreteModel()
        m.x = pyo.Var()
        m.y = pyo.Var()
        m.a1 = pyo.Param(mutable=True)
        m.a2 = pyo.Param(mutable=True)
        m.b1 = pyo.Param(mutable=True)
        m.b2 = pyo.Param(mutable=True)
        m.c1 = pyo.Constraint(expr=m.y == m.a1 * m.x + m.b1)
        m.c2 = pyo.Constraint(expr=m.y == m.a2 * m.x + m.b2)
    
        params_to_test = [(1, -1, 2, 1), (1, -2, 2, 1), (1, -1, 3, 1)]
        for a1, a2, b1, b2 in params_to_test:
            m.a1.value = a1
            m.a2.value = a2
            m.b1.value = b1
            m.b2.value = b2
            res: Results = opt.solve(m)
            self.assertEqual(res.solution_status, SolutionStatus.optimal)
            self.assertAlmostEqual(m.x.value, (b2 - b1) / (a1 - a2))
            self.assertAlmostEqual(m.y.value, a1 * (b2 - b1) / (a1 - a2) + b1)
            self.assertEqual(res.incumbent_objective, None)
            self.assertEqual(res.objective_bound, None)
            if check_duals:
                duals = res.solution_loader.get_duals()
>               self.assertAlmostEqual(duals[m.c1], 0)
E               KeyError: <pyomo.core.base.constraint.ScalarConstraint object at 0x11b9effc0>

pyomo/contrib/solver/tests/solvers/test_solvers.py:994: KeyError
```

If I return an empty ComponentMap or an empty dict, then you can't actually access that item anymore.

File: pyomo/contrib/solver/solvers/ipopt.py (hunk 4 above)
Comment (mrmundt) [reply pyomo/contrib/solver/solvers/ipopt.py]: Aha, okay, I had a logic error that I didn't realize because of how IPOPT sometimes smushes the first and second column together. I fixed that and now this is unnecessary.

File: pyomo/contrib/solver/solvers/ipopt.py (hunk 1 above)
Comment (dallan-keylogic) [reply pyomo/contrib/solver/solvers/ipopt.py]: @mrmundt  IPOPT switches objectives when it enters into Restoration mode, using the L1 norm of the constraint violation as an objective. If it converges to an infeasible point with the exit message `EXIT: Converged to a point of local infeasibility.` then it is at a local minimum of the constraint violation, and we expect nonzero dual variables and reduced costs for this objective. I don't know if IPOPT makes these values available to AMPL, though.

Comment (mrmundt) [reply pyomo/contrib/solver/solvers/ipopt.py]: I am not sure what to do with this information because I only understand what half of it means.

Comment (dallan-keylogic) [reply pyomo/contrib/solver/solvers/ipopt.py]: If IPOPT converges to an infeasible point, the dual variables and reduced costs are nonzero and contain useful information. I'm not sure whether IPOPT makes that information available via the interface you're using.

File: pyomo/contrib/solver/solvers/ipopt.py (hunk 5 above)
Comment (mrmundt) [reply pyomo/contrib/solver/solvers/ipopt.py]: Okay, empty ComponentMap is fine; it's empty dict for `get_duals` that causes these failures.

**Comment by mrmundt** [comment]
@AlexGisi - you pointed out a different missed thing in the IPOPT parser, though! We did only have lowercase `s` in the accepted chars for `alpha_pr`. I added that in just now. Thanks for pointing it out!

File: pyomo/contrib/solver/solvers/ipopt.py [hunk 14]
```diff
@@ -505,17 +546,19 @@ def solve(self, model, **kwds) -> Results:
...
-                            ADVANCED_OPTION
-                        )
-                    except KeyError as e:
+                        iter_log = results.extra_info.get("iteration_log", None)
+                        if iter_log is not None:
+                            # Only set if object supports the attribute
+                            try:
+                                iter_log._visibility = ADVANCED_OPTION
+                            except Exception:
+                                pass
```
jsiirola: How is this exception ever triggered?

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -610,41 +653,75 @@ def _parse_ipopt_output(self, output: Union[str, io.StringIO]) -> Dict[str, Any]
...
             iterations = []
+            n_expected_columns = len(columns)
 
             for line in iter_table:
                 tokens = line.strip().split()
-                if len(tokens) != len(columns):
-                    continue
+                # IPOPT sometimes mashes the first two column values together
+                # (e.g., "2r-4.93e-03"). We need to split them.
+                if tokens and (('-' in tokens[0][1:]) or ('+' in tokens[0][1:])):
```
jsiirola: You shouldn't need to test `if tokens` here: the RE guarantees that the line is not empty.

Also: why do you need to skip the first character of the token?

Also also: why are you testing "+"?  Is there any evidence that ipopt will ever insert a "+" for positive numbers?

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -610,41 +653,75 @@ def _parse_ipopt_output(self, output: Union[str, io.StringIO]) -> Dict[str, Any]
...
+                    m = re.match(
+                        r'^(\d+r?)([+-](?:\d+(?:\.\d*)?|\.\d+)(?:[eE][+-]?\d+)?)$',
+                        tokens[0],
+                    )
+                    if m:
+                        tokens = [m.group(1), m.group(2)] + tokens[1:]
```
jsiirola: The following should work and be significantly simpler:
```suggestion
                if '-' in tokens[0]:
                    tokens[:1] = tokens[0].split('-', 1)
                elif '+' in tokens[0]:
                    tokens[:1] = tokens[0].split('+', 1)
```

pyomo/contrib/solver/solvers/ipopt.py
jsiirola: Then please make another list for `numeric_columns` and iterate over that, or better yet, only create the dict for numeric values, convert them, and then add the extra stuff at the end.

pyomo/contrib/solver/solvers/sol_reader.py
```diff
@@ -133,21 +139,22 @@ def get_duals(
                 "have happened. Report this error to the Pyomo Developers."
             )
         res = {}
-        if self._nl_info.scaling is None:
-            scale_list = [1] * len(self._nl_info.constraints)
-            obj_scale = 1
-        else:
-            scale_list = self._nl_info.scaling.constraints
+        # Set a default scaling
+        obj_scale = 1
```
jsiirola: This should not be necessary.
```suggestion
```

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -610,41 +653,75 @@ def _parse_ipopt_output(self, output: Union[str, io.StringIO]) -> Dict[str, Any]
...
+                restoration = iter_str.endswith("r")
+                if restoration:
+                    iter_str = iter_str[:-1]
+
+                try:
+                    iter_num = int(iter_str)
+                except ValueError:
+                    logger.error(
+                        "Could not parse IPOPT iteration number from line:\n\t%s", line
+                    )
```
jsiirola: You will still need to insert/define `iter_num`, otherwise you will either get an AttributeError or an AssertionError below.  Regardless, you will end up raising a fatal exception and should just do so here.
```suggestion
                    logger.error(
                        "Could not parse IPOPT iteration number from line:\n\t%s", line
                    )
```

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -610,41 +653,75 @@ def _parse_ipopt_output(self, output: Union[str, io.StringIO]) -> Dict[str, Any]
...
-                iter_data['restoration'] = iter_data['iter'].endswith('r')
-                if iter_data['restoration']:
-                    iter_data['iter'] = iter_data['iter'][:-1]
+                iter_str = iter_data.pop("iter")
+                restoration = iter_str.endswith("r")
+                if restoration:
+                    iter_str = iter_str[:-1]
+
+                try:
+                    iter_num = int(iter_str)
```
jsiirola: There really isn't a good reason to incur the overhead of additional local variables (_str and _num) here, is there?  Just using a single local variable (`iter` or `iter_num`) should be just fine.

pyomo/contrib/solver/solvers/ipopt.py
mrmundt: This will drop the -/+ rather than preserving it, so I'll push back on this specific suggestion.

jsiirola: Right. Then make it
```python
idx = tokens[0].index('-')
tokens[:1] = (tokens[0][:idx], tokens[0][idx:])
```

mrmundt: Follow-up pushback on that: You are pretty much guaranteed that, if the lines are mashed together, you'll get something that looks like this: `1r-4.3126674e+00`

So if that actually ends up being something like: `1r-4.3126674e-01`, there are two minus symbols. I genuinely think it's more robust to do the regex.

EDIT: Yes, I know that you'll get "the first instance" when you do `index`. I just worry about overall robustness.

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -609,43 +648,79 @@ def _parse_ipopt_output(self, output: Union[str, io.StringIO]) -> Dict[str, Any]
...
                     iter_data['step_acceptance'] = None
 
+                # Capture optional IPOPT diagnostic tags if present
+                if extra_tokens:
+                    if all(tok in _DIAGNOSTIC_TAGS for tok in extra_tokens):
+                        iter_data['diagnostic_tags'] = "".join(extra_tokens)
+                    else:
+                        logger.warning(
+                            f"Unrecognized Ipopt diagnostic tags {extra_tokens} on line: {line}"
+                        )
```
jsiirola: I did some diving into the Ipopt codebase.  The diagnostic tags are collected using a method `Append_info_string`.   Some things I found (recursive `grep` is your friend):
- there are a large number of tags that are not documented in the Ipopt [documentation](https://coin-or.github.io/Ipopt/OUTPUT.html)
- the tags are appended and may or may not be space-delimited
- some tags may be output with a value (e.g., `xi={float}`

Given all this, I believe that we should:
- remove the _DIAGNOSTIC_TAGS set
- combine all extra tokens into a single string and just blindly return it in `diagnostic_tags` (note joining them with spaces):

```suggestion
                # Capture optional IPOPT diagnostic tags if present
                if extra_tokens:
                    iter_data['diagnostic_tags'] = " ".join(extra_tokens)
```

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -465,6 +505,7 @@ def solve(self, model, **kwds) -> Results:
                 # This is the data we need to parse to get the iterations
                 # and time
                 parsed_output_data = self._parse_ipopt_output(ostreams[0])
+                parsed_output_message = f"Parsed solver data: {parsed_output_data}"
```
jsiirola: Can we move this to the exception handler?  There is no reason to create teh string *unless* we are about to raise the exception.

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -609,43 +648,79 @@ def _parse_ipopt_output(self, output: Union[str, io.StringIO]) -> Dict[str, Any]
...
+                            f"Unrecognized Ipopt diagnostic tags {extra_tokens} on line: {line}"
+                        )
+
                 # Attempt to cast all values to float where possible
-                for key in columns:
-                    if iter_data[key] == '-':
+                for key, val in iter_data.items():
+                    if key not in numerical_columns:
+                        continue
+                    if val == '-':
```
jsiirola: It seems that the following would be simpler, and avoid the need to create / lookup in a set:
```suggestion
                for key in columns[1:]:
                    val = iter_data[key]
                    if val == '-':
```

Also note that `ls` should be an `int` (this should always succeed, as floats are always castable to int):
```python
iter_data['ls'] = int(iter_data['ls'])
```

(If you agree, then we can delete `numerical_columns` above.)

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -610,42 +611,70 @@ def _parse_ipopt_output(self, output: Union[str, io.StringIO]) -> Dict[str, Any]
...
             for line in iter_table:
                 tokens = line.strip().split()
-                if len(tokens) != len(columns):
-                    continue
+                # IPOPT sometimes mashes the first two column values together
+                # (e.g., "2r-4.93e-03"). We need to split them.
+                try:
+                    idx = tokens[0].index('-')
+                    head = tokens[0][:idx]
+                    if head and head.rstrip('r').isdigit():
```
jsiirola, line 623: Why is this test needed?  The only way for a `-` to be in tokens[0] is if the first two columns get merged...

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -610,42 +611,70 @@ def _parse_ipopt_output(self, output: Union[str, io.StringIO]) -> Dict[str, Any]
...
             iterations = []
+            n_expected_columns = len(columns)
 
             for line in iter_table:
                 tokens = line.strip().split()
-                if len(tokens) != len(columns):
-                    continue
+                # IPOPT sometimes mashes the first two column values together
+                # (e.g., "2r-4.93e-03"). We need to split them.
+                try:
```
jsiirola, line 620: I'd still to prefer to not rely on the exception for the "normal" case where the columns are properly separated.  That is, look for the `-`, and if it was found, then split the tokens.

pyomo/contrib/solver/solvers/sol_reader.py
jsiirola: (I am happy to defer sorting out the duals / rc until the next PR (which will be a more substantive rework of the SOL parser)

mrmundt:
Jenkins failures were related to #3750 , unrelated to anything else. So this is technically good to go.

Issue #3688, opened 2025-08-08
[PEP] Next Steps for Solver Interface Redesign


mrmundt:
## Summary
This is actually just a checklist of the tasks that we decided need to be done before we are comfortable porting over other interfaces.

- [ ] Finish the split of `available` and `license_is_valid`
- [ ] Update `SolutionLoader` to add in loader functions, change naming, convert to `NotImplemented`, etc.
- [ ] Assign "Generation" property to solvers

### Notes from 8/8/2025
------------ 
IPOPT solver exit codes: Do we want to revisit what we support? 

https://github.com/Pyomo/pyomo/issues/868 

IPOPT in particular: whatever the solver gave us, we are going to return the raw solver message as opposed to our own strings. 

We think it is best to follow the guideline that, in the cases of solvers like IPOPT that have a lot of possible return values, we will check for/process the common ones (80/20 rule), and all others we will return raw/unaltered and allow users to inspect it. 

We will map the result to SolverStatus & TerminationCondition.  We will also return whatever the solver gives us in extra info (e.g., return code, solver message, etc).  I don't know if we also decided to make "solver message" (or whatever) a standardized results field 

Non-persistent options: is that fixed? I think it is but I'm not sure 

https://github.com/Pyomo/pyomo/issues/798 

Gurobi file writing: are we writing files? If so, would we still see this issue? 

https://github.com/Pyomo/pyomo/issues/1567 

Mipgap on Results object: should we put it? 

https://github.com/Pyomo/pyomo/issues/2167 

Ambiguity here. Are they just asking us to report what the solver reported, sure, if we want to promise that we parse it all the time. We could put it in the extra_info and let them figure it out. From a pedantic point of view, what do we even report back? There are multiple ways to do it. 

If the solver returns a relative gap at termination, we should report that back to the user, and be clear in documentation that that is what we are doing. 

2362: Can Michael look over this again and see what is still valid / if we need to make some changes in the new interfaces? 

https://github.com/Pyomo/pyomo/issues/2362 

Pyomo command with new interfaces: how do we want this to work? 

Related which is why I thought about it: https://github.com/Pyomo/pyomo/issues/2703 

 

#### Michaels Notes - Remaining Items - 8/7/25 

Split available and license_available/acquire_license/release_license? 

John: Would like `available` to be able to be called on a class and on an instance; remove `config` from this. 

For the notes.  I propose the following API: 

def available(self, recheck=False): 
   # Return the best-guess of the availability 
[...]

PR #3743, opened 2025-10-03
Change how `available` and `license` work in new solver interfaces

mrmundt:
## Summary/Motivation:
As part of the solver redesign effort, we decided to change how `available` and licensing in general will work in the new solver interface paradigm. This implements the most recent comment in #3688 for all existing solvers.

- Redefine what `available` basically means
- Add a `license` attribute to each solver (default is no operation)
- Custom license manager for gurobi

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/solver/common/base.py
+++ b/pyomo/contrib/solver/common/base.py
@@ -64,7 +64,7 @@ class Availability(IntEnum):
 
     LicenseError = -2
     """The solver was found but no usable license is available. This could
-    indicate either that no license was found, an expired or misformed
+    indicate either that no license was found, an expired or malformed
     license was found, or the license is incorrect for the problem type."""
 
     def __bool__(self):
--- a/pyomo/contrib/solver/common/base.py
+++ b/pyomo/contrib/solver/common/base.py
@@ -55,6 +55,7 @@ class Availability(IntEnum):
     LimitedLicense = 1
     """The solver was found and a limited license (e.g., demo license) is
     accessible to use."""
+
     NotFound = 0
     """The solver was not found, either because the executable was not
     on the path or the solver package is not importable."""
--- a/pyomo/contrib/solver/common/base.py
+++ b/pyomo/contrib/solver/common/base.py
@@ -9,8 +9,9 @@
 #  This software is distributed under the 3-clause BSD License.
 #  ___________________________________________________________________________
 
-from typing import Sequence, Dict, Optional, Mapping, List, Tuple
 import os
+from contextlib import contextmanager
+from typing import Sequence, Dict, Optional, Mapping, List, Tuple
 
 from pyomo.core.base.constraint import ConstraintData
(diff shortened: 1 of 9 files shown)


pyomo/contrib/solver/common/base.py
```diff
@@ -45,12 +46,26 @@ class Availability(IntEnum):
...
     FullLicense = 2
+    """The solver was found and a full license is accessible to use."""
+
     LimitedLicense = 1
+    """The solver was found and a limited license (e.g., demo license) is
+    accessible to use."""
     NotFound = 0
-    BadVersion = -1
-    BadLicense = -2
-    NeedsCompiledExtension = -3
```
michaelbynum, line 53: Why throw out `NeedsCompiledExtension`? I could definitely see us using this occasionally. I don't think there is a current use case. I guess we could easily add it back in if needed?

pyomo/contrib/solver/solvers/gurobi_direct.py
```diff
@@ -325,59 +423,64 @@ def solve(self, model, **kwds) -> Results:
...
-                timer.start('transfer_model')
-                x = gurobi_model.addMVar(
-                    len(repn.columns),
-                    lb=lb,
-                    ub=ub,
-                    obj=repn.c.todense()[0] if repn.c.shape[0] else 0,
-                    vtype=vtype,
+
+            # Acquire a Gurobi env for the duration of solve (opt + postsolve):
+            with self.license():
```
michaelbynum, line 436: Will this release the license even if the user called `acquire` on their own?

michaelbynum (commented):
I have not done a detailed review, but I had a couple questions after skimming this.

pyomo/contrib/solver/common/base.py
mrmundt: We aren't using it currently so I took it out. We have no issues adding it back in if it actually becomes a valid case again.

pyomo/contrib/solver/solvers/gurobi_direct.py
mrmundt: I am reading this in two ways and have been stuck in analysis-paralysis on how to answer. So instead, I am going to walk through the logic of it:

1. User calls solver.license.acquire() -> client_count = 1
2. User calls solve() -> acquire again -> client_count = 2
3. Exiting the with block -> release once -> client_count = 1 (still held by the user)
4. User later calls solver.license.release() -> client_count = 0 -> env closed

That's how I intended it. A new `env` isn't created, just the `cls._num_gurobipy_env_clients` is incremented (assuming that an `env` already exists at this point).

michaelbynum: Perfect. Thanks!

mrmundt:
Closing this for a bit until I can figure out the multiprocessing stuff

PR #3745, opened 2025-10-06
Add Gurobi Direct MINLP solver in contrib.solver

### LINKED ISSUE #3522: Support Gurobi nonlinear expressions in the direct/persistent solvers
## Summary

Gurobi v12 supports nonlinear constraints, it would be good if the direct/persistent solvers supported adding these directly to the `gurobipy.Model` object.

### Rationale

Right now, to use this feature through Pyomo, one must use the `gurobinl` solver which uses the AMPL interface via an NL file.

### Description

I would like to leverage the Pyomo expressions to either create an expression tree to pass this to `gurobipy` via [`Model.addGenConstrNLAdv`](https://docs.gurobi.com/projects/optimizer/en/current/reference/python/model.html#Model.addGenConstrNLAdv) or even create a [`NLExpr`](https://docs.gurobi.com/projects/optimizer/en/current/reference/python/nlexpr.html#gurobipy-nlexpr) whatever is easiest. This requires a bit of knowledge of the Pyomo expressions and I haven't had time to dig into it yet.

### Additional information

This can be used to translate nonlinear functions such as `pyomo.kernel.sin` as well as polynomials, hence it would solve #3249

emma58:
## Summary/Motivation:

This adds a new solver interface `gurobi_direct_minlp` that supports sending MINLP problems to Gurobi via the Python API. currently this is implemented as a separate Gurobi direct interface in `contrib.solver` so that we don't have negative performance implications for `gurobi_direct_v2` (and to punt for the moment on integrating the matrix-vector inferface with the general MINLP one).

- Adds a walker that makes the (required, in Gurobi 12) conversion from general nonlinear constraints to a mixture of `aux_var == general_nonlinear_expr` equalities and linear/quadratic constraints involving the auxiliary and other variables.
- Implements a pretty minimalistic writer that uses the above walker: The special thing here is that some Pyomo constraints will create more than one constraint in the gurobi model because of the transformation mentioned in the bullet point above.
- Implements a new solver interface that inherits from `GurobiDirect` in `contrib.solver`: The only thing special here is the `solve` function which is again just unique because of the custom walker/writer explained above.
- Gets the test coverage to 89%, which *rounds* to 90%... The only thing interesting here is that I added a method for converting general nonlinear Gurobi expression trees back to Pyomo expressions so that we can test them that way.

Outstanding TODOs, none of which should prevent merging unless we want them to:
- [ ]  Move all the tests that manually test Gurobi's expression tree onto the `gurobi_nl_to_pyo_expr` function, which will make them a lot more readable, and easier to maintain if every there is a change to the public API in gurobipy.
- [ ] Add some tests for the sake of coverage--we're missing some cases

### Legal Acknowledgement


pyomo/contrib/solver/solvers/gurobi_direct_minlp.py
```diff
@@ -0,0 +1,677 @@
...
+        return self.exit_node_dispatcher[(node.__class__, *map(itemgetter(0), data))](
+            self, node, *data
+        )
+
+    def finalizeResult(self, result):
+        return result
+
+    # ESJ TODO: THIS IS COPIED FROM THE LINEAR WALKER--CAN WE PUT IT IN UTIL OR
+    # SOMETHING?
+    def check_constant(self, ans, obj):
```
emma58: @jsiirola I copied this over from the linear walker, but can we somehow centralize it?

pyomo/contrib/solver/solvers/gurobi_direct_minlp.py
```diff
@@ -0,0 +1,677 @@
...
+
+from pyomo.common.dependencies import attempt_import
+from pyomo.common.collections import ComponentMap, ComponentSet
+from pyomo.common.config import ConfigDict, ConfigValue
+from pyomo.common.errors import InvalidValueError
+from pyomo.common.numeric_types import native_complex_types
+from pyomo.common.timing import HierarchicalTimer
+
+# ESJ TODO: Could we move this to util or somewhere less bizarre?
+from pyomo.contrib.cp.repn.docplex_writer import collect_valid_components
```
emma58: @jsiirola this is my other oddity: I kind of like this function, and we hid it in the docplex writer. Where *should* we put it?

emma58: Ans: `repn.util`, if it or something similar isn't already there.

pyomo/contrib/solver/solvers/gurobi_direct_minlp.py
emma58: Answer: Promote to `pyomo.core.expr`. And does it really need the visitor?

pyomo/contrib/solver/solvers/gurobi_direct_minlp.py
emma58: Ah, the answer was `categorize_valid_components`.

pyomo/contrib/solver/solvers/gurobi_direct_minlp.py
emma58: It needs `InvalidNumber` so trying to put it in `core.expr` created a circular import situation. So I put it in `repn.util` for now... And it doesn't really need the visitor, but I think we maybe should defer that change because it takes it to pass to `complex_number_error` which only uses it to get the name of the visitor class... But it's public, so if we really want to be backwards compatible that's not a simple change.

emma58:
@jsiirola @mrmundt If the tests pass, I really am done messing with this now, I think!

pyomo/contrib/solver/solvers/gurobi_direct.py
```diff
@@ -117,7 +117,7 @@ def load_vars(self, vars_to_load=None, solution_number=0):
         if self._grb_model.SolCount == 0:
             raise NoSolutionError()
 
-        iterator = zip(self._pyo_vars, self._grb_vars.x.tolist())
+        iterator = zip(self._pyo_vars, map(operator.attrgetter('x'), self._grb_vars))
```
jsiirola, line 120: Is fetching the values one at a time slower than getting them in a single vector?  (Same question applies to get_primals, get_duals. and get_redued_costs below)

pyomo/contrib/solver/solvers/gurobi_direct.py
```diff
@@ -374,7 +376,12 @@ def solve(self, model, **kwds) -> Results:
             timer,
             config,
             GurobiDirectSolutionLoader(
-                gurobi_model, A, x, repn.rows, repn.columns, repn.objectives
+                gurobi_model,
+                A,
+                x.tolist(),
```
jsiirola, line 381: Why convert x to a list?

pyomo/contrib/solver/solvers/gurobi_direct.py
```diff
@@ -374,7 +376,12 @@ def solve(self, model, **kwds) -> Results:
...
+                list(map(operator.itemgetter(0), repn.rows)),
```
jsiirola, line 382: Why de-tupalize rows?  aren't both values needed by the loader?

pyomo/core/expr/numeric_expr.py
```diff
@@ -1206,7 +1206,11 @@ def _trunc_extend(self, other):
         return self.__class__(_args)
 
     def _apply_operation(self, result):
-        return sum(result)
+        # Avoid 0 being added to summations by specifying the start
+        if result:
+            return sum(result[1:], start=result[0])
```
jsiirola: We should avoid duplicating `result`:

```suggestion
            _it = iter(result)
            return sum(_iter, start=next(iter))
```

Justification:
```python
# (happened to test with Python 3.11.14)
>>> import timeit
>>> a = list(range(1, 1000))
>>> a[0]
1
>>> sum(a)
499500
>>> sum(a[1:], start=a[0])
499500
>>> i = iter(a)
>>> sum(i, start=next(i))
499500
>>> timeit.timeit('sum(a[1:], start=a[0])', 'from __main__ import a')
6.222495133988559
>>> timeit.timeit('i=iter(a); sum(i, start=next(i))', 'from __main__ import a')
4.841764358105138
```

pyomo/contrib/solver/solvers/gurobi_direct_minlp.py
```diff
@@ -0,0 +1,655 @@
...
+
+
+def _create_grb_var(visitor, pyomo_var, name=""):
+    pyo_domain = pyomo_var.domain
+    if pyo_domain in _domain_map:
+        domain, domain_lb, domain_ub = _domain_map[pyo_domain]
+    else:
+        raise ValueError(
+            "Unsupported domain for Var '%s': %s" % (pyomo_var.name, pyo_domain)
+        )
```
jsiirola, line 165: It would be nice to handle general domains:

```suggestion
        domain_lb, domain_ub, domain = pyo_domain.get_interval()
        if domain not in (0, 1)
            raise ValueError(
                "Unsupported domain for Var '%s': %s" % (pyomo_var.name, pyo_domain)
            )
        domain = GRB.INTEGER if domain else GRB.CONTINUOUS
        if domain_lb is None:
            domain_lb = -float('inf')
        if domain_ub is None:
            domain_ub = float('inf')
```

pyomo/contrib/solver/solvers/gurobi_direct_minlp.py
```diff
@@ -0,0 +1,655 @@
...
+            # 'tanh': None,
+            # 'atanh': None,
+            # 'ceil': None,
+            # 'floor': None,
+        }
+    )
+
+    _domain_map.update(
+        (
+            (Binary, (GRB.BINARY, -float('inf'), float('inf'))),
```
jsiirola: Why is this not

```suggestion
            (Binary, (GRB.BINARY, 0, 1)),
```

pyomo/contrib/solver/solvers/gurobi_direct_minlp.py
```diff
@@ -0,0 +1,655 @@
...
+def _handle_node_with_eval_expr_visitor_quadratic(visitor, node, *data):
+    return (
+        _QUADRATIC,
+        visitor._eval_expr_visitor.visit(node, tuple(map(itemgetter(1), data))),
+    )
+
+
+def _handle_node_with_eval_expr_visitor_nonlinear(visitor, node, *data):
+    # ESJ: _apply_operation for DivisionExpression expects that result is indexed, so
+    # I'm making it a tuple rather than a map.
```
jsiirola:
```suggestion
    # ESJ: _apply_operation for DivisionExpression expects that result supports __getitem__,
    # so, I'm expanding the map to a tuple.
```

pyomo/contrib/solver/solvers/gurobi_direct_minlp.py
```diff
@@ -0,0 +1,655 @@
...
+def _create_grb_var(visitor, pyomo_var, name=""):
+    pyo_domain = pyomo_var.domain
+    if pyo_domain in _domain_map:
+        domain, domain_lb, domain_ub = _domain_map[pyo_domain]
+    else:
+        raise ValueError(
+            "Unsupported domain for Var '%s': %s" % (pyomo_var.name, pyo_domain)
+        )
+    lb = max(domain_lb, pyomo_var.lb) if pyomo_var.lb is not None else domain_lb
+    ub = min(domain_ub, pyomo_var.ub) if pyomo_var.ub is not None else domain_ub
```
jsiirola: I think the following would be more efficient (and does the same thing):
```suggestion
    lb, ub = pyomo_var.bounds
    if lb is None:
        lb = -float("inf")
    if ub is None:
        ub = -float("inf")
```

pyomo/contrib/solver/solvers/gurobi_direct_minlp.py
```diff
@@ -0,0 +1,655 @@
...
+    return visitor.grb_model.addVar(lb=lb, ub=ub, vtype=domain, name=name)
```
jsiirola, line 178: Additional note: we should really consider providing an API that returns the bounds as a float (i.e., doesn't return `None`)

pyomo/contrib/solver/solvers/gurobi_direct_minlp.py
```diff
@@ -0,0 +1,655 @@
...
+                        expr = float(expr)
+                        if lb is not None:
+                            lb = float(lb)
+                        if ub is not None:
+                            ub = float(ub)
+                    if cons.equality:
+                        grb_cons.append(grb_model.addConstr(expr == lb))
+                        pyo_cons.append(cons)
+                    else:
+                        # TODO: should be have special handling if expr is a
```
jsiirola:
```suggestion
                        # TODO: should we have special handling if expr is a
```

pyomo/contrib/solver/solvers/gurobi_direct.py
emma58: I don't know? If it is, this is much harder to generalize because the Gurobi MINLP version doesn't have the gurobi vars as a vector--they're already a list.

pyomo/contrib/solver/solvers/gurobi_direct.py
emma58: Because of the thing you commented on above, actually. I'm making this look the same as Gurobi MINLP so that they can use an identical SolutionLoader. Gurobi MINLP doesn't have the luxury of declaring the variables as a vector (I don't think? Although if I can dynamically add to one, that's another option, maybe?), so they're a list at this point right now.

pyomo/contrib/solver/solvers/gurobi_direct.py
emma58: I don't think so--the only time we need anything from this at all is to get the duals. And then you have to get the `Pi` attribute off each constraint. The difference is that was vectorized before too...

pyomo/contrib/solver/solvers/gurobi_direct_minlp.py
emma58: That would be nice, actually! We've written that if tree a lot, I'm sure...

emma58: I opened #3757

pyomo/contrib/solver/tests/solvers/gurobi_to_pyomo_expressions.py
```diff
@@ -0,0 +1,78 @@
+from pyomo.common.dependencies import attempt_import
```
mrmundt, line 12: Needs copyright.

pyomo/contrib/solver/tests/solvers/test_gurobi_minlp.py
```diff
@@ -0,0 +1,226 @@
+import math
```
mrmundt, line 12: Needs copyright.

pyomo/contrib/solver/tests/solvers/test_gurobi_minlp_walker.py
```diff
@@ -0,0 +1,583 @@
+#  ___________________________________________________________________________
+#
+#  Pyomo: Python Optimization Modeling Objects
+#  Copyright (c) 2008-2024
```
mrmundt: Wrong copyright year.

pyomo/contrib/solver/solvers/gurobi_direct_minlp.py
```diff
@@ -0,0 +1,663 @@
+#  ___________________________________________________________________________
+#
+#  Pyomo: Python Optimization Modeling Objects
+#  Copyright (c) 2008-2024
```
mrmundt: Wrong copyright year.

pyomo/contrib/solver/tests/solvers/test_gurobi_minlp_writer.py
```diff
@@ -0,0 +1,570 @@
+#  ___________________________________________________________________________
+#
+#  Pyomo: Python Optimization Modeling Objects
+#  Copyright (c) 2008-2024
```
mrmundt: Wrong copyright year.

mrmundt:
I am not pushing the copyright changes yet - waiting for tests to finish since all the copyright stuff is NFC.

mrmundt:
But I DO have the changes ready to push as soon as the tests finish.

pyomo/contrib/solver/tests/solvers/test_gurobi_minlp_walker.py
```diff
@@ -0,0 +1,583 @@
...
+    def test_write_subtraction(self):
+        m = self.get_model()
+        m.c = Constraint(expr=m.x1 - m.x2 >= 3)
+        visitor = self.get_visitor()
+        _, expr = visitor.walk_expression(m.c.body)
+
+        x1 = visitor.var_map[id(m.x1)]
+        x2 = visitor.var_map[id(m.x2)]
+
+        # Also linear, whoot!
```
mrmundt, line 192: Not changing but this is an abomination of a word. Woot? Yes. Hoot? Yes. Whoo? Yes. But WHOOT???

jsiirola (approved):

PR #3746, opened 2025-10-07
Fix boolean conversion in DeferredImportIndicator class

eminyouskn:
## Summary/Motivation:

Some linters, such as Pyright, raise warnings when an object of class `DeferredImportIndicator` is used in an if statement, since its `__bool__` method may return `None`.
This PR explicitly casts the result of that method to `bool` to ensure a proper boolean value is returned.

- Cast the result of `__bool__` method of `DeferredImportIndicator` to be a `bool`

### Legal Acknowledgement

### DIFF
--- a/pyomo/common/dependencies.py
+++ b/pyomo/common/dependencies.py
@@ -346,7 +346,7 @@ class DeferredImportIndicator(_DeferredImportIndicatorBase):
 
     def __bool__(self):
         self.resolve()
-        return self._available
+        return bool(self._available)
 
     def resolve(self):
         # Only attempt the import once, then cache some form of result


jsiirola:
This is a nonfunctional change in that `_available` will always be `bool` after calling `resolve()`.  I am curious:  If you implement this change instead, does the linter recognize that fact?
```diff
diff --git a/pyomo/common/dependencies.py b/pyomo/common/dependencies.py
index 27167f12f..017b76c2c 100644
--- a/pyomo/common/dependencies.py
+++ b/pyomo/common/dependencies.py
@@ -344,13 +344,13 @@ class DeferredImportIndicator(_DeferredImportIndicatorBase):
         if callback is not None:
             DeferredImportCallbackFinder._callbacks.setdefault(name, []).append(self)

-    def __bool__(self):
+    def __bool__(self) -> bool:
         self.resolve()
         return self._available

     def resolve(self):
         # Only attempt the import once, then cache some form of result
-        if self._module is None:
+        if self._available is None:
             package = self._original_globals.get('__name__', '')
             try:
                 self._module, self._available = _perform_import(
(
```

yemine0x01:
Even with the proposed changes it does not managed to infer that `_available` has been edited to be a `bool` in the `resolve` function. If you don't like the `bool` casting, how about create another flag `_import_attempted` and from the begining put `_available` to `False` et not to `None`? I assume that the `None` status of `_available` is only to know if the import has been attempted or not.

pyomo/common/dependencies.py
```diff
@@ -346,7 +346,7 @@ def __init__(
 
     def __bool__(self):
         self.resolve()
-        return self._available
+        return bool(self._available)
```
jsiirola, line 352:
```suggestion
        # resolve() guarantees that _available has been resolved to a bool
        assert self._available.__class__ is bool
        # The following cast is to keep static code analysis linters happy
        return bool(self._available)
```

Issue #3753, opened 2025-10-15
highs iteration count always takes simplex iteration count


AlexGisi:
## Summary

In `contrib.solver.solvers.highs.py`, the `_postsolve` method records the simplex iteration count:
```
results.iteration_count = info.simplex_iteration_count
```

However, the highs solver might use the simplex, ipm, pdlp, mip, or qp solvers [depending on the problem](https://ergo-code.github.io/HiGHS/dev/solvers/#Primal-dual-hybrid-gradient-method). Each method has its [own iteration count](https://ergo-code.github.io/HiGHS/dev/structures/structs/HighsInfo/#simplex_iteration_count). The unused methods are populated with either -1 or 0 (I can't find rhyme or reason to which and when). Also, if the problem is infeasible, the entire highs info might be invalid.

### Steps to reproduce the issue

The latter case can be demonstrated with
```
from pyomo.environ import (
    ConcreteModel, Var, Reals, Objective, SolverFactory,
)


def main():
    m = ConcreteModel()
    m.x1 = Var(within=Reals, bounds=(0, 2), initialize=1.745)
    m.x4 = Var(within=Reals, bounds=(0, 5), initialize=3.048)
    m.x7 = Var(within=Reals, bounds=(0.9, 0.95), initialize=0.928)
    m.obj = Objective(expr=-6.3 * m.x4 * m.x7 + 5.04 * m.x1)

    opt = SolverFactory("highs")
    opt.solve(m, tee=False)


if __name__ == "__main__":
    main()

```

In general, we can check the iteration values for all types of problems:
```
from pyomo.environ import (
    ConcreteModel, Var, Objective, Constraint,
    NonNegativeReals, Binary, Reals, minimize, maximize,
    SolverFactory, value
)


def solve_lp(method: str):
    """Tiny LP, solved with a specific HiGHS LP method (simplex/ipm/pdlp)."""
    m = ConcreteModel()
    m.x = Var(domain=NonNegativeReals)
    m.y = Var(domain=NonNegativeReals)
    # min x + y  s.t.  x + y >= 1
    m.c = Constraint(expr=m.x + m.y >= 1)
    m.obj = Objective(expr=m.x + m.y, sense=minimize)

    opt = SolverFactory("highs")
    opt.options["solver"] = method  # 'simplex' | 'ipm' | 'pdlp'
    res = opt.solve(m, tee=False)
[...]

PR #3754, opened 2025-10-15
Get correct highs iteration count depending on method

AlexGisi:
## Fixes and Summary/Motivation # .
In the highs solver interface, instead of always reporting the simplex iteration count, we should get the iteration count of the method that was used. See #3753.

- Take the max of the available iteration counts, if the info is valid

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/solver/solvers/highs.py
+++ b/pyomo/contrib/solver/solvers/highs.py
@@ -769,10 +769,9 @@ class Highs(PersistentSolverMixin, PersistentSolverUtils, PersistentSolverBase):
                     results.iteration_count = 0
                 else:
                     assert (
-                        len(positive_iters) == 1,
+                        len(positive_iters) == 1
                     ), "Only one iteration count should have a positive value"
                     results.iteration_count = positive_iters[0]
-                results.iteration_count = max(counts)
             else:
                 results.iteration_count = 0
 
--- a/pyomo/contrib/solver/tests/solvers/test_highs.py
+++ b/pyomo/contrib/solver/tests/solvers/test_highs.py
@@ -126,9 +126,10 @@ class TestHighsMiniDemos(unittest.TestCase):
             solver.config.solver_options["solver"] = (
                 method  # 'simplex' | 'ipm' | 'pdlp'
             )
+            solver.config.solver_options["presolve"] = "off"
             results = solver.solve(m)
             self.assertTrue(results.solution_status == SolutionStatus.optimal)
-            self.assertTrue(results.iteration_count >= 0)
+            self.assertTrue(results.iteration_count > 0)
 
     def test_mip_demo(self):
         # Build MIP
@@ -139,9 +140,21 @@ class TestHighsMiniDemos(unittest.TestCase):
         m.obj = pyo.Objective(expr=3 * m.a + 2 * m.b, sense=pyo.maximize)
 
         solver = Highs()
+        solver.config.solver_options["presolve"] = "off"
         results = solver.solve(m)
         self.assertTrue(results.solution_status == SolutionStatus.optimal)
-        self.assertTrue(results.iteration_count >= 0)
+        self.assertTrue(results.iteration_count > 0)
+
+    def test_mip_pmedian(self):
+        # Build MIP
+        from pyomo.core.tests.examples.pmedian_concrete import create_model
+
+        M = create_model()
+
+        solver = Highs()
+        results = solver.solve(M)
+        self.assertEqual(results.solution_status, SolutionStatus.optimal)
+        self.assertTrue(results.iteration_count > 0)
 
     def test_qp_demo(self):
         # Build convex QP
@@ -152,6 +165,7 @@ class TestHighsMiniDemos(unittest.TestCase):
         m.obj = pyo.Objective(expr=(m.x - 1) ** 2 + (m.y - 2) ** 2, sense=pyo.minimize)
 
         solver = Highs()
+        solver.config.solver_options["presolve"] = "off"
         results = solver.solve(m)
         self.assertTrue(results.solution_status == SolutionStatus.optimal)
-        self.assertTrue(results.iteration_count >= 0)
+        self.assertTrue(results.iteration_count > 0)


mrmundt (approved):

pyomo/contrib/solver/solvers/highs.py
```diff
@@ -750,7 +750,20 @@ def _postsolve(self, stream: io.StringIO):
...
+                # The method that ran will have a non-negative iteration count
+                # and the others will be 0 or -1.
+                max_iters = max(
+                    info.simplex_iteration_count,
+                    info.ipm_iteration_count,
+                    info.mip_node_count,
+                    info.pdlp_iteration_count,
+                    info.qp_iteration_count,
+                )
+                results.iteration_count = max_iters if max_iters != -1 else 0
```
jsiirola: I feel like there should be an assertion that verifies <= 1 of these values is > 0.

jsiirola:
Can you add a test that exercises one (or a couple) different solvers?

AlexGisi:
> Can you add a test that exercises one (or a couple) different solvers?

@jsiirola Is this what you were thinking: https://github.com/Pyomo/pyomo/pull/3754/commits/ea5ba4f37daeaf4a2849869fa41d0c617a5f32f8

I added internal assertions so those will catch any oddities

pyomo/contrib/solver/tests/solvers/test_highs.py
```diff
@@ -109,3 +109,43 @@ def test_fix_and_unfix(self):
...
+            m.y = pyo.Var(domain=pyo.NonNegativeReals)
+            m.c = pyo.Constraint(expr=m.x + m.y >= 1)
+            m.obj = pyo.Objective(expr=m.x + m.y, sense=pyo.minimize)
+
+            solver = Highs()
+            solver.config.solver_options["solver"] = (
+                method  # 'simplex' | 'ipm' | 'pdlp'
+            )
+
+            solver.solve(m, tee=False)
```
mrmundt: What about adding some asserts to the tests? For example:
```suggestion
            results = solver.solve(m, tee=False)
            self.assertTrue(results.iteration_count > 0)
```

mrmundt: Also `tee` is False by default, so you don't need to set that.

pyomo/contrib/solver/solvers/highs.py
```diff
@@ -750,7 +750,26 @@ def _postsolve(self, stream: io.StringIO):
...
+                    info.mip_node_count,
+                    info.pdlp_iteration_count,
+                    info.qp_iteration_count,
+                ]
+                assert (
+                    len([c for c in counts if c > 1]) <= 1
+                ), "Only one method should have a positive iteration count"
+                assert any(
+                    [c >= 0 for c in counts]
+                ), "Should have a non-negative count if info valid"
```
jsiirola: What you have works, but the following is a little more selective in terms of the number of intermediate lists it creates and times it iterates over the counts.

```suggestion
                positive_iters = [c for c in counts if c > 0]
                if not positive_iters:
                    assert any(
                        (c == 0 for c in counts)
                    ), "At least one iteration count should have a non-negative value"
                    results.iteration_count = 0
                else:
                    assert (
                        len(positive_iters) == 1,
                    ), "Only one iteration count should have a positive value"
                    results.iteration_count = positive_iters[0]
```

pyomo/contrib/solver/tests/solvers/test_highs.py
```diff
@@ -109,3 +110,48 @@ def test_fix_and_unfix(self):
...
+            m.c = pyo.Constraint(expr=m.x + m.y >= 1)
+            m.obj = pyo.Objective(expr=m.x + m.y, sense=pyo.minimize)
+
+            solver = Highs()
+            solver.config.solver_options["solver"] = (
+                method  # 'simplex' | 'ipm' | 'pdlp'
+            )
+            results = solver.solve(m)
+            self.assertTrue(results.solution_status == SolutionStatus.optimal)
+            self.assertTrue(results.iteration_count >= 0)
```
jsiirola: Shouldn't the iteration count be strictly greater than 0 in this case?

pyomo/contrib/solver/tests/solvers/test_highs.py
```diff
@@ -109,3 +110,48 @@ def test_fix_and_unfix(self):
...
+        m = pyo.ConcreteModel()
+        m.a = pyo.Var(domain=pyo.Binary)
+        m.b = pyo.Var(domain=pyo.Binary)
+        m.cap = pyo.Constraint(expr=m.a + m.b <= 1)
+        m.obj = pyo.Objective(expr=3 * m.a + 2 * m.b, sense=pyo.maximize)
+
+        solver = Highs()
+        results = solver.solve(m)
+        self.assertTrue(results.solution_status == SolutionStatus.optimal)
+        self.assertTrue(results.iteration_count >= 0)
```
jsiirola: Shouldn't the iteration count be strictly greater than 0 in this case?

pyomo/contrib/solver/tests/solvers/test_highs.py
```diff
@@ -109,3 +110,48 @@ def test_fix_and_unfix(self):
...
+        m = pyo.ConcreteModel()
+        m.x = pyo.Var(domain=pyo.Reals, bounds=(0, None))
+        m.y = pyo.Var(domain=pyo.Reals, bounds=(0, None))
+        m.c = pyo.Constraint(expr=m.x + m.y >= 1)
+        m.obj = pyo.Objective(expr=(m.x - 1) ** 2 + (m.y - 2) ** 2, sense=pyo.minimize)
+
+        solver = Highs()
+        results = solver.solve(m)
+        self.assertTrue(results.solution_status == SolutionStatus.optimal)
+        self.assertTrue(results.iteration_count >= 0)
```
jsiirola: Shouldn't the iteration count be strictly greater than 0 in this case?

mrmundt:
Okay, I was playing with the tests and changed the MIP demo to this:
```python
    def test_mip_demo(self):
        # Build MIP
        from pyomo.core.tests.examples.pmedian_concrete import create_model

        M = create_model()

        solver = Highs()
        results = solver.solve(M)
        self.assertEqual(results.solution_status, SolutionStatus.optimal)
        self.assertTrue(results.iteration_count > 0)
```

Lo-and-behold, this causes an assertion error:
```
                positive_iters = [c for c in counts if c > 0]
                if not positive_iters:
                    assert any(
                        (c == 0 for c in counts)
                    ), "At least one iteration count should have a non-negative value"
                    results.iteration_count = 0
                else:
                    assert (
>                       len(positive_iters) == 1
                    ), "Only one iteration count should have a positive value"
E                   AssertionError: Only one iteration count should have a positive value

pyomo/contrib/solver/solvers/highs.py:773: AssertionError
```

`simplex_iteration_count` and `mip_node_count` are both positive values (in this case, 5 and 1). So we need to reconsider this PR / how we actually want to fix the problem since iteration_count in general is not well defined for MIP.

pyomo/contrib/solver/tests/solvers/test_highs.py
AlexGisi: Yes

pyomo/contrib/solver/tests/solvers/test_highs.py
AlexGisi: Yes

AlexGisi:
I suggest the problem raised by @mrmundt should be addressed by removing mip_node_count from the list of considered counts. It's a count of nodes, not solver iterations in the sense of the other methods. If we want to report the MIP node count, perhaps there would be a separate field in the results class. Since during MIP solve the other counts are populated corresponding to the method used within nodes (by default simplex) the count selection logic holds.

jsiirola:
This PR opened up a whole side discussion.  Talking with @bknueven (who happened to be sitting next to me), I think we came to the conclusion that we should argue for removing `iteration_count` from the "standard" set of results.  Instead, iteration count(s) should be put in the solver-specific results.  For MIP solvers, there really *isn't* a meaningful "iteration count" (@bknueven actually argued that "iteration count" may only be meaningful for interior point solvers ;) ), so trying to shoe-horn the concept into some kind of standard result would be misleading at best.

mrmundt:
@AlexGisi - We talked about this in the dev call today, and it's going to require a mini-rework of all the different solvers to address iteration counts. I'm going to close this PR but will incorporate the changes from it into the new PR that we open to address the "bigger problem."

PR #3756, opened 2025-10-16
Update to pypy3.11 and ignore pint

mrmundt:
## Summary/Motivation:
A new version of `pillow` was released, and it broke pypy-3.10. Update but also ignore pint.

-
-

### Legal Acknowledgement

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -240,7 +240,8 @@ setup_kwargs = dict(
             'openpyxl',  # dataportals
             'packaging',  # for checking other dependency versions
             #'pathos',   # requested for #963, but PR currently closed
-            'pint',  # units
+            # pint causes a segfault on pypy
+            'pint; implementation_name!="pypy"',  # units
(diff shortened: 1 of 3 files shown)


jsiirola (approved):

setup.py
```diff
@@ -240,7 +240,8 @@ def __ne__(self, other):
             'openpyxl',  # dataportals
             'packaging',  # for checking other dependency versions
             #'pathos',   # requested for #963, but PR currently closed
-            'pint',  # units
+            # pint causes a segfault on pypy
+            'pint; implementation_name!="pypy"',  # units
```
jsiirola, line 244: Can you move this down with the other "not on pypy" dependencies below?

mrmundt: I originally did that but those are clumped together due to "they are hard to install on pypy", which isn't true here. This causes a different problem. So... which do we want to do here?

PR #3766, opened 2025-10-21
Fix KNITRO solution status mapping

eminyouskn:
## Fixes # 

The initial implementation of KNITRO does not properly handle the mapping of solution statuses in cases where a time or iteration limit is reached, even when a feasible solution is available.

- Update the solution status when the iteration limit or time limit is reached on KNITRO.

### Legal Acknowledgement


mrmundt:
It might be worth adding some tests to ensure that the different statuses are caught!

yemine0x01:
@mrmundt that's a good idea. I will add them now then.

yemine0x01:
This PR is also ready.

yemine0x01:
@blnicho just to let you know, I added a link to the doc.

PR #3767, opened 2025-10-21
Remove `iteration_count` as a default Results object value

### LINKED ISSUE #3753: highs iteration count always takes simplex iteration count

mrmundt:
## Summary/Motivation:
As part of #3753 and #3754, we decided that `iteration_count` should be removed as a default piece of information on the Results object (particularly because it's not well defined for certain solvers, e.g., MIP solvers). This instead downgrades iteration count to being an `extra_info` declaration, if desired, and is specific to each solver.

- Remove `iteration_count` from defaults `Results` object
- Adjust each individual solver to move `iteration_count` into the `extra_info` ConfigDict
- Report ALL iteration counts from highs rather than whatever subset, with appropriate key-value pairs

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/solver/solvers/gurobi_direct.py
+++ b/pyomo/contrib/solver/solvers/gurobi_direct.py
@@ -442,7 +442,7 @@ class GurobiDirect(GurobiSolverMixin, SolverBase):
             results.incumbent_objective = None
             results.objective_bound = None
 
-        results.iteration_count = grb_model.getAttr('IterCount')
+        results.extra_info.iteration_count = grb_model.getAttr('IterCount')
 
         timer.start('load solution')
         if config.load_solutions:
--- a/pyomo/contrib/solver/solvers/gurobi_direct.py
+++ b/pyomo/contrib/solver/solvers/gurobi_direct.py
@@ -14,6 +14,7 @@ import io
 import math
 import operator
 import os
+import logging
 
 from pyomo.common.collections import ComponentMap, ComponentSet
 from pyomo.common.config import ConfigValue
@@ -43,6 +44,7 @@ from pyomo.contrib.solver.common.results import (
 )
 from pyomo.contrib.solver.common.solution_loader import SolutionLoaderBase
 
+logger = logging.getLogger(__name__)
 
 gurobipy, gurobipy_available = attempt_import('gurobipy')
 
@@ -442,7 +444,9 @@ class GurobiDirect(GurobiSolverMixin, SolverBase):
             results.incumbent_objective = None
             results.objective_bound = None
 
-        results.extra_info.iteration_count = grb_model.getAttr('IterCount')
+        results.extra_info.IterCount = grb_model.getAttr('IterCount')
+        results.extra_info.BarIterCount = grb_model.getAttr('BarIterCount')
+        results.extra_info.NodeCount = grb_model.getAttr('NodeCount')
 
         timer.start('load solution')
         if config.load_solutions:
--- a/pyomo/contrib/solver/solvers/gurobi_persistent.py
+++ b/pyomo/contrib/solver/solvers/gurobi_persistent.py
@@ -857,7 +857,9 @@ class GurobiPersistent(
             ):
                 results.incumbent_objective = None
 
-        results.extra_info.iteration_count = gprob.getAttr('IterCount')
+        results.extra_info.IterCount = gprob.getAttr('IterCount')
+        results.extra_info.BarIterCount = gprob.getAttr('BarIterCount')
+        results.extra_info.NodeCount = gprob.getAttr('NodeCount')
 
         timer.start('load solution')
         if config.load_solutions:
--- a/pyomo/contrib/solver/solvers/gurobi_persistent.py
+++ b/pyomo/contrib/solver/solvers/gurobi_persistent.py
@@ -857,7 +857,7 @@ class GurobiPersistent(
             ):
                 results.incumbent_objective = None
 
-        results.iteration_count = gprob.getAttr('IterCount')
+        results.extra_info.iteration_count = gprob.getAttr('IterCount')
 
         timer.start('load solution')
         if config.load_solutions:
--- a/pyomo/contrib/solver/solvers/highs.py
+++ b/pyomo/contrib/solver/solvers/highs.py
@@ -750,7 +750,15 @@ class Highs(PersistentSolverMixin, PersistentSolverUtils, PersistentSolverBase):
                     results.objective_bound = None
             else:
                 results.objective_bound = info.mip_dual_bound
-            results.iteration_count = info.simplex_iteration_count
+
+            if info.valid:
+                results.extra_info.iteration_counts = {
+                    'simplex_iteration_count': info.simplex_iteration_count,
+                    'ipm_iteration_count': info.ipm_iteration_count,
+                    'mip_node_count': info.mip_node_count,
+                    'pdlp_iteration_count': info.pdlp_iteration_count,
+                    'qp_iteration_count': info.qp_iteration_count,
+                }
 
         if config.load_solutions:
             if has_feasible_solution:
--- a/pyomo/contrib/solver/solvers/highs.py
+++ b/pyomo/contrib/solver/solvers/highs.py
@@ -752,13 +752,13 @@ class Highs(PersistentSolverMixin, PersistentSolverUtils, PersistentSolverBase):
                 results.objective_bound = info.mip_dual_bound
 
             if info.valid:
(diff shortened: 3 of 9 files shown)


pyomo/contrib/solver/solvers/highs.py
```diff
@@ -750,7 +750,15 @@ def _postsolve(self, stream: io.StringIO):
...
-            results.iteration_count = info.simplex_iteration_count
+
+            if info.valid:
+                results.extra_info.iteration_counts = {
+                    'simplex_iteration_count': info.simplex_iteration_count,
+                    'ipm_iteration_count': info.ipm_iteration_count,
+                    'mip_node_count': info.mip_node_count,
+                    'pdlp_iteration_count': info.pdlp_iteration_count,
+                    'qp_iteration_count': info.qp_iteration_count,
+                }
```
jsiirola: Why add this as a dictionary?  That makes accessing it awkward:
```python
results.extra_info.iteration_counts['simplex_iteration_count']
```
I think it would be better to either add these as a ConfigDict:
```python
_counts = ConfigDict()
_counts.declare('simplex_iteration_count', ConfigValue(domain=int))
# ....

    if info.valid:
        results.extra_info.iteration_counts = _counts({
            'simplex_iteration_count': info.simplex_iteration_count,
            'ipm_iteration_count': info.ipm_iteration_count,
            'mip_node_count': info.mip_node_count,
            'pdlp_iteration_count': info.pdlp_iteration_count,
            'qp_iteration_count': info.qp_iteration_count,
        })
```
OR just add the counts to the `extra_info`:
```python
results.extra_info.simplex_iteration_count = info.simplex_iteration_count
# ...
```

... I think I prefer the latter (just so that things don't get too deeply nested).  That said, I do like the explicit use of `declare` so that we could add documentation to the field (you can 'declare` into the extra_info, too - you don't have to create a special subclass of Results).

mrmundt: I also prefer the latter. Even though we are "losing" the documentation feature, that documentation should already exist somewhere in HiGHS, so it's redundant for us to do so.

pyomo/contrib/solver/solvers/gurobi_direct.py
```diff
@@ -442,7 +442,7 @@ def _postsolve(self, timer: HierarchicalTimer, config, loader):
             results.incumbent_objective = None
             results.objective_bound = None
 
-        results.iteration_count = grb_model.getAttr('IterCount')
+        results.extra_info.iteration_count = grb_model.getAttr('IterCount')
```
jsiirola: This is misleading: `IterCount` is the number of simplex iterations.  We should either call it `simplex_iteration_count` (and match HiGHS), or `IterCount` (to match the underlying Gurobi attribute).

In addition, we should report `BarIterCount` and `NodeCount`.

pyomo/contrib/solver/solvers/gurobi_persistent.py
```diff
@@ -857,7 +857,7 @@ def _postsolve(self, timer: HierarchicalTimer):
             ):
                 results.incumbent_objective = None
 
-        results.iteration_count = gprob.getAttr('IterCount')
+        results.extra_info.iteration_count = gprob.getAttr('IterCount')
```
blnicho: It seems a little strange that `IterCount` is stored as `iteration_count` in the persistent Gurobi interface but as `IterCount` in the direct Gurobi interface. Is there a reason why we aren't using consistent naming across they two interfaces?

mrmundt: The reason is, "I forgot that it also happened in gurobi_persistent" :P I'll go fix it.

blnicho (approved):

PR #3771, opened 2025-10-24
TYP: Enhance static typing

n-takumasa:
## Summary/Motivation:
To provide better autocompletion in vscode+pylance.
Complex dynamic patterns (e.g. `Var()` can return `ScalarVar` or `IndexedVar`) are out of scope as they are difficult to model accurately.

- TYP: Use `typing.overload` during `TYPE_CHECKING`
- TYP: `Model.__new__` should return an instance of the subclass
- TYP: better typing for `SolverFactory`s and `pyomo.future.solver_factory`
- TYP: reorder `@overload`s, overlapped overloads never be used

image

### Legal Acknowledgement


pyomo/common/pyomo_typing.py
```diff
@@ -18,16 +18,20 @@ def _get_fullqual_name(func: typing.Callable) -> str:
     return f"{func.__module__}.{func.__qualname__}"
 
 
-def overload(func: typing.Callable):
-    """Wrap typing.overload that remembers the overloaded signatures
+if typing.TYPE_CHECKING:
+    from typing import overload as overload
+else:
```
jsiirola: Why do you need to disable the overload?

pyomo/core/base/PyomoModel.py
```diff
@@ -572,6 +573,10 @@ def select(
         StaleFlagManager.mark_all_as_stale(delayed=True)
 
 
+# NOTE: Python 3.11+ use `typing.Self`
+ModelT = TypeVar("ModelT", bound="Model")
```
jsiirola: We haven't really settled on a typing convention for naming types, but appending `T` seems confusing.
- would standardizing on appendint `Type` be more clear?
- should local TypeVar objects be private by default (i.e., `_modelType` here)?

pyomo/core/base/block.py
```diff
@@ -2085,17 +2085,17 @@ class Block(ActiveIndexedComponent):
...
     def __new__(cls: Type[ScalarBlock], *args, **kwds) -> ScalarBlock: ...
 
     @overload
     def __new__(cls: Type[IndexedBlock], *args, **kwds) -> IndexedBlock: ...
 
+    @overload
+    def __new__(
+        cls: Type[Block], *args, **kwds
+    ) -> Union[ScalarBlock, IndexedBlock]: ...
+
```
jsiirola, line 2083: Why are you re-ordering the overloads?

pyomo/future.py
```diff
@@ -90,19 +113,19 @@ def solver_factory(version=None):
     if current is None:
         for ver, cls in versions.items():
             if cls._cls is _environ.SolverFactory._cls:
-                solver_factory._active_version = ver
+                solver_factory._active_version = ver  # type: ignore
```
jsiirola: We shouldn't need these annotations, should we?

pyomo/opt/base/solvers.py
```diff
@@ -144,6 +145,11 @@ def _solver_error(self, method_name):
 
 
 class SolverFactoryClass(Factory):
+    @overload
+    def __call__(self, _name: None = None, **kwds) -> "SolverFactoryClass": ...
+    @overload
+    def __call__(self, _name, **kwds) -> "OptSolver": ...
```
jsiirola, line 150: The [Python documentation](https://typing.python.org/en/latest/spec/overload.html#overload-definitions) indicates that you should only use the `Ellipsis` notation in `.pyi` files.  I think this should be:
```suggestion
    @overload
    def __call__(self, _name: None = None, **kwds) -> "SolverFactoryClass":
        pass
    @overload
    def __call__(self, _name, **kwds) -> "OptSolver":
        pass
```

jsiirola (commented):
Thanks for this!  I have several questions (see below).  Some are less about this PR specifically, and more about generally how we want to better support typing in Pyomo.  It would be useful to talk through that sometime in a Tuesday Dev Call.

pyomo/common/pyomo_typing.py
n-takumasa: A function calling `typing.overload` is not recognized as an overload by the type checker.
We need to replace it with `typing.overload` during type checking.

pyomo/core/base/PyomoModel.py
n-takumasa: `_ModelType` is better.

pyomo/core/base/block.py
n-takumasa: Because overloads are resolved sequentially from top to bottom when they overlap.
Therefore, the more specialized overloads need to be placed at the top.
ref: numpy/numpy#22193

pyomo/future.py
n-takumasa: Because we cannot add attributes to `FunctionType`.

pyomo/opt/base/solvers.py
n-takumasa: I do not believe the documentation states that you must not use `Ellipsis` in `.py` files.
In fact, it is uncommon to see `pass` used as the body for an `@overload`ed function.

pyomo/common/pyomo_typing.py
jsiirola: OK.  I went digging and it turns out that logic is primarily needed by a dependent project -- but only for Python versions through 3.10.  I think we should actually document that better in the code, with something like:
```python
if sys.version_info[:2] <= (3, 10) and not TYPE_CHECKING:
    def overload(func: typing.Callable):
        """Wrap typing.overload that remembers the overloaded signatures

        This provides a custom implementation of typing.overload that
        remembers the overloaded signatures so that they are available for
        runtime inspection (backporting `get_overloads` from Python 3.11+).

        """
        _overloads.setdefault(_get_fullqual_name(func), []).append(func)
        return typing.overload(func)

    def get_overloads_for(func: typing.Callable):
        return _overloads.get(_get_fullqual_name(func), [])
else:
    from typing import overload, get_overloads as get_overloads_for
```

pyomo/core/base/PyomoModel.py
jsiirola: Agreed.

pyomo/future.py
jsiirola: You can't annotate the function, plus the attribute whereit is first instantiated at the bottom of the file?
```python
def solver_factory(version: int | None = None) -> int:
# ...
solver_factory._active_version: int = solver_factory()
```

pyomo/common/pyomo_typing.py
n-takumasa: [typing_extensions](https://github.com/python/typing_extensions/blob/f684c9a36b8490985bf285404810a65ef6fdfc78/src/typing_extensions.py#L475) might be helpful.

pyomo/future.py
n-takumasa: It is allowed at runtime, but it's a type violation.
ref: microsoft/pyright#8838

jsiirola:
We were talking though this PR on the dev call this week.  I think there are 3 changes that need to be made to get it merged:
- rename `modelT` to `ModelType`
- rework the `TYPE_CHECKING` test, possibly using the suggestion from the review
- refactor the `solver_factory` function to promote the state to a module-level variable so that it can be statically-typed in a way that makes pyright happy.

n-takumasa:
Almost done.

> * rename `modelT` to `ModelType`

Is this intended to be public?
For now, I committed it as private with `_ModelType`.

pyomo/future.py
```diff
@@ -28,14 +26,49 @@
...
+import pyomo.environ as _environ
+import pyomo.opt.base.solvers as _solvers
+
+_SolverFactoryClassV1 = _solvers.SolverFactoryClass
+_SolverFactoryClassV2 = _appsi.SolverFactoryClass
+_SolverFactoryClassV3 = _contrib.SolverFactoryClass
+
+solver_factory_v1: _SolverFactoryClassV1 = _solvers.LegacySolverFactory
+solver_factory_v2: _SolverFactoryClassV2 = _appsi.SolverFactory
+solver_factory_v3: _SolverFactoryClassV3 = _contrib.SolverFactory
```
jsiirola: We don't want to declare these at the module scope: if we do, the module `__getattr__` won't be called, which was how we were implementing the Python idiom:

```python
from pyomo.__future__ import solver_factory_v3
```
as an alias for
```python
from pyomo.__future__ import solver_factory
solver_factory(3)
```

pyomo/future.py
```diff
@@ -28,14 +26,49 @@
...
+
+_versions = {1: solver_factory_v1, 2: solver_factory_v2, 3: solver_factory_v3}
```
jsiirola: Future is intended to manage all "future compatibility changes", so we should use a more specific name than `versions` here - there will eventually be several "versions" that this module will manage.

I would prefer *not* promoting this dict to the module scope: `solver_factory` will probably never be called more than once, so the performance gain of preserving the dict isn't worth the memory.

pyomo/future.py
```diff
@@ -28,14 +26,49 @@
...
+from typing import Any, Literal, overload
+
+import pyomo.contrib.appsi.base as _appsi
+import pyomo.contrib.solver.common.factory as _contrib
+import pyomo.environ as _environ
+import pyomo.opt.base.solvers as _solvers
+
+_SolverFactoryClassV1 = _solvers.SolverFactoryClass
+_SolverFactoryClassV2 = _appsi.SolverFactoryClass
+_SolverFactoryClassV3 = _contrib.SolverFactoryClass
```
jsiirola: I would prefer not caching these at the module scope.

jsiirola (commented):
Some minor comments.  The biggest is that I would prefer to see the changes to `pyomo.future.solver_factory()` be reverted *with the exception* of promoting the `_active_version` attribute on the function to a `_active_solver_factory_version` module attribute.  The changes add an unnecessary (admittedly slight) overhead and break some of the documented functionality.

pyomo/future.py
n-takumasa: My bad, it's regression.
I didn't realize that importing changes the state.

pyomo/future.py
```diff
@@ -84,25 +108,26 @@ def solver_factory(version=None):
...
+    try:
+        _ = _active_solver_factory_version
+    except NameError:
         for ver, cls in versions.items():
             if cls._cls is _environ.SolverFactory._cls:
-                solver_factory._active_version = ver
+                _active_solver_factory_version = ver
                 break
-        return solver_factory._active_version
+        return _active_solver_factory_version
```
jsiirola: [This is my fault] As I am thinking through this, I think this is actually a bug and this should be reworked.  Right now, everything is "fine" because we are guaranteed that the first time this function is called, it will be called with `version=None` (that is hard-coded below and will happen on import).  BUT that seems fragile.  I think I would propose combining this logic with the `if version is None` below:
```python
if version is None:
    if "_active_solver_factory_version" not in globals():
        for ver, cls in versions.items():
            if cls._cls is _environ.SolverFactory._cls:
                _active_solver_factory_version = ver
                break
    return _active_solver_factory_version
```

@n-takumasa: totally up to you if you want to make this change in this PR or leave it to me to get back to later

pyomo/future.py
```diff
@@ -28,14 +35,30 @@
 
 """
 
+solver_factory_v1: "_solvers.SolverFactoryClass"
+solver_factory_v2: "_appsi.SolverFactoryClass"
+solver_factory_v3: "_contrib.SolverFactoryClass"
+
```
jsiirola, line 41: We need to delete these lines so that the `from pyomo.__future__ import solver_factory_v3` logic works.

(And some time we should add unit tests for this [maybe put it in `pyomo/environ/tests/test_future.py`?].  @n-takumasa: I'm not going to argue to hold up this PR for more testing, though.)

jsiirola (commented):
This is *super* close.  If we can delete the 3 lines indicated, then I think this is good to go.

pyomo/future.py
n-takumasa: These lines declare types without actually defining variables, so the logic works via `__getattr__`.

jsiirola: Doh.  You're right, of course.

jsiirola (approved):

pyomo/future.py
n-takumasa: I agree because if we delete `_active_solver_factory_version = solver_factory()` and call `solver_factory(int)`, it will break.

blnicho (approved):

PR #3774, opened 2025-10-27
contrib.solver: Gurobi warm start option

emma58:
## Summary/Motivation:

The only Gurobi interface in the solver rewrite that was supporting warm starting was `gurobi_persistent`. This PR changes the option name to `warm_start` (from `use_mipstart`) and changes the behavior to pass any variable with non-`None` value as part of the warm start. It then adds an implementation of this option in `gurobi_direct` and `gurobi_direct_minlp`. I'm open to being argued with--my thinking is that this is simple and allows the user control over integer/not warm starting since they can change variable values. But maybe that isn't the best for persistent since that comes with some overhead?

- Adds warm start support in the regular and MINLP direct interfaces
- Adds testing for warm starts by capturing the log and making sure Gurobi took it
- Fixes bugs in Gurobi MINLP with `tee` ignoring being given a logger

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/piecewise/tests/test_nonlinear_to_pwl.py
+++ b/pyomo/contrib/piecewise/tests/test_nonlinear_to_pwl.py
@@ -19,6 +19,8 @@ from pyomo.contrib.piecewise import PiecewiseLinearFunction
 from pyomo.contrib.piecewise.transform.nonlinear_to_pwl import (
     NonlinearToPWL,
     DomainPartitioningMethod,
+    lineartree_available,
+    sklearn_available,
 )
 from pyomo.core.base.expression import _ExpressionData
 from pyomo.core.expr.compare import (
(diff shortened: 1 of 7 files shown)


pyomo/contrib/solver/solvers/gurobi_persistent.py
```diff
@@ -303,14 +303,22 @@ def _solve(self):
...
+                    if pyomo_var.value is not None:
                         self.set_var_attr(pyomo_var, 'Start', pyomo_var.value)
+            # elif config.warm_start_integer_vars:
+            #     for (
+            #         pyomo_var_id,
+            #         gurobi_var,
+            #     ) in self._pyomo_var_to_solver_var_map.items():
+            #         pyomo_var = self._vars[pyomo_var_id][0]
+            #         if pyomo_var.is_integer() and pyomo_var.value is not None:
+            #             self.set_var_attr(pyomo_var, 'Start', pyomo_var.value)
```
emma58, line 321: I was thinking we could do this in all the interfaces if we want a separate option for just warm starting integer vars, but dunno if we need to add the complexity?

emma58:
This is a duplicate of #3698, so I'm closing it. We can steal tests from here though.

PR #3778, opened 2025-10-30
Enhance TimingHandler class documentation comment

Somesh-Kennedy:
## Fixes # .
Added docstring (enhanced the comment) for TimingHandler class, detailing attributes and example data structure.

## Summary/Motivation:
Increasing the understandability of the code (for the whole TimingHandler class)

- added an example attribute in the comment in order to understand the returned object.

### Legal Acknowledgement

### DIFF
--- a/scripts/performance/main.py
+++ b/scripts/performance/main.py
@@ -41,7 +41,6 @@ class TimingHandler(logging.Handler):
 
     Attributes:
         _testRecord (OrderedDict): Current test's timing data collection
-        enabled (bool): Flag to enable/disable timing collection
 
     Example timing data structure:
         {


scripts/performance/main.py
```diff
@@ -39,6 +39,18 @@ class TimingHandler(logging.Handler):
     transformations.  It collects these records, extracts the timing
     information and adds it to the test data recorder.
 
+    Attributes:
+        _testRecord (OrderedDict): Current test's timing data collection
+        enabled (bool): Flag to enable/disable timing collection
```
mrmundt: I was looking over the class to make sure these changes are accurate, and I think `enabled` doesn't actually do anything. It's set in `__init__` but... then is never used again. Which leads me to believe that maybe we should just remove `enabled` (and this line from the docs).

Somesh-Kennedy: If the enabled attribute is only set in __init__ and then never actually checked or used within any other method (like to gate the timing collection logic around _testRecord updates), then you are absolutely right. It's probably a redundant code.

scripts/performance/main.py
```diff
@@ -39,6 +39,17 @@ class TimingHandler(logging.Handler):
     transformations.  It collects these records, extracts the timing
     information and adds it to the test data recorder.
 
+    Attributes:
+        _testRecord (OrderedDict): Current test's timing data collection
```
jsiirola: We generally prefer numpydoc format for docstrings

scripts/performance/main.py
```diff
@@ -39,6 +39,17 @@ class TimingHandler(logging.Handler):
...
+
+    Example timing data structure:
```
jsiirola: It would be better to make this a pair of `.. testcode::` / `.. testoutput::` blocks so that the doctests can ensure that the examples are correct.

blnicho:
@Somesh-Kennedy we discussed this at the dev call and decided to close this without merging. We don't think documenting a private attribute in an untested example script meets the minimum threshold of "usefulness" we expect from PRs. If you disagree with our decision please reach out or leave a comment.

PR #3781, opened 2025-11-11
Bugfixes and testing of linear templates

jsiirola:
## Summary/Motivation:
This PR is primarily adding  more comprehensive testing of templatizing (and compiling) linear models.  As part of flusing out the tests, this resolves a number of issues / bugs in the templatizer, the linear template compiler, and in general Pyomo:

General Pyomo changes
- Rework `Param.extract_values` to return `defaultdict` for Params with default values
- Provide a specialized `InvalidConstraintError` exception for constraints that cannot be emitted to the solvers
- Use `SetOf` when iterating over IndexedComponent objects indexed by non-finite Sets
- `SetOf.dimen` should return `UnknownSetDimen` (*not* 0) if the underlying reference has no data

Updates to the templatizer
- Leverage `attempt_import` to break circular dependencies in `template_expr` 
- Catch `UnknownSetDimen` when iterating over template generators and treat like jagged sets
- Expand (do not templatize) expressions with `sum()` that do not sum over generators
- Expand (do not templatize) expressions with `sum()` whose generator doesn't create a template index
- Expand (do not templatize) expressions with `sum()` that appear to iterate over things other than Pyomo Set objects

Updates to the linear template compiler
- remove `check_duplicates` and `remove_fixed_vars` options (this is not more efficiently implementable through sparse matrix operations0
- rework expression builders to be more testable and (slightly) more efficient
- Support `IndexedComponent.Skip`
- Resolve issue compiling constraints indexed by non-finite Sets

- (see above)

### Legal Acknowledgement


pyomo/core/base/constraint.py
```diff
@@ -219,7 +223,7 @@ def to_bounded_expression(self, evaluate_bounds=False):
                 and lb.is_potentially_variable()
                 and not lb.is_fixed()
             ):
-                raise ValueError(
+                raise InvalidConstraintError(
                     f"Constraint '{self.name}' is a Ranged Inequality with a "
                     "variable lower bound.  Cannot normalize the "
                     "constraint or send it to a solver."
```
emma58, line 229: While we're changing things here, is 'normalize' the right word? Maybe the second sentence of this error should be much simpler: "Pyomo doesn't allow Constraints of the form lb <= body <= ub where either of lb or ub contains variables."

pyomo/core/base/constraint.py
```diff
@@ -229,7 +233,7 @@ def to_bounded_expression(self, evaluate_bounds=False):
                 and ub.is_potentially_variable()
                 and not ub.is_fixed()
             ):
-                raise ValueError(
+                raise InvalidConstraintError(
                     f"Constraint '{self.name}' is a Ranged Inequality with a "
                     "variable upper bound.  Cannot normalize the "
                     "constraint or send it to a solver."
```
emma58, line 239: Same comment here.

pyomo/core/expr/template_expr.py
```diff
@@ -45,9 +46,64 @@
...
 )
 
+# Deferred imports to break circular dependencies
+pyomo_core_base_set, _ = attempt_import('pyomo.core.base.set')
+pyomo_core_base_param, _ = attempt_import('pyomo.core.base.param')
+
 logger = logging.getLogger(__name__)
 
 
+def _validate_generator(generator):
```
emma58, line 54: Naming question: Should this be `_check_generator_templatizable` or something of the like? It's not really invalid... Just impossible.

pyomo/core/expr/template_expr.py
```diff
@@ -45,9 +46,64 @@
...
+    return True
+
+
+def _validate_map(generator):
+    # We would love to validate that the map is actually iterating over
+    # Pyom Set objects (and yielding IndexTemplate objects), but there
+    # doesn't appear to be a way to interrogate the results of the list
+    # / generator that the map is iterating over.  So, we will just have
+    # to trust the user <shudder>.
+    #
```
emma58, line 91:
```suggestion
    # We would love to validate that the map is actually iterating over
    # Pyomo Set objects (and yielding IndexTemplate objects), but there
    # doesn't appear to be a way to interrogate the results of the list
    # / generator that the map is iterating over.  So, we will just have
    # to trust the user. <shudder>.
    #
```

pyomo/core/expr/template_expr.py
```diff
@@ -1032,7 +1082,10 @@ def __next__(self):
         else:
             d = _set.dimen
         grp = context.next_group()
-        if d is None or type(d) is not int:
+        if type(d) is not int:
+            # This covers None (jagged set) and UnknownSetDimen.  In
+            # both cases, we will not attempt to unpack the Set and just
+            # assume a single index template.
```
emma58, line 1086: Why is this safe for `UnknownSetDimen`? Or actually either of them?

emma58: Because will it catch if someone is trying to templatize an expression where *they* unpacked something multidimensional? Or is the idea that that will die then, so just keep going for now?

pyomo/repn/tests/test_linear_template.py
```diff
@@ -0,0 +1,1227 @@
...
+
+import pyomo.core.base.constraint as constraint
+import pyomo.core.base.objective as objective
+
+from pyomo.common.errors import InvalidConstraintError
+from pyomo.core.base.enums import SortComponents
+from pyomo.repn.linear_template import LinearTemplateRepnVisitor
+from pyomo.repn.util import TemplateVarRecorder
+
+from pyomo.environ import *
```
emma58: Miranda's going to be mad at you!

pyomo/core/expr/template_expr.py
```diff
@@ -43,9 +44,64 @@
...
+    #
+    # FIXME: rework IndexedComponent to return custom generators that
+    # wrap map so we can only accept them and not all maps?
+    return True
+
+
+# it is not clear what to import to get to the built-in "generator"
+# type.  We will just create a generator and query its __class__
+generator_validators = {
+    (_ for _ in ()).__class__: _validate_generator,
```
emma58, line 100: Wow, that one took me a minute, even with the comment! How strange that it's hard to import...

pyomo/core/expr/template_expr.py
```diff
@@ -1078,10 +1131,25 @@ def next_group(self):
...
+        except KeyError:
+            # We will only templatize sums over maps and generators.
+            # Expand everything else:
+            return _TemplateIterManager.builtin_sum(generator)
+        niters = -len(self.cache)
         expr = next(generator)
-        final_cache = len(self.cache)
-        return TemplateSumExpression((expr,), self.npop_cache(final_cache - init_cache))
+        niters += len(self.cache)
+        if niters:
```
emma58, line 1143: This is just me not understanding the old code either, but how does `next(generator)` change the state of `self.cache`?

emma58 (commented):
I kind of ran out of brainpower in the crux, so I'll look at `linear_template.py` again in the morning, but here are a few comments from elsewhere. I'm very exited that it's tested!!

pyomo/core/base/constraint.py
jsiirola: I am open to a better verb, but don't think we should use your proposed text.  The problem is that we *do* allow specifying constraints with variables in the LB/UB.  We just don't support sending them to most solvers (this was actually requested early on by users who "promised" to fix the variables or transform the model before solving it).

The idea here is we will happily carry around the relational expression provided by the user until the point that someone wants the constraint "normalized" into `(lb, body, ub)` form - where we promise that lb and ub are constants / fixed expressions (or raise this exception).

Maybe "standardized" is a better verb?

pyomo/core/expr/template_expr.py
jsiirola: I don't know.  Naming is hard.  This isn't checking that the generator is templatizable - that has already happened (when we call `next(generator)` in `sum_template` and don't get a `TemplateExpressionError`).  So at this point, we already *have* a template expression and we are looking into the state of the generator to check that the internal state of the generator is "valid" (i.e., it doesn't have any local variables that are not IndexTemplate objects).  So we really *are* validating that the generator is (in a) "correct" (state).

pyomo/core/expr/template_expr.py
jsiirola: What is happening here is that we don't know how many scalars (i.e., IndexTemplates) to unpack the set item into.  So we *won't* try and unpack anything (and hope that the user won't either).  In this case we will return a single IndexTemplate that is the placeholder for the *arbitrary tuple* that would be returned by the Set.  As long as the user doesn't try and unpack that tuple (and they shouldn't without first looking at the length of the tuple - which will generate the TemplateExpressionError that disables templatization).

Here's an example:

```python
m.I = Set(dimen=None, initialize=[(1,), (2, 3), (4, 5, 6)])
m.x = Var(m.I)
m.c = Constraint(expr=sum(m.x[i] for i in m.I) <= 0)
```

Here, `i` is a tuple of some "random" length.  So we want to templatize this as
```
sum(x[_1] for _1 in I)
```

pyomo/core/expr/template_expr.py
jsiirola: This is the core trick in Templatization of sum expressions: `next(generator)` calls the generator within the sum to return the first term. That in turn calls `iter(...)` on the thing (usually a Set) being summed over. For Set objects, that hits the overriden `__iter__` implementation (see `_set_iterator_template_generator`, managed by `_template_iter_context`) that adds to the context cache)

pyomo/repn/tests/test_linear_template.py
jsiirola: In my defense, this PR may predate the removal of `import *`... ;)

pyomo/core/base/constraint.py
emma58: I think I prefer "standardized." "Normalize is just confusing since it's mathematical meaning almost makes sense too, out of context.

pyomo/core/expr/template_expr.py
emma58: Ah, okay. Then maybe your name makes sense!

pyomo/core/expr/template_expr.py
emma58: OK, I see!

pyomo/core/expr/template_expr.py
emma58: Ohhhhh. This would be a nice thing to write down in developer docs someday! :)

pyomo/repn/linear_template.py
```diff
@@ -61,23 +78,27 @@ def multiplier_flag(val):
...
 
+    def to_expression(self, visitor):
+        # to_expression() is only used by the underlying
+        # LinearRepnVisitor to generate nonlinear expressions.  We are
+        # explicitly disallowing nonlinear expressions here, so we are
+        # going to bail now:
+        raise InvalidConstraintError(
+            "LinearTemplateRepn does not support constraints containing "
+            "general nonlinear terms."
+        )
```
emma58: Is it possible to give the name or some pointer to the offending Constraint in this error?

emma58 (approved):
OK, please don't get hit by a bus, but I kind of understand this, and I reallllly like it!

pyomo/repn/linear_template.py
jsiirola: Added.  But that took way more work/code than you might think.

pyomo/repn/tests/test_linear_template.py
```diff
@@ -0,0 +1,1257 @@
+#  ___________________________________________________________________________
+#
+#  Pyomo: Python Optimization Modeling Objects
+#  Copyright (c) 2008-2025
+#  National Technology and Engineering Solutions of Sandia, LLC
+#  Under the terms of Contract DE-NA0003525 with National Technology and
+#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain
+#  rights in this software.
+#  This software is distributed under the 3-clause BSD License.
+#  ___________________________________________________________________________
```
mrmundt: Oh noooo, John... Old copyright. I'll go fix it, this comment is just a heads up

mrmundt (approved):

pyomo/repn/linear_template.py
```diff
@@ -37,14 +46,29 @@ def __init__(self):
...
+        linear_sum = '[' + ', '.join(linear_sum) + ']'
         return (
-            f"{self.__class__.__name__}(mult={_str(self.multiplier)}, "
-            f"const={_str(self.constant)}, "
+            f"{self.__class__.__name__}(mult={val2str(self.multiplier)}, "
+            f"const={val2str(self.constant)}, "
             f"linear={linear}, "
-            f"linear_sum={self.linear_sum}, "
+            f"linear_sum={linear_sum}, "
             f"nonlinear={self.nonlinear})"
```
blnicho, line 72: Do you need the nonlinear line in the string representation if we're asserting that `self.nonlinear is None` for linear templates?

blnicho (approved):
I have a question but otherwise this can be merged!

PR #3790, opened 2025-11-21
Update Config classes / testing / documentation

jsiirola:
## Summary/Motivation:
This PR was motivated by two new features in the Config system:
- `ConfigDict.add()` should accept keyword arguments so that users can specify modifiers like `visibility` without explicitly creating the `ConfigValue`
- `ConfigDict` should derive from `MutableMapping` so that methods like `update()` are defined / available.

As part of implementing that, this PR made a number of other improvements
- [Performance] Split `ConfigBase.__call__` so that the `ConfigDict`-specific logic is in an overload method on `ConfigDict`
- [Performance] Avoid unnecessary calls to `generate_documentation()`
- [Performance] re-use method documenter in `document_class_CONFIG`
- [bugfix] fix logic when trapping recursion errors in domain pickler
- [feature] Expose additional classes from `enum` (`Flag`, `IntFlag`, `EnumType`)
- Update / expand unit testing
- Significant update / expansion on documentation

## Changes proposed in this PR:
- (see above)

### Legal Acknowledgement

By contributing to this software project, I have read the [contribution guide](https://pyomo.readthedocs.io/en/stable/contribution_guide.html) and agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #3792, opened 2025-12-02
Update Ipopt interface in `contrib.solver`

### LINKED ISSUE #3807: New solver interface doesn't return bound dual variables from IPOPT's restoration phase
## Summary

Presently, Pyomo's new solver interface returns dual variables for variable bonds [only if the NLP has an objective function](https://github.com/Pyomo/pyomo/blob/a079b2579a78cb8d1d8483d098860060c1984ff9/pyomo/contrib/solver/solvers/ipopt.py#L119). However, IPOPT returns dual variables not just when optimizing an objective function, but also when it returns `EXIT: Converged to a point of local infeasibility. Problem may be infeasible.` The dual variables in this case are extremely helpful for determining which variable bounds are causing the termination point to be a local minimizer for primal infeasibility.

I know these dual variables are getting returned by IPOPT, because they get loaded in when the suffixes `ipopt_zL_out` and `ipopt_zU_out` are created and the problem is solved through the old IPOPT interface. While it would be possible to manually add the function for primal infeasibility that IPOPT is minimizing in the restoration phase, it would also be extremely tedious because it involves creating slack variables for both the lower and upper bounds.

Pyomo version: 6.9.5
Python version: 3.12
Operating system: Windows 11
How Pyomo was installed (PyPI, conda, source): Some combination of IDAES and Watertap developer installations
Solver (if applicable): IPOPT

jsiirola:
## Summary/Motivation:
This started as an update the to SOL file parser to resolve some holes in the parse logic around parsing AMPL options, and turned into a significant rework of the Ipopt solver interface.

- rewrite of the SOL file parser.  The parser no longer requires any external information to parse a SOL file.
- update the ipopt log parser to make the iteration log parse cleaner
- rework how ipopt options are parsed so that both command line and option file options can be processed in a single pass through the user-provided options.  For efficiency, iterate over the (usually much shorter) list us user-provided options and not the list of all known options.
- refactor the ipopt interface so that in the future the bulk of it can be abstracted into a generic "ASL" interface that other ASL solvers can leverage.
- add a `delete=` option to the TempfileManager contexts so that we can create "unmanaged" temporary files through the context.
- update some documentation
- expand test coverage (cover 100% of the ipopt interface!)

### Legal Acknowledgement


pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -568,22 +472,142 @@ def solve(self, model, **kwds) -> Results:
...
+
+        # This is the data we need to parse to get the iterations
+        # and time
+        timer.start('parse_log')
+        parsed_output_data = self._parse_ipopt_output(results.solver_log)
+        results.extra_info.iteration_count = parsed_output_data.pop('iters', None)
+        _timing = parsed_output_data.pop('cpu_seconds', None)
+        if _timing:
+            # TODO: once #3790 is merged, this is just:
+            #   results.timing_info.update(_timing)
```
blnicho: #3790 has been merged

blnicho (approved):

pyomo/contrib/solver/solvers/ipopt.py
```diff
@@ -88,76 +110,62 @@ def __init__(
...
-        nl_info = self._nl_info
-        zl_map = sol_data.var_suffixes['ipopt_zL_out']
-        zu_map = sol_data.var_suffixes['ipopt_zU_out']
-        rc = {}
-        for ndx, v in enumerate(nl_info.variables):
-            scale = scale_list[ndx]
-            v_id = id(v)
-            rc[v_id] = (v, 0)
+        if self._nl_info.eliminated_vars:
+            raise MouseTrap(
```
michaelbynum, line 127: Why is this a mouse trap? I'm not sure I understand when MouseTrap should be expected?

michaelbynum: Oh, I see! It's in the doc string for MouseTrap!

PR #3802, opened 2025-12-16
[KNITRO] Separate Extra Information by MIP Relevance

yemine0x01:
## Summary/Motivation:

This PR updates the iteration count handling for KNITRO. Since the iteration count is not meaningful for MIP problems, the solver now returns None in those cases.

-
-

### Legal Acknowledgement


pyomo/contrib/solver/solvers/knitro/base.py
```diff
@@ -140,7 +140,10 @@ def _postsolve(self, config: KnitroConfig, timer: HierarchicalTimer) -> Results:
         results.solution_status = self._get_solution_status(status)
         results.termination_condition = self._get_termination_condition(status)
         results.incumbent_objective = self._engine.get_obj_value()
-        results.extra_info.iteration_count = self._engine.get_num_iters()
+        if self._is_mip():
+            results.extra_info.iteration_count = self._engine.get_num_iters()
+        else:
+            results.extra_info.iteration_count = None
```
mrmundt: This is kind of misleading. It reads as, "If this is a MIP, get an iteration count; otherwise, set to none." I think it should probably be the other way, right? If this is NOT a MIP, get an iteration count?

yemine0x01: Good catch. Thank you for that.

yemine0x01: By the way, I am working on this to add the info that we want to report when the problem is MIP.

yemine0x01:
@mrmundt I just update this PR with what we would like to report as extra info in both cases.

mrmundt (approved):
Love it! Thank you! 🥇

blnicho (approved):

PR #3803, opened 2025-12-16
Pyomo.DoE: Correct A optimality

smondal13:
## Summary/Motivation:
In the old definition of A-optimality (we will call it pseudo A-optimality from now on), DoE computed the trace of the Fisher Information Matrix (FIM), as few studies used that definition. However, that is not correct, and new literature mentions the trace of the inverse of the FIM as A-optimality. This PR has adopted this new definition of A-optimality in DoE. 

- Change the old A-optimality  to pseudo A-optimality
- Implement a new definition of A-optimality (=trace(inverse of FIM))
- Add new tests for the DoE plotting functionality

### Legal Acknowledgement


smondal13:
@adowling2 @blnicho This PR is ready for initial review.

pyomo/contrib/doe/doe.py
```diff
@@ -1353,12 +1429,86 @@ def cholesky_imp(b, c, d):
...
+            # If trace objective, need L * L^-1 = Identity matrix constraints
+            def cholesky_LLinv_imp(b, c, d):
+                """
+                Calculate Cholesky L * L inverse matrix using algebraic constraints
+                """
+                # If the row is greater than or equal to the column, we are in the
+                # lower triangle region of the L and L_inv matrices.
+                # This region is where our equations are well-defined.
+                m = b.model()
+                param_list = list(model.parameter_names)
```
blnicho:
```suggestion
                param_list = list(m.parameter_names)
```

pyomo/contrib/doe/doe.py
```diff
@@ -1663,12 +1846,13 @@ def compute_FIM_full_factorial(
...
             - "log10 D-opt": list of log10(D-optimality)
             - "log10 A-opt": list of log10(A-optimality)
+            - "log10 pseudo A-opt": list of log10(trace(FIM))
             - "log10 E-opt": list of log10(E-optimality)
             - "log10 ME-opt": list of log10(ME-optimality)
             - "eigval_min": list of minimum eigenvalues
             - "eigval_max": list of maximum eigenvalues
             - "det_FIM": list of determinants
-            - "trace_FIM": list of traces
+            - "trace_cov": list of traces
```
blnicho: Here you have replaced `trace_FIM` with `trace_cov` but down on line 1911, when setting up the `fim_factorial_results` object you have appended `trace_cov` so the dictionary contains both which seems inconsistent. Did you mean to include both or just one of `trace_FIM` and `trace_cov`?

pyomo/contrib/doe/doe.py
```diff
@@ -1718,11 +1902,13 @@ def compute_FIM_full_factorial(
...
                 "log10 D-opt": [],
                 "log10 A-opt": [],
+                "log10 pseudo A-opt": [],
                 "log10 E-opt": [],
                 "log10 ME-opt": [],
                 "eigval_min": [],
                 "eigval_max": [],
                 "det_FIM": [],
+                "trace_cov": [],
                 "trace_FIM": [],
```
blnicho, line 1914: See my previous comment but pointing out places where both `trace_cov` and `trace_FIM` are included.

pyomo/contrib/doe/doe.py
```diff
@@ -1785,21 +1971,32 @@ def compute_FIM_full_factorial(
 
             FIM = self._computed_FIM
 
-            (det_FIM, trace_FIM, E_vals, E_vecs, D_opt, A_opt, E_opt, ME_opt) = (
-                compute_FIM_metrics(FIM)
-            )
+            (
+                det_FIM,
+                trace_cov,
+                trace_FIM,
```
blnicho, line 1979: See my previous comment but pointing out places where both `trace_cov` and `trace_FIM` are included.

pyomo/contrib/doe/doe.py
```diff
@@ -1785,21 +1971,32 @@ def compute_FIM_full_factorial(
...
             fim_factorial_results["log10 D-opt"].append(D_opt)
             fim_factorial_results["log10 A-opt"].append(A_opt)
+            fim_factorial_results["log10 pseudo A-opt"].append(pseudo_A_opt)
             fim_factorial_results["log10 E-opt"].append(E_opt)
             fim_factorial_results["log10 ME-opt"].append(ME_opt)
             fim_factorial_results["eigval_min"].append(E_vals.min())
             fim_factorial_results["eigval_max"].append(E_vals.max())
             fim_factorial_results["det_FIM"].append(det_FIM)
+            fim_factorial_results["trace_cov"].append(trace_cov)
             fim_factorial_results["trace_FIM"].append(trace_FIM)
```
blnicho, line 2002: See my previous comment but pointing out places where both `trace_cov` and `trace_FIM` are included.

pyomo/contrib/doe/utils.py
```diff
@@ -141,7 +141,7 @@ def compute_FIM_metrics(FIM):
 
     det_FIM : float
         Determinant of the FIM.
-    trace_FIM : float
+    trace_cov : float
```
blnicho, line 144: Here again you have replaced `trace_FIM` with `trace_cov` but in most other places you have kept both.

pyomo/contrib/doe/utils.py
```diff
@@ -203,7 +220,7 @@ def get_FIM_metrics(FIM):
 
     "Determinant of FIM" : float
         determinant of the FIM
-    "Trace of FIM" : float
+    "Trace of cov" : float
```
blnicho, line 225: Again, replacement here instead of including both traces

pyomo/contrib/doe/tests/test_doe_build.py
blnicho, line 1: Looks like you replaced every instance of `trace` with `pseudo_trace` in these tests. `trace` is still a valid objective type so I'm curious why you wouldn't want a mix of tests with both options.

pyomo/contrib/doe/tests/test_doe_errors.py
blnicho, line 1: Same question about testing both `trace` and `pseudo_trace` objective types.

blnicho (changes requested):

pyomo/contrib/doe/tests/test_doe_build.py
smondal13: I have not done any testing for the new `trace`(trace of the covariance matrix). The previous Pyomo.DoE versions that used `trace` actually meant trace of Fisher Information Matrix (FIM). Now, following Dan's Greybox paper, we are referring to the `trace of Fisher Information Matrix (FIM)` as  `pseudo_trace`. That's why I have replaced all instances of `trace` with `pseudo_trace`. We need new tests for the new `trace` (trace of the covariance matrix). I plan to do that after the review.  Does this answer your question? It may be confusing.

pyomo/contrib/doe/tests/test_doe_errors.py
smondal13: I have not yet added any test for the new `trace (trace of the covariance matrix)`. So, all previous tests have changed to `pseudo_trace (trace of the FIM)`

pyomo/contrib/doe/utils.py
smondal13: I have added both now.

pyomo/contrib/doe/utils.py
smondal13: Added both now. Initially, I thought I would discard pseudo_A_optimality. But later thought about keeping it since it was already there. That's why there is this mismatch in some cases. Sorry for that.

pyomo/contrib/doe/tests/test_doe_build.py
blnicho: Thanks for the clarification. I think we should include those tests in this PR before it gets merged.

blnicho (commented):
@smondal13 thanks for addressing my initial review comments. I think these changes look reasonable so far but would like to see tests for the corrected A-optimality before I'll approve the PR.

pyomo/contrib/doe/tests/test_doe_build.py
smondal13: I will add tests for the new `trace (trace of the covariance matrix)`. This PR is for initial review to determine whether I need to make any major changes before I proceed with the testing. Thanks for reviewing.

pyomo/contrib/parmest/examples/rooney_biegler/doe_example.py
smondal13, line 1: @blnicho @adowling2 I have created a DoE example for the rooney_biegler experiment, mainly for testing DoE. Where should this live? I can move it to `parameter_estimation_example.py` and rename the file to `rooney_biegler_example.py` or something similar. Or are you okay with the current standalone script?

smondal13:
@blnicho @adowling2 I have added tests for A-optimality and added tests for the plotting functionality in `doe.py` as well. This is now ready for final review.

smondal13:
@adowling2, this PR is ready for final review.

pyomo/contrib/parmest/examples/rooney_biegler/doe_example.py
blnicho: @smondal13 I would suggest moving this example under `pyomo/contrib/doe/examples` so that it's easier for users to find. I think it's fine to import the `RooneyBieglerExperiment` class from parmest within a DoE example. Maybe rename the file to something like `rooney_biegler_doe_example.py` to distinguish it from the other DoE examples.

blnicho: @smondal13 I would suggest moving this example under `pyomo/contrib/doe/examples` so that it's easier for users to find. I think it's fine to import the RooneyBieglerExperiment class from parmest within a DoE example. Maybe rename the file to something like `rooney_biegler_doe_example.py` to distinguish it from the other DoE examples.

pyomo/contrib/parmest/examples/rooney_biegler/doe_example.py
```diff
@@ -0,0 +1,193 @@
...
+    if optimize_experiment_D:
+        # D-Optimality
+        rooney_biegler_doe_D = DesignOfExperiments(
+            experiment=rooney_biegler_experiment,
+            objective_option="determinant",
+            tee=tee,
+            prior_FIM=FIM,
+            improve_cholesky_roundoff_error=True,
+        )
+        rooney_biegler_doe_D.run_doe()
```
blnicho: Why do you need both `rooney_biegler_doe` and `rooney_biegler_doe_D`? Seems like they are identical instances of the `DesignOfExperiments` class. Why not just call `run_doe` directly on `rooney_biegler_doe` when `optimize_experiment_D` is `True`?

pyomo/contrib/parmest/examples/rooney_biegler/doe_example.py
```diff
@@ -0,0 +1,193 @@
...
+            print("Full Factorial Design Results:\n", results_container["results_dict"])
+
+    # Custom Plotting Functionality that shows the optimal designs and
+    # the max/min from the full factorial results
+    # Plotting Block
+    if plot_optimal_design:
+        plt = matplotlib.pyplot
+        res_dict = results_container["results_dict"]
+        opt_D = results_container["optimization"]["D"]
+        opt_A = results_container["optimization"]["A"]
```
blnicho: This will fail if either of `optimize_experiment_A` or `optimize_experiment_D` are set to `False`. You should either guard this code against different options sent to the `run_rooney_biegler_doe` function or remove the options and have this example always run DoE D/A optimization.

pyomo/contrib/parmest/examples/rooney_biegler/doe_example.py
smondal13: @blnicho I have moved it to `doe/examples` and renamed it.

pyomo/contrib/doe/doe.py
```diff
@@ -64,6 +64,7 @@
 class ObjectiveLib(Enum):
     determinant = "determinant"
```
adowling2: Let's add the following comments after each line

```
# det(FIM), D-optimality
# trace(inv(FIM)), A-optimalty
# trace(FIM), "psuedo"-A-optimality
# min(eig(FIM)), E-optimality
# cond(FIM), ME-optimality
# constant zero, useful for initialization and debugging

**Review by adowling2 — COMMENTED** [review]
Started to leave review comments. I will add more comments soon.

File: pyomo/contrib/doe/examples/rooney_biegler_doe_example.py [hunk 11]
```diff
@@ -0,0 +1,193 @@
...
+def run_rooney_biegler_doe(
+    optimize_experiment_A=False,
+    optimize_experiment_D=False,
+    compute_FIM_full_factorial=False,
+    draw_factorial_figure=False,
+    design_range={'hour': [0, 10, 40]},
+    plot_optimal_design=False,
+    tee=False,
+    print_output=False,
+):
```
adowling2, line 34: Please add a doc string explaining all of these function inputs (aka arguments).

pyomo/contrib/doe/examples/rooney_biegler_doe_example.py
```diff
@@ -0,0 +1,193 @@
+from pyomo.contrib.parmest.examples.rooney_biegler.rooney_biegler import (
+    RooneyBieglerExperiment,
+)
+from pyomo.contrib.doe import DesignOfExperiments
+from pyomo.common.dependencies import pandas as pd, numpy as np, matplotlib
+
+
```
adowling2, line 24: Please add a comment that gives the full citation for the source of this example and the data.

adowling2 (changes requested):
From a computational perspective, pseudo-A-optimality, i.e., trace(FIM), is the easiest objective to solve. I suggest we do a majority of the testing of permutations of DoE options with pseudo-A-optimality. I suggest that we then do a fewer number of tests with D-optimality (both Cholseky and GreyBox) and A-optimality (both Cholseky and GreyBox).

pyomo/contrib/doe/examples/rooney_biegler_doe_example.py
```diff
@@ -0,0 +1,193 @@
...
+    if optimize_experiment_D:
+        # D-Optimality
+        rooney_biegler_doe_D = DesignOfExperiments(
+            experiment=rooney_biegler_experiment,
+            objective_option="determinant",
+            tee=tee,
+            prior_FIM=FIM,
+            improve_cholesky_roundoff_error=True,
+        )
+        rooney_biegler_doe_D.run_doe()
```
blnicho: Why do you need both `rooney_biegler_doe` and `rooney_biegler_doe_D`? Seems like they are identical instances of the `DesignOfExperiments` class. Why not just call `run_doe` directly on `rooney_biegler_doe` when `optimize_experiment_D` is `True`?

pyomo/contrib/doe/examples/rooney_biegler_doe_example.py
```diff
@@ -0,0 +1,193 @@
...
+            print("Full Factorial Design Results:\n", results_container["results_dict"])
+
+    # Custom Plotting Functionality that shows the optimal designs and
+    # the max/min from the full factorial results
+    # Plotting Block
+    if plot_optimal_design:
+        plt = matplotlib.pyplot
+        res_dict = results_container["results_dict"]
+        opt_D = results_container["optimization"]["D"]
+        opt_A = results_container["optimization"]["A"]
```
blnicho: This will fail if either of `optimize_experiment_A` or `optimize_experiment_D` are set to `False`. You should either guard this code against different options sent to the `run_rooney_biegler_doe` function or remove the options and have this example always run DoE D/A optimization.

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -560,6 +567,215 @@ def test_doe_full_factorial(self):
...
+    def test_doe_A_optimality(self):
+        A_opt_value_expected = -2.2364242059539663
+        A_opt_design_value_expected = 9.999955457176451
+
+        A_opt_value = run_rooney_biegler_doe(optimize_experiment_A=True)[
+            "optimization"
+        ]["A"]["value"]
+        A_opt_design_value = run_rooney_biegler_doe(optimize_experiment_A=True)[
+            "optimization"
+        ]["A"]["design"][0]
```
blnicho: Calling `run_rooney_biegler_doe` twice here seems inefficient. Why not save the return value in a local variable? Then you can extract any of the values in the saved results

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -560,6 +567,215 @@ def test_doe_full_factorial(self):
...
+        self.assertAlmostEqual(A_opt_value, A_opt_value_expected, places=2)
+        print("A optimal design value:", A_opt_design_value)
+        self.assertAlmostEqual(
+            A_opt_design_value, A_opt_design_value_expected, places=2
+        )
+
+
+class TestRooneyBieglerExample(unittest.TestCase):
+    @unittest.skipUnless(pandas_available, "test requires pandas")
+    @unittest.skipUnless(matplotlib_available, "test requires matplotlib")
```
blnicho: The `run_rooney_biegler_doe` function has an option to plot or not. So why not have this test change the value of `plot_optimal_design` depending on the value of `matplotlib_available`? Then you can still test most of the example even when matplotlib isn't available.

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -560,6 +567,215 @@ def test_doe_full_factorial(self):
...
+    def test_rooney_bieglerdoe_example(self):
+        """
+        Tests the Design of Experiments (DoE) functionality, including
+        plotting logic, without displaying GUI windows.
+        """
+        plt = matplotlib.pyplot
+
+        # Switch backend to 'Agg' to prevent GUI windows from popping up
+        plt.switch_backend('Agg')
```
blnicho: Is this still needed? Didn't you set the backend to 'Agg` up on L28?

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -560,6 +567,215 @@ def test_doe_full_factorial(self):
...
+            compute_FIM_full_factorial=True,
+            draw_factorial_figure=True,
+            design_range={'hour': [0, 10, 3]},
+            plot_optimal_design=True,
+            tee=False,
+        )
+
+        # Assertions for Numerical Results
+        self.assertIn("D", results["optimization"])
+        self.assertIn("A", results["optimization"])
```
blnicho: You should add assertions that check for specific values otherwise this test isn't doing much besides confirming that the example runs without errors.

pyomo/contrib/doe/tests/test_doe_errors.py
```diff
@@ -808,6 +810,25 @@ def test_bad_compute_FIM_option(self):
         ):
             doe_obj.compute_FIM(method="Bad Method")
 
+    @unittest.skipIf(not ipopt_available, "The 'ipopt' command is not available")
+    def test_invalid_trace_without_cholesky(self):
+        fd_method = "central"
+        obj_used = "trace"
+
+        experiment = run_rooney_biegler_doe()["experiment"]
```
blnicho, line 818: Could you rework this to use `FullReactorExperiment` like the other tests in this file?

pyomo/contrib/doe/tests/test_doe_build.py
```diff
@@ -502,5 +503,119 @@ def test_reactor_update_suffix_items(self):
...
+        doe_obj = DesignOfExperiments(**DoE_args)
+
+        doe_obj.create_doe_model()
+        doe_obj.create_objective_function()
+
+        model = doe_obj.model
+        params = list(model.parameter_names)
+
+        # Check cov_trace initialization
+        cov_trace_from_fim_inv = sum(pyo.value(model.fim_inv[j, j]) for j in params)
```
blnicho: The expected value in a test should be a hard-coded number not a repeat of the implementation.

pyomo/contrib/doe/tests/test_doe_build.py
```diff
@@ -502,5 +503,119 @@ def test_reactor_update_suffix_items(self):
...
+        for i, c in enumerate(params):
+            for j, d in enumerate(params):
+                if i < j:
+                    continue  # upper triangle skipped by design
+
+                val = sum(
+                    pyo.value(model.L[c, params[k]])
+                    * pyo.value(model.L_inv[params[k], d])
+                    for k in range(len(params))
+                )
```
blnicho:
```suggestion
                val = pyo.value(sum(model.L[c, params[k]] * model.L_inv[params[k], d]
                    for k in range(len(params))
                ))
```
I'm pretty sure you can simplify this slightly by just calling `pyo.value` once. (Note: if you accept this change through the GitHub GUI the line indentation is off)

blnicho (changes requested):

pyomo/contrib/doe/examples/rooney_biegler_doe_example.py
smondal13: That was repeated. It's corrected now.

pyomo/contrib/doe/doe.py
smondal13: Added comments

pyomo/contrib/doe/examples/rooney_biegler_doe_example.py
smondal13: Added the citation.

pyomo/contrib/doe/examples/rooney_biegler_doe_example.py
smondal13: Added docstring.

pyomo/contrib/doe/examples/rooney_biegler_doe_example.py
smondal13: I have removed the plotting function. It is to visually confirm the result. It is not required.

pyomo/contrib/doe/tests/test_doe_build.py
smondal13: A hard-coded value is added.

pyomo/contrib/doe/tests/test_doe_build.py
smondal13: Added manually.

pyomo/contrib/doe/tests/test_doe_errors.py
smondal13: Since we talked about using inexpensive tests during the IDAES summit in Pittsburgh, I have added simpler `rooney_biegler` example here for the test instead of `FullReactorExperiment`, which is more complex and sometimes fails to converge.  Eventually, we want to make all tests inexpensive (or comparatively inexpensive) for `Pyomo.DoE`. However, it does not mean all tests that use `FullReactorExperiment` are expensive. I can add a test using `FullReactorExperiment`. Let me know whether I should add `FullReactorExperiment`.

pyomo/contrib/doe/tests/test_doe_solve.py
smondal13: Yeah, it was inefficient. I have saved the return as a variable and reused it.

pyomo/contrib/doe/tests/test_doe_solve.py
smondal13: Yes, I have done that.

pyomo/contrib/doe/tests/test_doe_solve.py
smondal13: Not needed, I have removed that

pyomo/contrib/doe/tests/test_doe_solve.py
smondal13: Assertions using hardcoded values are added.

pyomo/contrib/doe/examples/rooney_biegler_doe_example.py
smondal13: I agree. We can test most features when possible using `pseudo_A_optimality`. In this example, I have added `D-optimality` to check whether it gives us results similar to `A-optimality (trace(inv(FIM))`, and the example `rooney_biegler_doe_example.py` has an argument to activate or deactivate that. Although I have added a test for D-optimality in the `test_rooney_biegler_doe_example()` in `test_doe_solve.py`, the example itself is very lightweight and solves the full example in ~1 sec.

mrmundt:
@smondal13 - Gurobi just released a new version that supports Python 3.14, and we are seeing failures for our gurobi tests because of it. I cancelled your jobs because I have a fix already, just need to get it in.

TL;DR - if you see 3.14 failures in gurobi, not your fault, we're on it.

pyomo/contrib/doe/examples/reactor_example.py
```diff
@@ -101,7 +105,7 @@ def run_reactor_doe(
         )
     if plot_factorial_results:
         if save_plots:
-            figure_file_name = "example_reactor_compute_FIM"
+            figure_file_name = figure_file_name
```
adowling2: Is this needed anymore? Or could we remove `save_plots` as an option and interpret `figure_file_name = None` as indicating that figures should not be saved?

pyomo/contrib/doe/examples/rooney_biegler_doe_example.py
```diff
@@ -0,0 +1,176 @@
...
+    design_range={'hour': [0, 10, 40]},
+    tee=False,
+    print_output=False,
+):
+    """Run Rooney Biegler DoE Example. All the boolean options control whether
+    that section of the example is run. This is to facilitate testing using the example.
+
+    Parameters
+    ----------
+    optimize_experiment_A : bool, optional
```
adowling2: This does not seem very general... instead, do we want to support A, D, E, ME, pseudo A, ... objectives?

pyomo/contrib/doe/tests/test_doe_errors.py
adowling2: Can we do a lightweight version of `FullReactorExperiment`, perhaps with pseudo A opt

adowling2 (changes requested):
I have some concerns about how this testing framework can scale to all of the variants of Pyomo.DoE that we want to try, including:
- central finite difference versus symbolic differentiation
- GreyBox: A, D, E, ME optimality
- Cholesky: A, D

We do not need to add all of these variants in this PR. But I think we need to ensure this PR sets us up to easily add these extensions in other PRs. This is important because I could see multiple parallel PRs to add these features that will all be building on this testing foundation.

pyomo/contrib/doe/examples/reactor_example.py
smondal13: No, we do not need `save_plots`. I have removed that argument.

pyomo/contrib/doe/examples/rooney_biegler_doe_example.py
smondal13: Do you mean that we pass an argument, e.g., `objective="trace" / "pesudo_trace"... etc in `run_rooney_biegler_doe()` ? I think that is a very good idea.

pyomo/contrib/doe/tests/test_doe_errors.py
smondal13: @adowling2 @blnicho I have replaced all the tests in `test_doe_errors.py` with rooney_biegler experiment in #3828 . So, we are good here, right? 

Note: This branch is merged with #3828

pyomo/contrib/doe/examples/rooney_biegler_doe_example.py
adowling2: Yes, unless there are significant benefits to doing it as a single function (sequentially). If that is the case, you could make `objectives` the input and support either a single string or a list of strings.

smondal13: I have added the objective and the gradient/sensitivity computation option. Maybe symbolic can be added there when symbolic is merged. And solving the example for only one type of objective is fine. 

```
def run_rooney_biegler_doe(
    sensitivity_formula="central",
    optimization_objective="determinant",
    improve_cholesky_roundoff_error=False,
    compute_FIM_full_factorial=False,
    draw_factorial_figure=False,
    design_range={'hour': [0, 10, 40]},
    tee=False,
    print_output=False,
):
```

pyomo/contrib/doe/tests/test_doe_errors.py
adowling2: Sounds good.

adowling2 (approved):
I think this is ready to merge. Where are the changes to the documentation that explain A-optimality versus pseudo-A-optimality? Are they in this PR or a different PR?

adowling2:
@smondal13 Please also proof read the website version here: https://pyomo--3812.org.readthedocs.build/en/3812/explanation/analysis/doe/doe.html

adowling2:
@smondal13, I think we should update the workflow figure to match our tutorial material: https://dowlinglab.github.io/pyomo-doe/Readme.html

smondal13:
> @smondal13, I think we should update the workflow figure to match our tutorial material: https://dowlinglab.github.io/pyomo-doe/Readme.html

Added the workshop figure and proofread it. Had some minor corrections

pyomo/contrib/doe/examples/rooney_biegler_doe_example.py
```diff
@@ -0,0 +1,196 @@
...
+    )
+
+    # Run the optimization
+    rooney_biegler_doe.run_doe()
+
+    # Store results based on objective type
+    objective_name = optimization_objective.value
+    results_container["optimization"][objective_name] = {
+        "objective_type": str(optimization_objective).split('.')[-1],
+        "design": rooney_biegler_doe.results['Experiment Design'],
```
blnicho, line 136: Why are you saving the objective name twice in this `results_container` dictionary? Since you simplified the example to only run one type of DoE objective at a time, you can probably simplify this to something like:

```suggestion
    objective_name = optimization_objective.value
    results_container["optimization"] = {
        "objective_type": objective_name,
        "design": rooney_biegler_doe.results['Experiment Design'],
```

pyomo/contrib/doe/examples/rooney_biegler_doe_example.py
```diff
@@ -0,0 +1,196 @@
...
+    # Convert optimization_objective to enum if it's a string
+    if isinstance(optimization_objective, str):
+        optimization_objective = ObjectiveLib(optimization_objective)
+
+    # Initialize a container for all potential results
+    results_container = {
+        "experiment": None,
+        "results_dict": {},
+        "optimization": {},  # Will hold D/A optimization results if run
+        "plots": [],  # Store figure objects if created
```
blnicho: Since you removed some of the plotting code from the example, I don't think plots are being added to the `results_container` any more. This can be removed and the docstring describing the returned dictionary should be updated.

pyomo/contrib/doe/examples/rooney_biegler_doe_example.py
```diff
@@ -0,0 +1,196 @@
...
+from pyomo.common.dependencies import pandas as pd, numpy as np, matplotlib
+
+
+def run_rooney_biegler_doe(
+    sensitivity_formula="central",
+    optimization_objective="determinant",
+    improve_cholesky_roundoff_error=False,
+    compute_FIM_full_factorial=False,
+    draw_factorial_figure=False,
+    design_range={'hour': [0, 10, 40]},
```
blnicho: You should never use dictionaries as the default value for function arguments because it can lead to unexpected behavior. The dictionary is created once when the function is defined and not each time the function is called. The recommended practice is to set the default to `None` and then check for `None` in the function.
```suggestion
    design_range=None,
```

Then add this near the top of the function:
```
if design_range is None:
    design_range = {'hour': [0, 10, 40]}
```

pyomo/contrib/doe/examples/rooney_biegler_doe_example.py
```diff
@@ -0,0 +1,196 @@
...
+            print("Full Factorial Design Results:\n", results_container["results_dict"])
+
+    # Pyomo.DoE built-in factorial figure generation
+    if draw_factorial_figure:
+        rooney_biegler_doe.draw_factorial_figure(
+            sensitivity_design_variables=['hour'],
+            fixed_design_variables={},
+            log_scale=False,
+            figure_file_name="rooney_biegler",
+        )
```
blnicho, line 181: Image
You should double check the formatting of the plots being generated. I've attached an example of one of the plots I see when running this example locally. The title has an unexpected colon at the beginning and is missing an x-axis label.

blnicho (commented):
@smondal13 I think this is really close. I found a few more things in the example which should be straightforward to fix.

pyomo/contrib/doe/examples/rooney_biegler_doe_example.py
smondal13: Done

pyomo/contrib/doe/examples/rooney_biegler_doe_example.py
smondal13: Added labels and title

blnicho (approved):

PR #3804, opened 2025-12-18
Remove sorting from `tabular_writer`

jsiirola:
## Fixes # .

Supersedes #3578. 

## Summary/Motivation:
This removes all sorting from the `tabular_writer`, and prints out the data in the order in which the data iterator provides it.  Control for sorting int `pprint()` is now exposed through the `pprint()` API.  This makes the `tabular_writer` simpler / more generic and more usable in other contexts.

This is currently backwards compatible with the existing `pprint()` infrastructure and will produce the same output.  A subsequent PR will rework the `pprint()` design so that components are output in "ordered" (and not "sorted") order, as well as other changes that might break the output.

- remove sorting from tabular_writer
- update the `pprint()` private API so that sorting can be controlled by an optional argument to `pprint()`

### Legal Acknowledgement


pyomo/core/base/component.py
```diff
@@ -391,23 +399,37 @@ def _pprint_base_impl(
...
+                for _key, _val in _data:
+                    _fcn2(ostream, sort, _key, _val)
         elif _fcn is not None:
-            _data_dict = dict(_data)
-            for _key in sorted_robust(_data_dict):
-                _fcn(ostream, _key, _data_dict[_key])
+            # This is a non-standard component where we will not
+            # generate a table at all, and instead defer all formatting
+            # / printing to the callback.  This is primarily used by
+            # BLocks (and block-like things)
```
blnicho:
```suggestion
            # Blocks (and block-like things)
```

pyomo/core/base/param.py
```diff
@@ -466,23 +507,20 @@ def extract_values(self):
...
-            #
-            # The parameter is a scalar, so we need to create a temporary
-            # dictionary using the value for this parameter.
-            #
-            return {None: self()}
+            # The scalar could be defined (in which case items() will
+            # return the ScalarParam, OR it could be defined by a
+            # default value (in which case items() will return the
+            # actual numeric value.  to cover both cases we will use
+            # value():
```
blnicho, line 519:
```suggestion
            # The scalar could be defined (in which case items() will
            # return the ScalarParam), OR it could be defined by a
            # default value (in which case items() will return the
            # actual numeric value).  To cover both cases we will use
            # value():
```

pyomo/core/base/set.py
```diff
@@ -2022,6 +2006,29 @@ class _SortedSetData(metaclass=RenamedClass):
 _SETDATA_API = ('set_value', 'add', 'remove', 'discard', 'clear', 'update', 'pop')
 
 
+class _ScalarSetMixin:
+    # This mixin is required because scalar Sets implement __len__(),
+    # which raises an exception for non-finite sets.  Further finite
```
blnicho:
```suggestion
    # which raises an exception for non-finite sets. Further, finite
```

pyomo/core/base/set.py
```diff
@@ -2022,6 +2006,29 @@ class _SortedSetData(metaclass=RenamedClass):
...
+        # Scalar sets are always defined (we don't support Skip):
+        return iter(UnindexedComponent_set)
+
+    def values(self, sort=SortComponents.UNSORTED):
+        """Return an iterator of the component data objects in the dictionary"""
+        # Scalar sets are always defined (we don't support Skip):
+        yield self
+
+    def items(self, sort=SortComponents.UNSORTED):
+        """Return an iterator of (index,data) tuples from the dictionary"""
```
blnicho:
```suggestion
        """Return an iterator of (index, data) tuples from the dictionary"""
```

blnicho (approved):
I found a few minor typos which I will fix. Otherwise this looks good!

pyomo/common/formatting.py
```diff
@@ -154,8 +154,11 @@ def tabular_writer(ostream, prefix, data, header, row_generator):
             _rows[_key] = None
             continue
 
+        # Include the key for only the first line in a rowset, and only
+        # if we printed out a header (if there is no header, then then
```
mrmundt:
```suggestion
        # if we printed out a header (if there is no header, then the
```

pyomo/core/base/param.py
```diff
@@ -424,37 +425,77 @@ def mutable(self):
...
+            self._default_val = Param.NoValue
+            return self.items(sort)
+        finally:
+            self._default_val = tmp
 
+    @deprecated(
+        "The sparse_iterkeys method is deprecated.  Use sparse_keys()",
+        # This should have been deprecated when we dropped Python 2.7
+        version='6.10.0.dev0',
+    )
```
mrmundt, line 477: Should we not just... delete all of these then? Are they actually still in use anywhere?

mrmundt (approved):

pyomo/core/base/param.py
blnicho: We aren't using them but I'm in favor of the deprecation warning in case other downstream projects are using them.

mrmundt: Sounds good, happy with that.

PR #3811, opened 2026-01-09
GDP: Fix bug transforming Blocks in `gdp.mbigm` transformation

bammari:
## Fixes # .
The multiple big-M transformation fails to transform blocks.

## Summary/Motivation:
In many cases, we may want to apply the multiple big-M transformation on blocks rather than an entire model. 

For example, prior to this PR the following model would result in a failed transformation:

```
import pyomo.environ as pyo
from pyomo.gdp import Disjunction, Disjunct

m = pyo.ConcreteModel()
m.b = pyo.Block()
m.b.x = pyo.Var(bounds=(0, 5))
m.b.y = pyo.Var()
m.b.dis1 = Disjunct()
m.b.dis2 = Disjunct()

m.b.dis1.linear = pyo.Constraint(expr = m.b.x*0.5 + 3 == m.b.y)
m.b.dis2.linear = pyo.Constraint(expr = m.b.x*0.2 + 1 == m.b.y)
m.b.d = Disjunction(expr = [m.b.dis1, m.b.dis2])

pyo.TransformationFactory('gdp.mbigm').apply_to(m.b, threads=1)
```

- Add argument `relative_to=instance` on `constraint.getname()` and `other_disjunct.getname()` when setting up jobs for calculating M parameters.
- Add a test that applies the multiple big-M transformation on a block to ensure no failures occur.

### Legal Acknowledgement


pyomo/gdp/tests/test_mbigm.py
```diff
@@ -1191,3 +1192,24 @@ def test_complain_for_unrecognized_Suffix(self):
...
+        m.b.y = Var()
+        m.b.dis1 = Disjunct()
+        m.b.dis2 = Disjunct()
+
+        m.b.dis1.linear = Constraint(expr=m.b.x * 0.5 + 3 == m.b.y)
+        m.b.dis2.linear = Constraint(expr=m.b.x * 0.2 + 1 == m.b.y)
+        m.b.d = Disjunction(expr=[m.b.dis1, m.b.dis2])
+
+        TransformationFactory('gdp.mbigm').apply_to(m.b, threads=1)
+        self.assertEqual(m.b.find_component('dis1.linear').name, 'b.dis1.linear')
```
emma58: This really only tests that the transformation runs without error (which is progress, thank you!). But the assertion in line 1215 doesn't test anything about the transformed model. Could you use the tests above as an example and put in a couple assertions about the expressions of the transformed constraints?

emma58 (commented):
Thank you so much for fixing the bug, @bammari! Do you have time to add a few assertions about the transformed model to your test?

pyomo/gdp/tests/test_mbigm.py
bammari: @emma58  Thanks for the feedback! I went ahead and put in a couple of assertions that check whether the expressions in disjunct 1 are transformed correctly. The respective M values were verified independently and are correct.

pyomo/gdp/tests/test_mbigm.py
```diff
@@ -1191,3 +1192,35 @@ def test_complain_for_unrecognized_Suffix(self):
...
+        m = ConcreteModel()
+        m.b = Block()
+        m.b.x = Var(bounds=(0, 5))
+        m.b.y = Var()
+        m.b.dis1 = Disjunct()
+        m.b.dis2 = Disjunct()
+
+        m.b.dis1.linear = Constraint(expr=m.b.x * 0.5 + 3 == m.b.y)
+        m.b.dis2.linear = Constraint(expr=m.b.x * 0.2 + 1 == m.b.y)
+        m.b.d = Disjunction(expr=[m.b.dis1, m.b.dis2])
```
jsiirola, line 1211: Quick question: this tests the case where all the variables are defined by the subtree being transformed.  What happens if one of the Vars is defined outside the subtree (e.g., on `m` instead of under `m.b`)?

bammari: @jsiirola In this example, if we defined variable x as `m.x` instead of `m.b.x`, the transformed expression would be:

 `0.5 * m.x + 3 - m.b.y <= 3.5 * m.b.dis2.binary_indicator_var` instead of 

`0.5 * m.b.x + 3 - m.b.y <= 3.5 * m.b.dis2.binary_indicator_var`

Do you see any potential issues with this?

jsiirola: Nope - that should be fine.  I don't think it will be a huge issue, as the change this PR is making doesn't actually affect Vars (I wasn't thinking clearly last night -- you are only getting relative names for active components [Constraints & Disjuncts] - and not Vars).

BUT, this does raise the question: what do we expect the following to do??
```python
m.d1 = Disjunct()
m.b = Block()
m.b.d2 = Disjunct()
m.b.disj = Disjunction(expr=[m.d1, m.b.d2])
TransformationFactory('gdp.mbigm').apply_to(m.b, threads=2)
```
Thoughts? @emma58?

jsiirola: (Spot checking... this update to `mbigm` leads to an exception.  `bigm` will transform the model (including deactivating `m.d1`)

emma58: (@emma58 crawls under a table...)

(From under the table,) we expect the model above to put the transformed constraints on `m.b`, and to deactivate `m.d1` and `m.b.d2`. Scoping should be determined by the Disjunct*ion* and not care about where the Disjuncts are living at all. So basically, the `bigm` behavior is the desired behavior. I can look at this...

emma58: So actually, the model above transforms without error on this branch (both with `m` and `m.b` as the Block being transformed). Even this transforms correctly, as a fun fact:
```
from pyomo.environ import *
from pyomo.gdp import *

m = ConcreteModel()

m.x = Var(bounds=(0, 10))
m.d1 = Disjunct()
m.b = Block()
m.b.y = Var(bounds=(3, 15))
m.b2 = Block()
m.b2.d2 = Disjunct()
m.b2.d2.c1 = Constraint(expr=m.x <= 7)
m.b2.d2.c2 = Constraint(expr=m.b.y <= 7)

m.b.disj = Disjunction(expr=[m.d1, m.b2.d2])
TransformationFactory('gdp.mbigm').apply_to(m.b, threads=2)
```
We're somehow getting away with the relative names, though I admit I'm not fully following how at the moment.

emma58 (approved):

jsiirola (approved):
I think we've convinced ourselves that this is OK.

PR #3821, opened 2026-01-20
Factorable Programming for PWL Approximations

michaelbynum:
## Summary/Motivation:

The purpose of this module/transformation is to convert any nonlinear model 
to the following form:

```
min/max e(x_i)*f(x_j) + b^T*x
s.t.
        g_j(x_i)*h_j(x_k) + a_j^T*x >/</== 0
        g_j(x_i)/h_j(x_k) + a_j^T*x >/</== 0
        g_j(x_i)**h_j(x_k) + a_j^T*x >/</== 0
```

By doing so, each nonlinear function is only a function of one or two variables. 
If this transformation is used prior to the nonlinear_to_pwl transformation, 
it can, in some cases, significantly reduce the complexity of the PWL approximation.

### Legal Acknowledgement


pyomo/contrib/piecewise/tests/test_univariate_nonlinear_decomposition.py
```diff
@@ -0,0 +1,169 @@
+from pyomo.common.unittest import TestCase, skipUnless
```
jsiirola, line 10: Please add the copyright statement.

pyomo/contrib/piecewise/tests/test_univariate_nonlinear_decomposition.py
```diff
@@ -0,0 +1,169 @@
...
+import pyomo.environ as pyo
+from pyomo.contrib import piecewise
+from pyomo.core.expr.compare import assertExpressionsEqual
+from pyomo.common.dependencies import numpy_available, numpy
+from pyomo.core.expr.numeric_expr import ProductExpression
+
+pe = pyo
```
jsiirola: This is nonstandard, per the current project guidelines.

pyomo/contrib/piecewise/transform/factorable.py
```diff
@@ -0,0 +1,489 @@
+from pyomo.core.expr.visitor import StreamBasedExpressionVisitor
```
jsiirola, line 10: Please add the copyright statement.

pyomo/contrib/piecewise/transform/factorable.py
```diff
@@ -0,0 +1,489 @@
...
+            The reason to do this is to be able to bound z1 and z2.
+            """,
+        ),
+    )
+
+    def __init__(self):
+        super().__init__()
+        self._config = UnivariateNonlinearDecompositionTransformation.CONFIG()
+
+    def _check_for_unknown_active_components(self, model):
```
jsiirola, line 451: There is a standard/robust implementation of this as `pyomo.repn.util.categorize_valid_components`

pyomo/contrib/piecewise/transform/factorable.py
```diff
@@ -0,0 +1,489 @@
...
+        visitor.degree_map[res] = -1
+    visitor.substitution_map[node] = res
+    return res
+
+
+def _handle_named_expression(node, data, visitor):
+    assert len(data) == 1
+    res = data[0]
+    visitor.substitution_map[node] = res
+    return res
```
jsiirola: This looks like it has the effect of removing the Named Expression from the expression tree.  That will have implications on the efficiency of the ultimate solve (the NL writer exploits named expressions when writing NL files and the ASL exploits them when evaluating the model constraints / derivatives).

I would recommend leaving the Expression in the expression tree and just update the expression stored in it.

jsiirola (changes requested):
Some minor comments, but really a pretty big design question.

I think I will advocate refactoring this to avoid the `node_to_var_map` and `degree_map` entirely and restrict/repurpose the use of the `substitution_map` to only managing `Expression` objects (and maybe caching `1/x`?).  The issue is that as written, those maps will end up with one entry for every node of every expression of the entire model.  Even for modest models, that will be *huge*.  The other reason to not make the dict is that the only time things will be looked up in the maps will be when processing the parent node of the node in context.  I think a simpler / more efficient approach will be to return a tuple of `(node, degree, varset)` from `exitNode`.

I am still working through verifying the logic in the addition/product/division/pow handlers.

michaelbynum:
Thanks for the review, @jsiirola. I agree with most of your feedback. However, I am going to push back on the refactor for the following reasons:

1. I have a dream where we create a separate transformation that actually identifies common subexpressions. The current implementation would be able to exploit that. I would really like to avoid creating separate PWL approximations for x*y in one constraint and y*x in another constraint.
2. You did motivate me to do some profiling. The first time I profiled this, the time spent in FBBT was significant. I was able to cut the total time in the transformation by a factor of 2 by switching to `compute_bounds_on_expr` (I was originally being lazy - we don't actually need FBBT in most cases here). I'll push these changes shortly. After that fix, it does look like some improvement could be made by moving away from `ComponentMap`. However, I think there are much bigger fish to fry that I would prefer to focus on first. 
3. These are implementation details that we can change at any time without impacting users, so there is no pressing reason to make these changes now. 

I'll push fixes for your other comments shortly.

pyomo/contrib/piecewise/tests/test_univariate_nonlinear_decomposition.py
michaelbynum: Done.

pyomo/contrib/piecewise/tests/test_univariate_nonlinear_decomposition.py
michaelbynum: Done.

pyomo/contrib/piecewise/transform/factorable.py
michaelbynum: Done.

pyomo/contrib/piecewise/transform/factorable.py
michaelbynum: Done.

pyomo/contrib/piecewise/transform/factorable.py
michaelbynum: Done.

pyomo/contrib/piecewise/transform/factorable.py
```diff
@@ -0,0 +1,512 @@
...
+        setattr(model, bname, Block())
+        block = getattr(model, bname)
+        visitor = _UnivariateNonlinearDecompositionVisitor(
+            aux_block=block,
+            aggressive_substitution=self._config.aggressive_substitution,
+        )
+
+        for con in constraints:
+            lower, body, upper = con.to_bounded_expression(evaluate_bounds=True)
+            new_body = visitor.walk_expression(body)
```
jsiirola, line 487: This makes sense, as the walker *always* rebuilds the expression, even if nothing is being substituted out.  In the future, we should consider revisiting this design and borrow ideas from the `ExpressionReplacementVisitor` and only recreate the (parts of the) expressions that we have to in order to effect the substitutions.

pyomo/contrib/piecewise/transform/factorable.py
```diff
@@ -0,0 +1,512 @@
...
+            if lower == upper:
```
jsiirola: This can be true for unbounded expressions.  Is that a concern?

```suggestion
            if lower is not None and lower == upper:
```

pyomo/contrib/piecewise/transform/factorable.py
```diff
@@ -0,0 +1,512 @@
...
+        arg1_nvars = 1
+        arg1_degree = 1
+    if arg2_nvars > 1 or visitor.aggressive_substitution:
+        arg2 = visitor.create_aux_var(arg2)
+        arg2_vars = (arg2,)
+        arg2_nvars = 1
+        arg2_degree = 1
+    res = arg1 * arg2
+    # at this point arg1 should have at most 1 variable
+    # and arg2 should have at most 1 variable
```
jsiirola: Is this right?  If either `arg1` or `arg2` have no variables, then we have already returned above.  I believe that at this point we are guaranteed that both arg1 and arg2 have exactly 1 variable each.  The following two if bodies are uncovered, and I believe are unreachable.

pyomo/contrib/piecewise/transform/factorable.py
```diff
@@ -0,0 +1,512 @@
...
+        x = arg1_vars[0]
+        y = arg2_vars[0]
+        if x is y:
+            visitor.node_to_var_map[res] = (x,)
+        else:
+            visitor.node_to_var_map[res] = (x, y)
+    if arg1_degree == 0:
+        visitor.degree_map[res] = arg2_degree
+    elif arg2_degree == 0:
+        visitor.degree_map[res] = arg1_degree
```
jsiirola: This code should be unreachable...

pyomo/contrib/piecewise/transform/factorable.py
```diff
@@ -0,0 +1,512 @@
...
+        visitor.node_to_var_map[res] = arg2_vars
+    elif arg2_nvars == 0:
+        visitor.node_to_var_map[res] = arg1_vars
+    else:
+        x = arg1_vars[0]
+        y = arg2_vars[0]
+        if x is y:
+            visitor.node_to_var_map[res] = (x,)
+        else:
+            visitor.node_to_var_map[res] = (x, y)
```
jsiirola: This gets repeated a lot. Should it be a helper function?

pyomo/contrib/piecewise/transform/factorable.py
```diff
@@ -0,0 +1,512 @@
...
+        arg1 = visitor.create_aux_var(arg1)
+        arg1_vars = (arg1,)
+        arg1_nvars = 1
+        arg1_degree = 1
+
+    if arg2_nvars > 1 or visitor.aggressive_substitution:
+        arg2 = visitor.create_aux_var(arg2)
+        arg2_vars = (arg2,)
+        arg2_nvars = 1
+        arg2_degree = 1
```
jsiirola, line 259: This gets repeated a lot.  Should it be a helper function?

jsiirola (approved):
OK.  I don't think I found any real problems with this code.  I still feel strongly that the handlers should return tuples of `(node, degree, varlist)` and not rely on huge dictionaries to pass that information up the tree, but that can be implemented in a subsequent PR.

If you have a chance, it would be nice to resolve the unreachable code in the product handler before merging.

pyomo/contrib/piecewise/transform/factorable.py
michaelbynum: Good catch! Fixed.

pyomo/contrib/piecewise/transform/factorable.py
michaelbynum: Good catch! Fixed.

pyomo/contrib/piecewise/transform/factorable.py
michaelbynum: Good catch! Fixed.

pyomo/contrib/piecewise/transform/factorable.py
michaelbynum: Done.

pyomo/contrib/piecewise/transform/factorable.py
michaelbynum: I'm not sure there is anything for a helper function to do here? This is really just creating local variables. All of the "work" is already in `visitor.create_aux_var`.

PR #3828, opened 2026-01-28
[Pyomo.DoE] Replace computationally expensive test example with Rooney Biegler

smondal13:
## Summary/Motivation:
Most unit tests within `pyomo/contrib/doe/tests` utilized the `reactor_example.py` example, which is computationally expensive for the scope of these tests. This resulted in unnecessarily long execution times for the test suite.

This PR replaces the `reactor_example.py` with the simpler `rooney_biegler_doe_example.py`  example in the tests. This change maintains the validity of the tests while significantly reducing computational overhead.

Performance Impact (MacBook with M4 chip):

Before: ~20 seconds

After: ~12 seconds

Improvement: ~40% reduction in test execution time.

- Replaced usage of `reactor_example.py` with `rooney_biegler_doe_example.py` in pyomo/contrib/doe/tests.
- Updated relevant assertions to match the `rooney_biegler_doe_example.py`  model structure.
- Reduced the number of points calculated for sensitivity plotting.

## Note:
This PR relies on PR #3803 

### Legal Acknowledgement


smondal13:
@blnicho @adowling2 This PR is ready for initial design review. 

@blnicho, one of the tests is failing. Can you please take a look? 

**Note:** This PR depends on #3803

adowling2:
@sscini I recommend you take a look at this. It is related to some of your other PRs.

pyomo/contrib/doe/tests/experiment_class_example_flags.py
```diff
@@ -8,255 +8,70 @@
...
+            delattr(m, 'unknown_parameters')
+        elif flag == 5:
+            # Create mismatch: 1 experiment output but multiple measurement errors
+            # This tests the validation that checks output/error length match
+            if hasattr(m, 'measurement_error'):
+                import pyomo.environ as pyo
+
+                # Add a fake component to measurement_error to create length mismatch
+                m.fake_output = pyo.Var()
+                m.measurement_error[m.fake_output] = 0.1
```
blnicho: Use `del_component` to remove components from a Pyomo model. Also, this is an experiment model that we have control over so I don't think we need to check `hasattr` for expected suffixes. We want the tests using this flagged experiment class to fail if the suffix requirements are ever changed.

```suggestion
        if flag == 1:
            m.del_component(m.experiment_outputs)
        elif flag == 2:
            m.del_component(m.measurement_error)
        elif flag == 3:
            m.del_component(m.experiment_inputs)
        elif flag == 4:
            m.del_component(m.unknown_parameters)
        elif flag == 5:
            # Create mismatch: 1 experiment output but multiple measurement errors
            # This tests the validation that checks output/error length match
            # Add a fake component to measurement_error to create length mismatch
            m.fake_output = pyo.Var()
            m.measurement_error[m.fake_output] = 0.1
```

pyomo/contrib/doe/tests/test_doe_errors.py
```diff
@@ -30,20 +27,51 @@
...
+        flag: Model completeness flag (0=full, 1-4=missing specific suffix)
+    """
+    data = pd.DataFrame(data=[[5, 15.6]], columns=['hour', 'y'])
+    data_point = data.iloc[0]
+
+    return RooneyBieglerExperimentFlag(
+        data=data_point,
+        theta={'asymptote': 15, 'rate_constant': 0.5},
+        measure_error=0.1,
+    )
```
blnicho: These are pretty much identical. I think you can remove `get_rooney_biegler_experiment` and just use `get_rooney_biegler_experiment_flag` for everything. You also don't need the `flag` argument here because it isn't used until you ask for the labeled model.

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -61,6 +64,28 @@
...
+    """Get a fresh RooneyBieglerExperiment instance for testing.
+
+    Creates a new experiment instance to ensure test isolation.
+    Each test gets its own instance to avoid state sharing.
+    """
+    return RooneyBieglerExperiment(
+        data=get_rooney_biegler_data(),
+        theta={'asymptote': 15, 'rate_constant': 0.5},
+        measure_error=0.1,
+    )
```
blnicho, line 86: Why are you breaking these into two different functions when you have combined them in the other test files?

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -167,14 +193,29 @@ def test_reactor_fd_central_solve(self):
         # Make sure FIM and Q.T @ sigma_inv @ Q are close (alternate definition of FIM)
         self.assertTrue(np.all(np.isclose(FIM, Q.T @ sigma_inv @ Q)))
 
+    @unittest.skipIf(not pandas_available, "pandas is not available")
     def test_reactor_fd_forward_solve(self):
```
blnicho: You didn't update all of the test names in this file to not reference `reactor`

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -234,14 +279,22 @@ def test_reactor_obj_det_solve(self):
 
         self.assertEqual(doe_obj.results["Solver Status"], "ok")
```
blnicho, line 280: This is a weak test. It's only checking the solver status and not for any expected results.

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -292,11 +345,13 @@ def DISABLE_test_reactor_obj_cholesky_solve_bad_prior(self):
 
     # This test ensure that compute FIM runs without error using the
     # `sequential` option with central finite differences
```
blnicho, line 351: The next few tests that are checking that `compute_FIM` runs without error could be improved by adding assert statements that check for values or model/expression structure.

blnicho (changes requested):
I found a few spots where the code could be cleaned up and have some suggestions for making some of the tests stronger.

pyomo/contrib/doe/tests/test_doe_solve.py
smondal13: In this script, apart from `RooneyBieglerExperiment()` we have `RooneyBieglerExperimentBad()` in L493, which needs the data. That is why I have spareted them in this script

pyomo/contrib/doe/tests/test_doe_solve.py
smondal13: Updated the names when `RooneyBiegler` is used.

pyomo/contrib/doe/tests/test_doe_solve.py
smondal13: Added hardcode value.

pyomo/contrib/doe/tests/test_doe_solve.py
smondal13: I have added hardcoded values for `compute_FIM` for `central` finite difference and `kaug`. However, for `backward` and `forward` finite difference the output is not reliable. I can add values for them, but those values are numerically incorrect. We can probably add a structure/expression test for this in `test_doe_build.py` in a future PR. 

Note: `central` finite difference is the default `fd_method` in `DesignOfExperiments`.

pyomo/contrib/doe/tests/test_doe_errors.py
blnicho: @smondal13 I think this is the only comment that still needs to be addressed and then this can be merged.

blnicho (approved):

PR #3830, opened 2026-02-02
Improvements to LDSDA

bernalde:
## Summary/Motivation:
- A bug in ldsda.py was found related to unpacking a tuple
- Code coverage of ldsda.py was not 100%
- Documentation and formatting were lacking in this file
- An extra dependency (tabulate) was unnecessarily added to the code

- Modified line_search in pyomo/contrib/gdpopt/ldsda.py to unpack (primal_improved, primal_bound).
- Added unit tests in pyomo/contrib/gdpopt/tests/test_ldsda.py to verify the logic handles tuple returns correctly without infinite loops and raises coverage of ldsda.py to 100%
- Make behavior consistent with the docstring for _solve_GDP_subproblem: ensure primal_bound is None when the subproblem solve does not succeed (not only when preprocessing raises InfeasibleConstraintException).
- Avoid adding an unnecessary global dependency: remove use of the external `tabulate` package from ldsda and use simple logger output instead.

### Legal Acknowledgement


blnicho:
@bernalde could you give the Pyomo team edit permissions on this PR? Should be a checkbox at the bottom of the right side information summary about "Allowing maintainers to edit this pull request".

bernalde:
Oh no, since this is coming from the SECQUOIA fork I can’t do that https://github.com/orgs/community/discussions/5634
Should I open a PR from my personal account?

pyomo/contrib/gdpopt/tests/test_ldsda.py
```diff
@@ -57,5 +70,236 @@ def test_solve_four_stage_dynamic_model(self):
...
+        config = MagicMock()
+        solver.line_search(config)
+
+        # 5. Verify results
+        # The solver should have moved exactly ONCE (from 0,0 to 1,1)
+        self.assertEqual(solver.current_point, (1, 1))
+
+        # The solver should have attempted to solve exactly TWICE
+        # (Once for the success, once for the failure that stops the loop)
+        self.assertEqual(solver._solve_GDP_subproblem.call_count, 2)
```
emma58, line 107: This doesn't need to hold up this PR, but just a thing to keep in mind in general: You might consider trying to write tests that don't rely on calling private methods and accessing private attributes. I say this having shot myself in the foot this way before: If you ever do a major rewrite you will almost surely break all the tests, which is not just a lot of work, it also means you will be less sure you didn't break functionality. If you don't think that will be true, then you might think about if some of these things should actually be public.

pyomo/contrib/gdpopt/tests/test_ldsda.py
```diff
@@ -57,5 +70,236 @@ def test_solve_four_stage_dynamic_model(self):
...
+        # 3. FIX: Manually set the attribute that _get_external_information would have set
+        self.solver.number_of_external_variables = 1
+        self.config.starting_point = [0]
+
+        # 4. Run the method
+        self.solver._solve_gdp(self.model, self.config)
+
+        # 5. Verify we hit the break (neighbor_search was skipped)
+        self.solver.neighbor_search.assert_not_called()
+        self.solver.any_termination_criterion_met.assert_called()
```
emma58, line 154: Are these missing the test assertion? Like:
```
self.assertTrue(self.solver.neighbor_search.assert_not_called())
```

pyomo/contrib/gdpopt/tests/test_ldsda.py
```diff
@@ -57,5 +70,236 @@ def test_solve_four_stage_dynamic_model(self):
...
+        self.model.b = BooleanVar()
+        self.model.lc = LogicalConstraint(expr=self.model.b.implies(True))
+
+        self.config.logical_constraint_list = [self.model.lc]
+        self.model.util_block.config_logical_constraint_list = [self.model.lc]
+
+        with self.assertRaisesRegex(
+            ValueError, "should be a list of ExactlyExpression"
+        ):
+            self.solver._get_external_information(self.model.util_block, self.config)
```
emma58: Same question about testing private methods here (and in the tests following as well.)

pyomo/contrib/gdpopt/tests/test_ldsda.py
```diff
@@ -57,5 +70,236 @@ def test_solve_four_stage_dynamic_model(self):
...
+        (e.g., optimal, feasible).
+        """
+        # 1. Create a mock result object
+        mock_result = MagicMock()
+
+        # 2. Set termination condition to something NOT in the 'success' set
+        # The code checks for: optimal, feasible, globallyOptimal, locallyOptimal,
+        # maxTimeLimit, maxIterations, maxEvaluations.
+        # We use 'infeasible' or 'error' to trigger the False return.
+        from pyomo.opt import TerminationCondition as tc
```
emma58: Can you move this import to the top of the file?

pyomo/contrib/gdpopt/ldsda.py
```diff
@@ -60,14 +57,21 @@
...
 
-    Accepts models that can include nonlinear, continuous variables and
-    constraints, as well as logical conditions.
+    References
+    ----------
+    Ovalle, D.; Liñán, D. A.; Lee, A.; Gómez, J. M.; Ricardez-Sandoval, L.; Grossmann, I. E.; Bernal Neira, D. E. Logic-Based Discrete-Steepest Descent: A Solution Method for Process Synthesis Generalized Disjunctive Programs. Computers & Chemical Engineering 2025, 195, 108993. https://doi.org/10.1016/j.compchemeng.2024.108993.
     """
 
-    CONFIG = _GDPoptAlgorithm.CONFIG()
+    CONFIG = _GDPoptAlgorithm.CONFIG()  # CONFIG = ConfigBlock("GDPopt")
```
emma58: Did you intend to leave this comment?

pyomo/contrib/gdpopt/ldsda.py
```diff
@@ -334,21 +385,26 @@ def _get_directions(self, dimension, config):
             return directions
         elif config.direction_norm == 'Linf':
             directions = list(it.product([-1, 0, 1], repeat=dimension))
```
emma58: Also doesn't need to hold up this PR, but we should probably turn the norm into an Enum at some point rather than using strings.

pyomo/contrib/gdpopt/ldsda.py
```diff
@@ -421,47 +488,59 @@ def neighbor_search(self, config):
...
-            the values of the external variables
+            The subproblem model instance.
+        external_var_value : tuple
+            The external variable configuration used for this subproblem.
         config : ConfigBlock
-            GDPopt configuration block
+            The configuration block.
         search_type : str
-            the type of search, neighbor search or line search
+            The type of search ('Neighbor search', etc.).
```
emma58: Same comment here: Eventually it would be good to make this an Enum.

emma58 (commented):
Thank you for fixing the bug and adding so many tests! I put several comments in, but nothing needs to hold up the PR other than that I think you are missing assertions in one test, and there's on import that is inside of a test. The rest is all just preference and some design ideas that are totally optional.

blnicho:
> Oh no, since this is coming from the SECQUOIA fork I can’t do that https://github.com/orgs/community/discussions/5634 Should I open a PR from my personal account?

Ok, no problem. I didn't know about that limitation. We'll work with this PR, no need to reopen it from a personal fork.

pyomo/contrib/gdpopt/tests/test_ldsda.py
bernalde: This is great input. I went ahead and implemented your recommendations. Thank you for the lightning-fast review!

pyomo/contrib/gdpopt/tests/test_ldsda.py
bernalde: Yes, solved

pyomo/contrib/gdpopt/tests/test_ldsda.py
bernalde: Sure

pyomo/contrib/gdpopt/ldsda.py
bernalde: Oops

pyomo/contrib/gdpopt/ldsda.py
bernalde: Good point, I left some backward compatibility str way of specifying this, but this is good practice

pyomo/contrib/gdpopt/ldsda.py
bernalde: Already did in this one!

pyomo/contrib/gdpopt/tests/test_ldsda.py
bernalde: The current implementation self.solver.neighbor_search.assert_not_called() is correct. The assert_not_called() method from unittest.mock directly raises an AssertionError if the mock was called, and returns None on success. Wrapping it in assertTrue() would actually break the test since assertTrue(None) evaluates to False.

emma58: Oh, whoops, my bad. I don't know much about unittest.mock, clearly!

pyomo/contrib/gdpopt/ldsda.py
```diff
@@ -312,43 +401,51 @@ def fix_disjunctions_with_external_var(self, external_var_values_list):
...
+        Generate the search directions for the given dimension.
 
         Parameters
         ----------
         dimension : int
-            Dimension of the neighborhood
+            The dimensionality of the neighborhood (number of external variables).
         config : ConfigBlock
-            GDPopt configuration block
+            The configuration block specifying the norm ('L2' or 'Linf').
```
emma58, line 410: You probably want to update this docstring to reference your new `DirectionNorm` enum.

pyomo/contrib/gdpopt/ldsda.py
```diff
@@ -312,43 +401,51 @@ def fix_disjunctions_with_external_var(self, external_var_values_list):
...
         -------
-        list
-            the search directions.
+        list of tuple
+            A list of direction vectors (tuples).
+            - If DirectionNorm.L2: Standard basis vectors and their negatives.
+            - If DirectionNorm.Linf: All combinations of {-1, 0, 1} excluding the zero vector.
         """
-        if config.direction_norm == 'L2':
+        if config.direction_norm in (DirectionNorm.L2, 'L2'):
```
emma58: Since you set the value of the enum to be the string, the old version should work in either case.

pyomo/contrib/gdpopt/ldsda.py
```diff
@@ -312,43 +401,51 @@ def fix_disjunctions_with_external_var(self, external_var_values_list):
...
             directions = []
             for i in range(dimension):
                 directions.append(tuple([0] * i + [1] + [0] * (dimension - i - 1)))
                 directions.append(tuple([0] * i + [-1] + [0] * (dimension - i - 1)))
             return directions
-        elif config.direction_norm == 'Linf':
+        elif config.direction_norm in (DirectionNorm.Linf, 'Linf'):
```
emma58: Same comment here: Since you set the value of the enum to be the string, the old version should work in either case.

emma58 (approved):
A couple comments on the change to enums, but otherwise this looks good to me!

blnicho:
@bernalde or @AlbertLee125 could you reply to Emma's review comments or mark them resolved if there isn't anything to be done?

pyomo/contrib/gdpopt/ldsda.py
AlbertLee125: My apologies for the late response.  It is a good point. Since `DirectionNorm` is a str-Enum, the legacy == `'L2'`/`'Linf'` checks would work with either enum or string inputs. The explicit tuple membership check was retained mainly to make it visually clear that both `DirectionNorm` values and legacy string inputs are accepted. I am happy to simplify it if you'd prefer.

AlbertLee125:
My apologies for the late response.  It is a good point. Since `DirectionNorm` is a str-Enum, the legacy == `'L2'`/`'Linf'` checks would work with either enum or string inputs. The explicit tuple membership check was retained mainly to make it visually clear that both `DirectionNorm` values and legacy string inputs are accepted. I am happy to simplify it if you'd prefer.

pyomo/contrib/gdpopt/ldsda.py
emma58: I personally think simplifying it is better, and that you should encourage users to use the enum rather than relying on string arguments. We get burned relying on strings a lot and have been trying to move away from it. But I'm not opposed to this going in as is if that's your preference.

emma58: (Though I do think you may as well make the reversions in the comments below to just check the equality like before.)

pyomo/contrib/gdpopt/ldsda.py
AlbertLee125: Since `DirectionNorm` is a `str`-Enum, another clean option would be to normalize once using `norm = str(config.direction_norm)` and then compare against the string values. This keeps compatibility with legacy string inputs while avoiding redundant membership checks.
Something like,
```python
norm = str(config.direction_norm)
if norm == 'L2':
    ...
elif norm == 'Linf':
    ...

```

pyomo/contrib/gdpopt/ldsda.py
AlbertLee125: If normalizing the `DirectionNorm` using `norm = str(config.direction_norm)` is a good idea, I will implement it.

pyomo/contrib/gdpopt/ldsda.py
emma58: You don't need to cast it to a string--it's already a string. You can just make this line:
```
if config.direction_norm == DirectionNorm.L2
```
(This works because `DirectionNorm.L2 == 'L2'`)

emma58: (Sorry, just realized my original comment was confusing. What I meant is above.)

AlbertLee125:
I applied the suggested change. Backward compatibility remains intact. Thank you very much for your kind instruction.

pyomo/contrib/gdpopt/tests/test_ldsda.py
```diff
@@ -1,25 +1,13 @@
...
-    SolverFactory,
-    value,
-    Var,
-    Constraint,
-    TransformationFactory,
-    ConcreteModel,
-    BooleanVar,
-    LogicalConstraint,
-    Block,
-)
```
blnicho: @AlbertLee125 this looks like a bad merge. You lost some of the imports required for this test.

AlbertLee125: Thanks for pointing that out. It was a merge issue on my side. I restored the missing imports.

blnicho (approved):

blnicho:
Jenkins failure is unrelated (it's an intermittent cuOpt failure). This is safe to merge.

PR #3844, opened 2026-02-11
SuffixFinder: look in context for "external" data objects

### LINKED ISSUE #3800: nl writer does not apply scaling factors on temporary block
## Summary

When we create a temporary block and then solve using `ipopt_v2` with `scale_model=True`, it is known that scaling factors defined in the original model are not copied over or used. However, the solver also appears to ignore scaling factors that are manually defined on the temporary block.

```
# temporary_block_not_scaled.py
import pyomo.environ as pyo
from pyomo.util.subsystems import create_subsystem_block
# import idaes to allow use of ipopt
import idaes

m = pyo.ConcreteModel()
m.x = pyo.Var(initialize=1e-4, bounds=(0, None))
m.con = pyo.Constraint(expr=m.x ** 2 == 1+1e-12)
t_block = create_subsystem_block([m.con], [m.x])
t_block.scaling_factor = pyo.Suffix(direction=pyo.Suffix.EXPORT)
t_block.scaling_factor[m.x] = 1e4

solver_obj = pyo.SolverFactory("ipopt_v2")
solver_obj.config.writer_config.scale_model = True
solver_obj.options.max_iter = 0

print(f"x value: {m.x.value}")
solver_obj.solve(t_block)
# Scaling factor of 1e4 means the value shouldn't be pushed
print(f"x value: {m.x.value}")
```

The default IPOPT configuration has `bound_push=1e-2`. Since we're using a scaling factor of `1e4` for `m.x`, its initial value of `1e-4` should not be changed---the scaled value of `1` is further than the bound push of `1e-2`. However, `m.x` has its value pushed to `1e-2`, which indicates the unscaled value of `x` is getting passed to the solver.

```console
x value: 0.0001
x value: 0.00999999
```

Pyomo version: 6.9.5
Python version: 3.12
Operating system: Windows 11
Solver (if applicable): ipopt_v2

jsiirola:
## Summary/Motivation:
This updates the SuffixFinder to always look in the Suffix attached to the `context` (root) block for Components / ComponentDatas that are not part of the `context` model hierarchy.  

This resolves issues finding Suffixes for components that are part of References (i.e., they *look* like they are part of this tree, but are actually part of another tree elsewhere), and fixes #3800.

- SuffixFinder: When looking up components attached to blocks outside of `context`, always check the Suffix on `context` before giving up

### Legal Acknowledgement


pyomo/core/base/suffix.py
```diff
@@ -459,18 +459,31 @@ def __init__(self, name, default=None, context=None):
...
-           starting at the `root` and descending down the tree to
-           the component_data.  Return the first match found.
-        2. Search for a Suffix matching the component_data's container,
-           starting at the `root` and descending down the tree to
-           the component_data.  Return the first match found.
-        3. Search for a Suffix with key `None`, starting from the
-           component_data and working up the tree to the `root`.
+        If the `component_data` is attached to a block within the
+        `context` (that is, part of the model hierarchy rooted by the
+        `context` :py:`BlockData`), then :py:`Suffixe` components are
```
blnicho:
```suggestion
        `context` :py:`BlockData`), then :py:`Suffix` components are
```

dallan-keylogic:
I can confirm this fixes the example in #3800. However, if a suffix is on the original model, the `.nl` reader ignores it---it looks for suffixes only on the temporary block.

```
# temporary_block_not_scaled.py
import pyomo.environ as pyo
from pyomo.util.subsystems import create_subsystem_block
# import idaes to allow use of ipopt
import idaes

m = pyo.ConcreteModel()
m.x = pyo.Var(initialize=1e-4, bounds=(0, None))
m.con = pyo.Constraint(expr=m.x ** 2 == 1+1e-12)
m.scaling_factor[m.x] = 1e4

t_block = create_subsystem_block([m.con], [m.x])

solver_obj = pyo.SolverFactory("ipopt_v2")
solver_obj.config.writer_config.scale_model = True
solver_obj.options.max_iter = 0

print(f"x value: {m.x.value}")
solver_obj.solve(t_block)
# Value of 0.00999999 ~= 0.01 means that scaling suffix on m was ignored
print(f"x value: {m.x.value}")
```

This is a defensible design decision---if the user wants scaling on the temporary block, they should copy the scaling factors over themselves. (Indeed, this is what we do currently [for PETSc in IDAES.](https://github.com/IDAES/idaes-pse/blob/da90c69ce114d68be3294e271088ca196d3210c2/idaes/core/solvers/petsc.py#L379)) If this is how things are intended to work, then we can close #3800.

Thank you so much for getting this in before the Pyomo release.

pyomo/core/tests/unit/test_suffix.py
blnicho: I think this test needs to be updated
```suggestion
        self.assertEqual(_suffix_b2_finder.find(m.b1.v2), 2)
```

blnicho:
@dallan-keylogic thanks for checking this fix. For now you still have to manually transfer any suffixes you want over to the temporary block. We have some follow-on ideas related to "Suffix References" which could help with this in the future but that won't make it for this release.

blnicho (approved):

PR #3854, opened 2026-02-18
pytest: rework `conftest.py` to support solver/writer marker expressions

jsiirola:
## Summary/Motivation:
Pyomo's existing pytest configuration supports marking solver tests using `pytest.mark.solver(name)`, and then running those tests using a custom `--solver name` command line argument.  However, the test selection was done by skipping tests at runtime and not deselecting the tests after collection.  Further, the markers could not be used in "normal" marker expressions (because pytest does not support accessing positional arguments in marker expressions).

This PR redesigns how we handle `solver` markers:
- the positional argument is mapped to a keyword argument (`id`).  This enables using markers expressions like `solver(id='highs')` to select tests based on the solver marker.
- an additional keyword argument (`vendor`) is also defined using a substring of `id` up to, but not including the first underscore.  This supports selecting "all" tests for a vendor.  For example `solver(vendor='gurobi')` will match `gurobi`, `gurobi_direct`, and `gurobi_persistent` solver markers.
- the `--solver` command line option is now translated into a marker expression (instead of relying on skipping tests at runtime)
- move the `solver` marker definition into the `pyproject.toml` (with all the other markers)

In addition, this PR:
- adds a `writer` marker that works analogously to the `solver` marker
- removes a number of unused pytest markers from `pyproject.toml`
- propagates the use of the `solver` marker through the solver tests in `pyomo.solvers` and `pyomo.contrib.solver`
- found a way to mark tests generated by `paarameterized.expand`

- (see above)

### Legal Acknowledgement

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -44,7 +44,7 @@ def pytest_itemcollected(item):
 
     We have historically supported "category markers"::
 
-        @pytest.mark.solve("highs")
+        @pytest.mark.solver("highs")
 
     Unfortunately, pytest doesn't allow for building marker
     expressions (e.g., for "-m") based on the marker.args.  We will
--- a/conftest.py
+++ b/conftest.py
@@ -32,7 +32,7 @@ def pytest_configure(config):
 
 
 def pytest_itemcollected(item):
-    """Standardize all pyomo test markers.
+    """Standardize all Pyomo test markers.
 
     This callback ensures that all unmarked tests, along with all tests
     that are only marked by category markers (e.g., "solver" or
@@ -50,7 +50,7 @@ def pytest_itemcollected(item):
     expressions (e.g., for "-m") based on the marker.args.  We will
     map the positional argument (for pytest.mark.solver and
     pytest.mark.writer) to the keyword argument "id".  This will allow
-    querying againse specific solver interfaces in marker expressions
+    querying against specific solver interfaces in marker expressions
     with::
 
         solver(id='highs')
--- a/conftest.py
+++ b/conftest.py
@@ -10,20 +10,53 @@
 import pytest
 
 _implicit_markers = {'default'}
-_extended_implicit_markers = _implicit_markers.union({'solver'})
+_category_markers = {'solver', 'writer'}
+_extended_implicit_markers = _implicit_markers.union(_category_markers)
+
+def pytest_configure(config):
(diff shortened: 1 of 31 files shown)


conftest.py
```diff
@@ -46,17 +80,12 @@ def pytest_runtest_setup(item):
...
-    solvernames = [mark.args[0] for mark in item.iter_markers(name="solver")]
-    solveroption = item.config.getoption("--solver")
-    markeroption = item.config.getoption("-m")
-    item_markers = set(mark.name for mark in item.iter_markers())
-    if solveroption:
-        if solveroption not in solvernames:
-            pytest.skip("SKIPPED: Test not marked {!r}".format(solveroption))
-            return
-    elif markeroption:
+    if item.config.option.markexpr:
```
mrmundt: I believe the comment on this function will need to be updated to reflect the new logic.

pyproject.toml
mrmundt, line 1: Did you try the performance tests with these changes?

conftest.py
```diff
@@ -10,20 +10,54 @@
...
+    # (for pytest.mark.solver and pytest.mark.writer) to the keyword
+    # argument "id".
+    #
+    # We will take this opportunity to also set a keyword argument for
+    # the solver/writer "vendor" (defined as the id up to the first
+    # underscore).  This will allow running "all Gurobi tests"
+    # (including, e.g., lp, direct, and persistent) with
+    #
+    #     "-m solver(vendor='gurobi')"
+    #
```
mrmundt: It's worth amending this comment to show how the new categories can be combined, e.g., is it possible to do `-m solver(vendor='gurobi') -m writer(vendor='lp')`?

conftest.py
jsiirola: Good point.  Up[on further thought, we can remove this entire callback and implement everything with standard markers and marker expressions.

pyproject.toml
jsiirola: yes

conftest.py
jsiirola: It is replicating pytest documentation, but I added a comment that complex expressions are supported using the pytest expression operators.  How pytest parses multiple `-m` on the command line is really up to pytest (they currently silently ignore everything other than the last `-m` argument)

mrmundt (approved):

PR #3860, opened 2026-02-24
KKT Transform

ChrisLaliwala:
## Summary/Motivation:
This PR creates a Pyomo transformation for reformulating models using the KKT conditions.

-
-

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/mpc/data/interpolation.py
+++ b/pyomo/contrib/mpc/data/interpolation.py
@@ -71,7 +71,10 @@ def _get_interp_expr_vec(time_set, time_data, data, indexes=None):
     expr = [None] * len(time_set)
     for i, (h, t) in enumerate(zip(indexes, time_set)):
         l = h - 1
-        expr[i] = data[l] + (data[h] - data[l]) / (time_data[h] - time_data[l]) * (
-            t - time_data[l]
-        )
+        if data[l] is None or data[h] is None:
+            expr[i] = None
+        else:
+            expr[i] = data[l] + (data[h] - data[l]) / (time_data[h] - time_data[l]) * (
+                t - time_data[l]
+            )
     return expr
--- a/pyomo/contrib/mpc/data/tests/test_series_data.py
+++ b/pyomo/contrib/mpc/data/tests/test_series_data.py
@@ -144,6 +144,20 @@ class TestSeriesData(unittest.TestCase):
             new_data,
         )
 
+    def test_get_data_interpolate_none(self):
+        m = self._make_model()
+        data_dict = {m.var[:, "A"]: [1, 2, 3], m.var[:, "B"]: [None, None, None]}
+        data = TimeSeriesData(data_dict, m.time)
+
+        new_t = [0.05, 0.15]
+        new_data = data.get_interpolated_data(new_t)
+        self.assertEqual(
+            TimeSeriesData(
+                {m.var[:, "A"]: [1.5, 2.5], m.var[:, "B"]: [None, None]}, new_t
+            ),
+            new_data,
+        )
+
     def test_get_data_interpolate_range_check(self):
         m = self._make_model()
         data_dict = {m.var[:, "A"]: [1, 2, 3], m.var[:, "B"]: [2, 4, 6]}
--- a/pyomo/contrib/mpc/interfaces/load_data.py
+++ b/pyomo/contrib/mpc/interfaces/load_data.py
@@ -7,6 +7,7 @@
 # software.  This software is distributed under the 3-clause BSD License.
 # ____________________________________________________________________________________
 
+from pyomo.core.base import Expression
 from pyomo.contrib.mpc.data.dynamic_data_base import _is_iterable
 from pyomo.contrib.mpc.data.find_nearest_index import (
     find_nearest_index,
@@ -18,7 +19,7 @@ def _raise_invalid_cuid(cuid, model):
     raise RuntimeError("Cannot find a component %s on block %s" % (cuid, model))
 
 
-def load_data_from_scalar(data, model, time):
+def load_data_from_scalar(data, model, time, ignore_named_expressions=False):
     """A function to load ScalarData into a model
 
     Arguments
@@ -26,6 +27,7 @@ def load_data_from_scalar(data, model, time):
     data: ~scalar_data.ScalarData
     model: BlockData
     time: Iterable
+    ignore_named_expressions: Bool
 
     """
     data = data.get_data()
@@ -34,6 +36,10 @@ def load_data_from_scalar(data, model, time):
         var = model.find_component(cuid)
         if var is None:
             _raise_invalid_cuid(cuid, model)
+        elif var.ctype == Expression:
+            if ignore_named_expressions:
+                continue
+            raise TypeError("Cannot load data for named Expression")
         # TODO: Time points should probably use find_nearest_index
         # This will have to happen in the calling function, as data
         # doesn't have a list of time points to check.
@@ -44,7 +50,9 @@ def load_data_from_scalar(data, model, time):
             var.set_value(val)
 
 
-def load_data_from_series(data, model, time, tolerance=0.0):
+def load_data_from_series(
+    data, model, time, tolerance=0.0, ignore_named_expressions=False
+):
     """A function to load TimeSeriesData into a model
 
     Arguments
@@ -52,6 +60,7 @@ def load_data_from_series(data, model, time, tolerance=0.0):
     data: TimeSeriesData
     model: BlockData
     time: Iterable
+    ignore_named_expressions: Bool
 
     """
     time_list = list(time)
@@ -72,6 +81,10 @@ def load_data_from_series(data, model, time, tolerance=0.0):
         var = model.find_component(cuid)
         if var is None:
             _raise_invalid_cuid(cuid, model)
+        elif var.ctype == Expression:
+            if ignore_named_expressions:
+                continue
+            raise TypeError("Cannot load data for named Expression")
         for idx, val in zip(time_indices, vals):
             t = time_list[idx]
             var[t].set_value(val)
@@ -85,6 +98,7 @@ def load_data_from_interval(
     prefer_left=True,
     exclude_left_endpoint=True,
     exclude_right_endpoint=False,
+    ignore_named_expressions=False,
 ):
     """A function to load IntervalData into a model
 
@@ -104,6 +118,7 @@ def load_data_from_interval(
     prefer_left: Bool
     exclude_left_endpoint: Bool
     exclude_right_endpoint: Bool
+    ignore_named_expressions: Bool
 
     """
     if prefer_left and exclude_right_endpoint and not exclude_left_endpoint:
@@ -169,6 +184,10 @@ def load_data_from_interval(
         var = model.find_component(cuid)
         if var is None:
             _raise_invalid_cuid(cuid, model)
+        elif var.ctype == Expression:
+            if ignore_named_expressions:
+                continue
+            raise TypeError("Cannot load data for named Expression")
         for i, t in zip(idx_list, time):
             if i is None:
                 # t could not be found in an interval. This is fine.
--- a/pyomo/contrib/mpc/interfaces/model_interface.py
+++ b/pyomo/contrib/mpc/interfaces/model_interface.py
@@ -170,6 +170,7 @@ class DynamicModelInterface:
         prefer_left=None,
         exclude_left_endpoint=None,
         exclude_right_endpoint=None,
+        ignore_named_expressions=False,
     ):
         """Method to load data into the model.
 
@@ -189,6 +190,10 @@ class DynamicModelInterface:
         time_points: Iterable (optional)
             Subset of time points into which data should be loaded.
             Default of None corresponds to loading into all time points.
+
+        ignore_named_expressions: (optional)
+            ignore data for named expressions, otherwise a TypeError will
+            be raise on encountering a named expression.
 
         """
         if time_points is None:
@@ -211,10 +216,21 @@ class DynamicModelInterface:
             # This covers the case of non-time-indexed variables
             # as keys.
             _error_if_used(prefer_left, excl_left, excl_right, type(data))
-            load_data_from_scalar(data, self.model, time_points)
+            load_data_from_scalar(
+                data,
+                self.model,
+                time_points,
+                ignore_named_expressions=ignore_named_expressions,
+            )
         elif isinstance(data, TimeSeriesData):
             _error_if_used(prefer_left, excl_left, excl_right, type(data))
-            load_data_from_series(data, self.model, time_points, tolerance=tolerance)
+            load_data_from_series(
+                data,
+                self.model,
+                time_points,
+                tolerance=tolerance,
+                ignore_named_expressions=ignore_named_expressions,
+            )
         elif isinstance(data, IntervalData):
             prefer_left = True if prefer_left is None else prefer_left
             excl_left = prefer_left if excl_left is None else excl_left
@@ -227,6 +243,7 @@ class DynamicModelInterface:
                 prefer_left=prefer_left,
                 exclude_left_endpoint=excl_left,
                 exclude_right_endpoint=excl_right,
+                ignore_named_expressions=ignore_named_expressions,
             )
 
     def copy_values_at_time(self, source_time=None, target_time=None):
--- a/pyomo/contrib/mpc/interfaces/tests/test_interface.py
+++ b/pyomo/contrib/mpc/interfaces/tests/test_interface.py
@@ -29,6 +29,7 @@ class TestDynamicModelInterface(unittest.TestCase):
         )
         m.input = pyo.Var(m.time, initialize={i: 1.0 - i * 0.1 for i in m.time})
         m.scalar = pyo.Var(initialize=0.5)
+        m.scalar_squared = pyo.Expression(expr=m.scalar**2)
         m.var_squared = pyo.Expression(
             m.time,
             m.comp,
@@ -128,6 +129,21 @@ class TestDynamicModelInterface(unittest.TestCase):
         interface.load_data(data)
         self.assertEqual(m.scalar.value, 6.0)
 
+    def test_load_scalar_data_expr(self):
+        # load_scalar_data has been removed. Instead we simply call
+        # load_data
+        m = self._make_model()
+        interface = DynamicModelInterface(m, m.time)
+        data = {
+            pyo.ComponentUID(m.scalar): 6.0,
+            pyo.ComponentUID(m.scalar_squared): 6.0,
+        }
+        interface.load_data(data, ignore_named_expressions=True)
+        self.assertEqual(m.scalar.value, 6.0)
+        self.assertEqual(pyo.value(m.scalar_squared), 36.0)
+        with self.assertRaises(TypeError):
+            interface.load_data(data)
+
     def test_load_data_at_time_all(self):
         # NOTE: load_data_at_time has been deprecated
         m = self._make_model()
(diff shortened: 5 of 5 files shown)


ChrisLaliwala:
Need to add a couple of more test before PR is ready to be reviewed. Will be done in 1-2 days.

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,438 @@
+#  ___________________________________________________________________________
+#
+#  Pyomo: Python Optimization Modeling Objects
+#  Copyright (c) 2008-2025
+#  National Technology and Engineering Solutions of Sandia, LLC
+#  Under the terms of Contract DE-NA0003525 with National Technology and
+#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain
+#  rights in this software.
+#  This software is distributed under the 3-clause BSD License.
+#  ___________________________________________________________________________
```
blnicho:
```suggestion
# ____________________________________________________________________________________
#
# Pyomo: Python Optimization Modeling Objects
# Copyright (c) 2008-2026 National Technology and Engineering Solutions of Sandia, LLC
# Under the terms of Contract DE-NA0003525 with National Technology and Engineering
# Solutions of Sandia, LLC, the U.S. Government retains certain rights in this
# software.  This software is distributed under the 3-clause BSD License.
# ____________________________________________________________________________________
```
We updated our copyright statement last week.

pyomo/core/tests/unit/test_kkt.py
```diff
@@ -0,0 +1,464 @@
+#  ___________________________________________________________________________
+#
+#  Pyomo: Python Optimization Modeling Objects
+#  Copyright (c) 2008-2025
+#  National Technology and Engineering Solutions of Sandia, LLC
+#  Under the terms of Contract DE-NA0003525 with National Technology and
+#  Engineering Solutions of Sandia, LLC, the U.S. Government retains certain
+#  rights in this software.
+#  This software is distributed under the 3-clause BSD License.
+#  ___________________________________________________________________________
```
blnicho:
```suggestion
# ____________________________________________________________________________________
#
# Pyomo: Python Optimization Modeling Objects
# Copyright (c) 2008-2026 National Technology and Engineering Solutions of Sandia, LLC
# Under the terms of Contract DE-NA0003525 with National Technology and Engineering
# Solutions of Sandia, LLC, the U.S. Government retains certain rights in this
# software.  This software is distributed under the 3-clause BSD License.
# ____________________________________________________________________________________
```
We updated our copyright statement last week

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,432 @@
...
+            - (Constraint, bound) tuple for ranged constraints
+            - (Var, bound) tuple for variable bounds
+        """
+
+        info = model.private_data()
+        if multiplier_var in info.equality_con_from_multiplier.keys():
+            return info.equality_con_from_multiplier[multiplier_var]
+        if multiplier_var in info.inequality_con_from_multiplier.keys():
+            # if this multiplier var maps to a ranged constraint, we will return a tuple
+            # so that we can indicate which bound the multplier var maps to
```
blnicho:
```suggestion
            # so that we can indicate which bound the multiplier var maps to
```

pyomo/core/tests/unit/test_kkt.py
```diff
@@ -0,0 +1,514 @@
...
+
+        m_reform = m.clone()
+        TransformationFactory('core.kkt').apply_to(
+            m_reform, parametrize_wrt=[m_reform.outer1, m_reform.outer2]
+        )
+        TransformationFactory("mpec.simple_nonlinear").apply_to(m_reform)
+
+        self.check_primal_kkt_transformation_solns(m, m_reform)
+
+    def test_multipe_obj_error(self):
```
blnicho:
```suggestion
    def test_multiple_obj_error(self):
```

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,432 @@
...
+        # we should check that all vars the user fixed are included
+        # in parametrize_wrt
+        params = config.parametrize_wrt or ComponentSet()
+        vars_in_cons = ComponentSet(
+            get_vars_from_components(model, Constraint, active=True, descend_into=True)
+        )
+        vars_in_obj = ComponentSet(
+            get_vars_from_components(model, Objective, active=True, descend_into=True)
+        )
+        vars_in_model = ComponentSet(vars_in_cons | vars_in_obj)
```
jsiirola: The extra call to `ComponentSet` is redundant (and actually causes the ComponentSet returned by the `|` operator to be duplicated)

```suggestion
        vars_in_model = vars_in_cons | vars_in_obj
```

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,432 @@
...
+        fixed_vars_in_model = ComponentSet(v for v in vars_in_model if v.is_fixed())
+        missing = [v for v in fixed_vars_in_model if v not in params]
+        if missing:
+            raise ValueError("All fixed variables must be included in parametrize_wrt.")
+
+        # we should also check that all vars the user passes in parametrize_wrt
+        # exist on an active constraint or objective within the model
+        unknown = [v for v in params if v not in vars_in_model]
```
jsiirola: ComponentSet objects support normal set operations... this could just be:

```suggestion
        unknown = params - vars_in_model
```

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,432 @@
...
+        config = self.CONFIG(kwds.pop('options', {}))
+        config.set_value(kwds)
+
+        if hasattr(model, config.kkt_block_name):
+            raise ValueError(f"""model already has an attribute with the 
+                specified kkt_block_name: '{config.kkt_block_name}'""")
+
+        # we should check that all vars the user fixed are included
+        # in parametrize_wrt
+        params = config.parametrize_wrt or ComponentSet()
```
jsiirola: You can make this simpler by declaring a default value for `parameterize_wrt` above:
```python
   default=[],
```

If you do that, then it should never be None and you can just do

```suggestion
        params = config.parametrize_wrt
```

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,432 @@
...
+            raise ValueError(
+                "A variable passed in parametrize_wrt does not exist on an "
+                "active constraint or objective within the model."
+            )
+
+        kkt_block = Block()
+        model.add_component(config.kkt_block_name, kkt_block)
+        kkt_block.parametrize_wrt = params
+
+        return self._reformulate(model, kkt_block)
```
jsiirola: This is (another) nit, but if you end up raising an exception from `_reformulate`, then you will leave the model with a "partial kkt block".  It would be more robust to build the whole block, and then - only after you know everything is OK - attach it tot he model:

```suggestion
        kkt_block = Block(concrete=True)
        kkt_block.parametrize_wrt = params
        self._reformulate(model, kkt_block)
        model.add_component(config.kkt_block_name, kkt_block)
        return model
```

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,432 @@
...
+    def _construct_lagrangean(self, model, kkt_block):
+        # we need to loop through the model and store the
+        # equality and inequality constraints
+        info = model.private_data()
+        equality_cons = []
+        inequality_cons = []
+        for con in model.component_data_objects(
+            Constraint, descend_into=True, active=True
+        ):
+            if con.has_lb() and con.has_ub() and (con.lower == con.upper):
```
jsiirola: This is a bit of a nuanced thing, but Pyomo stores the "raw" relational expression from the user in the Constraint.  Asking for lower, or body, or upper forces it to "normalize" the constraint with a call to `to_bounded_expression`.  It would be significantly more efficient to only call that *once* with something like:
```python
lower, body, upper = con.to_bounded_expression()
```
and then `has_lb` is equivalent to `lower is not None`

Testing for equality is also tricky: `con.equality()` is probably your best bet.  It will return True for EqualityExpressions (i.e., constraints created with `==`), *and* for RangedExpressions where `lower is upper`.  That is a bit restrictive, but the problem with testing with `==` is if the bounds are "not potentially variable" expressions (i.e., expressions of mutable Params), then `==` will create a new equality expression that will raise an exception in your "if" (evaluating a non-constant expression in a boolean context).

In the end, I would suggest logic something like:
```python
lower, body, upper = con.to_bounded_expression()
if con.equality():
    equality_cons.append(con)
    info.equality_con_to_expr[con] = con.upper - con.body
else:
    if lower is not None:
        inequality_cons.append((con, "lb"))
        info.inequality_con_to_expr.setdefault(con, {})["lb"] = lower - body
    if upper is not None:
        inequality_cons.append((con, "ub"))
        info.inequality_con_to_expr.setdefault(con, {})["ub"] = body - upper

    if lower is not None and upper is not None:
        # we want to keep track of the ranged constraints because the mapping between
        # multipliers and ranged constraints will be a tuple (to indicate bound as well)
        # instead of simply the model object
        info.ranged_constraints.add(con)
```
(you could, of course save a redundant comparison to None by putting the check for ranged inside one of the previous (lower / upper is not None) tests.  But that might be less clear to readers...)

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,432 @@
...
+                    info.inequality_con_to_expr.setdefault(con, {})["ub"] = (
+                        con.body - con.upper
+                    )
+                # we want to keep track of the ranged constraints because the mapping between
+                # multipliers and ranged constraints will be a tuple (to indicate bound as well)
+                # instead of simply the model object
+                if con.has_lb() and con.has_ub() and (con.lower != con.upper):
+                    info.ranged_constraints.add(con)
+
+        kkt_block.equality_cons_set = Set(initialize=equality_cons, ordered=True)
```
jsiirola: FWIW, `ordered=True` is the default

That said, why does this need to be a Set?  I don't think that you ever use it to index anything, nor do you look for membership in it.

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,432 @@
...
+        # formed by the user specifying variable bounds
+        var_bound_sides = []
+        vars_in_cons_all = ComponentSet(
+            get_vars_from_components(model, Constraint, active=True, descend_into=True)
+        )
+        vars_in_kkt_cons = ComponentSet(
+            get_vars_from_components(
+                kkt_block, Constraint, active=True, descend_into=True
+            )
+        )
```
jsiirola: When is this set not empty?  At this point, it doesn't look like you have added any Constraints to the `kkt_block`?

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,432 @@
...
+        vars_in_cons = ComponentSet(vars_in_cons_all - vars_in_kkt_cons)
+        vars_in_obj = ComponentSet(
+            get_vars_from_components(model, Objective, active=True, descend_into=True)
+        )
+        kkt_block.var_set = ComponentSet(vars_in_cons | vars_in_obj)
```
jsiirola: Again, you don't have to duplicate the ComponentSet here.

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,432 @@
...
+        # the corresponding multiplier
+        obj = list(
+            model.component_data_objects(Objective, active=True, descend_into=True)
+        )
+        # maximize is -1 and minimize is +1
+        sense = obj[0].sense
+        if sense == maximize:
+            lagrangean = -obj[0].expr
+        elif sense == minimize:
+            lagrangean = obj[0].expr
```
jsiirola: Nothing wrong with this implementation -- but "if trees" that don't have an `else` make reviewers nervous (what happens if an unexpected value comes in?).  In this case, it's OK - `lagrangean` will be undefined and we will get an exception (the thing I usually worry about with bare if trees is accidentally doing *nothing*).  Further, `sense` is guaranteed to be an `ObjectiveSense` enum, which only has `minimize` and `maximize` members.  So, I suspect that you could also omit the test against minimize.

I am not necessary advocating the following, but one of the reasons that `ObjectiveSense` is an IntEnum (and that it has the values it does) is you *could* do this:

```suggestion
        # Note: maximize is -1 and minimize is +1
        lagrangean = obj[0].sense * obj[0].expr
```

(@mrmundt, is that "too cute"?)

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,432 @@
...
+        Returns
+        -------
+        Constraint or Tuple
+            - Constraint object for simple constraints
+            - (Constraint, bound) tuple for ranged constraints
+            - (Var, bound) tuple for variable bounds
+        """
+
+        info = model.private_data()
+        if multiplier_var in info.equality_con_from_multiplier.keys():
```
jsiirola: The `keys()` is unnecessary (and slower).  Omitting it hits the container's `__contains__` method directly (which is preferable):
```suggestion
        if multiplier_var in info.equality_con_from_multiplier:
```

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,432 @@
...
+            return info.equality_con_from_multiplier[multiplier_var]
+        if multiplier_var in info.inequality_con_from_multiplier.keys():
```
jsiirola: Same here: omit `.keys()`

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,432 @@
...
+            # if this multiplier var maps to a ranged constraint, we will return a tuple
+            # so that we can indicate which bound the multiplier var maps to
+            return info.inequality_con_from_multiplier[multiplier_var]
+        raise ValueError(
+            f"The KKT multiplier: {multiplier_var.name}, does not exist on {model.name}."
+        )
+
+    def get_multiplier_from_constraint(self, model, constraint=None, variable=None):
```
jsiirola: I wonder if a different / simpler API might be preferable?  I am wondering about something like:

```python
def get_multipliers(self, model, component) -> VarData | tuple[VarData, VarData]:
```

That is, if component is an equality or (simple) inequality, then return the multiplier.  If it is a ranged inequality of a bounded variable, then return a tuple of the multipliers for the lower and upper bounds (for the case of variables, one [or I suppose both] could be None).

ChrisLaliwala:
> This is really good. I don't think that I found anything that is incorrect, but lots of opportunities to make the code cleaner / more efficient.
> 
> One general note: you create a lot of intermediate data structures as sets (usually `ComponentSet` -- but you never test for membership in the set, and the sets don't look like they are being used for deduplication (only iteration later). In that case it is significantly more efficient to make them lists and not hashed (set or dict) data structures.
> 
> The following is a suggestion (that I though would make this a lot simpler / easier to follow), but could probably be argued into deferring for later:
> 
> In addition to the notes below, I wonder if the transformation could be significantly simpler / easier to follow if there was a bigger refactor? Right now the transformation
> 
> 1. walks the whole model and collects the variables from all constraints (and objectives) and does a fair amount of preliminary error checking (e.g., verifying the completeness of the `parametrize_wrt` set)
> 2. walks the model a second time collecting the constraints and sorts them into internal data structures (equality / inequality constraint lists, with bound information)
> 3. loop through the variables and build bound expressions
> 4. walk the model and collect objectives to create the Lagrangean
> 5. loop through the constraint lists and add them to the Lagrangean
> 6. loop through the variable bounds and add them to the Lagrangean
> 7. Compute the derivative of the Lagrangean and add the stationarity constraints
> 8. loop through the inequality constraint list and define the complementary slackness constraints
> 9. loop through the variable bounds and define the complimentary slackness
> 
> This is a lot of repeated work (and requires making a lot of somewhat unintuitive internal data structures. I wonder if you can avoid most of these data structures and the complexity by basically reversing the order you do things:
> 
> 1. initialize
>    
>    * `lagrangean = 0`
>    * var_set = ComponentSet()
>    * create the block to hold the KKT system (but don't add it to the model)
>    * 2 empty mappings (obj -> dual variable(s); dual variable -> obj)
> 2. collect the active Objectives from the model.  This should be a single-element list (if not, error).
>    
>    * collect the  variables form the objective into the var_set
>    * add the expression to the Lagrangean
> 3. collect the active ComoponentDatas from the model; for each one:
>    
>    * collect the  variables form the objective into the var_set
>    * determine if it is an equality constraint; if yes:
>      
>      * get the multiplier, add the term (gamma * (body - ub)) to the Lagrangean
>      * add the gamma -> con and con -> gamma to the mappings
>    * if no:
>      
>      * convert con to `lower, body, upper`
>      * if lower is not None:
>        
>        * get a multiplier, add the term (alpha1 * (lower - body)) to the Lagrangean
>        * add the dual / expression complimentarity condition to the model (you can use the same ComplimentarityList you were using before)
>        * add the alpha1 -> con to the mapping
>      * if upper is not None:
>        
>        * get a multiplier, add the term (alpha2 * (body - upper)) to the Lagrangean
>        * add the dual / expression complimentarity condition to the model
>        * add the alpha2 -> con to the mapping
>      * add the con -> (alpha, alpha) to the mappings
> 4. Now that we have all the variables, do the error checking on `parametrize_wrt`
> 5. Loop through the variables
>    
>    * get multipliers and add terms to the Lagrangean and define complimentary slackness (as with inequality constraints above)
> 6. Now we have a full Lagrangean and the correct var_set: take the derivative and define the stationarity constraints.
> 
> Last note (that should NOT be part of this PR, but we should consider for the future): As the model is likely to go out through the NL writer, we could save on subsequent computation time by the judicious use of `Expression` objects. This could involve rewriting the primal constraints or making the NL writer smarter about identifying repeated expressions.

Thank you for these comments and suggestions. There is definitely a lot of repeated work, and the transformation can be simplified. Now that I've addressed the other issues, I will open a PR to implement this ASAP.

pyomo/core/plugins/transform/kkt.py
ChrisLaliwala: Thanks for the feedback. I have tried to save a redudant comparison to None and added a comment to make it a bit more clear for users: 

```
            lower, body, upper = con.to_bounded_expression()
            if con.equality:
                equality_cons.append(con)
                info.equality_con_to_expr[con] = upper - body
            else:
                if lower is not None:
                    inequality_cons.append((con, "lb"))
                    info.inequality_con_to_expr.setdefault(con, {})["lb"] = lower - body
                if upper is not None:
                    inequality_cons.append((con, "ub"))
                    info.inequality_con_to_expr.setdefault(con, {})["ub"] = body - upper

                    # lower is not None and upper is not None -> ranged constraint
                    if lower is not None:
                        # we want to keep track of the ranged constraints because the mapping between
                        # multipliers and ranged constraints will be a tuple (to indicate bound as well)
                        # instead of simply the model object
                        info.ranged_constraints.add(con)
```

pyomo/core/plugins/transform/kkt.py
ChrisLaliwala: This is a good catch. I was trying to exclude variables from KKT constraints here, but you are right to point out that the constraints have not been added to the KKT block, so there is no need to check for this, and thus the set is always empty. I have removed this code.

pyomo/core/plugins/transform/kkt.py
ChrisLaliwala: Your are correct, this does not need to be a set. I have switched this to a list. Additionally, I have gone through the code and changed all the unecessary sets to lists.

pyomo/core/plugins/transform/kkt.py
ChrisLaliwala: I have simplified the API as per your suggestions.

blnicho:
Superseded by #3881

PR #3861, opened 2026-02-24
Fix no results returned when no discrete variables are present in MindtPy

### LINKED ISSUE #3855: MindtPy fails to return solutions for NLP or LP problem because of the short-circuit mechanism.
## Summary

I ran `gdpopt.ldsda` using mindtpy as the MINLP subproblem solver. The algorithm returned a results object that contains no solutions.

This appears related to MindtPy’s short-circuit (e.g., all discrete variables are fixed or just an NLP problem), where MindtPy may directly solve an LP/NLP but does not reliably return a populated SolverResults / does not load primal values back onto the model instance provided to solve().

**Reproducer A** (direct MindtPy; “no discrete decisions” path)
Run the following script:

```python
# example.py
  from pyomo.environ import ConcreteModel, Var, Objective, Constraint, Binary, NonNegativeReals, value, SolverFactory
  
  m = ConcreteModel()
  m.x = Var(domain=NonNegativeReals)
  m.y = Var(domain=Binary)
  
  # Fix the discrete variable so MindtPy detects "no discrete decisions"
  m.y.fix(0)
  
  # NLP constraint so MindtPy takes the NLP direct-solve branch
  m.c = Constraint(expr=m.x**2 >= 1 + m.y)
  m.obj = Objective(expr=m.x)
  
  res = SolverFactory("mindtpy").solve(
      m,
      strategy="OA",
      nlp_solver="ipopt",
      # mip_solver can be provided but should not matter if there are no discretes
      mip_solver="appsi_highs",
  )
  
  print("res is None:", res is None)
```

**Observed in the test output:**
`res is None: True`  

**Expected behavior**
`res is None: False` 

**Reproducer B** (GDPopt LD-SDA gams free test)

Please change the pyomo/contrib/gdpopt/test/test_ldsda.py to:

```python
class TestGDPoptLDSDA(unittest.TestCase):

....notations...

    @unittest.skipUnless(
        all(
            (
                SolverFactory("mindtpy").available(False),
                SolverFactory("appsi_highs").available(False),
                SolverFactory("ipopt").available(False),
            )
        ),
        "mindtpy/appsi_highs/ipopt not available",
    )
    def test_solve_four_stage_dynamic_model_minimize(self):
    
        model = build_model(mode_transfer=True)
        
        discretizer = TransformationFactory("dae.collocation")
        discretizer.apply_to(model, nfe=10, ncp=3, scheme="LAGRANGE-RADAU")
      
        for disjunct in model.component_data_objects(ctype=Disjunct):
            for constraint in disjunct.component_objects(ctype=Constraint):
                constraint._constructed = False
                constraint.construct()

        for dxdt in model.component_data_objects(ctype=Var, descend_into=True):
            if "dxdt" in dxdt.name:
                dxdt.setlb(-300)
                dxdt.setub(300)

        for direction_norm in ["L2", "Linf"]:
            results = SolverFactory("gdpopt.ldsda").solve(
                model,
                direction_norm=direction_norm,
                minlp_solver="mindtpy",
                minlp_solver_args={"mip_solver": "appsi_highs", "nlp_solver": "ipopt"},
                # minlp_solver="gams",
                # minlp_solver_args={'solver': "ipopt"},
                starting_point=[1, 2],
                logical_constraint_list=[
                    model.mode_transfer_lc1,
                    model.mode_transfer_lc2,
                ],
                time_limit=100,
            )
        
            self.assertAlmostEqual(value(model.obj), -23.305325, places=4)

...the rest of test unchanged...

```

```cmd
pytest <path_to_test.py>
```

### Error Message

The error message 
```txt
============================================= test session starts =============================================
platform win32 -- Python 3.12.12, pytest-8.4.2, pluggy-1.5.0
rootdir: 
configfile: pyproject.toml
plugins: cov-7.0.0
collected 10 items

pyomo\contrib\gdpopt\tests\test_ldsda.py F.........                                                      [100%]

================================================== FAILURES ===================================================
________________________ TestGDPoptLDSDA.test_solve_four_stage_dynamic_model_minimize _________________________

self = <pyomo.contrib.gdpopt.tests.test_ldsda.TestGDPoptLDSDA testMethod=test_solve_four_stage_dynamic_model_minimize>

    @unittest.skipUnless(
        all(
            (
                SolverFactory("mindtpy").available(False),
                SolverFactory("appsi_highs").available(False),
                SolverFactory("ipopt").available(False),
            )
        ),
        "mindtpy/appsi_highs/ipopt not available",
    )
    def test_solve_four_stage_dynamic_model_minimize(self):

        model = build_model(mode_transfer=True)

        discretizer = TransformationFactory("dae.collocation")
        discretizer.apply_to(model, nfe=10, ncp=3, scheme="LAGRANGE-RADAU")

        for disjunct in model.component_data_objects(ctype=Disjunct):
            for constraint in disjunct.component_objects(ctype=Constraint):
                constraint._constructed = False
                constraint.construct()

        for dxdt in model.component_data_objects(ctype=Var, descend_into=True):
            if "dxdt" in dxdt.name:
                dxdt.setlb(-300)
                dxdt.setub(300)

        for direction_norm in ["L2", "Linf"]:
            results = SolverFactory("gdpopt.ldsda").solve(
                model,
                direction_norm=direction_norm,
                minlp_solver="mindtpy",
                minlp_solver_args={"mip_solver": "appsi_highs", "nlp_solver": "ipopt"},
                # minlp_solver="gams",
                # minlp_solver_args={'solver': "ipopt"},
                starting_point=[1, 2],
                logical_constraint_list=[
                    model.mode_transfer_lc1,
                    model.mode_transfer_lc2,
                ],
                time_limit=100,
            )

>           self.assertAlmostEqual(value(model.obj), -23.305325, places=4)
                                   ^^^^^^^^^^^^^^^^

pyomo\contrib\gdpopt\tests\test_ldsda.py:75:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
pyomo\common\numeric_types.py:378: in value
    tmp = obj(exception=True)
          ^^^^^^^^^^^^^^^^^^^
pyomo\core\base\objective.py:462: in __call__
    return super().__call__(exception)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
pyomo\core\base\expression.py:63: in __call__
    return arg(exception=exception)
           ^^^^^^^^^^^^^^^^^^^^^^^^
pyomo\core\expr\base.py:116: in __call__
    return visitor.evaluate_expression(self, exception)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pyomo\core\expr\visitor.py:1312: in evaluate_expression
    ans = visitor.dfs_postorder_stack(exp)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pyomo\core\expr\visitor.py:930: in dfs_postorder_stack
    flag, value = self.visiting_potential_leaf(_sub)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pyomo\core\expr\visitor.py:1213: in visiting_potential_leaf
    return True, value(node, exception=self.exception)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

obj = <pyomo.core.base.var.VarData object at 0x000002E32D106F90>, exception = True

    def value(obj, exception=True):
        """
        A utility function that returns the value of a Pyomo object or
        expression.

        Args:
            obj: The argument to evaluate. If it is None, a
                string, or any other primitive numeric type,
                then this function simply returns the argument.
                Otherwise, if the argument is a NumericValue
                then the __call__ method is executed.
            exception (bool): If :const:`True`, then an exception should
                be raised when instances of NumericValue fail to
                evaluate due to one or more objects not being
                initialized to a numeric value (e.g, one or more
                variables in an algebraic expression having the
                value None). If :const:`False`, then the function
                returns :const:`None` when an exception occurs.
                Default is True.

        Returns: A numeric value or None.
        """
        if obj.__class__ in native_types:
            return obj
        #
        # Test if we have a duck typed Pyomo expression
        #
        if not hasattr(obj, 'is_numeric_type'):
            #
            # TODO: Historically we checked for new *numeric* types and
            # raised exceptions for anything else.  That is inconsistent
            # with allowing native_types like None/str/bool to be returned
            # from value().  We should revisit if that is worthwhile to do
            # here.
            #
            if check_if_numeric_type(obj):
                return obj
            else:
                if not exception:
                    return None
                raise TypeError(
                    "Cannot evaluate object with unknown type: %s" % obj.__class__.__name__
                )
        #
        # Evaluate the expression object
        #
        if exception:
            #
            # Here, we try to catch the exception
            #
            try:
                tmp = obj(exception=True)
                if tmp is None:
>                   raise ValueError(
                        "No value for uninitialized %s object %s"
                        % (type(obj).__name__, obj.name)
                    )
E                   ValueError: No value for uninitialized VarData object x1[0.015505]

pyomo\common\numeric_types.py:380: ValueError
---------------------------------------------- Captured log call ----------------------------------------------
ERROR    pyomo.common.numeric_types:numeric_types.py:391 evaluating object as numeric value: x1[0.015505]
    (object: <class 'pyomo.core.base.var.VarData'>)
No value for uninitialized VarData object x1[0.015505]
ERROR    pyomo.common.numeric_types:numeric_types.py:391 evaluating object as numeric value: obj
    (object: <class 'pyomo.core.base.objective.ScalarObjective'>)
No value for uninitialized VarData object x1[0.015505]
=========================================== short test summary info ===========================================
FAILED pyomo/contrib/gdpopt/tests/test_ldsda.py::TestGDPoptLDSDA::test_solve_four_stage_dynamic_model_minimize - ValueError: No value for uninitialized VarData object x1[0.015505]
========================================= 1 failed, 9 passed in 3.90s =========================================
```

Pyomo version: Pyomo 6.10.0.dev0
Python version: CPython 3.12.12
Operating system: Windows 11
How Pyomo was installed (PyPI, conda, source): Git clone, pip install -e 
Solver (if applicable): MIP: appsi_highs & NLP: ipopt

bernalde:
## Fixes #3855 .
Fixes https://github.com/Pyomo/pyomo/issues/3855
## Summary/Motivation:
This PR fixes an issue where MindtPy can short-circuit on “no discrete decisions” (LP/NLP) and then fails to reliably return a proper SolverResults and/or load primal values onto the input model, even when the direct LP/NLP solve succeeds. This behavior breaks downstream meta-solvers (e.g., GDPopt subproblem solves) that depend on Var.value to capture an incumbent.

Reference: https://github.com/https://github.com/Pyomo/pyomo/issues/3855

MindtPy contains a validation/short-circuit path intended to directly solve models that do not require decomposition (e.g., LP/NLP, or models where all discrete variables are fixed). In this path, MindtPy may:

return None from solve() (bare return)

- Ensure MindtPy always returns a SolverResults object (never None) even when the model is handled by the LP/NLP short-circuit.
- Initialize and populate self.results before early exits, and propagate termination/status/message from the direct LP/NLP solve into MindtPy’s results.
- Add a standalone test/test_mindtpy_no_discrete.py script

### Legal Acknowledgement


Toflamus:
Hi team,

I’ve investigated the current CI failures across Linux, macOS, and Windows. The failures in pyomo/contrib/solver/tests/solvers are not caused by the code changes in this PR.

The root cause is an expired GAMS license in the test environment. It appears the license might have expired a few days ago, which is why we are seeing identical failures across all platforms.

Interestingly, the tests for MindtPy are still passing (likely because they use different solver paths or have different fallback mechanisms), but the core solver tests are blocked.

Once the GAMS license is renewed in the CI environment, these tests should return to green.

blnicho:
@Toflamus we are aware of the issue and discussed it during the developer call today. This is indeed an infrastructure issue and we're working on getting it fixed.

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -317,6 +359,50 @@ def model_is_valid(self):
...
+        lb = getattr(results.problem, 'lower_bound', None)
+        ub = getattr(results.problem, 'upper_bound', None)
+        if lb is not None:
+            prob.lower_bound = lb
+        if ub is not None:
+            prob.upper_bound = ub
+
+        # Fallback: if the solver reports optimal termination but does not provide
+        # explicit bounds, infer them from the objective value. For a direct
+        # continuous optimal solve, primal==dual.
```
jsiirola: Doesn't that imply convexity?  I feel like you are making a promise that may not hold for the actual model you are solving...

jsiirola (commented):
One question about results bounds, but otherwise, this looks good to me.

pyomo/contrib/mindtpy/algorithm_base_class.py
Toflamus: Good catch — the "primal==dual" phrasing is too strong. What this fallback actually relies on is not convexity but the solver's own claim: if the solver reports `tc.optimal` and didn't populate its own bounds, we use the objective value as the best available information for both bounds.

That said, you're right that for a local NLP solver like IPOPT, `tc.optimal` really only guarantees local optimality, so setting both bounds equal overstates what we actually know. Do you want me to fix the comment?

bernalde: Absolutely! Good catch. We cannot promise dual==primal, so in general we should inherit the dual and primal from the subsolver if they provide any. If they don't provide a dual (such as IPOPT) we only report the primal

jsiirola: I'll let others weigh in, but I think I would prefer to change the results to only return what we know; in this case, only return the objective value for the primal (feasible) bound and then return None for the dual (infeasible) bound.

Toflamus: Btw, just to mention, the "primal==dual" phrasing in the comment is admittedly loose, but the behavior here is safe. A few points:

1. `prob` is the result-reporting object (`self.results.problem`), not any subproblem bound used algorithmically. This code only runs in the short-circuit path (no discrete variables), so these bounds never feed back into any decomposition logic. It's purely for populating the solver results that get returned to the user. We're just filling in the `lower_bound`, `upper_bound`fields so the returned `SolverResults` isn't incomplete. 

2. This only fires in the short-circuit path — when `model_is_valid()` detects zero discrete variables and solves the original model directly as an LP/NLP. MindtPy then returns False and never enters the decomposition loop. So these bounds don't influence any cuts or iterations.

3. The fallback only triggers when the solver claims `tc.optimal` but didn't populate its own bounds. We're just mirroring what the solver already told us — if it says optimal, using the objective value as the reported bound is the best available information.

Toflamus: Anyway, I agree that we should only keep what we know.

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -317,6 +359,56 @@ def model_is_valid(self):
...
+            lb is None or ub is None
+        ) and self.results.solver.termination_condition == tc.optimal:
+            obj_val = value(obj.expr, exception=False)
+            if obj_val is not None:
+                if obj.sense == minimize:
+                    if ub is None:
+                        prob.upper_bound = obj_val
+                else:
+                    if lb is None:
+                        prob.lower_bound = obj_val
```
emma58, line 410: I'm confused about this: Isn't the results object from the solver already reporting whatever bounds *it* got? Why can't you just pass those through? If it's local, it should only have the primal bound, and not the dual bound, which is exactly what you want.

pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py
```diff
@@ -0,0 +1,423 @@
...
+# Under the terms of Contract DE-NA0003525 with National Technology and Engineering
+# Solutions of Sandia, LLC, the U.S. Government retains certain rights in this
+# software.  This software is distributed under the 3-clause BSD License.
+# ____________________________________________________________________________________
+
+
+from unittest.mock import MagicMock
+
+from pyomo.opt import TerminationCondition
+from pyomo.opt import TerminationCondition as tc
```
emma58: Looks like these could be consolidated.

emma58 (commented):
I have one question related to the conversation about bounds @jsiirola started, but otherwise this looks good to me.

pyomo/contrib/mindtpy/algorithm_base_class.py
bernalde: The code does pass through bounds from the solver first (lines 385-391). The fallback logic here only activates when the solver doesn't provide bounds at all. This happens with NLP solvers like IPOPT, which never populate lower_bound or upper_bound in their results objects (you can confirm this in sol.py - the sol reader sets constraint/variable/objective counts but never sets bounds). So the fallback is necessary to infer the primal bound from the objective value when the solver reports optimal but provides no bound information at all.

jsiirola (approved):
This looks good.  I do have a couple design questions to think about for the future:
- `model_is_valid` is a really expensive implementation.  In particular, computing the polynomial degree for every expression is pretty much the same cost as doing a full LP/NL write.  Is that duplicate effort really necessary?
- if the model is a QP, and the `mip_solver` can handle QPs, shouldn't we use it and not the `nlp_solver`?

I wonder if a future improvement would be to skip the polynomial degree checks and just try solving with the MIP solver. If that fails, then fall back on the NLP solver?

bernalde:
Following up on @jsiirola's design questions: I agree the duplicate-work concern is real. `model_is_valid()` currently walks polynomial degree again, even though utility-block setup has already classified constraints into linear and nonlinear lists, so there is a clear cleanup opportunity there.

I do not think #3861 is the right place to change the short-circuit routing behavior, though. Today, the continuous-no-discrete path does not choose between `mip_solver` and `nlp_solver` based on explicit subsolver QP/QCP capabilities; it is tied to MindtPy's current degree-based logic (and the existing `quadratic_strategy` behavior), so changing that here would broaden this PR beyond the bug fix. I also do not think a blanket "try MIP and fall back to NLP on failure" policy is safe, because a MIP failure can mean unsupported structure, but it can also mean infeasible, unbounded, time-limit, or solver-error outcomes, and silently retrying would risk masking real failures.

To keep this PR focused on the results-loading fix, I opened `#3906` to track the follow-up work:
- remove the duplicate structure walk by reusing the preclassified constraint lists and cached objective degree
- evaluate capability-aware routing for continuous QP/QCP short-circuit cases
- explicitly keep "fallback on any MIP failure" out of scope

So my proposal is to land #3861 on its current scope and handle the solver-routing refactor separately in #3906.

pyomo/contrib/mindtpy/algorithm_base_class.py
emma58: Oh, sorry, I think I was getting mixed up between old interfaces and new! This makes sense.

emma58 (approved):

PR #3865, opened 2026-03-01
Update GHA TPL cache; resolve new test failures

jsiirola:
## Summary/Motivation:
This updates the GHA TPL caches.  Sort of.  We are updating GAMS to the almost-current version (52.5) because the current version (53.1) causes segfaults (at least under pytest).  The updated versions led to some test errors that this PR also resolves:
- Rework the BARON license test to rely on the LST ("times") file and not parsing the log (in the most recent BARON version the log message changed).
   - This also takes the opportunity to significantly simplify the logic around parsing and caching the BARON version / licensing information.
   - This also adds some "defensive" exceptions to catch incomplete results parsing if the Baron API ever expands/
- Update Xpress "locally optimal" test so Xpress doesn't recognize it as a conic problem it "knows how to solve"
- Update GAMS on GHA:
  - Stop installing gams distribution on OSX (self-extracting zip is no longer distributed, and I couldn't see how to (easily) install the package on GHA.
  - Update the gams Python installation for recent package layout
  - Add fallback on gamspy from PyPI if we either don't recognize the layout or didn't install the package

While working on things, some other things were cleaned up:
- Rework/simplify the `download-extensions` command
  - simplify (and document!) function control flow.
  - move retry parameters from module constants to command-line arguments.
- GHA drivers
  - standardize on use of the `curlrc` for passing standard options to curl
  - remove creation of local variables that led to obfuscation of the driver logic
  - update `test-branches` python versions to get better coverage without adding tests
- cplex: IBMDecisionOptimization released new versions of cplex and docplex that are both a bit buggy
  - cplex 22.1.2.1 errors when computing the IIS on Python 3.13 & 3.14.  It looks like the file they write to is getting closed/corrupted before the routine exits.  Either way, it results in some (fatal) error is coming out of the compiled library.  I don't see how to fix or workaround, so we are just going to ignore that release.
  - docplex 2.32.257 is missing required dependencies for Python<=3.12.  This highlighted that our tests for whether docplex is available were fragile / incomplete.  This PR updates those tests to be more robust.
- Quit attempting to verify links to SourceForge: it appears that SF is categorically rejecting connections from GHA.

NOTE: the intent is to let this PR build and generate the new caches, then we will update the PR to *unpin* GAMS (but not trigger the caches to be rebuilt).

- (see above)

### Legal Acknowledgement

### DIFF
--- a/.github/workflows/test_branches.yml
+++ b/.github/workflows/test_branches.yml
@@ -95,13 +95,13 @@ jobs:
           PYENV: pip
 
         - os: macos-latest
-          python: 3.13
+          python: 3.12
           test_docs: 1
           TARGET: osx
           PYENV: pip
 
         - os: windows-latest
-          python: 3.14
+          python: 3.13
           test_docs: 1
           TARGET: win
           PYENV: conda
--- a/.github/workflows/test_branches.yml
+++ b/.github/workflows/test_branches.yml
@@ -477,8 +477,7 @@ jobs:
         echo "$IPOPT_DIR" >> $GITHUB_PATH
         echo "LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$IPOPT_DIR" >> $GITHUB_ENV
         echo "DYLD_LIBRARY_PATH=$DYLD_LIBRARY_PATH:$IPOPT_DIR" >> $GITHUB_ENV
-        NEW_PYOMO_PATH="$IPOPT_DIR:$PYOMO_PATH"
-        echo "PYOMO_PATH=$NEW_PYOMO_PATH" >> $GITHUB_ENV
+        echo "PYOMO_PATH=$IPOPT_DIR:$PYOMO_PATH" >> $GITHUB_ENV
         mkdir -p $IPOPT_DIR
         IPOPT_TAR=${DOWNLOAD_DIR}/ipopt.tar.gz
         if test ! -e $IPOPT_TAR; then
@@ -528,6 +527,10 @@ jobs:
             Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
         echo "DYLD_LIBRARY_PATH=${env:DYLD_LIBRARY_PATH}:$GAMS_DIR" | `
             Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
+        # Note: we put GAMS_DIR at the *end* of PYOMO_PATH so explicitly
+        #    installed solvers will be found first.
+        echo "PYOMO_PATH=${env:PYOMO_PATH}:$GAMS_DIR" | `
+            Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
         $INSTALLER = "${env:DOWNLOAD_DIR}/gams_install.exe"
         # Demo licenses are included for 5mo from the newest release
         $URL = "https://d37drm4t2jghv5.cloudfront.net/distributions/52.5.0"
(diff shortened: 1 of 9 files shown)


.github/workflows/test_branches.yml
```diff
@@ -95,13 +95,13 @@ jobs:
...
         - os: macos-latest
-          python: 3.13
+          python: 3.12
           test_docs: 1
           TARGET: osx
           PYENV: pip
 
         - os: windows-latest
-          python: 3.14
+          python: 3.13
```
mrmundt, line 105: This one was intentionally on 3.14 because that's where a lot of Windows weirdness seems to happen lately.

.github/workflows/test_branches.yml
```diff
@@ -477,61 +479,66 @@ jobs:
...
-        $URL = "https://d37drm4t2jghv5.cloudfront.net/distributions/latest"
+        $URL = "https://d37drm4t2jghv5.cloudfront.net/distributions/52.5.0"
         if ( "${{matrix.TARGET}}" -eq "win" ) {
             $URL = "$URL/windows/windows_x64_64.exe"
         } elseif ( "${{matrix.TARGET}}" -eq "osx" ) {
-            $URL = "$URL/macosx/osx_x64_64_sfx.exe"
+            $osxarch = (uname -m).Trim()
+            if ( "$osxarch" -eq "arm64" ) {
+                $URL = "$URL/macosx/osx_arm64_sfx.exe"
+            } else {
```
mrmundt, line 551: Why have this logic if we are skipping OSX?

pyomo/scripting/plugins/download.py
```diff
@@ -47,49 +43,55 @@ def _call_impl(self, args, unparsed, logger):
...
+                    if getattr(ext, "skip", bool)():
+                        # If the extension has a "skip" attribute, and
+                        # calling it returns Truth, then mark this as a
+                        # skipped download and move on.
                         result = "SKIP"
-                    elif hasattr(ext, "__call__"):
-                        ext()
-                        result = " OK "
                     else:
+                        # If the extension has a __cell__(), then call
```
mrmundt:
```suggestion
                        # If the extension has a __call__(), then call
```

mrmundt (changes requested):

.github/workflows/test_branches.yml
jsiirola: True - but most of that weirdness is because things aren't installing.  I would argue that it is more important to try and cover things that *are* on windows than to have the one Windows conda build be intentionally in an incomplete environment?  The 3.14 job will still get hit by the full PR builder, so things won't sneak in...

.github/workflows/test_branches.yml
jsiirola: Mostly because it took too long to find/derive the URLs and this documents it (way easier than remembering that I *had* figured it out then having to do log archeology)?  Plus I have hope that in the future the self-extracting zips may come back?

.github/workflows/test_branches.yml
```diff
@@ -477,61 +483,73 @@ jobs:
...
+                curl $URL/idaes-solvers-ubuntu2204-x86_64.tar.gz > "$IPOPT_TAR"
             elif test "${{matrix.TARGET}}" == osx; then
-                curl --max-time 150 --retry 8 \
-                    -L $URL/idaes-solvers-darwin-aarch64.tar.gz \
-                    -o $IPOPT_TAR
+                curl $URL/idaes-solvers-darwin-aarch64.tar.gz > "$IPOPT_TAR"
             else
-                curl --max-time 150 --retry 8 \
-                    -L $URL/idaes-solvers-windows-x86_64.tar.gz \
-                    $URL/idaes-lib-windows-x86_64.tar.gz > $IPOPT_TAR
```
blnicho, line 502: We no longer need `idaes-lib` on Windows?

pyomo/scripting/plugins/download.py
```diff
@@ -47,49 +43,55 @@ def _call_impl(self, args, unparsed, logger):
...
         )
 
+        returncode = 0
         for target in DownloadFactory:
-            attempt = 0
-            result = "FAIL"
-
-            while attempt < MAX_RETRIES:
+            attempt = 1
+            while attempt <= args.retry:
```
blnicho, line 51: Do we constrain `args.retry` to always be at least 1? Seems like we might get unexpected behavior if `args.retry` was ever set to 0 which might be a reasonable thing to do if you want to attempt the download only once.

.github/workflows/test_branches.yml
mrmundt: I am squinting at you but won't complain any further.

mrmundt (approved):

blnicho (commented):
I have a couple (hopefully easy) questions

.github/workflows/test_branches.yml
mrmundt: I think the better question here is, "Did we ever actually need it in the first place?", and I think the answer is "no."

pyomo/scripting/plugins/download.py
mrmundt: Extremely good point. Fixed in most recent commit.

jsiirola: But did you add a test?  ;)

mrmundt: I refuse, get outta here.

.github/workflows/test_branches.yml
jsiirola: I think there was a time when we *did* need it (we were distributing some "standard libraries" that are always present on *nix).  I think at some point we started compiling things more statically, which cut out the need for the extra libraries (or we changed how things were packaged so that the libraries we needed for the solvers were bundled with the solvers).  Either way, I don't think there is anything in `idaes-lib` that we need for Pyomo testing.

pyomo/scripting/plugins/download.py
mrmundt: (For posterity, I added the test)

blnicho (approved):

PR #3789, opened 2025-11-20
Simplify _Q_opt in parmest with block scenario structure.

sscini:
## Summary/Motivation:
_Q_opt, the main model building and solving function within parmest, is still heavily dependent on MPISPPY, and the code is becoming outdated and difficult to interpret. The goal of this PR is to redesign _Q_opt to work without this dependence, retaining all the current functionality

## Changes proposed in this PR:
-_Q_opt redesign, using a block structure, with minimal changes needed to functions that utilize it.

### Legal Acknowledgement

By contributing to this software project, I have read the [contribution guide](https://pyomo.readthedocs.io/en/stable/contribution_guide.html) and agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #3866, opened 2026-03-02
[Pyomo.DoE] Add simultaneous design of multiple experiments

smondal13:
## Summary/Motivation:
This PR adds a new `DesignOfExperiments.optimize_experiments()` API in `pyomo/contrib/doe/doe.py` to support simultaneous optimization of multiple experiments in one workflow. The motivation is to provide a  multi-experiment DoE interface with stronger initialization options, clearer mode handling (template vs. user-initialized experiments), and richer diagnostics/results than the existing single-experiment path.

- Added API `optimize_experiments()` for multi-experiment DoE optimization.
- Added Cholesky-based D- and A-optimality objective and Greybox-based D-, A-, E-, and ME- optimality objective
- Implemented two operating modes:
  - Template mode: pass one experiment and set `n_exp`.
  - User-initialized mode: pass a list of experiments; `n_exp` is inferred/validated.
- Added optional LHS-based initialization (`initialization_method="lhs"`) with controls for:
  - sample count, seed, candidate evaluation parallelism, combination fim metric scoring parallelism,
  - worker count, chunk size, parallel threshold, and optional wall-clock budget.
- Added symmetry-breaking constraints for multi-experiment solves:
  - supports user-specified variable through `sym_break_cons` suffix,
  - falls back to the first experiment input with a diagnostic warning when not provided.
- Expanded output for this API:
  - per-scenario and per-experiment results (designs, outputs, measurement errors, FIM/sensitivities),
  - aggregated FIM metrics, timings, settings, names, diagnostics, and structured `run_info`.
- Added JSON-safe serialization via `_DoEResultsJSONEncoder` for numpy/Pyomo-enum values when writing `results_file`.

## Note:
- Added new `documentation.md` which describes the API. This documentation is to help the reviewers to understand the API and will not be merged into `Pyomo:main`

## Remove before merging
- **documentation.md**
- **rb_multi.py**

### Legal Acknowledgement


pyomo/contrib/doe/tests/test_doe_build.py
```diff
@@ -97,9 +104,11 @@ def get_FIM_FIMPrior_Q_L(doe_obj=None):
         for i in model.output_names
         for j in model.parameter_names
     ]
-    sigma_inv = [1 / v for k, v in model.scenario_blocks[0].measurement_error.items()]
+    sigma_inv = [
+        1 / v for k, v in model.fd_scenario_blocks[0].measurement_error.items()
```
smondal13, line 128: I have renamed the block from `scenario_blocks` to `fd_scenario_blocks` to make it clearer and avoid ambiguity with `param_scenario_blocks`, which we will use in the near future.

pyomo/contrib/doe/doe.py
```diff
@@ -75,10 +86,38 @@ class FiniteDifferenceStep(Enum):
...
+            ObjectiveLib.determinant,
+            ObjectiveLib.pseudo_trace,
+            ObjectiveLib.minimum_eigenvalue,
+        }
+    )
+
     def __init__(
         self,
         experiment=None,
+        experiment_list=None,
```
smondal13: @adowling2 , @blnicho Should we keep both `experiment` and `experiment_list`? The user can pass the `experiment_list` argument if the user wants to pass the user-initialized experiment object. Another option is to keep the `experiment` argument and make it able to accept `list` objects as well, and maybe use a hint, e.g., `experiment: list`. 


Currently, I have added a deprecation warning for the `experiment` argument.

pyomo/contrib/doe/doe.py
```diff
@@ -186,17 +234,43 @@ def __init__(
...
         """
-        if experiment is None:
-            raise ValueError("Experiment object must be provided to perform DoE.")
+        # Handle backward compatibility - experiment -> experiment_list
+        if experiment is not None:
+            warnings.warn(
+                "The 'experiment' parameter in DesignOfExperiments is deprecated and "
+                "will be removed in a future version. "
+                "Please use 'experiment_list' instead.",
+                DeprecationWarning,
```
smondal13: @blnicho Is this `DeprecationWarning` acceptable, or is there a different approach Pyomo recommends?

pyomo/contrib/doe/doe.py
```diff
@@ -537,6 +610,1564 @@ def run_doe(self, model=None, results_file=None):
...
+                "not yet supported."
+            )
+        # Add parameter scenario blocks to the model
+        self.model.param_scenario_blocks = pyo.Block(range(n_param_scenarios))
+        symmetry_breaking_info = {
+            "enabled": n_exp > 1,
+            "variable": None,
+            "source": None,
+        }
+        diagnostics_warnings = []
```
smondal13: I want to propagate the warning to the final result, which may help the user. If this is not a good choice, let me know @adowling2, @blnicho

smondal13:
@blnicho @adowling2 This PR is ready for an initial review. Please take a look when you have a moment.

pyomo/contrib/doe/doe.py
```diff
@@ -75,10 +86,43 @@ class FiniteDifferenceStep(Enum):
     backward = "backward"
 
 
+class InitializationMethod(Enum):
+    lhs = "lhs"
+
+
+class _DoEResultsJSONEncoder(json.JSONEncoder):
```
smondal13, line 80: Numpy floats specifically are shown in a weird format, which is why everything is converted to Python objects.

pyomo/contrib/doe/doe.py
```diff
@@ -537,6 +619,1547 @@ def run_doe(self, model=None, results_file=None):
...
+        n_exp: int = None,
+        init_method=None,
+        init_n_samples: int = 5,
+        init_seed: int = None,
+        init_parallel: bool = False,
+        init_combo_parallel: bool = False,
+        init_n_workers: int = None,
+        init_combo_chunk_size: int = 5000,
+        init_combo_parallel_threshold: int = 20000,
+        init_max_wall_clock_time: float = None,
```
smondal13: @adowling2, @blnicho, It may be worth adding an `init_solver = None` where the user can add flexible solver options that will be passed for initialization. It may save time for initialization. If the user does not pass that, then the default solver from `DesignOfExperiments` will be used. Let me know if this is a good idea.

pyomo/contrib/doe/doe.py
```diff
@@ -537,6 +619,1547 @@ def run_doe(self, model=None, results_file=None):
...
+        self.model.param_scenario_blocks = pyo.Block(range(n_param_scenarios))
+        symmetry_breaking_info = {
+            "enabled": n_exp > 1,
+            "variable": None,
+            "source": None,
+        }
+        diagnostics_warnings = []
+
+        # Add experiment(s) for each scenario
+        # TODO: Add s_prev = 0 to handle parameter scenarios
```
smondal13: This is the extension that I will be working on next. So, I have added this as a reference for myself. This is the reason we are using `param_scenarios`, although we are not using it this time. It does not interfere with the user interface.

smondal13:
@blnicho One of the tests is failing.

pyomo/contrib/doe/tests/_review.md
```diff
@@ -0,0 +1,362 @@
+# Review: `_DoEResultsJSONEncoder`, `_enum_label`, and `optimize_experiments()` API
```
adowling2: How was this file generated? I am looking for some additional context to help me digest the contents of this file. @smondal13

pyomo/contrib/doe/tests/test_doe_build.py
adowling2: I like this!

pyomo/contrib/doe/tests/test_doe_build.py
```diff
@@ -640,5 +664,292 @@ def test_trace_initialization_consistency(self):
...
+        solver.options["max_iter"] = 3000
+        return solver
+
+    def test_get_experiment_input_vars_direct_and_fd_fallback(self):
+        doe_obj = DesignOfExperiments(
+            experiment_list=[RooneyBieglerMultiExperiment(hour=2.0)],
+            objective_option="pseudo_trace",
+        )
+
+        model_direct = pyo.ConcreteModel()
```
adowling2, line 755: What is this block of code doing? Comments would help.

pyomo/contrib/doe/tests/test_doe_build.py
```diff
@@ -640,5 +664,292 @@ def test_trace_initialization_consistency(self):
...
+            model_fd.fd_scenario_blocks[0].x
+        ] = 3.0
+        vars_fd = doe_obj._get_experiment_input_vars(model_fd)
+        self.assertEqual(
+            [v.name for v in vars_fd], [model_fd.fd_scenario_blocks[0].x.name]
+        )
+
+    @unittest.skipIf(not ipopt_available, "The 'ipopt' command is not available")
+    def test_multi_experiment_structure_and_results(self):
+        solver = self._make_solver()
```
adowling2: What is this test doing? I recommend adding a comment block with a few-sentence description of the contents and goals of this test.

adowling2: This applies to all of the tests in this file.

pyomo/contrib/doe/doe.py
```diff
@@ -25,11 +25,20 @@
...
+    product,
+    combinations as _combinations,
+    islice as _islice,
+)
+import concurrent.futures as _cf
 import json
 import logging
 import math
+import os
+import threading
```
adowling2: We want to discuss with the Pyomo team if this is the recommended approach for threading.

pyomo/contrib/doe/doe.py
adowling2: This looks strange to me. Are you saving DoE results to a JSON?

pyomo/contrib/doe/doe.py
adowling2: I like allowing `experiment` to be either an instance of the `Experiment` class (or a child) or a list of `Experiment` objects. That seems cleaner to me.

pyomo/contrib/doe/doe.py
```diff
@@ -108,14 +147,22 @@ def __init__(
...
         ----------
-        experiment:
-            Experiment object that holds the model and labels all the components. The
-            object should have a ``get_labeled_model`` where a model is returned with
-            the following labeled sets:
+        experiment_list:
+            Experiment object(s) that hold the model and labels all the components.
+            Can be a single Experiment object or a list of Experiment objects.
+            For single experiments, you can pass the object directly: experiment_list=experiment
+            or as a list: experiment_list=[experiment].
```
adowling2: We can automate this for the user if they pass in an `Experiment` object. We can detect that case and put it into a list for them.

pyomo/contrib/doe/doe.py
```diff
@@ -108,14 +147,22 @@ def __init__(
...
+            Each object should have a ``get_labeled_model`` method that returns a model with
+            the following labeled Pyomo Suffixes:
 
               - ``unknown_parameters``,
               - ``experimental_inputs``,
               - ``experimental_outputs``
+              - ``measurement_error``.
+
+        experiment:
+            **DEPRECATED** - Use 'experiment_list' instead. This parameter will be removed
```
adowling2: I do not think we should do this.

pyomo/contrib/doe/doe.py
adowling2: See my comments above.

pyomo/contrib/doe/doe.py
adowling2: I have been thinking something similar to this... it would be helpful to allow for different solver options: initialization and the full optimization solve.

pyomo/contrib/doe/doe.py
```diff
@@ -537,6 +619,1555 @@ def run_doe(self, model=None, results_file=None):
             with open(results_file, "w") as file:
                 json.dump(self.results, file)
 
+    def optimize_experiments(
```
adowling2, line 656: Is this function new? Does it replace an exisiting function?

adowling2 (commented):
There is a lot of added code in this PR. I think it would be helpful to draft a documentation page explaining the key features of this proposed interface. This would be a simple `.md` file that eventually becomes a RTD page. I find that drafting documentation helps provide more robust motivation and context to engage with the PR.

pyomo/contrib/doe/tests/_review.md
smondal13: This was for me to review. I forgot to remove this. I asked `AI` to review the code for edge cases and best practices.

pyomo/contrib/doe/tests/test_doe_build.py
smondal13: This is used to check the build structure of the new method. This one particularly tests the helper function that returns the `experiment_input` vars

pyomo/contrib/doe/doe.py
smondal13: That is one of the options. The user can choose to save the result.

pyomo/contrib/doe/doe.py
smondal13: Yes, I did that. So, now I am implementing your previous suggestion, `experiment: list`; however, it will accept either a list or a single experiment object and will remove the deprecationwarning

pyomo/contrib/doe/doe.py
smondal13: Yes, since we are keeping `experiment` argument, we can remove this.

pyomo/contrib/doe/doe.py
smondal13: This is a new function. Instead of writing this code multiple times, I have added a function that just do this.

pyomo/contrib/doe/rb_multi.py
```diff
@@ -0,0 +1,206 @@
+#### NO NEED TO CHECK THIS SCRIPT. THIS IS FOR ME TO UNDERSTAND THE RESULT AND WILL BE DELETED LATER.
```
smondal13: NO NEED TO REVIEW THIS SCRIPT

pyomo/contrib/doe/doe.py
```diff
@@ -25,18 +25,20 @@
...
-from itertools import permutations, product
-
+from itertools import permutations, product, combinations as _combinations
 import json
 import logging
 import math
+import time
+import warnings
 
 from pyomo.common.dependencies import (
```
smondal13, line 34: @blnicho , I think the current imports will have deferred imports, right? I have removed the `if scipy_availble:` statement and moved those imports inside the method

smondal13:
@adowling2 @slilonfe5 I have removed the `LHS parallel initialization` code from `optimize_experiments()`  in `doe.py`. Feel free to review it. 

@blnicho, considering the May 15 release, we have decided to remove the `LHS parallel initialization` code that we had. I have also implemented the deferred import. Please review at your earliest convenience.

pyomo/contrib/doe/tests/test_doe_build.py
```diff
@@ -640,5 +709,394 @@ def test_trace_initialization_consistency(self):
...
+            self.assertGreaterEqual(point[0], 1.0)
+            self.assertLessEqual(point[0], 10.0)
+
+        # For Rooney-Biegler there is one design variable (hour), so:
+        # n_candidates = lhs_n_samples^1 and n_combinations = C(n_candidates, n_exp).
+        self.assertEqual(lhs_init["number_of_candidate_designs"], lhs_n_samples)
+        self.assertEqual(lhs_init["number_of_design_combinations"], 3)
+
+
+class TestDoeResultsSerialization(unittest.TestCase):
```
mrmundt, line 1058: Should this be protected by a "if numpy available" check?

pyomo/contrib/doe/tests/test_doe_build.py
```diff
@@ -640,5 +709,394 @@ def test_trace_initialization_consistency(self):
...
+        }
+        encoded = json.dumps(payload, cls=_DoEResultsJSONEncoder)
+        decoded = json.loads(encoded)
+
+        self.assertEqual(decoded["scalar"], 7)
+        self.assertEqual(decoded["array"], [1.0, 2.0])
+        self.assertEqual(decoded["objective"], str(ObjectiveLib.trace))
+
+
+class TestDoeBuildHelpers(unittest.TestCase):
```
mrmundt, line 1078: Should this be protected by a "if numpy available" check?

pyomo/contrib/doe/tests/test_doe_errors.py
```diff
@@ -762,6 +808,36 @@ def test_bad_compute_FIM_option(self):
...
+                hour=3.5, y=12.0, theta={'asymptote': 16, 'rate_constant': 0.5}
+            ),
+        ]
+        doe_obj = DesignOfExperiments(**DoE_args)
+
+        def _fake_sequential(*args, **kwargs):
+            # This is only used if execution reaches the FIM solve call.
+            doe_obj.seq_FIM = np.eye(2)
+
+        with patch.object(doe_obj, "_sequential_FIM", side_effect=_fake_sequential):
```
mrmundt: We aren't that keen on using mock / patch (see https://github.com/Pyomo/pyomo/pull/3789#discussion_r3219623927). Please rework any tests in this PR to _not_ use mock/patch unless absolutely necessary. Happy to provide more details if you'd like.

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -54,6 +59,38 @@
 ipopt_available = SolverFactory("ipopt").available()
 k_aug_available = SolverFactory("k_aug", solver_io="nl", validate=False)
 
+
+def k_aug_runtime_available():
```
mrmundt: Why is this also necessary (in addition to `k_aug_available` two lines up)? Theoretically it won't return "available" from the solver factory if it's not actually runnable.

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -383,12 +425,58 @@ def test_compute_FIM_seq_forward(self):
...
+            model = kwargs.get("model")
+            hour = float(pyo.value(model.hour))
+            if np.isclose(hour, 1.5):
+                doe_multi.seq_FIM = fim_exp1.copy()
+            elif np.isclose(hour, 3.5):
+                doe_multi.seq_FIM = fim_exp2.copy()
+            else:
+                raise RuntimeError(f"Unexpected hour value in mocked test: {hour}")
+
+        with patch.object(
```
mrmundt: For visibility - same comment about mocking / patch in this file.

pyomo/contrib/doe/tests/test_greybox.py
```diff
@@ -346,6 +350,138 @@ def make_greybox_and_doe_objects_rooney_biegler(objective_option):
...
+    def __getattr__(self, name):
+        return getattr(self._solver, name)
+
+
+def _make_ipopt_solver():
+    solver = SolverFactory("ipopt")
+    solver.options["linear_solver"] = "ma57"
+    solver.options["halt_on_ampl_error"] = "yes"
+    solver.options["max_iter"] = 3000
+    return solver
```
mrmundt: This same code is also in at least test_doe_errors. It is probably worth consolidating duplicate helper functions like this into something like a utils.py file in the tests directory and then calling from there. That way you don't have to modify a bunch of files if you decide to change something like ipopt options of max number of iterations in the future.

pyomo/contrib/doe/tests/test_greybox.py
```diff
@@ -1202,5 +1339,600 @@ def test_solve_ME_optimality_condition_number(self):
...
+            places=7,
+        )
+
+    def test_optimize_experiments_greybox_outputs_match_numpy_for_all_supported_objectives(
+        self,
+    ):
+        # Validate the deterministic wiring: for each supported greybox metric,
+        # the external block output should match direct NumPy on scenario.total_fim.
+        prior_fim = np.array([[6.0, 0.75], [0.75, 4.0]])
+        for objective_option in (
```
mrmundt: The way you did this is technically fine, but if you wanted to, we pretty commonly use `parametrize` as a way to do exactly what you're doing here. (See parmest for an example: https://github.com/Pyomo/pyomo/blob/main/pyomo/contrib/parmest/tests/test_parmest.py#L50)

pyomo/contrib/doe/documentation.md
```diff
@@ -0,0 +1,175 @@
+# DesignOfExperiments `optimize_experiments()` Proposed Interface
+
+This is a brief documentation explaining the changes. This document will not get merged.
```
mrmundt: This probably wasn't supposed to be committed?

pyomo/contrib/doe/grey_box_utilities.py
```diff
@@ -93,15 +118,21 @@ def __init__(self, doe_object, objective_option="determinant", logger_level=None
 
         # If logger level is None, use doe_object's logger level
         if logger_level is None:
-            logger_level = doe_object.logger.level
+            logger_level = (
+                doe_object.logger.level if doe_object is not None else logging.WARNING
+            )
```
mrmundt: This seems like a kind of convoluted way to do this. Why are there multiple ways to set the logger level?

pyomo/contrib/doe/doe.py
```diff
@@ -75,10 +72,37 @@ class FiniteDifferenceStep(Enum):
     backward = "backward"
 
 
+class InitializationMethod(Enum):
+    lhs = "lhs"
```
mrmundt: Assuming a different method will be added later? (like `rhs`)

pyomo/contrib/doe/doe.py
```diff
@@ -186,17 +215,33 @@ def __init__(
...
-        self.experiment = experiment
+        # Auto-convert single experiment to list
+        if not isinstance(experiment, list):
+            experiment_list = [experiment]
+        else:
+            experiment_list = experiment
+
+        # Validate list is not empty
+        if len(experiment_list) == 0:
+            raise ValueError("The 'experiment' list cannot be empty.")
```
mrmundt: This seems reversed. Wouldn't you want to error if they don't pass in an experiment first (because that's essential), then convert to a list, assuming something is given?

pyomo/contrib/doe/doe.py
```diff
@@ -537,6 +622,1235 @@ def run_doe(self, model=None, results_file=None):
             with open(results_file, "w") as file:
                 json.dump(self.results, file)
 
+    def optimize_experiments(
```
mrmundt, line 656: This is _huge_ and needs to be split up.

pyomo/contrib/doe/doe.py
```diff
@@ -1349,6 +2787,15 @@ def create_objective_function(self, model=None):
...
             model = self.model
 
+        if self.objective_option in [
+            ObjectiveLib.minimum_eigenvalue,
+            ObjectiveLib.condition_number,
+        ]:
+            raise ValueError(
+                f"objective_option='{self.objective_option.value}' requires "
+                "use_grey_box_objective=True."
+            )
```
mrmundt, line 2792: This check confuses me. When is `use_grey_box_objective` set? Shouldn't this check be "if option in this set AND use_grey_box_objective=False"?

pyomo/contrib/doe/doe.py
```diff
@@ -1699,30 +3502,71 @@ def check_model_labels(self, model=None):
...
+                "'experiment_inputs' Suffix to your model with at least one variable."
+            )
+
         # Check that unknown parameters exist
         try:
-            outputs = [k.name for k, v in model.unknown_parameters.items()]
+            parameters = [k.name for k, v in model.unknown_parameters.items()]
         except:
             raise RuntimeError(
                 "Experiment model does not have suffix " + '"unknown_parameters".'
```
mrmundt: I know this isn't part of this PR but this looks like a weird line split that might have happened because of `black`

pyomo/contrib/doe/tests/test_doe_build.py
smondal13: yes, added

pyomo/contrib/doe/tests/test_doe_build.py
smondal13: added the check

pyomo/contrib/doe/tests/test_doe_errors.py
smondal13: Reworking on mock/patch. Will keep updating

pyomo/contrib/doe/tests/test_doe_solve.py
smondal13: Removed the mock test

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -383,12 +390,43 @@ def test_compute_FIM_seq_forward(self):
...
+            np.all(np.isfinite(np.array(doe_multi._computed_FIM_by_experiment[1])))
+        )
+
     # This test ensure that compute FIM runs without error using the
     # `kaug` option. kaug computes the FIM directly so no finite difference
     # scheme is needed.
     @unittest.skipIf(not scipy_available, "Scipy is not available")
     @unittest.skipIf(
-        not k_aug_available.available(False), "The 'k_aug' command is not available"
+        not k_aug_available, "The 'k_aug' is not available in this environment"
```
smondal13, line 423: @mrmundt this is probably fine, right? I mean, do we need exactly `not k_aug_available.available(False)` for the code to work?

pyomo/contrib/doe/tests/test_greybox.py
smondal13: Agreed. I have consolidated the IPOPT test solver construction into a shared helper under `pyomo/contrib/doe/tests/utils_for_doe_tests.py` and updated DOE tests to use it, so option changes are centralized.

pyomo/contrib/doe/tests/test_greybox.py
smondal13: Thanks for the suggestion. I switched this to a parameterized test and now run one case per objective. Please take a look to ensure that I am doing it correctly.

pyomo/contrib/doe/documentation.md
smondal13: It was intentional. This markdown is intended to help reviewers understand the math and arguments. @adowling2 suggested this. This will be removed before merging.

pyomo/contrib/doe/grey_box_utilities.py
smondal13: Good point. If `doe_object` and `logger_level` are not passed, then we will have no logger level set here. In that case, it will default to `WARNING` which was the intent.

pyomo/contrib/doe/doe.py
smondal13: Yes.  "lhs" stands for `Latin Hypercube Sampling". In future extensions, we can add other sampling techniques (e.g. Sobol) .

pyomo/contrib/doe/doe.py
smondal13: @mrmundt we are ensuring that `experiment` exists before this in line 219-220

pyomo/contrib/doe/doe.py
smondal13: I have split it up to multiple helper functions.

pyomo/contrib/doe/doe.py
slilonfe5: @smondal13 I suggest that you use a more descriptive name, e.g., latin_hypercube_sampling

blnicho: I agree with the more descriptive name. I definitely thought "lhs" was "left hand side" upon first glance.

slilonfe5: I figured that was the understanding when Miranda thought there would be a “rhs” (right hand side)

pyomo/contrib/doe/doe.py
smondal13: You’re right about the intent. `use_grey_box_objective` is set in `__init__` (stored as `self.use_grey_box`), and this function handles both greybox and non-greybox paths. The current check is in the `else` branch (`self.use_grey_box == False`), so it is effectively “if objective is E-opt/ME-opt and `use_grey_box_objective=False`, raise.” We have the following code:

```
        # Add the objective function to the model
        if self.use_grey_box:
            self.create_grey_box_objective_function(model=model)
        else:
            self.create_objective_function(model=model)
```

pyomo/contrib/doe/doe.py
smondal13: I have removed the string addition `+` and made the whole string a single line. There were 3 other instances of this, fixed those as well.

pyomo/contrib/doe/doe.py
smondal13: Good suggestion @slilonfe5. I have added a more descriptive name.

smondal13:
@mrmundt @blnicho , I have addressed the comments from @mrmundt. Wating for the CI tests to pass

pyomo/contrib/doe/doe.py
mrmundt: Yup, I fell victim to assuming we're always talking about sides of an equation. Thanks for renaming!

pyomo/contrib/doe/tests/test_doe_solve.py
mrmundt: No, this is perfectly fine. I actually infinitely prefer this - much easier to read.

pyomo/contrib/doe/tests/test_greybox.py
```diff
@@ -10,6 +10,7 @@
 import itertools
 import json
 import os.path
+from parameterized import parameterized
```
mrmundt: Oh I should have warned you, apologies. We normally add a guard just in case `parameterized` isn't available (we love our guards):
```
parameterized, param_available = attempt_import('parameterized')

if not param_available:
    raise unittest.SkipTest('Parameterized is not available.')
```

pyomo/contrib/doe/doe.py
mrmundt: Gotcha. Maybe you should in that earlier step be checking `if experiment is None or []` - if you didn't supply one or you gave us an empty list.

pyomo/contrib/doe/doe.py
mrmundt: Much better, thank you!

pyomo/contrib/doe/doe.py
mrmundt: Okay, cool. As long as both cases are being tested, then this is fine.

pyomo/contrib/doe/doe.py
smondal13: Yes, I think that will be cleaner. I have added the following immediately after the constructor.
```
        if experiment is None or (isinstance(experiment, list) and not experiment):
            raise ValueError(
                "The 'experiment' argument is required and cannot be an empty list. "
                "Pass a single Experiment object or a non-empty list of Experiment "
                "objects."
            )
 ```

File: pyomo/contrib/doe/tests/test_greybox.py (hunk 32 above)
Comment (smondal13) [reply pyomo/contrib/doe/tests/test_greybox.py]: When I used 
```
parameterized, param_available = attempt_import('parameterized')
```

It throws
```
ERROR test_greybox.py - AttributeError: module 'parameterized' has no attribute 'expand'
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
=============================== 1 error in 1.48s ===============================
```

I am not sure why that is. Now, I use the following as a workaround.
```
try:
    from parameterized import parameterized

    parameterized_available = True
except ImportError:
    parameterized_available = False

@unittest.skipIf(
        not parameterized_available, "The 'parameterized' package is not available"
    )
```
Is this acceptable?

Comment (smondal13) [reply pyomo/contrib/doe/tests/test_greybox.py]: The issue is fixed.  `attempt_import` is fine.

**Comment by smondal13** [comment]
@blnicho @mrmundt  All the tests are passing except `win/3.11` with
```
INFO: The following extensions were downloaded:
    [FAIL]  gjh
    [ OK ]  mcpp
 ```

pyomo/contrib/doe/tests/test_doe_errors.py
```diff
@@ -84,19 +103,36 @@ def get_standard_args(experiment, fd_method, obj_used, flag):
...
         ):
             # Experiment provided as None
             DoE_args = get_standard_args(None, fd_method, obj_used, flag_val)
 
             doe_obj = DesignOfExperiments(**DoE_args)
 
+    def test_experiment_empty_list_error(self):
+        with self.assertRaisesRegex(
+            ValueError,
+            "The 'experiment' argument is required and cannot be an empty " "list",
```
mrmundt: This is definitely another "black did weird things with whitespace" situation.
```suggestion
            "The 'experiment' argument is required and cannot be an empty list",
```

pyomo/contrib/doe/tests/test_greybox.py
```diff
@@ -1202,5 +1327,597 @@ def test_solve_ME_optimality_condition_number(self):
...
+        self,
+    ):
+        # Rooney-Biegler integration check for the LHS candidate-combination
+        # scorer on greybox-only objectives (E-opt and ME-opt). Compare the
+        # selected initial designs against an independent exhaustive reference
+        # built from real candidate FIM evaluations.
+        lhs_n_samples = 4
+        lhs_seed = 19
+
+        for objective_option in ("minimum_eigenvalue", "condition_number"):
```
mrmundt: Did you want to also convert this one and the other usage of subTest in L#1904 to use parameterize?

smondal13: Yes, I have added a parameterized test here. And, replaced other subtests in `doe/tests` with parameterized tests.

smondal13:
@mrmundt and @blnicho , all the comments are addressed, and all the tests are passing now.

slilonfe5:
@mrmundt @blnicho, Shuvo has addressed all the comments. This looks pretty good to me. Any final review would be appreciated; otherwise, I think we are good to go.

pyomo/contrib/doe/documentation.md
slilonfe5: @smondal13 @mrmundt Maybe we should remove this before merging?

smondal13: @slilonfe5 , yes that is the plan. I will remove `rb_multi.py` script as well.  @mrmundt should we remove these two scripts now?

mrmundt (approved):

adowling2:
@smondal13 This PR is now showing merge conflicts.

pyomo/contrib/doe/tests/test_doe_build.py
```diff
@@ -640,5 +705,401 @@ def test_trace_initialization_consistency(self):
...
+        self.assertIn("best_initial_objective_value", lhs_init)
+        self.assertIsInstance(lhs_init["best_initial_objective_value"], float)
+        self.assertTrue(np.isfinite(lhs_init["best_initial_objective_value"]))
+        self.assertGreater(lhs_init["best_initial_objective_value"], 0.0)
+        self.assertIn("best_initial_objective_value_log10", lhs_init)
+        self.assertIsInstance(lhs_init["best_initial_objective_value_log10"], float)
+        self.assertAlmostEqual(
+            lhs_init["best_initial_objective_value_log10"],
+            np.log10(lhs_init["best_initial_objective_value"]),
+            places=12,
```
adowling2: Do we really want to test this to 12 places? This could make the test very fragile, especially across platforms and Ipopt versions.

pyomo/contrib/doe/tests/test_doe_build.py
```diff
@@ -640,5 +705,401 @@ def test_trace_initialization_consistency(self):
...
+            experiment=[RooneyBieglerMultiExperiment(hour=2.0)],
+            objective_option="pseudo_trace",
+            step=1e-2,
+            solver=self._make_solver(),
+        )
+        doe_obj.optimize_experiments(
+            n_exp=2,
+            init_method="latin_hypercube_sampling",
+            init_n_samples=2,
+            init_seed=11,
```
adowling2, line 1001: I am 99% sure you are setting the random seed here to make the example deterministic. I suggest adding a comment "# setting random seed to make the test deterministic" everywhere you are doing this. I can also guarantee a future Pyomo.DoE developer will ask "why are we setting the random seed"?

pyomo/contrib/doe/tests/test_doe_build.py
```diff
@@ -640,5 +705,401 @@ def test_trace_initialization_consistency(self):
...
+            experiment=[RooneyBieglerMultiExperiment(hour=2.0)],
+            objective_option="pseudo_trace",
+            step=1e-2,
+            solver=self._make_solver(),
+        )
+        doe_obj.optimize_experiments(
+            n_exp=n_exp,
+            init_method="latin_hypercube_sampling",
+            init_n_samples=lhs_n_samples,
+            init_seed=17,
```
adowling2, line 1040: See comment above about comment to explain this line

adowling2 (approved):
I think this is practically ready. I requested a few small comments. Also, there is a merge conflict (I think).

pyomo/contrib/doe/tests/test_doe_build.py
smondal13: Yes. The seed makes the random number generator deterministic. Now, I have added comments everywhere I used the seed.

pyomo/contrib/doe/tests/test_doe_build.py
smondal13: That's a good point. I have now added up to 4 decimal places.

pyomo/contrib/doe/tests/test_doe_build.py
smondal13: Resolved

smondal13:
> I think this is practically ready. I requested a few small comments. Also, there is a merge conflict (I think).

@adowling2 I have addressed your comments. @adowling2 @mrmundt, I have resolved the conflicts. The conflicts were from @adowling2's recent PR #3867 merge into `Pyomo:main`.

smondal13:
@mrmundt @blnicho, the following two tests are failing. I think these two are not part of my PR
```
FAILED pyomo/contrib/iis/tests/test_iis.py::TestIIS::test_write_iis_any_solver - SystemError: <method 'write' of '_io.FileIO' objects> returned a result with an exception set
FAILED pyomo/contrib/iis/tests/test_iis.py::TestIIS::test_write_iis_cplex - SystemError: <method 'write' of '_io.FileIO' objects> returned a result with an exception set
```

blnicho:
@smondal13 those failures were resolved in #3965 and I merged main into your PR to pick up the fix.

smondal13:
@blnicho @mrmundt this one is failing due to download issue


downloaded:
    [FAIL]  gjh
    [ OK ]  mcpp
Error: Process completed with exit code 1.

smondal13:
@blnicho @mrmundt Two tests are failing in this PR, which are not related to the PR.

pyomo/contrib/doe/tests/experiment_class_example_flags.py
```diff
@@ -70,3 +70,80 @@ def get_labeled_model(self):
...
 
         return m
+
+
+class RooneyBieglerMultiExperiment(RooneyBieglerExperiment):
+    """
+    Experiment class based on the multi-experiment Rooney-Biegler prototype.
+
+    This mirrors the implementation in
+    ``examples/multiexperiment-prototype/rooney_biegler_multiexperiment.py``
```
blnicho: I don't see this example file in your branch or this PR. Did you mean to include an example in this PR? Or is this docstring out of date?

pyomo/contrib/doe/tests/experiment_class_example_flags.py
```diff
@@ -70,3 +70,80 @@ def get_labeled_model(self):
...
+        m = super().get_labeled_model()
+        hour_lb, hour_ub = self.hour_bounds
+        m.hour.setlb(hour_lb)
+        m.hour.setub(hour_ub)
+
+        if hasattr(m, "sym_break_cons"):
+            m.sym_break_cons.clear()
+        else:
+            m.sym_break_cons = pyo.Suffix(direction=pyo.Suffix.LOCAL)
+        m.sym_break_cons[m.hour] = None
```
blnicho, line 101: Why is this needed? Are you adding a new required `sym_break_cons` Suffix as part of this PR?

pyomo/contrib/doe/tests/experiment_class_example_flags.py
```diff
@@ -70,3 +70,80 @@ def get_labeled_model(self):
...
+        m.temp.fix()
+
+        # Replace base Rooney-Biegler response with two-input synthetic variant
+        # used only for symmetry-breaking tests.
+        m.del_component(m.response_function)
+        m.response_function = pyo.Constraint(
+            expr=m.y
+            == m.asymptote * (1 - pyo.exp(-m.rate_constant * m.hour)) + 0.01 * m.temp
+        )
+        m.experiment_inputs[m.temp] = self.temp
```
blnicho, line 141: If this is only used for symmetry-breaking tests, should this code be moved inside the if-statement below that is checking if the `sym_break_flag` is set?

pyomo/contrib/doe/tests/test_doe_errors.py
```diff
@@ -21,17 +23,41 @@
...
+)
+
 if scipy_available:
     from pyomo.contrib.doe import DesignOfExperiments
+    from pyomo.contrib.doe.doe import InitializationMethod, _DoEResultsJSONEncoder
     from pyomo.contrib.doe.tests.experiment_class_example_flags import (
         BadExperiment,
         RooneyBieglerExperimentFlag,
+        RooneyBieglerMultiExperiment,
+        RooneyBieglerMultiInputExperimentFlag,
```
blnicho: Looks like these imports are duplicated and can be removed.

pyomo/contrib/doe/tests/test_doe_build.py
```diff
@@ -633,5 +705,404 @@ def test_trace_initialization_consistency(self):
                 self.assertAlmostEqual(val, expected, places=4)
 
 
+@unittest.skipIf(not pandas_available, "pandas is not available")
+class TestOptimizeExperimentsBuildStructure(unittest.TestCase):
+    """Coverage for optimize_experiments() build, output, and diagnostics behavior."""
+
+    def _make_solver(self):
+        return make_ipopt_solver()
```
blnicho: Why do you need the `_make_solver` method? Why not just call the `make_ipopt_solver` function directly in the tests instead of going through this method?

pyomo/contrib/doe/tests/test_doe_build.py
```diff
@@ -633,5 +705,404 @@ def test_trace_initialization_consistency(self):
...
+            experiment=[RooneyBieglerMultiExperiment(hour=2.0)],
+            objective_option="pseudo_trace",
+            step=1e-2,
+            solver=main_solver,
+        )
+
+        doe_obj.optimize_experiments(n_exp=2, init_solver=init_solver)
+
+        # The exact number of initialization solves is implementation-dependent,
+        # but they must all occur before the one final main-solver call.
```
blnicho: Shouldn't this number be deterministic and known for this test?

pyomo/contrib/doe/tests/test_doe_build.py
```diff
@@ -633,5 +705,404 @@ def test_trace_initialization_consistency(self):
...
+            init_solver.calls, 1
+        )  # At least one initialization solve
+        self.assertEqual(main_solver.calls, 1)  # Exactly one main optimization solve
+        self.assertEqual(call_order[-1], "main")
+        self.assertTrue(all(tag == "init" for tag in call_order[:-1]))
+        # Distinct option markers provide a second check that solver routing
+        # matches the expected init-versus-final phase split.
+        option_markers = init_solver.option_markers + main_solver.option_markers
+        self.assertTrue(all(marker == 123 for marker in option_markers[:-1]))
+        self.assertEqual(option_markers[-1], 321)
```
blnicho: I don't think checking `option_markers` is adding anything to this test. It's basically repeating what you just did in checking `call_order`. I would recommend removing this and removing the `option_markers` stuff from the `_TrackingSolver` class to avoid redundancy.

pyomo/contrib/doe/tests/test_doe_build.py
```diff
@@ -45,6 +56,26 @@
...
+        self.option_markers = []
+
+    def solve(self, *args, **kwargs):
+        self.calls += 1
+        self._call_order.append(self._phase_tag)
+        self.option_markers.append(self._solver.options.get("max_iter"))
+        return self._solver.solve(*args, **kwargs)
+
+    def __getattr__(self, name):
+        return getattr(self._solver, name)
```
blnicho: Why are you doing this through `__getattr__` instead of adding a `name` property?

pyomo/contrib/doe/tests/test_doe_build.py
```diff
@@ -633,5 +705,404 @@ def test_trace_initialization_consistency(self):
...
+        # Result payloads should report the same phase-specific solver names that
+        # were observed through the tracked solve() calls above.
+        self.assertEqual(
+            doe_obj.results["initialization"]["solver"],
+            getattr(init_solver, "name", str(init_solver)),
+        )
+        self.assertEqual(
+            doe_obj.results["optimization_solve"]["solver"],
+            getattr(main_solver, "name", str(main_solver)),
+        )
```
blnicho, line 745: Same comment about adding a `name` property to your `_TrackingSolver` to avoid using `getattr` for a simple check like this.

blnicho (commented):
@smondal13 I still have 4 more files to go through but here are my comments so far.

pyomo/contrib/doe/tests/experiment_class_example_flags.py
smondal13: I had that example but later deleted it since it was a prototype. The docstring was out of date. There isn’t a new example file in this PR, so I’ve updated the docstring to describe this as a test-only multi-experiment variant instead of referencing `examples/multiexperiment-prototype/rooney_biegler_multiexperiment.py`

pyomo/contrib/doe/tests/experiment_class_example_flags.py
smondal13: Yes — in multi-experiment design, `sym_break_cons` is an optional suffix that lets the user specify which experiment input should be used for symmetry breaking. If the user does not provide it, we automatically use the first variable in the experiment_inputs suffix.

We do this to reduce permutation symmetry and save solver time. In multi-experiment design, the order of experiments does not matter. For example, if we are designing two experiments with design variable T, then the designs (`exp 1: T = 300 K, exp 2: T = 350 K`) and (`exp 1: T = 350 K, exp 2: T = 300 K`) are equivalent. The symmetry-breaking constraint helps restrict these equivalent permutations so the solver only considers one ordering.

pyomo/contrib/doe/tests/experiment_class_example_flags.py
smondal13: We keep that code outside the `sym_break_flag` conditional because this helper is intended to be a two-input experiment in all cases. The flag only controls whether we add an explicit `sym_break_cons` suffix.

We still need the temp input and modified response when `sym_break_flag=0`, because that case is used to test the automatic fallback behavior when no explicit symmetry-breaking marker is provided. In that path, the code should infer the symmetry-breaking variable from experiment_inputs. Moving the two-input setup inside the conditional would remove that coverage.

pyomo/contrib/doe/tests/test_doe_build.py
smondal13: We only need `_TrackingSolver` to intercept `solve()` and expose the wrapped solver name for the result checks, so `__getattr__` was broader than necessary. I updated the helper to use an explicit `name` property instead.

pyomo/contrib/doe/tests/test_doe_build.py
smondal13: Agreed. `_make_solver()` was only a thin pass-through to `make_ipopt_solver()` and did not add any class-specific behavior, so I removed it and now call `make_ipopt_solver()` directly in the tests.

pyomo/contrib/doe/tests/test_doe_build.py
smondal13: In this test, `init_solver` is also used during finite-difference scenario construction and the square initialization solve, so it is expected to be called multiple times (deterministically) before the final optimization solve.

pyomo/contrib/doe/tests/test_doe_build.py
smondal13: I agree. The `option_markers` check was redundant with the recorded `call_order`, so I removed that extra bookkeeping from `_TrackingSolver` and kept the test focused on the solver-routing behavior itself.

pyomo/contrib/doe/tests/test_doe_errors.py
smondal13: Yes, those are duplicated. The file already imports them unconditionally after the early `SkipTest` guard, so the extra `if scipy_available:` import block is unnecessary. I have removed the duplicate block.

pyomo/contrib/doe/tests/test_doe_build.py
blnicho: Right but my point is that for this specific example problem you should know exactly how many times `init_solver` will be called so why does the comment imply that the exact number is "unknowable" and the test check for a positive number of calls instead of the exact number?

smondal13: You’re right. For this test, the `init_solver` call count is deterministic, so the assertion should be exact rather than a lower bound. I have added `self.assertEqual(init_solver.calls, 11)` instead of the previously used `assertGreaterEqual`.

pyomo/contrib/doe/tests/test_doe_build.py
```diff
@@ -45,6 +56,25 @@
...
+        self.calls = 0
+
+    def solve(self, *args, **kwargs):
+        self.calls += 1
+        self._call_order.append(self._phase_tag)
+        return self._solver.solve(*args, **kwargs)
+
+    @property
+    def name(self):
+        return getattr(self._solver, "name", str(self._solver))
```
blnicho:
```suggestion
        return self._solver.name
```

pyomo/contrib/doe/tests/test_doe_build.py
```diff
@@ -633,5 +704,392 @@ def test_trace_initialization_consistency(self):
...
+        # Result payloads should report the same phase-specific solver names that
+        # were observed through the tracked solve() calls above.
+        self.assertEqual(
+            doe_obj.results["initialization"]["solver"],
+            getattr(init_solver, "name", str(init_solver)),
+        )
+        self.assertEqual(
+            doe_obj.results["optimization_solve"]["solver"],
+            getattr(main_solver, "name", str(main_solver)),
+        )
```
blnicho, line 745:
```suggestion
        self.assertEqual(
            doe_obj.results["initialization"]["solver"],
            init_solver.name
        )
        self.assertEqual(
            doe_obj.results["optimization_solve"]["solver"],
            main_solver.name
        )
```

pyomo/contrib/doe/tests/test_doe_errors.py
```diff
@@ -774,10 +822,442 @@ def test_invalid_trace_without_cholesky(self):
...
+            doe_obj.optimize_experiments(**kwargs)
+
+
+@unittest.skipIf(not ipopt_available, "The 'ipopt' command is not available")
+@unittest.skipIf(not numpy_available, "Numpy is not available")
+@unittest.skipIf(not scipy_available, "scipy is not available")
+@unittest.skipIf(not pandas_available, "pandas is not available")
+class TestDoEErrorsRequiringSolver(unittest.TestCase):
+    def _make_solver(self):
+        return make_ipopt_solver()
```
blnicho: Just like in the other testing file, use the `make_ipopt_solver` function directly instead of going through this `_make_solver` method.

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -917,5 +953,621 @@ def cleanup_files():
...
+
+    def _build_template_model_for_multi_experiment(self, doe_obj, n_exp):
+        doe_obj.model.param_scenario_blocks = pyo.Block(range(1))
+        doe_obj.model.param_scenario_blocks[0].exp_blocks = pyo.Block(range(n_exp))
+        for k in range(n_exp):
+            doe_obj.create_doe_model(
+                model=doe_obj.model.param_scenario_blocks[0].exp_blocks[k],
+                experiment_index=0,
+                _for_multi_experiment=True,
+            )
```
blnicho: Should these be static methods?

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -917,5 +953,621 @@ def cleanup_files():
...
+            class _SolverResults:
+                solver = _SolverInfo()
+
+            return _SolverResults()
+
+        class _SquareSolveThenStubSolver:
+            def __init__(self, wrapped_solver):
+                self._wrapped_solver = wrapped_solver
+                self.name = getattr(wrapped_solver, "name", "stub-solver")
+                self.options = getattr(wrapped_solver, "options", {})
```
blnicho:
```suggestion
                self.name = wrapped_solver.name
                self.options = wrapped_solver.options
```

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -917,5 +953,621 @@ def cleanup_files():
...
+        total_fim = np.array(
+            _optimize_experiments_param_scenario(doe.results)["total_fim"]
+        )
+        min_eig = np.min(np.real(np.linalg.eigvals(total_fim)))
+        if min_eig < _SMALL_TOLERANCE_DEFINITENESS:
+            jitter = min(
+                -min_eig + _SMALL_TOLERANCE_DEFINITENESS, _SMALL_TOLERANCE_DEFINITENESS
+            )
+        else:
+            jitter = 0.0
```
blnicho: Why is this copying the implementation for computing the jitter instead of hard-coding the known expected value?

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -917,5 +953,621 @@ def cleanup_files():
...
+
+        best_combo = None
+        best_obj = np.inf
+        for combo in combinations(range(len(candidate_points)), n_exp):
+            fim_total = sum((candidate_fims[idx] for idx in combo), np.zeros((2, 2)))
+            obj_val = doe_ref._evaluate_objective_from_fim(fim_total)
+            if obj_val < best_obj:
+                best_obj = obj_val
+                best_combo = combo
+        expected_points = [list(candidate_points[i]) for i in best_combo]
```
blnicho: You shouldn't repeat the implementation in the test. In most cases you should be testing against hard-coded, known values.

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -917,5 +953,621 @@ def cleanup_files():
...
+
+        per_dim_samples = []
+        for i in range(len(exp_input_vars)):
+            dim_seed = int(rng.integers(0, 2**31))
+            sampler = LatinHypercube(d=1, seed=dim_seed)
+            s_unit = sampler.random(n=lhs_n_samples).flatten()
+            s_scaled = lb_vals[i] + s_unit * (ub_vals[i] - lb_vals[i])
+            per_dim_samples.append(s_scaled.tolist())
+        candidate_points = list(product(*per_dim_samples))
+        expected_points = [list(candidate_points[i]) for i in range(3)]
```
blnicho: Again, why are you copying the actual implementation into the test instead of hard-coding `expected_points`?

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -917,5 +953,621 @@ def cleanup_files():
...
+        ]
+        best_obj = -np.inf
+        best_combo = None
+        for combo in combinations(range(len(candidate_points)), n_exp):
+            fim_total = sum((fims[i] for i in combo), np.zeros((2, 2)))
+            obj = doe_ref._evaluate_objective_from_fim(fim_total)
+            if obj > best_obj:
+                best_obj = obj
+                best_combo = combo
+        expected_points = [list(candidate_points[i]) for i in best_combo]
```
blnicho: Same comment about repeating the implementation in the test.

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -917,5 +953,621 @@ def cleanup_files():
...
+        best_obj = -np.inf
+        best_combo = None
+        for combo in combinations(range(len(candidate_points)), 2):
+            f1 = _fake_fim(0, list(candidate_points[combo[0]]))
+            f2 = _fake_fim(0, list(candidate_points[combo[1]]))
+            obj_val = float(np.trace(f1 + f2))
+            if obj_val > best_obj:
+                best_obj = obj_val
+                best_combo = combo
+        expected_points = [list(candidate_points[i]) for i in best_combo]
```
blnicho: Same comment about repeating the implementation in the test

blnicho (commented):
@smondal13 here is my next set of comments. I have 2 more files to look at.

pyomo/contrib/doe/tests/test_doe_errors.py
smondal13: Done.

pyomo/contrib/doe/tests/test_doe_solve.py
smondal13: Yes - both helpers are instance-independent, so I converted them to static methods and removed the unused `self` parameter.

pyomo/contrib/doe/tests/test_doe_solve.py
smondal13: You're right.  I’ve switched it to use a synthetic singular FIM and to check the resulting regularized Cholesky matrix directly.

pyomo/contrib/doe/tests/test_doe_solve.py
smondal13: I have added a hard-coded test for the specific seed.

pyomo/contrib/doe/tests/test_doe_solve.py
smondal13: I have added hard-coded tests.

pyomo/contrib/doe/tests/test_doe_solve.py
smondal13: Added hard-coded test

pyomo/contrib/doe/tests/test_doe_solve.py
smondal13: added hard-coded test

pyomo/contrib/doe/doe.py
```diff
@@ -270,6 +312,129 @@ def __init__(
...
+            output_value = np.min(np.linalg.eigvalsh(fim_np))
+        elif self.objective_option == ObjectiveLib.condition_number:
+            eig = np.linalg.eigvalsh(fim_np)
+            output_value = np.log(np.abs(np.max(eig) / np.min(eig)))
+        else:
+            output_value = 0.0
+
+        egb_block.outputs[self._grey_box_output_name()].set_value(float(output_value))
+
+    def _initialize_standard_objective_block(self, block, fim_np, parameter_names):
```
smondal13, line 376: @blnicho I have removed `_initialize_cholesky_from_fim()` and consolidated the shared FIM-derived initialization updates into `_initialize_standard_objective_block()`. Both the single- and multi-experiment designs were doing the same initialization work (L, L_inv, fim_inv, cov_trace, determinant/pseudo-trace state), so consolidating that logic avoids duplication and reduces the risk of the two paths drifting out of sync.

smondal13:
@blnicho @mrmundt  can you run this again? It is showing the `gjh` failure

smondal13:
@blnicho , @mrmundt , can you please run the tests again?

PR #3867, opened 2026-03-04
Pyomo.DoE: fix trace/Cholesky initialization consistency

adowling2:
## Summary/Motivation:
This PR ~~adds advanced debugging/inspection controls for `pyomo.contrib.doe.DesignOfExperiments.run_doe()` while preserving default behavior for typical users~~ fixes a trace-objective initialization inconsistency that can produce large residuals in Cholesky-related constraints at the start of the final NLP solve.

The motivating workflow showed:
- large early infeasibilities in optimize mode,
- ~~inability to use `max_iter=0` for final-model probing because scenario solves reused the same solver options,~~
- and inconsistent initialization among `fim`, `L`, `fim_inv`, `L_inv`, and `cov_trace`.

- ~~Added an advanced `run_config` (ConfigBlock/dict) argument to `run_doe(model=None, results_file=None, run_config=None)`:~~
  - ~~`scenario_solver_options`~~
  - ~~`final_solver_options`~~
  - ~~`final_solve`~~
  - ~~`inspection.enabled`~~
  - ~~`inspection.top_constraints`~~
- ~~Implemented scoped solver option application:~~
  - ~~scenario-generation and square-initialization solves can use different options from the final optimization solve.~~
- ~~Added optional structured residual reporting utility:~~
  - ~~reports top violated active constraints with fields:~~
    - ~~`constraint_name`~~
    - ~~`body`~~
    - ~~`lower_bound`~~
    - ~~`upper_bound`~~
    - ~~`violation`~~
    - ~~`constraint_type`~~
- ~~Added assemble-and-inspect path (`final_solve=False`) to inspect assembled NLP state without running final optimization.~~
- Fixed Cholesky jitter computation and added robust helper usage.
- Added `_initialize_cholesky_from_fim()` to re-synchronize `L`, `L_inv`, `fim_inv`, and `cov_trace` from current FIM immediately before final solve / inspection.
- ~~Updated docstrings and added advanced usage documentation + example path in `reactor_example.py`.~~
- Added/updated regression tests in `test_doe_initialization.py` for jitter behavior and FIM synchronization.

### Backward compatibility:
- ~~Default usage (`run_doe()` with no `run_config`) remains unchanged.~~
- ~~Advanced functionality is available only when `run_config` is provided.~~

### Testing:
I ran:

- ~~`pytest -q pyomo/contrib/doe/tests/test_doe_debug.py`~~
  - ~~Result: `8 passed`~~
- ~~`pytest -q pyomo/contrib/doe/tests/test_doe_solve.py -k "rooney_biegler_fd_central_solve"`~~
  - ~~Result: `1 passed, 17 deselected`~~
- `python3 -m unittest -q pyomo.contrib.doe.tests.test_doe_initialization`
  - Result: `3 skipped` in the current interpreter because SciPy/numpy are unavailable there
- `python3 -m py_compile pyomo/contrib/doe/doe.py pyomo/contrib/doe/tests/test_doe_initialization.py`
  - passed

New/updated tests cover:
- negative-jitter helper behavior,
- zero-jitter helper behavior,
- Cholesky/FIM inverse consistency after initialization in trace mode.

### Outstanding Tasks (before removing WIP):
- ~~Finalize proposed interface for advanced initialization diagnostics~~
- ~~Decide the best way to document the proposed interface. Are examples sufficient or should we build out the Pyomo documentation?~~
- ~~Run black~~
- ~~@adowling2 carefully reviews proposed changes~~

### Why the strikethought?

The original version of this PR included a new interface to `run_doe()` that makes debugging the initialization and, more broadly, solver failures easier. We decided that the new `run_doe()` interface added too much complexity and exposed some inconsistencies in other parts of the Pyomo solver interface. (This led to a real-time Codex experiment documented in https://github.com/Pyomo/pyomo/pull/3871.) To expedite this current PR, I decided to remove these contributions and strike through the above parts of the PR text. These contributions are still available in the git history. Thus, this current PR only contributes a targeted patch to the Pyomo FIM initialization.

## Legal Acknowledgement


adowling2:
@smondal13 @sscini @slilonfe5 @snarasi2 @djlaky I found a small bug in how we initialize the Cholesky factorization when the FIM at the initial point needs "inertia correction". This PR fixes the initialization inconsistency. This PR also proposes an interface to allow for more advanced solver options when diagnosing initialization issues for the Pyomo.DoE (dynamic) optimization formulation.

Edit: This is a work in progress right now. I am not 100% happy with the proposed changes.

adowling2:
@sscini @blnicho @mrmundt @smondal13 @slilonfe5 This is ready for review. This PR is now narrowly targeted at fixing a Pyomo.DoE FIM initialization issue.

adowling2:
Test failures appear to be unrelated to this PR: GAMS solvers (Windows) and NEOS solver (all)

slilonfe5:
> Test failures appear to be unrelated to this PR: GAMS solvers (Windows) and NEOS solver (all)

These failures are global. The Pyomo team mentioned that they are working on resolving the issues.

pyomo/contrib/doe/doe.py
```diff
@@ -537,6 +486,105 @@ def run_doe(self, model=None, results_file=None):
...
+            fim_inv_vals = np.linalg.inv(fim_pd)
+            for i, c in enumerate(model.parameter_names):
+                for j, d in enumerate(model.parameter_names):
+                    if self.only_compute_fim_lower and i < j:
+                        model.fim_inv[c, d].value = 0.0
+                    else:
+                        model.fim_inv[c, d].value = fim_inv_vals[i, j]
+
+        if hasattr(model, "cov_trace"):
+            fim_inv_np = np.linalg.inv(fim_pd)
```
smondal13: Since you are making the fim PD by adding the jitter, I believe the fim_pd will probably be non-singular. But if it is singular for some edge cases, then isn't it better to use the pseudo-inverse here just to be safe?

pyomo/contrib/doe/tests/test_doe_initialization.py
```diff
@@ -0,0 +1,197 @@
...
+        fd_formula="central",
+        step=1e-3,
+        objective_option="trace",
+        solver=_MutatingRecordingSolver(),
+        tee=False,
+    )
+
+
+@unittest.skipIf(
+    not (numpy_available and scipy_available),
```
smondal13: I think `scipy` is not required for the following tests. So, it may unnecessarily skip tests. Only `greybox` and `multi-experiment initialization` with LHS require `scipy` as far as I can remember.

pyomo/contrib/doe/doe.py
```diff
@@ -376,6 +375,11 @@ def run_doe(self, model=None, results_file=None):
             if self.objective_option == ObjectiveLib.trace:
                 trace_val = np.trace(np.linalg.pinv(self.get_FIM()))
                 model.obj_cons.egb_fim_block.outputs["A-opt"].set_value(trace_val)
+            elif self.objective_option == ObjectiveLib.pseudo_trace:
```
smondal13, line 379: I did not see any tests for the `pseudo_trace` objective in `test_greybox.py`

pyomo/contrib/doe/doe.py
adowling2: I like this suggestion. I am going to switch to `pinv` for most of the initialization code.

pyomo/contrib/doe/tests/test_doe_initialization.py
adowling2: Good catch

pyomo/contrib/doe/doe.py
adowling2: Good catch

pyomo/contrib/doe/tests/test_doe_initialization.py
```diff
@@ -0,0 +1,194 @@
...
+        self.options = {}
+
+    def solve(self, model, tee=False):
+        """Mutate 2x2 FIM values during init-stage solve."""
+        if (
+            hasattr(model, "dummy_obj")
+            and model.dummy_obj.active
+            and hasattr(model, "fim")
+            and len(list(model.parameter_names)) == 2
+        ):
```
blnicho: You know the attributes of the model being tested with this dummy solver so is this if-statement really necessary?

pyomo/contrib/doe/tests/test_doe_initialization.py
```diff
@@ -0,0 +1,194 @@
...
+            p1, p2 = list(model.parameter_names)
+            model.fim[p1, p1].set_value(16.0)
+            model.fim[p2, p1].set_value(4.0)
+            if (p1, p2) in model.fim:
+                model.fim[p1, p2].set_value(0.0)
+            model.fim[p2, p2].set_value(9.0)
+        return _FakeResult()
+
+
+class _TwoParamExperiment:
```
blnicho: Should this inherit from the `Experiment` class?

pyomo/contrib/doe/tests/test_doe_initialization.py
```diff
@@ -0,0 +1,194 @@
...
+        This is a direct unit test for the helper used by the initialization
+        sync path, so we can catch arithmetic regressions independently of the
+        solver/modeled example below.
+        """
+        doe_obj = _make_trace_doe_object()
+        min_eig = -1.0e-3
+        jitter = doe_obj._compute_cholesky_jitter(min_eig)
+        self.assertAlmostEqual(
+            jitter, _SMALL_TOLERANCE_DEFINITENESS - min_eig, places=14
+        )
```
blnicho: I know this is a simple method but in general we don't want to repeat the calculation implementation in the test and instead check against a hard-coded value. Checking up to 14 places also seems excessive. I would also recommend adding a test verifying that the return value is non-negative which is the intention of this test mentioned in the docstring.

pyomo/contrib/doe/tests/test_greybox.py
```diff
@@ -243,6 +265,18 @@ def get_numerical_second_derivative(grey_box_object=None, return_reduced=True):
...
+    Build common DesignOfExperiments keyword arguments used in this test module.
+
+    Parameters
+    ----------
+    experiment : object
+        Experiment fixture implementing ``get_labeled_model``.
+    fd_method : str
+        Finite-difference mode passed to DoE (for example, ``"central"``).
+    obj_used : str
+        Objective option name (for example, ``"trace"`` or ``"determinant"``).
```
blnicho:
```suggestion
        Finite-difference mode passed to DoE (for example, "central").
    obj_used : str
        Objective option name (for example, "trace" or "determinant").
```
I would recommend removing the extra formatting around the quotations

pyomo/contrib/doe/tests/test_greybox.py
```diff
@@ -389,6 +439,7 @@ class TestFIMExternalGreyBox(unittest.TestCase):
     # set the inputs for the
     # Grey Box object
     def test_set_inputs(self):
+        """Verify packed upper-triangular inputs reconstruct the full FIM."""
```
blnicho: This docstring doesn't really match the comment above the method describing the intended purpose. Maybe the comment should just be used as the docstring instead?

pyomo/contrib/doe/tests/test_greybox.py
```diff
@@ -405,6 +456,7 @@ def test_set_inputs(self):
     # Testing that getting the
     # input names works properly
     def test_input_names(self):
+        """Verify the grey-box input ordering matches the expected FIM entries."""
```
blnicho: This docstring doesn't really match the comment above the method describing the intended purpose. Maybe the comment should just be used as the docstring instead?

pyomo/contrib/doe/tests/test_doe_initialization.py
```diff
@@ -0,0 +1,194 @@
...
+
+    class _SolverData:
+        status = "ok"
+        termination_condition = "optimal"
+        message = "fake solve"
+
+    solver = _SolverData()
+
+
+class _MutatingRecordingSolver:
```
mrmundt: Are there any tests in any of the other files that exercise initialization using a real solver? (I do see changes in other tests in other files, but it's unclear to me if any of them are actually exercising the initialization code.)

If not, can you please add a test that uses a real solver / real result as well?

pyomo/contrib/doe/tests/test_greybox.py
```diff
@@ -869,7 +1029,28 @@ def test_D_opt_greybox_build(self):
...
+        # build path. This fixture may still carry the shared default flag, so
+        # we do not treat it as a grey-box invariant here.
+        self.assertTrue(doe_obj.use_grey_box)
+
+        doe_obj.create_grey_box_objective_function()
+
+        self.assertTrue(hasattr(doe_obj.model.obj_cons, "egb_fim_block"))
+        self.assertIn("log-D-opt", list(grey_box_object.output_names()))
+        self.assertIsInstance(doe_obj.model.objective, pyo.Objective)
+        self.assertEqual(doe_obj.model.objective.sense, pyo.maximize)
```
blnicho: I don't really understand the need for this test. It's checking for certain attributes and types but not checking any values or calculations. Aren't there separate unit tests verifying that objective function is being constructed with the right type and sense? Could these assertions be added to the previous test, `test_D_opt_greybox_build`?

The comment about "Grey-box objective construction not depending on the Cholesky build path" seems like it's trying to highlight something subtle in the implementation but I'm not following the subtlety based on the surrounding assertions.

pyomo/contrib/doe/doe.py
```diff
@@ -537,6 +486,109 @@ def run_doe(self, model=None, results_file=None):
...
+        min_eig: float
+            Minimum eigenvalue of the current FIM estimate.
+
+        Returns
+        -------
+        float
+            Nonnegative diagonal shift needed so the minimum eigenvalue
+            is at least ``_SMALL_TOLERANCE_DEFINITENESS``.
+        """
+        return max(0.0, _SMALL_TOLERANCE_DEFINITENESS - float(min_eig))
```
blnicho: Is there a reason to have this as a method on the `DesignOfExperiments` class instead of a standalone utility function? It isn't using anything from the class instance.

Also, it's a 1-line method that is only being used once on L555. Why did you even abstract this out as a separate method?

pyomo/contrib/doe/doe.py
```diff
@@ -537,6 +486,109 @@ def run_doe(self, model=None, results_file=None):
...
+        Returns
+        -------
+        None
+            Updates model values in place for available variables:
+            ``L``, ``L_inv``, ``fim_inv``, and ``cov_trace``.
+        """
+        if model is None:
+            model = self.model
+        if not hasattr(model, "L"):
+            return
```
blnicho, line 537: Could you add a comment explaining why this is a valid check to skip the rest of this method?

pyomo/contrib/doe/doe.py
```diff
@@ -537,6 +486,109 @@ def run_doe(self, model=None, results_file=None):
...
+                    model.L[c, d].value = 0.0
+
+        if hasattr(model, "L_inv"):
+            L_inv_vals = np.linalg.inv(L_vals)
+            for i, c in enumerate(model.parameter_names):
+                for j, d in enumerate(model.parameter_names):
+                    if i >= j:
+                        model.L_inv[c, d].value = L_inv_vals[i, j]
+                    else:
+                        model.L_inv[c, d].value = 0.0
```
blnicho: This implementation is slightly different from the implementation where these values are originally calculated on L1443-1465. Is there a reason for the difference? Can these be combined to avoid duplication?

pyomo/contrib/doe/grey_box_utilities.py
```diff
@@ -108,9 +108,10 @@ def __init__(self, doe_object, objective_option="determinant", logger_level=None
     def _get_FIM(self):
         # Grabs the current FIM subject
         # to the input values.
-        # This function currently assumes
-        # that we use a lower triangular
-        # FIM.
+        # Inputs store one triangular half
+        # of a symmetric FIM. Reconstruct
+        # the full symmetric matrix here,
+        # consistent with manuscript S5.
```
blnicho: What is "manuscript S5"?

blnicho (changes requested):

pyomo/contrib/doe/tests/test_greybox.py
adowling2: Good suggestion, change made

pyomo/contrib/doe/doe.py
adowling2: Done

pyomo/contrib/doe/grey_box_utilities.py
adowling2: Done.

pyomo/contrib/doe/tests/test_doe_initialization.py
adowling2: Based on your suggestion, I removed these tests and instead created a simple "pathological" example that has a singular FIM and thus exercises the new jitter logic. This is a much better solution (IMHO).

pyomo/contrib/doe/tests/test_doe_initialization.py
adowling2: Good catch. I agree.

pyomo/contrib/doe/tests/test_doe_initialization.py
adowling2: Yes. I changed the example to be a pathological non-identifiable model. It now inherets from Experiment.

pyomo/contrib/doe/tests/test_doe_initialization.py
adowling2: Good suggestion. I streamlined this and made the test more meaningful.

pyomo/contrib/doe/tests/test_greybox.py
adowling2: I cleaned up the doc string

pyomo/contrib/doe/tests/test_greybox.py
adowling2: This is a good suggestion. I merged parts of this into `test_D_opt_greybox_build` and removed this extra test.

pyomo/contrib/doe/doe.py
adowling2: Excellent suggestion. I combined these. You are correct they should not be different.

adowling2:
@blnicho @mrmundt Thank you for the great comments. I have implemented changes to address your comments and ran black. The tests are passing locally and are running on GitHub right now.

adowling2:
Tests now pass! (I had failures on tests that were being skipped in my local environment.)

pyomo/contrib/doe/tests/test_greybox.py
```diff
@@ -426,7 +426,7 @@ def make_greybox_and_doe_objects_rooney_biegler(objective_option):
         cyipopt_call_working = not (
             bad_message in doe_object.results["Termination Message"]
         )
-    except:
+    except Exception:
```
blnicho, line 429: Is there a more specific exception type that can be used here?

blnicho (approved):
@adowling2 I have one minor question but I'm also fine with merging as-is.

pyomo/contrib/doe/tests/test_greybox.py
mrmundt: I'll push back and say that it doesn't matter for this test suite. This is a catch-all for, "In case anything else goes wrong while trying to run cyipopt."

mrmundt (approved):

PR #3868, opened 2026-03-06
MPC Load Data Ignore Expressions

eslickj:
## Summary/Motivation:
When loading MPC data into a model, if a data key is an ``Expression``, the model will get messed up because the expression will be replaced.  This PR makes loading MPC data into a model ignore expressions.

- Ignore expressions when loading MPC data into a model
- When interpolating MPC data if a data point is None make the interpolated value None

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/mpc/data/interpolation.py
+++ b/pyomo/contrib/mpc/data/interpolation.py
@@ -71,7 +71,10 @@ def _get_interp_expr_vec(time_set, time_data, data, indexes=None):
     expr = [None] * len(time_set)
     for i, (h, t) in enumerate(zip(indexes, time_set)):
         l = h - 1
-        expr[i] = data[l] + (data[h] - data[l]) / (time_data[h] - time_data[l]) * (
-            t - time_data[l]
-        )
+        if data[l] is None or data[h] is None:
+            expr[i] = None
+        else:
+            expr[i] = data[l] + (data[h] - data[l]) / (time_data[h] - time_data[l]) * (
+                t - time_data[l]
+            )
     return expr
--- a/pyomo/contrib/mpc/interfaces/load_data.py
+++ b/pyomo/contrib/mpc/interfaces/load_data.py
@@ -81,9 +81,9 @@ def load_data_from_series(
         var = model.find_component(cuid)
         if var is None:
             _raise_invalid_cuid(cuid, model)
-        elif var.ctype == Expression and ignore_named_expressions:
-            continue
         elif var.ctype == Expression:
+            if ignore_named_expressions:
+                continue
             raise TypeError("Cannot load data for named Expression")
         for idx, val in zip(time_indices, vals):
             t = time_list[idx]
--- a/pyomo/contrib/mpc/interfaces/load_data.py
+++ b/pyomo/contrib/mpc/interfaces/load_data.py
@@ -36,9 +36,9 @@ def load_data_from_scalar(data, model, time, ignore_named_expressions=False):
         var = model.find_component(cuid)
         if var is None:
             _raise_invalid_cuid(cuid, model)
-        elif var.ctype == Expression and ignore_named_expressions:
-            continue
         elif var.ctype == Expression:
+            if ignore_named_expressions:
+                continue
             raise TypeError("Cannot load data for named Expression")
         # TODO: Time points should probably use find_nearest_index
         # This will have to happen in the calling function, as data
--- a/pyomo/contrib/mpc/interfaces/load_data.py
+++ b/pyomo/contrib/mpc/interfaces/load_data.py
@@ -184,9 +184,9 @@ def load_data_from_interval(
         var = model.find_component(cuid)
         if var is None:
             _raise_invalid_cuid(cuid, model)
-        elif var.ctype == Expression and ignore_named_expressions:
-            continue
         elif var.ctype == Expression:
+            if ignore_named_expressions:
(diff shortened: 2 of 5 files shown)


blnicho (approved):

pyomo/contrib/mpc/data/interpolation.py
mrmundt, line 1: Is it possible to add a test to check that this works as intended?

eslickj: Sure.  I'll do that.

eslickj: I added tests that have data for expressions, and I make sure it doesn't load and that the expressions evaluate to what they should.

pyomo/contrib/mpc/interfaces/load_data.py
```diff
@@ -32,6 +33,8 @@ def load_data_from_scalar(data, model, time):
     t_iter = time if _is_iterable(time) else (time,)
     for cuid, val in data.items():
         var = model.find_component(cuid)
+        if var.ctype == Expression:
+            continue
```
Robbybp: Should we silently continue or raise an error? You could imagine calling `load_data({"myexpr": 1.0})` then being confused when the value of the expression doesn't change.

Robbybp (commented):
I imagine this is coming up because, on one side, you're saving data with `include_named_expressions=True`, then later loading that data into the model gives an error? 

What do you mean by this?

> the model will get messed up because the expression will be replaced

Can you provide an example?

Handling `None` like this for interpolation makes sense to me.

pyomo/contrib/mpc/interfaces/load_data.py
eslickj: The reason this comes because I'm using the TimeSeriesData class as a basis for trajectory from Sundials, then loading that into a discretized model.  I'm also calculating named expressions because sometimes I need them, and don't want to run the discretized model. But when you load the data now it overwrites the expressions.  By mess up the model, I mean loading names expressions changes the model constraints potentially.

How do you feel about raising an exception by default and having an ``ignore_named_expressions`` option?

Robbybp: Oh I see, `expr[1] = 1.0` actually overrides the value of the named expression. Yep that's a problem. I think I'd prefer raising an error with an `ignore_named_expressions` option. Or I'd be fine with ignoring them as long as a warning is raised, in which case there should probably be the same option to suppress the warning...

eslickj: Ok, I think that sounds good.  I have that ready.  I just need to write a few tests.

eslickj: Okay, I updated it to raise an exception or take an ignore_named_expressions argument.

Robbybp (approved):
Looks good, thanks for this fix.

pyomo/contrib/mpc/interfaces/load_data.py
```diff
@@ -7,6 +7,7 @@
 # software.  This software is distributed under the 3-clause BSD License.
 # ____________________________________________________________________________________
 
+from pyomo.environ import Expression
```
jsiirola: I would prefer this be imported from `pyomo.core.base`.  Loading from `environ` triggers all the plugins, which could potentially cause strange circular imports.

pyomo/contrib/mpc/interfaces/load_data.py
```diff
@@ -34,6 +36,10 @@ def load_data_from_scalar(data, model, time):
         var = model.find_component(cuid)
         if var is None:
             _raise_invalid_cuid(cuid, model)
+        elif var.ctype == Expression and ignore_named_expressions:
+            continue
+        elif var.ctype == Expression:
+            raise TypeError("Cannot load data for named Expression")
```
jsiirola: Stylistic suggestion (avoids repeated attribute access and comparison):
```suggestion
        elif var.ctype == Expression:
            if ignore_named_expressions:
                continue
            raise TypeError("Cannot load data for named Expression")
```
(I'm indifferent if you want to have this or put the raise in an explicit "else")

pyomo/contrib/mpc/interfaces/load_data.py
```diff
@@ -72,6 +81,10 @@ def load_data_from_series(data, model, time, tolerance=0.0):
...
```
jsiirola: Same...

```suggestion
        elif var.ctype == Expression:
            if ignore_named_expressions:
                continue
            raise TypeError("Cannot load data for named Expression")
```

pyomo/contrib/mpc/interfaces/load_data.py
```diff
@@ -169,6 +184,10 @@ def load_data_from_interval(
...
```
jsiirola: and again

```suggestion
        elif var.ctype == Expression:
            if ignore_named_expressions:
                continue
            raise TypeError("Cannot load data for named Expression")
```

jsiirola (changes requested):
Some super-minor changes; otherwise LGTM.

pyomo/contrib/mpc/interfaces/load_data.py
eslickj: I'll have to remember that.  Thanks.

jsiirola (approved):

PR #3869, opened 2026-03-06
New options and mappings in the core.add_slack_variables Transformation

emma58:
## Summary/Motivation:

We've always used the `core.add_slack_variables` transformation as purely a debugging tool for infeasible models. However, there are other reasons one could want to add slack variables, such as relaxing certain constraints of a model by bounded amounts. This PR adds a couple features to the add slack variables transformation to make that easier, specifically an API so that a user can *get* the slack variables from the constraints and adding an option to not change the objective.

- Adds a config option to not touch the objective
- Adds public methods for getting relaxed constraints corresponding to given slack variables and vice versa on the transformation

### Legal Acknowledgement


pyomo/core/plugins/transform/add_slack_vars.py
```diff
@@ -171,6 +201,70 @@ def _apply_to_impl(self, instance, **kwds):
...
+        Parameters
+        ----------
+        model: ConcreteModel
+            A model, having had the 'core.add_slack_variables' transformation
+            applied to it
+        constraint: Constraint
+            A constraint that was relaxed by the transformation (either
+            because no targets were specified or because it was a target)
+        """
+        slack_variables = model.private_data().slack_variables
```
jsiirola: Do you get a meaningful error if `get_slack_variables` is called on a model that was not transformed by this Transformation?

pyomo/core/plugins/transform/add_slack_vars.py
```diff
@@ -171,6 +201,70 @@ def _apply_to_impl(self, instance, **kwds):
...
-        # make a new objective that minimizes sum of slack variables
-        xblock._slack_objective = Objective(expr=obj_expr)
+
+        if config.add_slack_objective:
+            # deactivate the objective
+            for o in instance.component_data_objects(Objective):
+                o.deactivate()
+
+            # make a new objective that minimizes sum of slack variables
+            xblock._slack_objective = Objective(expr=obj_expr)
```
jsiirola, line 227: Is there a use-case for saving the `obj_expr` in the `privae_data()` and adding an API for users to get it?  For example, that would make it easier to add it as a penalty to the existing Objective...  Without it, a user has to search the new block for a bunch of scalar variables...

pyomo/core/plugins/transform/add_slack_vars.py
```diff
@@ -161,6 +189,8 @@ def _apply_to_impl(self, instance, **kwds):
                 body += posSlack
                 # penalize slack in objective
                 obj_expr += posSlack
+                trans_info.slack_variables[cons].append(posSlack)
+                trans_info.relaxed_constraint[posSlack] = cons
```
jsiirola, line 204: Philosophical question: now that we are providing a useful API for mapping slacks to constraints and back, does it make sense to avoid all the name munging and just make a single indexed `xblock.slacks = Var(NonNegativeIntegers, bounds=(0, None))`?

jsiirola (commented):
I don't see any problems with this PR, but I do have a couple philosophical questions...

pyomo/core/plugins/transform/add_slack_vars.py
emma58: Wow, significantly more meaningful than I was expecting, actually:
```
>>> m = ConcreteModel()
>>> m.x = Var()
>>> m.y = Var()
>>> m.c = Constraint(expr=(3, m.x + m.y, 4))
>>> m.obj = Objective(expr= m.x - m.y)
>>> TransformationFactory('core.add_slack_variables').get_slack_variables(m, m.c)
  File "<python-input-6>", line 1, in <module>
    TransformationFactory('core.add_slack_variables').get_slack_variables(m, m.c)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/Users/esjohn/src/pyomo/pyomo/core/plugins/transform/add_slack_vars.py", line 240, in get_slack_variables
    raise ValueError(
    ...<3 lines>...
    )
ValueError: It does not appear that c is a constraint on model unknown that was relaxed by the 'core.add_slack_variables' transformation.
```
We were having a very smart day when we implemented the private data Blocks... This is working because, in the scope of that file, `slack_variables` gets created when we get the private_data, so then it's just empty and we get a good error.

pyomo/core/plugins/transform/add_slack_vars.py
emma58: I'm open to this, but a user wouldn't have to go searching for the scalars: they can get them via the Constraints. But maybe since this is probably a relatively common use case, we should make it easier.

pyomo/core/plugins/transform/add_slack_vars.py
emma58: Maaaaybe. Though the way it is now, `pprint` is enough for debugging... I am certainly guilty of this method.

pyomo/core/plugins/transform/add_slack_vars.py
emma58: OK, I added it.

emma58:
I decided to make this error for unknown active components. This means it will now die on an untransformed GDP, but I think that's okay because if someone wants to add slacks on a Disjunct, they can just call it on the Disjunct directly. (It just seems like that would be a rare enough use case compared to the more likely forgot-to-transform-first situation that it's okay to make it harder to accomplish.)

jsiirola (approved):

pyomo/core/plugins/transform/add_slack_vars.py
```diff
@@ -171,6 +212,141 @@ def _apply_to_impl(self, instance, **kwds):
...
+            for block in components[Constraint]:
+                for cons in block.component_data_objects(
+                    Constraint,
+                    active=True,
+                    descend_into=False,
+                    sort=SortComponents.deterministic,
+                ):
+                    yield cons
+
+    def get_slack_variables(self, model, constraint):
```
blnicho, line 276: Do you need to pass the model as an argument? Why not get the model from the constraint object directly, `constraint.model()`?

blnicho (approved):
@emma58 I have one minor question but otherwise this looks great!

pyomo/core/plugins/transform/add_slack_vars.py
jsiirola: `model` is probably a misnomer.  It is really the "root block" that was passed to the transformation, which may or may not be the the `model()` (which is where the transformation hides the mapping between constraints and slack variables.

@emma58, question: would it make sense to put a `private_data` on every block holding a transformation target that has a pointer to the "root block" from the transformation that holds the mappings?  If we do that, then we wouldn't need the `model` attribute (as @blnicho was suggesting).

emma58: I fixed the misnomer, but we discussed this during the dev call and realized that for the reverse API (getting constraints from variables), we *do* need to know what Block/model we are talking about since Vars are not active components. So for consistency we will leave this, but make the name and documentation better.

emma58:
When the tests pass, let's merge this!

PR #3871, opened 2026-03-10
[WIP][Do Not Merge] Refactor `cyipopt` solve-time options to match classic Pyomo solver behavior

adowling2:
## Summary/Motivation:

This PR makes the PyNumero `cyipopt` solver interface behave more like the classic Pyomo solver interfaces for solver options.

Before this change, `SolverFactory("cyipopt")` only consumed options stored on the solver object via `solver.config.options`, which meant callers needing one-shot overrides had to mutate solver state around a solve. That was inconsistent with the rest of Pyomo, where callers can generally pass ephemeral options directly to `solve(...)`.

With this change:

- `SolverFactory("cyipopt").solve(model, solver_options={...})` is supported
- `SolverFactory("cyipopt").solve(model, options={...})` is also supported for legacy-style compatibility
- per-call options override any defaults stored on the solver object for that solve only
- solver defaults stored on `solver.config.options` are not permanently mutated by solve-time overrides
- passing both `options=` and `solver_options=` raises an error, matching existing Pyomo conventions

1. `pyomo.contrib.pynumero.algorithms.solvers.cyipopt_solver.PyomoCyIpoptSolver`
- Added solve-time option merging in the public solver plugin used by `SolverFactory("cyipopt")`
- Preserved existing solver defaults from `config.options`
- Applied per-call overrides only to the active solve

2. `pyomo.contrib.doe.doe.DesignOfExperiments`
- Updated the grey-box `cyipopt` solve path to pass options through the solve call instead of relying solely on mutable solver state during the solve

3. `pyomo.contrib.mindtpy`
- Updated the legacy `cyipopt` paths so ephemeral settings like `max_cpu_time` and `constr_viol_tol` are added to solve-time kwargs instead of mutating `opt.config.options`
- Kept existing behavior for non-`cyipopt` solvers unchanged

4. Tests
- Added focused regression tests for the public `cyipopt` interface covering:
  - default solver options still apply
  - per-call `solver_options` override defaults for a single solve
  - legacy `options=` support
  - defaults are restored after solve
  - mixed `options=` and `solver_options=` is rejected
- Added MindtPy unit tests for the new cyipopt solve-time option plumbing

### Public API Behavior

Before:

```python
solver = pyo.SolverFactory("cyipopt")
solver.config.options["max_iter"] = 10
solver.solve(model)
```

One-shot overrides generally required mutating solver.config.options before solve and then restoring them afterward.

After:

```python
solver = pyo.SolverFactory("cyipopt")
solver.config.options["max_iter"] = 10

solver.solve(model, solver_options={"max_iter": 1})
# or
solver.solve(model, options={"max_iter": 1})
```

In both cases, the solve uses max_iter=1 for that call only, and the stored default remains 10

### Compatibility / Risk Notes

- This is intentionally narrow and limited to solve-time option handling.
- Existing code that sets persistent defaults on solver.config.options continues to work.
options= remains supported for compatibility, but solver_options= is also accepted and is the preferred solve-time override path where applicable.
- Passing both options= and solver_options= now raises a ValueError for cyipopt, matching behavior already used elsewhere in Pyomo.

### Outstanding Tasks Before Review

- Rerun the targeted tests in a Python 3.11 environment. I could not run the pytest targets in the current shell because the available interpreter is Python 3.9, while this checkout now imports Python 3.10+ syntax.

Suggested test commands once the environment is updated:

```
python -m pytest pyomo/contrib/pynumero/algorithms/solvers/tests/test_pyomo_cyipopt_solver_interface.py
python -m pytest pyomo/contrib/mindtpy/tests/test_cyipopt_options.py
python -m pytest pyomo/contrib/pynumero/algorithms/solvers/tests/test_cyipopt_solver.py
python -m pytest pyomo/contrib/doe/tests/test_greybox.py
python -m pytest pyomo/contrib/mindtpy/tests/test_mindtpy_grey_box.py
```

I am running a few minutes late; my previous meeting is running over.

## Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/doe/doe.py
+++ b/pyomo/contrib/doe/doe.py
@@ -452,7 +452,21 @@ class DesignOfExperiments:
 
         # Solve the full model, which has now been initialized with the square solve
         if self.use_grey_box:
-            res = self.grey_box_solver.solve(model, tee=self.grey_box_tee)
+            grey_box_solver_options = None
+            if hasattr(self.grey_box_solver, 'config') and hasattr(
+                self.grey_box_solver.config, 'options'
+            ):
+                grey_box_solver_options = dict(
+                    self.grey_box_solver.config.options.items()
+                )
+            if grey_box_solver_options:
+                res = self.grey_box_solver.solve(
+                    model,
+                    tee=self.grey_box_tee,
+                    solver_options=grey_box_solver_options,
+                )
+            else:
+                res = self.grey_box_solver.solve(model, tee=self.grey_box_tee)
         else:
             res = self.solver.solve(model, tee=self.tee)
 
--- a/pyomo/contrib/mindtpy/algorithm_base_class.py
+++ b/pyomo/contrib/mindtpy/algorithm_base_class.py
@@ -275,13 +275,18 @@ class _MindtPyAlgorithm:
                     'Your model is a NLP (nonlinear program). '
                     'Using NLP solver %s to solve.' % config.nlp_solver
                 )
+                nlp_args = dict(config.nlp_solver_args)
                 update_solver_timelimit(
-                    self.nlp_opt, config.nlp_solver, self.timing, config
+                    self.nlp_opt,
+                    config.nlp_solver,
+                    self.timing,
+                    config,
+                    solve_args=nlp_args,
                 )
                 self.nlp_opt.solve(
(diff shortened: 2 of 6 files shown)


adowling2:
This PR was created with Codex as part of an educational experiment.

PR #3872, opened 2026-03-14
Add SetOf domain; improve config's mock yaml dumper

jsiirola:
## Summary/Motivation:

@dallan-keylogic requested that Pyomo's config system had a `SubsetOf` domain that paralleled the `ListOf` domain.  This PR refactors `ListOf` so that `SetOf` is a trivial extension.  `SubsetOf` can be easily acomplished using a combination of `SetOf` and `In` (and is documented in the SetOf docs).

While I was messing with config, I refactored the mock yaml dumper so that we could reliably test it (even on systems with yaml).  This included a subtle change to the dumper so that types are more intelligently dumped (that is, we will consistently dump the actual type and not "<class 'type'>"), and so that `None` is output in proper YAML (as `null`).

- Add `SetOf`
- Refactor the mock yaml dumper to improve type output and support better testing

### Legal Acknowledgement

### DIFF
--- a/doc/OnlineDocs/explanation/experimental/solvers.rst
+++ b/doc/OnlineDocs/explanation/experimental/solvers.rst
@@ -264,9 +264,9 @@ solver's ``writer_config`` configuration option (see the
    file_determinism: FileDeterminism.ORDERED
    symbolic_solver_labels: false
    scale_model: true
-   export_nonlinear_variables: None
-   row_order: None
-   column_order: None
+   export_nonlinear_variables: null
+   row_order: null
+   column_order: null
    export_defined_variables: true
    linear_presolve: true
 
--- a/pyomo/common/config.py
+++ b/pyomo/common/config.py
@@ -753,6 +753,8 @@ def _get_dump():
         # dump = lambda x,**y: str(x)
         # YAML uses lowercase True/False
         def dump(x, **args):
+            if x is None:
+                return "null"
             if type(x) is bool:
                 return str(x).lower()
             if type(x) is type:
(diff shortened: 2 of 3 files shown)


dallan-keylogic:
My original problem was the result of me failing to use a domain validator for `ListOf`. Looking again at the problem today, it looks like I can solve my problem by correctly using domain validation in `ListOf`.

I'm not sure adding `SetOf` as a domain validator is the right solution, however. I am wary of using Python's default set objects because they are unordered.

When working on SOC dynamics, I ended up with an extremely obscure bug. I was specifying the phases in a certain property package as the set `{"Liq", "Vap"}`. The way IDAES phase equilibrium works is that we specify the phases in equilibrium with a tuple. However, because the phase set was unordered, it was sometimes defined as `("Liq", "Vap")` and sometimes as `("Vap", "Liq")`. Theoretically the order you get variables by iterating over a set should be the same within each Python version. However, we found that it would change even within the same version of Python. 

The way we specified the initial condition was to solve the steady state flowsheet, dump it to a `.json` file, then load that `.json` file to serve as the initial condition. Some IDAES variables, such as bubble and dew points, were indexed by the phase equilibrium pair. But if the `.json` file had variables indexed by `("Liq", "Vap")` and the simulation had variables indexed by `("Vap", "Liq")`, then those variables wouldn't have the correct values. If we tried to solve the initial condition problem, IPOPT could take a long time to solve and would sometimes fail. If we *didn't* try to solve the initial condition problem, the results might be nonsensical. 

So, in short, I think the user should be discouraged from providing items using unordered objects because it sets the stage for code either failing intermittently, or succeeding in one version of Python and failing in the other.

mrmundt (approved):

blnicho (approved):

PR #3874, opened 2026-03-17
Exact hull reformulation for quadratic constraints

sergey-gusev94:
## Summary/Motivation:
This PR adds support for the exact hull reformulation for quadratic constraints in Generalized Disjunctive Programming (GDP).

The current `gdp.hull` transformation applies perspective-function based relaxations for nonlinear constraints. While effective for many nonlinear expressions, this approach can produce weaker relaxations for quadratically constrained disjunctions.

This PR implements the exact hull reformulation for quadratic constraints described in:

Gusev, S., & Bernal Neira, D. E. (2025).
Exact Hull Reformulation for Quadratically Constrained Generalized Disjunctive Programs.
https://arxiv.org/abs/2508.16093

The implementation extends the existing hull transformation with an optional configuration flag (which is `False` by default) that detects quadratic constraints and applies the exact hull formulation when appropriate.

For convex quadratic constraints, the reformulation is expressed as a rotated second-order cone (SOC) representable constraint. For nonconvex quadratics and equality constraints, the general exact hull formulation is used.

Convexity is determined automatically through eigenvalue analysis of the Hessian matrix.

This provides a tighter relaxation for quadratic GDP models and maintains the quadratic structure of constraints in hull reformulation.

- Extend the `gdp.hull` transformation with a new configuration option:

  - `exact_hull_quadratic` (default: `False`)

- When enabled, quadratic constraints inside disjuncts are reformulated using the exact hull formulation instead of the standard perspective function.

- Implement automatic convexity detection via eigenvalue decomposition of the quadratic Hessian matrix.

- Apply different reformulations depending on structure:

  - Convex quadratic inequalities  
    Reformulated using a rotated second-order cone representation.

  - Nonconvex quadratics or equalities  
    Reformulated using the general exact hull formulation.

- Add a configuration parameter:

  - `eigenvalue_tolerance`  
    Controls numerical tolerance when checking positive or negative semidefiniteness of the Hessian.

- Extend the implementation in `pyomo.gdp.plugins.hull` to support these reformulations.

- Add unit tests verifying:
  - correctness of the generated constraints
  - compatibility with existing hull behavior
  - correctness of variable disaggregation and mappings.

The default behavior of `gdp.hull` is unchanged unless the new option is enabled.

Example usage:

```python
TransformationFactory("gdp.hull").apply_to(
    model,
    exact_hull_quadratic=True
)
```

### Legal Acknowledgement


pyomo/gdp/plugins/hull.py
```diff
@@ -820,6 +949,403 @@ def _transform_constraint(
...
+                "GDP(Hull): Constraint '%s' contains mutable parameters %s. "
+                "The exact hull reformulation evaluates these to their current "
+                "numeric values for the eigenvalue-based convexity check; the "
+                "transformed constraint will not update if these parameters "
+                "change later.",
+                c.getname(fully_qualified=True),
+                [p.name for p in mutable_params],
+            )
+
+        repn = generate_standard_repn(c.body)
```
emma58: Sorry, this isn't documented yet, but we are gradually replacing `generate_standard_repn` with more robust expression visitors. For your purposes, you should use the `QuadraticRepnVisitor` from `pyomo.repn.quadratic`.

You will also want to account for the `assume_fixed_vars_permanent` option when you do this. Right now, the easiest way to do that when you are not assuming they are permanent (the default) is to actually *unfix* them before you walk the expression and then fix them back to their old values after the fact. The same Hessian issue would happen if they are treated as params (the `True` option) but in this case it's actually okay because setting that option to True is promising to not unfix them or change their values in any  way.

pyomo/gdp/plugins/hull.py
```diff
@@ -658,6 +746,25 @@ def _get_local_var_list(self, parent_disjunct):
     def _transform_constraint(
         self, obj, disjunct, var_substitute_map, zero_substitute_map
     ):
+        """Transform a single Constraint on a Disjunct.
+
+        Applies the appropriate hull reformulation to each
+        ``ConstraintData`` in ``obj``. When ``exact_hull_quadratic`` is
+        enabled and the constraint body has polynomial degree 2, an exact
+        hull formulation is used instead of the perspective function.
+
```
emma58, line 780: Thanks for the docstring!!

pyomo/gdp/plugins/hull.py
```diff
@@ -676,20 +783,42 @@ def _transform_constraint(
             unique = len(newConstraint)
             name = c.local_name + "_%s" % unique
 
-            NL = c.body.polynomial_degree() not in (0, 1)
+            polynomial_degree = c.body.polynomial_degree()
```
emma58: There are some known issues with `polynomial_degree`... I think that, since we're going to need it anyway, it probably makes sense to use the `QuadraticRepnVisitor` here to determine the nature of the Constraint body. That will be more robust to some strange edge cases.

pyomo/gdp/plugins/hull.py
```diff
@@ -820,6 +949,403 @@ def _transform_constraint(
...
+        if mutable_params:
+            logger.warning(
+                "GDP(Hull): Constraint '%s' contains mutable parameters %s. "
+                "The exact hull reformulation evaluates these to their current "
+                "numeric values for the eigenvalue-based convexity check; the "
+                "transformed constraint will not update if these parameters "
+                "change later.",
+                c.getname(fully_qualified=True),
+                [p.name for p in mutable_params],
+            )
```
emma58: I would lean towards putting this fact in the documentation rather than logging a warning, since there's nothing a user can do to prevent this warning other than actually going through and replacing mutable params with immutable ones. But we should definitely make this loud in the docs: It's a departure from the usual behavior of GDP transformations.

pyomo/gdp/plugins/hull.py
```diff
@@ -820,6 +949,403 @@ def _transform_constraint(
...
+        for var_i, var_j in repn.quadratic_vars:
+            if id(var_i) not in var_ids_seen:
+                all_vars.append(var_i)
+                var_ids_seen.add(id(var_i))
+            if id(var_j) not in var_ids_seen:
+                all_vars.append(var_j)
+                var_ids_seen.add(id(var_j))
+
+        n_vars = len(all_vars)
+        var_to_idx = {id(var): idx for idx, var in enumerate(all_vars)}
```
emma58: It looks like you're building several data structures you don't need here. You never use `all_vars` for anything, and the keys of `var_to_index` are the `var_ids_seen` set. So I propose replacing these lines with:
```
var_to_idx = {}
for quad_vars in repn.quadratic_vars:
    for var_i in quad_vars:
        if id(var_i) not in var_to_idx:
            var_to_idx[id(var_i)] = len(var_to_idx)
n_vars = len(var_to_idx)
```

emma58 (changes requested):
This looks pretty good, but I'm submitting a partial review because there's one major revision: See my notes below, but can you please rewrite this to use the `QuadraticRepnVisitor` rather than `generate_standard_repn`? You'll be able to get all the same data, though it will come in a slightly different structure. But it is much more robust to edge cases--I think `generate_standard_repn` is not long for this world, actually.

pyomo/gdp/plugins/hull.py
sergey-gusev94: Addressed in 15d7fb7.I switched from generate_standard_repn to QuadraticRepnVisitor

pyomo/gdp/plugins/hull.py
sergey-gusev94: Addressed in 15d7fb7.I switched from generate_standard_repn to QuadraticRepnVisitor

pyomo/gdp/plugins/hull.py
sergey-gusev94: That's right! Thank you! Addressed in 15d7fb7

pyomo/gdp/plugins/hull.py
sergey-gusev94: I removed the warning and documented this behavior instead in 15d7fb7

bernalde (approved):

mrmundt:
@jsiirola , @blnicho  - This one is ready for a re-review!

pyomo/gdp/plugins/hull.py
```diff
@@ -180,6 +218,74 @@ class Hull_Reformulation(GDP_to_MIP_Transformation):
...
+        all nonlinear constraints.
+
+        **Departure from typical GDP hull behavior (mutable Parameters):**
+        Coefficients are obtained via ``QuadraticRepnVisitor``, which freezes
+        mutable :class:`Param` values (and other fixed numeric leaves) when the
+        expression is walked. The transformed model is not updated if those
+        Params change later. Most other hull paths keep Param references in the
+        transformed constraints; this option does not. Set mutable Params to
+        their intended values before transforming, or avoid mutable Params in
+        the quadratic body, if you need algebraic linkage to parameters.
```
emma58, line 263: It might also be worth mentioning that using fixed Vars as mutable parameters and setting `assume_fixed_vars_permanent` to `False` is another viable workaround.

pyomo/gdp/plugins/hull.py
```diff
@@ -676,9 +801,70 @@ def _transform_constraint(
...
+                visitor = QuadraticRepnVisitor(
+                    {},
+                    var_recorder=OrderedVarRecorder(
+                        var_map, var_order, SortComponents.deterministic
+                    ),
+                )
+                qrepn = visitor.walk_expression(c.body)
+                vm = visitor.var_map
+
+                is_quadratic = qrepn.quadratic is not None and len(qrepn.quadratic) > 0
```
emma58: Is there a case where `qrepn.quadratic` is `None` and `len(qrepn.quadratic) == 0`?

pyomo/gdp/plugins/hull.py
```diff
@@ -676,9 +801,70 @@ def _transform_constraint(
...
+                    vm=vm,
+                )
+                for var in to_refix:
+                    var.fix()
+                continue
+
+            # Not using exact quad: classify via polynomial_degree and refix.
+            polynomial_degree = c.body.polynomial_degree()
+            for var in to_refix:
+                var.fix()
```
emma58: We can do this in a follow-on PR if you want, but if it's easy enough, we should always use the QuadraticRepnVisitor instead of `polynomial_degree`, regardless of if we're using the exact quadratic hull or not.

pyomo/gdp/plugins/hull.py
```diff
@@ -820,6 +1007,397 @@ def _transform_constraint(
...
+        constraint_map : object
+            Private data object tracking constraint mappings.
+        var_substitute_map : dict
+            Mapping from ``id(original_var)`` to disaggregated variable.
+        newConstraint : Constraint
+            The indexed Constraint container for transformed constraints.
+        name : str
+            Base name for the transformed constraint indices.
+        i : object
+            The index of the constraint in its parent component.
```
emma58: Minor nit, but I might name this `idx` or something a bit more verbose.

pyomo/gdp/plugins/hull.py
```diff
@@ -820,6 +1007,397 @@ def _transform_constraint(
...
+        obj : Constraint
+            The parent Constraint component (needed for ``is_indexed``).
+        qrepn : QuadraticRepn
+            Result of ``QuadraticRepnVisitor.walk_expression`` on the constraint
+            body (walked by the caller with fixed Vars temporarily unfixed).
+        vm : dict
+            The visitor's ``var_map``: maps ``id(Var)`` to ``Var`` objects.
```
emma58: Also a minor nit, but this might be clearer if it was named `var_map`

pyomo/gdp/plugins/hull.py
```diff
@@ -820,6 +1007,397 @@ def _transform_constraint(
...
+                    newConstraint[name, 'eq']
+                )
+                constraint_map.src_constraint[newConstraint[name, 'eq']] = c
+            return
+
+        # --- Lower bound ---
+        if c.lower is not None:
+            if self._generate_debug_messages:
+                _name = c.getname(fully_qualified=True)
+                logger.debug("GDP(Hull): Transforming constraint '%s'", _name)
```
emma58, line 1202: I think the equality case is missing this debug message.

pyomo/gdp/tests/test_hull.py
```diff
@@ -2890,3 +2894,907 @@ class NestedDisjunctsInFlatGDP(unittest.TestCase):
...
+        self.hull.apply_to(m, exact_hull_quadratic=True)
+
+        relaxBlock = m._pyomo_gdp_hull_reformulation.relaxedDisjuncts[0]
+        v_x = relaxBlock.disaggregatedVars.x
+        v_y = relaxBlock.disaggregatedVars.y
+        y_ind = m.d1.binary_indicator_var
+
+        # An auxiliary t variable should have been created.
+        t_var = relaxBlock.component('_conic_aux_t_c')
+        self.assertIsNotNone(t_var, "Expected auxiliary variable '_conic_aux_t_c'")
```
emma58: I would recommend against testing that depends on private names since any naming changes will break tests. It might be worth including an API (using the private_data Block) to retrieve the auxiliary variables, and then you can test them that way.

pyomo/gdp/tests/test_hull.py
```diff
@@ -2890,3 +2894,907 @@ class NestedDisjunctsInFlatGDP(unittest.TestCase):
...
+        # Two transformed constraints: conic SOC + linear bound.
+        trans_cons = self.hull.get_transformed_constraints(m.d1.c)
+        self.assertEqual(len(trans_cons), 2)
+
+        conic_cons, linear_cons = trans_cons
+
+        # --- Conic constraint: v_x**2 + v_y**2 - t*y_ind <= 0 ---
+        self.assertIsNone(conic_cons.lower)
+        self.assertEqual(value(conic_cons.upper), 0)
+        repn = generate_standard_repn(conic_cons.body, compute_values=False)
```
emma58: It doesn't have to hold up this PR, but could you update the tests to use the `QuadraticRepnVisitor` as well, rather than `generate_standard_repn`?

emma58 (commented):
Thank you so much for making the change to the `QuadraticRepnVisitor`! This is looking quite good. Most of the changes suggested below are fairly minor, though if you can, it would be nice to abandon `generate_standard_repn` in the tests as well and to add public APIs where you are testing private components added by the transformation.

pyomo/gdp/plugins/hull.py
sergey-gusev94: Addressed in 1b1f81c

pyomo/gdp/plugins/hull.py
sergey-gusev94: I don't think so. I changed it in 1b1f81c

pyomo/gdp/plugins/hull.py
sergey-gusev94: Addressed in 1b1f81c

pyomo/gdp/plugins/hull.py
sergey-gusev94: Addressed in 1b1f81c

pyomo/gdp/plugins/hull.py
sergey-gusev94: Addressed in 1b1f81c

pyomo/gdp/plugins/hull.py
sergey-gusev94: Addressed in 1b1f81c

pyomo/gdp/tests/test_hull.py
sergey-gusev94: Updated for tests done for exact quadratic functionality in 1b1f81cc. For the rest of the hull reformulation tests, it should probably be done in a separate PR

pyomo/gdp/tests/test_hull.py
sergey-gusev94: Addressed in  73e309e

pyomo/gdp/tests/test_hull.py
emma58: This is reasonable--thank you!

pyomo/gdp/plugins/hull.py
emma58: Oh, but thinking about this again, is there a problem here because changing the values can effect convexity? I guess there's still a user-beware thing we should mention, right? I think it's fine--we can't really save people from themselves all the way. But it's probably worth mentioning in the docs.

emma58: Oh, sorry, I'm wrong because they get treated as vars in the transformation when we do this. Never mind!

pyomo/gdp/plugins/hull.py
```diff
@@ -963,6 +1544,45 @@ def get_var_bounds_constraint(self, v, disjunct=None):
...
+
+        Parameters
+        ----------
+        cons : ConstraintData
+            Original constraint on a Disjunct that may have been reformulated
+            using the conic exact hull.
+        raise_exception : bool
+            Whether to raise a ``GDP_Error`` when ``cons`` is not on a
+            transformed Disjunct. Default is ``True``.
+
```
emma58: I would just always raise the exception. That's the pattern we've followed in other APIs like this.

pyomo/gdp/plugins/hull.py
```diff
@@ -963,6 +1544,45 @@ def get_var_bounds_constraint(self, v, disjunct=None):
...
+        """
+        disjunct = cons.parent_block()
+        if disjunct is None or disjunct.ctype is not Disjunct:
+            if raise_exception:
+                raise GDP_Error(
+                    "Constraint '%s' does not appear to belong to a Disjunct"
+                    % cons.name
+                )
+            return None
+        if disjunct._transformation_block is None:
```
emma58: There's a public `transformation_block` property on Disjuncts--please use that.

pyomo/gdp/plugins/hull.py
```diff
@@ -963,6 +1544,45 @@ def get_var_bounds_constraint(self, v, disjunct=None):
...
+            if raise_exception:
+                raise GDP_Error(
+                    "Disjunct '%s' has not been transformed" % disjunct.name
+                )
+            return None
+        return (
+            disjunct._transformation_block()
```
emma58: Same as above.

emma58 (commented):
Thank you for the lightning-fast response! A couple more minor comments and then I really will be done complaining!

pyomo/gdp/plugins/hull.py
sergey-gusev94: Addressed in 7187aab

pyomo/gdp/plugins/hull.py
sergey-gusev94: Addressed in 7187aab

pyomo/gdp/plugins/hull.py
sergey-gusev94: Addressed in 7187aab

emma58 (approved):

mrmundt (approved):

PR #3876, opened 2026-03-20
Remove `pyomo.duality` and `pyomo.core.plugins.transform.model`

jsiirola:
## Summary/Motivation:
This removes two {old, deprecated, poorly tested} pieces of Pyomo:
- `pyomo.duality` - this package has been completely superseded by:
    - `pyomo.duality.lagrangian_dual` (deprecated since 5.6.2) is completely replaced by `pyomo.core.plugins.transform.lp_dual` 
    - `pyomo.duality.collect` is replaced by `pyomo.repn.plugins.standard_form`.  This is a more generalized and robust implementation (and with simpler, easier to use return data types)
       - **Note we never officially deprecated this module.**  That said, it is buggy and not documented.  To our knowledge, no one in the wild is using this module (otherwise they should have been complaining about the bugs).  As it is not core functionality and does not appear in the Book, this PR proposes deleting the module *without deprecation*.  An alternative path would be to create a small wrapper around the `standard_form` writer that re-creates the old return types (which will resolve the bugs) and then *immediately deprecate that method.*  Providing the deprecation path would be several hours of work.
- `pyomo.core.plugins.transform.model` (deprecated since 6.7.3) has been replaced by `pyomo.repn.plugins.standard_form`

- remove `pyomo.duality`
- remove `pyomo.core.plugins.transform.model`

### Legal Acknowledgement


mrmundt (approved):

pyomo/duality/tests/test_linear_dual.py
```diff
@@ -1,189 +0,0 @@
...
-                getattr(self, key).__doc__ = " (%s)" % getattr(self, key).__name__
-
-    def test_t5(self):
-        self.problem = 'test_t5'
-        self.run_bilevel(join(exdir, 't5.py'))
-        self.check('t5', 'linear_dual')
-
-    def test_t1(self):
-        self.problem = 'test_t1'
-        self.run_bilevel(join(exdir, 't1.py'))
```
blnicho, line 96: There are a couple example py files used in these tests that are not deleted in this PR. I'm guessing they aren't being tested anywhere else so I wonder if they should be removed.

jsiirola: Good catch!  Removed.

blnicho (approved):

PR #3733, opened 2025-09-23
Restructure and Add Tutorial to Online PyROS Documentation

shermanjasonaf:
## Summary/Motivation:

This PR restructures the PyROS online documentation and adds a new tutorial to the documentation.

## Changes proposed in this PR:
- Reorganize the PyROS online documentation and split it into multiple webpages
- Add a new, extensive tutorial to the PyROS online documentation
- Replace the original quickstart example problem with a problem that is smaller, easier to write down, and less sensitive to the subsolver(s) used (see [this comment of #3649](https://github.com/Pyomo/pyomo/pull/3649#discussion_r2220293365))
- Add the mathematical definition of each pre-implemented PyROS `UncertaintySet` class to the class's docstring

## TODO (after #3732 merged)
- [x] Account for changes to the logging system in the restructured online docs
- [x] Rename the new PyROS doc folder as `"pyros"`, and make the original PyROS reST file reference the main page of the new PyROS doc folder
- [x] Update version number, changelog

### Legal Acknowledgement

By contributing to this software project, I have read the [contribution guide](https://pyomo.readthedocs.io/en/stable/contribution_guide.html) and agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #3877, opened 2026-03-20
PyROS Add caching for computed uncertain parameter bounds

jas-yao:
## Fixes 
`_fbbt_parameter_bounds` in `uncertainty_sets.py`

## Summary/Motivation:
PyROS solves optimization bounding problems for every uncertain parameter multiple times throughout its routine using `_compute_exact_parameter_bounds`. There are up to 4 instances of PyROS accessing this method during its runtime.
1. Validation in `is_bounded`. This occurs when no parameter bounds are provided and FBBT fails to find bounds. Only the bounds that FBBT has not found are evaluated.
2. Validation in `is_nonempty`. This occurs for intersection, polyhedral, and custom uncertainty sets, where a feasibility problem is constructed.
3. Preprocessing in `get_effective_uncertain_dimensions` if parameter bounds are not provided or the provided ones are not exact.
4. Separation problem construction in `add_uncertainty_set_constraints` if no parameter bounds are provided.

The time taken to repeatedly solve these bounding problems may add up and be significant for larger uncertainty sets. 

This PR adds a method for caching the solutions of these bounding problems so that subsequent processes do not need to solve the bounding problems again.

This PR also fixes a small bug in `_fbbt_parameter_bounds` where the returned bounds are expressions rather than values (e.g., a variable with both domain and bounds set such as `m.Var = pyo.Var(within=pyo.UnitInterval, bounds=(0,1))` or a variable with `within=pyo.UnitInterval ` in a model that FBBT has been used on to set bounds has lower bound max(0,0) and upper bound min(1,1)).

- Add a `_solve_bounds_optimization` method that caches solutions for bounding problems solved for any uncertain parameter that is used by `_compute_exact_parameter_bounds`
- Add cache checks and clearing at the end of every PyROS solve.
- Add tests for checking caching behavior of `_solve_bounds_optimization`
- Fix the issue in `_fbbt_parameter_bounds`

### Legal Acknowledgement

6. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.


jas-yao:
@shermanjasonaf @jsiirola

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -788,45 +793,78 @@ def _compute_exact_parameter_bounds(self, solver, index=None):
...
+        of `self` at a specified index by solving a bounding model.
+        Results are cached as efficiency for large uncertainty sets.
+
+        Parameters
+        ----------
+        solver : Pyomo solver type
+            Optimizer to invoke on the bounding problems.
+        index : int
+            The index of the parameter to solve for bounds.
+        sense : Pyomo objective sense
```
jsiirola:
```suggestion
        sense : ~pyomo.common.ObjectiveSense
```

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -788,45 +793,78 @@ def _compute_exact_parameter_bounds(self, solver, index=None):
...
+    @functools.cache
+    def _solve_bounds_optimization(self, solver, index, sense):
+        """
+        Compute value of bounds for a single parameter
+        of `self` at a specified index by solving a bounding model.
+        Results are cached as efficiency for large uncertainty sets.
+
+        Parameters
+        ----------
+        solver : Pyomo solver type
```
jsiirola:
```suggestion
        solver : ~pyomo.opt.base.solvers.OptSolver
```

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -788,45 +793,78 @@ def _compute_exact_parameter_bounds(self, solver, index=None):
...
+            the corresponding dimension at the specified index.
+
+        Raises
+        ------
+        ValueError
+            If solver failed to compute a bound for a
+            coordinate.
+        """
+        # create bounding model and get all objectives
+        bounding_model = self._create_bounding_model()
```
jsiirola: You are re-creating the bounding model for every variable and both directions.  This seems wasteful.  Given that this is typically called in a loop, maybe the model should be created once outside the look and passed in to this function?  

That may foul up the use of @functools.cache, but we could just directly implement our own cache as an instance / model attribute.

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -667,6 +668,10 @@ def validate(self, config):
         ValueError
             If nonemptiness check or boundedness check fails.
         """
+        # clear any cached exact parameter bounds
+        self._solve_bounds_optimization.cache_clear()
```
jsiirola: Is this the right place to clear the cache?  This clears it at the very beginning of the PyROS algorithm, right?  It should be empty now.  Does it make more sense to clear the cache at the end of the algorithm so that we don't leave unnecessary models hanging around in memory?

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -788,45 +793,78 @@ def _compute_exact_parameter_bounds(self, solver, index=None):
...
+            raise ValueError(
+                "Could not compute "
+                f"{'lower' if sense == minimize else 'upper'} "
+                f"bound in dimension {index + 1} of {self.dim}. "
+                f"Solver status summary:\n {res.solver}."
+            )
+
+        # ensure sense is minimize when done, deactivate
+        obj.sense = minimize
+        obj.deactivate()
```
jsiirola: If the solve fails, the objective is left activated.  I recommend moving this earlier:
```python
try:
     res = solver.solve(bounding_model, load_solutions=False)
finally:
     # ensure sense is minimize when done, deactivate
     obj.sense = minimize
     obj.deactivate()
```

jsiirola (commented):
This looks pretty good.  Some questions and a couple minor suggestions.

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -858,7 +896,8 @@ def _fbbt_parameter_bounds(self, config):
             )
 
         param_bounds = [
-            (var.lower, var.upper) for var in bounding_model.param_vars.values()
+            (value(var.lower), value(var.upper))
```
shermanjasonaf: We should consider adding a test that is affected by this change.

In light of the write-up for this PR, it may also be worth noting that as of PR #3733, the PyROS Uncertainty Sets documentation page explicitly states that mixed-integer uncertainty sets are not supported. More broadly, and perhaps in a future PR, we may want to make that documentation and/or the docstring for the `UncertaintySet.set_as_constraint()` method more specific/explicit about can/can't be done to model components within that method. (E.g., should altering the domains of the uncertain parameter `Var` objects be generally supported?)

pyomo/contrib/pyros/uncertainty_sets.py
shermanjasonaf: I anticipate that clearing the cache here will work for most use cases, but I am questioning whether this is the safest place to clear the cache. If, for example, the user implements a custom `UncertaintySet` subclass where the `validate()` method is overriden such that the cache clearing statement is omitted, then the `_compute_exact_parameter_bounds()` method may return an inaccurate result when, for example, the state of a subclass instance is altered between successive calls to `PyROS.solve()` that use the subclass instance.

Perhaps cache clearing should be invoked in `PyROS.solve()`, just before argument validation? We probably also want the cache cleared near the end of `PyROS.solve()` to clear memory (see @jsiirola's [comment](https://github.com/Pyomo/pyomo/pull/3877/changes#r2989819984)). To this end, I would consider modifying `PyROS.solve()` to the following:

```python
def solve(...):
    ...  # everything up to disclaimer logging

    ...  # <-- clear cache here
    try:
        # everything from argument validation
        # to finalization of results object
        ...  
    finally:
        ...  # <-- clear cache here

    ... # log termination-related messages

    return ...
```

If we proceed with this change, then we may want to add to `test_grcs.py` a test to check that successive calls to `PyROS.solve()` using the same `UncertaintySet` object yield accurate results, particularly for a case where:
- The `validate()` method is overriden
- The exact bounds are computed exclusively using `_compute_exact_parameter_bounds`.
- Between the successive calls to `PyROS.solve()`, the attributes of the `UncertaintySet` object are changed such that the exact bounds are changed.

shermanjasonaf (commented):
Good PR. I have a few questions about edge cases and testing.

jsiirola:
@jas-yao: we are adopting a new [review process](https://pyomo.readthedocs.io/en/latest/contribution_guide.html#review-process) where we convert PRs that are "waiting on the author" back to "draft" (to signal the PR state to both the author and the dev team).  Once you have had a chance to address the comments, please mark it as "ready for review" so the developers can get it back into the review queue.

pyomo/contrib/pyros/uncertainty_sets.py
jas-yao: I have added a test for this. 

Yes, I noticed that the FBBT issue comes up when uncertain parameter `Var` objects are given domains (e.g., Binary, NonNegativeReals, UnitInterval, etc.) and bounds at the same time. Setting bounds and domains leads to an expression object being returned by `var.lower` and `var.upper` rather than a value.

I think it is reasonable to not support altering the domains of uncertain parameters if they can already be specified through setting parameter bounds.

pyomo/contrib/pyros/uncertainty_sets.py
jas-yao: I have added some test to `test_grcs.py` for this with a modified `BoxSet` that uses `_compute_exact_parameter_bounds` for all uncertain parameter bounds calculations including this test. The `BoxSet` `validate()` overrides the the base `UncertaintySet` method.

jas-yao:
Thank you for all the detailed feedback @jsiirola and @shermanjasonaf .

I have updated the caching implementation and added additional tests. This PR is ready for review.

I had some questions come to mind when working on this and would appreciate any insights:
1. Where is the best place for the `._cache` attribute to be created? It is currently being created when needed in `_solve_exact_bounds_optimization`.
2. Does it make sense to have the default behavior of `parameter_bounds` in `UncertaintySet` return `_compute_exact_parameter_bounds`?

pyomo/contrib/pyros/tests/test_grcs.py
```diff
@@ -5146,5 +5146,276 @@ def test_discrete_set_subsolver_error_recovery(self, name, sec_con_UB):
...
+
+    def test_pyros_cache_time(self):
+        """
+        Check that caching improves solve time.
+        """
+        m = build_leyffer_two_cons()
+
+        # Define the uncertainty set
+        interval_cache = CustomExactBoundsUncertaintySet(
+            bounds=[(0.25, 2)], sleep_time=0.5, cache=True
```
jsiirola: Can this be made smaller (maybe 0.1)?  I worry about how long this test takes (it's about 5 seconds).

pyomo/contrib/pyros/tests/test_grcs.py
```diff
@@ -5146,5 +5146,276 @@ def test_discrete_set_subsolver_error_recovery(self, name, sec_con_UB):
...
+            bounds=[(0.25, 2)], sleep_time=0.5, cache=False
+        )
+
+        # Instantiate the PyROS solver
+        pyros_solver = SolverFactory("pyros")
+
+        # Define subsolvers utilized in the algorithm
+        # the error solver will cause the first separation problem to fail
+        local_subsolver = SolverFactory("baron")
+        global_subsolver = SolverFactory("baron")
```
jsiirola: Do we really need to use baron here?

pyomo/contrib/pyros/tests/test_grcs.py
```diff
@@ -5146,5 +5146,276 @@ def test_discrete_set_subsolver_error_recovery(self, name, sec_con_UB):
...
+        interval.bounds = [(0.25, 2)]
+        self.assertEqual(interval.parameter_bounds, [(25, 200)])
+
+        # Instantiate the PyROS solver
+        pyros_solver = SolverFactory("pyros")
+
+        # Define subsolvers utilized in the algorithm
+        # the error solver will cause the first separation problem to fail
+        local_subsolver = SolverFactory("baron")
+        global_subsolver = SolverFactory("baron")
```
jsiirola: Do we really need Baron here?

jsiirola (commented):
Some more minor questions,a nd a potential significant design simplification:  Instead of putting everything inside another context block, you could use a custom `dict` that supports context management:
```diff
diff --git a/pyomo/contrib/pyros/uncertainty_sets.py b/pyomo/contrib/pyros/uncertainty_sets.py
index a925ab1939..8a056e05b2 100644
--- a/pyomo/contrib/pyros/uncertainty_sets.py
+++ b/pyomo/contrib/pyros/uncertainty_sets.py
@@ -460,6 +460,15 @@ def validate_array(
         )


+class ContextDict(dict):
+    def __enter__(self):
+        assert not self
+        return self
+
+    def __exit__(self, et, e, tb):
+        self.clear()
+
+
 class Geometry(Enum):
     """
     Geometry classifications for PyROS uncertainty set objects.
@@ -524,6 +533,14 @@ class UncertaintySet(object, metaclass=abc.ABCMeta):
         """
         raise NotImplementedError

+    @property
+    def _cache(self):
+        try:
+            return self.__cache
+        except AttributeError:
+            self.__cache = ContextDict()
+            return self.__cache
+
     def _create_bounding_model(self):
         """
         Make uncertain parameter value bounding problems (optimize
```

Then you could just include the cache management in with the main timer:
```diff
diff --git a/pyomo/contrib/pyros/pyros.py b/pyomo/contrib/pyros/pyros.py
index 38104e98fa..c9fb37e611 100644
--- a/pyomo/contrib/pyros/pyros.py
+++ b/pyomo/contrib/pyros/pyros.py
@@ -386,10 +386,13 @@ class PyROS(object):
         )

         model_data = ModelData(original_model=model, timing=TimingData(), config=None)
-        with time_code(
-            timing_data_obj=model_data.timing,
-            code_block_name="main",
-            is_main_timer=True,
+        with (
+            uncertainty_set._cache,
+            time_code(
+                timing_data_obj=model_data.timing,
+                code_block_name="main",
+                is_main_timer=True,
+            ),
         ):
             kwds.update(
                 dict(
```

pyomo/contrib/pyros/tests/test_grcs.py
jas-yao: No. In general, we would still need a subsolver to calculate uncertain parameter bounds the first time before caching, but ipopt works as well. I have changed this to use ipopt.

pyomo/contrib/pyros/tests/test_grcs.py
jas-yao: I have changed this to use ipopt.

pyomo/contrib/pyros/tests/test_grcs.py
jas-yao: Thanks for the suggestion here. I have changed it to 0.1.

jas-yao:
@jsiirola Thanks for all the suggestions. I have added the `ContextDict` and simplified the cache clearing for the `pyros.py` code.

A small difference was that I implemented `__enter__` with `self.clear()` rather than `assert not self` to clear the cache before starting the PyROS routine rather than erroring if the cache is not empty.
```python
class ContextDict(dict):
    def __enter__(self):
        self.clear()
        return self

    def __exit__(self, et, e, tb):
        self.clear()
```

pyomo/contrib/pyros/uncertainty_sets.py
jsiirola: @jas-yao: Note that `var.lower` returns the expression for the lower bound of the variable.  For cases where the bound is defined by a Param, then you get the Param back.  If the bound is determined by both the bound from `bounds=` and the implied bound from the domain, then you will get a `max` expression back.  The same applies to the upper bound.

If you always want the current numerical bound, then use `var.lb` / `var.ub`

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -460,6 +460,15 @@ def generate_shape_str(shape, required_shape):
         )
 
 
+class ContextDict(dict):
+    def __enter__(self):
+        self.clear()
```
jsiirola: The concern I have around clearing this here, is that the cache should only be non-empty if we are already in the middle of a PyROS solve.  That should never happen (hence why I originally suggested checking that with an assertion).  With this implementation, the cache will be cleared out from under whatever was using it, and would end up with a) we would never know the assumption of only one PyROS using the cache at a time was being violated, and b) whatever was using the cache could unexpectedly fail in a very hard-to-debug manner.

pyomo/contrib/pyros/uncertainty_sets.py
jas-yao: Understood. I have made this change to use `var.lb` / `var.ub`. It will be included in my next push.

pyomo/contrib/pyros/uncertainty_sets.py
jas-yao: Thanks for this feedback. I'll change it back to the assertion then.
To give a bit more context, the idea to clear in `__enter__` is if the user had modified the cache or performed bounds calculations and stored them in the cache of the uncertainty set before running PyROS with it.
The tests in `test_pyros_cache_solutions` are to check that clearing the cache prevents keeps PyROS from using incorrect/outdated cached bounds to end up at a wrong solution.

I think that checking with an assertion should have no problems, as it does the same thing in preventing incorrectly cached bounds from messing with PyROS, along with guarding from unexpected scenarios. There should be no issues in changing the uncertainty set within a loop as well, since the cache is cleared after every run. I think the main trigger for the assertion error would be caused by someone experimenting with a custom uncertainty set.

How does the following look for the error message? 
```python
class ContextDict(dict):
    def __enter__(self):
        assert not self, "Uncertainty set cache has been modified."
        return self

    def __exit__(self, et, e, tb):
        self.clear()
```

shermanjasonaf:
@jas-yao @jsiirola  As PR #3927 has been merged ahead of this PR, we will need to add bounds caching for the new `CartesianProductSet` class here. Also, the PyROS version number and changelog were not updated ahead of merging #3927. I have opened #3951 to address that:

1. If it makes sense to have separate version numbers for #3927 and this PR, then let's wait until #3951 is merged before updating the version number and changelog in the present PR.
2. Otherwise, we can close #3951 and perform a single version number + changelog bump here.

blnicho:
@shermanjasonaf I vote for option 2 and doing a single update to the PyROS version and changelog

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -460,6 +460,15 @@ def generate_shape_str(shape, required_shape):
         )
 
 
+class ContextDict(dict):
+    def __enter__(self):
+        assert not self, "Uncertainty set cache has been modified."
```
shermanjasonaf: Consider modifying this to the effect of:

```suggestion
        assert not self, "Nonempty cache for uncertainty set's exact parameter bounds."
```

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -524,6 +533,14 @@ def parameter_bounds(self):
         """
         raise NotImplementedError
 
+    @property
+    def _cache(self):
```
shermanjasonaf, line 529: Consider documenting this attribute by adding here a docstring to the effect of:

```
        """
        dict : Cache for the bounds defining the minimum bounding box
        of `self`. Each key is a 2-tuple containing the positional
        index of a coordinate and an
        :class:`~pyomo.common.enums.ObjectiveSense` object
        indicating the type of bound (`minimize` for a lower bound,
        `maximize for an upper bound) being specified for the
        coordinate.
        """
```

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -790,44 +808,94 @@ def _compute_exact_parameter_bounds(self, solver, index=None):
...
             # add parameter bounds for current dimension
             param_bounds.append(tuple(bounds))
 
+        return param_bounds
+
+    def _solve_exact_bounds_optimization(self, bounding_model, index, sense, solver):
+        """
+        Compute value of bounds for a single parameter
+        of `self` at a specified index by solving a bounding model.
+        Results are cached as efficiency.
```
shermanjasonaf:
```suggestion
        Compute an exact lower or upper bound for a specified
        coordinate of the points contained in `self`, by solving
        a bounding model.
        
        For efficiency, the result is cached if the bounding model
        is solved successfully. Further, if the cache already contains
        an entry corresponding to the coordinate bound of interest,
        then that entry is returned and solution of the bounding model
        is skipped.
```

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -790,44 +808,94 @@ def _compute_exact_parameter_bounds(self, solver, index=None):
...
+
+        Parameters
+        ----------
+        bounding_model : ConcreteModel
+            Bounding model, with an indexed minimization sense
+            Objective with name 'param_var_objectives' consisting
+            of `N` entries, all of which have been deactivated.
+        index : int
+            The index of the parameter to solve for bounds.
```
shermanjasonaf:
```suggestion
            The positional index for the coordinate of interest.
```

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -790,44 +808,94 @@ def _compute_exact_parameter_bounds(self, solver, index=None):
...
+        sense : ~pyomo.common.ObjectiveSense
+            A Pyomo objective sense to optimize for the bounding model.
+            `maximize` solves for an upper bound and
+            `minimize` solves for a lower bound.
+        solver : ~pyomo.opt.base.solvers.OptSolver
+            Optimizer to invoke on the bounding problems.
```
shermanjasonaf:
```suggestion
            Optimizer to invoke on the bounding model.
```

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -790,44 +808,94 @@ def _compute_exact_parameter_bounds(self, solver, index=None):
...
+        bounding_model : ConcreteModel
+            Bounding model, with an indexed minimization sense
+            Objective with name 'param_var_objectives' consisting
+            of `N` entries, all of which have been deactivated.
+        index : int
+            The index of the parameter to solve for bounds.
+        sense : ~pyomo.common.ObjectiveSense
+            A Pyomo objective sense to optimize for the bounding model.
+            `maximize` solves for an upper bound and
+            `minimize` solves for a lower bound.
```
shermanjasonaf:
```suggestion
            Optimization sense for the bounding model objective.
            This also indicates the type of bound (lower or upper)
            to be computed.
            Select `minimize` to compute the lower bound or
            `maximize` to compute the upper bound.
```

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -790,44 +808,94 @@ def _compute_exact_parameter_bounds(self, solver, index=None):
...
+        -------
+        bound : float
+            A value of the lower or upper bound for
+            the corresponding dimension at the specified index.
+
+        Raises
+        ------
+        ValueError
+            If solver failed to compute a bound for a
+            coordinate.
```
shermanjasonaf:
```suggestion
            If there was an unsuccessful attempt to solve
            the bounding model.
```

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -790,44 +808,94 @@ def _compute_exact_parameter_bounds(self, solver, index=None):
...
-                if check_optimal_termination(res):
-                    bounding_model.solutions.load_from(res)
-                else:
-                    raise ValueError(
-                        "Could not compute "
-                        f"{'lower' if sense == minimize else 'upper'} "
-                        f"bound in dimension {idx + 1} of {self.dim}. "
-                        f"Solver status summary:\n {res.solver}."
+                bounds.append(
+                    self._solve_exact_bounds_optimization(
```
shermanjasonaf, line 835: It may be worth noting/commenting here that (due to the setup of `self._solve_exact_bounds_optimization()`) the variables of the bounding model are initialized to the solution that was loaded to the model in most recent iteration where the model was solved successfully; also, the variables are not initialized at all the first time `self._solve_exact_bounds_optimization()` is called in this loop.

pyomo/contrib/pyros/CHANGELOG.txt
```diff
@@ -3,6 +3,12 @@ PyROS CHANGELOG
 ===============
 
 
+-------------------------------------------------------------------------------
+PyROS 1.3.14    20 Mar 2026
```
shermanjasonaf: After PR #3951 is either merged or closed, you may update the PyROS version number (in `pyros.py`) and further update the changelog here according to [this comment](https://github.com/Pyomo/pyomo/pull/3877#issuecomment-4434296082). In particular, if #3951 is closed rather than merged, then the final version number here will be 1.3.14.

pyomo/contrib/pyros/tests/test_grcs.py
```diff
@@ -5232,5 +5232,252 @@ def test_discrete_set_subsolver_error_recovery(self, name, sec_con_UB):
...
+        self.sleep_time = sleep_time
+        self.cache = cache
+
+    @property
+    def parameter_bounds(self):
+        """
+        Solve bounding problems to calculate exact parameter bounds.
+        """
+        solver = SlowSolver(
+            sub_solver=SolverFactory("baron"), sleep_time=self.sleep_time
```
shermanjasonaf: Are you sure that BARON is needed here? If so, then shouldn't the tests defined in `TestPyROSCache` be skipped unless BARON is available?

pyomo/contrib/pyros/tests/test_grcs.py
```diff
@@ -5232,5 +5232,252 @@ def test_discrete_set_subsolver_error_recovery(self, name, sec_con_UB):
...
+        print("I slept. Now, I will solve.")
+        # invoke subsolver
+        results = self.sub_solver.solve(model, **kwargs)
+
+        return results
+
+
+class CustomExactBoundsUncertaintySet(BoxSet):
+    """
+    Custom uncertainty set that always solves optimization bounding problems.
```
shermanjasonaf: Consider modifying this to be more precise about the "optimization bounding problems". Something to the effect of:

```suggestion
    A box set, modified such that the `parameter_bounds`
    attribute calculation involves globally solving the coordinate
    value bounding problems.
```

pyomo/contrib/pyros/tests/test_grcs.py
```diff
@@ -5232,5 +5232,252 @@ def test_discrete_set_subsolver_error_recovery(self, name, sec_con_UB):
...
+
+        # check cache has been cleared after solve
+        self.assertTrue(hasattr(interval, "_cache"))
+        self.assertEqual(
+            interval._cache, {}, msg="Did not clear uncertainty set cache after solve."
+        )
+
+    def test_pyros_cache_time(self):
+        """
+        Check that caching improves solve time.
```
shermanjasonaf:
```suggestion
        Check that the PyROS solve time is improved when
        the uncertainty set's exact coordinate value bounds are cached.
```

pyomo/contrib/pyros/tests/test_grcs.py
```diff
@@ -5232,5 +5232,252 @@ def test_discrete_set_subsolver_error_recovery(self, name, sec_con_UB):
...
+            self._cache.clear()
+
+        return bounds
+
+
+@unittest.skipUnless(ipopt_available, "IPOPT is not available.")
+class TestPyROSCache(unittest.TestCase):
+    """
+    Test PyROS cache creation and clearing.
+    """
```
shermanjasonaf: Consider altering the name and docstring here to be more descriptive (of the "cache"). Perhaps something to the effect of


```suggestion
class TestPyROSCacheUncertaintySetBounds(unittest.TestCase):
    """
    Test behavior of PyROS solver with respect to caching of
    an uncertainty set's exact coordinate value bounds.
    """
```

pyomo/contrib/pyros/tests/test_grcs.py
```diff
@@ -5232,5 +5232,252 @@ def test_discrete_set_subsolver_error_recovery(self, name, sec_con_UB):
...
+                uncertain_params=[m.u],
+                uncertainty_set=interval,
+                local_solver=local_subsolver,
+                global_solver=global_subsolver,
+                options={
+                    "objective_focus": ObjectiveType.worst_case,
+                    "solve_master_globally": True,
+                },
+            )
+
```
shermanjasonaf, line 5468: You may also want to check here that state of the cache is unchanged by PyROS if the `AssertionError` is raised:

```suggestion

        self.assertAlmostEqual(interval._cache[0, minimize], 25)
        self.assertAlmostEqual(interval._cache[0, maximize], 200)
```

pyomo/contrib/pyros/tests/test_grcs.py
```diff
@@ -5232,5 +5232,252 @@ def test_discrete_set_subsolver_error_recovery(self, name, sec_con_UB):
...
+                second_stage_variables=[m.x2],
+                uncertain_params=[m.u],
+                uncertainty_set=interval,
+                local_solver=local_subsolver,
+                global_solver=global_subsolver,
+                options={
+                    "objective_focus": ObjectiveType.worst_case,
+                    "solve_master_globally": True,
+                },
+            )
```
shermanjasonaf: It seems that the first two assignments here:

```python
        interval._cache[0, minimize] = 25
        interval._cache[0, maximize] = 200
```

do not alter the state of `interval._cache` (from `{(0, minimize): 25, (0, maximize): 200}`) and that the `AssertionError` triggered by calling PyROS with a nonempty bounds cache was already tested for in the earlier lines. Consequently, are you sure that we need any of these additional checks?

pyomo/contrib/pyros/uncertainty_sets.py
shermanjasonaf, line 4223: With #3927 merged, `CartesianProductSet._compute_exact_parameter_bounds()` should be modified to more carefully manage caching of the bounds for instances of `CartesianProductSet`, including all the multiplicand `UncertaintySet` instances stored in `CartesianProductSet._all_sets`. I would change this to something like:

```python
        obj_sense_order = (minimize, maximize)

        if index is None:
            index = [(True, True)] * self.dim

        # for convenience, cast to array
        index = np.array(index, dtype=bool)

        # we will compute the desired bounds as follows:
        # 1. retrieve what we can from the bounds cache.
        # 2. retrieve everything else by solving bounding problems.
        # we will use the following array to track what "everything else" will be
        compute_bounds_index = np.array(index, dtype=bool)

        # retrieve what we can from the bounds cache
        cache_param_bounds = np.ma.array(
            [(np.nan, np.nan)] * self.dim, mask=~index, fill_value=None, dtype=object
        )
        for (dim_idx, bound_idx), bound_bool in np.ndenumerate(index):
            obj_sense = obj_sense_order[bound_idx]
            if bound_bool and (dim_idx, obj_sense) in self._cache:
                cache_param_bounds[dim_idx, bound_idx] = self._cache[dim_idx, obj_sense]
                compute_bounds_index[dim_idx, bound_idx] = False

        # retrieve everything else we need by solving bounding problems
        comp_param_bounds = np.ma.array(
            [(np.nan, np.nan)] * self.dim,
            mask=~compute_bounds_index,
            fill_value=None,
            dtype=object,
        )
        comp_param_bounds.harden_mask()
        for start_dim, stop_dim, uset in self._iterate_over_all_sets():
            # use context manager to ensure that the bounds
            # cache of the multiplicand set is cleared when done
            with uset._cache:
                comp_param_bounds[start_dim:stop_dim] = np.array(
                    # compute only the desired bounds that were not already
                    # retrieved from the cache of `self`
                    uset._compute_exact_parameter_bounds(
                        solver, index=compute_bounds_index[start_dim:stop_dim]
                    )
                )

        # cache all the new bounding model solutions results
        for dim_idx, bound_idx in np.argwhere(~comp_param_bounds.mask).tolist():
            obj_sense = obj_sense_order[bound_idx]
            self._cache[dim_idx, obj_sense] = comp_param_bounds[dim_idx, bound_idx]

        # combine cache retrieval and bounding model solution results
        param_bounds = cache_param_bounds.copy()
        param_bounds[~comp_param_bounds.mask] = comp_param_bounds[
            ~comp_param_bounds.mask
        ]

        return param_bounds.tolist()
```

We may also want to add to `test_grcs::TestPyROSSolveCartesianProductSet` a test or two to confirm that management of the caches is performed according to expectations.

Thoughts?

shermanjasonaf, line 4265: With #3927 merged, should we place this block under a `with uset._cache:` statement, to ensure that the bounds cache for the multiplicand set `uset` is cleared?

shermanjasonaf, line 3952: Isn't it possible that `a_set._cache` gets populated here, particularly for cases where calling `a_set.validate()` entails calling `a_set._compute_exact_parameter_bounds()`? Do we need to be careful to clear `a_set._cache` here?

pyomo/contrib/pyros/tests/test_grcs.py
```diff
@@ -5232,5 +5232,252 @@ def test_discrete_set_subsolver_error_recovery(self, name, sec_con_UB):
...
+        )
+
+        # caching should always result in less time,
+        # as not caching reruns the slow solver multiple times
+        self.assertGreater(results_no_cache.time, results_cache.time)
+
+    def test_pyros_cache_solutions(self):
+        """
+        Check that PyROS clears cache before/after and yields accurate results.
+        """
```
shermanjasonaf: The test name and docstring here can be modified for consistency with what this test is checking. It seems this test checks that an `AssertionError` is raised by `ContextDict.__enter__()` if the cache is nonempty when PyROS is called.

```suggestion
    def test_pyros_solve_assert_bounds_cache_empty(self):
        """
        Check that calling PyROS on an uncertainty set
        with a nonempty bounds cache results in an exception.
        """
```

pyomo/contrib/pyros/tests/test_grcs.py
```diff
@@ -5232,5 +5232,252 @@ def test_discrete_set_subsolver_error_recovery(self, name, sec_con_UB):
...
+
+        # Define subsolvers utilized in the algorithm
+        local_subsolver = SolverFactory("ipopt")
+        global_subsolver = SolverFactory("ipopt")
+
+        # check cache exists
+        self.assertTrue(hasattr(interval, "_cache"))
+
+        # Call the PyROS solver
+        results = pyros_solver.solve(
```
shermanjasonaf: Since `results` is not used after this line:

```suggestion
        pyros_solver.solve(
```

pyomo/contrib/pyros/tests/test_grcs.py
```diff
@@ -5232,5 +5232,252 @@ def test_discrete_set_subsolver_error_recovery(self, name, sec_con_UB):
...
+        pyros_solver = SolverFactory("pyros")
+
+        # Define subsolvers utilized in the algorithm
+        local_subsolver = SolverFactory("ipopt")
+        global_subsolver = SolverFactory("ipopt")
+
+        # Solve with PyROS
+        exc_str = r"Uncertainty set cache has been modified."
+        with self.assertRaisesRegex(AssertionError, exc_str):
+            results = pyros_solver.solve(
```
shermanjasonaf: Since `results` is not used after this line:

```suggestion
            pyros_solver.solve(
```

pyomo/contrib/pyros/tests/test_grcs.py
```diff
@@ -5232,5 +5232,252 @@ def test_discrete_set_subsolver_error_recovery(self, name, sec_con_UB):
...
+        # modify the cache
+        interval._cache[0, minimize] = 25
+        interval._cache[0, maximize] = 200
+
+        self.assertEqual(interval.parameter_bounds, [(25, 200)])
+
+        # Solve with PyROS
+        exc_str = r"Uncertainty set cache has been modified."
+        with self.assertRaisesRegex(AssertionError, exc_str):
+            results = pyros_solver.solve(
```
shermanjasonaf: Since `results` is not used after this line:

```suggestion
            pyros_solver.solve(
```

shermanjasonaf (commented):
After the more recent changes, this PR looks good overall. I have comments on functionality for `IntersectionSet` and `CartesianProductSet` cases. I also have comments on testing and several suggestions on documentation.

@jas-yao To address the second question of [your earlier comment](https://github.com/Pyomo/pyomo/pull/3877#issuecomment-4392041686): I do not think that we should make `UncertaintySet.parameter_bounds` return the results of `UncertaintySet._compute_exact_parameter_bounds()` by default, for a few reasons:

- `_compute_exact_parameter_bounds()` formally requires the user to provide a solver, whereas `parameter_bounds` does not.
- `parameter_bounds` is a property (basically an attribute), so we may want the calculations performed within to be somewhat quick. In contrast,  `_compute_exact_parameter_bounds()` is generally slow when the bounds are not cached, due to the time required to construct and repeatedly solve the bounding model.
- `parameter_bounds` is public (and therefore accessible). If the default behavior of `parameter_bounds` is changed in the suggested way, then accessing `parameter_bounds` results by default in population of the cache, a protected attribute that must be clear(ed) ahead of calling `PyROS.solve()`.

pyomo/contrib/pyros/tests/test_grcs.py
jas-yao: Thanks for catching this. I have changed it to IPOPT to be in line with all the other solvers used in `TestPyROSCache`

pyomo/contrib/pyros/tests/test_grcs.py
jas-yao: That makes sense, I will remove it. 

Originally, this test was intended to check that the cache was cleared before solving, and so the first and second parts were different ways of modifying the `._cache` (indirectly through parameter bounds calculation and directly changing `._cache` values, with the test checking that the solution was not impacted.

Now, the purpose of this test has changed to check that the assert `._cache` empty works correctly.

pyomo/contrib/pyros/uncertainty_sets.py
jsiirola: Is that necessary?  The constitutive uncertainty sets all have their own caches, which will cache the results for the calls to `_solve_exact_bounds_optimization` as necessary (via `_compute_exact_parameter_bounds`).  Given the recent updates to the cache manager, I don't think there is any special processing we need here?

I do agree that some additional tests would be helpful.

jsiirola:
@shermanjasonaf @jas-yao: I pushed the updated cache context manager.  Are we good to "final review" and merge?  (@shermanjasonaf: are you satisfied with how caching interacts with the `CartesianProductSet`, or do you want to wait for additional testing?)

pyomo/contrib/pyros/uncertainty_sets.py
shermanjasonaf: I agree that, with the addition of `CartesianProductSet._cache_manager()`, these changes are no longer necessary.

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -3991,6 +4078,19 @@ def parameter_bounds(self):
...
+            # Verify this (CartesianProductSet's) cache is empty
+            stack.enter_context(super()._cache_manager())
+            for uset in self._all_sets:
+                # Verify all component caches are empty
+                stack.enter_context(uset._cache_manager())
+            yield self
+            # This will re-enter when this context manager is exited,
+            # which will exit the stack context, triggering all the
+            # context managers entered above to exit.
+
```
shermanjasonaf, line 4106: This works nicely. Should `IntersectionSet._cache_manager()` be implemented similarly, to ensure that the bounds caches of an `IntersectionSet`'s operand `UncertaintySet` objects are also cleared?

pyomo/contrib/pyros/tests/test_grcs.py
```diff
@@ -5232,5 +5232,242 @@ def test_discrete_set_subsolver_error_recovery(self, name, sec_con_UB):
...
+                global_solver=global_subsolver,
+                options={
+                    "objective_focus": ObjectiveType.worst_case,
+                    "solve_master_globally": True,
+                },
+            )
+
+        self.assertAlmostEqual(interval._cache[0, minimize], 25, places=2)
+        self.assertAlmostEqual(interval._cache[0, maximize], 200, places=2)
+
```
shermanjasonaf, line 5471: I would add here a test based on a problem with a `CartesianProductSet`, to formally check that `CartesianProductSet._cache_manager()` works according to expectations. Something like:

```diff
diff --git a/pyomo/contrib/pyros/tests/test_grcs.py b/pyomo/contrib/pyros/tests/test_grcs.py
index 410593c82..8f577d879 100644
--- a/pyomo/contrib/pyros/tests/test_grcs.py
+++ b/pyomo/contrib/pyros/tests/test_grcs.py
@@ -75,6 +75,7 @@ from pyomo.contrib.pyros.uncertainty_sets import (
     FactorModelSet,
     Geometry,
     IntersectionSet,
+    PolyhedralSet,
     UncertaintyQuantification,
     UncertaintySet,
 )
@@ -5468,6 +5469,71 @@ class TestPyROSCacheUncertaintySetBounds(unittest.TestCase):
         self.assertAlmostEqual(interval._cache[0, minimize], 25, places=2)
         self.assertAlmostEqual(interval._cache[0, maximize], 200, places=2)
 
+    def test_solve_cartesian_product_set_bounds_cache(self):
+        """
+        Test management of uncertainty set bounds caches
+        is carried out as expected in the context of a PyROS solve.
+        """
+        # deterministic model
+        m = ConcreteModel()
+        m.q = Param(range(4), initialize=0, mutable=True)
+        m.x = Var(initialize=0, bounds=(0, 100))
+        m.obj = Objective(expr=m.x, sense=minimize)
+        m.con = Constraint(expr=m.x >= sum(m.q.values()))
+
+        # uncertainty set(s)
+        poly_set = PolyhedralSet(
+            # this is just the cube [-1, 1]^3
+            lhs_coefficients_mat=np.vstack([np.eye(3), -np.eye(3)]),
+            rhs_vec=[1] * 6,
+        )
+        # inclusion of PolyhedralSet means bounds caching takes place
+        cpset = CartesianProductSet([BoxSet([[0, 1]]), poly_set])
+        res1 = SolverFactory("pyros").solve(
+            model=m,
+            first_stage_variables=m.x,
+            second_stage_variables=[],
+            uncertain_params=m.q,
+            uncertainty_set=cpset,
+            local_solver=SolverFactory("ipopt"),
+            global_solver=SolverFactory("ipopt"),
+            objective_focus="worst_case",
+            solve_master_globally=True,
+        )
+
+        # check caches cleared
+        self.assertEqual(cpset._cache, {})
+        self.assertEqual(cpset._all_sets[0]._cache, {})
+        self.assertEqual(cpset._all_sets[1]._cache, {})
+        # check results: worst case objective is just sum of the
+        #                uncertainty set upper bounds
+        self.assertEqual(res1.iterations, 2)
+        self.assertAlmostEqual(res1.final_objective_value, 4.0, places=6)
+        self.assertAlmostEqual(m.x.value, 4.0, places=6)
+
+        # expand the polyhedralset to the cube [-2, 2]^3
+        poly_set.rhs_vec = [2] * 6
+        # solve again. since caches cleared, PyROS should work normally
+        res2 = SolverFactory("pyros").solve(
+            model=m,
+            first_stage_variables=m.x,
+            second_stage_variables=[],
+            uncertain_params=m.q,
+            uncertainty_set=cpset,
+            local_solver=SolverFactory("ipopt"),
+            global_solver=SolverFactory("ipopt"),
+            objective_focus="worst_case",
+            solve_master_globally=True,
+        )
+        # check caches cleared
+        self.assertEqual(cpset._cache, {})
+        self.assertEqual(cpset._all_sets[0]._cache, {})
+        self.assertEqual(cpset._all_sets[1]._cache, {})
+        # results have changed, since the polyhedral set was expanded
+        self.assertEqual(res2.iterations, 2)
+        self.assertAlmostEqual(res2.final_objective_value, 7.0, places=6)
+        self.assertAlmostEqual(m.x.value, 7.0, places=6)
+
 
 if __name__ == "__main__":
     unittest.main()
```

shermanjasonaf (commented):
@jsiirola @jas-yao I think we are close to merging. The new cache context manager works nicely. I am satisfied with how caching interacts with the `CartesianProductSet`; my comments include suggestions for additional testing.  We may also want a similar custom implementation for `IntersectionSet._cache_manager()`, as bounds caching of the operand `UncertaintySet` attributes may also occur in some (seemingly rare) cases.

pyomo/contrib/pyros/tests/test_grcs.py
jas-yao: I have added this. I renamed `test_solve_cartesian_product_set_bounds_cache` to `test_solve_cartesian_product_and_intersection_set_bounds_cache` and added a couple additional checks for `IntersectionSet`.

pyomo/contrib/pyros/uncertainty_sets.py
jas-yao: Yes, that makes sense to me. I have added this.

jas-yao:
@jsiirola @shermanjasonaf , I've taken a look at the caching behavior with the latest updates and everything works as expected. I added the suggestions made by @shermanjasonaf for `IntersectionSet._cache_manager()` and tests.

Looks good to me for merging, pending any final feedback on my latest changes.

jsiirola (approved):

pyomo/contrib/pyros/tests/test_grcs.py
```diff
@@ -5232,5 +5233,359 @@ def test_discrete_set_subsolver_error_recovery(self, name, sec_con_UB):
             )
 
 
+# @SolverFactory.register("slow_solver")
```
mrmundt, line 5236: Remove line?

pyomo/contrib/pyros/uncertainty_sets.py
```diff
@@ -3744,6 +3831,19 @@ def parameter_bounds(self):
...
+        with contextlib.ExitStack() as stack:
+            # Verify this (IntersectionSet's) cache is empty
+            stack.enter_context(super()._cache_manager())
+            for uset in self.all_sets:
+                # Verify all component caches are empty
+                stack.enter_context(uset._cache_manager())
+            yield self
+            # This will re-enter when this context manager is exited,
+            # which will exit the stack context, triggering all the
+            # context managers entered above to exit.
```
mrmundt, line 3845: It's not a showstopper, but this exact code is repeated, no differences, in 4095-4105. It'd be better to _not_ have repeat code.

blnicho (approved):

pyomo/contrib/pyros/tests/test_grcs.py
```diff
@@ -5232,5 +5233,359 @@ def test_discrete_set_subsolver_error_recovery(self, name, sec_con_UB):
...
+        self.assertAlmostEqual(m.x.value, 7.0, places=6)
+
+        # repeat above checks for intersection set
+        # update poly_set
+        poly_set = PolyhedralSet(
+            # this is just the cube [-1, 1]^4
+            lhs_coefficients_mat=np.vstack([np.eye(4), -np.eye(4)]),
+            rhs_vec=[2] * 8,
+        )
+        iset = IntersectionSet(set1=BoxSet([[-3, 3]] * 4), set2=poly_set)
```
shermanjasonaf: I would modify this set so that the results of the `IntersectionSet`-based checks are more responsive to the implementation of `IntersectionSet._cache_manager()`. 

When `PyROS.solve()` is called on this `IntersectionSet` instance, the cache for neither of the two operand sets is populated, as `_compute_exact_parameter_bounds()` is not invoked at all for either the `BoxSet` (since `BoxSet.parameter_bounds` returns finite, exact parameter bounds) or the `PolyhedralSet` (since, for this case, `PolyhedralSet.is_bounded()` uses FBBT). Thus, the checks that follow would pass even if `IntersectionSet._cache_manager()` were identical to `UncertaintySet._cache_manager()`.

In contrast, when `PyROS.solve()` is called on an `IntersectionSet` with an operand `CustomExactBoundsUncertaintySet(..., cache=True)` instance, the bounds cache for the operand `CustomExactBoundsUncertaintySet` is populated, as accessing `CustomExactBoundsUncertaintySet.parameter_bounds` populates the cache. So the result of checking the state of that operand's cache after calling `PyROS.solve()` would change if `IntersectionSet._cache_manager()` were identical to `UncertaintySet._cache_manager()`.

Thus, consider this change:
 
```suggestion
        iset = IntersectionSet(
            set1=CustomExactBoundsUncertaintySet(
                bounds=[[-3, 3]] * 4, sleep_time=0, cache=True
            ),
            set2=poly_set,
        )
```

shermanjasonaf (approved):
@jas-yao Looks good to me. I have just one suggestion for testing.

pyomo/contrib/pyros/tests/test_grcs.py
jas-yao: Thanks for pointing this out. I have added this change.

At least on my end, FBBT fails to return bounds for the polyhedral set, so `_compute_exact_parameter_bounds` is run and the test is useful for checking that the `PolyhedralSet` cache is clear.
Regardless, adding this change is good since it guarantees that the caching will occur.

pyomo/contrib/pyros/tests/test_grcs.py
jas-yao: I added this to copy the format the other solvers in `test_grcs.py` were written with.
If this is removed, should we also remove `# @SolverFactory.register("time_delay_solver")` and `# @SolverFactory.register("subsolver_error__solver")`?

pyomo/contrib/pyros/uncertainty_sets.py
jas-yao: I tried to update the implementation to move to move the repeated code to a single class (shown below). Would this make sense?

```diff
1 file changed, 21 insertions(+), 28 deletions(-)
pyomo/contrib/pyros/uncertainty_sets.py | 49 ++++++++++++++-------------------

modified   pyomo/contrib/pyros/uncertainty_sets.py
@@ -1283,6 +1283,25 @@ class UncertaintySetList(MutableSequence):
         return self._dim
 
 
+class UncertaintySetOperation:
+    """
+    Mixin class for uncertainty sets that perform operations
+    on other uncertainty sets (e.g., IntersectionSet, CartesianProductSet).
+    """
+    @contextlib.contextmanager
+    def _cache_manager(self):
+        with contextlib.ExitStack() as stack:
+            # Verify this set's cache is empty
+            stack.enter_context(super()._cache_manager())
+            for uset in self._all_sets:
+                # Verify all component caches are empty
+                stack.enter_context(uset._cache_manager())
+            yield self
+            # This will re-enter when this context manager is exited,
+            # which will exit the stack context, triggering all the
+            # context managers entered above to exit.
+
+
 class BoxSet(UncertaintySet):
     """
     A hyperrectangle (or box).
@@ -3676,7 +3695,7 @@ class DiscreteScenarioSet(UncertaintySet):
         )
 
 
-class IntersectionSet(UncertaintySet):
+class IntersectionSet(UncertaintySetOperation, UncertaintySet):
     """
     An intersection of two or more uncertainty sets, each of which
     is represented by an `UncertaintySet` object.
@@ -3831,19 +3850,6 @@ class IntersectionSet(UncertaintySet):
 
         return []
 
-    @contextlib.contextmanager
-    def _cache_manager(self):
-        with contextlib.ExitStack() as stack:
-            # Verify this (IntersectionSet's) cache is empty
-            stack.enter_context(super()._cache_manager())
-            for uset in self.all_sets:
-                # Verify all component caches are empty
-                stack.enter_context(uset._cache_manager())
-            yield self
-            # This will re-enter when this context manager is exited,
-            # which will exit the stack context, triggering all the
-            # context managers entered above to exit.
-
     def point_in_set(self, point):
         """
         Determine whether a given point lies in the intersection set.
@@ -3955,7 +3961,7 @@ class IntersectionSet(UncertaintySet):
         super().validate(config)
 
 
-class CartesianProductSet(UncertaintySet):
+class CartesianProductSet(UncertaintySetOperation, UncertaintySet):
     """
     A Cartesian product of uncertainty sets.
 
@@ -4091,19 +4097,6 @@ class CartesianProductSet(UncertaintySet):
                 return []
         return parameter_bounds
 
-    @contextlib.contextmanager
-    def _cache_manager(self):
-        with contextlib.ExitStack() as stack:
-            # Verify this (CartesianProductSet's) cache is empty
-            stack.enter_context(super()._cache_manager())
-            for uset in self._all_sets:
-                # Verify all component caches are empty
-                stack.enter_context(uset._cache_manager())
-            yield self
-            # This will re-enter when this context manager is exited,
-            # which will exit the stack context, triggering all the
-            # context managers entered above to exit.
-
     def _iterate_over_all_sets(self):
         """
         Iterate over the sets contained in `self`.
```

PR #3880, opened 2026-03-23
GDP Hull transform: add handling for cases of constraint functions not well-defined at the origin

sadavis1:
## Summary/Motivation:
When applying the gdp.hull transformation to a model containing a constraint function whose evaluation is not well-defined when every variable is fixed at zero, the perspective functions used by the hull transformation are not well-defined either and lead to errors, especially when the mode is set to FurmanSawayaGrossmann (which is the useful one). This PR slightly alters the mathematical formulation to permit a nonzero base point which can be used in place of the origin, and adds a heuristic to try to magically find one by calling gurobi. When the zero point works, it uses that so Gurobi is never invoked.

Note: this is a minor merge conflict with #3874, if that is merged first I can rebase this

- Alter mathematical formulation for gdp.hull transformation to include a base point
- Add a heuristic to get a well-defined point. This is done at the pyomo level by solving a subproblem using a nonlinear solver, even though solvers usually have a way to do this sort of thing directly, because doing it directly requires setting many solver-specific options and in some cases triggers bugs in solvers (this happened for BARON during testing). This is done using a walker to generate constraints corresponding to each potentially ill-defined function evaluation.
- Change the way LocalVars works to permit a category of variables that still need to be disaggregated now, but do not need to worry about global constraints and also do not need to be re-disaggregated in any parent disjunct if the GDP is nested ("generalized local vars"). This category is naturally occupied by local variables that have an offset (since the offset handling happens during disaggregation) but other variables may be placed in this category by the user by marking them LocalVars even though they appear on multiple disjuncts. Ordinary LocalVars occupying only one disjunct behave the same way as before.

### Legal Acknowledgement


pyomo/gdp/plugins/hull.py
```diff
@@ -197,6 +231,43 @@ class Hull_Reformulation(GDP_to_MIP_Transformation):
         """,
         ),
     )
+    CONFIG.declare(
+        'well_defined_points',
+        cfg.ConfigValue(
+            default=ComponentMap(),
+            domain=ComponentMap,
```
emma58: Disjunctions are hashable, so there's no need to require this be a `ComponentMap`: It just needs to be dict-like.

pyomo/gdp/plugins/hull.py
```diff
@@ -197,6 +231,43 @@ class Hull_Reformulation(GDP_to_MIP_Transformation):
...
+    )
+    CONFIG.declare(
+        'well_defined_points_heuristic_solver',
+        cfg.ConfigValue(
+            default='gurobi_direct_minlp',
+            domain=Solver,
+            description="Solver used for the base points heuristic",
+            doc="""
+        Solver used for the base points heuristic. This must be a
+        nonlinear solver in general.
```
emma58: Should we mention that it also needs to be global if the problem is nonconvex?

pyomo/gdp/plugins/hull.py
```diff
@@ -292,6 +363,116 @@ def _add_transformation_block(self, to_block):
...
+            logger.error(
+                "While trying to evaluate an expression, got unexpected exception type "
+                f"{e.__class__.__name__} (was prepared for success or a ValueError)."
+            )
+            raise
+        # Second, try making it well-defined by editing only the regular vars
+        for x in fallback_vars:
+            x.fix(0)
+        for x in regular_vars:
+            x.set_value(0)
```
emma58: Do you know that 0 is in the domain in these two cases?

pyomo/gdp/plugins/hull.py
```diff
@@ -292,6 +363,116 @@ def _add_transformation_block(self, to_block):
...
+            feasible = self._solve_for_first_feasible_solution(test_model)
+            if not feasible:
+                raise GDP_Error(
+                    f"Unable to find a well-defined point on disjunction {disj_name}. "
+                    "To carry out the hull transformation, each disjunction must have a "
+                    "point at which every constraint function appearing in its "
+                    "disjuncts is well-defined and finite. Please ensure such a point "
+                    "actually exists, then if we still cannot find it, override our "
+                    "search process using the `well_defined_points` option."
+                )
```
emma58: This is a bit confusing to me. If we end up here, it means that we haven't found it, right? So isn't the only recourse (if it exists) to use the `well_defined_points` option?

pyomo/gdp/plugins/hull.py
```diff
@@ -292,6 +363,116 @@ def _add_transformation_block(self, to_block):
...
+        # Found a point
+        if not math.isfinite(value(test_expr)):
+            raise DeveloperError("unreachable")
```
emma58: I agree that this is unreachable--I think it can be removed

pyomo/gdp/plugins/hull.py
```diff
@@ -292,6 +363,116 @@ def _add_transformation_block(self, to_block):
...
+    # a nonlinear solver.
+    def _solve_for_first_feasible_solution(self, test_model):
+        results = self._config.well_defined_points_heuristic_solver.solve(
+            test_model, load_solutions=False
+        )
+        if results.solver.termination_condition is TerminationCondition.infeasible:
+            return False
+        if results.solver.status is not SolverStatus.ok:
+            raise GDP_Error(f"Unexpected solver status {results.solver.status}.")
+        test_model.solutions.load_from(results)
```
emma58, line 591: This won't work with any solvers in contrib.solver right now, unless it's through the legacy interface. I think that's okay, but perhaps we should document it in the docstring for `well_defined_points_heuristic_solver`?

pyomo/gdp/plugins/hull.py
```diff
@@ -354,11 +535,18 @@ def _transform_disjunctionData(
...
+                seen_vars = set()
+                for var in IdentifyVariableVisitor(
+                    include_fixed=not self._config.assume_fixed_vars_permanent
+                ).walk_expression(con.expr):
+                    if id(var) in seen_vars:
+                        continue
+                    seen_vars.add(id(var))
+                    if var not in var_order:
+                        var_order.add(var)
+                        disjuncts_var_appears_in[var] = ComponentSet([disjunct])
```
emma58, line 661: This is past Emma being silly, but this can just be a `set` actually, since we only add Disjuncts to it.

pyomo/gdp/plugins/hull.py
```diff
@@ -354,11 +535,18 @@ def _transform_disjunctionData(
...
-                    disjuncts_var_appears_in[var] = ComponentSet([disjunct])
-                else:
-                    disjuncts_var_appears_in[var].add(disjunct)
+                seen_vars = set()
+                for var in IdentifyVariableVisitor(
+                    include_fixed=not self._config.assume_fixed_vars_permanent
+                ).walk_expression(con.expr):
+                    if id(var) in seen_vars:
+                        continue
+                    seen_vars.add(id(var))
```
emma58: The `IdentifyVariableVisitor` returns the unique Vars, so you don't need the `seen_vars` set at all.

pyomo/gdp/plugins/hull.py
```diff
@@ -371,28 +559,58 @@ def _transform_disjunctionData(
...
-                if self._generate_debug_messages:
-                    logger.debug(
-                        "Assuming '%s' is not a local var since it is"
-                        "used in multiple disjuncts." % var.name
-                    )
+            for disj in disjuncts:
+                if disj in local_vars_by_disjunct:
+                    if var in local_vars_by_disjunct[disj]:
+                        if len(disjuncts) == 1:
+                            # was this a noop before?
```
emma58: Yeah, I think we had a bug here before, right? But let's remove this comment because it won't make sense to future us, I don't think.

pyomo/gdp/plugins/hull.py
```diff
@@ -371,28 +559,58 @@ def _transform_disjunctionData(
...
-            if len(disjuncts) > 1:
-                if self._generate_debug_messages:
-                    logger.debug(
-                        "Assuming '%s' is not a local var since it is"
-                        "used in multiple disjuncts." % var.name
-                    )
+            for disj in disjuncts:
+                if disj in local_vars_by_disjunct:
+                    if var in local_vars_by_disjunct[disj]:
+                        if len(disjuncts) == 1:
```
emma58, line 687: This is probably something to save for a separate PR, but this makes me wonder if `local_vars_by_disjunct` is just the wrong data structure... I could be wrong, but I'm wondering if we can get what we need in only one pass through all the Disjuncts x Vars?

pyomo/gdp/plugins/hull.py
```diff
@@ -371,28 +559,58 @@ def _transform_disjunctionData(
...
+            x0_map, offset_vars = self._get_well_defined_point(
+                test_expr=sum(con.body for con in active_constraints),
+                regular_vars=all_vars_to_disaggregate,
+                fallback_vars=all_local_vars,
+                disj_name=obj.name,
+            )
+        transBlock.private_data().well_defined_points_map[obj] = x0_map
+        # Any var that got an offset cannot be local anymore, but it can
+        # still be generalized local
+        # breakpoint()
```
emma58: Looks like a debugging vestige

pyomo/gdp/plugins/hull.py
```diff
@@ -371,28 +559,58 @@ def _transform_disjunctionData(
...
+        for var in offset_vars:
+            if var in all_local_vars:
+                var_disjunct = next(iter(disjuncts_var_appears_in[var]))
```
emma58, line 720: We know there's only one Disjunct here because it's in `all_local_vars`?

pyomo/gdp/plugins/hull.py
```diff
@@ -469,14 +705,17 @@ def _transform_disjunctionData(
             else:
                 disaggregatedExpr = 0
             for disjunct in disjuncts_var_appears_in[var]:
+                # breakpoint()
```
emma58: Can be removed

pyomo/gdp/plugins/hull.py
```diff
@@ -571,19 +813,17 @@ def _transform_disjunct(
...
+        # call _transform_constraint on each constraint.
         self._transform_block_components(
-            obj, obj, var_substitute_map, zero_substitute_map
+            obj, obj, var_substitute_map, x0_substitute_map
         )
 
-        # Anything that was local to this Disjunct is also local to the parent,
-        # and just got "promoted" up there, so to speak.
-        parent_disjunct_local_vars.update(local_vars)
         # deactivate disjunct so writers can be happy
```
emma58, line 935: Why is this removed? (This is for nested--the disaggregated variables for an inner Disjunction are local to the parent Disjunct.)

pyomo/gdp/plugins/hull.py
```diff
@@ -292,6 +363,116 @@ def _add_transformation_block(self, to_block):
...
+    # fixed statuses when we return successfully.
+    def _get_well_defined_point(
+        self, test_expr, regular_vars, fallback_vars, disj_name
+    ):
+        # First, see if test_expr is well-defined at the origin.
+        x0_map = ComponentMap()
+        orig_values = ComponentMap()
+        orig_fixed = ComponentMap()
+        for x in itertools.chain(regular_vars, fallback_vars):
+            x0_map[x] = 0  # ZeroConstant?
```
emma58: @jsiirola, should this be ZeroConstant? Using 0 instead would be a change from the old implementation, but I don't know if it matters?

pyomo/gdp/plugins/hull.py
```diff
@@ -494,7 +733,10 @@ def _transform_disjunct(
         parent_local_var_suffix,
         parent_disjunct_local_vars,
         disjunct_disaggregated_var_map,
+        x0_map,
+        offset_vars,
     ):
+        # breakpoint()
```
emma58: Can be removed

pyomo/gdp/plugins/hull.py
```diff
@@ -676,16 +917,36 @@ def _transform_constraint(
             unique = len(newConstraint)
             name = c.local_name + "_%s" % unique
 
-            NL = c.body.polynomial_degree() not in (0, 1)
+            # HACK: If I'm not treating fixed vars as permanent, I need
+            # the LinearRepnVisitor to pick them up, but it really
+            # doesn't want to. So I'm going to unfix and refix, which
+            # involves walking the expression twice, plus it's generally
+            # stupid
```
emma58: @sadavis1 and I talked about this yesterday. I'm thinking that we might want a "fixed-variable aware" `LinearRepnVisitor` so that we don't have to do this. We could play a similar game as we did with the `ParameterizedLinearRepnVisitor` and inherit from `LinearRepnVisitor` and just reimplement a few handlers so that we can treat fixed variables as variables when we want to. Does this make sense to you, @jsiirola? If so, we can probably leave this for now and do that in a separate PR?

emma58: (And the reason we wouldn't touch the current LinearRepnVisitor is performance, just to be clear.)

pyomo/gdp/plugins/hull.py
```diff
@@ -676,16 +917,36 @@ def _transform_constraint(
...
             # disaggregated variables
             if not NL or mode == "FurmanSawayaGrossmann":
-                h_0 = clone_without_expression_components(
-                    c.body, substitute=zero_substitute_map
+                # Only permanently fixed vars are not being substituted
+                # by the x0_substitute_map
+                h_x0 = value(
+                    clone_without_expression_components(
+                        c.body, substitute=x0_substitute_map
+                    )
```
emma58, line 1110: Was this a bug before?

pyomo/gdp/plugins/hull.py
```diff
@@ -711,42 +973,76 @@ def _transform_constraint(
...
-                    c.body, substitute=var_substitute_map
+                # For a linear constraint that looks like a^Tx + b <= c,
+                # the transformed constraint will be a^Tv <= lambda (c -
+                # b - a^Tx_0). Get the a^Tv here and note that b +
+                # a^Tx_0 is exactly h_x0 from earlier, so we will have
+                # it when we need it.
+                expr = linear_repn.multiplier * sum(
+                    coef * var_substitute_map[var]
+                    for var, coef in linear_repn.linear.items()
+                    if coef != 0
```
emma58: `linear_repn.multiplier` is guaranteed to be 1, so you don't need it here.

pyomo/gdp/plugins/hull.py
```diff
@@ -711,42 +973,76 @@ def _transform_constraint(
...
+                    if len(linear_repn.linear) == 1:
+                        var, coef = next(iter(linear_repn.linear.items()))
+                        # Second clause of this condition happens iff
+                        # the constraint implies x = x0
+                        if (
+                            coef != 0
+                            and (
+                                c.lower - linear_repn.constant * linear_repn.multiplier
+                            )
+                            / (coef * linear_repn.multiplier)
```
emma58: Same comment on the multiplier here.

pyomo/gdp/plugins/hull.py
```diff
@@ -972,6 +1268,18 @@ def get_transformed_constraints(self, cons):
...
+        """
+        Retrieve the well-defined points originally used to transform
+        a Block. Format is a ComponentMap of ComponentMaps identical to
+        that of the parameter well_defined_points.
+
+        Parameters
+        ----------
+        b: a Block that was transformed by gdp.hull
+        """
+        return b._pyomo_gdp_hull_reformulation.private_data().well_defined_points_map
```
emma58: We can't assume that the transformation block is called `_pyomo_gdp_hull_reformulation`--it actually gets a unique name if that name is already taken. So I think the easiest solution is just to put the private_data on `b` itself.

pyomo/gdp/plugins/hull.py
```diff
@@ -986,3 +1294,106 @@ def get_transformed_constraints(self, cons):
...
+    if exp.__class__ in EXPR.native_types or not exp.is_potentially_variable():
+        val = value(exp)
+        if round(val) == val:
+            if val >= 0:
+                return ()
+            else:
+                # return base != 0
+                return (base >= EPS,)
+        elif val >= 0:
+            return (base >= 0,)
```
emma58: If the exponent is nonnegative, why have any restriction on the base?

pyomo/gdp/plugins/hull.py
```diff
@@ -986,3 +1294,106 @@ def get_transformed_constraints(self, cons):
...
+        return (arg >= -1, arg <= 1)
+    # It can't be exactly pi/2 plus a multiple of pi. Rather difficult
+    # to enforce, so make a conservative effort by instead keeping it in
+    # (-pi/2, pi/2).
+    if node.name == 'tan':
+        return (arg >= -(math.pi / 2) + EPS, arg <= (math.pi / 2) - EPS)
+    if node.name == 'acosh':
+        return (arg >= 1,)
+    if node.name == 'atanh':
+        return (arg >= -1 + EPS, arg <= 1 - EPS)
```
emma58: I think it's more efficient to make a dictionary of unary function handlers and use indirection here too. At the very least, these should be `elif`s.

pyomo/gdp/tests/test_hull.py
```diff
@@ -2882,6 +2887,87 @@ def test_dill_pickle(self):
...
+        self.assertNotEqual(m_pts[m.disjunction][m.x], 0)
+        # using the recorded map is the same thing as the process that
+        # created that map
+        TransformationFactory('gdp.hull').apply_to(m2, well_defined_points=m_pts)
+        # how to assert that two blocks have (value) identical contents?
+        for c1, c2 in zip(
+            m.component_data_objects(Constraint, sort=SortComponents.deterministic),
+            m2.component_data_objects(Constraint, sort=SortComponents.deterministic),
+        ):
+            self.assertEqual(str(c1.expr), str(c2.expr))
```
emma58: I think you can write a more robust test here by using `assertExpressionsStructurallyEqual` from `pyomo.core.expr.compare`

pyomo/gdp/tests/test_hull.py
```diff
@@ -2890,3 +2976,45 @@ class NestedDisjunctsInFlatGDP(unittest.TestCase):
...
+            m.x + 4 <= 1,
+            m.x + 5 >= -1,
+            m.x + 5 <= 1,
+            m.x + 6 <= (math.pi / 2) - walker_eps,
+            m.x + 6 >= -(math.pi / 2) + walker_eps,
+            m.x + 7 >= 1,
+            m.x + 8 <= 1 - walker_eps,
+            m.x + 8 >= -1 + walker_eps,
+        ]:
+            self.assertIn(str(con), walked_strings)
```
emma58: Is there a way to use an `assertExpressionsEqual` variant here too?

emma58 (commented):
Thank you so much for this @sadavis1, some of this is quite unpleasant to think through! :P A lot of comments and questions, but overall this is looking good. @jsiirola, I left a couple questions for you in the comments too.

pyomo/gdp/plugins/hull.py
emma58: (If so, can we test it?)

pyomo/gdp/plugins/hull.py
sadavis1: Added a note here

pyomo/gdp/plugins/hull.py
sadavis1: I agree

pyomo/gdp/plugins/hull.py
sadavis1: Added it to the doc. I do validate that the solver is correct since it passes through this Solver method that I'm currently importing from multiple_bigm.py (maybe that function should be moved somewhere more central?)

pyomo/gdp/plugins/hull.py
sadavis1: Fixed

pyomo/gdp/plugins/hull.py
sadavis1: Fixed

pyomo/gdp/plugins/hull.py
sadavis1: fixed

pyomo/gdp/plugins/hull.py
sadavis1: Yes, when we collected local vars earlier, it wouldn't have been put in all_local_vars unless it only appeared in one disjunct. Here we're changing our mind and deciding that our local var can't be a local var anymore.

pyomo/gdp/plugins/hull.py
sadavis1: I think it's the same as before when there is no offset, I just added a comment. Whether fixed vars end up in this map is ultimately controlled by what ends up in `var_order`, and the logic for determining that should not have changed (I do it in a different way because I need to collect constraints too, but that should be the only difference)

pyomo/gdp/plugins/hull.py
sadavis1: Fixed. Actually I'm not sure if it was a bug because I cannot find the thing that I decided was a no-op...

pyomo/gdp/plugins/hull.py
sadavis1: This whole scheme is convoluted so I would not be surprised if it can be done better. I think this loop here is a subset of Disjuncts x Vars since we loop through all Vars, then only the disjuncts that var appears in in a nested loop. Of course, I only know that subset because earlier I first looped Disjuncts then Vars appearing in them, which is a somewhat suspicious structure, but it's an inherently awkward operation because there's no easy built-in way to get what disjuncts a var appears in.

pyomo/gdp/plugins/hull.py
sadavis1: My guess was that it's just as likely we get into this situation by a modeling error as by a failure of the heuristic (maybe more likely), so I want the user to make sure their model is right before they start assembling intricate dicts of componentmaps to pass us.

pyomo/gdp/plugins/hull.py
sadavis1: This was moved to the caller (line 639 now) because we want to promote everything in `generalized_local_vars` to the status of full LocalVars in the parent disjunct, and `_transform_disjunct` doesn't have that list (and if it did, it would be still be unnecessary to do it there since it would be the same in every loop iteration).

pyomo/gdp/plugins/hull.py
emma58: I was confused why you had to add `value` though?

pyomo/gdp/plugins/hull.py
emma58: Ah, I missed that! Thank you!

pyomo/gdp/plugins/hull.py
sadavis1: Oh.. I think you're right. This is surely not going to work correctly with permanent fixed vars without the value() call in place. I'll come up with a test for this.

pyomo/gdp/plugins/hull.py
sadavis1: Fixed this

pyomo/gdp/plugins/hull.py
sadavis1: Fixed

pyomo/gdp/plugins/hull.py
sadavis1: Moved it there

pyomo/gdp/plugins/hull.py
emma58: Ah, okay, makes sense.

pyomo/gdp/plugins/hull.py
emma58: It probably should be moved somewhere more central... @jsiirola, where?

pyomo/gdp/plugins/hull.py
sadavis1: For the x's in fallback_vars, it's an intentional assumption; we want to try it this way first before unfixing those, since the fallback_vars are costlier than the others to set offsets. For the regular_vars we are relying on the solver to disregard the warm start values if it produces an infeasible or ill-defined starting point, however we would like to start with all zeros when we can because a sparser solution is better here. Also as a practical matter we can't leave variables on None here unless we clean up after the solve, since load_from after solve may in some cases not fill in variables we need values for (like if they were eliminated in presolve? not sure all the situations where this can happen). Finally note that this is all moot if we don't have warmstart, which we don't by default (but maybe should?).

pyomo/gdp/plugins/hull.py
sadavis1: Actually this is already effectively tested by `test_handle_fixed_no_disagg` from this PR. If the value() call is removed, then the tested expression has x in it, like in `test_handle_fixed_disagg` above, instead of being trivial like it should be since x is replaced with its value.

pyomo/gdp/plugins/hull.py
sadavis1: I think this should stay a ComponentSet for the sake of determinism. We later iterate through these sets multiple times in ways that will affect the ordering of variables in the final transformed model.

pyomo/gdp/tests/test_hull.py
sadavis1: Fixed

pyomo/gdp/tests/test_hull.py
sadavis1: Fixed this one too

pyomo/gdp/plugins/hull.py
sadavis1: Fixed, but note that the inner maps still need to be ComponentMap.

pyomo/gdp/plugins/hull.py
sadavis1: Fixed

pyomo/gdp/plugins/hull.py
sadavis1: I changed this to also be done with handlers for consistency, though I'm not sure which way is actually faster in the end (arguments could be made either way, since function calls have overhead too).

sadavis1:
Updated for review comments

pyomo/gdp/plugins/hull.py
emma58: OK, thank you, this makes sense!

pyomo/gdp/plugins/hull.py
emma58: Ohhh, right, good catch! Maybe let's add a comment here so future me doesn't forget that again.

emma58 (approved):

pyomo/gdp/plugins/hull.py
```diff
@@ -371,28 +535,56 @@ def _transform_disjunctionData(
...
-                        "used in multiple disjuncts." % var.name
-                    )
+            for disj in disjuncts:
+                if disj in local_vars_by_disjunct:
+                    if var in local_vars_by_disjunct[disj]:
+                        if len(disjuncts) == 1:
+                            local_vars[disj].add(var)
+                            all_local_vars.add(var)
+                        else:
+                            vars_to_disaggregate[disjunct].add(var)
```
bernalde: Blocking: This branch uses the stale outer-loop `disjunct` variable instead of the current `disj`. For a variable marked `LocalVars` while appearing in multiple disjuncts, this only adds the variable to `vars_to_disaggregate` for the last active disjunct, leaving other disjuncts without a substitute. A simple two-disjunct model with `x` in both disjuncts and `LocalVars` entries for both currently fails during transformation with `ValueError: No value for uninitialized ScalarVar object x`.

Please use the current disjunct and add regression coverage for the generalized-local-vars case. The fix may need to ensure every disjunct in `disjuncts_var_appears_in[var]` gets a disaggregated variable once the variable is treated as generalized local.

pyomo/gdp/plugins/hull.py
```diff
@@ -292,6 +345,114 @@ def _add_transformation_block(self, to_block):
...
+            pass
+        except ZeroDivisionError:
+            pass
+        except Exception as e:  # can anything else be thrown here?
+            logger.error(
+                "While trying to evaluate an expression, got unexpected exception type "
+                f"{e.__class__.__name__} (was prepared for success or a ValueError)."
+            )
+            raise
+        # Second, try making it well-defined by editing only the regular vars
```
bernalde: Blocking: `_get_well_defined_point()` mutates original model variables before calling the heuristic solver, but restores values/fixed states only on the successful path. If the solver is unavailable, errors, or no point is found, the user model is left changed. For example, with a fixed initialized variable, a failed hull transform due to missing `gurobipy` left `x` changed from `value=5, fixed=True` to `value=0, fixed=False`.

Please wrap the solve/search section in `try/finally` and restore `orig_values` and `orig_fixed` on all exits, including solver exceptions and the `GDP_Error` path.

pyomo/gdp/tests/test_hull.py
```diff
@@ -2882,6 +2887,87 @@ def test_dill_pickle(self):
...
+    def test_no_nonlinear(self):
+        m = models.makeTwoTermDisj()
+        TransformationFactory('gdp.hull').apply_to(m)
+
+        m_pts = TransformationFactory('gdp.hull').get_well_defined_points_map(m)
+        # both points have zero offset
+        self.assertEqual(m_pts[m.disjunction][m.a], 0)
+        self.assertEqual(m_pts[m.disjunction][m.x], 0)
+
+    @unittest.skipUnless(gurobi_available, "Gurobi is not available")
```
bernalde, line 2916: Nonblocking: Most new domain-restriction coverage is skipped when Gurobi is unavailable, including the documented escape hatch for users who pass `well_defined_points` directly. Please add at least one non-Gurobi test that supplies a `ComponentMap` manually, for example transforming `log(m.x - 1)` with `{m.disjunction: ComponentMap([(m.x, 2)])}`.

pyomo/gdp/plugins/hull.py
```diff
@@ -197,6 +212,44 @@ class Hull_Reformulation(GDP_to_MIP_Transformation):
         """,
         ),
     )
+    CONFIG.declare(
```
bernalde, line 337: Nonblocking: The class docstring says the transformation accepts keyword arguments but still lists only `perspective_function`, `EPS`, and `targets`. Since this PR adds public options, please document `well_defined_points` and `well_defined_points_heuristic_solver` there or point readers to the generated CONFIG documentation.

bernalde (dismissed):
Please address the inline comments.

pyomo/gdp/tests/test_hull.py
emma58: I don't think this is a concern--we have Gurobi in our testing infrastructure.

pyomo/gdp/plugins/hull.py
sadavis1: Added those entries to the class doc

pyomo/gdp/tests/test_hull.py
sadavis1: It's no trouble so I went ahead and added one anyway

pyomo/gdp/plugins/hull.py
sadavis1: Fixed -- now handled in a finally block, so the original model state is restored even when we exit by throwing

pyomo/gdp/plugins/hull.py
sadavis1: Good catch. I think all that needs to be done here is to use the correct variable `disj` instead of `disjunct` (I really dislike how python keeps that variable in scope after its loop is finished...). I fixed that and added a test for generalized localvars to ensure it behaves as intended.

sadavis1:
Merged in the latest changes from main, fixing conflicts, and addressed review comments - thank you David for the review.

sadavis1:
(1 character fix; test was throwing the right exception but not for the exact reason I wanted)

pyomo/gdp/plugins/hull.py
jsiirola: I think we have (slowly) moving away from using `NumericConstant` (including `ZeroConstant`) anywhere in Pyomo.

pyomo/gdp/plugins/hull.py
```diff
@@ -405,6 +467,118 @@ def _add_transformation_block(self, to_block):
...
+    def _get_well_defined_point(
+        self, test_expr, regular_vars, fallback_vars, disj_name
+    ):
+        # First, see if test_expr is well-defined at the origin.
+        x0_map = ComponentMap()
+        orig_values = ComponentMap()
+        orig_fixed = ComponentMap()
+        for x in itertools.chain(regular_vars, fallback_vars):
+            x0_map[x] = 0  # ZeroConstant?
+            orig_values[x] = value(x, exception=False)
```
jsiirola: If these are all going to be Vars, any reason not to just use `x.value`?

pyomo/gdp/plugins/hull.py
```diff
@@ -405,6 +467,118 @@ def _add_transformation_block(self, to_block):
...
+                    EXPR.ExpressionReplacementVisitor(
+                        substitute=x0_map
+                    ).walk_expression(test_expr)
+                )
+                if math.isfinite(val):
+                    return x0_map, ComponentSet()
+            except ValueError:  # ('math domain error')
+                pass
+            except ZeroDivisionError:
+                pass
```
jsiirola: This could be
```suggestion
            except (ValueError, ZeroDivisionError):  # ('math domain error')
                pass
```

pyomo/gdp/plugins/hull.py
jsiirola: I wonder if we should just extend the `SolverFactory` so that it could be used as a domain validator?  It already handles the `str` -> solver mapping.  We would just need to augment it to pass through any solver instances that are instances of registered classes (that way the LegacySolverFactory would admit legacy solvers but not (unwrapped) `contrib.solver` solvers).

pyomo/gdp/plugins/hull.py
```diff
@@ -405,6 +467,118 @@ def _add_transformation_block(self, to_block):
...
+    # Use gurobi_direct_minlp for the heuristic for now. It needs to be
+    # a nonlinear solver.
+    def _solve_for_first_feasible_solution(self, test_model):
+        results = self._config.well_defined_points_heuristic_solver.solve(
+            test_model, load_solutions=False
+        )
+        if results.solver.termination_condition is TerminationCondition.infeasible:
+            return False
+        if results.solver.status is not SolverStatus.ok:
+            raise GDP_Error(f"Unexpected solver status {results.solver.status}.")
```
jsiirola, line 590: Would it be more consistent / future-proof to just `check_optimal_termination`?  Or maybe not, because we don't care about optimality -- but then, should we make a peer utility for `check_feasible_termination` that we can use here?

pyomo/gdp/plugins/hull.py
```diff
@@ -1595,3 +1871,142 @@ def get_transformed_constraints(self, cons):
...
+    return (base >= EPS_HEURISTIC,)
+
+
+def _handleDivisionExpression(node):
+    # No division by zero. Dividing by an NPV is always allowed.
+    arg = node.args[1]
+    if arg.__class__ in EXPR.native_types or not arg.is_potentially_variable():
+        return ()
+    # Same LP vs MIP problem as before
+    return (arg >= EPS_HEURISTIC,)
```
jsiirola, line 1950: What if the arg is allowed to be negative? (or by bounds is strictly negative?)

pyomo/gdp/plugins/hull.py
```diff
@@ -1595,3 +1871,142 @@ def get_transformed_constraints(self, cons):
...
+
+
+def _handleUnaryFunctionExpression(node):
+    arg = node.args[0]
+    if (
+        arg.__class__ in EXPR.native_types
+        or not arg.is_potentially_variable()
+        or node.name not in _unary_function_handlers
```
jsiirola: I wuold generally prefer "positive inaction" - that is, maintain a list (set) of non-restrictive unary functions, and a map of unary functions that imply domain restrictions, and then raise an exception if the node is not in either set.

jsiirola (commented):
A bunch of minor questions.  I am comfortable deferring all except for what looks like a bug in the division domain handler.

pyomo/gdp/plugins/hull.py
emma58: I like that idea!

pyomo/gdp/plugins/hull.py
emma58: I think this is a problem if arg is strictly negative. Since it's a heuristic, it's okay so long as it can be positive, but I think this would actually go infeasible if it was non-positive?

pyomo/gdp/plugins/hull.py
emma58: Whoot! Can you remove the comment then, @sadavis1 , so that I don't ask the same thing next time?

pyomo/gdp/plugins/hull.py
```diff
@@ -1595,3 +1871,142 @@ def get_transformed_constraints(self, cons):
...
+    # invoke a MIP solve here, so replace "x is nonzero" with "x is >=
+    # eps". It's a heuristic so this is not critical.
+    if exp.__class__ in EXPR.native_types or not exp.is_potentially_variable():
+        val = value(exp)
+        if round(val) == val:
+            if val >= 0:
+                return ()
+            else:
+                # return base != 0
+                return (base >= EPS_HEURISTIC,)
```
bernalde, line 1938: Blocking: This treats every negative-integer power as if the only acceptable nonzero base is positive. A model with `x` bounded in `[-2, -1]` and `x**-1` is well-defined, but the helper model gets `x >= EPS_HEURISTIC` and becomes infeasible before the transformation can find a base point. Please use a shared nonzero-domain helper that chooses `base <= -EPS_HEURISTIC` when expression bounds rule out the positive side, and add a regression test for a strictly negative base.

pyomo/gdp/plugins/hull.py
bernalde, line 1950: Blocking: This has the same sign problem for denominators. For `x` bounded in `[-2, -1]`, `1 / x` is well-defined, but the generated domain constraint is `x >= EPS_HEURISTIC`, so the heuristic reports no well-defined point. Please choose the nonzero side compatible with the denominator bounds, for example by using the negative side when the upper bound is already below zero, and test a negative-denominator case.

pyomo/gdp/plugins/hull.py
```diff
@@ -405,6 +467,118 @@ def _add_transformation_block(self, to_block):
...
+    # fixed statuses when we return successfully.
+    def _get_well_defined_point(
+        self, test_expr, regular_vars, fallback_vars, disj_name
+    ):
+        # First, see if test_expr is well-defined at the origin.
+        x0_map = ComponentMap()
+        orig_values = ComponentMap()
+        orig_fixed = ComponentMap()
+        for x in itertools.chain(regular_vars, fallback_vars):
+            x0_map[x] = 0  # ZeroConstant?
```
bernalde: Nonblocking: This looks like a stale note from the earlier `ZeroConstant` discussion. Since the code intentionally stores the numeric zero here, please remove the comment or replace it with a short rationale.

bernalde (changes requested):
Blocking issues:
- The generated domain constraints for nonzero expressions still force the argument to be positive. This makes valid strictly negative denominators and negative-integer-power bases infeasible during the well-defined-point search.

Nonblocking issues:
- There is one stale comment around the numeric zero used for the origin substitution.

Questions: none.

Tests run:
- `python -m pytest -q pyomo/gdp/tests/test_hull.py::DomainRestrictionTest pyomo/gdp/tests/test_hull.py::WellDefinedConstraintWalkerTest pyomo/gdp/tests/test_hull.py::TestGeneralizedLocalVars` -> 4 passed, 6 skipped.
- `python -m pytest -q pyomo/gdp/tests/test_hull.py` -> 151 passed, 10 skipped.
- `python -m black --check --diff pyomo/gdp/plugins/hull.py pyomo/gdp/tests/test_hull.py` -> 2 files would be left unchanged.
- Direct walker reproduction: `_WellDefinedConstraintGenerator` emits `0.0001 <= x` for both `1 / x` and `x**-1` when `x` has bounds `(-2, -1)`.

Gurobi is not available in my local environment, so the Gurobi-dependent heuristic tests were skipped locally. Hosted CI is currently green.

I would not merge this until the blocking issues above are addressed.

bernalde:
Prepared fixes for the remaining actionable review comments, but I could not push them to the PR head branch.

Commits pushed:
- `9d9c0800d Address hull domain review comments` on `bernalde:review/pr-3880-comment-fixes`

Push status:
- Pushing to `sadavis1:hull-fix` failed with `remote: Permission to sadavis1/pyomo.git denied to bernalde` / HTTP 403.
- Because the PR head was not updated, I am not replying to inline threads or marking anything resolved yet.

Main changes prepared:
- Added a shared nonzero-domain helper for the well-defined-point heuristic. It uses expression bounds when available to choose `arg <= -EPS_HEURISTIC` for strictly negative domains, covering both division denominators and negative-integer power bases.
- Removed the stale `ZeroConstant` comment, switched saved Var values to `x.value`, and combined the expected `ValueError` / `ZeroDivisionError` handling.
- Added positive-inaction handling for known unrestricted unary functions and an explicit error for unhandled unary functions.
- Added regression coverage for strictly negative nonzero domains and unary-function handling.

Tests run:
- `conda run -n pyomo-local-test python -m pytest -q pyomo/gdp/tests/test_hull.py::DomainRestrictionTest::test_nonzero_domain_strictly_negative pyomo/gdp/tests/test_hull.py::WellDefinedConstraintWalkerTest` -> 5 passed.
- `conda run -n pyomo-local-test python -m pytest -q pyomo/gdp/tests/test_hull.py::DomainRestrictionTest pyomo/gdp/tests/test_hull.py::WellDefinedConstraintWalkerTest pyomo/gdp/tests/test_hull.py::TestGeneralizedLocalVars` -> 14 passed.
- `conda run -n pyomo-local-test python -m pytest -q pyomo/gdp/tests/test_hull.py` -> 143 passed, 22 skipped.
- `conda run -n pyomo-local-test python -m black --check --diff pyomo/gdp/plugins/hull.py pyomo/gdp/tests/test_hull.py` -> 2 files would be left unchanged.
- Manual strict-negative reproduction for `1 / x` and `x**-1` with `x` bounded `(-2, -1)` now transforms successfully with `x0 = -2.0`.

Comments intentionally not changed in this commit:
- Solver validator centralization / `SolverFactory` extension: out of scope for this PR; the current option documentation already notes the V1 solver API expectation.
- New feasible-termination utility: out of scope for this PR; this would be a broader solver-results API addition.
- Threads already addressed by earlier commits, such as deterministic `ComponentSet` ordering, manual `well_defined_points` coverage, class option docs, generalized `LocalVars`, and restoration on error, were left unchanged.

Remaining risk:
- The prepared commit still needs to be applied to `sadavis1:hull-fix` before the PR reflects these fixes.

sadavis1:
This business with positive versus nonzero values is a very good point. Keeping bounds is preferable for a nonlinear solver, and I originally thought it was fine to keep them since it was unlikely to break feasibility of the subproblem (though there are somewhat contrived counterexamples). But since the constraint associated with certain functions was changed from x != 0 to x > 0, it becomes much easier to break feasibility here just by bounding the variable on the wrong side of the number line. There are a few different ways to resolve this, each with some benefits and downsides.

- I don't want to have x != 0 because that implies a big-M constraint and a MIP solve, and I don't want to do something potentially expensive in a simple heuristic unless the user is really asking for it.
- It's not really possible for me to choose ahead of time whether to strengthen to x < 0 or x > 0, since answering the question of which of those is feasible is as hard as the original problem in the general case. Of course I could use fbbt to switch to negative when the expression is strictly negative by its bounds, but that is really ad hoc
-  I could extend the bounds out symmetrically around zero (e.g. (1, 3) is converted to (-3, 3)) which eliminates some failure cases such as 1/x for -2 <= x <= 1.
- I can just remove the bounds entirely which of course eliminates any failure caused by a bound, including these. But it won't remove every failure case (e.g. 1/(-x^2) still fails here). Since I have no evidence of harm caused by unbounded variables (gurobi and baron pass the tests at least) I think this is probably the best option for now - remember that it isn't the end of the world if we fail here sometimes, since a dedicated user can still pick their base point manually.

pyomo/gdp/plugins/hull.py
sadavis1: Changed - we now check against a set for all UnaryFunctionExpression. Note that this doesn't cover AbsExpression as it's a subclass, but that's currently the only one

pyomo/gdp/plugins/hull.py
sadavis1: I think the best change here it to just relax the bounds while solving, although it does not eliminate every failure case -- see the discussion below

pyomo/gdp/plugins/hull.py
sadavis1: Made that change

pyomo/gdp/plugins/hull.py
sadavis1: Changed to using .value throughout when dealing with Vars

pyomo/gdp/plugins/hull.py
sadavis1: See the discussion below

bernalde:
I think that at least keeping track of these comments as an issue would be the right way to not have it be lost in this PR. I would commend you to try whether the problem is in fact resolved for instances that suffer from this in GDPLib (kaibel and gdp_col are among them). I have PRs in that repo that avoid the issue altogether by a hand-made transformation, but maybe with this work, it is unnecessary.

bernalde (approved):

sadavis1:
I can confirm that both those models transform successfully (with the heuristic, so no user intervention required) using this branch.

jsiirola (approved):

PR #3881, opened 2026-03-24
KKT Transform Code Refactor

ChrisLaliwala:
## Summary/Motivation:
This PR refactors the KKT transformation code (see PR #3860) to make it more efficient. In the original PR, it was suggested by @jsiirola that the order of operations be reversed to avoid repeated work and the use of unintuitive data structures.

- Refactors code to increase efficiency
- Simplifies API for getting multipliers/components from model

### Legal Acknowledgement


ChrisLaliwala:
@jsiirola

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,281 @@
...
+            raise ValueError(
+                f"model must have only one active objective; found {len(active_objs)}"
+            )
+        # collect vars from active objective
+        obj = active_objs[0]
+        for v in identify_variables(obj.expr, include_fixed=True):
+            if v.is_fixed():
+                fixed_var_set.add(v)
+            else:
+                var_set.add(v)
```
jsiirola: This is a super-minor thing, but it would be more efficient to just collect everything into var_set, e.g., with

```suggestion
        all_vars_set.update(identify_variables(obj.expr, include_fixed=True))
```

and then at the very end (after walking all expressions), split the fixed variables out.  That way, you only have to test `.fixed` once for every variable (instead of testing every time the variable is encountered in a constraint/objective).  Something like this would work well:

```python
fixed_vars = tuple(v for v in all_vars_set if v.fixed)
var_set = ComponentSet(all_vars_set)
# Note: set - other loops over set and looks values up in other; set -= other
# loops over other and looks up in set.  fixed_vars is expected to be smaller
# than var_set, so the latter is preferable.
var_set -= fixed_vars
```

TODO (in the future; not in this PR): we should consider extending the API for `identify_variables` so that we can provide the `_seen` cache.  If that is the case, then we could (efficiently) guarantee that each variable is returned only once.

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,281 @@
...
+        # add objective to lagrangean
+        lagrangean += obj.sense * obj.expr
+
+        # list of equality multipliers
+        kkt_block.gamma = VarList()
+        # list of inequality multipliers
+        kkt_block.alpha1 = VarList(domain=NonNegativeReals)
+        kkt_block.alpha2 = VarList(domain=NonNegativeReals)
```
jsiirola: Why are you splitting `alpha` into `alpha1` and `aplha2`?  I think you can just use a single `VarList` here?

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,281 @@
...
+        kkt_block.lagrangean = Expression(expr=lagrangean)
+
+        # enforce stationarity conditions
+        deriv_lagrangean = reverse_sd(kkt_block.lagrangean.expr)
+        kkt_block.stationarity_conditions = ConstraintList()
+        for var in var_set:
+            kkt_block.stationarity_conditions.add(deriv_lagrangean[var] == 0)
+
+        active_objs[0].deactivate()
+        kkt_block.dummy_obj = Objective(expr=1)
```
jsiirola: Why add a constant Objective?

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,281 @@
...
+            For ranged constraints/variables, returns (lb_mult, ub_mult),
+            where an entry is 'None' if that bound doesn't exist.
+        """
+
+        info = model.private_data()
+        if component in info.obj_dual_map:
+            return info.obj_dual_map[component]
+        else:
+            raise ValueError(
+                f"The component '{component.name}' either does not exist on '{model.name}', or is not associated with a multiplier."
```
jsiirola: Please break this string to keep lines <= 88 characters.

pyomo/core/tests/unit/test_kkt.py
```diff
@@ -0,0 +1,535 @@
...
+        self.assertIsNone(lower_bound_mult)
+        self.assertAlmostEqual(
+            value(abs(upper_bound_mult)), value(abs(m.dual[m.c4])), delta=1e-6
+        )
+
+        for v in [(m.x, m_reform.x), (m.y, m_reform.y)]:
+            primal_var, kkt_reform_var = v
+            self.assertAlmostEqual(value(primal_var), value(kkt_reform_var))
+
+    def test_kkt_solve(self):
```
jsiirola, line 83: This needs to be skipped if Ipopt isn't available.

pyomo/core/tests/unit/test_kkt.py
```diff
@@ -0,0 +1,535 @@
...
+            m.kkt.complements[3]._args[1],
+            2 * m.x - m.y - (4 + 0.5 * m.outer1) <= 0,
+        )
+
+        self.assertFalse(m.obj.active)
+
+        self.assertTrue(m.kkt.dummy_obj.active)
+        self.assertEqual(m.kkt.dummy_obj.expr, 1.0)
+
+    def test_solve_parametrized_kkt(self):
```
jsiirola, line 402: This needs to be skipped if Ipopt isn't available.

jsiirola (changes requested):
This is really close! Just a couple minor changes...

pyomo/core/plugins/transform/kkt.py
ChrisLaliwala: This is a good suggestion. The code has been updated to use a single VarList here.

pyomo/core/plugins/transform/kkt.py
ChrisLaliwala: This has been removed.

jsiirola (approved):

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,251 @@
...
+        config.set_value(kwds)
+
+        if hasattr(model, config.kkt_block_name):
+            raise ValueError(
+                "model already has an attribute with the "
+                f"specified kkt_block_name: '{config.kkt_block_name}'"
+            )
+
+        # we should check that all vars the user fixed are included
+        # in parametrize_wrt
```
emma58:
```suggestion
# We will check below that all vars the user fixed are included in
# parameterize_wrt
```

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,251 @@
...
+        lagrangean = 0
+        all_vars_set = ComponentSet()
+
+        # collect the active Objectives
+        active_objs = list(
+            model.component_data_objects(Objective, active=True, descend_into=True)
+        )
+        if len(active_objs) != 1:
+            raise ValueError(
+                f"model must have only one active objective; found {len(active_objs)}"
```
emma58: Minor nit, but I'd say "exactly one" since a user can hit this error for having no objectives as well.

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,251 @@
...
+        'kkt_block_name',
+        ConfigValue(
+            default='kkt',
+            doc="""
+           Name of the block on which the kkt variables and constraints will be stored. 
+           """,
+        ),
+    )
+    CONFIG.declare(
+        'parametrize_wrt',
```
emma58: I think your spelling is also correct, but for consistency with other parts of Pyomo, can we change this to `parameterize_wrt`?

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,251 @@
...
+                info.obj_dual_map[con] = (alpha_l, alpha_u)
+
+        fixed_vars = ComponentSet(v for v in all_vars_set if v.is_fixed())
+        var_set = ComponentSet(all_vars_set)
+        var_set -= fixed_vars
+
+        # do error checking on parametrize_wrt
+        missing = fixed_vars - params
+        if missing:
+            raise ValueError("All fixed variables must be included in parametrize_wrt.")
```
emma58: Can you also print the names of the Vars in `missing` in this error message?

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,251 @@
...
+
+        if not params <= all_vars_set:
+            raise ValueError(
+                "A variable passed in parametrize_wrt does not exist on an "
```
emma58:
```suggestion
                "A variable passed in parametrize_wrt does not exist in an "
```

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,251 @@
...
+
+        # enforce stationarity conditions
+        deriv_lagrangean = reverse_sd(kkt_block.lagrangean.expr)
+        kkt_block.stationarity_conditions = ConstraintList()
+        for var in var_set:
+            kkt_block.stationarity_conditions.add(deriv_lagrangean[var] == 0)
+
+        active_objs[0].deactivate()
+
+    def get_object_from_multiplier(self, model, multiplier_var):
```
emma58, line 203: I would prefer this be named `get_constraint_from_multiplier` even though it could return a Var if it is a multiplier on bounds. But I'm open to being argued with.

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,251 @@
...
+        """
+
+        info = model.private_data()
+        if multiplier_var in info.dual_obj_map:
+            return info.dual_obj_map[multiplier_var]
+        raise ValueError(
+            f"The KKT multiplier: {multiplier_var.name}, does not exist on {model.name}."
+        )
+
+    def get_multiplier_from_object(self, model, component):
```
emma58, line 230: Same naming thought as above

pyomo/core/plugins/transform/kkt.py
```diff
@@ -0,0 +1,251 @@
...
+        """
+        Return the multiplier for the object. If the object is a normal constraint, a single
+        multiplier is returned. If the object is a ranged constraint or a variable, a tuple
+        containing the lower and upper bound multipliers is returned.
+
+        Parameters
+        ----------
+        model: ConcreteModel
+            The model on which the kkt transformation was applied to
```
emma58:
```suggestion
            The model to which the kkt transformation was applied
```

emma58 (approved):
Apologies for the much-delayed review. This looks really good! I have a few minor nits, and then one (perhaps more controversial) naming question.

pyomo/core/plugins/transform/kkt.py
jsiirola: I have to admit that I would prefer `object` (for consistency with `component_data_objects`) or `component` (for the same reason).

pyomo/core/plugins/transform/kkt.py
ChrisLaliwala: I've added the following (and updated the tests) to print missing and extra vars (lines 158-171 in newest commit):

        # do error checking on parameterize_wrt
        missing = fixed_vars - params
        if missing:
            raise ValueError(
                "All fixed variables must be included in parameterize_wrt. "
                "Missing variables:\n\t" + "\n\t".join(v.name for v in missing)
            )

        extra = params - all_vars_set
        if extra:
            raise ValueError(
                "A variable passed in parameterize_wrt does not exist in an "
                "active constraint or objective within the model. "
                "Invalid variables:\n\t" + "\n\t".join(v.name for v in extra)
            )

pyomo/core/plugins/transform/kkt.py
emma58: I was afraid you would say that! :P OK, let's leave it then.

PR #3892, opened 2026-03-26
Use Hessian approximation in CyIpopt when grey-box without Hessian is present

Robbybp:
- Set (potentially override) `hessian_approximation=limited-memory` in CyIpopt solver when the model contains a grey-box that doesn't support Hessians
- Log a warning if this happens and the user had set `hessian_approximation=exact`
- Add a `has_hessian_support` method to `PyomoNLPWithGreyBoxBlocks` (previously only existed on `_ExternalGreyBoxAsNLP`, although the former has a private `_has_hessian_support` attribute)

### Legal Acknowledgement


blnicho (approved):

jsiirola (approved):
Looks good.  

One design question, though: should we promote `has_hessian_support` up to PyomoNLP?  (Maybe this should be added to a future "Let's revisit the design of Pynumero and ExternalGreyBox"?)

pyomo/contrib/pynumero/algorithms/solvers/cyipopt_solver.py
```diff
@@ -371,6 +372,20 @@ def solve(self, model, **kwds):
                 set_scaling = cyipopt_solver.setProblemScaling
             set_scaling(obj_scaling, x_scaling, g_scaling)
 
+        # has_hessian_support only exists for PyomoNLPWithGreyBoxBlocks, but if
+        # we have grey box blocks, we know our NLP is of this class.
+        if grey_box_blocks and not nlp.has_hessian_support():
+            # Note that `config` is a copy of the instance-level self.config so
+            # we don't have to reset the option after the solve.
+            if config.options.get("hessian_approximation", None) == "exact":
+                logger.warning(
```
jsiirola, line 381: Can you add a test exercising / verifying this warning?

Robbybp:
> should we promote `has_hessian_support` up to PyomoNLP?

For now, no. PyomoNLP always has Hessian support. However, if we eventually merge grey box support into PyomoNLP (to eliminate the need for PyomoNLPWithGreyBoxBlocks), then we will need to do this.

jsiirola:
> > should we promote `has_hessian_support` up to PyomoNLP?
> 
> For now, no. PyomoNLP always has Hessian support. However, if we eventually merge grey box support into PyomoNLP (to eliminate the need for PyomoNLPWithGreyBoxBlocks), then we will need to do this.

I was mostly thinking about an implementation like
```python
def has_hessian_support(self):
    return True
```
and including it so that we had a "standard" interface API.

Robbybp:
I see. I still would rather not add this before we need it.

pyomo/contrib/pynumero/algorithms/solvers/cyipopt_solver.py
```diff
@@ -371,6 +372,20 @@ def solve(self, model, **kwds):
                 set_scaling = cyipopt_solver.setProblemScaling
             set_scaling(obj_scaling, x_scaling, g_scaling)
 
+        # has_hessian_support only exists for PyomoNLPWithGreyBoxBlocks, but if
+        # we have grey box blocks, we know our NLP is of this class.
+        if grey_box_blocks and not nlp.has_hessian_support():
+            # Note that `config` is a copy of the instance-level self.config so
+            # we don't have to reset the option after the solve.
+            if config.options.get("hessian_approximation", None) == "exact":
```
jsiirola, line 380: `ConfigDict.get()` returns the `ConfigValue`/`ConfigList`/`ConfigDict` and not the value, so as written this test will *always* be False.  Suggested change:

```suggestion
            hessian_approx = config.options.get("hessian_approximation", None)
            if hessian_approx and hessian_approx.value() == "exact":
```

pyomo/contrib/pynumero/algorithms/solvers/tests/test_cyipopt_solver.py
```diff
@@ -414,3 +419,67 @@ def intermediate(
...
+        m.reactor = ExternalGreyBoxBlock(
+            external_model=ReactorConcentrationsOutputModel()
+        )
+        m.k1con = pyo.Constraint(expr=m.reactor.inputs["k1"] == 5 / 6)
+        m.k2con = pyo.Constraint(expr=m.reactor.inputs["k2"] == 5 / 3)
+        m.k3con = pyo.Constraint(expr=m.reactor.inputs["k3"] == 1 / 6000)
+        m.cafcon = pyo.Constraint(expr=m.reactor.inputs["caf"] == 10000)
+        m.obj = pyo.Objective(expr=m.reactor.outputs["cb"], sense=pyo.maximize)
+        solver = pyo.SolverFactory("cyipopt")
+        solver.config.options["hessian_approximation"] = "exact"
```
jsiirola, line 460: Please wrap this in a LoggingIntercept and verify that the expected log message is being emitted.

jsiirola (changes requested):
I couldn't figure out why you had an uncovered line... but it was because there is a bug in the logic.

Robbybp:
> I couldn't figure out why you had an uncovered line... but it was because there is a bug in the logic.

Good catch, thanks. I think this should be ready once checks pass.

jsiirola (approved):

Issue #24, opened 2016-08-20
Add SMPS conversion to the 'convert' subcommand


whart222:
Converting SP problems is only supported with a separate subcommand, but it should be integrated into the 'pyomo convert' subcommand.

PR #3893, opened 2026-03-26
Improved `ConstraintData.equality`, relational expression generation

### LINKED ISSUE #2093: Constraint's equality attribute sometimes fails to properly identify equality constraints

jsiirola:
## Summary/Motivation:

The original motivation of this PR was to resolve #2093.  The Issue stemmed from the `ConstraintData.equality` property sometimes returning False for things that are obviously equality constraints due to the use of `is` for comparing the lower and upper bounds.  This PR makes a modification of the property so that `native_types` in the lower/upper bounds are compared using `==` and everything else is compared with `is`.  This is related to the (abandoned) PR #2097), where we used `is_constant()` to determine if we should use `==` or `is`.  The difference here is the switch to `native_types` - both for efficiency and clarity of the intent.

As part of working through this, we came across inconsistencies in when constant expressions were resolved: the move of the relational expression generation logic to a multiple dispatch model caused constant relational expressions generated from operator overloading to be resolved to `bool`.  This behavior was *not* promulgated to the `inequality()` function, or to the specification of constraints using "tuple notation".  This PR overhauls `inequality()` and the tuple processing to be consistent with the relational expression logic.  It also removed redundant logic, and funnels all expression generation through the relational expression dispatchers.

### Update [13 Apr]

Resolved a number of additional edge cases, mostly related to resolving trivial portions of ranged constraints, including 
- `inequality(0, 0, None)` should be `True` (and not 0), and 
- Ranged constraints should check that trivial bounds are feasible (e.g., `inequality(1, m.x, 0)` should be `False`).

As part of this update, I added about 2000 lines of relational expression tests.  Unfortunately, GitHub gets confused when trying to render the diff.  The actual changes are adding 3 arguments (`native2` (a larger native constant) and `le2` / `lt2` which are the same as `le` / `lt`, but with the argument order reversed).  This meant adding 3 tests to the end of each test method, plus adding 3 new test methods (`test_XX_native2`, `test_XX_le2`, and `test_XX_lt2`) for each of the 3 test classes.

Buried in this are a handful of baseline changes, all due to the expression system being more thorough and consistent with resolving constant portions of expressions when generating expressions.

- use `==` for comparing constraint bounds for `native_types` in `ConstraintData.equality`
- rework `inequality()` to build on relational expression dispatcher
- improve implementation of `tuple_to_relational_expr` to make it consistent with constraint behavior (and build on `inequality()`)
- simplify `ConstraintData.set_value()` to call `tuple_to_relational_expr`
- add workarounds for unbounded constraints in the `nonnegative_transform` and `Complementarity`
- update tests

### Legal Acknowledgement

### DIFF
--- a/pyomo/core/tests/unit/test_relational_expr_dispatcher.py
+++ b/pyomo/core/tests/unit/test_relational_expr_dispatcher.py
@@ -27,7 +27,7 @@ logger = logging.getLogger(__name__)
 
 
 class BaseRelational(Base):
-    NUM_TESTS = 26
+    NUM_TESTS = 28
 
     def tearDown(self):
         pass
@@ -66,6 +66,8 @@ class BaseRelational(Base):
         self.eq = self.m.x == self.m.q
         self.le = self.m.x <= self.m.q
         self.lt = self.m.x < self.m.p
+        self.le2 = self.m.q <= self.m.x
+        self.lt2 = self.m.p < self.m.x
         self.ranged = inequality(self.m.p, self.m.x, self.m.q)
 
         # self.TEMPLATE = [
@@ -105,6 +107,8 @@ class BaseRelational(Base):
                 # 24:
                 self.ranged,
                 self.native2,
+                self.le2,
+                self.lt2,
             ]
         )
 
@@ -155,6 +159,8 @@ class TestEquality(BaseRelational, unittest.TestCase):
             # 24:
             (self.invalid, self.ranged, False),
             (self.invalid, self.native2, False),
+            (self.invalid, self.le2, False),
+            (self.invalid, self.lt2, False),
         ]
         self._run_cases(tests, operator.eq)
 
@@ -224,6 +230,18 @@ class TestEquality(BaseRelational, unittest.TestCase):
                 "sub-expressions is a relational expression",
             ),
             (self.asbinary, self.native2, EqualityExpression((self.bin, 8))),
+            (
+                self.asbinary,
+                self.le2,
+                "Cannot create an EqualityExpression where one of the "
+                "sub-expressions is a relational expression",
+            ),
+            (
+                self.asbinary,
+                self.lt2,
+                "Cannot create an EqualityExpression where one of the "
+                "sub-expressions is a relational expression",
+            ),
         ]
         self._run_cases(tests, operator.eq)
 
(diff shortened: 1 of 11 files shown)


blnicho (approved):

pyomo/core/tests/unit/test_relational_expr_dispatcher.py
```diff
@@ -2191,2556 +2515,3595 @@ def test_eq_ranged(self):
...
+                self.native,
+                self.le,
+                RangedExpression((5,) + self.le.args, (False, False)),
+            ),
+            (
+                self.native,
+                self.lt,
+                RangedExpression((5,) + self.lt.args, (False, True)),
+            ),
+            # 24
```
mrmundt: This is going to be the world's stupidest nitpick and does NOT need to be addressed, but I am me, so I'm gonna say it. `#24` is the only one that doesn't have a trailing colon.

mrmundt (approved):

PR #3898, opened 2026-04-01
Add core symbolic gradient support to Pyomo.DoE

snarasi2:
## Fixes # .
Summary/Motivation:
This PR ports the core symbolic-gradient functionality from the historical pyomo-doe-symbolic work into the current pyomo.contrib.doe implementation.

## Summary/Motivation:

This PR ports the core symbolic-gradient functionality from the historical `pyomo-doe-symbolic` work into the current `pyomo.contrib.doe` implementation.

Rather than merging the old branch directly, this change transplants the symbolic DoE pieces onto current `main` so that symbolic sensitivities work with the newer DoE implementation already present in Pyomo, including the current objective and GreyBox-oriented code paths.

This PR is intentionally scoped to the core symbolic-gradient integration. It does not attempt to port every example, test refactor, or auxiliary change from the historical `adowling2/pyomo-doe-symbolic` branch in one step.

This contribution was prepared with coding assistance from OpenAI Codex. All design decisions, validation, testing, and quality-assurance responsibility remain with Shilpa Narasimhan.

Local validation included:
- `pyomo/contrib/doe/tests`: `133 passed, 4 skipped, 5 warnings, 10 subtests passed`
- `pyomo/contrib/doe/tests/test_greybox.py`: `33 passed` in HSL-enabled local environments

The remaining public-CI GreyBox `cyipopt` skips are environment-dependent and tied to MA57/HSL runtime availability.

- add `GradientMethod` support to `DesignOfExperiments`
- preserve the existing finite-difference workflow through the new gradient-method interface
- add a symbolic / `pynumero` gradient path for DoE
- add an analytic FIM computation path for non-optimization FIM evaluation
- export `ExperimentGradients` from `pyomo.contrib.doe`
- refactor `ExperimentGradients` so symbolic and automatic differentiation are set up together
- add supporting symbolic-gradient utilities for labeled experiment models
- update DoE model construction so symbolic sensitivities can be used in the Jacobian-based DoE machinery
- add a guard preventing `run_doe()` from being called with `GradientMethod.kaug`
- add the `polynomial` example and polynomial-focused regression coverage
- add broader symbolic-versus-automatic gradient consistency tests
- add factorial-result dataframe tests
- add reactor regression tests against expected solutions
- clarify the GreyBox `cyipopt` test skip reason when the MA57/HSL runtime is unavailable

### Legal Acknowledgement


adowling2:
This is a port of https://github.com/adowling2/pyomo/pull/7

snarasi2:
Outstanding item from the historical symbolic branch still remain to be completed:
- updating documentation

Closed out:

- refactoring `ExperimentGradients` so symbolic and automatic differentiation are set up together
- adding tests for the polynomial example
- checking symbolic and automatic gradient correctness more broadly
- testing the data frame functions
- adding reactor regression tests against expected solutions
- checking code coverage
- removing diagnostic print statements
- removing stale or outdated comments
- running remaining lint and spellcheck steps

snarasi2:
@sscini @smondal13 @slilonfe5 ready for preliminary review. Could you please take a look when you are free?

pyomo/contrib/doe/tests/test_utils.py
```diff
@@ -18,6 +32,72 @@
...
+file_path = os.path.join(currdir, "..", "examples", "result.json")
+
+with open(file_path) as f:
+    data_ex = json.load(f)
+
+data_ex["control_points"] = {float(k): v for k, v in data_ex["control_points"].items()}
+
+ipopt_available = SolverFactory("ipopt").available()
+
+
```
sscini, line 39: It looks like this is the same model from the new example polynomial.py. It is usually preferred to import from the example file instead of copying the code here. Please correct me if this is unique to the testing file.

pyomo/contrib/doe/tests/test_utils.py
```diff
@@ -154,5 +234,260 @@ def test_get_FIM_metrics(self):
         )
```
sscini: This testing class also uses ipopt, so you can add the flag to skipIf ipopt is not available.

pyomo/contrib/doe/utils.py
```diff
@@ -266,3 +269,166 @@ def get_FIM_metrics(FIM):
         "log10(E-Optimality)": E_opt,
         "log10(Modified E-Optimality)": ME_opt,
     }
+
+
+class ExperimentGradients:
```
sscini, line 292: To assist reviewers and future developers, it would be helpful to add more comments in the functions to explain what math steps are being taken.

pyomo/contrib/doe/doe.py
```diff
@@ -1036,50 +1163,79 @@ def initialize_fim_inv(m, j, d):
...
+            experiment_grad.construct_sensitivity_constraints(model.scenario_blocks[0])
 
-            if self.scale_nominal_param_value:
-                return (
-                    m.sensitivity_jacobian[n, p]
-                    == (var_up - var_lo)
-                    / param_diff
-                    * param_val
-                    * self.scale_constant_value
+            def jacobian_rule(m, n, p):
```
sscini: I am slightly confused in this portion of the code. I think this could be restructured to avoid defining different functions inside of the if else statement. It might be preferred to pass the gradient method as an argument for the jacobian rule, and define the if else statement inside. More comments in this section would also be appreciated to explain decisions.

pyomo/contrib/doe/doe.py
```diff
@@ -1170,6 +1326,13 @@ def _generate_scenario_blocks(self, model=None):
         if model is None:
             model = self.model
 
+        fd_formula = (
```
sscini, line 1333: It looks like previously fd formula was stored in the class. Why remove it from the class (self.fd_formula -> fd_formula)?

pyomo/contrib/doe/doe.py
```diff
@@ -1274,13 +1442,13 @@ def build_block_scenarios(b, s):
...
                 diff = self.step * (
                     (-1) ** s
                 )  # Positive perturbation, even; negative, odd
-            elif self.fd_formula == FiniteDifferenceStep.backward:
+            elif fd_formula == FiniteDifferenceStep.backward:
                 diff = self.step * -1  # Backward always negative perturbation
-            elif self.fd_formula == FiniteDifferenceStep.forward:
+            elif fd_formula == FiniteDifferenceStep.forward:
                 diff = self.step  # Forward always positive
             else:
```
sscini, line 1457: Does this error message still need to be added?

.github/workflows/test_branches.yml
```diff
@@ -950,7 +950,9 @@ jobs:
         fi
 
     - name: Upload codecov reports
```
sscini, line 952: I am not used to seeing these files get modified or included in PRs? (In .github/workflows folder). Please explain these changes if they are needed or remove if possible.

pyomo/contrib/doe/examples/polynomial.py
```diff
@@ -0,0 +1,101 @@
...
+        m.b = pyo.Var(bounds=(-5, 5), initialize=-1)
+        m.c = pyo.Var(bounds=(-5, 5), initialize=0.5)
+        m.d = pyo.Var(bounds=(-5, 5), initialize=-1)
+
+        m.y = pyo.Var(initialize=0)
+
+        @m.Constraint()
+        def output_equation(m):
+            return m.y == m.a * m.x1 + m.b * m.x2 + m.c * m.x1 * m.x2 + m.d
+
```
sscini, line 50: If there is no finalize model, you can leave the function out of the model, and then remove from line 27.

pyomo/contrib/doe/tests/test_greybox.py
```diff
@@ -373,11 +375,17 @@ def make_greybox_and_doe_objects_rooney_biegler(objective_option):
...
-        )
-    except:
+        termination_message = str(doe_object.results.get("Termination Message", ""))
+        cyipopt_call_working = bad_message not in termination_message
+        if not cyipopt_call_working:
+            cyipopt_skip_reason = (
+                "cyipopt GreyBox solve path cannot access the MA57/HSL runtime"
+            )
+    except Exception:
         cyipopt_call_working = False
```
sscini, line 355: This is nitpicky but why specifically MA57/HSL? I think old error message was more general. Please document why this specific solver if it is important to function.

sscini (commented):
Nice work. I think this is close, but more comments throughout would be appreciated to go over mathematical reasoning for certain decision areas, and help reviewer/future developers understand steps and flow. 

@slilonfe5 @smondal13 Is it normal to change the github workflow files? I have never done that before but that not sure if you have.

pyomo/contrib/doe/examples/polynomial.py
```diff
@@ -0,0 +1,101 @@
...
+import pyomo.environ as pyo
+
+from pyomo.contrib.doe import DesignOfExperiments
+from pyomo.contrib.parmest.experiment import Experiment
+
+
+class PolynomialExperiment(Experiment):
+    """A small algebraic experiment used for symbolic-gradient testing."""
+
+    def __init__(self, data=None):
```
smondal13: The `data` argument is not used anywhere, so why are you passing that argument? If the user can pass data, then you should clearly mention how to pass data and what structure the data should have.

pyomo/contrib/doe/tests/test_doe_build.py
```diff
@@ -25,6 +25,7 @@
 
 if scipy_available:
     from pyomo.contrib.doe import DesignOfExperiments
+    from pyomo.contrib.doe.examples.polynomial import PolynomialExperiment
```
smondal13: Do we need scipy for the`PolynomialExperiment`? If not, then you should not make a general import, not a conditional one.

pyomo/contrib/doe/tests/test_doe_errors.py
```diff
@@ -503,6 +503,86 @@ def test_reactor_figure_drawing_bad_sens_names(self):
                 fixed_design_variables={"hour": 1},
             )
 
+    @unittest.skipIf(not ipopt_available, "The 'ipopt' command is not available")
+    def test_reactor_figure_drawing_more_than_two_sens_vars(self):
```
smondal13: Does it test `symbolic gradient` functionality, or is it a different test?

pyomo/contrib/doe/tests/test_doe_errors.py
```diff
@@ -503,6 +503,86 @@ def test_reactor_figure_drawing_bad_sens_names(self):
...
+        ):
+            doe_obj.draw_factorial_figure(
+                results=synthetic_results,
+                sensitivity_design_variables=["hour", "temperature", "pressure"],
+                fixed_design_variables={},
+                full_design_variable_names=["hour", "temperature", "pressure"],
+            )
+
+    @unittest.skipIf(not ipopt_available, "The 'ipopt' command is not available")
+    def test_reactor_figure_drawing_requires_all_other_design_vars_fixed(self):
```
smondal13: Same as above.

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -457,6 +512,146 @@ def test_reactor_grid_search(self):
...
+        doe_obj = DesignOfExperiments(**DoE_args)
+        doe_obj.run_doe()
+
+        self.assertEqual(doe_obj.results["Solver Status"], "ok")
+
+        FIM, Q, L, sigma_inv = get_FIM_Q_L(doe_obj=doe_obj)
+        self.assertTrue(np.all(np.isclose(FIM, L @ L.T)))
+        self.assertTrue(np.all(np.isclose(FIM, Q.T @ sigma_inv @ Q)))
+
+    def test_reactor_run_doe_determinant_regression(self):
```
smondal13: We wanted to avoid the reactor experiment and use Rooney-Biegler as much as possible. Can this test not be replaced by RB?

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -457,6 +512,146 @@ def test_reactor_grid_search(self):
...
+        )
+        self.assertAlmostEqual(
+            doe_obj.results["log10 A-opt"], -2.936916407094173, places=4
+        )
+
+        fim = np.array(doe_obj.results["FIM"])
+        self.assertAlmostEqual(fim[0, 0], 173179.08904385, places=1)
+        self.assertAlmostEqual(fim[3, 3], 8218128.34162048, places=1)
+
+    def test_reactor_run_doe_gradient_matrix(self):
```
smondal13: Same here.

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -457,6 +512,146 @@ def test_reactor_grid_search(self):
...
+                self.assertTrue(np.all(np.isclose(FIM, Q.T @ sigma_inv @ Q)))
+
+    def test_polynomial_example_compute_fim_pynumero(self):
+        """Check that the transplanted polynomial example computes the expected FIM."""
+        fim = run_polynomial_doe()
+        expected = get_expected_polynomial_fim()
+        self.assertEqual(fim.shape, expected.shape)
+        self.assertTrue(np.allclose(fim, expected))
+
+    def test_polynomial_example_compute_fim_central_smoke(self):
```
smondal13: In `DesignOfExperiments`, we have `_only_compute_fim_lower=True`, so it will always be symmetric. I think we have a test that ensures that. If we don't, then this test is okay. Otherwise, this test is probably redundant.

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -457,6 +512,146 @@ def test_reactor_grid_search(self):
...
+        """Check that the finite-difference polynomial path returns a symmetric FIM."""
+        doe_obj = DesignOfExperiments(**get_polynomial_args())
+        fim = doe_obj.compute_FIM()
+
+        self.assertEqual(fim.shape, (4, 4))
+        self.assertTrue(np.all(np.isfinite(fim)))
+        self.assertTrue(np.allclose(fim, fim.T))
+
+    def test_polynomial_example_measurement_error_scaling(self):
```
smondal13, line 596: Why do we need this test? Are you not doing something similar in `test_polynomial_example_compute_fim_pynumero()`?

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -457,6 +512,146 @@ def test_reactor_grid_search(self):
...
+            **get_polynomial_args(
+                gradient_method="pynumero",
+                prior_FIM=prior_FIM,
+                objective_option="trace",
+            )
+        )
+
+        doe_obj.run_doe()
+
+        self.assertEqual(doe_obj.results["Solver Status"], "ok")
```
smondal13, line 638: You need to test it against the optimal value here as well. If there is a bug in the solver in the future, we can test it that way.

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -789,6 +984,91 @@ def cleanup_file():
                 f"Expected plot file '{expected_d_plot}' was not created.",
             )
 
+    @unittest.skipUnless(pandas_available, "test requires pandas")
+    @unittest.skipUnless(ipopt_available, "test requires ipopt")
+    def test_rooney_biegler_factorial_results_dataframe_schema(self):
```
smondal13, line 981: I am quite confused. Why are you adding these tests? Are they related to `symbolic gradient`? I see that you are using the "central" finite difference here. **If you are testing something else, you should open a different PR**

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -789,6 +984,91 @@ def cleanup_file():
...
+        }
+
+        self.assertTrue(expected_columns.issubset(results_pd.columns))
+        self.assertEqual(len(results_pd), 3)
+        self.assertEqual(sorted(results_pd["hour"].tolist()), [0.0, 5.0, 10.0])
+        self.assertTrue(np.all(np.isfinite(results_pd["solve_time"])))
+
+    @unittest.skipUnless(pandas_available, "test requires pandas")
+    @unittest.skipUnless(ipopt_available, "test requires ipopt")
+    def test_draw_factorial_figure_accepts_dataframe_input(self):
```
smondal13, line 1013: Same here.

pyomo/contrib/doe/tests/test_utils.py
smondal13: I agree with @sscini. Since you have tested the `polynomial.py`, you should import that instead of copying and pasting.

pyomo/contrib/doe/tests/test_utils.py
```diff
@@ -154,5 +234,260 @@ def test_get_FIM_metrics(self):
...
+    def test_polynomial_automatic_only_still_sets_both_jacobians(self):
+        """Check that both Jacobian maps are prepared in the unified setup path."""
+        experiment = PolynomialExperiment()
+        model = experiment.get_labeled_model()
+
+        experiment_gradients = ExperimentGradients(
+            model, symbolic=False, automatic=True
+        )
+
+        self.assertIsNotNone(experiment_gradients.jac_dict_sd)
```
smondal13, line 286: This is a weak test. I'd recommend using the values (numpy array/list) for this line and the one below.

pyomo/contrib/doe/tests/test_utils.py
```diff
@@ -154,5 +234,260 @@ def test_get_FIM_metrics(self):
...
+        self.assertIsNotNone(experiment_gradients.jac_dict_ad)
+
+    def test_polynomial_metric_helpers_match_numpy(self):
```
smondal13: I am not quite sure whether this test will add any value. It seems like you are testing `compute_FIM_metrics()`, but that is not your purpose since that is an already existing function.

pyomo/contrib/doe/tests/test_utils.py
```diff
@@ -154,5 +234,260 @@ def test_get_FIM_metrics(self):
...
+    def test_polynomial_symbolic_only_still_sets_both_jacobians(self):
+        """Check that symbolic-only requests still initialize both Jacobian maps."""
+        experiment = PolynomialExperiment()
+        model = experiment.get_labeled_model()
+
+        experiment_gradients = ExperimentGradients(
+            model, symbolic=True, automatic=False
+        )
+
+        self.assertIsNotNone(experiment_gradients.jac_dict_sd)
```
smondal13, line 306: Need to test against values.

pyomo/contrib/doe/tests/test_utils.py
```diff
@@ -154,5 +234,260 @@ def test_get_FIM_metrics(self):
...
+    def test_reactor_symbolic_and_automatic_jacobians_agree(self):
+        """Check reactor Jacobians from symbolic and automatic differentiation."""
+        experiment = self._get_reactor_experiment()
+        model = self._initialize_reactor_model(experiment.get_labeled_model())
+
+        experiment_gradients = self._assert_symbolic_and_automatic_jacobians_agree(
+            model, atol=1e-6, rtol=1e-6
+        )
+
+        self.assertGreater(len(experiment_gradients.jac_dict_sd), 0)
```
smondal13: Instead of using `assertGreater`, why not use `assertEqual`? If it is an array/nested list, you can use shape (n_outputs, n_params)

smondal13:
@snarasi2, I think it would be helpful for the reviewers if you add a markdown file explaining the PR and the math.

pyomo/contrib/doe/examples/polynomial.py
snarasi2: I also removed the unused data argument from the example constructor and updated the remaining call sites (d708d5522).

pyomo/contrib/doe/tests/test_utils.py
snarasi2: Thank you, I updated the tests to import PolynomialExperiment from pyomo.contrib.doe.examples.polynomial and removed the duplicated local definition (80552f209).

pyomo/contrib/doe/utils.py
snarasi2: Thank you, I added comments and docstrings in utils.py to clarify the main mathematical steps in the symbolic sensitivity construction and FIM-related calculations, using the same notation as in the paper (f53fe17f4).

pyomo/contrib/doe/examples/polynomial.py
snarasi2: Thank you, agreed. Since finalize_model() was not used within this example, I removed the method and the corresponding call in get_labeled_model() (11d475ef9).

pyomo/contrib/doe/tests/test_doe_build.py
snarasi2: Thank you. PolynomialExperiment does not require SciPy. Made the import non-conditional (e704f686a).

pyomo/contrib/doe/tests/test_doe_errors.py
snarasi2: The test is not intended to validate the symbolic gradient functionality, but I left it in because it covers the plotting/input-validation guard in draw_factorial_figure(). I added for coverage

pyomo/contrib/doe/tests/test_doe_errors.py
snarasi2: The test is also not intended to validate the symbolic gradient functionality. I added it as an additional coverage test.

pyomo/contrib/doe/tests/test_doe_solve.py
snarasi2: @adowling2 Should I replace the reactor experiment with RB?

pyomo/contrib/doe/tests/test_doe_solve.py
snarasi2: @adowling2 Should I replace the reactor experiment with RB?

pyomo/contrib/doe/tests/test_doe_solve.py
snarasi2: Thank you for pointing that out. I agree. Removed in commit 0d0a14ff5

pyomo/contrib/doe/tests/test_doe_solve.py
snarasi2: This test checks a different property. test_polynomial_example_compute_fim_pynumero() verifies that the baseline FIM is correct for the polynomial example, while this test checks that the measurement-error scaling in fim_rule in doe.py is applied correctly, i.e., that the FIM scales with 1/\sigma^2

pyomo/contrib/doe/tests/test_doe_solve.py
snarasi2: Thank you, I strengthened this from a smoke test to a regression test that checks for optimal values in addition to solver status. Change made in commit 401c03a3f

pyomo/contrib/doe/tests/test_doe_solve.py
snarasi2: Thank you. I updated the PR-added DoE tests so that the symbolic-path cases explicitly select the symbolic gradient backend (pynumero/kaug) rather than relying on the default finite-difference backend, while keeping fd_method="central"  (65571dad0).

pyomo/contrib/doe/tests/test_doe_solve.py
snarasi2: Thank you. I updated the PR-added DoE tests so that the symbolic-path cases explicitly select the symbolic gradient backend (pynumero/kaug) rather than relying on the default finite-difference backend, while keeping fd_method="central" as the shared setup convention (65571dad0).

pyomo/contrib/doe/tests/test_utils.py
snarasi2: Thank you, good point. I added the ipopt availability guard to the test_utils.py testing classes that use IPOPT so they skip cleanly when the solver is unavailable (9eb49593e).

pyomo/contrib/doe/tests/test_utils.py
snarasi2: Thank you, agreed. I strengthened the automatic-only and symbolic-only Jacobian tests to check the actual polynomial sensitivity values returned by compute_gradient_outputs_wrt_unknown_parameters() rather than only checking that the Jacobian containers were populated (702dc8990).

pyomo/contrib/doe/tests/test_utils.py
snarasi2: That’s fair. This test was exercising the pre-existing FIM metric helper utilities rather than the symbolic-gradient functionality itself, so I removed it from this PR to keep the scope focused (d652ceaeb).

pyomo/contrib/doe/tests/test_utils.py
snarasi2: Thank you, agreed. I strengthened this test to check the actual polynomial sensitivity values returned by compute_gradient_outputs_wrt_unknown_parameters() rather than only checking that the Jacobian containers were populated (702dc8990).

pyomo/contrib/doe/tests/test_utils.py
snarasi2: Thank you, agreed. I replaced the weak nonempty Jacobian assertions with stronger checks. The element-wise symbolic-versus-automatic Jacobian comparison is already done in _assert_symbolic_and_automatic_jacobians_agree() in test_utils.py, and in this update I also replaced the leftover assertGreater(...) checks with exact expected Jacobian-entry counts (b5ea7f18c).

snarasi2:
@smondal13 @slilonfe5 @sscini  For workflow modifications: 
I removed the workflow-related changes from this PR as requested, but that appears to have re-exposed CI issues in the workflow/config layer rather than in the symbolic DoE code itself. In particular, the style-and-typos failure is likely due to removing the writeable allowlist entry from .github/workflows/typos.toml, and the process-coverage-* failures are likely due to reverting the added hashFiles('coverage.xml') != '' guard around the Codecov upload step. The symbolic DoE test changes themselves are still passing locally. Since the guidance was not to modify workflow files in this PR, I’ve left those workflow/config fixes out for now, but I’m happy to handle them separately if needed.

pyomo/contrib/doe/tests/test_doe_solve.py
snarasi2: Thank you. I treated this as a broader preference to avoid the reactor example unless it was specifically needed, and addressed it in three cleanup batches:

c593f0bf1: replaced the general-purpose run_doe / regression coverage in test_doe_solve.py with Rooney-Biegler, converted the generic 2D grid-search coverage to the polynomial example, and removed the redundant plain reactor Jacobian-agreement test.
36d569485: converted the perturbed-point symbolic/automatic Jacobian-agreement test in test_utils.py to Rooney-Biegler, removed the now-unused reactor helper machinery from test_utils.py, and cleaned up the remaining stale bad-model comments.
dae72f4a2: finished the remaining reactor cleanup in the touched DoE test files by replacing the remaining reactor-based error/build/solve coverage with Rooney-Biegler, using the polynomial example for the remaining generic 2D factorial cases, and rebasing the non-solve GreyBox helper/tests onto the Rooney-Biegler path with the corresponding 2-parameter FIM expectations.
After these updates, there are no remaining reactor references in the touched DoE test files, and the focused DoE bundle passes locally:

130 passed, 4 skipped, 5 warnings in 35.67s

pyomo/contrib/doe/tests/test_doe_solve.py
snarasi2: Thank you. I treated this as a broader preference to avoid the reactor example unless it was specifically needed, and addressed it in three cleanup batches:

c593f0bf1: replaced the general-purpose run_doe / regression coverage in test_doe_solve.py with Rooney-Biegler, converted the generic 2D grid-search coverage to the polynomial example, and removed the redundant plain reactor Jacobian-agreement test.
36d569485: converted the perturbed-point symbolic/automatic Jacobian-agreement test in test_utils.py to Rooney-Biegler, removed the now-unused reactor helper machinery from test_utils.py, and cleaned up the remaining stale bad-model comments.
dae72f4a2: finished the remaining reactor cleanup in the touched DoE test files by replacing the remaining reactor-based error/build/solve coverage with Rooney-Biegler, using the polynomial example for the remaining generic 2D factorial cases, and rebasing the non-solve GreyBox helper/tests onto the Rooney-Biegler path with the corresponding 2-parameter FIM expectations.
After these updates, there are no remaining reactor references in the touched DoE test files, and the focused DoE bundle passes locally:

130 passed, 4 skipped, 5 warnings in 35.67s

pyomo/contrib/doe/tests/test_greybox.py
snarasi2: Thank you. I added a short comment to clarify why the skip message names MA57/HSL here: this is specific to the GreyBox cyipopt solve path, which is distinct from plain IPOPT executable availability, and the local failure mode we diagnosed on that path was a missing MA57/HSL runtime (0545dcf47).

.github/workflows/test_branches.yml
snarasi2: Please see comment on the PR in response to this

snarasi2:
@smondal13  I added a reviewer-facing markdown note to the PR that summarizes the scope, the main sensitivity/FIM equations, how those map onto the implementation, the test strategy, and the key caveats around Rooney-Biegler/polynomial coverage and the GreyBox cyipopt path (270bf8c3f).

pyomo/contrib/doe/doe.py
snarasi2: Thank you, that makes sense. I refactored this section so jacobian_rule is now defined once and branches internally on the gradient method instead of defining separate nested rule functions, and I added comments to clarify the symbolic/pynumero path, the finite-difference path, and the nominal-parameter scaling logic in that block (33d0b9874).

pyomo/contrib/doe/doe.py
snarasi2: self.fd_formula was not removed from the class; it is still the stored source of truth. The local fd_formula here is just the validated value used inside _generate_scenario_blocks(), since that method only needs an FD formula when the selected gradient method is finite-difference rather than symbolic.

pyomo/contrib/doe/doe.py
snarasi2: That old TODO no longer made sense once the finite-difference scheme was being validated earlier in the flow, so I replaced the silent fallback with an explicit DeveloperError for the unexpected branch (4209df54c).

mrmundt:
Closing in favor of #3928

PR #3907, opened 2026-04-07
Internal/mindtpy short circuit base

bernalde:
## Summary/Motivation:

Based on a comment from @jsiirola in #3861 I went ahead and revisited how we are dealing with a problem that has no discrete variables. Now we check whether the MIP solver (eventually we need to change that nomenclature, but I will leave that to a future PR) can address the problem directly. This would be the case with quadratic programs (QPs) or quadratically constrained programs (QCPs). Although Gurobi can deal with NLPs now, usually it would try to solve them to global optimality, defeating the whole purpose of MindtPy, so we are still passing that down to the NLP (again, I need to change the naming here) solver.

This PR depends on #3861, so it is worth looking into it only after merging that

- Modification in the 'shortcircuit' part of MindtPy and added tests
- 
### Legal Acknowledgement


jsiirola:
@bernalde, @Toflamus: Note that this PR is currently blocked by merge conflicts introduced by #3861.

bernalde:
@jsiirola Addressed in https://github.com/bernalde/pyomo/commit/888bf231f01d28796052895a2ff1439a32cc69d3.

I merged current `main` into `internal/mindtpy-short-circuit-base`, which clears the branch-blocking state noted in your comment. GitHub is now reporting this PR as mergeable again.

I also reran `python -m pytest pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py` on the updated branch.

bernalde:
As far as I could see, none of the failed tests had to do with code introduced in this PR. We can restart the tests later to make sure everything is working

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -291,25 +293,53 @@ def model_is_valid(self):
...
-                )
-                or obj.expr.polynomial_degree()
-                not in self.mip_objective_polynomial_degree
-            ):
-                config.logger.info(
-                    'Your model is a NLP (nonlinear program). '
-                    'Using NLP solver %s to solve.' % config.nlp_solver
+            original_obj = next(
+                self.original_model.component_data_objects(ctype=Objective, active=True)
+            )
```
jsiirola, line 350: This will raise an exception for models with no objective, because `model_is_valid()` is called before `objective_reformulation()`.  Further, it will "silently pass" for multiobjective models...?

It might make sense to move the basic Objective validation (e.g., lines 832-849) to the beginning of this function?

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -291,25 +293,53 @@ def model_is_valid(self):
...
+            problem_type_names = {
+                'LP': 'LP (linear program)',
+                'QP': 'QP (quadratic program)',
+                'QCP': 'QCP (quadratically constrained program)',
+                'NLP': 'NLP (nonlinear program)',
+            }
+            problem_type_articles = {'LP': 'an', 'QP': 'a', 'QCP': 'a', 'NLP': 'an'}
+            obj_degree = MindtPy.objective_polynomial_degree
+            if obj_degree is None:
```
jsiirola: Note that `obj_degree` is None at this point if there is no objective (even though later the objective will be a constant).

I actually don't understand the logic, either: if the `obj_degree` is None coming in, why/how would re-checking it change the degree?

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -359,12 +397,88 @@ def model_is_valid(self):
...
+            return 'QCP', 'mip', None
+
+        if obj_degree == 2:
+            if not self._mip_solver_supports_capability('quadratic_objective'):
+                return 'QP', 'nlp', 'quadratic objectives'
+            return 'QP', 'mip', None
+
+        return 'LP', 'mip', None
+
+    def _mip_solver_supports_capability(self, capability):
```
jsiirola, line 493: I think this is OK for the current solver interfaces.  However, I believe that the new interfaces (both `contrib.appsi` and `contrib.solver`) will be removing the "capability" flags all together.  In the future, I think the expectation is to just "try and solve the model", and the solver interface will raise an exception or return an error status if the solver cannot solve the model.

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -431,21 +545,31 @@ def build_ordered_component_lists(self, model):
...
         util_block.objective_list = list(
             model.component_data_objects(
                 ctype=Objective, active=True, descend_into=(Block)
             )
         )
+        util_block.objective_polynomial_degree = None
+        if len(util_block.objective_list) == 1:
+            util_block.objective_polynomial_degree = util_block.objective_list[
+                0
+            ].expr.polynomial_degree()
```
jsiirola, line 614: I think that caching this information is a reasonable idea: `polynomial_degree()` is convenient, but it requires walking the entire expression to compute it.  Repeating that work is probably an unnecessary performance hit.  A good follow-on PR would be to remove redundant calls to `polynimoal_degree()` from MindtPy.

jsiirola (changes requested):
Overall, this looks pretty good.  There is one case where woul will end up with an unexpected exception because of the order in which things are being processed (you are doing the short circuit handling before you have checked / normalized the objective).

pyomo/contrib/mindtpy/algorithm_base_class.py
bernalde, line 350: Short-circuit routing now depends on objective metadata before MindtPy has normalized the objective state. This path, together with `set_up_solve_data()`, still assumes there is exactly one active objective, so models with no objective will still raise unexpectedly and multiobjective models can take the first objective silently. Please validate or normalize the objective once before the short-circuit classification path and then use the cached degree directly here.

pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py
```diff
@@ -405,3 +405,316 @@ def test_solver_status_and_message_mirrored(self):
...
+        solver=_SimpleNamespace(
+            termination_condition=termination_condition,
+            status=SolverStatus.ok,
+            message=None,
+        ),
+        problem=_SimpleNamespace(lower_bound=None, upper_bound=None),
+    )
+
+
+class TestMindtPyShortCircuitRouting(unittest.TestCase):
```
bernalde, line 499: These routing tests cover the LP, QP, QCP, and NLP branches well, but they do not exercise the objective-normalization edge cases that this short-circuit path now depends on. Please add regressions for both a no-objective model and a multiobjective model so the behavior stays aligned with `process_objective()` and `set_up_solve_data()`.

pyomo/contrib/mindtpy/algorithm_base_class.py
bernalde: Fixed in [ee55f3921](https://github.com/bernalde/pyomo/commit/ee55f39210d9ba0e6340c31a418af731091ee8a6). I moved the single-objective normalization into `set_up_solve_data()`, so MindtPy now rejects multiobjective models before any short-circuit logic and adds a temporary dummy objective for no-objective models. The temporary objective is removed before returning to the user model. I also added regressions for both cases in `pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py`.

pyomo/contrib/mindtpy/algorithm_base_class.py
bernalde: Fixed in [ee55f3921](https://github.com/bernalde/pyomo/commit/ee55f39210d9ba0e6340c31a418af731091ee8a6). After normalizing the objective earlier, `MindtPy.objective_polynomial_degree` is the cached value used for the short-circuit check, so I removed the redundant `obj_degree is None` re-check. In this path, `None` now only means a genuinely non-polynomial objective.

pyomo/contrib/mindtpy/algorithm_base_class.py
bernalde: I left this behavior unchanged in [ee55f3921](https://github.com/bernalde/pyomo/commit/ee55f39210d9ba0e6340c31a418af731091ee8a6). For the current solver interfaces, the explicit capability checks are still the narrowest way to route continuous QP and QCP models without changing failure semantics. I did not broaden this PR to a "try MIP and fall back on any error" policy. The narrower short-circuit follow-up remains tracked in #3906.

pyomo/contrib/mindtpy/algorithm_base_class.py
bernalde: Agreed. [ee55f3921](https://github.com/bernalde/pyomo/commit/ee55f39210d9ba0e6340c31a418af731091ee8a6) keeps the cached objective degree in the short-circuit path, but I left the broader migration for a separate follow-up. I opened #3920 to track replacing the remaining MindtPy structural `polynomial_degree()` checks with `QuadraticRepnVisitor`, following the discussion in https://github.com/Pyomo/pyomo/pull/3874#discussion_r2991334592.

pyomo/contrib/mindtpy/algorithm_base_class.py
bernalde: Addressed in [ee55f3921](https://github.com/bernalde/pyomo/commit/ee55f39210d9ba0e6340c31a418af731091ee8a6). Objective normalization now happens once in `set_up_solve_data()`, the short-circuit path consumes the cached degree directly, multiobjective models raise early, and no-objective models use a temporary dummy objective that is cleaned up before returning.

pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py
bernalde: Addressed in [ee55f3921](https://github.com/bernalde/pyomo/commit/ee55f39210d9ba0e6340c31a418af731091ee8a6). I added `test_short_circuit_model_with_no_objective_uses_temporary_dummy_objective` and `test_short_circuit_multiobjective_model_raises` in `pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py`.

bernalde:
Addressed the requested review changes in [ee55f3921](https://github.com/bernalde/pyomo/commit/ee55f39210d9ba0e6340c31a418af731091ee8a6). The fix normalizes the objective before short-circuit routing, removes the redundant cached-degree re-check, adds regressions for missing-objective and multiobjective models, and keeps the current capability-based QP and QCP routing behavior in place for this PR. Tests run: `python -m pytest pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py -q` and `python -m pytest pyomo/contrib/mindtpy/tests -q` (52 passed, 32 skipped). For the broader `QuadraticRepnVisitor` follow-up, I opened #3920.

jsiirola (approved):
(pending passing tests)

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -359,12 +451,88 @@ def model_is_valid(self):
...
+            return 'QP', 'mip', None
+
+        return 'LP', 'mip', None
+
+    def _mip_solver_supports_capability(self, capability):
+        """Return whether the selected MIP solver supports a model feature."""
+        appsi_capabilities = {
+            'appsi_cplex': {'quadratic_objective': True, 'quadratic_constraint': True},
+            'appsi_gurobi': {'quadratic_objective': True, 'quadratic_constraint': True},
+            'appsi_highs': {
```
mrmundt, line 498: Not a this-PR problem most likely, but it's probably a good idea for MindtPy to migrate away from supporting / relying on APPSI sometime in the future and instead move towards the `contrib.solver` interfaces.

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -2878,72 +3052,78 @@ def solve(self, model, **kwds):
...
-            self.initial_var_values = list(v.value for v in MindtPy.variable_list)
+                # Save model initial values.
+                self.initial_var_values = list(v.value for v in MindtPy.variable_list)
 
-            # TODO: if the MindtPy solver is defined once and called several times to solve models. The following two lines are necessary. It seems that the solver class will not be init every time call.
-            # For example, if we remove the following two lines. test_RLPNLP_L1 will fail.
-            self.best_solution_found = None
-            self.best_solution_found_time = None
-            self.initialize_mip_problem()
+                # TODO: if the MindtPy solver is defined once and called several times to solve models. The following two lines are necessary. It seems that the solver class will not be init every time call.
```
mrmundt: Nitpick - can you please break this comment onto multiple lines? It's really long.

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -245,20 +249,76 @@ def create_utility_block(self, model, name):
...
+        if (
+            self._temporary_original_objective_name is None
+            or self.original_model is None
+        ):
+            return
+        if (
+            self.original_model.component(self._temporary_original_objective_name)
+            is not None
+        ):
+            self.original_model.del_component(self._temporary_original_objective_name)
```
emma58: `del_component` doesn't actually raise an error if you're deleting by name and the name doesn't exist. I only know because I just tried it out of curiosity... But for better or for worse, you don't need the `if` clause. In fact, `m.del_component(None)` is silent too, so you could just call it as long as you have the model, even if the name is None.

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -245,20 +249,76 @@ def create_utility_block(self, model, name):
...
+        if objective_count == 0:
+            if not create_dummy_objective:
+                return None, objective_count, None
+            if logger is not None:
+                logger.warning(
+                    'Model has no active objectives. Adding dummy objective.'
+                )
+            dummy_name = unique_component_name(model, 'MindtPy_dummy_objective')
+            setattr(model, dummy_name, Objective(expr=1))
+            return getattr(model, dummy_name), objective_count, dummy_name
```
emma58: What you have works fine, but just a style thing, you might do:
```
obj = Objective(expr=1)
model.add_component(dummy_name, obj)
return obj, objective_count, dummy_name
```

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -359,12 +451,88 @@ def model_is_valid(self):
         #  need to do some kind of transformation (Glover?) or throw an error message
         return True
 
+    def _classify_short_circuit_problem(self, MindtPy, obj_degree):
+        """Classify a no-discrete model for short-circuit direct solves.
```
emma58:
```suggestion
        """Classify a continuous model for short-circuit direct solves.
```

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -359,12 +451,88 @@ def model_is_valid(self):
...
+
+        This classification is intentionally independent of
+        ``quadratic_strategy``. In the no-discrete short-circuit path we are
```
emma58:
```suggestion
        ``quadratic_strategy``. In the continuous-model short-circuit path we are
```
(Or something like that?)

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -359,12 +451,88 @@ def model_is_valid(self):
...
+    def _mip_solver_supports_capability(self, capability):
+        """Return whether the selected MIP solver supports a model feature."""
+        appsi_capabilities = {
+            'appsi_cplex': {'quadratic_objective': True, 'quadratic_constraint': True},
+            'appsi_gurobi': {'quadratic_objective': True, 'quadratic_constraint': True},
+            'appsi_highs': {
+                # TODO: revisit if/when the APPSI HiGHS wrapper adds QP support.
+                'quadratic_objective': False,
+                'quadratic_constraint': False,
+            },
```
emma58, line 502: Also not a this-PR problem, but there is quadratic support in the `contrib.solver` HiGHS interface now.

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -245,20 +249,76 @@ def create_utility_block(self, model, name):
             self.build_ordered_component_lists(model)
             self.add_cuts_components(model)
 
+    def _get_main_objective(self, model, create_dummy_objective=False, logger=None):
```
emma58: It looks like you always call this with `create_dummy_objective=True`. Do you need the argument?

pyomo/contrib/mindtpy/algorithm_base_class.py
```diff
@@ -829,8 +998,14 @@ def set_up_solve_data(self, model):
             The original model to be solved in MindtPy.
         """
         config = self.config
+        self.original_model = model
+        self._temporary_original_objective_name = None
```
emma58: This line looks unnecessary? I think it has to already be `None`, and even if not, `_get_main_objective` doesn't use it and you are setting it in line 1006 below.

emma58 (approved):
Several comments below, but they're all picky or optional--nothing that needs to block the PR.

pyomo/contrib/mindtpy/algorithm_base_class.py
bernalde: Not addressed in this PR. This would change the solver-interface strategy beyond the continuous short-circuit fix, so I opened https://github.com/Pyomo/pyomo/issues/3936 to track evaluating a MindtPy migration from APPSI-specific routing to `contrib.solver` interfaces.

pyomo/contrib/mindtpy/algorithm_base_class.py
bernalde: Addressed in [ca776cdee](https://github.com/bernalde/pyomo/commit/ca776cdee086aac3419576ad3860d54a28bf7fbb). I split the long TODO into multiple comment lines and kept the existing behavior unchanged: [https://github.com/bernalde/pyomo/blob/ca776cdee086aac3419576ad3860d54a28bf7fbb/pyomo/contrib/mindtpy/algorithm_base_class.py#L3075](https://github.com/bernalde/pyomo/blob/ca776cdee086aac3419576ad3860d54a28bf7fbb/pyomo/contrib/mindtpy/algorithm_base_class.py#L3075).

pyomo/contrib/mindtpy/algorithm_base_class.py
bernalde: Addressed in [ca776cdee](https://github.com/bernalde/pyomo/commit/ca776cdee086aac3419576ad3860d54a28bf7fbb). `_cleanup_temporary_original_objective()` now calls `del_component()` whenever `original_model` is available and then clears the stored temporary objective name: [https://github.com/bernalde/pyomo/blob/ca776cdee086aac3419576ad3860d54a28bf7fbb/pyomo/contrib/mindtpy/algorithm_base_class.py#L290](https://github.com/bernalde/pyomo/blob/ca776cdee086aac3419576ad3860d54a28bf7fbb/pyomo/contrib/mindtpy/algorithm_base_class.py#L290).

pyomo/contrib/mindtpy/algorithm_base_class.py
bernalde: Addressed in [ca776cdee](https://github.com/bernalde/pyomo/commit/ca776cdee086aac3419576ad3860d54a28bf7fbb). The dummy objective is now constructed first, added with `model.add_component(...)`, and returned directly: [https://github.com/bernalde/pyomo/blob/ca776cdee086aac3419576ad3860d54a28bf7fbb/pyomo/contrib/mindtpy/algorithm_base_class.py#L282](https://github.com/bernalde/pyomo/blob/ca776cdee086aac3419576ad3860d54a28bf7fbb/pyomo/contrib/mindtpy/algorithm_base_class.py#L282).

pyomo/contrib/mindtpy/algorithm_base_class.py
bernalde: Addressed in [ca776cdee](https://github.com/bernalde/pyomo/commit/ca776cdee086aac3419576ad3860d54a28bf7fbb). The docstring now says `continuous model` for the short-circuit direct-solve classifier: [https://github.com/bernalde/pyomo/blob/ca776cdee086aac3419576ad3860d54a28bf7fbb/pyomo/contrib/mindtpy/algorithm_base_class.py#L443](https://github.com/bernalde/pyomo/blob/ca776cdee086aac3419576ad3860d54a28bf7fbb/pyomo/contrib/mindtpy/algorithm_base_class.py#L443).

pyomo/contrib/mindtpy/algorithm_base_class.py
bernalde: Addressed in [ca776cdee](https://github.com/bernalde/pyomo/commit/ca776cdee086aac3419576ad3860d54a28bf7fbb). The paragraph now uses `continuous-model short-circuit path`: [https://github.com/bernalde/pyomo/blob/ca776cdee086aac3419576ad3860d54a28bf7fbb/pyomo/contrib/mindtpy/algorithm_base_class.py#L445](https://github.com/bernalde/pyomo/blob/ca776cdee086aac3419576ad3860d54a28bf7fbb/pyomo/contrib/mindtpy/algorithm_base_class.py#L445).

pyomo/contrib/mindtpy/algorithm_base_class.py
bernalde: Not addressed in this PR. This belongs with the solver-interface migration rather than this behavior-preserving short-circuit change, so I opened https://github.com/Pyomo/pyomo/issues/3936 to track evaluating `contrib.solver` routing and HiGHS quadratic support for MindtPy.

pyomo/contrib/mindtpy/algorithm_base_class.py
bernalde: Addressed in [ca776cdee](https://github.com/bernalde/pyomo/commit/ca776cdee086aac3419576ad3860d54a28bf7fbb). I removed the unused `create_dummy_objective` argument from `_get_main_objective()` and updated both callers to use the simplified signature: [https://github.com/bernalde/pyomo/blob/ca776cdee086aac3419576ad3860d54a28bf7fbb/pyomo/contrib/mindtpy/algorithm_base_class.py#L252](https://github.com/bernalde/pyomo/blob/ca776cdee086aac3419576ad3860d54a28bf7fbb/pyomo/contrib/mindtpy/algorithm_base_class.py#L252), [https://github.com/bernalde/pyomo/blob/ca776cdee086aac3419576ad3860d54a28bf7fbb/pyomo/contrib/mindtpy/algorithm_base_class.py#L874](https://github.com/bernalde/pyomo/blob/ca776cdee086aac3419576ad3860d54a28bf7fbb/pyomo/contrib/mindtpy/algorithm_base_class.py#L874), [https://github.com/bernalde/pyomo/blob/ca776cdee086aac3419576ad3860d54a28bf7fbb/pyomo/contrib/mindtpy/algorithm_base_class.py#L988](https://github.com/bernalde/pyomo/blob/ca776cdee086aac3419576ad3860d54a28bf7fbb/pyomo/contrib/mindtpy/algorithm_base_class.py#L988).

pyomo/contrib/mindtpy/algorithm_base_class.py
bernalde: Addressed in [ca776cdee](https://github.com/bernalde/pyomo/commit/ca776cdee086aac3419576ad3860d54a28bf7fbb). I removed the redundant reset before `_get_main_objective()`; `set_up_solve_data()` now assigns `_temporary_original_objective_name` only from the returned dummy objective name: [https://github.com/bernalde/pyomo/blob/ca776cdee086aac3419576ad3860d54a28bf7fbb/pyomo/contrib/mindtpy/algorithm_base_class.py#L987](https://github.com/bernalde/pyomo/blob/ca776cdee086aac3419576ad3860d54a28bf7fbb/pyomo/contrib/mindtpy/algorithm_base_class.py#L987).

bernalde:
For the top-level Codecov report, no separate coverage-only change was made. The report shows 98.24561% patch coverage, and the author revision in [ca776cdee](https://github.com/bernalde/pyomo/commit/ca776cdee086aac3419576ad3860d54a28bf7fbb) only addresses review cleanup without adding new behavior branches.

Local checks for this revision:
`python -m pytest pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py -q` passed with 28 tests.
`python -m pytest pyomo/contrib/mindtpy/tests -q` passed with 52 tests and 32 skipped.
`python -m black --check --diff --fast pyomo/contrib/mindtpy/algorithm_base_class.py` passed.
`git diff --check` passed.

bernalde:
Could you please share the failure reason in Jenkins so I can fix this PR? Thanks!

jsiirola:
> Could you please share the failure reason in Jenkins so I can fix this PR? Thanks!

It's unrelated (it has to do with cuOpt and #3916)

PR #3912, opened 2026-04-15
Initialization Module

michaelbynum:
## Summary/Motivation:

This PR adds a module to `devel` called `initialization`. The goal of the module is to provide methods to help initialize nonconvex nonlinear programming problems.

- Add `initialization` module
- Add an initialization method that uses a global optimization solver to find a feasible solution
- Add an initialization method that builds a linear programming approximation using linear least squares
- Add an initialization method that builds a piecewise linear approximation and iteratively refines it

### Legal Acknowledgement


michaelbynum:
See #3878 for the changes I had to make to contrib.piecewise.

michaelbynum:
Thanks for updating the imports, @blnicho!

doc/OnlineDocs/conf.py
```diff
@@ -428,6 +428,7 @@ def check_output(self, want, got, optionflags):
 baron_available = bool(_opt.check_available_solvers('baron'))
 glpk_available = bool(_opt.check_available_solvers('glpk'))
 gurobipy_available = bool(_opt.check_available_solvers('gurobi_direct'))
+scip_available = bool(_opt.check_available_solvers('scip_direct'))
```
mrmundt: Can this maybe be named `pyscip_available` or `pyscipopt_available`? I think `scip_available` implies the long-time-existing command line version, but it's actually the Python interface that we are looking for.

pyomo/devel/initialization/examples/__init__.py
mrmundt, line 1: Missing copyright

pyomo/devel/initialization/global_init.py
```diff
@@ -0,0 +1,52 @@
...
+def _initialize_with_global_solver(
+    nlp: BlockData, global_solver: SolverBase, nlp_solver: SolverBase
+):
+    if isinstance(global_solver, (ScipDirect, ScipPersistent)):
+        opts = {'limits/solutions': 1}
+    elif isinstance(global_solver, (GurobiDirectMINLP,)):
+        opts = {'SolutionLimit': 1}
+    else:
+        raise NotImplementedError(
+            'Currently, the initialization module only works with new solver interface, so the global solvers are limited to ScipDirect, ScipPersistent, and GurobiDirectMINLP.'
```
mrmundt:
```suggestion
            'Currently, the initialization module only works with new solver interfaces, so the global solvers are limited to ScipDirect, ScipPersistent, and GurobiDirectMINLP.'
```

pyomo/devel/initialization/lp_approx_init.py
```diff
@@ -0,0 +1,218 @@
...
+#
+# Pyomo: Python Optimization Modeling Objects
+# Copyright (c) 2008-2026 National Technology and Engineering Solutions of Sandia, LLC
+# Under the terms of Contract DE-NA0003525 with National Technology and Engineering
+# Solutions of Sandia, LLC, the U.S. Government retains certain rights in this
+# software.  This software is distributed under the 3-clause BSD License.
+# ____________________________________________________________________________________
+
+from pyomo.core.base.block import BlockData
+import pyomo.environ as pe
```
mrmundt: _Michael_! `pyo`, not `pe` 😜

pyomo/devel/initialization/lp_approx_init.py
```diff
@@ -0,0 +1,218 @@
...
+import math
+from pyomo.contrib.solver.common.base import SolverBase
+import logging
+from pyomo.common.modeling import unique_component_name
+from pyomo.devel.initialization.pwl_init import _minimize_infeasibility
+from pyomo.contrib.fbbt.fbbt import fbbt
+from pyomo.repn.linear import LinearRepnVisitor, LinearRepn
+from pyomo.core.expr.visitor import identify_variables
+from pyomo.common.dependencies import numpy as np
+from pyomo.common.dependencies import attempt_import
```
mrmundt: Preferred import order (plus changed to pyo instead of pe).

```suggestion
import logging
import math
from typing import List, MutableMapping, Sequence

import pyomo.environ as pyo
from pyomo.common.collections import ComponentMap, ComponentSet
from pyomo.common.dependencies import attempt_import
from pyomo.common.dependencies import numpy as np
from pyomo.common.modeling import unique_component_name
from pyomo.common.numeric_types import native_numeric_types
from pyomo.contrib.fbbt.fbbt import fbbt
from pyomo.contrib.piecewise.piecewise_linear_expression import (
    PiecewiseLinearExpression,
)
from pyomo.contrib.piecewise.piecewise_linear_function import PiecewiseLinearFunction
from pyomo.contrib.solver.common.base import SolverBase
from pyomo.contrib.solver.common.results import SolutionStatus
from pyomo.core.base.block import BlockData
from pyomo.core.base.constraint import ConstraintData
from pyomo.core.base.expression import ExpressionData, ScalarExpression
from pyomo.core.base.param import ParamData, ScalarParam
from pyomo.core.base.var import ScalarVar, VarData
from pyomo.core.expr.numvalue import NumericConstant
from pyomo.core.expr.numeric_expr import (
    DivisionExpression,
    LinearExpression,
    MonomialTermExpression,
    NegationExpression,
    NPV_DivisionExpression,
    NPV_NegationExpression,
    NPV_PowExpression,
    NPV_ProductExpression,
    NPV_SumExpression,
    NPV_UnaryFunctionExpression,
    PowExpression,
    ProductExpression,
    SumExpression,
    UnaryFunctionExpression,
)
from pyomo.core.expr.relational_expr import (
    EqualityExpression,
    InequalityExpression,
    RangedExpression,
)
from pyomo.core.expr.visitor import (
    StreamBasedExpressionVisitor,
    identify_components,
    identify_variables,
)
from pyomo.devel.initialization.bounds.bound_variables import (
    bound_all_nonlinear_variables,
)
from pyomo.devel.initialization.pwl_init import _minimize_infeasibility
from pyomo.devel.initialization.utils import (
    fix_vars_with_equal_bounds,
    get_vars,
    shallow_clone,
)
from pyomo.repn.linear import LinearRepn, LinearRepnVisitor
from pyomo.repn.util import ExitNodeDispatcher
```

pyomo/devel/initialization/pwl_init.py
```diff
@@ -0,0 +1,376 @@
...
+)
+from pyomo.repn.util import ExitNodeDispatcher
+from pyomo.core.base.var import ScalarVar, VarData
+from pyomo.core.base.param import ScalarParam, ParamData
+from pyomo.core.base.expression import ScalarExpression, ExpressionData
+import math
+from pyomo.contrib.solver.common.base import SolverBase
+import logging
+from pyomo.common.modeling import unique_component_name
+from pyomo.contrib.solver.common.results import SolutionStatus
```
mrmundt:
```suggestion
import logging
import math
from typing import List, MutableMapping, Sequence

import pyomo.environ as pyo
from pyomo.common.collections import ComponentMap, ComponentSet
from pyomo.common.modeling import unique_component_name
from pyomo.common.numeric_types import native_numeric_types
from pyomo.contrib.piecewise.piecewise_linear_expression import (
    PiecewiseLinearExpression,
)
from pyomo.contrib.piecewise.piecewise_linear_function import PiecewiseLinearFunction
from pyomo.contrib.solver.common.base import SolverBase
from pyomo.contrib.solver.common.results import SolutionStatus
from pyomo.core.base.block import BlockData
from pyomo.core.base.constraint import ConstraintData
from pyomo.core.base.expression import ExpressionData, ScalarExpression
from pyomo.core.base.param import ParamData, ScalarParam
from pyomo.core.base.var import ScalarVar, VarData
from pyomo.core.expr.numvalue import NumericConstant
from pyomo.core.expr.numeric_expr import (
    DivisionExpression,
    LinearExpression,
    MonomialTermExpression,
    NegationExpression,
    NPV_DivisionExpression,
    NPV_NegationExpression,
    NPV_PowExpression,
    NPV_ProductExpression,
    NPV_SumExpression,
    NPV_UnaryFunctionExpression,
    PowExpression,
    ProductExpression,
    SumExpression,
    UnaryFunctionExpression,
)
from pyomo.core.expr.relational_expr import (
    EqualityExpression,
    InequalityExpression,
    RangedExpression,
)
from pyomo.core.expr.visitor import (
    StreamBasedExpressionVisitor,
    identify_components,
)
from pyomo.devel.initialization.bounds.bound_variables import (
    bound_all_nonlinear_variables,
)
from pyomo.devel.initialization.utils import (
    fix_vars_with_equal_bounds,
    get_vars,
    shallow_clone,
)
from pyomo.repn.util import ExitNodeDispatcher
```

pyomo/devel/initialization/pwl_init.py
```diff
@@ -0,0 +1,376 @@
...
+            else:
+                val = v.value
+                if val <= v.lb + 1e-6 + 1e-6 * abs(v.lb):
+                    val += 1e-5
+                if val >= v.ub - 1e-6 - 1e-6 * abs(v.ub):
+                    val -= 1e-5
+            var_vals.append(val)
+        # var_vals = tuple(i.value for i in expr.args)
+        # for v, val in zip(expr.args, var_vals):
+        #     print(f'{str(v):<20}{val:<20.5f}{v.lb:<20.5f}{v.ub:<20.5f}{id(v):<20}')
```
mrmundt: Residual?

pyomo/devel/initialization/utils.py
```diff
@@ -0,0 +1,53 @@
...
+# Under the terms of Contract DE-NA0003525 with National Technology and Engineering
+# Solutions of Sandia, LLC, the U.S. Government retains certain rights in this
+# software.  This software is distributed under the 3-clause BSD License.
+# ____________________________________________________________________________________
+
+import pyomo.environ as pe
+from pyomo.common.collections import ComponentSet
+from pyomo.core.base.block import BlockData
+from pyomo.core.expr.visitor import identify_variables
+import math
```
mrmundt:
```suggestion
import math

import pyomo.environ as pyo
from pyomo.common.collections import ComponentSet
from pyomo.core.base.block import BlockData
from pyomo.core.expr.visitor import identify_variables
```

mrmundt (changes requested):

michaelbynum:
Looks like tests are failing due to a cplex release yesterday...

pyomo/devel/initialization/README.md
```diff
@@ -0,0 +1 @@
+The purpose of this module is to provide methods for initializing nonlinear programming models.
```
jsiirola: Is this readme necessary?  Should it just have a pointer to the documentation?

doc/OnlineDocs/explanation/analysis/nlp_initialization/nlp_initialization.rst
```diff
@@ -0,0 +1,131 @@
...
+
+.. warning::
+
+   This package lives in :mod:`pyomo.devel`. APIs, options, and behavior may
+   change without notice.
+
+The initialization module within ``pyomo.devel.initialization`` is intended to 
+provide methods to help initialize nonconvex nonlinear programs (NLPs). The 
+goal is to increase the chance of finding a local minimizer (i.e., decrease the
+chance of getting stuck a point that locally minimizes infeasibility). If
```
blnicho:
```suggestion
chance of getting stuck at a point that locally minimizes infeasibility). If 
```

doc/OnlineDocs/explanation/index.rst
```diff
@@ -43,6 +43,7 @@ Explanations
       `Design of Experiments`
       `MPC`
       `AOS`
+      `NLP Initialization`
```
blnicho: I don't think this change was necessary. I'm pretty sure this is just an old comment block we were using to track the documentation reorganization effort.

pyomo/devel/initialization/initialize.py
```diff
@@ -0,0 +1,178 @@
...
+    # in all cases, try to solve the nlp before doing extra work
+    res = nlp_solver.solve(
+        nlp, load_solutions=False, raise_exception_on_nonoptimal_result=False
+    )
+    logger.info(f'solved NLP: {res.solution_status}, {res.termination_condition}')
+
+    if res.solution_status == SolutionStatus.optimal:
+        res.solution_loader.load_vars()
+        logger.info('NLP solved without any initialization')
+        return res
```
blnicho: What do you think about adding an option to skip this initial NLP solve? I feel like this could add a lot of extra time for complex models and I can picture a scenario where a user has already tried and failed to solve it directly so they turn to this initialization tool and have to wait again for the NLP solver to fail (which could take a while even with the default solver settings).

pyomo/devel/initialization/initialize.py
```diff
@@ -0,0 +1,178 @@
...
+
+    # get all variable bounds, domains, etc. to restore them later
+    orig_vars = get_vars(nlp)
+    orig_var_data = ComponentMap(
+        (v, (v.lower, v.upper, v.domain, v.fixed, v.value)) for v in orig_vars
```
blnicho: Should we cache the variable's `stale` status also?

pyomo/devel/initialization/initialize.py
```diff
@@ -0,0 +1,178 @@
...
+# Copyright (c) 2008-2026 National Technology and Engineering Solutions of Sandia, LLC
+# Under the terms of Contract DE-NA0003525 with National Technology and Engineering
+# Solutions of Sandia, LLC, the U.S. Government retains certain rights in this
+# software.  This software is distributed under the 3-clause BSD License.
+# ____________________________________________________________________________________
+
+from typing import Optional
+from pyomo.core.base.block import BlockData
+from enum import Enum
+from pyomo.devel.initialization.utils import get_vars, shallow_clone
```
blnicho:
```suggestion
from pyomo.devel.initialization.utils import get_vars
```
It doesn't look like `shallow_clone` is used here

pyomo/devel/initialization/initialize.py
```diff
@@ -0,0 +1,178 @@
...
+    # restore variable bounds, domain, etc.
+    for v, (lb, ub, domain, fixed, value) in orig_var_data.items():
+        v.setlb(lb)
+        v.setub(ub)
+        v.domain = domain
+        if fixed:
+            assert v.value == value
+            assert v.fixed
+        else:
+            v.unfix()
```
blnicho: Why do we revert the variables to their original state here? Wouldn't you only want to revert the state if the initialization was unsuccessful?

pyomo/devel/initialization/lp_approx_init.py
```diff
@@ -0,0 +1,232 @@
...
+    vlist = list(vset)
+    n_vars = len(vlist)
+    bnds_list = []
+    for v in vlist:
+        if v.lb is None:
+            lb = -1e6
+        else:
+            lb = v.lb
+        if v.ub is None:
+            ub = 1e6
```
blnicho: Should these be using the `default_bound` instead of being hard-coded to `1e6`?

blnicho (commented):
@michaelbynum I have a few questions and suggestions but overall I think this looks great and I'm excited to try it out!

pyomo/devel/initialization/bounds/bound_variables.py
```diff
@@ -0,0 +1,36 @@
...
+logger = logging.getLogger(__name__)
+
+
+def bound_all_nonlinear_variables(m: BlockData, default_bound: float = 1.0e8):
+    """
+    Attempt to obtain valid bounds on all nonlinear variables based on the
+    constraints in the model, m. If variable bounds cannot be obtained,
+    we use default_bound.
+    """
+    fbbt(m)
```
jsiirola, line 24: Is it worth only running FBBT on constraints that actually involve unbounded variables (instead of attempting to tighten all variables in the model)?

pyomo/devel/initialization/initialize.py
```diff
@@ -0,0 +1,178 @@
...
+    if opt.available():
+        logger.info(f'Using {sname} for {reason} because a solver was not specified')
+    else:
+        raise RuntimeError(
+            f'No solver was specified for {reason} and the default ({sname}) is not available'
+        )
+    return opt
+
+
+def initialize_nlp(
```
jsiirola: Design question: would it better to make this a context manager instead of a "global gateway for all initialization routines"?  The problem with the "gateway function" is that:
- it has to be updated every time we add a new strategy/method
- the API must be the superset of all the options / arguments that any method takes (and then there is the question about warning about arguments that the user provided, but are ignored by the chosen method)

Instead, if you made this a context manager, it might look like:
```python
with initialization_context(m):
    res = initialize_with_piecewise_linear_approximation(
        nlp=m,
        mip_solver="xpress",
        max_iter=5,
    )
```

And InitializationContext could be as simple as
```python
@context_manager
def initialization_context(nlp):
    # get all variable bounds, domains, etc. to restore them later
    orig_var_data = [
        (v, v.lower, v.upper, v.domain, v.fixed, v.value) for v in get_vars(nlp)
    ]
    yield
    for v, lb, ub, domain, fixed, value in orig_var_data:
        v.setlb(lb)
        v.setub(ub)
        v.domain = domain
        if fixed:
            assert v.value == value
            assert v.fixed
        else:
            v.unfix()
```

pyomo/devel/initialization/initialize.py
```diff
@@ -0,0 +1,178 @@
...
+    logger.info(f'solved NLP: {res.solution_status}, {res.termination_condition}')
+
+    if res.solution_status == SolutionStatus.optimal:
+        res.solution_loader.load_vars()
+        logger.info('NLP solved without any initialization')
+        return res
+
+    # get all variable bounds, domains, etc. to restore them later
+    orig_vars = get_vars(nlp)
+    orig_var_data = ComponentMap(
```
jsiirola: This doesn't need to be a ComponentMap; a list is sufficient

pyomo/devel/initialization/initialize.py
jsiirola: Probably not: `stale` isn't settable (the bool is the relative value between the local vale and the global stale identifier).

pyomo/devel/initialization/global_init.py
```diff
@@ -0,0 +1,52 @@
...
+from pyomo.contrib.solver.solvers.scip.scip_direct import ScipDirect
+from pyomo.contrib.solver.solvers.scip.scip_persistent import ScipPersistent
+from pyomo.contrib.solver.solvers.gurobi.gurobi_direct_minlp import GurobiDirectMINLP
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+def _initialize_with_global_solver(
+    nlp: BlockData, global_solver: SolverBase, nlp_solver: SolverBase
```
jsiirola, line 22: Style question: the general convention elsewhere in Pyomo (e.g., in transformations), is for this to be `model`.  Is there a reason not to follow that convention here?

pyomo/devel/initialization/global_init.py
```diff
@@ -0,0 +1,52 @@
...
+):
+    if isinstance(global_solver, (ScipDirect, ScipPersistent)):
+        opts = {'limits/solutions': 1}
+    elif isinstance(global_solver, (GurobiDirectMINLP,)):
+        opts = {'SolutionLimit': 1}
+    else:
+        raise NotImplementedError(
+            'Currently, the initialization module only works with new solver interfaces, so the global solvers are limited to ScipDirect, ScipPersistent, and GurobiDirectMINLP.'
```
jsiirola: Please break/wrap to <88 characters

pyomo/devel/initialization/global_init.py
```diff
@@ -0,0 +1,52 @@
...
+    logger.info(
+        f'solved NLP with {global_solver.name}: {res.solution_status}, {res.termination_condition}'
+    )
+    res = nlp_solver.solve(
+        nlp, load_solutions=False, raise_exception_on_nonoptimal_result=False
+    )
+    logger.info(
+        f'solved NLP with {nlp_solver.name}: {res.solution_status}, {res.termination_condition}'
+    )
+    if res.solution_status in {SolutionStatus.feasible, SolutionStatus.optimal}:
```
jsiirola, line 49: Should this be a helper (e.g., `check_feasible_solution(res)`)? That way we can standardize / centralize the logic around SolutionStatus (similar to `check_optimal_termination`)?

pyomo/devel/initialization/lp_approx_init.py
```diff
@@ -0,0 +1,232 @@
...
+from pyomo.devel.initialization.pwl_init import _minimize_infeasibility
+from pyomo.devel.initialization.utils import (
+    fix_vars_with_equal_bounds,
+    get_vars,
+    shallow_clone,
+)
+from pyomo.repn.linear import LinearRepn, LinearRepnVisitor
+from pyomo.repn.util import ExitNodeDispatcher
+
+qmc, qmc_avail = attempt_import('scipy.stats.qmc')
```
jsiirola: You don't need `attempt_import` here.  It is sufficient to do:
```
from pyomo.common.dependencies.scipy import stats
```
(and then just do `stats.qmc.LatinHypercube` below)

pyomo/devel/initialization/initialize.py
blnicho: Nevermind, I read the code again and realized you aren't resetting the variable values, just their other attributes.

pyomo/devel/initialization/lp_approx_init.py
```diff
@@ -0,0 +1,232 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def _replace_expression_with_linear_approx(expr, num_samples=100, seed=None):
+    vset = ComponentSet(identify_variables(expr, include_fixed=False))
```
jsiirola: You shouldn't need ComponentSet here - identify_variables should remove duplicates

pyomo/devel/initialization/lp_approx_init.py
```diff
@@ -0,0 +1,232 @@
...
+    A = np.zeros((num_samples, n_coefs), dtype=float)
+    b = np.zeros(num_samples, dtype=float)
+    A[:, :n_vars] = sample
+    A[:, n_vars] = 1
+    for sample_ndx in range(num_samples):
+        for v, val in zip(vlist, sample[sample_ndx, :]):
+            v.value = float(val)
+        b[sample_ndx] = pyo.value(expr)
+    coefs = np.linalg.solve(A.transpose().dot(A), A.transpose().dot(b))
+    coefs = [float(i) for i in coefs]
```
jsiirola: Is the following equivalent / faster / more numpythic?
```suggestion
    coefs = np.linalg.solve(A.transpose().dot(A), A.transpose().dot(b)).tolist()
```

pyomo/devel/initialization/lp_approx_init.py
```diff
@@ -0,0 +1,232 @@
...
+    # now we want to build the matrix and the right hand side
+    # we have a linear coefficient for each variable plus a constant
+    n_coefs = n_vars + 1
+    A = np.zeros((num_samples, n_coefs), dtype=float)
+    b = np.zeros(num_samples, dtype=float)
+    A[:, :n_vars] = sample
+    A[:, n_vars] = 1
+    for sample_ndx in range(num_samples):
+        for v, val in zip(vlist, sample[sample_ndx, :]):
+            v.value = float(val)
```
jsiirola: Is the following equivalent / faster / more numpythic?

```suggestion
        for v, val in zip(vlist, sample[sample_ndx, :].tolist()):
            v.value = val
```

pyomo/devel/initialization/lp_approx_init.py
```diff
@@ -0,0 +1,232 @@
...
+    if objs:
+        if len(objs) > 1:
+            raise NotImplementedError(
+                'lp approximation does not support multiple objectives'
+            )
+        obj = objs[0]
+        repn = visitor.walk_expression(obj)
+        assert repn.multiplier == 1
+        linear_part = LinearRepn()
+        linear_part.multiplier = 1
```
jsiirola: FWIW, I have a branch somewhere that removes multiplier from the LinearRepn (mostly to avoid this confusion).  The walker guarantees that non-unity values are distributed into the compiled representations.

pyomo/devel/initialization/lp_approx_init.py
```diff
@@ -0,0 +1,232 @@
...
+        linear_part.constant = repn.constant
+        linear_part.linear = repn.linear
+        linear_part.nonlinear = None
+        new_obj_expr = linear_part.to_expression(visitor=visitor)
+        if repn.nonlinear is not None:
+            replacement = _replace_expression_with_linear_approx(
+                repn.nonlinear, num_samples=num_samples, seed=seed_handler.spawn(1)[0]
+            )
+            new_obj_expr += replacement
+        lp.obj = pyo.Objective(expr=new_obj_expr, sense=obj.sense)
```
jsiirola: There is no reason to regenerate the expression if we have already proven it's linear:
```suggestion
        if repn.nonlinear is not None:
            linear_part = LinearRepn()
            linear_part.constant = repn.constant
            linear_part.linear = repn.linear
            new_obj_expr = (
                linear_part.to_expression(visitor=visitor) +
                _replace_expression_with_linear_approx(
                    repn.nonlinear, num_samples=num_samples, seed=seed_handler.spawn(1)[0]
                )
            )
        else:
            new_obj_expr = obj.expr
        lp.obj = pyo.Objective(expr=new_obj_expr, sense=obj.sense)
```

(the same comment applies to constraints below)

pyomo/devel/initialization/lp_approx_init.py
```diff
@@ -0,0 +1,232 @@
...
+)
+from pyomo.repn.linear import LinearRepn, LinearRepnVisitor
+from pyomo.repn.util import ExitNodeDispatcher
+
+qmc, qmc_avail = attempt_import('scipy.stats.qmc')
+
+logger = logging.getLogger(__name__)
+
+
+def _replace_expression_with_linear_approx(expr, num_samples=100, seed=None):
```
jsiirola: This technically doesn't replace expressions; it generates a linear approximation:

```suggestion
def _generate_linear_approx(expr, num_samples=100, seed=None):
```

pyomo/devel/initialization/lp_approx_init.py
```diff
@@ -0,0 +1,232 @@
...
+    lp_solver: SolverBase,
+    nlp_solver: SolverBase,
+    default_bound=1.0e8,
+    num_samples=100,
+    seed=None,
+):
+    orig_nlp = nlp
+    logger.info('Starting initialization using a linear programming approximation')
+    nlp = shallow_clone(nlp)
+    logger.info('created a shallow clone of the model')
```
jsiirola, line 185: Why shallow clone?  the driver should restore the model state, right?

pyomo/devel/initialization/lp_approx_init.py
```diff
@@ -0,0 +1,232 @@
...
+
+    # Now, we need to fix variables with equal (or nearly equal) bounds.
+    # Otherwise, the PWL transformation complains
+    fix_vars_with_equal_bounds(nlp)
+    logger.info('fixed variables with equal bounds')
+
+    # now we modify the model by introducing slacks to make sure the LP
+    # approximation is feasible
+    _minimize_infeasibility(nlp)
+    logger.info('reformulated model to minimize infeasibility')
```
jsiirola: instead of doing this to the original model, we could just add this objective to the new `lp` block we generate next...

Follow-up: why add slack variables first?  Don't they just have to be filtered out as part of the "linear part" of the constraint?  Can you save that work by adding the slacks after you have reformulated the constraints?

pyomo/devel/initialization/utils.py
```diff
@@ -0,0 +1,56 @@
...
+
+import math
+
+import pyomo.environ as pyo
+from pyomo.common.collections import ComponentSet
+from pyomo.core.base.block import BlockData
+from pyomo.core.expr.visitor import identify_variables
+
+
+def get_vars(m: BlockData):
```
jsiirola, line 19: This is basically a reimplementation of 
```
pyomo.util.vars_from_expressions.get_vars_from_components(
    m,
    ctype=(pyo.Constraint, pyo.Objective), 
    include_fixed=False,
    active=True
)
```

pyomo/devel/initialization/utils.py
```diff
@@ -0,0 +1,56 @@
...
+    )
+    assert len(objlist) <= 1
+    if objlist:
+        obj = objlist[0]
+        m2.obj = pyo.Objective(expr=obj.expr, sense=obj.sense)
+
+    return m2
+
+
+def fix_vars_with_equal_bounds(m):
```
jsiirola: Should this take the tolerance as an argument?

pyomo/devel/initialization/pwl_init.py
```diff
@@ -0,0 +1,373 @@
...
+            ns.setlb(0)
+            con.set_value(body - lb - ps + ns == 0)
+        elif lb is None:
+            ps = m.slacks.add()
+            ps.setlb(0)
+            con.set_value(body - ub - ps <= 0)
+        elif ub is None:
+            ns = m.slacks.add()
+            ns.setlb(0)
+            con.set_value(body - lb + ns >= 0)
```
jsiirola: Will you get an exception if the constraint is unbounded?  Maybe catch that case above with
```python
if lb == ub:
    if lb is not None: 
        ps = m.slacks.add()
        # ...
    else:
        # trivial constraint; no slacks needed
        pass
```

pyomo/devel/initialization/pwl_init.py
```diff
@@ -0,0 +1,373 @@
...
+    fix_vars_with_equal_bounds,
+    get_vars,
+    shallow_clone,
+)
+from pyomo.repn.util import ExitNodeDispatcher
+
+logger = logging.getLogger(__name__)
+
+
+def _minimize_infeasibility(m):
```
jsiirola, line 66: This is basically a reimplementation of the `core.add_slack_variables` transformation.  Why not just use it?

pyomo/devel/initialization/pwl_init.py
```diff
@@ -0,0 +1,373 @@
...
+    NPV_SumExpression,
+    NPV_UnaryFunctionExpression,
+    EqualityExpression,
+    InequalityExpression,
+    RangedExpression,
+]:
+    _handlers[t] = _handle_node
+
+
+class _PWLRefinementVisitor(StreamBasedExpressionVisitor):
```
jsiirola, line 150: I lost the thread on the logic here, but this seems to do way more expression manipulation than I would think is necessary.  I would have thought this could have been done in one pass through the model where you substituted the nonlinear expressions with a new variable.  If you kept a list of all the "new" constraints, you would have only had to rebuild the refined piecewise linearization of those expressions (no walker needed)?

pyomo/devel/initialization/pwl_init.py
```diff
@@ -0,0 +1,373 @@
...
+            if math.isclose(v.lb, v.ub, rel_tol=1e-6, abs_tol=1e-6):
+                val = 0.5 * (v.lb + v.ub)
+            elif v.value is None:
+                val = None
+            else:
+                val = v.value
+                if val <= v.lb + 1e-6 + 1e-6 * abs(v.lb):
+                    val += 1e-5
+                if val >= v.ub - 1e-6 - 1e-6 * abs(v.ub):
+                    val -= 1e-5
```
jsiirola: Should these embedded constants be moved to either an argument or a global named constant?

pyomo/devel/initialization/pwl_init.py
```diff
@@ -0,0 +1,373 @@
...
+            continue
+        approx_value = func(*var_vals)
+        true_value = func._func(*var_vals)
+        err = abs(true_value - approx_value)
+        violations.append((err, expr))
+    violations.sort(key=lambda i: i[0], reverse=True)
+
+    if len(violations) == 0:
+        raise RuntimeError(
+            'Did not find any piecewise linear functions with variable values'
```
jsiirola: I do not understand this exception message.  Can you clarify it?

jsiirola (commented):
This is really cool.  Lots of comments / questions, but I don't think there was anything that should block merging.

pyomo/devel/initialization/tests/test_initialization.py
```diff
@@ -0,0 +1,236 @@
...
+from pyomo.common import unittest
+from pyomo.common.dependencies import scipy_available
+from pyomo.contrib.solver.common.factory import SolverFactory
+from pyomo.contrib.solver.common.results import (
+    SolutionStatus,
+    Results,
+    TerminationCondition,
+)
+from pyomo.contrib.solver.common.base import Availability, SolverBase
+import pytest
```
blnicho:
```suggestion
from pyomo.common.unittest import pytest
```

michaelbynum:
Thanks for all the feedback! Working on this now.

pyomo/devel/initialization/lp_approx_init.py
michaelbynum: I don't think this code should exist. We set the bounds to `default_bound` before we ever get to this function. I'm going to delete it. Good catch.

pyomo/devel/initialization/initialize.py
michaelbynum: I don't know how to ensure users use the context manager. However, I agree. I'm going to just put this code into helper functions and ensure all the initialization functions call them.

pyomo/devel/initialization/initialize.py
michaelbynum: Done.

pyomo/devel/initialization/README.md
michaelbynum: Removed.

pyomo/devel/initialization/bounds/bound_variables.py
michaelbynum: No, I don't think so. This is easiest explained with an example:

```
In [1]: import pyomo.environ as pyo
   ...: from pyomo.contrib.fbbt.fbbt import fbbt
   ...: m = pyo.ConcreteModel()
   ...: m.x = pyo.Var(bounds=(0, 100))
   ...: m.y = pyo.Var(bounds=(0, 1))
   ...: m.z = pyo.Var()
   ...: m.c1 = pyo.Constraint(expr=m.x + m.y == 1)
   ...: m.c2 = pyo.Constraint(expr=m.x == pyo.log(m.z))

In [2]: fbbt(m.c2)
   ...: m.z.pprint()
z : Size=1, Index=None
    Key  : Lower : Value : Upper                  : Fixed : Stale : Domain
    None :   1.0 :  None : 2.6881171418161356e+43 : False :  True :  Reals

In [3]: fbbt(m)
   ...: m.z.pprint()
z : Size=1, Index=None
    Key  : Lower : Value : Upper             : Fixed : Stale : Domain
    None :   1.0 :  None : 2.718281828459045 : False :  True :  Reals
```

Only constraint `c2` has unbounded variables. However, if we only do FBBT on `c2`, we get an unhelpful upper bound on `z`. If we do FBBT on the whole model, then we get good bounds on `z`.

pyomo/devel/initialization/global_init.py
michaelbynum: I think I was just emphasizing the type of model this is useful for. If you would like me to change it, let me know.

pyomo/devel/initialization/global_init.py
michaelbynum: I don't think I understand what you are suggesting. This is a 1-line check. What would the helper function do?

pyomo/devel/initialization/lp_approx_init.py
michaelbynum: Why does this matter? What's wrong with the way I have done it (so that I can do it right the first time in the future)?

pyomo/devel/initialization/lp_approx_init.py
michaelbynum: Good catch! Fixed.

pyomo/devel/initialization/lp_approx_init.py
michaelbynum: The driver only restores the variable bounds and domains. The shallow clone let's us modify constraints and objectives without modifying the original model. Plus, it is cheap.

pyomo/devel/initialization/lp_approx_init.py
michaelbynum: Good point. I don't think the order should matter here. I'll swap them.

pyomo/devel/initialization/utils.py
michaelbynum: Oh, absolutely.

pyomo/devel/initialization/pwl_init.py
michaelbynum: I always forget to consider unbounded constraints...

pyomo/devel/initialization/pwl_init.py
michaelbynum: That is indeed a very bad message, haha!

pyomo/devel/initialization/global_init.py
jsiirola: I was suggesting that we might want to move this 1-liner into a standard helper alongside `check_optimal_termination`, not because it would make *this* PR materially better, but so that we could help later PRs.

My thought was that this is a common check that people need to do, and that check may need to be solver-interface-specific.  Centralizing it would enable us to better support old/new solver interfaces and would prevent users from having to "look up" how to "correctly" check for a feasible solution from a solver.

pyomo/devel/initialization/lp_approx_init.py
jsiirola: First, nothing is wrong with the way you have done it - your implementation will work just fine.

My suggestion was to reduce the overhead of setting up (and later resolving) a second `attempt_import` (and to "show off" a bit of useful semantic sugar that the deferred importers support).  The `scipy` importer in `dependencies` also has a custom callback that both allows you to import a "deferred submodule importer" for the `stats` submodule that doesn't immediately trigger the resolution, but also *does* trigger the import of the stats module when scipy is imported (either directly or through the deferred importer objects).

pyomo/devel/initialization/global_init.py
michaelbynum: Ah. Got it. That makes sense. Thanks.

pyomo/devel/initialization/pwl_init.py
michaelbynum: I deleted this code in favor of core.add_slack_variables.

pyomo/devel/initialization/pwl_init.py
michaelbynum: As far as I can tell, it is nearly impossible to rebuild/refine a piecewise linear function/expression in place. My memory is a little fuzzy, but I recall that I needed to actually do an expression replacement type thing to replace the piecewise linear functions/expressions with new ones that had different points/segments. Hence, the visitor. However, I am careful in `_refine_pwl_approx` to only walk expressions that have piecewise linear functions/expressions.

pyomo/devel/initialization/pwl_init.py
michaelbynum: For sure.

pyomo/devel/initialization/lp_approx_init.py
michaelbynum: I removed this assertion.

michaelbynum:
I think I am done making changes. This should be ready for a re-review. If I forgot something or did not address a comment appropriately, please let me know.

pyomo/devel/initialization/global_init.py
michaelbynum: I think that the helper function should be a separate PR?

michaelbynum:
And I already lied. I forgot one change. Should be good for a re-review now.

pyomo/devel/initialization/examples/init_polynomial_ex.py
```diff
@@ -0,0 +1,60 @@
...
+    results = ini.initialize_with_global_opt(
+        nlp=m, nlp_solver=nlp_solver, global_solver=global_solver
+    )
+
+    return results.solution_status, m.x.value
+
+
+if __name__ == '__main__':
+    # stat, x = lp_init_ex()
+    # stat, x = pwl_init_ex()
```
mrmundt, line 58: Intentionally commented out?

pyomo/devel/initialization/global_init.py
mrmundt: I agree with Michael - this can be in a separate PR.

pyomo/devel/initialization/initialize.py
```diff
@@ -0,0 +1,288 @@
...
+    logger.info(f'solved NLP: {res.solution_status}, {res.termination_condition}')
+
+    if res.solution_status == SolutionStatus.optimal:
+        res.solution_loader.load_vars()
+        logger.info('NLP solved without any initialization')
+    return res
+
+
+def initialize_with_piecewise_linear_approximation(
+    nlp: BlockData,
```
mrmundt, line 81: John asked a similar question on I think a test file earlier ("why nlp instead of model?"). For tests, I really don't think it matters, but here, I think it should be model. It'll make more sense to our users if they look at the docs to see `model` (the Pyomo style standard) rather than `nlp`.

pyomo/devel/initialization/initialize.py
```diff
@@ -0,0 +1,288 @@
...
+        bounds_tol=bounds_tol,
+    )
+
+    _cleanup(orig_var_data)
+
+    return res
+
+
+def initialize_with_LP_approximation(
+    nlp: BlockData,
```
mrmundt, line 163: Same as above (`model` instead of `nlp`).

pyomo/devel/initialization/initialize.py
```diff
@@ -0,0 +1,288 @@
...
+        aggressive_substitution=aggressive_substitution,
+    )
+
+    _cleanup(orig_var_data)
+
+    return res
+
+
+def initialize_with_global_opt(
+    nlp: BlockData,
```
mrmundt, line 246: Same as above (`model` instead of `nlp`).

pyomo/devel/initialization/pwl_init.py
```diff
@@ -0,0 +1,360 @@
...
+        new_expr = new_func(*variables)
+        for v, val in zip(variables, var_values):
+            v.set_value(val, skip_validation=True)
+        self.named_expr_map[node] = new_expr
+        self.substitution[child] = new_expr.expr
+        return False, new_expr.expr
+
+
+def _refine_pwl_approx(
+    m,
```
mrmundt: Total nitpick
```suggestion
    m: BlockData,
```

mrmundt (changes requested):
I do have somewhat stronger feelings about `model` vs `nlp` on the public-facing functions. Otherwise, this is looking great.

pyomo/devel/initialization/initialize.py
michaelbynum: I didn't realize there was a pyomo style standard here. I don't think I agree that `model` is any more descriptive than `nlp`. Also, we are not consistent at all. In fact, I think we use `m` more than `model`:

`find ./ -name "*.py" | xargs grep "(m," | wc -l`:    1685
`find ./ -name "*.py" | xargs grep "(model," | wc -l`: 1025
`find ./ -name "*.py" | xargs grep "(instance," | wc -l`: 104
`find ./ -name "*.py" | xargs grep "(block," | wc -l`: 126
`find ./ -name "*.py" | xargs grep "(b," | wc -l`: 246

pyomo/devel/initialization/examples/init_polynomial_ex.py
michaelbynum: Yes. This was intentional. Just to make it easy to run the example with different methods.

pyomo/devel/initialization/initialize.py
michaelbynum: There are also many other examples, such as `working_blk`, `working_model`, `separation_model`, `subproblem`, `root`, `blk`, `scenario_instance`, `ef`, etc. Some of these argument names have meanings that are significantly more useful than `m` or `model`.

blnicho (approved):

pyomo/devel/initialization/initialize.py
```diff
@@ -0,0 +1,288 @@
...
+        mip_solver=mip_solver,
+        nlp_solver=nlp_solver,
+        default_bound=default_bound,
+        max_iter=max_pwl_refinement_iter,
+        num_cons_to_refine_per_iter=num_pwl_cons_to_refine_per_iter,
+        aggressive_substitution=aggressive_substitution,
+        bounds_tol=bounds_tol,
+    )
+
+    _cleanup(orig_var_data)
```
jsiirola: Should this be in a `try ... finally` block to guarantee that `_cleanup()` gets called?

pyomo/devel/initialization/initialize.py
```diff
@@ -0,0 +1,288 @@
...
+        lp_solver=lp_solver,
+        nlp_solver=nlp_solver,
+        default_bound=default_bound,
+        num_samples=num_samples_per_nonlinear_constraint,
+        seed=seed,
+        use_univariate_nonlinear_decomposition=use_univariate_nonlinear_decomposition,
+        aggressive_substitution=aggressive_substitution,
+    )
+
+    _cleanup(orig_var_data)
```
jsiirola: Again,, should this be in a `try .. finally`?

pyomo/devel/initialization/initialize.py
```diff
@@ -0,0 +1,288 @@
...
+    if opt.available():
+        logger.info(f'Using {sname} for {reason} because a solver was not specified')
+    else:
+        raise RuntimeError(
+            f'No solver was specified for {reason} and the default ({sname}) is not available'
+        )
+    return opt
+
+
+def _setup(nlp):
```
jsiirola, line 40: Should setup also relax any discrete variables to continuous domains?

pyomo/devel/initialization/utils.py
```diff
@@ -0,0 +1,56 @@
...
+
+def get_vars(m: BlockData):
+    return ComponentSet(
+        get_vars_from_components(
+            m, ctype=(pyo.Constraint, pyo.Objective), include_fixed=False, active=True
+        )
+    )
+
+
+def shallow_clone(m1):
```
jsiirola, line 27: In the future, we should promote this to a standard transformation.  Not only does it make a fast shallow copy, but it collapses the entire block structure, which could be useful in a lot of contexts.  It could generate more of a "flattened view" into a model than a "shallow copy"?

jsiirola (changes requested):
Two minor changes / questions (and some notes for later development)

pyomo/devel/initialization/initialize.py
michaelbynum: Absolutely. Thank you.

pyomo/devel/initialization/initialize.py
michaelbynum: I added an exception for now. I want some time to think about what it would mean to allow models with discrete variables.

pyomo/devel/initialization/utils.py
michaelbynum: That is a great suggestion. I don't think "view" is the right term. What you get is a valid model. Flattened is very accurate, though...

michaelbynum: Would this fit into the normal transformation workflow? `apply_to` wouldn't really make sense, and `create_using` might also not do what you expect. Maybe I'm thinkin about it wrong...

jsiirola: My initial thought would be to have `apply_to` raise an exception and only support `create_using` (similar to the dual transform).  I would argue that "view" is correct, as when you solve the flattened view, the solution "appears" in the original structured model (because they share variables).  If we go down the transformation route, there are questions around other components, though (e.g., should we map results for suffixes back?).

jsiirola (approved):

mrmundt (approved):

blnicho:
One of the GHA jobs is randomly getting stuck so I'm going to merge this without waiting for the last check.

pyomo/devel/initialization/utils.py
michaelbynum: That is a good point. I thought that there was an expectation of `create_using` that a full clone would get created?

michaelbynum: I guess that is where naming and documentation comes in.

jsiirola: The idea was `apply_to` is "in-place" and `create_using` is "out-of-place".  There was concern that "in place" and "out of place" were less user friendly, and `apply_to` and `create_using` (short for "create a new model using data from model `m`") were more clear.

"Cloning" only came into play because many transformations (like GDP) are most efficiently written in-place, so the default implementation of `create_using` is simply to clone and then run apply_to...

michaelbynum: Got it. That makes sense.

PR #3915, opened 2026-04-16
Fix memory leak in appsi LegacySolverInterface wrapper

Marl0nL:
## Fixes # .
Prevents a memory leak observed when iteratively solving via the LegacySolverInterface wrapper.
## Summary/Motivation:
Observed an issue with significant increase in memory use when iteratively solving a model (that fundamentally wasn't changing size) via `SolverFactory('appsi_cbc')`, which uses the the APPSI LegacySolverInterface.
Found that the growing memory use was the symbol maps.

The memory leak can be avoided by not using the LegacySolverInterface wrapper in the first place, but it seems worth fixing the legacy wrapper to avoid unsuspecting users swapping "cbc" for "appsi_cbc" when calling `SolverFactory()`, and then having memory issues.

While I can't think of why someone would intensionally leverage the current behaviour, it's possible I'm naive to other usage paradigms which may conflict with this fix. If that is the case, it may be preferable to add a new option which handles symbol map deletion or retention.

- LegacySolverInterface.solve() now ensures the clean-up of symbol maps after solving, which prevents an effective memory-leak when iteratively re-solving a model via the interface.
- Added test_legacy_leak.py, which can be used to demonstrate the memory leak issue, and the resolution thereof. 
I recommend checking out commit `65402fa992afeae269decc9e69b1e4bac9c47d37`, which adds only the test script, to verify the existence of the issue prior to the application of the fix in subsequent commits.

### Legal Acknowledgement


pyomo/contrib/appsi/tests/test_legacy_leak.py
```diff
@@ -0,0 +1,70 @@
+import pyomo.environ as pyo
```
mrmundt, line 10: Thanks for this! We love new external contributions. Can you please add the copyright statement to the top of this file?

pyomo/contrib/appsi/base.py
```diff
@@ -1634,6 +1634,11 @@ def solve(
         legacy_results.solution.insert(legacy_soln)
```
mrmundt, line 1634: It looks like this same code was copied over to `pyomo.contrib.solver.common.base`: https://github.com/Pyomo/pyomo/blob/main/pyomo/contrib/solver/common/base.py#L597

Would you mind updating it there, too? We don't have the cbc solver ported over to the new solver interface structure yet, but it would be good to avoid having this problem in the future.

mrmundt (changes requested):

pyomo/contrib/appsi/base.py
```diff
@@ -1634,6 +1634,11 @@ def solve(
...
         if delete_legacy_soln:
             legacy_results.solution.delete(0)
```
jsiirola, line 1636: @michaelbynum: why do you insert the solution just to immediately delete it?

pyomo/contrib/appsi/base.py
```diff
@@ -1634,6 +1634,11 @@ def solve(
...
+            model.solutions.delete_symbol_map(legacy_results._smap_id)
+            legacy_results._smap_id = None
+        else:
+            legacy_results._smap = model.solutions.symbol_map[legacy_results._smap_id]
+            model.solutions.delete_symbol_map(legacy_results._smap_id)
```
jsiirola, line 1641: Instead of deleting the symbol map, if we follow this suggestion of storing the SymbolMap on the legacy_results as `_smap` (which I think I like), can we just never call `model.solutions.add_symbol_map()` in the first place?

jsiirola (commented):
First, this is great.  It also points to a fundamental design flaw in the old model `model.solutions` cache - this is inherently a source of memory "leaks" (not a leak, per se, but an implicit, monotonically increasing cache of questionable value).

I just looked, and the only thing that was documented in The Book was `model.solutions.load_from()` ... we may be able to start reworking that entire infrastructure in advance of Pyomo 7.

pyomo/contrib/appsi/tests/test_legacy_leak.py
```diff
@@ -0,0 +1,70 @@
...
+import pyomo.common.unittest as unittest
+import gc
+from pyomo.contrib.appsi.cmodel import cmodel_available
+
+try:
+    import tracemalloc
```
jsiirola, line 16: When is `tracemalloc` not available?

Marl0nL: @jsiirola on the first CI run, I saw that the run with PyPy 3.11 failed with the following:
```
pyomo/contrib/appsi/tests/test_legacy_leak.py:3: in <module>
    import tracemalloc
/opt/hostedtoolcache/PyPy/3.11.13/x64/lib/pypy3.11/tracemalloc.py:9: in <module>
    from _tracemalloc import *
E   ModuleNotFoundError: No module named '_tracemalloc'
```
I'm not familiar with pypy, so I assumed perhaps it excludes some built-in packages like tracemalloc. A quick google now indicates that tracemalloc should be available for pypy 3.11, so I'm unsure what actually went wrong there.

pyomo/contrib/appsi/tests/test_legacy_leak.py
Marl0nL: Done!

pyomo/contrib/appsi/base.py
Marl0nL: I've added the same fix to `pyomo.contrib.solver.common.base`. I haven't added a test case that covers that addition - unsure what the preference there is, but I lean towards not adding more duplication, on the assumption that these Legacy components will be consolidated or deprecated eventually.

pyomo/contrib/appsi/tests/test_legacy_leak.py
jsiirola: That makes sense.  I can confirm that it doesn't import for me with the current pypy release.  As a matter of style, we generally use the following:

```python

# Note: tracemalloc is not always available, e.g., under PyPy
tracemalloc, tracemalloc_available = attempt_import('tracemalloc')
```

pyomo/contrib/appsi/base.py
jsiirola: Following up here: I spoke with @michaelbynum: we think:
- both the `solutions.delete_symbol_map()` calls and the `solutions.add_symbol_map()` call above can be deleted.
- We can probably get rid of the `delete_legacy_soln` logic and just move:
  ```python
  legacy_results.solution.insert(legacy_soln)
  legacy_results._smap_id = id(symbol_map)
  legacy_results._smap = symbol_map
  ```
  into the `elif results.best_feasible_objective is not None:` block above.
- finally, have the `_smap_id` default to None:
  ```diff
  --- a/pyomo/contrib/appsi/base.py
  +++ b/pyomo/contrib/appsi/base.py
  @@ -1601,8 +1601,7 @@ class LegacySolverInterface:
           symbol_map.bySymbol = dict(self.symbol_map.bySymbol)
           symbol_map.aliases = dict(self.symbol_map.aliases)
           symbol_map.default_labeler = self.symbol_map.default_labeler
  -        model.solutions.add_symbol_map(symbol_map)
  -        legacy_results._smap_id = id(symbol_map)
  +        legacy_results._smap_id = None
 
           delete_legacy_soln = True
           if load_solutions:
  ```

pyomo/contrib/appsi/base.py
jsiirola: (see discussion in the next comment)

pyomo/contrib/appsi/base.py
Marl0nL: I've implemented this, and also removed the block below from contrib/solver/common/base.py, as I believe it should no longer be needed, as the only reference to `model.solutions` was the call to `model.solutions.add_symbol_map()`.
```
        if not hasattr(model, 'solutions'):
            # This logic gets around Issue #2130 in which
            # solutions is not an attribute on Blocks
            from pyomo.core.base.PyomoModel import ModelSolutions

            setattr(model, 'solutions', ModelSolutions(model))
```

jsiirola (approved):
(pending tests)

mrmundt (approved):

PR #3928, opened 2026-05-04
Add core symbolic gradient support to Pyomo.DoE

snarasi2:
## Fixes # .
This PR ports the core symbolic-gradient functionality from the historical pyomo-doe-symbolic work into the current pyomo.contrib.doe implementation.
## Summary/Motivation:

Rather than merging the old branch directly, this change transplants the symbolic DoE pieces onto current main so that symbolic sensitivities work with the newer DoE implementation already present in Pyomo, including the current objective and GreyBox-oriented code paths.

This PR ports the core symbolic-gradient functionality from adowling2/pyomo#7 onto current Pyomo main and adapts it to the current pyomo.contrib.doe implementation.

This contribution was prepared with coding assistance from OpenAI Codex. All design decisions, validation, testing, and quality-assurance responsibility remain with Shilpa Narasimhan.

- Add `GradientMethod` support to `DesignOfExperiments`
- Preserve the existing finite-difference workflow through the new gradient-method interface
- Add a symbolic / PyNumero gradient path for DoE
- Add an analytic FIM computation path for non-optimization FIM evaluation
- Export `ExperimentGradients` from `pyomo.contrib.doe`
- Refactor `ExperimentGradients` so symbolic and automatic differentiation are set up together
- Add supporting symbolic-gradient utilities for labeled experiment models
- Update DoE model construction so symbolic sensitivities can be used in the Jacobian-based DoE machinery
- Add a guard preventing `run_doe()` from being called with `GradientMethod.kaug`
- Add the polynomial example and polynomial-focused regression coverage
- Add broader symbolic-versus-automatic gradient consistency tests
- Add factorial-result dataframe tests
- Add reactor regression tests against expected solutions
- Clarify GreyBox `cyipopt` / MA57-HSL test behavior where relevant

Validation performed locally:
- Black: `python -m black -S -C --check --diff pyomo/contrib/doe` → passed
- Typos: `typos --config .github/workflows/typos.toml pyomo/contrib/doe DOE_SYMBOLIC_PR_NOTES.md` → passed
- Focused DoE/GreyBox validation: `134 passed, 0 failed, 0 skipped`
- GreyBox `cyipopt` tests were run with MA57/HSL available in the local environment
- The existing K_AUG-dependent DoE test path was validated locally after configuring the required runtime library path

### Legal Acknowledgement


snarasi2:
@adowling2 @slilonfe5 @smondal13 @sscini This is the new symbolic PR (created from the unpolluted branch)

snarasi2:
@blnicho — this PR (#3928) is the clean replacement for #3898. The original PR included unintended extra changes; this one contains only the intended symbolic DoE work and has been fully validated.

Could you please switch review/merge to this PR and we can close #3898?

mrmundt:
> @blnicho — this PR (#3928) is the clean replacement for #3898. The original PR included unintended extra changes; this one contains only the intended symbolic DoE work and has been fully validated.
> 
> Could you please switch review/merge to this PR and we can close #3898?

@snarasi2 - Done. #3898 is closed in favor of this one.

pyomo/contrib/doe/doe.py
```diff
@@ -927,6 +1048,12 @@ def create_doe_model(self, model=None):
             ]
         )
 
+        experiment_grad = None
+        if self._gradient_method == GradientMethod.pynumero:
+            experiment_grad = ExperimentGradients(
+                model.scenario_blocks[0], automatic=True, symbolic=True
```
snarasi2, line 1057: @adowling2 @smondal13 , per discussion with Shuvo on 05/11: Shuvo and I decided that the cleanest path forward to address potential conflicts with Shuvo's multi-experiment PR #3866 is to leave symbolic PR as is for now. When PR3866 is merged, we will address the conflict with fd_scenario_blocks and fd_scenarios within multi-experiment vs scenatio_blocks and scenarios within the current Pyomo main and the symbolic PR.

Alternative would be to merge this PR with multi-experiment as of now, but it will create a dependency and we are not sure we should proceed this way.

snarasi2: Shilpa to double-check and address: If main does not use model.scenarios currently, @snarasi2 will modify all occurrences of model.scenarios to to model.fd_scenarios to facilitate multi-experiment merge. Additionally, this will help distinguish scenario building for finite difference methods for parameter perturbations vs parameter scenarios for uncertainty-aware design.

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -312,15 +335,20 @@ def test_rooney_biegler_obj_cholesky_solve(self):
         # Note: When using prior_FIM, the relationship FIM = Q.T @ sigma_inv @ Q + prior_FIM
         self.assertTrue(np.all(np.isclose(FIM, Q.T @ sigma_inv @ Q + prior_FIM)))
 
-    def DISABLE_test_reactor_obj_cholesky_solve_bad_prior(self):
+    # This legacy Cholesky/bad-prior case is kept disabled in
+    # this PR. It is not part of the active regression signal, and this cleanup is
+    # focused on replacing active general-purpose coverage with
+    # Rooney-Biegler or polynomial examples rather than rewriting inactive,
+    # branch-specific tests.
+    def DISABLE_test_rooney_biegler_obj_cholesky_solve_bad_prior(self):
```
mrmundt: Maybe this should be deleted? It's been disabled for a while, and it'll still live in the commit history.

pyomo/contrib/doe/tests/test_greybox.py
```diff
@@ -373,11 +335,20 @@ def make_greybox_and_doe_objects_rooney_biegler(objective_option):
...
-            bad_message in doe_object.results["Termination Message"]
-        )
-    except:
+        termination_message = str(doe_object.results.get("Termination Message", ""))
+        cyipopt_call_working = bad_message not in termination_message
+        if not cyipopt_call_working:
+            # The GreyBox path here uses cyipopt rather than the standalone
+            # IPOPT executable, and the local failure mode we diagnosed was a
+            # missing MA57/HSL runtime on that cyipopt solve path.
+            cyipopt_skip_reason = (
```
mrmundt, line 342: This is a nitpick. All of the skip reasons are variations of the same thing (MA57/HSL issues). Do you really need all three different messages? Or can you just standardize to the one that's on line 315?

pyomo/contrib/doe/tests/test_utils.py
```diff
@@ -154,5 +174,196 @@ def test_get_FIM_metrics(self):
         )
 
 
+@unittest.skipIf(not ipopt_available, "The 'ipopt' command is not available")
+@unittest.skipIf(not numpy_available, "Numpy is not available")
```
mrmundt: Aren't these numpy_available checks redundant (because of lines 29-30)?

pyomo/contrib/doe/tests/test_utils.py
```diff
@@ -154,5 +174,196 @@ def test_get_FIM_metrics(self):
...
+                    1.0 - np.exp(-rate_constant * hour),
+                    asymptote * hour * np.exp(-rate_constant * hour),
+                ]
+            ]
+        )
+
+        self.assertEqual(jacobian.shape, expected.shape)
+        self.assertTrue(np.allclose(jacobian, expected, atol=1e-7, rtol=1e-7))
+
+    @unittest.skipIf(not scipy_available, "scipy is not available")
```
mrmundt: Similar comment - isn't this scipy_available check redundant?

pyomo/contrib/doe/doe.py
```diff
@@ -283,6 +312,12 @@ def run_doe(self, model=None, results_file=None):
                       default: None --> don't save
 
         """
+        if self._gradient_method == GradientMethod.kaug:
```
mrmundt, line 318: Maybe I am misunderstanding - `kaug` is listed as an acceptable option in the docstring, but it's not actually?

pyomo/contrib/doe/doe.py
snarasi2: @smondal13 @adowling2 I checked upstream/main and confirmed that literal model.scenarios already exists in pyomo/contrib/doe/doe.py and pyomo/contrib/doe/tests/test_doe_build.py. Since the proposed rename was conditional on main not currently using model.scenarios, I did not rename model.scenarios to model.fd_scenarios in this PR. I’m leaving this deferred to the multi-experiment merge as discussed on 05/11.

pyomo/contrib/doe/tests/test_doe_solve.py
snarasi2: Removed the disabled Rooney-Biegler Cholesky/bad-prior test based on Miranda’s review comment and alignment from @slilonfe5 and @smondal13 on Slack. Commit: 5ad99f008

pyomo/contrib/doe/tests/test_greybox.py
snarasi2: Thanks — standardized the MA57/HSL-related skip reasons to use the same message as line 315 in ab2f729ef.

pyomo/contrib/doe/tests/test_utils.py
snarasi2: Thanks — removed the redundant NumPy skip decorators in `test_utils.py` because the file-level `SkipTest` already handles missing NumPy in f1c9a8f3b.

pyomo/contrib/doe/tests/test_utils.py
snarasi2: Thanks — removed the redundant SciPy skip decorator in `test_utils.py` because the file-level `SkipTest` already handles missing SciPy in dcbfee6a4.

snarasi2: @mrmundt, @adowling2  I also noticed the same pattern in test_greybox.py: the file has a file-level SkipTest for NumPy/SciPy, but the TestFIMExternalGreyBox class still has NumPy/SciPy skipIf decorators. Would you prefer that I keep this PR scoped to the test_utils.py comments, or should I also remove the redundant decorators within test_greybox.py and similar files for consistency?

pyomo/contrib/doe/doe.py
snarasi2: Thanks — I checked the code path more closely. `kaug` is supported for FIM computation through `compute_FIM()`, but `run_doe()` explicitly rejects `GradientMethod.kaug` because it is not supported for DoE optimization. I clarified that scope in the `gradient_method` docstring in 1096e52d8.

snarasi2:
Hi @mrmundt , I addressed the latest review comments and pushed the updates.

Summary of changes:

* Removed the disabled Rooney-Biegler Cholesky test, since it has been disabled for a while and remains available in Git history.
* Clarified the `gradient_method` docstring: `kaug` is supported for FIM computation through `compute_FIM()`, but not for DoE optimization through `run_doe()`.
* Updated the DoE optional dependency imports in `doe.py`, `test_utils.py`, and `test_greybox.py` to use `attempt_import(..., defer_import=True)`.
* Applied Black formatting after the import update.

Local checks:

* `black -S -C --check pyomo/contrib/doe/doe.py pyomo/contrib/doe/tests/test_greybox.py pyomo/contrib/doe/tests/test_utils.py` passed.
* `git diff --check` passed.
* `python -m compileall pyomo/contrib/doe/doe.py pyomo/contrib/doe/tests/test_greybox.py pyomo/contrib/doe/tests/test_utils.py` passed.
* The targeted changed-file tests pass fully: `test_utils.py` + `test_greybox.py`, `48 passed`.
* I also reran the full DoE test suite in the HSL-enabled environment. The full suite:`134 passed`.

One CI job failed during Windows environment setup, before the Pyomo tests ran. The failure appears to be a conda/libmamba segmentation fault while installing dependencies:

`Segmentation fault "$CONDA_EXE" "$@"`
`Error: Process completed with exit code 139`

This does not look related to the PR changes. Could this job please be rerun?


I think the PR is otherwise ready for you to re-review. Could you please re-review when you have a chance?

pyomo/contrib/doe/tests/test_utils.py
snarasi2: Update, one more follow-up: after @adowling2  confirmed the broader cleanup was okay, I removed the remaining redundant downstream NumPy/SciPy skip decorators in the DoE tests and added comments noting that NumPy/SciPy availability is checked by the file-level `SkipTest`.

Checks run:

* No remaining downstream `skipIf(not numpy_available)` / `skipIf(not scipy_available)` decorators in `pyomo/contrib/doe/tests`.
* Black passed: `conda run -n base black -S -C --check pyomo/contrib/doe/tests`.
* `git diff --check` passed.
* `python -m compileall pyomo/contrib/doe/tests` passed.
* Full DoE test suite passed: `134 passed, 5 warnings`.

Latest commit: `ba6d769c6`.

pyomo/contrib/doe/tests/test_utils.py
```diff
@@ -154,5 +174,196 @@ def test_get_FIM_metrics(self):
...
+            .ravel()
+            .astype(float)
+        )
+        base_values = {"a": 2.0, "b": -1.0, "c": 0.5, "d": -1.0}
+        step = 1e-6
+        finite_difference = []
+        for parameter in ("a", "b", "c", "d"):
+            forward_values = dict(base_values)
+            backward_values = dict(base_values)
+            forward_values[parameter] += step
```
slilonfe5: @snarasi2 I believe you wanted to implement a relative perturbation (which is what we do in ParmEst and Pyomo.DoE). I suggest you define a `rel_perturbation = step * base_values[parameter]` before the `forward_values` line. Then replace all the `step` with `rel_perturbation`.

pyomo/contrib/doe/tests/test_utils.py
```diff
@@ -154,5 +174,196 @@ def test_get_FIM_metrics(self):
...
+        model = experiment.get_labeled_model()
+        experiment_gradients = ExperimentGradients(model, symbolic=True, automatic=True)
+
+        symbolic = (
+            experiment_gradients.compute_gradient_outputs_wrt_unknown_parameters()
+            .ravel()
+            .astype(float)
+        )
+        base_values = {"a": 2.0, "b": -1.0, "c": 0.5, "d": -1.0}
+        step = 1e-6
```
slilonfe5: Also, the step is too small. I suggest 1e-3

pyomo/contrib/doe/examples/polynomial.py
```diff
@@ -0,0 +1,95 @@
...
+
+        y = a*x1 + b*x2 + c*x1*x2 + d
+        """
+
+        m = self.model = pyo.ConcreteModel()
+
+        m.x1 = pyo.Var(bounds=(-5, 5), initialize=2.0)
+        m.x2 = pyo.Var(bounds=(-5, 5), initialize=3.0)
+
+        m.a = pyo.Var(bounds=(-5, 5), initialize=2)
```
slilonfe5, line 40: It appears a, b, c, and d are parameters. I think these should be fixed after defining them. E.g., after you defined `m.a = pyo.Var(bounds=(-5, 5), initialize=2)`, you should add a `m.a.fix()` to the new line after the definition. Similar for m.b, m.c, and m.d.

pyomo/contrib/doe/tests/test_utils.py
snarasi2: Thanks, modified as suggested

pyomo/contrib/doe/tests/test_utils.py
snarasi2: Thanks, modified as suggested

pyomo/contrib/doe/examples/polynomial.py
snarasi2: Thanks, modified as suggested

snarasi2:
@slilonfe5 @mrmundt  I addressed the review comments by fixing the polynomial model parameters after declaration, updating the finite-difference check to use relative perturbations, and increasing the step size to 1e-3.

Checks passed locally:

git diff --check
Black check
compileall
Full DoE test suite: 134 passed

Latest commit: 512451e25.

slilonfe5:
@mrmundt, this also looks pretty good. We will appreciate any final comments; otherwise, I think this is ready to be merged.

DOE_SYMBOLIC_PR_NOTES.md
mrmundt, line 1: This file needs to be removed before this PR can be approved.

snarasi2:
@adowling2 @sscini @slilonfe5 @smondal13 Adding context here: the markdown file explaining the math was included based on earlier internal review feedback from April 8th. Since the Sandia team would prefer that the `.md` file not be included, should I remove it entirely from the PR, or would you prefer that context be captured elsewhere (for example, in the PR description or inline comments)?

adowling2:
Describing the math in the PR text is not a bad idea, but also not necessary IMO. The markdown file will still be in the git history.

pyomo/contrib/doe/tests/test_doe_build.py
```diff
@@ -510,21 +525,26 @@ def test_generate_blocks_without_model(self):
             )
 
 
-class TestReactorExample(unittest.TestCase):
-    def test_reactor_update_suffix_items(self):
-        """Test the reactor example with updating suffix items."""
-        from pyomo.contrib.doe.examples.update_suffix_doe_example import main
+class TestRooneyBieglerExample(unittest.TestCase):
```
adowling2, line 528: I am a little confused why this PR changes these files. I thought we already swapped out the reactor example for the consolidated RooneyBiegler example in a previous PR. My practical recommendation is to merge https://github.com/Pyomo/pyomo/pull/3866 first, as it is really close and it is the last active Pyomo.DoE PR besides this current PR. After we merge @smondal13's PR, we can look at what this PR is changing relative to the new `main`.

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -425,43 +432,191 @@ def test_compute_FIM_seq_backward(self):
...
+        self.assertTrue((len(set(x1_vals)) == 2) and (len(set(x2_vals)) == 2))
 
-        # assert unique values are correct
+        # Check that each polynomial design variable spans the requested grid values.
+        self.assertTrue(
+            (set(x1_vals).issuperset(set([0, 5])))
+            and (set(x2_vals).issuperset(set([0, 5])))
+        )
+
+    def test_rooney_biegler_run_doe_pynumero(self):
```
adowling2, line 467: @blnicho @mrmundt PyNumero now provides an alternative to the finite difference approximation. That means for Pyomo.DoE, the user gets to choose:

Sensitivity calculation method: finite_difference or PyNumero

Objective: trace (Cholseky), det (Cholseky), pseudo-trace, trace (GreyBox), det (GreyBox), min_eigenvalue (GreyBox), condition_number (GreyBox), pseudo-trace (GreyBox -- I think, but I could be wrong)

This leaves 2 times 8 = 16 different reasonable configurations for Pyomo.DoE. Is there a preferred way to test all 16 of these configurations with the RooneyBiegler example in an "extended" or "long runtime" test suite? Is this something that makes the most sense for a parameterized test?

I think this PR is adding tests for the PyNumero for many of these objectives. I want to test this with RooneyBiegler, but I also appreciate if the everyday Pyomo testing suite is getting bloated.

@slilonfe5 @smondal13 @sscini Did we decide on a similar "long runtime" or "extended" test suite for ParmEst? Or is this just on the todo list?

pyomo/contrib/doe/doe.py
adowling2: Once we are happy with this feature, I want to deprecated `k_aug` in both Pyomo.DoE and ParmEst. This will be a future PR. The new `GradientMethod` should provide us with sufficient infrastructure to replicate with `k_aug` is doing in both DoE and ParmEst. That then means one less solver to maintain.

@slilonfe5 fyi, another thing to think about for the fall Pyomo release

pyomo/contrib/doe/doe.py
```diff
@@ -611,6 +658,79 @@ def compute_FIM(self, model=None, method="sequential"):
...
+            dsdp_re, col = get_dsdp(model, params_names, params_dict, tee=self.tee)
+            dsdp_array = dsdp_re.toarray().T
+
+            dsdp_extract = []
+            for k, _ in model.experiment_outputs.items():
+                name = k.name
+                try:
+                    kaug_no = col.index(name)
+                    dsdp_extract.append(dsdp_array[kaug_no])
+                except Exception:
```
adowling2: We should avoid `except Exception` if possible.

pyomo/contrib/doe/doe.py
adowling2: I think this is a good plan. Also, this PR changes tests outside of the new symbolic/automatic differentiation capabilities. I am curious if any of these changes are also being made in @smondal13's PR.

adowling2 (commented):
In my mind, there are two important issues to discuss as a team, possibly at the next Pyomo dev call:
- [ ] Why are some tests outside the symbolic feature getting changed in this PR? Are these intentional changes to improve code coverage, etc., or an artifact of help from Codex? I think we have had a few small group conversations quickly about this as a team, but not with the code pulled up, just in the abstract.
- [ ] Do we want to add RooneyBiegler tests for (almost) all of the objective options with the PyNumero derivatives? If yes, how should this be added to the testing infrastructure? Do we want to tag these as "long" or "extended" tests? Should we use parameterized tests?

pyomo/contrib/doe/tests/test_doe_build.py
smondal13: @snarasi2 I am really confused. Why are you adding this test? Is it doing something symbolic is supposed to do? `update_model_from_suffix` was a function @sscini added last year, and I added tests for RB.

pyomo/contrib/doe/tests/test_doe_solve.py
smondal13: @adowling2, the decision was for `Pyomo.DoE`. I think for `Parmest` that is on `TODO` list. @sscini @slilonfe5 Correct me if I am wrong

pyomo/contrib/doe/doe.py
```diff
@@ -31,14 +31,14 @@
 import logging
 import math
 
-from pyomo.common.dependencies import (
```
smondal13, line 34: @mrmundt @adowling2 I think the imports from `pyomo.common.dependencies` are deferred imports. So, is it necessary to use `attempt_import` specifically? Is there a preferred style choice between these two? I am curious, and this will help me in the future.

pyomo/contrib/doe/tests/test_doe_build.py
smondal13: @snarasi2 , I think the class name here is misleading

pyomo/contrib/doe/tests/test_doe_errors.py
```diff
@@ -503,7 +504,91 @@ def test_reactor_figure_drawing_bad_sens_names(self):
...
+
+        experiment = PolynomialExperiment()
+
+        DoE_args = get_standard_args(experiment, fd_method, obj_used, flag=None)
+        DoE_args["gradient_method"] = "pynumero"
+        DoE_args["scale_nominal_param_value"] = False
+
+        doe_obj = DesignOfExperiments(**DoE_args)
+
+        synthetic_results = {
```
smondal13: since this is being used multiple times, why don't you make this a function ?

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -425,43 +432,191 @@ def test_compute_FIM_seq_backward(self):
         doe_obj.compute_FIM(method="sequential")
 
     @unittest.skipIf(not pandas_available, "pandas is not available")
```
smondal13, line 435: Is `pandas` required for polynomial?

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -425,43 +432,191 @@ def test_compute_FIM_seq_backward(self):
...
+
+        FIM, Q, L, sigma_inv = get_FIM_Q_L(doe_obj=doe_obj)
+        self.assertTrue(np.all(np.isclose(FIM, L @ L.T)))
+        self.assertTrue(np.all(np.isclose(FIM, Q.T @ sigma_inv @ Q + prior_FIM)))
+
+    def test_rooney_biegler_run_doe_determinant_regression(self):
+        """Check a stable Rooney-Biegler optimum fingerprint against expected values."""
+        experiment = get_rooney_biegler_experiment()
+        DoE_args = get_standard_args(experiment, "central", "determinant")
+        DoE_args["gradient_method"] = "pynumero"
```
smondal13, line 495: I think we have tests for the ' sequential` method. So, it would make more sense to use `parameterized` tests for maintainability. @adowling2 also mentioned something like that.

pyomo/contrib/doe/tests/test_doe_solve.py
```diff
@@ -425,43 +432,191 @@ def test_compute_FIM_seq_backward(self):
...
+        self.assertAlmostEqual(fim[0, 0], 41155.59271917, places=2)
+        self.assertAlmostEqual(fim[1, 1], 973.06126181, places=2)
+        self.assertAlmostEqual(
+            doe_obj.results["log10 D-opt"], 7.179982499524086, places=4
+        )
+        self.assertAlmostEqual(
+            doe_obj.results["log10 A-opt"], -2.5554049159721415, places=4
+        )
+
+    def test_rooney_biegler_run_doe_pynumero_objective_matrix(self):
```
smondal13: same here

pyomo/contrib/doe/tests/test_doe_solve.py
slilonfe5: @smondal13 @adowling2 The ParmEst test file is very robust. We have already implemented something like that in the past, where we used parameterization to test 24 configurations for parameter and covariance matrix estimation using the Rooney-Biegler example. The current set of parameterized tests in Pyomo.DoE were inspired by the infrastructure we have implemented in ParmEst.

pyomo/contrib/doe/doe.py
slilonfe5: @adowling2 my understanding is that we want to move the `k_aug` FIM method to the `pynumero` method because we currently have the option to set `automatic=True` when using `self._analytic_FIM` to compute the FIM for the `pynumero` method. And we are not able to currently do this because irrespective of what the `automatic` and `symbolic` arguments of `self._analytic_FIM` is, it returns the FIM for symbolic because of this code in `ExperimentGradients`: `if automatic or symbolic:`.

pyomo/contrib/doe/tests/test_doe_solve.py
adowling2: Excellent. I think we want to further extend the parameterized tests in Pyomo.DoE to cover the symbolic option for derivatives.

pyomo/contrib/doe/doe.py
adowling2: @slilonfe5 Not exactly. In a future PR (not this one), I want to extend `ExperimentGradients` to support computing numerical derivatives (measurement sensitivities with respect to parameters) using `pynumero`. Once we do that, `k_aug` will become obsolete and we can remove it from both `parmest` and `DoE`. You are correct that this will require some changes to `ExperimentGradients`.

blnicho:
This is waiting on #3866

pyomo/contrib/doe/tests/test_doe_build.py
smondal13: After merging and reviewing merge conflicts, decide if this test should be moved to `test_utils.py`

DOE_SYMBOLIC_PR_NOTES.md
snarasi2: Removed in 70f7a596e. Thanks!

pyomo/contrib/doe/tests/test_doe_solve.py
snarasi2: I checked this. The pandas skip is currently attached at the `TestDoeBuild` class level. Most of that class uses the Rooney-Biegler helper, which constructs the test data with `pd.DataFrame(...).iloc[0]`, so the class-level pandas guard is still needed. The polynomial tests inside that class do not appear to use pandas directly, but I would prefer not to split the class only for that unless you want that change. Marking @adowling2  too

pyomo/contrib/doe/doe.py
snarasi2: Thank you, I replaced the except Exception with except ValueError, so this only handles the expected “output name not found” case from the column lookup instead of suppressing unrelated errors. I also checked for similar DOE usage. The other DOE-side explicit `ValueError` handler is in `pyomo/contrib/doe/tests/test_utils.py` at line 48, where the test intentionally catches a `ValueError` as the expected failure mode. Addressed in e6ce422d7.

pyomo/contrib/doe/tests/test_doe_errors.py
snarasi2: I found two occurrences of the duplicated synthetic-results dictionary in test_doe_errors.py, both in polynomial plotting error tests. I factored the duplicated dictionary into a helper and reused it in both places. The targeted error tests pass, and the full DOE test suite passes locally with 134 passed and 5 plotting-backend warnings. Addressed in d9b7c5514.

pyomo/contrib/doe/doe.py
snarasi2: @adowling2 @mrmundt This comment was made on a deleted line. I updated this import block to use explicit `attempt_import(..., defer_import=True)` calls for NumPy, pandas, matplotlib, and SciPy.

This was my understanding of the May 6 discussion item about deferring imports of external packages such as NumPy, SciPy, and pandas and checking `attempt_import` as the Pyomo convention for doing that. If this is not the preferred style, or if the existing `pyomo.common.dependencies` import form was already the intended deferred-import style, please let me know and I can adjust.

pyomo/contrib/doe/tests/test_doe_solve.py
snarasi2: I looked at this one separately. This test is a PyNumero-gradient-method determinant regression: `fd_formula="central"`, `objective_option="determinant"`, and `gradient_method="pynumero"`. I kept it separate because the assertions are a determinant-specific numerical fingerprint of the optimal design, FIM entries, and objective metrics.

The repeated symbolic objective cases below were parameterized in fc9e96634. I did not parameterize this regression test because adding other objective/gradient-method cases here would require separate expected fingerprints for each configuration.

pyomo/contrib/doe/tests/test_doe_solve.py
snarasi2: Addressed in fc9e96634. I converted this PyNumero/symbolic Rooney-Biegler objective matrix from an internal `subTest` loop to `parameterized.expand(...)`, so the `trace`, `determinant`, and `zero` objective cases are now collected as separate symbolic `run_doe()` tests.

Checks run:

* targeted symbolic objective cases: `3 passed`
* Rooney-Biegler solving class with `k_aug`: `23 passed`
* full DoE test suite: `136 passed, 5 warnings`

PR #3979, opened 2026-06-19
KNITROAMPL: use \$KNITRODIR/knitroampl/knitroampl for executable

yemine0x01:
## Fixes # .

NA

## Summary/Motivation:

Some `KNITRO` users install it as a standalone package (not via the Python knitro module) and rely on the `KNITRODIR` environment variable to locate their installation. Previously, `_default_executable()` only checked the knitro Python package path first, then fell back to a plain PATH search, meaning users with a `KNITRODIR`-based installation but no Python package would only succeed if `knitroampl` happened to be on their PATH.

This PR adds an explicit lookup for `$KNITRODIR/knitroampl/knitroampl` as the highest-priority candidate, reflecting the standard layout of a standalone KNITRO installation.

-
-

## AI-Use Disclosure

or

        - [ ] Added and ran tests
    
AI was used only for generating the commit message.

### Legal Acknowledgement

### DIFF
--- a/pyomo/solvers/plugins/solvers/KNITROAMPL.py
+++ b/pyomo/solvers/plugins/solvers/KNITROAMPL.py
@@ -7,6 +7,7 @@
 # software.  This software is distributed under the 3-clause BSD License.
 # ____________________________________________________________________________________
 import logging
+import os
 
 from pyomo.common import Executable
 from pyomo.common.collections import Bunch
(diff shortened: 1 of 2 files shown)


jsiirola (commented):
This looks reasonable; however, should the order in which paths are tested be reversed?  That is, if the `knitro` module is available, we should probably use the `knitroampl` from it (for consistency) before falling back on KNITRODIR?

yemine0x01:
Thanks for the review @jsiirola. I checked with the KNITRO dev team on the ordering, and they'd rather keep `KNITRODIR` first, ahead of the `knitro` module. Their reasoning is that if a user has set `KNITRODIR`, that's a deliberate pointer to the install they want used, so it should win. Does that line up with how you're thinking about it?

blnicho (approved):

mrmundt (approved):

PR #3982, opened 2026-06-22
Simple disjunctive transform

cjohnston1:
## Summary/Motivation:
This PR is a draft meant to facilitate conversation and is not ready to be reviewed and committed.

This PR is a draft of the simple disjunctive transformation module. This creates a new transformed disjunction which has a single inequality per disjunct for either user given disjunctions or all disjunctions on a model if no input is given. The transformation can be done in multiple ways which are selected by the user. This is motivated by the development of another module which can generate a family of inequalities for these transformed disjunctive constraints.

- Adds the simple disjunction transform module
- Adds tests for the simple disjunction transform module.

## AI-Use Disclosure

or

        - [ ] Added and ran tests
    
### Legal Acknowledgement


pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,539 @@
...
+            )
+    result = ComponentMap()
+    for disjunct, value in items:
+        result[disjunct] = _as_constraint_list(value)
+    return result
+
+
+@TransformationFactory.register(
+    'gdp.simple_disjunction',
+    doc="Relax selected Disjunctions by building, for each one, a 'simple' "
```
emma58: This transformation is not a relaxation, I don't think. I think I'd call this "reformulation"

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,539 @@
...
+    "Disjunction whose Disjuncts each retain a single Constraint derived from "
+    "the corresponding original Disjunct.",
+)
+class SimpleDisjunctionTransformation(Transformation):
+    """Create a relaxation of one or more Disjunctions as *simple* Disjunctions.
```
emma58: Same here

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,539 @@
...
+
+        build_expression = getattr(self, self._EXPRESSION_BUILDERS[method])
+        for disjunction in self._get_disjunctions_to_transform(
+            instance, config.targets
+        ):
+            self._transform_disjunction(
+                disjunction, build_expression, selected_constraints
+            )
+
+    def _get_disjunctions_to_transform(self, instance, targets):
```
emma58: I'd vote for not making this a method and just moving the logic it into the method above: You can just transform Disjunctions as you find them and not have to build a separate list of them (which could be quite long for big models.)

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,539 @@
...
+                        continue
+                    disjunctions.extend(self._gather_disjunctions(block))
+            else:
+                raise GDP_Error(
+                    "Target '%s' was not a Block or Disjunction. It was of type "
+                    "%s and can't be transformed." % (t.name, type(t))
+                )
+        return disjunctions
+
+    def _gather_disjunctions(self, block):
```
emma58: You can get this from the GDP Tree when you switch to it.

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,539 @@
...
+                % disjunction.name
+            )
+        if _parent_disjunct(disjunction) is not None:
+            raise GDP_Error(
+                "Disjunction '%s' is nested in another Disjunct. This "
+                "transformation does not create simple disjunctions from nested "
+                "Disjunctions." % disjunction.name
+            )
+        nested = self._nested_disjunction_owner(disjunction)
+        if nested is not None:
```
emma58: Let's move all this logic to use the GDP Tree.

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,539 @@
...
+                    "Disjunct '%s' has no active constraints to select, so it "
+                    "is skipped in the simple disjunction generated from "
+                    "Disjunction '%s'." % (disjunct.name, disjunction.name)
+                )
+                continue
+            chosen.append((disjunct, expression))
+
+        if not chosen:
+            raise GDP_Error(
+                "Cannot create a simple disjunction from Disjunction '%s': none "
```
emma58, line 375:
```suggestion
                "Cannot create a simple disjunction from Disjunction '%s': None "
```

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,539 @@
...
+            new_disjunct.constraint = Constraint(expr=expression)
+            new_disjuncts.append(new_disjunct)
+
+        trans_block.simple_disjunction = Disjunction(expr=new_disjuncts)
+
+        # Record the mapping in both directions so callers can move between an
+        # original Disjunction and the simple Disjunction generated from it.
+        disjunction._transformation_map[self.transformation_name] = (
+            trans_block.simple_disjunction
+        )
```
emma58: This was super astute of you to notice this map, I had forgotten about it! 

@jsiirola, in the current implementation, the public API to map from original to transformed is using this. I would definitely have stored it separately on the MFE block, but I'm thinking this is actually better. Do you agree?

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,539 @@
...
+            if constraint not in own:
+                raise GDP_Error(
+                    "The constraint '%s' selected for Disjunct '%s' is not one "
+                    "of that Disjunct's own active constraints. (Constraints "
+                    "inside a nested Disjunct cannot be selected.)"
+                    % (constraint.name, disjunct.name)
+                )
+
+    @staticmethod
+    def _own_active_constraints(disjunct):
```
emma58, line 498: I'd suggest `_get_active_constraints` here. Names of methods/functions should almost always be verb phrases. This is...but it doesn't quite mean what it's doing.

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,539 @@
...
+            )
+
+        parent_block = disjunction.parent_block()
+        trans_block = Block()
+        parent_block.add_component(
+            unique_component_name(
+                parent_block, '_pyomo_gdp_simple_disjunction_reformulation'
+            ),
+            trans_block,
+        )
```
emma58: Right now this means that you will create a new transformation block for every Disjunction. This is a memory-hungry way to do it for large models that might have indexed disjunctions with many disjunctions in them. I would instead keep track of where you have already created transformation blocks and only create a new one if you need it. Then on each transformation block, make just one Disjunction component to hold simple disjunctions, indexed by Any (or NonNegativeIntegers or something). Since you have the API to map between anyway, you can just make this part compact, basically. The other GDP transformations all follow this model, so you can look elsewhere for examples.

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,539 @@
...
+            new_disjuncts.append(new_disjunct)
+
+        trans_block.simple_disjunction = Disjunction(expr=new_disjuncts)
+
+        # Record the mapping in both directions so callers can move between an
+        # original Disjunction and the simple Disjunction generated from it.
+        disjunction._transformation_map[self.transformation_name] = (
+            trans_block.simple_disjunction
+        )
+        trans_block._src_disjunction = weakref_ref(disjunction)
```
emma58: This will break when you make the change I suggested above since you'll have multiplied transformed Disjunctions to a Block. Can you redo the APIs to use private data classes like the other transformations? You will want to define a class that will hold the mappings, so it will have one dictionaries in your case: to get from transformed to original. It should inherit from `AutoSlots.Mixin` for memory's sake. Then you will want to call `Block.register_private_data_initializer(_YourMappingDataClass)` in this file. See the bigm transformation (and/or the GDP-to-MIP base class) for an example.

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,539 @@
...
+        that aggregate several Constraints into one will perform that reduction
+        here (for example, by combining the Constraint bodies into a new
+        expression).
+        """
+        if len(sources) != 1:
+            raise GDP_Error(
+                "Disjunct '%s' (in Disjunction '%s') was assigned %d constraints, "
+                "but the current selection methods keep exactly one constraint "
+                "per Disjunct. (Aggregating several constraints into one is not "
+                "yet implemented.)" % (disjunct.name, disjunction.name, len(sources))
```
emma58, line 471: This error should be unreachable if you have coded everything correctly, right? I'm not sure you need to check for it.

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,539 @@
...
+    def _validate_selected_sources(self, disjunction, disjunct, sources):
+        own = ComponentSet(self._own_active_constraints(disjunct))
+        for constraint in sources:
+            if not isinstance(constraint, ConstraintData):
+                raise GDP_Error(
+                    "An object selected for Disjunct '%s' in "
+                    "'selected_constraints' is not a Constraint. Expected a "
+                    "ConstraintData, but got an object of type %s."
+                    % (disjunct.name, type(constraint).__name__)
+                )
```
emma58: In the spirit of duck-typing, I think you can just assume it's a Constraint until it fails to act like one.

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,539 @@
...
+        """
+        trans_block = simple_disjunction.parent_block()
+        src = getattr(trans_block, '_src_disjunction', None)
+        if type(src) is not weakref_ref:
+            raise GDP_Error(
+                "It appears that '%s' is not a simple disjunction generated by "
+                "the 'gdp.%s' transformation. No source disjunction found."
+                % (simple_disjunction.name, self.transformation_name)
+            )
+        return src()
```
emma58: I would neither use a weakref here nor validate what you got... If it was there, it should be the right thing, so in this case if a user hits this error, you are complaining to them about your bug. You should however, probably raise a helpful error if you don't find `simple_disjunction` in your map. You can look at other transformations for an example of that.

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,539 @@
...
+        simple = src_disjunction._transformation_map.get(self.transformation_name)
+        if simple is None:
+            raise GDP_Error(
+                "Disjunction '%s' has not been transformed with the 'gdp.%s' "
+                "transformation, so it has no simple disjunction."
+                % (src_disjunction.name, self.transformation_name)
+            )
+        return simple
+
+    def get_src_disjunction(self, simple_disjunction):
```
emma58, line 527: You'll need to update this function when you switch to private data classes.

pyomo/gdp/tests/test_simple_disjunction_transform.py
```diff
@@ -0,0 +1,425 @@
...
+from pyomo.common.collections import ComponentMap, ComponentSet
+from pyomo.common.log import LoggingIntercept
+from pyomo.core.expr import identify_variables
+from pyomo.core.expr.compare import assertExpressionsEqual
+from pyomo.environ import Block, ConcreteModel, Constraint, TransformationFactory, Var
+from pyomo.gdp import Disjunct, Disjunction, GDP_Error
+from pyomo.gdp.plugins import simple_disjunction_transform
+
+
+class CommonModels:
```
emma58: There are a bunch of GDP models for testing in `tests/models.py`. I was trying at one point to avoid a proliferation of very similar models... I don't really have strong feelings about it, but your nested one might actually be very similar to some that are already there. If it's hard to switch don't worry, but maybe put these in that file.

pyomo/gdp/tests/test_simple_disjunction_transform.py
```diff
@@ -0,0 +1,425 @@
...
+    # 'first' constraint selection (the default)                         #
+    # ------------------------------------------------------------------ #
+    def test_first_method_builds_simple_disjunction(self):
+        m = self.make_two_term_model()
+        TransformationFactory('gdp.simple_disjunction').apply_to(m)
+
+        trans = self.get_transformation_block(m)
+        self.assertIsNotNone(trans)
+        simple = trans.simple_disjunction
+        self.assertIsInstance(simple, Disjunction.__mro__[0])
```
emma58: `Disjunction.__mro__[0]` is the class itself, I think. This should just be `Disjunction`.

emma58 (changes requested):
@cjohnston1 , this is looking good! A couple more major things to change though:
0) Please switch to using the GDP Tree for all the nested cases--that should completely eliminate a lot of your methods for handling nested structures and simplify some other things too.
1) Can you please use a private data block for the mapping from the transformed to the original model? Look to bigm for an example of that.
2) Only create one transformation block per parent block and put multiple transformed Disjunctions on the transformation blocks.

After you do this, I can take another look.

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,544 @@
...
+    Holds the mapping from a generated (simple) Disjunction back to the
+    original Disjunction it was built from. The map holds the components
+    directly (a hard reference), so callers can always recover the source
+    Disjunction from a simple one via ``get_src_disjunction``.
+    """
+
+    __slots__ = ('src_disjunction',)
+
+    def __init__(self):
+        self.src_disjunction = ComponentMap()
```
emma58, line 40: Unless you need this to be ordered, you can just use a dictionary, since you're mapping Disjunctions to each other (and they're hashable). It'll be a little faster.

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,544 @@
...
+    structure ready for selection methods that aggregate several Constraints
+    into one.
+    """
+    if isinstance(value, ConstraintData):
+        return [value]
+    if isinstance(value, ComponentBase) and value.ctype is Constraint:
+        return list(value.values())
+    if isinstance(value, (list, tuple, set, ComponentSet)):
+        return list(value)
+    return [value]
```
emma58, line 82: You only check that singletons are Constraints here, not iterables. So I think you can just use the `target_list` function from https://github.com/Pyomo/pyomo/blob/b953cf90d6ed26df46cb5924301a89e2e76716fa/pyomo/core/util.py#L240 here. I think that's alright (and reduces duplicated code) since you'll crash if something fails to behave like a Constraint later anyway.

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,544 @@
...
+        except (TypeError, ValueError):
+            raise ValueError(
+                "Expected a dict or ComponentMap mapping Disjuncts to "
+                "Constraints for 'selected_constraints', but received an object "
+                "of type %s" % (type(arg).__name__,)
+            )
+    result = ComponentMap()
+    for disjunct, value in items:
+        result[disjunct] = _as_constraint_list(value)
+    return result
```
emma58, line 108: I think this is more complicated than it needs to be and you probably don't need this method at all. I would just assume that `arg` has the `items` method later on in the code: If it doesn't that's the error Python will throw, which should be clear enough in combination with the fact that you have documentation elsewhere of what the input should be. In addition to duck-typing, my motive here is that this function has a side effect that you *always* copy the dictionary, which could get expensive for large models.

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,544 @@
...
+)
+class SimpleDisjunctionTransformation(Transformation):
+    """Reformulate one or more Disjunctions as *simple* Disjunctions.
+
+    A *simple* Disjunction is one in which every Disjunct holds exactly one
+    Constraint. For each Disjunction that is transformed, this transformation
+    derives a single Constraint for each of its Disjuncts and assembles those
+    Constraints into a brand new Disjunction. Each new Disjunct keeps a single
+    Constraint derived from the Disjunct it was generated from, and reuses the
+    original model (problem) variables.
```
emma58, line 125:
```suggestion
    Constraint from the Disjunct it was generated from.
```

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,544 @@
...
+
+    The original Disjunction is never modified: the generated simple Disjunction
+    (together with its Disjuncts) is placed on a transformation Block added to
+    the parent Block of the Disjunction it was generated from. A single such
+    transformation Block is shared by every simple Disjunction built on the same
+    parent Block. The new Disjuncts get their own indicator variables, so the
+    simple Disjunction is an independent component that the caller may transform
+    or otherwise use however they see fit.
```
emma58, line 133: Whoops, sorry I didn't think of this sooner, but we probably should add LogicalConstraints making an equivalence between the simple disjunction's indicator_vars and the old guy's. Else this really isn't a transformation: The transformed model isn't equivalent to the original without that.

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,544 @@
...
+
+      * ``'first'`` (the default) keeps the first active Constraint encountered
+        on each Disjunct.
+      * ``'user_specified'`` keeps the Constraint that the user assigned to each
+        Disjunct through the ``selected_constraints`` mapping.
+
+    Both currently-implemented methods keep a single existing Constraint. The
+    selection machinery is, however, organized around lists of source
+    Constraints and a separate combination step (``_combine_sources``) so that
+    future methods which aggregate several Constraints into a single one can be
```
emma58, line 148:
```suggestion
    future methods that aggregate several Constraints into a single one can be
```

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,544 @@
...
+    Only active Constraints are ever considered. If a Disjunct has no active
+    Constraints, it is skipped (no corresponding Disjunct is created in the
+    simple Disjunction). If *every* Disjunct of a Disjunction is skipped, so
+    that the simple Disjunction would be empty, a GDP_Error is raised and no
+    simple Disjunction is created for that Disjunction.
+
+    Nested Disjunctions are not supported: a Disjunction is not transformed if
+    it is itself nested inside a Disjunct or if any of its Disjuncts contains a
+    Disjunction of its own. When no targets are given, such Disjunctions are
+    skipped automatically; when a nested Disjunction is supplied as an explicit
```
emma58, line 160: Sorry if we had a misunderstanding about this: I think we need to raise an error in this case too, to be consistent with other transformations. I think of not specifying targets as implicitly specifying *every* Disjunction as a target. That means that we should raise an error whenever we try to transform a nested GDP. (A user can avoid this error on a nested model by explicitly specifying only non-nested Disjunctions as targets.)

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,544 @@
...
+            """,
+        ),
+    )
+    CONFIG.declare(
+        'selected_constraints',
+        ConfigValue(
+            default=None,
+            domain=_selected_constraints_map,
+            description="Mapping from Disjuncts to the Constraint(s) to keep for "
+            "each of them",
```
emma58, line 214:
```suggestion
            description="Mapping from Disjuncts to the Constraint(s) to keep/aggregate to "
            "create the simple disjunction",
```

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,544 @@
...
+            the (active) Constraint, or list of Constraints, to retain for each
+            of those Disjuncts. This is only used (and is required) when
+            'constraint_selection_method' is 'user_specified'. A Disjunct that
+            owns active Constraints but is absent from this mapping is an error;
+            a Disjunct with no active Constraints may be omitted and is skipped.
+
+            The currently-implemented selection methods keep a single Constraint
+            per Disjunct, so a single Constraint is expected for each entry.
+            Values are nevertheless stored as lists so that selection methods
+            which aggregate several Constraints into one can reuse this option.
```
emma58, line 226:
```suggestion
           that aggregate several Constraints into one can use this option.
```

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,544 @@
...
+    #: selected source Constraints into the single expression placed in the
+    #: simple Disjunct. New strategies only need a new entry and builder.
+    _EXPRESSION_BUILDERS = {
+        ConstraintSelectionMethod.first: '_first_constraint_expression',
+        ConstraintSelectionMethod.user_specified: '_user_specified_expression',
+    }
+
+    def __init__(self):
+        super().__init__()
+        self.logger = logger
```
emma58, line 242: I don't think you need to put this on the class.

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,544 @@
...
+                "use it." % (method.value,)
+            )
+
+        build_expression = getattr(self, self._EXPRESSION_BUILDERS[method])
+
+        targets = config.targets
+        # Track the Disjunctions the user named explicitly (as opposed to ones
+        # we merely came across while descending into a Block or the instance).
+        # An explicit, ineligible target is an error; one we only discovered is
+        # skipped quietly.
```
emma58, line 277: See above, but let's change this so there's not a distinction here. If it's something we were supposed to transform and it's nested, that's an error.

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,544 @@
...
+        # rather than walking the model ourselves. (get_gdp_tree also validates
+        # that every target lives on the instance.)
+        gdp_tree = get_gdp_tree(targets, instance)
+        for disjunction in gdp_tree.topological_sort():
+            if disjunction.ctype is not Disjunction:
+                continue
+            if not disjunction.active:
+                # Discovery only turns up active Disjunctions, so an inactive one
+                # here was named explicitly.
+                if disjunction in explicit:
```
emma58, line 301: You can skip creating `explicit` and skip all these checks: Let's treat everything the same.

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,544 @@
...
+        # we merely came across while descending into a Block or the instance).
+        # An explicit, ineligible target is an error; one we only discovered is
+        # skipped quietly.
+        explicit = ComponentSet()
+        if targets is None:
+            targets = (instance,)
+        else:
+            for t in targets:
+                if t.ctype is Disjunction:
+                    explicit.update(t.values() if t.is_indexed() else (t,))
```
emma58, line 284: I think the only thing you need here is a check if the target is inactive, because that is an error. But you don't need to build `explicit`. (Honestly, we should probably have an option to check for active stuff when we build the gdp_tree... But we don't right now.)

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,544 @@
...
+
+    @staticmethod
+    def _nested_disjunction_owner(disjunction, gdp_tree):
+        # Return the first Disjunct of `disjunction` that owns a nested
+        # Disjunction, or None. In the GDP tree, a Disjunct with a nested
+        # Disjunction has that Disjunction as a child, so it is not a leaf.
+        for disjunct in disjunction.disjuncts:
+            if disjunct in gdp_tree.vertices and not gdp_tree.is_leaf(disjunct):
+                return disjunct
+        return None
```
emma58, line 352: You don't need this method: Just ask for `gdp_tree.parent(disjunction)`. (That will return `None` if `disjunction` is a root.)

pyomo/gdp/plugins/simple_disjunction_transform.py
```diff
@@ -0,0 +1,544 @@
...
+            if not disjunct.active:
+                continue
+            expression = build_expression(disjunction, disjunct, selected_constraints)
+            if expression is None:
+                logger.debug(
+                    "Disjunct '%s' has no active constraints to select, so it "
+                    "is skipped in the simple disjunction generated from "
+                    "Disjunction '%s'." % (disjunct.name, disjunction.name)
+                )
+                continue
```
emma58, line 370: Sorry I didn't catch this before yet either, but shouldn't you still create an empty Disjunct in the simple disjunction so that the transformed model will be equivalent to the original? If that messes up the constraint generation, maybe we put a trivial constraint on it, like just a redundant bounds constraint or something?

pyomo/gdp/plugins/simple_disjunction_transform.py
emma58: Oh, just kidding, I take it back. You could get here with bad user input to `selected_constraints` right?

emma58 (commented):
@cjohnston1 this is getting close! I'm sorry for not thinking of this earlier, but I have one more concern: Skipping Disjuncts without any active Constraints is not the right approach because the simple disjunction still needs to be allowed to choose "nothing" in that case just like the original. I'm okay with raising an error in that case because it's kind of insane: optimization will always choose to not constrain something if it can--that at least has to be an optimal indicator variable assignment. But the fundamental thing is that we should have whatever we do here not change the model.

pyomo/gdp/plugins/simple_disjunction_transform.py
emma58: Never mind, this is wrong: Because we don't deactivate the constraints on the original disjunction, this is okay, we don't need to explicitly link the indicator variables.

PR #3989, opened 2026-07-02
contrib.piecewise: Adding option for convexity/concavity and an epigraph/hypograph transformation

emma58:
## Summary/Motivation:

If your piecewise-linear function is convex/concave you can use a formulation with no discrete variables if you either have the correct objective pressure or are dealing in inequalities anyway. This enables such a formulation.

## Changes proposed in this PR:
- Add `convex` argument to `PiecewiseLinearFunction` construction (values `True`, `False`, and `None` (the default).
- Adds `contrib.piecewise.epigraph_hypograph` to provide a transformation of piecewise-linear functions to epigraphical or hypographical form. Currently we do no error-checking on the user's promise of convexity/concavity, so it is on their head to make sure that this is the formulation they really want.
- Tests of the above, with some help from Claude

## AI-Use Disclosure

- [ ] **AI tools were NOT used during the preparation of this PR**

or

- [x] **AI tools contributed to the development of this PR**
    - [x] AI tools generated documentation (including the PR description/comments, code comments, and/or Sphinx documentation)
    - [x] AI tools generated tests (baselines, examples, and/or code)
    - [ ] AI tools generated code (apart from tests)
    
    *Review process (select ONE)*:
    - [ ] **Rewritten**: All AI-generated content was rewritten by me before being committed.
    - [x] **Reviewed/verified**: I retained AI-generated content and verified it before committing. Verification included (as applicable):
        - [x] Ran the code and fixed issues
        - [ ] Added and ran tests
        - [x] Checked correctness/logic of code and tests
        - [x] Checked for alignment with the contribution guide
        - [x] Considered security implications
    - [ ] **As-is**: AI-generated content was commited directly to the repository
    
 **Notes for reviewers (optional):** I wouldn't worry about AI here: I completely rewrote the first two tests and then had it imitate them for the others, which seems to have worked well (the only telling evidence being waaaay more (boring) comments than an Emma would put in.)

### Legal Acknowledgement

By contributing to this software project, I have read the [contribution guide](https://pyomo.readthedocs.io/en/stable/contribution_guide.html) and agree to the following terms and conditions for my contribution:

1. I agree my contributions are submitted under the BSD license.
2. I represent I am authorized to make the contributions and grant the license. If my employer has rights to intellectual property that includes these contributions, I represent that I have received permission to make contributions and grant the required license on behalf of that employer.

PR #3990, opened 2026-07-02
contrib.piecewise: Add special case for single-segment "piecewise" linear functions

emma58:
## Summary/Motivation:

@michaelbynum and others have observed that we don't have any way to handle the special case of piecewise-linear in which the model is just plain linear (one segment). This catches that case before all of the GDP/MIP transformations so that we don't create unnecessary Booleans/binaries in this case.

- Adds handling in the transformation base class to return linear expressions for single-segment cases rather than creating disjunctive or other logical structures. This can still be bypassed when it's not relevant, and the epigraph/hypograph transformation (as written in #3989) already will.
- Tests single-segment functions for all transformations.

## AI-Use Disclosure

or

    - [x ] AI tools generated code (apart from tests)
    
        - [ ] Added and ran tests
    
 **Notes for reviewers (optional):** Note that this is completely independent of #3989: The epigraph/hypograph should not have this same handling because then it does something unexpected, which is to *not* return an epigraph/hypograph. So the fact that it will still redefine `_transform_pw_linear_expr` is good.

### Legal Acknowledgement


michaelbynum (approved):
Looks great!

PR #3991, opened 2026-07-08
Fix bug in initialization routines to return solve with bounded nlp

stephenscini:
## Fixes # .
Fixes issue in initialization routines where the model is run through nlp solver before bounds are removed.

## Summary/Motivation:

Currently: Original variables are bounded in methods and then there is cleanup but the initial point from lp, pwl, and global approximations is not verified to work with nlp_solver

- After each method "Finally: _cleanup()", run the original model again through the nlp solver and return that result. 
- Additional warning added for using the global solver without specifying a time limit.

## AI-Use Disclosure

- [x] **AI tools were NOT used during the preparation of this PR**

or

- [ ] **AI tools contributed to the development of this PR**
    
        - [ ] Added and ran tests
    
### Legal Acknowledgement


stephenscini:
@michaelbynum Please review when available.

pyomo/devel/initialization/initialize.py
```diff
@@ -156,6 +156,23 @@ def initialize_with_piecewise_linear_approximation(
     finally:
         _cleanup(orig_var_data)
 
+        # Commented out while I fix testing
```
stephenscini: @michaelbynum If I add this in the test for pwl_approx fails. The mock solver is not set up for more than one run of the nlp, and the nlp is used in the private call (line 146) to check the refinement of the pwl approximation. Working on adjusting mock solver.

pyomo/devel/initialization/initialize.py
```diff
@@ -156,7 +156,21 @@ def initialize_with_piecewise_linear_approximation(
...
+            nlp, load_solutions=False, raise_exception_on_nonoptimal_result=False
+        )
+        logger.info(
+            f'solved NLP: {nlp_res.solution_status}, {nlp_res.termination_condition}'
+        )
+
+        if nlp_res.solution_status in {SolutionStatus.feasible, SolutionStatus.optimal}:
+            nlp_res.solution_loader.load_vars()
+        else:
+            logger.warning('initialization was not successful via LP approximation')
```
michaelbynum: This should not be in the finally block.

pyomo/devel/initialization/initialize.py
```diff
@@ -239,7 +253,20 @@ def initialize_with_LP_approximation(
...
```
michaelbynum: Same here. This should not be in the finally block.

pyomo/devel/initialization/initialize.py
```diff
@@ -293,4 +320,15 @@ def initialize_with_global_opt(
...
+        )
+        logger.info(
+            f'solved NLP with {nlp_solver.name}: {nlp_res.solution_status}, {nlp_res.termination_condition}'
+        )
+        if nlp_res.solution_status in {SolutionStatus.feasible, SolutionStatus.optimal}:
+            nlp_res.solution_loader.load_vars()
+        else:
+            logger.warning('initialization was not successful via global optimization')
+
+    return nlp_res
```
michaelbynum: And same here.

michaelbynum (changes requested):
This is almost perfect. One comment below. Also, the code for the final nlp solve is almost identical in all three methods. Could you move this to a helper function?

stephenscini:
@michaelbynum Addressed comments. Please take another look when available. Thanks!

michaelbynum (approved):
Thanks!

blnicho (approved):

Issue #3573, opened 2025-04-22
Handle bounds in `solve_strongly_connected_components`


Robbybp:
## Summary

The algorithm in `solve_strongly_connected_components` is not robust to the following situation:
1. A subsystem has multiple solutions. We converge to one of them.
2. For this particular local solution (but not some other local solution), a downstream subsystem is infeasible.

We would like the first subsystem to converge to the "right" local solution.

### Rationale

Inspired by some problems @dallan-keylogic is having. See #3571.

### Description

I propose we do this with FBBT. Here's a sketch of what this looks like in `solve_strongly_connected_components`:
```python
    scc_input_list = list(
        generate_strongly_connected_components(constraints, variables, igraph=igraph)
    )

    if propagate_bounds:
        orig_bounds = [(var.lb, var.ub) for var in variables]
        for scc, inputs in scc_input_list:
            # Propagate bounds forward. This computes bounds on decision variables
            # based on fixed "degree of freedom" values. This may help us propagate
            # bound backwards in the reverse pass.
            fbbt(scc)
        for scc, inputs in reversed(scc_input_list):
            # Propagate bounds in the "reverse direction". This attempts to strengthen
            # "upstream" bounds based on existing bounds on decision variables.
            fbbt(scc)

    for scc, inputs in scc_input_list:
        # Solve scc...

    if propagate_bounds:
        # Reset bounds to their original values
        for var, (lb, ub) in zip(variables, orig_bounds):
            var.setlb(lb)
            var.setub(ub)
```

I've implemented this in my [`scc-fbbt`](https://github.com/Robbybp/pyomo/tree/scc-fbbt) branch. @dallan-keylogic, can you check if this branch helps with your infeasibility issue? Here's an example:
```python
import pyomo.environ as pyo
from pyomo.contrib.incidence_analysis import solve_strongly_connected_components

m = pyo.ConcreteModel()
m.x = pyo.Var([1, 2, 3, 4], initialize=1.0)
m.eq = pyo.Constraint(pyo.PositiveIntegers)
m.eq[1] = m.x[1]**2 == 1.0
m.x[1] = -2.0
m.eq[2] = m.x[2]**2 == 4.0
m.x[2].setlb(1.0)
m.eq[3] = m.x[2] == m.x[3]
m.eq[4] = m.x[1] * m.x[3] * m.x[4] == 1.0
m.x[4].setlb(0.0)
solver = pyo.SolverFactory("ipopt")
[...]

PR #3993, opened 2026-07-13
Clip and scale calculate_variable_from_constraint

dallan-keylogic:
## Summary/Motivation:
Projects like PrOMMiS are increasingly relying on IDAES's `BlockTriangularizationInitializer`, which in turn relies on `solve_strongly_connected_components`.  However, in models that need significant amounts of scaling, the initialization raises an error that the constraint residuals are not small. That's because `BlockTriangularizationInitializer` takes constraint scaling into account but `solve_strongly_connected_components` does not.

This PR begins addressing this issue by adding an option to scale the variable and constraint in `calculate_variable_from_constraint`. Additionally, it adds an option to clip variable values that fall outside their bounds back inside those bounds, because we have a ton of log spam from Pyomo warning about variables getting set to values barely outside their bounds.

- Add option to allow scaling of variable and constraint in `calculate_variable_from_constraint
- Add option to clip variables back to their bounds

## AI-Use Disclosure

or

        - [ ] Added and ran tests
    
 **Notes for reviewers (optional):** I set the default values of these new options to `False` in order to avoid anything breaking from using the new options. However, I think users will usually want to set those options to `True`. On the IDAES side, I can set those options in the configuration of `BlockTriangularizationInitializer`, but consider what the default behavior in Pyomo should be.

Additionally, when clipping the values back to their bounds, I check the residual to see whether to accept the clipping or not. If the clipped value does not satisfy the convergence condition, it gets set to the unclipped value. What should the default behavior be?

Finally, I noticed the line search done here doesn't use the Armijo sufficient decrease condition. Is that intended behavior?

### Legal Acknowledgement

### DIFF
--- a/pyomo/util/calc_var_value.py
+++ b/pyomo/util/calc_var_value.py
@@ -11,6 +11,7 @@ from pyomo.common.errors import IterationLimitError
 from pyomo.common.numeric_types import native_numeric_types, native_complex_types, value
 from pyomo.core.expr.calculus.derivatives import differentiate
 from pyomo.core.base.constraint import Constraint
+from pyomo.core.base.suffix import SuffixFinder
 
 import logging
 
@@ -24,6 +25,53 @@ _symbolic_modes = {
 }
 
 
+def _clip_variable_to_bounds(variable, expr, eps):
+    """
+    Function to try to clip a variable back within its bounds.
+    If the expression residual is less than eps, then the new
+    value of the variable is retained. If the value is greater
+    than eps, then the original, bound-violating value of
+    variable is retained.
+
+    Parameters:
+    -----------
+    variable: :py:class:`VarData`
+        The variable to clip to within its bounds
+    expr: :py:class:`NumericExpression`
+        The expression to evaluate the constraint residual
+    eps: `float`
+        The tolerance to use to determine constraint feasibility.
+
+    Returns:
+    --------
+    None
+
+    """
+    # Should we check against variable domain here as well?
+    variable_value = value(variable)
+    if (
+        variable.lb is not None
+        and variable.ub is not None
+        and variable.lb > variable.ub
+    ):
+        raise ValueError(
+            f"Variable {variable} has inconsistent bounds. "
+            f"It has a lower bound of {variable.lb} and an "
+            f"upper bound of {variable.ub}."
+        )
+    elif variable.lb is not None and variable_value < variable.lb:
+        variable.set_value(variable.lb)
+        if abs(value(expr)) < eps:
+            return
+        # Else proceed to set the variable to its old value
+    elif variable.ub is not None and variable_value > variable.ub:
+        variable.set_value(variable.ub)
+        if abs(value(expr)) < eps:
+            return
+        # Else proceed to set the variable to its old value
+    variable.set_value(variable_value)
+
+
 def calculate_variable_from_constraint(
     variable,
     constraint,
@@ -32,6 +80,8 @@ def calculate_variable_from_constraint(
     linesearch=True,
     alpha_min=1e-8,
     diff_mode=None,
+    scale_problem=False,
+    clip_to_bounds=False,
 ):
     """Calculate the variable value given a specified equality constraint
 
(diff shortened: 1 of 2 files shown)


pyomo/util/calc_var_value.py
```diff
@@ -310,7 +391,9 @@ def calculate_variable_from_constraint(
                     f"variable '{variable}' using constraint '{constraint}'; "
                     f"remaining residual = {residual}."
                 )
-    #
-    # Re-set the variable value to trigger any warnings WRT the final
-    # variable state
-    variable.set_value(variable.value)
```
Robbybp, line 316: We could just add `skip_validation=True` here to silence the warnings. And potentially expose this option in the user-facing function. Not saying this is necessarily better than clipping, just want to make sure it's considered.

We could also transform `calculate_variable_from_constraint` into a bound-constrained solver by "clipping" the variable when taking the Newton step, as done by interior point methods.

dallan-keylogic: I'd rather have the warnings if we end up setting the variable value to something outside its bounds.

I would be okay with converting it to a bound-constrained solver. For the initial version of this PR, I wanted to keep the changes minor as to not break any existing workflow. I wonder whether turning on bound clipping would cause some of the `CubicComplementarityVLE` examples to break in IDAES---we've seen examples where earlier blocks converge to the wrong root, causing later blocks to have no feasible solution that satisfies the variable bounds, even though a solution exists if the entire VLE were solved on a single block. I think the moral of that story is "You need a better initialization method for VLE".

Robbybp: Maybe `skip_validation` if it violates the bound by more than the tolerance?

I guess my problem with clipping is that you're just trading some bound infeasibility for some equality infeasibility. I would want to make sure that we didn't make the constraint infeasible (by more than tolerance) after the clipping. But I'm also now realizing that, since these are single-variable problems, clipping is not so different than respecting bounds during the solve iterations.

Robbybp: > we've seen examples where earlier blocks converge to the wrong root, causing later blocks to have no feasible solution that satisfies the variable bounds, even though a solution exists if the entire VLE were solved on a single block

I would love to see an example of this. I have some ideas related to bound propagation among blocks (e.g., https://github.com/Pyomo/pyomo/issues/3573) that I would like to test on a real problem.

dallan-keylogic: > Maybe skip_validation if it violates the bound by more than the tolerance?
> 
> I guess my problem with clipping is that you're just trading some bound infeasibility for some equality infeasibility. I would want to make sure that we didn't make the constraint infeasible (by more than tolerance) after the clipping. 

Right now, if clipping the variable causes the constraint to become infeasible, it reverts to the unclipped value and logs a warning.

> But I'm also now realizing that, since these are single-variable problems, clipping is not so different than respecting bounds during the solve iterations.

With any sort of nonlinear solution method, there's always the risk that changing the intermediate iterations causes the problem to converge to either a different solution or a stationary point for the constraint residual. In general, though, I would guess that clipping as part of the line search would give a better chance at converging to a solution that satisfies the bound (if such a solution exists). 

There are two possible implementations:

1. Do the thing same as IPOPT: if a candidate step violates a bound, set `alpha = 0.5 * alpha` and repeat. 
2. Just jump directly to the bound. If taking a step at the bound also results in bound violation, then we've found a local minimizer and terminate.

I think I prefer option (2), because we don't have to worry about how to determine we've reached a bound---we just jump straight to the bound. With option (1), we could end up spending a long time taking tiny line search steps before the sufficient decrease condition is violated.

Edit: Another thing to keep in mind is consistency between the three methods used in the function. The first shortcut method assumes the constraint is of the form `x == y`, the second assumes it's of the form `y == m*x + b`. Neither one of those methods checks bounds, and I'd prefer not to slow them down by adding checks.

Robbybp: > In general, though, I would guess that clipping as part of the line search would give a better chance at converging to a solution that satisfies the bound (if such a solution exists).

Good point. I don't have a strong preference on the implementation if it is demonstrated to work, as long as this is opt-in behavior (for now).

pyomo/util/calc_var_value.py
```diff
@@ -72,6 +122,16 @@ def calculate_variable_from_constraint(
     diff_mode: :py:enum:`pyomo.core.expr.calculus.derivatives.Modes`
         The mode to use to differentiate the expression.  If
         unspecified, defaults to `Modes.sympy`
+    scale_problem: `bool`
+        Whether to take variable and constraint scaling into account
+        when calculating if the constraint residual tolerance is
+        satisfied or if the initial derivative is zero.
```
Robbybp, line 128: If this is just scaling tolerances, I'm not sure it belongs in this function. The caller can always use the scaling factor to compute a new tolerance if they want. I also don't love calling SuffixFinder within this function.

dallan-keylogic: > If this is just scaling tolerances, I'm not sure it belongs in this function.

It also applies scaling factors when calculating if a problem has a derivative of effectively zero.

> The caller can always use the scaling factor to compute a new tolerance if they want.

That's true, but that means I'm going to move the scaling here into `solve_strongly_connected_components`.

>  I also don't love calling SuffixFinder within this function.

Is the concern that it adds too much overhead to each function call?

PR #3994, opened 2026-07-13
Update multistart solver, add method for multistart NLP initialization

stephenscini:
## Summary/Motivation:
Additional method in devel/initialization to use multistart optimization as a way for initializing nonliner models.

- Updating contrib/multistart solver for reproducibility, and adding support for new solver interface.
- Adding support for uniform, lhs, and sobol sampling within the multistart solver
- Adding method to devel/initialization

## AI-Use Disclosure

    - [x] **Rewritten**: All AI-generated content was rewritten by me before being committed.
    
### Legal Acknowledgement


stephenscini:
Opening PR, but still in active development. Dependent on #3991 first as it includes those changes.

stephenscini:
Note for 7/21 Dev meeting

Method implemented with support for new solver interface. All original tests passing locally. 

Tasks before ready for review:
- [x] It looks like there is an issue with changing the SolverFactory version in a different place, which I will address.
- [ ] Try and simplify interface, check with mentors.
- [ ] Add tests for new functionality

pyomo/contrib/solver/tests/solvers/test_solvers.py
```diff
@@ -48,6 +48,8 @@
 from pyomo.core.expr.compare import assertExpressionsEqual
 from pyomo.core.expr.numeric_expr import LinearExpression
 
+from pyomo.contrib.multistart.multi import MultiStart
```
stephenscini, line 51: I received an error message when I added the solver to the new SolverFactory that it needed to be in this list. This would mean it is tested with the other solvers from contrib/solver and I am not sure that if that is expected? If so, will all contrib solvers need to eventually be switched over or should the opt/solvers still be the best location.

In this case, I wanted to update it to use these solvers so thought it would also go here. Maybe Legacy?

pyomo/contrib/multistart/reinit.py
```diff
@@ -11,35 +11,50 @@
...
+from pyomo.core.expr.visitor import identify_variables
 
 from pyomo.core import Var
 
 logger = logging.getLogger('pyomo.contrib.multistart')
 
 
-def rand(val, lb, ub):
-    return random.uniform(lb, ub)  # uniform distribution between lb and ub
+def rand(val, lb, ub, rng):
```
stephenscini: Previously, there was not a set way I could find to make the solver result reproducible, so I made the default random generator np.random with the current RNG interface. 

I wanted to otherwise leave the established strategies alone, but could also make all random strategies support the different random generator options? Kept all newer options under rand_vector to use a vectorized approach.

pyomo/contrib/multistart/reinit.py
```diff
@@ -75,8 +94,34 @@ def reinitialize_variables(model, config):
...
-        # apply reinitialization strategy to variable
-        var.set_value(
-            strategies[config.strategy](val, var.lb, var.ub), skip_validation=True
-        )
+
+        eligible_vars.append(var)
+
+    if config.strategy == "rand_vector":
+
+        if len(eligible_vars) == 0:
```
stephenscini, line 100: Before, if all the variables are unbounded, it would not reset them but just give the warning. I thought it would be better to error out.

pyomo/contrib/multistart/multi.py
```diff
@@ -12,23 +12,169 @@
...
+from pyomo.contrib.solver.common.factory import SolverFactory
+from pyomo.contrib.solver.common.results import SolutionStatus
+
+from pyomo.common.dependencies.scipy import stats
+from pyomo.common.dependencies import numpy as np
 
 logger = logging.getLogger('pyomo.contrib.multistart')
 
 
+@document_configdict()
```
stephenscini, line 51: Made changes so it is now loaded as a solver for the new SolverFactory since I wanted to use the new solver interfaces like the other devel/initialization methods, and from discussion with Bethany. Causing testing errors but would like to discuss best way to do this if other method preferred.

pyomo/contrib/multistart/multi.py
```diff
@@ -130,6 +197,12 @@ def available(self, exception_flag=True):
         """
         return True
 
+    def version(self):
+        """Get solver version
+        TODO: This is a solver wrapper, unsure how to define version in this case."""
+
```
stephenscini, line 267: Currently no versions for this. If old was v0.0, would this be v0.1?

pyomo/contrib/multistart/multi.py
```diff
@@ -138,7 +211,15 @@ def solve(self, model, **kwds):
         config = self.CONFIG(kwds.pop('options', {}))
         config.set_value(kwds)
 
-        # initialize the solver
+        # Create centralized sampler once
+        sampler = SamplingManager(
+            method=config.sampling_method, rng=config.rng, seed=config.seed
+        )
+
```
stephenscini, line 303: The goal of this solver is to run multiple times, so I thought having these as default solver options would make sense. In this case solver_args are for the subsolver.

pyomo/contrib/multistart/multi.py
```diff
@@ -12,23 +12,169 @@
...
+            - "rand": random choice between variable bounds
+            - "rand_vector": random choice, vectorized approach with sampler
+            - "midpoint_guess_and_bound": midpoint between current value and farthest bound
+            - "rand_guess_and_bound": random choice between current value and farthest bound
+            - "rand_distributed": random choice among evenly distributed values
+            - "midpoint": exact midpoint between the bounds. If using this option, multiple iterations are useless.
+            """,
+            ),
+        )
+        self.solver = self.declare(
```
stephenscini: Should I make this have different solver acceptances like MindtPy where there are supported categories, or make it so it accepts the solver object? How consistent should the contrib solvers that accept other solvers be with handling internal solver?

pyomo/contrib/multistart/multi.py
```diff
@@ -12,23 +12,169 @@
...
+    ):
+        super().__init__(
+            description=description,
+            doc=doc,
+            implicit=implicit,
+            implicit_domain=implicit_domain,
+            visibility=visibility,
+        )
+
+        self.strategy = self.declare(
```
stephenscini, line 69: Could/should this also support a user supplying a list of initial points? Thinking of obj_at_theta in parmest. Goal in my head is to make as generalized as possible.

pyomo/contrib/multistart/multi.py
```diff
@@ -149,14 +230,14 @@ def solve(self, model, **kwds):
             raise RuntimeError(
                 "Multistart solver is unable to handle model with multiple active objectives."
             )
-        if obj is None:
```
stephenscini, line 152: Previously it was restricted from using square models or constant objectives in the multistart solver. Is there a valid reason for this or can it be removed? Is there a case where a model would have no objective or would that cause a different error? Removing would make more sense in my opinion.

pyomo/contrib/multistart/multi.py
```diff
@@ -12,23 +12,169 @@
...
+            ),
+        )
+        self.HCS_tolerance = self.declare(
+            "HCS_tolerance",
+            ConfigValue(
+                default=0,
+                description="Tolerance on HCS objective value equality. Defaults to Python float equality precision.",
+            ),
+        )
+        self.break_on_solution = self.declare(
```
stephenscini, line 154: For initialization, I thought this made sense, other thought would be implementing a SolutionLimit argument like Gurobi.

pyomo/contrib/multistart/multi.py
```diff
@@ -204,15 +293,27 @@ def solve(self, model, **kwds):
...
+                    SolutionStatus.feasible,
+                    SolutionStatus.optimal,
+                }:
+                    result.solution_loader.load_vars()
+                    logger.info(
+                        f'solved NLP: {result.solution_status}, {result.termination_condition}'
+                    )
+                    # If we are looking for the first feasible solution, then return immediately
+                    if config.break_on_solution:
+                        return best_result
```
stephenscini: Would it be a good idea to have this return a dictionary or something instead of just the best result if someone wanted to inspect? Or is this better for a user to set up? Should a user be able to access subsolver information too?

stephenscini (commented):
Went through and added some questions in the code for design meeting.

stephenscini:
Had discussion with Bethany and Michael. Main takeaway points:

- Invest to make this work for newer SolverFactory. Do not keep in opt.SolverFactory.
- Try and resolve all test failures with the new SolverFactory as default. More discussion needed on expectations for contrib solvers as they transition to new SolverFactory.
- If all results are infeasible/not optimal, have it throw the 'NoOptimalSolution' error.
-Keep this PR focused on a minimum working version that has key functionality we want. Can always add additional touches later.
- LHS makes sense to be default for all methods/strategies, and should be looped in to the old strategies too.
- Availability of multistart should depend on availablility of chosen sub-solver (default is ipopt)
- Version 0.1.0 makes sense for this new iteration.
- Subsolver has its own configuration, but still needs to comply with 'meta-solver' configurations. If subsolver timelimit is 1 minute for 50 runs, but metasolver is 10 minutes, when metasolver time runs out the methods stop.

pyomo/contrib/multistart/multi.py
stephenscini: From design meeting consensus: That would be nice, but get a minimum viable version working and this can be an improvement later

pyomo/contrib/multistart/multi.py
```diff
@@ -40,221 +225,284 @@ class MultiStart:
...
+        )
+        # As we are about to run a solver, update the stale flag
+        StaleFlagManager.mark_all_as_stale()
+
+        # Create centralized sampler once
+        sampler = SamplingManager(
+            method=config.sampling_method, rng=config.rng, seed=config.seed
+        )
+
+        # Set sub-solver options
```
stephenscini: Note: Michael made case that all arguments should be independent for solver and subsolver. Fix before converting from draft

pyomo/contrib/multistart/reinit.py
```diff
@@ -75,8 +92,34 @@ def reinitialize_variables(model, config):
...
                 )
             continue
-        val = var.value if var.value is not None else (var.lb + var.ub) / 2
-        # apply reinitialization strategy to variable
-        var.set_value(
-            strategies[config.strategy](val, var.lb, var.ub), skip_validation=True
-        )
+
+        eligible_vars.append(var)
+
```
stephenscini, line 97: Originally I made new method "rand_vector" to distinguish a method for sampling using stratified methods to retain original methods. Now all methods support stratified sampling. Is there a need/value for a vectorized form?

pyomo/contrib/solver/tests/solvers/test_solvers.py
stephenscini: Consensus from design meeting: Yes, this solver should and does comply with the tests in the new solver interface.

pyomo/contrib/multistart/multi.py
stephenscini: Removed other than load_solution and raise_exception_on_nonoptimal_result. These are handled by the meta-solver.

stephenscini:
@michaelbynum @blnicho I think this is ready for review. Please take a look when you have a chance. Thanks!

stephenscini:
Note from 08/25 dev meeting: Not necessarily a bad design choice but technically precedent in Pyomo is to allow users to change the solver during solve after initialization. Worth future discussion.

stephenscini:
> Note from 08/25 dev meeting: Not necessarily a bad design choice but technically precedent in Pyomo is to allow users to change the solver during solve after initialization. Worth future discussion.

Followed precedent and added back config option. Also adjusted so availability checks subsolver, and is rechecked during solve in case of change.

stephenscini:
@michaelbynum @blnicho 
Can the osx/3.14 be rerun please? Not sure why it failed. Thanks!

PR #3942, opened 2026-05-11
Fix GDPopt LBB time-limit results

bernalde:
Summary
- Replace GDPopt LBB's stale `_get_final_results_object()` call on the time-limit path with `_get_final_pyomo_results_object()`.
- Add a solver-independent LBB regression test that forces the time-limit branch and verifies `maxTimeLimit` results.
- Add MindtPy GOA time-limit coverage to verify its inherited GOA path sets `maxTimeLimit` on `SolverResults`.

Tests run
- `pytest -q pyomo/contrib/gdpopt/tests/test_LBB.py::TestGDPopt_LBB_TimeLimit::test_time_limit_returns_pyomo_results_object pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py::TestMindtPyGOATimeLimit::test_goa_time_limit_sets_solver_results_condition` - passed, 2 passed in 0.52s
- `pytest -q pyomo/contrib/gdpopt/tests/test_LBB.py pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py` - passed, 32 passed, 9 skipped, 2 deselected in 1.86s
- `python -m black --check --diff pyomo/contrib/gdpopt/branch_and_bound.py pyomo/contrib/gdpopt/tests/test_LBB.py pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py` - passed, 3 files would be left unchanged
- `git diff --check -- pyomo/contrib/gdpopt/branch_and_bound.py pyomo/contrib/gdpopt/tests/test_LBB.py pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py` - passed

Notes about tests not run
- Full repository pytest and full-repository Black were not run locally because the targeted change is narrow and the full CI matrix covers multiple Python versions, platforms, optional solver installs, documentation, and coverage jobs.

Closes #3941

PR #4010, opened 2026-08-15
Fix GDPopt LBB time-limit results

### LINKED ISSUE #3941: GDPopt LBB time-limit path calls stale _get_final_results_object
## Summary

`GDP_LBB_Solver._solve_gdp()` calls a stale private method, `self._get_final_results_object()`, when GDPopt LBB exits through the time-limit branch. The current GDPopt base class exposes `_get_final_pyomo_results_object()` instead, so LBB raises `AttributeError` instead of returning a normal `SolverResults` object with `maxTimeLimit`.

This appears to be distinct from, but related to, the broader open LBB correctness issue in #2483.

## Minimal reproducer

This MWE forces the LBB time-limit branch to isolate the stale method call. It does not require external solvers or the downstream GDPlib model.

```python
import pyomo
import pyomo.environ as pyo
from pyomo.contrib.gdpopt.branch_and_bound import GDP_LBB_Solver

print("pyomo", pyomo.__version__)
print(
    "has _get_final_results_object",
    hasattr(GDP_LBB_Solver, "_get_final_results_object"),
)
print(
    "has _get_final_pyomo_results_object",
    hasattr(GDP_LBB_Solver, "_get_final_pyomo_results_object"),
)

# Force the time-limit exit path without depending on solver runtime.
GDP_LBB_Solver.reached_time_limit = lambda self, config: True

m = pyo.ConcreteModel()
m.x = pyo.Var(bounds=(0, 2))
m.d1 = Disjunct()
m.d2 = Disjunct()
m.d1.c = pyo.Constraint(expr=m.x <= 0.5)
m.d2.c = pyo.Constraint(expr=m.x >= 1.5)
m.obj = pyo.Objective(expr=m.x)

pyo.SolverFactory("gdpopt.lbb").solve(m, time_limit=1, tee=False)
```

## Actual output

```text
pyomo 6.10.0
has _get_final_results_object False
has _get_final_pyomo_results_object True
WARNING: 09/06/22: The GDPopt LBB algorithm currently has known issues. Please
use the results with caution and report any bugs!
  File "<string>", line 1, in <module>
  File ".../pyomo/contrib/gdpopt/branch_and_bound.py", line 90, in solve
    return super().solve(model, **kwds)
  File ".../pyomo/contrib/gdpopt/algorithm_base_class.py", line 128, in solve
    self._solve_gdp(model, config)
  File ".../pyomo/contrib/gdpopt/branch_and_bound.py", line 239, in _solve_gdp
    return self._get_final_results_object()
AttributeError: 'GDP_LBB_Solver' object has no attribute '_get_final_results_object'. Did you mean: '_get_final_pyomo_results_object'?
```

## Expected behavior

LBB should return a normal Pyomo `SolverResults` object from the time-limit branch, with the termination condition set by `reached_time_limit(config)`.

## Likely fix

In `pyomo/contrib/gdpopt/branch_and_bound.py`, the time-limit branch currently contains:

```python
return self._get_final_results_object()
```

The leaf-node return path in the same method already uses the current method name:

```python
return self._get_final_pyomo_results_object()
```

Changing the time-limit branch to call `_get_final_pyomo_results_object()` appears to resolve the `AttributeError`. A local monkeypatch aliasing the old name to the current method let the time-limit path return normally in the downstream reproducer.

## Downstream context

This was encountered while benchmarking GDPlib's `gdp_col` model:

- GDPlib model issue: https://github.com/SECQUOIA/gdplib/issues/66
- GDPlib central benchmark tracker: https://github.com/SECQUOIA/gdplib/issues/60
- GDPlib PR where this surfaced: https://github.com/SECQUOIA/gdplib/pull/124
- Local solver campaign evidence: https://github.com/SECQUOIA/gdplib/pull/124#issuecomment-4421060451
- Root-cause note in the GDPlib PR: https://github.com/SECQUOIA/gdplib/pull/124#issuecomment-4421111392

In the GDPlib run, both DICOPT-backed and BARON-backed `gdpopt.lbb` campaigns reached this same `AttributeError` after the LBB time-limit path. With a local method-name monkeypatch, the same `gdp_col` LBB path returned `maxTimeLimit` cleanly, although it still did not find a feasible incumbent. That suggests the stale method call is a reporting/control-flow bug independent of the model formulation.

## Environment

Observed locally with:

- Pyomo: 6.10.0
- Python: 3.12.13
- OS/platform: Linux WSL2, glibc 2.39

I also checked `pyomo/contrib/gdpopt/branch_and_bound.py` on upstream `Pyomo/pyomo` `main` via GitHub and saw the same stale call in the time-limit branch.

bernalde:
## Summary/Motivation:

The GDPopt LBB time-limit path calls `self._get_final_results_object()`, which no longer exists. The method was renamed to `_get_final_pyomo_results_object()` during an earlier GDPopt rewrite, and this one call site was left stale. Any LBB solve that terminates by hitting its time limit raises `AttributeError` instead of returning a results object.

The same file already uses the current name at the other termination site (`branch_and_bound.py:288`), so this is an isolated leftover rather than an intentional difference.

This is a replacement for #3942, which GitHub would not allow me to reopen after it was closed. That PR was approved by @emma58 and was closed for a PR-template/CI-state reason rather than a code concern; the prior review discussion and history remain available there.

- Call `_get_final_pyomo_results_object()` on the GDPopt LBB time-limit termination path in `pyomo/contrib/gdpopt/branch_and_bound.py`, matching the other termination site in the same file.
- Add `TestGDPopt_LBB_TimeLimit.test_time_limit_returns_pyomo_results_object`, a solver-independent regression that forces the time-limit branch and asserts a Pyomo results object is returned with `maxTimeLimit` and unbounded problem bounds.
- Add `TestMindtPyGOATimeLimit.test_goa_time_limit_sets_solver_results_condition`, documenting that the corresponding MindtPy GOA time-limit path already sets its termination condition correctly.
- Validation performed locally on this branch, rebased onto current `main`:
  - `python -m pytest -q pyomo/contrib/gdpopt/tests/test_LBB.py pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py`
    - Result: `32 passed, 9 skipped, 2 deselected in 2.55s`
  - `python -m pytest -q pyomo/contrib/gdpopt/tests/`
    - Result: `76 passed, 35 skipped, 5 deselected in 81.25s`
  - `python -m black --check --diff pyomo/contrib/gdpopt/branch_and_bound.py pyomo/contrib/gdpopt/tests/test_LBB.py pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py`
    - Result: passed, `3 files would be left unchanged`
  - `typos --config ./.github/workflows/typos.toml` on the three changed files
    - Result: passed, no findings

## AI-Use Disclosure

or

        - [x] Added and ran tests
    
 **Notes for reviewers (optional):** This replacement PR carries over the implementation and tests from #3942 unchanged, then refreshes the branch against current `main`. The replacement PR description and branch-refresh workflow were prepared with AI assistance and reviewed before posting. The runtime change is a single call-site rename; reviewers may want to focus on whether the forced-time-limit regression test in `test_LBB.py` is the right way to exercise that path without requiring a MINLP subsolver, since it temporarily patches `GDP_LBB_Solver.reached_time_limit` and restores it in a `finally` block.

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/gdpopt/tests/test_LBB.py
+++ b/pyomo/contrib/gdpopt/tests/test_LBB.py
@@ -20,8 +20,17 @@ from pyomo.common.fileutils import import_file
 from pyomo.common.log import LoggingIntercept
 import pyomo.contrib.gdpopt.tests.common_tests as ct
 from pyomo.contrib.satsolver.satsolver import z3_available
-from pyomo.environ import SolverFactory, value, ConcreteModel, Var, Objective, maximize
-from pyomo.gdp import Disjunction
+from pyomo.contrib.gdpopt.branch_and_bound import GDP_LBB_Solver
+from pyomo.environ import (
+    SolverFactory,
+    value,
+    ConcreteModel,
+    Constraint,
+    Var,
+    Objective,
+    maximize,
+)
+from pyomo.gdp import Disjunct, Disjunction
 from pyomo.opt import TerminationCondition
 
 currdir = dirname(abspath(__file__))
@@ -33,6 +42,40 @@ solver_available = SolverFactory(minlp_solver).available()
 license_available = (
     SolverFactory(minlp_solver).license_is_valid() if solver_available else False
 )
+
+
+class TestGDPopt_LBB_TimeLimit(unittest.TestCase):
+    """Tests for solver-independent LBB termination paths."""
+
+    def test_time_limit_returns_pyomo_results_object(self):
+        m = ConcreteModel()
+        m.x = Var(bounds=(0, 2))
+        m.d1 = Disjunct()
+        m.d2 = Disjunct()
+        m.d1.c = Constraint(expr=m.x <= 0.5)
+        m.d2.c = Constraint(expr=m.x >= 1.5)
+        m.disj = Disjunction(expr=[m.d1, m.d2])
+        m.obj = Objective(expr=m.x)
+
+        orig_reached_time_limit = GDP_LBB_Solver.reached_time_limit
+
+        def force_time_limit(solver, config):
+            solver.pyomo_results.solver.termination_condition = (
+                TerminationCondition.maxTimeLimit
+            )
+            return True
+
+        GDP_LBB_Solver.reached_time_limit = force_time_limit
+        try:
+            results = SolverFactory('gdpopt.lbb').solve(m, time_limit=1, tee=False)
+        finally:
+            GDP_LBB_Solver.reached_time_limit = orig_reached_time_limit
+
+        self.assertEqual(
+            results.solver.termination_condition, TerminationCondition.maxTimeLimit
+        )
+        self.assertEqual(results.problem.lower_bound, float('-inf'))
+        self.assertEqual(results.problem.upper_bound, float('inf'))
 
 
 @unittest.skipUnless(
--- a/pyomo/contrib/gdpopt/tests/test_LBB.py
+++ b/pyomo/contrib/gdpopt/tests/test_LBB.py
@@ -13,6 +13,7 @@ from io import StringIO
 import logging
 from math import fabs
 from os.path import abspath, dirname, join, normpath
+from unittest.mock import patch
 
 import pyomo.common.unittest as unittest
 
@@ -57,19 +58,14 @@ class TestGDPopt_LBB_TimeLimit(unittest.TestCase):
         m.disj = Disjunction(expr=[m.d1, m.d2])
         m.obj = Objective(expr=m.x)
 
-        orig_reached_time_limit = GDP_LBB_Solver.reached_time_limit
-
         def force_time_limit(solver, config):
             solver.pyomo_results.solver.termination_condition = (
                 TerminationCondition.maxTimeLimit
             )
             return True
 
-        GDP_LBB_Solver.reached_time_limit = force_time_limit
-        try:
+        with patch.object(GDP_LBB_Solver, 'reached_time_limit', new=force_time_limit):
             results = SolverFactory('gdpopt.lbb').solve(m, time_limit=1, tee=False)
-        finally:
-            GDP_LBB_Solver.reached_time_limit = orig_reached_time_limit
 
         self.assertEqual(
             results.solver.termination_condition, TerminationCondition.maxTimeLimit
--- a/pyomo/contrib/gdpopt/tests/test_LBB.py
+++ b/pyomo/contrib/gdpopt/tests/test_LBB.py
@@ -13,7 +13,6 @@ from io import StringIO
 import logging
 from math import fabs
 from os.path import abspath, dirname, join, normpath
-from unittest.mock import patch
 
 import pyomo.common.unittest as unittest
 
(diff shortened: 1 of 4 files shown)


pyomo/contrib/gdpopt/tests/test_LBB.py
```diff
@@ -35,6 +44,40 @@
...
+            solver.pyomo_results.solver.termination_condition = (
+                TerminationCondition.maxTimeLimit
+            )
+            return True
+
+        GDP_LBB_Solver.reached_time_limit = force_time_limit
+        try:
+            results = SolverFactory('gdpopt.lbb').solve(m, time_limit=1, tee=False)
+        finally:
+            GDP_LBB_Solver.reached_time_limit = orig_reached_time_limit
```
emma58: I don't understand the try-finally here: It looks like if line 70 throws an exception `results` will be undefined, which will lead to an exception in the assertions below.

emma58 (commented):
One question about the LBB test, but otherwise this look good!

bernalde:
Addressed the LBB test review in `b63db6190`.

- Replaced manual `GDP_LBB_Solver.reached_time_limit` assignment/restoration with a scoped `patch.object(...)`. If `solve()` raises, its exception propagates after cleanup and the assertions are not reached; the context manager makes that lifecycle explicit.
- Local verification:
  - `python -m pytest -q pyomo/contrib/gdpopt/tests/test_LBB.py::TestGDPopt_LBB_TimeLimit::test_time_limit_returns_pyomo_results_object` — 1 passed
  - `python -m pytest -q pyomo/contrib/gdpopt/tests/test_LBB.py pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py` — 32 passed, 9 skipped, 2 deselected
  - `python -m pytest -q pyomo/contrib/gdpopt/tests/` — 73 passed, 38 skipped, 5 deselected
  - `python -m black . --check --diff` — passed
  - `typos --config ./.github/workflows/typos.toml` — passed
- Current-head CI: GitHub Actions completed successfully across all 30 jobs, and Read the Docs succeeded.
- Intentionally not addressed: none.
- Remaining: `Jenkins CI / inspection` is still pending. The live review decision is `REVIEW_REQUIRED`; the review thread remains unresolved for the reviewer to close.

pyomo/contrib/gdpopt/tests/test_LBB.py
bernalde: Addressed in `b63db6190`. If `solve()` raises, the original `try`/`finally` restores the method and then propagates that exception, so the assertions are not reached and `results` is not read. I replaced the manual assignment/restoration with `patch.object(...)` to make that lifecycle explicit; the focused regression test and full GDPopt suite pass.

bernalde:
@emma58, the LBB test follow-up is addressed in `b63db6190`. Could you please take another look?

pyomo/contrib/gdpopt/tests/test_LBB.py
```diff
@@ -13,6 +13,7 @@
 import logging
 from math import fabs
 from os.path import abspath, dirname, join, normpath
+from unittest.mock import patch
```
emma58: We prefer to not use `mock` when there is not a good reason it's needed (see [this page](https://pyomo.readthedocs.io/en/stable/contribution_guide.html#use-of-generative-ai)). I think in this case it should be sufficient to just set an very short (near-zero) solve time limit in your call to solve.

emma58 (commented):
I still think the test could be simpler. We prefer to not use mock if it can be avoided.

PR #3948, opened 2026-05-11
Fix non-rigorous crossed bound results

bernalde:
## Summary

- Add crossed-bound result handling so non-certified crossed bounds return `feasible` and report no certified dual bound instead of manufacturing `optimal` with equal bounds.
- Detect LOA models that may have non-rigorous dual bounds, including non-polynomial expressions and basic nonconvex quadratic structure, while preserving certified behavior for GLOA and convex quadratic LOA cases.
- Apply the same non-certified crossed-bound result semantics to MindtPy GOA.

## Tests run

- `pytest -q pyomo/contrib/gdpopt/tests/test_gdpopt.py::TestGDPoptUnit::test_loa_crossed_bounds_are_not_reported_as_global_optimal pyomo/contrib/gdpopt/tests/test_gdpopt.py::TestGDPoptUnit::test_loa_nonpolynomial_model_may_have_nonrigorous_dual_bound pyomo/contrib/gdpopt/tests/test_gdpopt.py::TestGDPoptUnit::test_loa_distinguishes_basic_quadratic_convexity pyomo/contrib/gdpopt/tests/test_gdpopt.py::TestGDPoptUnit::test_gloa_crossed_bounds_preserve_certified_optimal_behavior pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py::TestMindtPyGOAResults::test_goa_crossed_bounds_are_not_reported_as_global_optimal pyomo/contrib/gdpopt/tests/test_gdpopt.py::TestGDPopt::test_logical_constraints_on_disjuncts_nonlinear_convex pyomo/contrib/gdpopt/tests/test_gdpopt.py::TestGDPopt::test_reverse_numeric_differentiation_in_LOA` -> 7 passed
- `pytest -q pyomo/contrib/gdpopt/tests/test_gdpopt.py::TestGDPoptUnit pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py` -> 42 passed, 2 skipped
- `pytest -q pyomo/contrib/gdpopt/tests/test_gdpopt.py` -> 64 passed, 19 skipped, 3 deselected
- `pytest -q pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py` -> 29 passed
[...]

PR #4013, opened 2026-08-15
Fix non-rigorous crossed bound results

### LINKED ISSUE #3947: GDPopt LOA reports non-rigorous nonconvex bounds as optimal LB=UB
## Summary

GDPopt LOA can report `TerminationCondition.optimal` with `problem.lower_bound == problem.upper_bound` on nonconvex GDPs even when those bounds are not rigorous.

This is not a request for LOA to prove global optimality on nonconvex problems. Pyomo already documents that "For nonconvex problems, LOA may not report rigorous lower/upper bounds." The issue is that the returned `SolverResults` object still uses the unqualified global-looking fields:

- `results.solver.termination_condition = optimal`
- `results.problem.lower_bound = <local incumbent>`
- `results.problem.upper_bound = <same local incumbent>`

For benchmark and downstream tooling, that is indistinguishable from a globally certified solve unless the caller already knows the model is nonconvex and knows this GDPopt LOA caveat.

## Minimal result-finalization reproducer

This reproducer does not require external solvers. It isolates the relevant result semantics: when a LOA run has a local primal incumbent and a non-rigorous OA/discrete bound that has crossed it, the generic GDPopt convergence logic force-updates the dual bound to the primal bound and returns `optimal` with equal bounds.

```python
import logging

import pyomo.environ as pyo
from pyomo.common.timing import default_timer
from pyomo.contrib.gdpopt.loa import GDP_LOA_Solver
from pyomo.opt import SolverResults

solver = GDP_LOA_Solver()
solver.pyomo_results = SolverResults()
solver.pyomo_results.problem.sense = pyo.maximize

# This mimics a nonconvex LOA state:
# - LB is a feasible incumbent from a local NLP subproblem.
# - UB is the OA/discrete bound. For a nonconvex problem this bound is not
#   necessarily globally valid, and it has crossed the incumbent.
solver.LB = 1011.6577899409375
solver.UB = 1000.0
solver.iteration = 1
solver.timing.main_timer_start_time = default_timer()
solver.timing.total = 0.0

config = solver.CONFIG()
config.bound_tolerance = 1e-6
config.logger = logging.getLogger("gdpopt-loa-mwe")
config.logger.handlers[:] = [logging.StreamHandler()]
config.logger.setLevel(logging.INFO)

print("before", solver.LB, solver.UB, solver.pyomo_results.solver.termination_condition)
print("converged", solver.bounds_converged(config))
print("after", solver.LB, solver.UB, solver.pyomo_results.solver.termination_condition)

results = solver._get_final_pyomo_results_object()
print(
    "results",
    results.problem.lower_bound,
    results.problem.upper_bound,
    results.solver.termination_condition,
)
```

Observed output with Pyomo 6.10.0:

```text
        1                      1011.65779    1011.65779      0.00%      0.00
GDPopt exiting--bounds have converged or crossed.
before 1011.6577899409375 1000.0 unknown
converged True
after 1011.6577899409375 1011.6577899409375 optimal
results 1011.6577899409375 1011.6577899409375 optimal
```

The problematic behavior is not that the internal crossed-bound logic exists for algorithms with rigorous bounds. The problematic behavior is that LOA uses the same finalization path even though its dual bound can be non-rigorous on nonconvex models.

## Downstream evidence from GDPLib

This surfaced while benchmarking GDPLib's HDA model:

- GDPLib HDA issue: https://github.com/SECQUOIA/gdplib/issues/67
- GDPLib PR: https://github.com/SECQUOIA/gdplib/pull/130
- GDPLib central benchmark tracker: https://github.com/SECQUOIA/gdplib/issues/60

For the HDA instance, the local LOA benchmark row reported:

```text
strategy: gdpopt.loa
solver profile: gams-local
termination: optimal
lower bound: 1011.6577899409375
upper bound: 1011.6577899409375
iterations: 1
```

However, the HDA model is nonconvex, and other GDPopt/global-capable routes found a better feasible solution:

```text
gdpopt.gloa + gams-local: optimal, objective 1105.3441871686007
gdpopt.ric + gams-local: optimal, objective 1105.3441871686007
gdpopt.ric + gams-gurobi: optimal, objective 1105.3441871686007
gdpopt.ric + gams-baron: optimal, objective 1105.339918385673
```

For a maximization problem, a reported LOA upper bound of `1011.6577899409375` is therefore not a valid global upper bound once a feasible point around `1105.34` is known. The `LB = UB` result gives the appearance of a globally certified incumbent even though it is a local/non-rigorous LOA result.

## Expected behavior

For LOA on models where the OA/discrete bound is not known to be rigorous, GDPopt should avoid returning globally certified-looking result metadata.

Possible approaches:

- Return a non-global termination condition such as `feasible` or `locallyOptimal` when LOA converges only by non-rigorous/crossed bounds on a nonconvex problem.
- Preserve the incumbent as the primal bound but avoid force-setting the dual bound equal to the incumbent when the dual bound is not rigorous.
- Add explicit result metadata indicating that the reported bound is not rigorous.
- Add an option such as `assume_convex=True/False` or `certify_bounds=True/False`, with conservative default result semantics for LOA unless convexity/global validity is asserted.

The important downstream requirement is that JSON/benchmark consumers should not have to infer from `solver.name == "GDPopt ... - LOA"` that `optimal` and `LB = UB` may not mean globally optimal.

## Environment

Observed locally with:

```text
Python 3.12.13
Pyomo 6.10.0
OS/platform: Linux WSL2, glibc 2.39
```

bernalde:
## Summary/Motivation:

When GDPopt LOA is applied to a nonconvex problem, its linearizations are not valid relaxations, so the "dual bound" it accumulates is not a rigorous bound. If those bounds cross, GDPopt currently snaps the dual bound to the primal bound and reports `optimal` with `LB == UB`. That presents a non-rigorous result as a certified global optimum, which is the failure reported in #3947.

The fix reports what is actually known in that case: the incumbent is feasible, and no certified dual bound is available.

This is a replacement for #3948, which GitHub would not allow me to reopen after it was closed. The prior review discussion and history remain available there.

- Add a `_crossed_bounds_are_certified` class flag to `_GDPoptAlgorithm` (default `True`), set to `False` on `GDP_LOA_Solver`. GLOA keeps the existing certified behavior.
- When bounds cross on a solver whose bounds are not certified *and* the model is detected as possibly non-rigorous, terminate with `feasible` instead of `optimal`, and report the uncertified side of the bound as infinite rather than snapping it to the primal bound.
- Add `_problem_may_have_nonrigorous_dual_bound()` in `pyomo/contrib/gdpopt/loa.py`, which inspects the active objective and constraints: anything non-polynomial or of degree > 2 is treated as possibly non-rigorous, quadratic equalities are treated as possibly non-rigorous, and quadratic inequalities are classified by curvature against their bound sense.
- Classify quadratic curvature from the full symmetric Q matrix rather than diagonal terms only, so convex forms with cross terms such as `x**2 + x*y + y**2 <= 4` are still certified.
- Use NumPy `eigvalsh` when available, with a pure-Python Jacobi eigenvalue fallback so the classification also works in Pyomo's NumPy-free test environments.
- Apply the same crossed-bound result semantics to MindtPy GOA.
- Add regressions in `pyomo/contrib/gdpopt/tests/test_gdpopt.py` and `pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py` covering crossed-bound reporting, non-polynomial detection, quadratic convexity classification (including the no-NumPy path), and preservation of certified GLOA behavior.
- Validation performed locally on this branch, rebased onto current `main`:
  - `python -m pytest -q pyomo/contrib/gdpopt/tests/test_gdpopt.py pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py`
    - Result: `96 passed, 19 skipped, 3 deselected in 43.01s`
  - `python -m black --check` on the six changed files
    - Result: passed, `6 files would be left unchanged`
  - `typos --config ./.github/workflows/typos.toml` on the six changed files
    - Result: passed, no findings

## AI-Use Disclosure

or

        - [x] Added and ran tests
    
 **Notes for reviewers (optional):** This replacement PR carries over the implementation and tests from #3948 unchanged, then refreshes the branch against current `main`. The replacement PR description and branch-refresh workflow were prepared with AI assistance and reviewed before posting.

This is the largest of the four GDPopt fixes I have open and it changes reported termination conditions, so it deserves the closest look. Specific points where I would like reviewer judgement:

1. **Behavior change in reported status.** Nonconvex LOA runs that previously returned `optimal` with `LB == UB` will now return `feasible` with one bound infinite. That is the intent of #3947, but it is user-visible and may affect downstream code that keys on `optimal`.
2. **Asymmetry between GDPopt and MindtPy.** GDPopt gates the new behavior on both `not _crossed_bounds_are_certified` and `_problem_may_have_nonrigorous_dual_bound(model)`. The MindtPy GOA path in `bounds_converged()` gates only on `not _crossed_bounds_are_certified`, so any crossed bound there becomes `feasible`. I would like a decision on whether MindtPy should carry the same model-structure gate.
3. **GLOA vs GOA.** GDPopt GLOA keeps `_crossed_bounds_are_certified = True` while MindtPy GOA is set to `False`, even though they are the analogous global algorithms. That was deliberate — GLOA's certified behavior is covered by a regression test — but the inconsistency is worth an explicit maintainer decision.
4. **Conservatism of the classifier.** `_problem_may_have_nonrigorous_dual_bound()` is intentionally conservative: quadratic equality constraints and anything of degree > 2 or non-polynomial are all treated as possibly non-rigorous. Convex problems expressed in ways the classifier cannot certify will lose their `optimal` status.
5. **The pure-Python Jacobi fallback.** It exists so the classification works without NumPy. Sampled eigenvalues were checked against NumPy and a no-NumPy regression test is included, but the rotation cap (`50 * order**2`) returning `None` on non-convergence — which is then treated as "not certified" — is a heuristic worth confirming.

### Legal Acknowledgement


bernalde:
Pushed `f4c13fd52` — this resolves review points 2 and 3 in the PR description above, which flagged an inconsistency between the GDPopt and MindtPy halves of this PR. Rather than leaving them for a reviewer to arbitrate, I traced the cause and generalized the mechanism.

## The inconsistency

The two toolkits have matching algorithm pairs, and the distinguishing property is how each builds its relaxation:

| Algorithm | Relaxation | Rigorous dual bound? | Previously in this PR |
|---|---|---|---|
| GDPopt LOA | linearize at trial points | only if convex | `certified = False` — correct |
| GDPopt GLOA | McCormick envelopes | yes | `certified = True` — correct |
| MindtPy OA | linearize at trial points (`add_oa_cuts`) | only if convex | `certified = True` — wrong |
| MindtPy ECP | linearize at trial points (`add_ecp_cuts`) | only if convex | `certified = True` — wrong |
| MindtPy GOA | McCormick envelopes (`add_affine_cuts`, `use_mcpp = True`) | yes | `certified = False` — wrong |

The flag was on the wrong MindtPy solver. GOA is the McCormick-based algorithm — the direct analogue of GLOA, which this PR deliberately leaves certified — while OA and ECP are the analogues of LOA and were left untouched.

## What changed

- Extracted the curvature/convexity classification out of `GDP_LOA_Solver` into a new shared module, `pyomo/contrib/gdpopt/nonrigorous_bounds.py`.
- Both algorithm base classes now implement `_problem_may_have_nonrigorous_dual_bound()` once: return `False` when the algorithm is certified, otherwise run the model check. Individual solvers only declare `_crossed_bounds_are_certified`, so no algorithm carries its own copy of this logic.
- Moved the MindtPy flag: GOA is certified, OA and ECP are not.
- Gave MindtPy the model-structure gate GDPopt already had, so a crossed bound only downgrades to `feasible` when the model itself could not be certified convex. Previously any crossed bound in MindtPy downgraded unconditionally.

A side benefit of routing through the flag: certified algorithms now skip the convexity scan entirely instead of relying on a base method that returned `False`.

## Effect on reported results

For MindtPy this is a behavior change in both directions, and both directions look like the intended fix:

- **GOA** no longer downgrades to `feasible` on crossed bounds. Its bounds are rigorous, so a crossing is numerical and `optimal` is correct.
- **OA and ECP** now downgrade to `feasible` on crossed bounds, but only for models that fail the convexity check. Convex models are unaffected.

## Tests

`TestMindtPyGOAResults` became `TestMindtPyCrossedBoundResults` and now covers OA and ECP downgrading, OA on a convex model staying `optimal`, GOA preserving certified behavior, the flag matrix itself, and certified algorithms skipping the model scan. The GDPopt no-NumPy test was repointed at the new module.

- `pytest -q pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py pyomo/contrib/gdpopt/tests/` → `115 passed, 35 skipped, 5 deselected`
- `pytest -q pyomo/contrib/mindtpy/tests/` → `58 passed, 32 skipped`
- `black --check` on all nine changed files → `9 files would be left unchanged`
- `typos --config ./.github/workflows/typos.toml pyomo/contrib/gdpopt/ pyomo/contrib/mindtpy/` → clean

## Still worth a maintainer decision

1. **Module location.** `nonrigorous_bounds.py` sits under `pyomo/contrib/gdpopt/` because MindtPy already imports from `pyomo.contrib.gdpopt` in eight modules, so this follows existing precedent. If a neutral location is preferred, it moves cleanly.
2. **`MindtPy_FP_Solver` is left at the default (certified).** The feasibility pump is a primal heuristic rather than a bounding algorithm, and it does not reference `bounds_converged` or the dual bound directly, so I did not classify it. Worth a second opinion.
3. **The original crossing evidence.** The GOA test in the previous revision used primal `1011.6577899409375` against dual `1000.0` — a crossing of roughly 11.7 against a `1e-6` tolerance, which is far too large to be numerical noise. If that came from a real GOA run, then GOA's dual bound was not rigorous in that instance, and the underlying cause is a separate bug in the McCormick cut generation rather than something this flag should paper over. I could not reproduce the run to confirm it, so I have preserved the number in the tests but classified GOA on the basis of what the algorithm is designed to do. If anyone recognizes that instance, it is worth chasing separately.

bernalde:
Following up on open point 3 from my previous comment, where I flagged that the crossed-bound figures used in the tests (primal `1011.6577899409375` against dual `1000.0`, a gap of roughly 11.7 against a `1e-6` tolerance) looked too large to be numerical noise, and asked whether they came from a real MindtPy GOA run that would indicate a non-rigorous GOA dual bound.

**They did not. The point is resolved, and no code change is needed.**

Tracing the provenance:

1. The value originates in the reproducer in #3947, which is a **GDPopt LOA** reproducer. The issue text labels it explicitly — "This mimics a nonconvex LOA state" — and constructs the state by hand:

   ```python
   solver.LB = 1011.6577899409375   # feasible incumbent from a local NLP subproblem
   solver.UB = 1000.0               # non-rigorous OA/discrete bound that has crossed it
   ```

   The precise `LB` is an incumbent objective from GDPLib's HDA model; the `UB` is an arbitrary round number chosen to force the crossing. It is a synthetic fixture for exercising the result-finalization path, not a measurement.

2. The same pair is used by `_setup_crossed_bound_state()` in the GDPopt tests here. The MindtPy test added in the previous revision reused those same two numbers in a GOA-shaped unit test. So it was the LOA fixture copied across, not an independent GOA observation.

3. The number appears nowhere else in Pyomo — only in tests introduced by this PR.

4. Searching Pyomo issues for reports of MindtPy GOA, or MindtPy generally, producing crossed bounds returns nothing.

5. Consistent with this, the GDPLib HDA tracking issue (SECQUOIA/gdplib#67, referenced from #3947) lists LOA among the strategies causing benchmark problems, while GLOA is not checked off.

So there is no evidence that a McCormick-based algorithm ever produced a genuinely crossed bound, and no separate cut-generation bug implied. The reclassification in `f4c13fd52` — GOA certified, OA and ECP not — rests on what each algorithm's relaxation actually guarantees, and nothing in the record contradicts it.

The figures remain in the tests as a synthetic fixture, which is all they ever were. Open points 1 (module location) and 2 (`MindtPy_FP_Solver` left unclassified) from my previous comment still stand for maintainer input.

pyomo/contrib/gdpopt/nonrigorous_bounds.py
```diff
@@ -0,0 +1,184 @@
...
+positively certify as convex is reported as possibly non-rigorous.
+"""
+
+from math import copysign, sqrt
+
+from pyomo.common.collections import ComponentMap
+from pyomo.common.dependencies import numpy as np, numpy_available
+from pyomo.core import Block, Constraint, Objective, minimize, value
+from pyomo.gdp import Disjunct
+from pyomo.repn import generate_standard_repn
```
emma58: Please replace use of `generate_standard_repn` with the `LinearRepnVisitor` or the `QuadraticRepnVisitor`. It has known issues and we are hoping to eventually deprecate it. In this case, it looks like you want the QuadraticRepnVisitor.

pyomo/contrib/gdpopt/nonrigorous_bounds.py
```diff
@@ -0,0 +1,184 @@
...
+    return q_matrix
+
+
+def symmetric_matrix_eigenvalues(matrix):
+    """Return the eigenvalues of a real symmetric matrix.
+
+    Uses NumPy when it is available. Otherwise falls back to a cyclic Jacobi
+    iteration so that curvature classification still works in environments
+    without NumPy. Returns None if the Jacobi iteration does not converge.
+    """
```
emma58: Is it crucial that this work in an environment without numpy? Implementing a fallback seems like a lot of technical debt, and I'm not sure for what gain. It would not be unprecedented in Pyomo to have a piece of functionality that does truly depend on numpy--that would be fine for GDPopt.

pyomo/contrib/gdpopt/nonrigorous_bounds.py
```diff
@@ -0,0 +1,184 @@
...
+    repn = generate_standard_repn(expr, quadratic=True)
+    if not repn.quadratic_coefs:
+        return 0
+
+    q_matrix = _quadratic_matrix(repn)
+    if q_matrix is None:
+        return None
+
+    eigenvalue_tolerance = 1e-10
```
emma58: This should be configurable in the options for the global algorithms

pyomo/contrib/gdpopt/nonrigorous_bounds.py
```diff
@@ -0,0 +1,184 @@
...
+    if is_psd and is_nsd:
+        return 0
+    if is_psd:
+        return 1
+    if is_nsd:
+        return -1
+    return None
+
+
+def model_may_have_nonrigorous_dual_bound(model):
```
emma58: I'd propose renaming this to `model_is_not_certified_convex` or something along those lines since that's what you're actually checking.

pyomo/contrib/gdpopt/nonrigorous_bounds.py
```diff
@@ -0,0 +1,184 @@
...
+    """Return True if linearization-based dual bounds for this model may be invalid.
+
+    A True result means the model could not be certified as convex, so an outer
+    approximation dual bound computed for it must not be treated as rigorous.
+    """
+    for obj in model.component_data_objects(Objective, active=True, descend_into=True):
+        degree = obj.expr.polynomial_degree()
```
emma58: `polynomial_degree` has some known issues also. It may be safer to walk this with the `QuadraticRepnVisitor` and draw your own conclusion from the result.

pyomo/contrib/gdpopt/nonrigorous_bounds.py
```diff
@@ -0,0 +1,184 @@
+# ____________________________________________________________________________________
+#
+# Pyomo: Python Optimization Modeling Objects
+# Copyright (c) 2008-2026 National Technology and Engineering Solutions of Sandia, LLC
+# Under the terms of Contract DE-NA0003525 with National Technology and Engineering
+# Solutions of Sandia, LLC, the U.S. Government retains certain rights in this
+# software.  This software is distributed under the 3-clause BSD License.
+# ____________________________________________________________________________________
+
+"""Detection of models whose outer approximation dual bounds may not be rigorous.
```
emma58: This file seems misnamed to me, it's actually entirely about convexity detection.

pyomo/contrib/gdpopt/algorithm_base_class.py
```diff
@@ -57,6 +61,8 @@ def __init__(self, **kwds):
 
         self.incumbent_boolean_soln = None
         self.incumbent_continuous_soln = None
+        self._nonrigorous_crossed_bounds = False
+        self._nonrigorous_dual_bound_possible = False
```
emma58: I'd advocate renaming these to something about whether or not convexity is certified, but if you prefer this, that's okay.

pyomo/contrib/gdpopt/algorithm_base_class.py
```diff
@@ -37,6 +40,7 @@
 class _GDPoptAlgorithm:
     CONFIG = ConfigBlock("GDPopt")
     _add_common_configs(CONFIG)
+    _crossed_bounds_are_certified = True
```
emma58: This name is confusing to me. Certified what? Seems like this should be `_treat_problem_as_convex` or something like that.

pyomo/contrib/gdpopt/tests/test_gdpopt.py
```diff
@@ -78,6 +83,108 @@
...
+        m = ConcreteModel()
+        m.x = Var(bounds=(0, 4), initialize=0.2)
+        m.y = Var(bounds=(-2, 2), initialize=0)
+        m.nonconvex = Constraint(expr=m.y <= sin(3 * m.x) + 0.2 * m.x)
+        m.choose_region = Disjunction(expr=[[m.x <= 2], [m.x >= 2]], xor=True)
+        m.obj = Objective(expr=m.y, sense=maximize)
+
+        self.assertTrue(GDP_LOA_Solver()._problem_may_have_nonrigorous_dual_bound(m))
+
+    def test_loa_distinguishes_basic_quadratic_convexity(self):
```
emma58, line 126: I'm surprised you don't have to guard this test on some of the ones below for numpy being available?

pyomo/contrib/gdpopt/tests/test_gdpopt.py
```diff
@@ -78,6 +83,108 @@
...
+        solver.UB = 1000.0
+        solver.iteration = 1
+        solver._nonrigorous_dual_bound_possible = True
+        solver.timing.main_timer_start_time = default_timer()
+        solver.timing.total = 0.0
+
+        config = solver.CONFIG()
+        config.bound_tolerance = 1e-6
+        config.logger = logging.getLogger(__name__)
+        return config
```
emma58, line 100: This is fine for checking the `bounds_converged` method and a couple other things, but I think a stronger test would be to actually set up a call to solve on a nonconvex problem where this will happen and see it through end to end.

pyomo/contrib/gdpopt/tests/test_gdpopt.py
```diff
@@ -78,6 +83,108 @@
...
+        try:
+            self.assertFalse(
+                GDP_LOA_Solver()._problem_may_have_nonrigorous_dual_bound(convex)
+            )
+            self.assertTrue(
+                GDP_LOA_Solver()._problem_may_have_nonrigorous_dual_bound(nonconvex)
+            )
+        finally:
+            nonrigorous_bounds_module.numpy_available = original_numpy_available
+
```
emma58: This is unnecessary: Just import `numpy_available` here and guard the whole test to only run if it is available.

emma58 (commented):
Thank you: This is a nice addition to keep users using from shooting themselves in the foot when we can guarantee a problem is convex. I had quite a few comments: I find all the names confusing--I'd vote for names that actually say something about certified or not certified convexity rather than about the crossed bounds. I also would strongly prefer that the implementation not depend on `generate_standard_repn` and instead use the `QuadraticRepnVisitor`. Also I'm curious why it is imperative that this work without numpy?

pyomo/contrib/gdpopt/nonrigorous_bounds.py
bernalde: Replaced generate_standard_repn with QuadraticRepnVisitor in convexity.py.

pyomo/contrib/gdpopt/nonrigorous_bounds.py
bernalde: Removed the custom Jacobi fallback. Quadratic curvature checks now use numpy.linalg.eigvalsh

pyomo/contrib/gdpopt/nonrigorous_bounds.py
bernalde: Added eigenvalue_tolerance to the GDPopt and MindtPy solver options, with a default of 1e-10.

pyomo/contrib/gdpopt/nonrigorous_bounds.py
bernalde: Renamed the check to model_is_not_certified_convex.

pyomo/contrib/gdpopt/nonrigorous_bounds.py
bernalde: Removed the polynomial_degree() checks. QuadraticRepnVisitor now determines whether each expression is quadratic or contains remaining nonlinear terms.

pyomo/contrib/gdpopt/nonrigorous_bounds.py
bernalde: Renamed nonrigorous_bounds.py to convexity.py.

pyomo/contrib/gdpopt/algorithm_base_class.py
bernalde: Renamed the state to _model_is_not_certified_convex and _bounds_crossed_without_certified_convexity.

pyomo/contrib/gdpopt/algorithm_base_class.py
bernalde: Replaced the ambiguous certification flag with _requires_model_convexity. LOA, OA, and ECP set it to True; GLOA and GOA retain the default False.

pyomo/contrib/gdpopt/tests/test_gdpopt.py
bernalde: Added numpy_available guards to the tests that require eigendecomposition.

pyomo/contrib/gdpopt/tests/test_gdpopt.py
bernalde: Added test_loa_nonconvex_solve_reports_crossed_bounds_as_feasible. It runs GDPopt LOA with GLPK and Ipopt and verifies that crossed uncertified bounds produce feasible, preserve the incumbent bound, and leave the uncertified side infinite.

pyomo/contrib/gdpopt/tests/test_gdpopt.py
bernalde: Removed the Jacobi-specific test and guarded the NumPy-dependent curvature tests with numpy_available.

bernalde:
Addressed in 4f51f8ea3: the naming now describes convexity certification, expression inspection uses QuadraticRepnVisitor, the custom NumPy fallback was removed, the eigenvalue tolerance is configurable, and an end-to-end nonconvex LOA regression test was added.

pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py
```diff
@@ -264,6 +266,90 @@ def __init__(self, **kwargs):
...
+    solver.best_solution_found_time = None
+    solver.primal_integral = 0.0
+    solver.dual_integral = 0.0
+    solver.primal_dual_gap_integral = 0.0
+    solver.update_result()
+
+
+class TestMindtPyCrossedBoundResults(unittest.TestCase):
+    def test_oa_crossed_bounds_are_not_reported_as_global_optimal(self):
+        from pyomo.contrib.mindtpy.outer_approximation import MindtPy_OA_Solver
```
emma58: Can you please move this import and the others below to the top of the file?

emma58 (commented):
One last comment on import statements, but otherwise this looks good, thank you!

bernalde:
Addressed Emma's final import-placement review in [`bb76bc3b3`](https://github.com/Pyomo/pyomo/commit/bb76bc3b324f50cef3d92b55cec9ccb2727b0a6f).

- Moved all local MindtPy algorithm and solver imports in `test_mindtpy_no_discrete.py` to module scope. No items from the latest review were intentionally left unaddressed.
- Local verification: `pytest -q pyomo/contrib/mindtpy/tests/` — `57 passed, 32 skipped`; `black . --check --diff` — `1,830 files would be left unchanged`.
- At the five-minute bounded cutoff, exact-head CI had `lint/style-and-typos` and `linux/3.10/bare-env` successful. The remaining Actions matrix jobs and Read the Docs were still queued or running, with no failures reported.
- GitHub currently reports `REVIEW_REQUIRED`; review threads remain for maintainer resolution.

pyomo/contrib/mindtpy/tests/test_mindtpy_no_discrete.py
bernalde: Addressed in [`bb76bc3b3`](https://github.com/Pyomo/pyomo/commit/bb76bc3b324f50cef3d92b55cec9ccb2727b0a6f): I moved `_MindtPyAlgorithm` and the OA, ECP, and GOA solver classes to module-level imports and removed all local imports from this test file. The MindtPy suite passes locally (`57 passed, 32 skipped`); exact-head CI had no failures at the bounded cutoff and remains in progress.

bernalde:
@emma58, the final import-placement comment is addressed in [`bb76bc3b3`](https://github.com/Pyomo/pyomo/commit/bb76bc3b324f50cef3d92b55cec9ccb2727b0a6f). Could you take another look when convenient? The current-head CI matrix is still running; completed jobs are green so far.

PR #4022, opened 2026-08-17
Remove expression `_args_` tuple

jsiirola:
## Summary/Motivation:
@bknueven reached out a couple months ago pointing out that the recent work to standardize the internals of the expression system (and in particular, that `LinearExpression` should more directly inherit from `SumExpression`) had the indirect effect of tripling the memory used by a `LinearExpression` object:

> I traced the issue we are having with constraints and memory, and I think it goes back to changes made to `LinearExpression` a few years back. When I calculate the memory needed to store a single (coefficient, variable) entry in a `LinearExpression` object, I come up with 120 bytes per-entry for our use case:
>
> ```
> MonomialTermExpression : 40 bytes
>     MonomialTermExpression._args_ : 40 bytes
>           pointer-to-Pyomo-var: 8 bytes
>           pointer-to-Python-double: 8 bytes
>                     Python-double: 24 bytes
> ```
> Blow this up to 40000 entries per row and a few hundred rows, and we’re in GBs of memory.
>
> In the old implementation, `LinearExpression` would store two lists, which for the marginal entry in this use case would cost only 40 bytes instead of 120 bytes.

@bknueven's request was to revert the `LinerExpression` storage model back to using a pair of lists.  The challenge there is that the "pair of lists" model doesn't gracefully handle incrementally building `LinearExpression` objects through the normal operator overloading process (in large part because of challenges with handling constants that appear within the summation) -- and the fact that Pyomo's expression system now "automatically" recognizes and creates `LinearExpression` objects is a big part of the benefit from that previous rewrite.

As a result, this PR does not re-implement the "pair of lists" data structure (nor does it resolve all the memory degradation).  Instead, it proposes a compromise that removes the `_args_` tuple from `MonomialTermExpression` (and all other Expression objects).  Storage now depends on the Expression object:
- unary operators store the single argument in the `_arg` attribute
- binary operators store the arguments in the `_l_arg` and `_r_arg` attributes
- some specialized ternary operators (e.g., Expr_IF) store the operands in named attributes
- n-ary operators continue to store the arguments in either a tuple or list as the `_args` attribute.

This reduces the marginal memory overhead from 120 bytes/entry to 80 bytes per entry.

It also significantly improves model creation time (and marginally improves the lp and nl writers).  In my opinion, the slight performance degradation in the GAMS writer is OK, as that writer needs to be completely redesigned (to bring it in line with the current LP/NL writers).

image

- remove standard `_args_` attribute from all expression classes
- implement custom (tuple-free) storage solutions for 1, 2, and some 3-operand expression nodes.
- standardize the CP expression system to leverage / reuse more of the infrastructure developed for the numeric expression system
- remove most (but not all) instances of accessing private argument attributes outside of the class that defines them (and the core expression generation callbacks)
- clean up some inconsistencies between numeric and boolean expression systems (e.g., `value` should be a property)

## AI-Use Disclosure

- [X] **AI tools were NOT used during the preparation of this PR**

or

- [ ] **AI tools contributed to the development of this PR**
    
        - [ ] Added and ran tests
    
### Legal Acknowledgement


jsiirola:
I do think there there is still a need / use for a more memory-efficient expression object for @bknueven's use case, but I think that should be done separately (probably through a re-thought MatrixConstraint expression object).

pyomo/contrib/cp/scheduling_expr/sequence_expressions.py
```diff
@@ -36,14 +36,13 @@ class FirstInSequenceExpression(BooleanExpression):
                       SequenceVar
     """
 
-    def nargs(self):
-        return 2
+    __slots__ = ('_l_arg', '_r_arg')
```
mrmundt, line 39: For `left` and `right`?

pyomo/contrib/cp/scheduling_expr/sequence_expressions.py
```diff
@@ -65,14 +61,26 @@ class BeforeInSequenceExpression(BooleanExpression):
...
-                      IntervalVar that must be after, and the SequenceVar
+    Parameters
+    ----------
+    args : tuple["IntervalVar", "IntervalVar", "SequenceVar"]
+
+        The child nodes, consisting of the predecessor IntervalVar, the successor
+        IntervalVar, and the SequenceVar
     """
 
+    __slots__ = ('_before', '_after', '_sequence')
```
mrmundt, line 72: You use before/after here but predecessor/successor in L#95. It would be better to pick one probably, considering that you are referring to them in the same way?

pyomo/contrib/solver/tests/solvers/gurobi_to_pyomo_expressions.py
```diff
@@ -26,62 +27,97 @@
...
+            fcn, args = grb_op_to_pyo[op]
+            if dat.__class__ is list:
+                dat.reverse()
+            node = fcn(dat, *args)
+        # Data holds "-1" for any operators.  If parent is anything
+        # other than -1, then that is point to an operator whose data
+        # starts with -1.  We can safely replace that entry with the
+        # list of processed operands.
+        #
+        # Note also, the parent of the root node (i.e., parent[0] is -1,
```
mrmundt, line 115:
```suggestion
        # Note also, the parent of the root node (i.e., parent[0]) is -1,
```

pyomo/core/base/expression.py
```diff
@@ -33,34 +33,46 @@
...
 
+    def arg(self, i):
+        """Return the i-th child node.
+
+        Parameters
+        ----------
+        i: int
+            Index of the child argument to return
+
+        Returns: The i-th child node.
```
mrmundt, line 64: Feels a bit redundant to the main description 😛

pyomo/core/expr/__init__.py
```diff
@@ -208,4 +206,15 @@
     "Please import expression symbols from pyomo.core.expr",
     version='6.6.2',
 )
-del moved_module
+relocated_module_attribute(
+    'UnaryBooleanExpression',
+    'pyomo.core.expr.logical_expr.UnaryBooleanExpression',
+    version='6.10.1.dev0',
```
mrmundt, line 212: Isn't it `6.10.2.dev0`?

pyomo/core/expr/logical_expr.py
```diff
@@ -304,26 +297,34 @@ def count_if(*args):
 
     Usage: count_if(m.Y1, m.Y2, ...)
     """
-    return CountIfExpression(list(_flattened_boolean_args(args)))
+    return CountIfExpression(tuple(_flattened_boolean_args(args)))
 
 
+@deprecated(
+    "UnaryBooleanExpression is no longer used; see UnaryExpression_Mixin",
+    version="6.10.1.dev0",
```
mrmundt, line 305: Same comment as earlier - I think it should be `6.10.2.dev0`

mrmundt (changes requested):

PR #4029, opened 2026-08-23
GAMS writer: map non-finite float values to GAMS special values (NA / INF)

### LINKED ISSUE #3911: Handling of float('nan') especially in GAMSWriter
## Summary
TLDR: Variable values are converted from float('nan') to nan. However, nan doesn't mean anything in GAMS. It should be converted to NA instead. 

Here is a minimal example (can be run with GAMS demo license)
```
import pyomo.environ as pyo

# Create a simple model
model = pyo.ConcreteModel()

# Define a variable and explicitly set it to NaN
model.x = pyo.Var(initialize=float('nan'))

# Add a dummy objective or constraint so the solver has something to do
model.obj = pyo.Objective(expr=model.x**2, sense=pyo.minimize)
model.c = pyo.Constraint(expr=model.x >= 1)

# Set up the GAMS solver link
# Note: Ensure 'gams' is in your PATH or specify the executable path
solver = pyo.SolverFactory('gams')
results = solver.solve(model, tee=True , keepfiles=True, tmpdir="tmp")
```

Pyomo produces a warning but the error comes from GAMS (that the symbol nan is undefined):

```
WARNING (W1001): Setting Var 'x' to a value `nan` (float) not in domain Reals.
```

Pyomo version: 6.10.1
Python version: 3.13.4
Operating system: Windows
Solver (if applicable): GAMS

### Additional information

At first, I don't think this is a bigger issue for Pyomo beyond just the GAMS solver. A couple of conflicting arguments come to my mind:
- If the variable initialized to float('nan') is eventually not part of the model after presolve, this might be OK and this could be why Pyomo chooses to throw a warning for this instead of an error.
- GAMS is in general stricter and always throws an error but I got the same error by directly calling IPOPT from Pyomo. I wonder if there are any cases where the solver automatically switches float('nan') to a default initial value without complaining. If there are no such solvers, it could be better to catch this early and throw an error.

A potential fix: instead of using `ftoa()` for gams_writer.py, use a wrapper `gams_ftoa()` that uses `ftoa` and switches `nan` to `NA`

cjck944084735-dot:
## Summary/Motivation:

The GAMS writer emits Python's string rendering of non-finite floats (`nan`, `inf`, `-inf`), which are **not valid GAMS numeric literals**. Exporting a model whose variables or fixed expressions hold such values produces a `.gms` file that GAMS rejects with errors like `the symbol nan is undefined` (reported in #3911). All numeric output in `pyomo/repn/plugins/gams_writer.py` goes through `ftoa()` (format `'%.17g'`), which renders non-finite floats as `nan`/`inf`, e.g. on current `main` with default `warmstart=True`:

```gams
c2.. GAMS_OBJECTIVE =e= power(x1, 2) + x2 + nan ;
x1.l = nan;
x2.l = inf;
```

GAMS's special numeric literals are `NA` and `INF`/`-INF`, so every one of those lines is rejected by GAMS. This hits all value-emitting paths of the writer: constraint bounds, expression constants (fixed Vars/Params), warmstart `.l` values, and fixed-variable `.fx` values.

- Add a `gams_ftoa()` wrapper that post-processes the `ftoa()` result: `nan` maps to GAMS `NA`, and `±inf` maps to `±INF` (including the parenthesized `(-inf)` branch).
- Route all 18 numeric output sites of the GAMS writer through `gams_ftoa()`.
- Add two regression tests: `test_nonfinite_var_to_string` (expression path with fixed `nan`/`inf`/`-inf` Vars) and `test_nonfinite_var_values_write_gams` (full model write asserting `.l` warmstart values, the fixed-expression constant, and that no bare `nan`/`inf` appears in the exported file).

After the change the same model exports as:

```gams
c2.. GAMS_OBJECTIVE =e= power(x1, 2) + x2 + NA ;
x1.l = NA;
x2.l = INF;
```

Two notes on float semantics that the tests document: `-1 * nan` is `nan` (the sign is lost), so `- z` with `z = nan` exports as `+ NA`; and `-inf` values in monomials are parenthesized by `ftoa`, hence `+ (-INF)`.

Local verification: full existing suite `pyomo/repn/tests/gams/` passes (24 passed, 1 skipped), and `black --check` is clean.

**Note for reviewers:** `pyomo/repn/plugins/gams_writer_v2.py` has the same pattern for warmstart values (`ostream.write(f"{v}.l = {pyomo_v.value};\n")`, line ~537). I left it untouched to keep this PR focused on the default `gams` writer from the issue; happy to apply the equivalent mapping there in this PR or a follow-up, whichever the maintainers prefer.

## AI-Use Disclosure

or

        - [x] Added and ran tests

 **Notes for reviewers (optional):** Where should reviewers focus? (1) The `gams_ftoa()` mapping of `nan`/`inf` to `NA`/`INF` — please confirm `NA` (rather than an error) is the desired semantic for undefined variable values in exported GAMS files. (2) Whether `gams_writer_v2` should be fixed in this PR or a follow-up.

### Legal Acknowledgement


blnicho:
@cjck944084735-dot thank you for the bug fix. We welcome contributions from the community but please make sure to review our contributor guide: https://pyomo.readthedocs.io/en/latest/contribution_guide.html

In particular, we require all contributors to use our PR template so this will be closed by policy. Please feel free to reopen with the correct template and we will happily review it.

cjck944084735-dot:
@blnicho thank you for the pointer and sorry for missing the template! I've reopened the contribution as #4033 from the same branch, now using the PR template (including the AI-use disclosure and legal acknowledgement). Happy to adjust anything per the contributor guide.

PR #4031, opened 2026-08-27
Make GDP enumeration lazy

### LINKED ISSUE #3953: GDPopt enumerate materializes all discrete solutions before checking time_limit
## Summary

`gdpopt.enumerate` materializes the complete list of discrete solutions before it checks `time_limit` or `iterlim`. For GDPs with many disjunctions, this can exhaust memory or appear to ignore the configured time limit before any subproblem solve starts.

The issue is in the setup phase of `GDP_Enumeration_Solver._solve_gdp()`: it does

```python
discrete_solns = list(
    self._discrete_solution_iterator(...)
)
self.num_discrete_solns = len(discrete_solns)
for soln in discrete_solns:
    if self.reached_time_limit(config) or self.reached_iteration_limit(config):
        break
```

so the time-limit check happens only after the full Cartesian product has been built.

This solver-free example forces the time-limit predicate to be true before enumeration starts, then raises if the discrete-solution iterator is still called. It demonstrates the ordering issue without requiring a large model, GAMS, Gurobi, IPOPT, or another external solver.

```console
$ python enumerate_time_limit_mwe.py
```

```python
# enumerate_time_limit_mwe.py
import pyomo.environ as pyo
from pyomo.contrib.gdpopt.enumerate import GDP_Enumeration_Solver

print("has enumerate solver", pyo.SolverFactory("gdpopt.enumerate").available())


def fail_if_enumerated(self, *args, **kwargs):
    raise RuntimeError("enumeration happened before the time-limit check")


# Simulate entering _solve_gdp after the time limit has already expired.
GDP_Enumeration_Solver.reached_time_limit = lambda self, config: True
GDP_Enumeration_Solver._discrete_solution_iterator = fail_if_enumerated

m = pyo.ConcreteModel()
m.x = pyo.Var(bounds=(0, 2))
m.d1 = Disjunct()
m.d2 = Disjunct()
m.d1.c = pyo.Constraint(expr=m.x <= 0.5)
m.d2.c = pyo.Constraint(expr=m.x >= 1.5)
m.obj = pyo.Objective(expr=m.x)

pyo.SolverFactory("gdpopt.enumerate").solve(m, time_limit=1, tee=False)
```

### Error Message

The current behavior calls `_discrete_solution_iterator()` before returning on the already-expired time limit:

```console
$ python enumerate_time_limit_mwe.py
has enumerate solver True
  File "enumerate_time_limit_mwe.py", line 25, in <module>
    pyo.SolverFactory("gdpopt.enumerate").solve(m, time_limit=1, tee=False)
  File ".../pyomo/contrib/gdpopt/enumerate.py", line 71, in solve
    return super().solve(model, **kwds)
  File ".../pyomo/contrib/gdpopt/algorithm_base_class.py", line 128, in solve
    self._solve_gdp(model, config)
  File ".../pyomo/contrib/gdpopt/enumerate.py", line 128, in _solve_gdp
    self._discrete_solution_iterator(
  File "enumerate_time_limit_mwe.py", line 10, in fail_if_enumerated
    raise RuntimeError("enumeration happened before the time-limit check")
RuntimeError: enumeration happened before the time-limit check
```

For a real downstream model with 32 active binary disjunction choices, this same eager `list(...)` construction attempts to materialize `2**32` discrete choices before the time limit can be honored, which can exhaust memory or make the machine unresponsive.

Pyomo version: 6.10.0
Python version: 3.12.13
Operating system: Linux WSL2, glibc 2.39
How Pyomo was installed (PyPI, conda, source): conda-forge package through Pixi
Solver (if applicable): not applicable for the MWE; downstream evidence used GAMS/DICOPT, GAMS/IPOPTH, and GAMS/Gurobi role solvers

### Additional information

Expected behavior: `gdpopt.enumerate` should not need to build the entire discrete solution list before enforcing `time_limit` or `iterlim`. It should iterate lazily, or otherwise check limits while generating choices. This would avoid memory exhaustion and would make the documented time limit meaningful for large discrete spaces.

Downstream context from GDPlib:

- GDPlib model issue: https://github.com/SECQUOIA/gdplib/issues/71
- GDPlib PR with the model-side mitigation: https://github.com/SECQUOIA/gdplib/pull/140
- GDPlib central benchmark tracker: https://github.com/SECQUOIA/gdplib/issues/60

In GDPlib's `mod_hens` model, issue SECQUOIA/gdplib#71 exposed this behavior. Before the model-side mitigation, the default formulation had 32 active exchanger disjunctions, 20 of which represented structurally fixed absent choices. `gdpopt.enumerate` appeared not to respect a 60-second time limit because it was materializing the full discrete solution list first. After pruning those fixed choices from the active GDP structure in SECQUOIA/gdplib#140, the same 60-second `gdpopt.enumerate` probe returned normally with `maxTimeLimit` at about 60.10 seconds and reported 12 disjunctions.

This is distinct from, but adjacent to, the LBB time-limit finalization issue reported in Pyomo/pyomo#3941.

alvaroborras:
## Summary/Motivation:

`gdpopt.enumerate` previously materialized every discrete solution before checking the time and iteration limits. For large discrete spaces, this could exhaust available memory before solving began.

This change counts the discrete solution space without materializing it and requests solutions lazily after checking the configured limits. It preserves correct completion behavior and supports reusing the solver instance.

- Enumerate discrete solutions lazily, checking time and iteration limits before requesting each solution.
- Add concise regression tests for large discrete spaces, completion behavior, and solver reuse.

## AI-Use Disclosure

or

        - [ ] Added and ran tests
    

### Legal Acknowledgement

### DIFF
--- a/pyomo/contrib/gdpopt/enumerate.py
+++ b/pyomo/contrib/gdpopt/enumerate.py
@@ -7,9 +7,7 @@
 # software.  This software is distributed under the 3-clause BSD License.
 # ____________________________________________________________________________________
 
-import math
 from itertools import product
-from operator import index
 
 from pyomo.common.collections import ComponentSet
 from pyomo.common.config import document_kwargs_from_configdict
@@ -29,11 +27,7 @@ from pyomo.contrib.gdpopt.create_oa_subproblems import (
     get_subproblem,
 )
 from pyomo.contrib.gdpopt.solve_subproblem import solve_subproblem
-from pyomo.contrib.gdpopt.util import (
-    fix_discrete_solution_in_subproblem,
-    time_code,
-    get_main_elapsed_time,
-)
+from pyomo.contrib.gdpopt.util import fix_discrete_solution_in_subproblem, time_code
 
 from pyomo.core import value
 from pyomo.opt import TerminationCondition as tc
@@ -95,22 +89,6 @@ class GDP_Enumeration_Solver(_GDPoptAlgorithm):
                             integer_realization,
                         )
 
-    # Override logging so that we print progress in terms of the number of
-    # iterations needed to fully enumerate the discrete space.
-    def _log_current_state(self, logger, subproblem_type, primal_improved=False):
-        star = "*" if primal_improved else ""
-        logger.info(
-            self.log_formatter.format(
-                "{}/{}".format(self.iteration, self.num_discrete_solns),
-                subproblem_type,
-                self.LB,
-                self.UB,
-                self.relative_gap(),
-                get_main_elapsed_time(self.timing),
-                star,
-            )
-        )
-
     def _solve_gdp(self, original_model, config):
         util_block = self.original_util_block
         # From preprocessing to make sure this *is* a GDP, we already have
@@ -124,29 +102,16 @@ class GDP_Enumeration_Solver(_GDPoptAlgorithm):
 
         subproblem, subproblem_util_block = get_subproblem(original_model, util_block)
 
-        disjunctions = subproblem_util_block.disjunction_list
-        non_indicator_boolean_vars = (
-            subproblem_util_block.non_indicator_boolean_variable_list
-        )
-        discrete_vars = subproblem_util_block.discrete_variable_list
-
-        self.num_discrete_solns = math.prod(
-            len(disjunction.disjuncts) for disjunction in disjunctions
-        )
-        if config.force_subproblem_nlp:
-            self.num_discrete_solns *= 2 ** len(non_indicator_boolean_vars) * math.prod(
-                max(0, index(v.ub) - index(v.lb) + 1) for v in discrete_vars
-            )
-
         if self.reached_time_limit(config) or self.reached_iteration_limit(config):
             return
-        discrete_solns = self._discrete_solution_iterator(
-            disjunctions, non_indicator_boolean_vars, discrete_vars, config
-        )
-        for _ in range(self.num_discrete_solns):
+        for soln in self._discrete_solution_iterator(
+            subproblem_util_block.disjunction_list,
+            subproblem_util_block.non_indicator_boolean_variable_list,
+            subproblem_util_block.discrete_variable_list,
+            config,
+        ):
             if self.reached_time_limit(config) or self.reached_iteration_limit(config):
                 return
-            soln = next(discrete_solns)
 
             self.iteration += 1
 
--- a/pyomo/contrib/gdpopt/enumerate.py
+++ b/pyomo/contrib/gdpopt/enumerate.py
@@ -7,7 +7,9 @@
 # software.  This software is distributed under the 3-clause BSD License.
 # ____________________________________________________________________________________
 
+import math
 from itertools import product
+from operator import index
 
 from pyomo.common.collections import ComponentSet
 from pyomo.common.config import document_kwargs_from_configdict
@@ -110,8 +112,6 @@ class GDP_Enumeration_Solver(_GDPoptAlgorithm):
         )
 
     def _solve_gdp(self, original_model, config):
-        logger = config.logger
-
         util_block = self.original_util_block
         # From preprocessing to make sure this *is* a GDP, we already have
         # lists of:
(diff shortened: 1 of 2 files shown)


pyomo/contrib/gdpopt/enumerate.py
```diff
@@ -124,19 +124,30 @@ def _solve_gdp(self, original_model, config):
...
+            subproblem_util_block.non_indicator_boolean_variable_list
+        )
+        discrete_vars = subproblem_util_block.discrete_variable_list
+
+        self.num_discrete_solns = math.prod(
+            len(disjunction.disjuncts) for disjunction in disjunctions
+        )
+        if config.force_subproblem_nlp:
+            self.num_discrete_solns *= 2 ** len(non_indicator_boolean_vars) * math.prod(
+                max(0, index(v.ub) - index(v.lb) + 1) for v in discrete_vars
```
emma58: While it would make sense for the bounds of integer variables to be integer, Pyomo doesn't enforce that. Your call to `index` here will raise an error in that case. I think instead you should round correctly and cast the result to int.

pyomo/contrib/gdpopt/enumerate.py
```diff
@@ -124,19 +124,30 @@ def _solve_gdp(self, original_model, config):
...
+
+        if self.reached_time_limit(config) or self.reached_iteration_limit(config):
+            return
+        discrete_solns = self._discrete_solution_iterator(
+            disjunctions, non_indicator_boolean_vars, discrete_vars, config
         )
-        self.num_discrete_solns = len(discrete_solns)
-        for soln in discrete_solns:
-            # We will interrupt based on time limit or iteration limit:
+        for _ in range(self.num_discrete_solns):
```
emma58: On second thought, there's no reason to calculate `num_discrete_solutions` ever. We could just iterate through the solutions in `self._discrete_solution_iterator` here.

emma58 (changes requested):
Thank you for this--I agree it's a good change since someone could always want to cut off something unreasonable at a time or iteration limit. I think the implementation could be simpler though: There's no need to calculate the number of discrete solutions, we can just iterate through the generator until we hit a termination condition or are done.

alvaroborras:
Thanks for the feedback. I agree that precomputing num_discrete_solns is unnecessary. I’ve updated the implementation to iterate directly over _discrete_solution_iterator, so the index calculation (and the bounds issue you identified) is removed entirely.
  
I also removed the custom _log_current_state because its only difference from the base implementation was displaying iteration/num_discrete_solns. Without that count, the override would either reference a removed attribute or duplicate the base logger (and show AttributeError).

pyomo/contrib/gdpopt/enumerate.py
```diff
@@ -93,25 +89,7 @@ def _discrete_solution_iterator(
...
-            self.log_formatter.format(
-                "{}/{}".format(self.iteration, self.num_discrete_solns),
-                subproblem_type,
-                self.LB,
-                self.UB,
-                self.relative_gap(),
-                get_main_elapsed_time(self.timing),
-                star,
-            )
-        )
```
emma58, line 110: Oh, I see. Sorry, I wasn't being very careful. It may be worth keeping the calculations for how many discrete solutions there are so that unsuspecting users get this info in the log. Do you agree?

emma58 (commented):
This looks much better, thank you! On question though--I forgot that we overrode the progress logger to show progress towards complete enumeration. I think that may have been wise and we should add back in the calculation of how many solutions there are for that. Do you agree?

alvaroborras:
I think the cleanest approach is to keep iterating directly over _discrete_solution_iterator, while calculating num_discrete_solns separately only for the progress logger (which I would keep). The count would be computed arithmetically without materializing any solutions, and would use ceil(lb) and floor(ub) for integer-variable bounds. Does that look right to you?

emma58:
@alvaroborras, yes, I like that plan!

pyomo/contrib/gdpopt/enumerate.py
```diff
@@ -124,19 +125,28 @@ def _solve_gdp(self, original_model, config):
...
-            # We will interrupt based on time limit or iteration limit:
+        discrete_vars = subproblem_util_block.discrete_variable_list
+
+        self.num_discrete_solns = math.prod(
+            len(disjunction.disjuncts) for disjunction in disjunctions
+        )
+        if config.force_subproblem_nlp:
+            self.num_discrete_solns *= 2 ** len(non_indicator_boolean_vars) * math.prod(
+                max(0, math.floor(v.ub) - math.ceil(v.lb) + 1) for v in discrete_vars
+            )
```
emma58, line 146: Sorry, this isn't actually a bug introduced by this PR, but I just caught it: `add_discrete_variable_list` doesn't check for bounds, so this will error if `v.lb` or `v.ub` is `None`. I think we should check that case here because this is the first place we'll catch it. It will also be a problem in `_discrete_solution_iterator`.

emma58 (commented):
This looks good, thank you! On comment below because I realized we don't correctly handle unbounded discrete variables. That's not this PR's fault, but is worth fixing before we merge if you're willing.

alvaroborras:
Thanks for catching this. I added an explicit check for missing lb or ub before calculating the enumeration size or constructing the iterator, and I also added a regression test for this case.
Let me know if it looks good now.

alvaroborras:
Thanks, @emma58  I’ve review again all your feedback but I think everything looks OK and CI is passing.
Could you take a look and see if there is anything else missing?

pyomo/contrib/gdpopt/enumerate.py
mrmundt: @alvaroborras - I am going through unresolved comments (and generally just trying to help move some PRs forward...). I don't think this was resolved, correct?

mrmundt (approved):
From a code quality and style perspective, this is great. Pending re-review by @emma58 .

PR #4034, opened 2026-08-29
Fix LinearRepnVisitor to substitute fixed vars already in var_map

### LINKED ISSUE #3851: LinearStandardFormCompiler sometimes says expression are nonlinear even when variables are fixed
## Summary

Sometimes, the `LinearStandardFormCompiler` throws an error saying that an expression is nonlinear even if the variables that would make it nonlinear are fixed. 

In order to get a simple example that illustrates the problem, I created an instance of `LinearRepnVisitor` the same way that `LinearStandardFormCompiler` makes one. See [this](https://github.com/Pyomo/pyomo/blob/87f383ecc381b66de84615cb9a5c0c0d9dbfe911/pyomo/repn/plugins/standard_form.py#L332) line and the few lines before it. 

```
   ...: from pyomo.repn.linear import LinearRepnVisitor
   ...: from pyomo.repn.util import TemplateVarRecorder, SortComponents
   ...: m = pe.ConcreteModel()
   ...: m.a = pe.Set(initialize=[1,2,3])
   ...: m.x = pe.Var(m.a)
   ...: e1 = m.x[1] * (m.x[2] + m.x[3])
   ...: e2 = m.x[1] - 1/m.x[2]
   ...: m.x[2].fix(2)
   ...: sorter = SortComponents.ORDERED_INDICES
   ...: vmap = {}
   ...: vr = TemplateVarRecorder(vmap, sorter)
   ...: visitor = LinearRepnVisitor(subexpression_cache={}, var_recorder=vr)
   ...: r1 = visitor.walk_expression(e1)

In [2]: r2 = visitor.walk_expression(e2)

In [3]: print(r2)
LinearRepn(mult=1, const=0, linear={140176910789184: 1}, nonlinear=- (1/x[2]))
```

Here, `e2` should be linear because `x[2]` is fixed, but the `LinearRepnVisitor` says it is nonlinear. The problem is that the `TemplateVarRecorder` inserts variables in bulk based on the parent component. So, when the `LinearRepnVisitor` is used to walk `e1`, `x[2]` gets added to the `var_map` even though it is fixed. Then, `LinearBeforeChildDispatcher._before_var` does not behave as expected. With the variable fixed, we want to hit [this](https://github.com/Pyomo/pyomo/blob/87f383ecc381b66de84615cb9a5c0c0d9dbfe911/pyomo/repn/linear.py#L662) line. However, because `x[2]` is in the `var_map`, we instead continue to [this](https://github.com/Pyomo/pyomo/blob/87f383ecc381b66de84615cb9a5c0c0d9dbfe911/pyomo/repn/linear.py#L664) line, and the visitor thinks the variable is not fixed.

Note that this does not happen if we create the `LinearRepnVisitor` without the `var_recorder`.

Pyomo version: just pulled main

cjck944084735-dot:
## Summary/Motivation:

`LinearRepnVisitor.record_monomial` only applied the fixed-variable (constant) handling when the variable had **not yet** been registered in the visitor's `var_map`. When a variable was registered in bulk by the `TemplateVarRecorder` while it was still free and was fixed afterwards, subsequent `walk_expression` calls found it in the `var_map` and treated it as an unrestricted variable: fixed values were not substituted and expressions involving them were misclassified as nonlinear. This is exactly the failure mode reported in #3851, including through the `LinearStandardFormCompiler`'s use of a `TemplateVarRecorder`.

- `record_monomial` now checks `var.fixed` **before** the `var_map` membership test, so a fixed variable always contributes its value as a constant, regardless of when it was registered.
- Two regression tests in `pyomo/repn/tests/test_linear.py::TestFixedVarAfterRegistration`, using the reporter's scenario (a variable registered by walking one expression while free, fixed, then used in another expression):
  - `test_fixed_var_after_registration`: `x[1] - 1/x[2]` with `x[2].fix(2)` now yields `constant=-0.5, linear={x[1]: 1}, nonlinear=None`;
  - `test_fixed_var_substituted_before_registration`: the fixed variable never becomes a linear term nor leaks into the nonlinear remainder.

After the change, the reporter's `r2` is `LinearRepn(mult=1, const=-0.5, linear={x[1]: 1}, nonlinear=None)` instead of carrying `nonlinear=-(1/x[2])`.

## Local verification

- `pyomo/repn/tests/`: 357 passed (355 existing + 2 new), 48 skipped.
- Broader `pyomo/core/tests/`: 4392 passed — no regressions.
- `black --check` clean on the changed files.

## AI-Use Disclosure

or

        - [x] Added and ran tests

 **Notes for reviewers (optional):** The behavior change is narrow: a fixed variable that was registered in `var_map` while free is now handled as a constant. Callers that relied on such variables appearing as linear terms would have received incorrect coefficients (the fixed value was never substituted), so this should be strictly a correctness improvement. One thing I would appreciate a second look at: whether any writer depends on fixed-but-registered variables still occupying a column in `var_map`/`var_list`.

### Legal Acknowledgement


jsiirola:
Thank you for this PR!  While I agree that it resolves the observed behavior, I am not sure that this is the best implementation.  In particular:

- this is in the core, and your change could affect the performance of the LinearRepn / LP writers.  Please document the performance impacts of this PR in the PR description / comments.  At a minimum, this will involve using the driver script (`scripts/performance/main.py`) to run the performance tests in the pyomo-model-libraries repo and then pasting the summary generated by `scripts/performance/compare.py` to compare the results from main and this branch on a couple versions of Python (generally at least for 3.10 and 3.14)
- I don't think this is the correct place to make this change: the TemplateVarRecorder is designed to be used by the LinearTemplateRepnVisitor, and you are making the change in LinearRepnVisitor, which makes different assumptions on when the `fixed` flag is checked.  I think the correct place to check this is in the LinearTemplateRepnVisitor, and because of the desire to work on templates, the correct solution might be to actually perform the check somewhere else.  There is a tension between wanting to only process the template once (which is structural and generally cannot know the fixed flag) and correctly removing constants to reduce/simplify higher-order terms.  This is likely to require a more detailed design discussion.

